PR #2884, opened 2015-10-07
Fix cli base _handle_response

elyezer:
As stderr can contain unicode characters, string format should expected
that.

PR #2892, opened 2015-10-08
Support unicode in exception messages for cli factory

abalakh:
Related to #2884 
`stderr` may contain unicode, so exception messages should be unicode as well.


sthirugn:
ACK

elyezer:
ACK

PR #2934, opened 2015-10-23
Implement CRUD tests for CLI host parameter

### LINKED ISSUE #2933: Implement CRUD tests for CLI host parameter commands
Tests for `hammer host set-parameter` and `hammer host delete-parameter` are missing. As we had BZ for this functionality risen recently, it would be good to have the tests implemented.

abalakh:
Closes #2933
Tests results:
```
nosetests tests/foreman/cli/test_host.py:HostParameterTests
........
Ran 8 tests in 321.766s

OK
```

### DIFF
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -1,8 +1,12 @@
+# pylint: disable=invalid-name
 """CLI tests for ``hammer host``."""
+from fauxfactory import gen_string
 from nailgun import entities
+from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.host import Host
 from robottelo.cli.proxy import Proxy
 from robottelo.config import settings
+from robottelo.helpers import invalid_values_list, valid_data_list
 from robottelo.test import CLITestCase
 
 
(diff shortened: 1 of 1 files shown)


elyezer:
ACK

tests/foreman/cli/test_host.py
```diff
@@ -1,5 +1,8 @@
 """CLI tests for ``hammer host``."""
+from fauxfactory import gen_string
 from nailgun import entities
+from robottelo.helpers import invalid_values_list, valid_data_list
```
sthirugn: move this 4 lines down

tests/foreman/cli/test_host.py
```diff
@@ -79,3 +82,205 @@ def test_create_libvirt_without_mac(self):
...
+        """Create host to tests parameters for"""
+        super(HostParameterTests, cls).setUpClass()
+        cls.puppet_proxy = Proxy.list()[0]
+        cls.host = entities.Host()
+        cls.host.create_missing()
+        cls.host = Host.create({
+            u'architecture-id': cls.host.architecture.id,
+            u'domain-id': cls.host.domain.id,
+            u'environment-id': cls.host.environment.id,
+            u'location-id': cls.host.location.id,  # pylint:disable=no-member
```
sthirugn: why does this specific line requires pylint silence but not others?

tests/foreman/cli/test_host.py
```diff
@@ -79,3 +82,205 @@ def test_create_libvirt_without_mac(self):
...
+            u'mac': cls.host.mac,
+            u'medium-id': cls.host.medium.id,
+            u'name': cls.host.name,
+            u'operatingsystem-id': cls.host.operatingsystem.id,
+            # pylint:disable=E1101
```
sthirugn: Please use text instead of E1101 which is easier to read

tests/foreman/cli/test_host.py
```diff
@@ -79,3 +82,205 @@ def test_create_libvirt_without_mac(self):
...
+
+
+class HostParameterTests(CLITestCase):
+    """Tests targeting host parameters"""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create host to tests parameters for"""
+        super(HostParameterTests, cls).setUpClass()
+        cls.puppet_proxy = Proxy.list()[0]
```
omaciel: Can I suggest that we add the following assertion to make sure that the test can handle the possible situation where there's no default capsule (aka 'proxy')?

```python
assert len(Proxy.list()) > 0
```

If this were to fail without an assertion, then all tests would fail and it would be hard to determine what caused the failure. Jenkins would report that `setUp` failed with the traceback, but with an assertion we would be able to know right away what caused it.

tests/foreman/cli/test_host.py
```diff
@@ -79,3 +82,205 @@ def test_create_libvirt_without_mac(self):
...
+                name = name.lower()
+                Host.set_parameter({
+                    'host-id': self.host['id'],
+                    'name': name,
+                    'value': gen_string('alphanumeric'),
+                })
+                self.host = Host.info({'id': self.host['id']})
+                self.assertIn(name, self.host['parameters'].keys())
+
+    def test_add_parameter_diff_values(self):
```
sthirugn: test_add_parameter_by_host_id? to make it consistent with next test.

tests/foreman/cli/test_host.py
```diff
@@ -79,3 +82,205 @@ def test_create_libvirt_without_mac(self):
...
+        Host.set_parameter({
+            'host': self.host['name'],
+            'name': name,
+            'value': value,
+        })
+        self.host = Host.info({'id': self.host['id']})
+        self.assertIn(name, self.host['parameters'].keys())
+        self.assertEqual(value, self.host['parameters'][name])
+
+    def test_update_parameter_host_id(self):
```
sthirugn: test_update_parameter_by_host_id?

tests/foreman/cli/test_host.py
```diff
@@ -79,3 +82,205 @@ def test_create_libvirt_without_mac(self):
...
+            with self.subTest(new_value):
+                Host.set_parameter({
+                    'host-id': self.host['id'],
+                    'name': name,
+                    'value': new_value,
+                })
+                self.host = Host.info({'id': self.host['id']})
+                self.assertIn(name, self.host['parameters'].keys())
+                self.assertEqual(new_value, self.host['parameters'][name])
+                self.assertNotEqual(old_value, self.host['parameters'][name])
```
sthirugn: The above two lines are redundant :) please remove one.

tests/foreman/cli/test_host.py
```diff
@@ -79,3 +82,205 @@ def test_create_libvirt_without_mac(self):
...
+
+    def test_update_parameter_host_name(self):
```
sthirugn: test_update_parameter_by_host_name?

tests/foreman/cli/test_host.py
```diff
@@ -1,5 +1,8 @@
 """CLI tests for ``hammer host``."""
```
sthirugn: I am pretty sure some test names are larger in length to what pylint expects.  In that case implement this to silence it for the module:
https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/ui/test_ldapauthsource.py#L1

tests/foreman/cli/test_host.py
```diff
@@ -79,3 +82,205 @@ def test_create_libvirt_without_mac(self):
...
+            with self.subTest(new_value):
+                Host.set_parameter({
+                    'host': self.host['name'],
+                    'name': name,
+                    'value': new_value,
+                })
+                self.host = Host.info({'id': self.host['id']})
+                self.assertIn(name, self.host['parameters'].keys())
+                self.assertEqual(new_value, self.host['parameters'][name])
+                self.assertNotEqual(old_value, self.host['parameters'][name])
```
sthirugn: same comment as above

tests/foreman/cli/test_host.py
```diff
@@ -79,3 +82,205 @@ def test_create_libvirt_without_mac(self):
...
+
+    def test_delete_parameter_host_id(self):
```
sthirugn: test_delete_paramter_by_host_id?

tests/foreman/cli/test_host.py
```diff
@@ -79,3 +82,205 @@ def test_create_libvirt_without_mac(self):
...
+            'value': gen_string('alphanumeric'),
+        })
+        Host.delete_parameter({
+            'host-id': self.host['id'],
+            'name': name,
+        })
+        self.host = Host.info({'id': self.host['id']})
+        self.assertNotIn(name, self.host['parameters'].keys())
+
+    def test_delete_parameter_host_name(self):
```
sthirugn: test_delete_parameter_by_host_name?

sthirugn: I am curious why none of the delete tests are datadriven (subTest)?

sthirugn:
Pending comments

tests/foreman/cli/test_host.py
abalakh: nope, i'm using pylint too :) all the names here are ok, no need to silence pylint (at least for now)

tests/foreman/cli/test_host.py
abalakh: and then pylint will start complaining. so either i leave short names, or silence pylint. your suggestion?

tests/foreman/cli/test_host.py
abalakh: :+1:

tests/foreman/cli/test_host.py
abalakh: Take a look at `nailgun.entities.Host.create_missing()`. All the others objects like `domain`, `environment` etc. are initialized there, but not `location` or `organization`. `create_missing()` is using existing location and organization; and as both are required fields, nailgun will create them if they were not provided.
And code analyzers are complaining as no member `location` or `organization` was directly initialized in `entities.Host` class.

tests/foreman/cli/test_host.py
abalakh: :+1:

tests/foreman/cli/test_host.py
abalakh: :+1:

tests/foreman/cli/test_host.py
abalakh: but right now it's more consistent with previous test - previous checked different names, this one checks different values.

tests/foreman/cli/test_host.py
abalakh: :open_mouth:  no idea why i put it there. of course it's redundant, thanks for catching it.

tests/foreman/cli/test_host.py
abalakh: no need to check it for `...delete_by_id` test, but for `...delete_by_name` it's good suggestion, will update the test.

abalakh:
Updated with all the comments except test naming ones. As original thread became hidden because of oudated code, will post my question to @sthirugn here:
Right now pylint is ok with these names. I can either leave them as they are or update the names and disable pylint naming checks. What do you prefer?

abalakh:
```
nosetests tests/foreman/cli/test_host.py:HostParameterTests
........
Ran 8 tests in 383.048s

OK
```

tests/foreman/cli/test_host.py
sthirugn: :+1:

tests/foreman/cli/test_host.py
sthirugn: @abalakh any reason why you cant just use cls.host = entities.Host().create()?

tests/foreman/cli/test_host.py
sthirugn: :+1:

tests/foreman/cli/test_host.py
sthirugn: in my experience, many folks like descriptive names, so silencing pylint for invalid-name is the way to go.

sthirugn:
Pending comments/questions.

tests/foreman/cli/test_host.py
abalakh: It's CLI test, even more, CLI test from file `test_host.py`. IMO It's not ok to create host through API here :) API is used here only to create dependencies.

sthirugn: @abalakh got it now. To make it more clear, can you please make a note that you are using nailgun here to construct dummy host arguments which is eventually used to make the Host using cli. This will really help others folks reading this code.

abalakh:
Comments addressed.

tests/foreman/cli/test_host.py
```diff
@@ -79,3 +83,211 @@ def test_create_libvirt_without_mac(self):
...
+    """Tests targeting host parameters"""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create host to tests parameters for"""
+        super(HostParameterTests, cls).setUpClass()
+        cls.proxies = Proxy.list()
+        assert len(cls.proxies) > 0
+        cls.puppet_proxy = cls.proxies[0]
+        # using nailgun to create dependencies
```
sthirugn: :+1:

tests/foreman/cli/test_host.py
sthirugn: :+1:

PR #2935, opened 2015-10-23
Automate BZ1184442 for CLI

### LINKED ISSUE #2911: Please automate BZ 1184442
Please automate Bugzilla issue https://bugzilla.redhat.com/show_bug.cgi?id=1184442 using the CLI to verify that the content has been deleted and the disk usage has decreased.

abalakh:
Added `test_bz1184442` to verify deleting content-view actually deletes files from hard drive.
Closes #2911 
Test results:
```
nosetests tests/foreman/cli/test_contentviews.py -m test_bz1184442
.
Ran 1 test in 74.263s

OK
```

### DIFF
--- a/tests/foreman/cli/test_contentviews.py
+++ b/tests/foreman/cli/test_contentviews.py
@@ -3,6 +3,7 @@
 import random
 
 from fauxfactory import gen_alphanumeric, gen_string
+from robottelo import manifests, ssh
 from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.contentview import ContentView
 from robottelo.cli.factory import (
@@ -16,7 +17,6 @@ from robottelo.cli.factory import (
     make_repository,
     make_user,
 )
-from robottelo import manifests
 from robottelo.cli.repository import Repository
 from robottelo.cli.repository_set import RepositorySet
 from robottelo.cli.puppetmodule import PuppetModule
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_contentviews.py
```diff
@@ -142,6 +142,46 @@ def setUp(self):
                 {u'organization-id': TestContentView.org['id']},
                 cached=True)
 
+    def test_bz1184442(self):
```
elyezer: This should be skipped while that bug is open?

elyezer:
ACK, pending question.

tests/foreman/cli/test_contentviews.py
abalakh: [BZ1184442](https://bugzilla.redhat.com/show_bug.cgi?id=1184442) is already closed.

omaciel:
ACK

tests/foreman/cli/test_contentviews.py
elyezer: Then I think is better to change the test name. And need to remove `@bz` from the docstring since it is not blocking the test.

sthirugn: good point :+1:

tests/foreman/cli/test_contentviews.py
```diff
@@ -142,6 +142,46 @@ def setUp(self):
...
+        """@Test: Delete content view and verify it was actually deleted from
+        hard drive.
+
+        @Feature: Content View
+
+        @Assert: Content view was deleted and pulp folder doesn't contain
+        content view files anymore
+
+        @BZ: 1184442
```
sthirugn: like @elyezer mentioned, we add BZ number here only if it is blocked by the BZ.  May be you can add a note inside the test saying that this test is originally written to test BZ 1184442

tests/foreman/cli/test_contentviews.py
```diff
@@ -142,6 +142,46 @@ def setUp(self):
...
+        # Create a content view and publish it
+        content_view = make_content_view({'organization-id': self.org['id']})
+        ContentView.publish({u'id': content_view['id']})
+        content_view = ContentView.info({u'id': content_view['id']})
+        # Check content view files presence before deletion
+        result = ssh.command(
+            'find /var/lib/pulp -name "*{0}*"'.format(content_view['name']))
+        self.assertEqual(result.return_code, 0)
+        self.assertNotEqual(len(result.stdout), 0)
+        # Completely delete the content view
```
sthirugn: Move this comment one line down

tests/foreman/cli/test_contentviews.py
```diff
@@ -142,6 +142,46 @@ def setUp(self):
...
+            'content-view': content_view['name'],
+            'organization': self.org['name'],
+            'version': content_view['versions'][0]['version'],
+        })
+        ContentView.delete({'id': content_view['id']})
+        # Check content view files presence after deletion
+        result = ssh.command(
+            'find /var/lib/pulp -name "*{0}*"'.format(content_view['name']))
+        self.assertEqual(result.return_code, 0)
+        self.assertEqual(len(result.stdout), 0)
```
sthirugn: You are testing the same asserts before and after the deletion.  Please help me understand how this assert makes sure that the disk space is reclaimed (or) the content is actually deleted (as per the bz description)

sthirugn:
Pending questions/comments

tests/foreman/cli/test_contentviews.py
omaciel: @sthirugn they are not the same. First batch of assertions checks that:
* Return code is '0'
* Something got returned (len > 0)
* Version is 1

then,

* Return code is '0'
* Nothing is returned (len == 0)

sthirugn: @omaciel you are right, thanks

abalakh: Not the same. Before deletion it's `self.assertNotEqual`, after - `self.assertEqual`. 
It's bad idea to check the disk space as numerous tasks may run in background and hard disk space may be consumed by something else. Locally 1 of 10 runs was failing because of this.
Then I came up with another idea - each published content view stores some files in `/var/lib/pulp`, and all the related files can be easily found by content view name in the filenames. So I'm verifyng that content view files are actually present before deletion, and they actually disappear after deletion. IMO it's even better to check this way, as we're 100% sure that files of the specific content view were deleted.

abalakh:
Updated & rebased.

tests/foreman/cli/test_contentviews.py
```diff
@@ -250,6 +250,44 @@ def test_cv_delete(self):
...
+        hard drive.
+
+        @Feature: Content View
+
+        @Assert: Content view was deleted and pulp folder doesn't contain
+        content view files anymore
+
+        """
+        # Create a content view and publish it
+        content_view = make_content_view({'organization-id': self.org['id']})
```
sthirugn: Will this not create an empty content view? It will be great to add a real repo for the test to make more sense?

omaciel: @sthirugn it does. Perhaps add a second test that adds a repo, synchronizes it, adds it to the content view, etc, etc?

sthirugn: Not sure if there is a real need for second test (as this test is a dummy test without repo IMO).  I vote for updating this test with real repo, synchronize, add it to content view before publishing.

sthirugn: I learnt that empty content view creates few files as well so yeah please create second test for my suggestion.

abalakh:
```
nosetests tests/foreman/cli/test_contentviews.py -m test_delete_cv_files
.
Ran 1 test in 71.425s

OK
nosetests tests/foreman/cli/test_contentviews.py -m test_delete_cv_custom_repo_files
S
Ran 1 test in 25.796s

OK (SKIP=1)
```

omaciel:
ACK, round TWO

tests/foreman/cli/test_contentviews.py
```diff
@@ -250,6 +250,94 @@ def test_cv_delete(self):
         with self.assertRaises(CLIReturnCodeError):
             ContentView.info({'id': con_view['id']})
 
+    def test_delete_cv_files(self):
+        """@Test: Delete content view and verify it was actually deleted from
```
sthirugn: to make the test description more clear: 
s/Delete content view/Delete empty content view

sthirugn:
Pending comment

abalakh:
Comment addressed.

sthirugn:
ACK

omaciel:
Karma given

PR #2977, opened 2015-11-03
UI: Update user search according to #2930

abalakh:
Part of #2930 
```
.SSSSSSSSSSSSSSSS.....SSSSSSSSSSSSSSSSSSSSSSSSSSSSS.S....ES.....SSSSSSSSSSSSSSSSSSSSSSSSSSSSFSSSSS..
ERROR: @Test: Create User for all variations of Username
FAIL: @Test: Update Username in User
Ran 100 tests in 1406.566s

FAILED (SKIP=80, errors=1, failures=1)
```

### DIFF
--- a/robottelo/ui/user.py
+++ b/robottelo/ui/user.py
@@ -91,11 +91,18 @@ class User(Base):
                 )
         self.click(common_locators['submit'])
 
-    def search(self, name, search_key):
+    def search(self, element_name, element_locator=None, search_key=None,
+               katello=None, button_timeout=15, result_timeout=15):
         """Searches existing user from UI."""
         Navigator(self.browser).go_to_users()
-        return self.search_entity(
-            name, locators['users.user'], search_key=search_key)
+        return super(User, self).search(
+            element_name,
+            locators['users.user'] or element_locator,
+            search_key=search_key,
+            katello=katello,
+            button_timeout=button_timeout,
+            result_timeout=result_timeout,
+        )
 
     def delete(self, name, search_key, really=True):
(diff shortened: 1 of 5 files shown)


abalakh:
`[DONOTMERGE]` tag while I'm figuring out what's wrong with 2 failed tests.

abalakh:
Ok, the first test failed as I was working while running the tests. Re-run multiple times without switching to another window - it's passing:
```
nosetests tests/foreman/ui/test_user.py -m test_positive_create_user_different_usernames
.
Ran 1 test in 107.849s

OK
```
Regarding the second one - I guess there's a bug in test and I already have fix for that. Jenkins is executing UI tests right now, and if it shows the bug is reproducing not just for me, then I'll create a PR for `master` branch later today. As mentioned failure is not related to the current PR, removing `[DONOTMERGE]` tag.

robottelo/ui/user.py
```diff
@@ -91,11 +91,18 @@ def create(self, username=None, email=None, password1=None,
                 )
         self.click(common_locators['submit'])
 
-    def search(self, name, search_key):
+    def search(self, element_name, element_locator=None, search_key=None,
```
kbidarkar: As we are re-factoring things, how about we set search_key='login' by default here?

We can just get rid of handling this stuff for all the tests. thoughts?

sthirugn: :+1: to @kbidarkar

abalakh:
Closing as we changed our approach with #2984

PR #2982, opened 2015-11-04
UI: Fix test_positive_update_user_username

abalakh:
The test was designed to be run as admin user.
In the end of each sub-test we're verifying changes actually
applied by logging in as our custom user.
But logging out was missing, causing test to fail.
Resolved this by starting a new session for each subtest.
Test results:
```
nosetests tests/foreman/ui/test_user.py -m test_positive_update_user_username
.
Ran 1 test in 217.426s

OK
```

### DIFF
--- a/tests/foreman/ui/test_user.py
+++ b/tests/foreman/ui/test_user.py
@@ -853,8 +853,9 @@ class User(UITestCase):
                 edit=True,
                 roles=['Site'],
             )
-            for new_username in valid_strings():
-                with self.subTest(new_username):
+        for new_username in valid_strings():
+            with self.subTest(new_username):
+                with Session(self.browser):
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_user.py
```diff
@@ -853,8 +853,9 @@ def test_positive_update_user_username(self):
                 edit=True,
                 roles=['Site'],
             )
-            for new_username in valid_strings():
-                with self.subTest(new_username):
+        for new_username in valid_strings():
+            with self.subTest(new_username):
+                with Session(self.browser):
```
sthirugn: Why do you need a new Session? It is already there in 846.

sthirugn:
ACK

elyezer:
ACK

PR #3022, opened 2015-11-11
Updated `valid_labels_list` to return 128 characters.

omaciel:
Our automation has shown that when creating new Products and providing `labels` via the `valid_labels_list` heper, we get an error claiming that 'Label cannot contain more than 128 characters'.

Since this helper seems to be used only by one test, I have updated it to return 128 character strings.

### DIFF
--- a/robottelo/datafactory.py
+++ b/robottelo/datafactory.py
@@ -184,8 +184,8 @@ def valid_environments_list():
 def valid_labels_list():
     """Generates a list of valid labels."""
     return [
-        gen_string('alphanumeric', random.randint(1, 255)),
-        gen_string('alpha', random.randint(1, 255)),
+        gen_string('alphanumeric', random.randint(1, 128)),
+        gen_string('alpha', random.randint(1, 128)),
     ]
 
 


omaciel:
```bash
(robottelo)┌[omaciel☮Ogs-MacBook-Pro.local]-(~/hacking/robottelo)-[git://test-positive-create-2 ✔]-
└> nosetests -v tests/foreman/cli/test_product.py:TestProduct.test_positive_create_2
@Test: Check if product can be created with random labels ... ok

Ran 1 test in 21.983s

OK
```

elyezer:
`valid_labels_list` is being used by other entities. I think is better to add an optional argument length that defaults to 255 and override that for products.

sthirugn:
ACK

sthirugn:
@elyezer as mentioned by @omaciel I see only one test using it so far.

omaciel:
@elyezer 

```bash
git grep -i 'valid_labels_list'
robottelo/datafactory.py:def valid_labels_list():
tests/foreman/cli/test_product.py:    valid_labels_list,
tests/foreman/cli/test_product.py:        for label in valid_labels_list():
tests/robottelo/test_datafactory.py:    valid_labels_list,
tests/robottelo/test_datafactory.py:        self.assertEqual(len(valid_labels_list()), 2)
tests/robottelo/test_datafactory.py:        self.assertEqual(len(valid_labels_list()), 1)
tests/robottelo/test_datafactory.py:        7. :meth:`robottelo.datafactory.valid_labels_list`
tests/robottelo/test_datafactory.py:                valid_labels_list(),
```

elyezer:
ACK

Issue #3021, opened 2015-11-11
UI: Add option to specify firefox path


lpramuk:
With this option one can easily switch between various firefox versions.
(when debugging frameworks issues etc...)
This option will be configured in robotello.properties

sthirugn:
:+1:

lpramuk:
PR #3036 merged. Issue closed

PR #3036, opened 2015-11-12
UI: config option to specify firefox binary location

lpramuk:
UI: config option to specify firefox binary location

(Issue #3021)

### DIFF
--- a/robottelo/test.py
+++ b/robottelo/test.py
@@ -131,6 +131,7 @@ class UITestCase(TestCase):
         cls.katello_user = settings.server.admin_username
         cls.katello_passwd = settings.server.admin_password
         cls.driver_name = settings.webdriver
+        cls.driver_binary = settings.webdriver_binary
         cls.locale = settings.locale
         cls.server_name = settings.server.hostname
         cls.screenshots_dir = settings.screenshots_path
@@ -170,11 +171,20 @@ class UITestCase(TestCase):
     def setUp(self):  # noqa
         """We do want a new browser instance for every test."""
         if self.driver_name.lower() == 'firefox':
-            self.browser = webdriver.Firefox()
+            self.browser = webdriver.Firefox(
+                firefox_binary=webdriver.firefox.firefox_binary.FirefoxBinary(
+                    self.driver_binary)
+            )
         elif self.driver_name.lower() == 'chrome':
-            self.browser = webdriver.Chrome()
+            self.browser = (
+                webdriver.Chrome() if self.driver_binary is None
+                else webdriver.Chrome(executable_path=self.driver_binary)
+            )
         elif self.driver_name.lower() == 'ie':
-            self.browser = webdriver.Ie()
+            self.browser = (
+                webdriver.Ie() if self.driver_binary is None
+                else webdriver.Ie(executable_path=self.driver_binary)
+            )
         elif self.driver_name.lower() == 'phantomjs':
             service_args = ['--ignore-ssl-errors=true']
             self.browser = webdriver.PhantomJS(
--- a/robottelo.properties.sample
+++ b/robottelo.properties.sample
@@ -40,6 +40,9 @@ ssh_key=
 # verbosity=debug
 # Selenium webdriver to use
 # webdriver=firefox
+# Binary location for selected wedriver
+# webdriver_binary=/usr/bin/firefox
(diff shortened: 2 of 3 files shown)


robottelo/test.py
```diff
@@ -170,7 +171,10 @@ def setUpClass(cls):  # noqa
     def setUp(self):  # noqa
         """We do want a new browser instance for every test."""
         if self.driver_name.lower() == 'firefox':
-            self.browser = webdriver.Firefox()
+            self.browser = webdriver.Firefox(
+                firefox_binary=webdriver.firefox.firefox_binary.FirefoxBinary(
+                    self.driver_binary
```
sthirugn: what happens if I dont specify `webdriver_binary` in properties file?  Will this line fail? If so, can you use a default value so it wont fail?

lpramuk: No, it wont fail. FirefoxBinary() init is robust enough to handle None. There is logic within for this. By omitting webdriver_binary you stay on the default

sthirugn: :+1:

sthirugn:
ACK

robottelo/test.py
```diff
@@ -170,7 +171,10 @@ def setUpClass(cls):  # noqa
...
+                ))
         elif self.driver_name.lower() == 'chrome':
             self.browser = webdriver.Chrome()
```
elyezer: The driver_binary don't work for Chrome too? If yes it is better to add it here using the Chrome way.

lpramuk:
@elyezer:  Extended option to specify binary location for other webdrivers too (chrome, ie)

lpramuk:
@elyezer: I may prefer being ternary here:
```
        elif self.driver_name.lower() == 'chrome':
            self.browser = webdriver.Chrome() if self.driver_binary is None \
                else webdriver.Chrome(executable_path=self.driver_binary)
```
not to rely on hardcoded default param value in _init_():
```
        elif self.driver_name.lower() == 'chrome':
            if self.driver_binary is None:
                self.driver_binary = 'chromedriver'
            self.browser = webdriver.Chrome(executable_path=self.driver_binary)
```
WYT?

lpramuk:
I used ternary solution as it is better.

robottelo/test.py
```diff
@@ -170,11 +171,16 @@ def setUpClass(cls):  # noqa
...
         """We do want a new browser instance for every test."""
         if self.driver_name.lower() == 'firefox':
-            self.browser = webdriver.Firefox()
+            self.browser = webdriver.Firefox(
+                firefox_binary=webdriver.firefox.firefox_binary.FirefoxBinary(
+                    self.driver_binary
+                ))
         elif self.driver_name.lower() == 'chrome':
-            self.browser = webdriver.Chrome()
+            self.browser = webdriver.Chrome() if self.driver_binary is None \
```
elyezer: `\` should not be used.

And having an if/else block is more readable than a ternary operator in this case because it needs more than one line.

robottelo/test.py
```diff
@@ -170,11 +171,16 @@ def setUpClass(cls):  # noqa
...
+                else webdriver.Chrome(executable_path=self.driver_binary)
         elif self.driver_name.lower() == 'ie':
-            self.browser = webdriver.Ie()
+            self.browser = webdriver.Ie() if self.driver_binary is None \
```
elyezer: Same as above.

elyezer:
ACK, pending comments.

robottelo/test.py
lpramuk: RE \ should not be used - but flake8 was silent about it?

Wrapping ternary into ( ) looks really ugly, hmm I better use if/else block...

elyezer: It is a valid python way to do multiple lines, but it causes issues related to the end of the line depending on the OS you are running. To be safe, is better to avoid its usage.

lpramuk: I can't help myself I stuck with ternary wrapped in ( ). Thanks Elyezer.
if/else blocked looked even uglier, due to additional line wrapping :-)

lpramuk:
Comments resolved, I've stuck with 2 line ternary operator wrapped in (), it looks more consistent with code above

elyezer:
ACK

sthirugn:
:+1:  ACK

Issue #3063, opened 2015-11-23
Changes required to fix discovery tests as per new infra.


sghai:
I was trying to fix setup related to host provisioning and host discovery. But looks like now decent changes are required as per new infra. Also, earlier we were creating host on Satellite itself by enabling nested virt. It means we were using Satellite server as hypervisor. But now plan is to use some other external compute-resource. 

I'm planning to use external libvirt. 

what all changes are required ?

1. Like provisioning server, we may need to define the libvirt_server setting along with image_dir in robottelo properties file as well as in jenkins.

2. Since external libvirt host is accessible via ssh. So we need to configure ssh key-pairs between Satellite server and hypervisor.

3. For host_discovery, the first variable we need is 'bridge_interface' . This is based on the different vlan configurations. Currently we have four bridges based on different distros.
I'm not sure best place to fetch/configure this but as per discussion with @kbidarkar, I think we can fetch this with hostname. Please let us know if anyone has any other way to get that ?

4. we may also required to fetch Gateway/Subnet etc depending on selected bridge interface.

Note: I tested manually host discovery by configuring external libvirt server along with our Satellite automation rhel67 server.

kbidarkar:
@sghai  will be working on 1) and 2) 

I will be working on 3) and 4)

sghai:
Thank you @kbidarkar for picking 3) and 4).

sghai:
1) and 2) are covered in  #3064 and https://github.com/SatelliteQE/automation-tools/issues/264

sghai:
This issue is well covered and implemented. Thank you @kbidarkar for helping out to populate Vlan networking stuff. It was a great help.

Closing this issue.

PR #3064, opened 2015-11-23
Added params for libvirt_host

sghai:
This is required to cover step 1) in issue #3063

### DIFF
--- a/robottelo.properties.sample
+++ b/robottelo.properties.sample
@@ -188,3 +188,12 @@ ssh_key=
 # Parameter for deciding whether conduct initial sync or resync
 # 'resync' denotes resync; 'sync' denotes initial sync
 # sync_type='sync'
+
+# [compute_resources]
+# External Libvirt Hostname
+# libvirt_hostname=
+
+# Path on the Libvirt host, where the virtual images will be stored.
+# The default libvirt image dir will be used
+# "/var/lib/libvirt/images/".
+# libvirt_image_dir=/var/lib/libvirt/images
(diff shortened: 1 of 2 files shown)


robottelo.properties.sample
```diff
@@ -188,3 +188,12 @@ ssh_key=
 # Parameter for deciding whether conduct initial sync or resync
 # 'resync' denotes resync; 'sync' denotes initial sync
 # sync_type='sync'
+
+# [libvirt_host]
```
kbidarkar: How about we making this [compute_resources] ?
I am not asking to change this, but rather so we discuss and think about it, if it is OK ?

Thinking In future all the other(vmware, rhevm, openstack) CR details also could go below this.

sghai: +1
Updating now.. anyway, I'm using `libvirt_` keyword.

sthirugn:
ACK

robottelo.properties.sample
kbidarkar: :+1:  for `libvirt_` keyword.

elyezer: The main point here is, when we have more than one compute resource, like openstack and we just want to use libvirt, having all under compute_resources is good in terms of organization, but we may find problems with validation, like, in the example mentioned, don't knowing what should be required the libvirt_hostname and/or the openstack_hostname...

kbidarkar: ok, if using separate section helps isolate things then that's good. Ok, I agree separating things would quickly help us know what is required and what is not , for a particular CR.

Also sometimes certain CR's have multiple variables too.

elyezer:
ACK

PR #3084, opened 2015-11-26
Updating entities to use only alpha string.

kbidarkar:
a) We are testing for other strings, for these entities in it's own
   entity modules. Hence dropping it.
b) This will help troubleshoot oscap related testcase failure more
   easier with alpha strings for navigating to these entities as
   and when required.
c) Please note that oscap sits on top of the entire stack and hence
   just using alpha strings would help here immensely.

### DIFF
--- a/tests/foreman/longrun/test_oscap.py
+++ b/tests/foreman/longrun/test_oscap.py
@@ -50,9 +50,12 @@ class OpenScap(UITestCase):
         ]
 
         # step 1: Create new organization and environment.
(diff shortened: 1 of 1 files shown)


sghai:
ACK
+1 to alpha strings.. and probably irrelevant to this PR. But I would suggest we should only use alpha strings in `tier3` tests `(integration, end_2_end_tests)`

elyezer:
ACK

PR #3108, opened 2015-12-04
Add skip_if_not_set decorator

elyezer:
The decorator skip a test if some robottelo configuration is missing.
For example if client provisioning settings is not available then it
will skip the test decorated.

### DIFF
--- a/robottelo/decorators.py
+++ b/robottelo/decorators.py
@@ -40,6 +40,65 @@ _redmine = {
 }
 
 
+def skip_if_not_set(*options):
+    """Skips test if expected configuration is not set.
+
+    Decorating a method::
+
+        @skip_if_not_set('clients')
+        def test_something(self):
+            self.assertTrue(True)
+
+    Decorating an entire class::
+
+        class FeatureTestCase(robottelo.test.TestCase):
+
+            @skip_if_not_set('clients')
+            def setUp(self):
+                pass
+
+            def test_something(self):
+                self.assertTrue(True)
+
+    The above approach is required for decorating all test methods of a class
+    because the ``skip_if_not_set`` decorator is intended to run at runtime and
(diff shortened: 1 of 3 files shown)


omaciel:
ACK

robottelo/decorators.py
```diff
@@ -40,6 +40,65 @@
...
+            @skip_if_not_set('clients')
+            def setUp(self):
+                pass
+
+            def test_something(self):
+                self.assertTrue(True)
+
+    The above approach is required for decorating all test methods of a class
+    because the ``skip_if_not_set`` decorator is intended to run at runtime and
+    not import time. Decorating a class or the ``setUpClass`` method is not
```
sthirugn: :+1:

robottelo/decorators.py
```diff
@@ -40,6 +40,65 @@
...
+    if not options_set.issubset(settings.all_features):
+        invalid = options_set.difference(settings.all_features)
+        raise ValueError(
+            'Feature(s): "{0}" not found. Available ones are: "{1}"'
+            .format(
+                ', '.join(invalid),
+                ', '.join(settings.all_features)
+            )
+        )
+
```
sthirugn: Should Lines 74 - 83 be moved to wrapper() function instead?  I understand that program logic wise it is the same but I am thinking if it would make more sense to have this logic in the wrapper().

elyezer: No, if the code is placed there, then it will fail only when running. In this case is better to fail at import time since we don't want to generate a failure for miss typing for example.

sthirugn: ah ok, you want to fail the import of test module incase of incorrect spelling. makes sense.

sthirugn: This fails the test import of the module without running any tests. :+1:

sthirugn:
ACK

PR #3109, opened 2015-12-04
Cover robottelo.decorators with 100%

elyezer:
Add tests to ensure 100% coverage for tests.robottelo.test_decorators.

Bonus: Cover robottelo.datafactory with 100%.

### DIFF
--- a/robottelo/decorators.py
+++ b/robottelo/decorators.py
@@ -10,7 +10,7 @@ from functools import wraps
 from robottelo.config import settings
 from robottelo.constants import BZ_OPEN_STATUSES, NOT_IMPLEMENTED
 from six.moves.xmlrpc_client import Fault
-from xml.parsers.expat import ExpatError, errors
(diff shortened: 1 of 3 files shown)


robottelo/decorators.py
```diff
@@ -321,6 +322,12 @@ def __call__(self, func):
         :param func: The function being decorated.
 
         """
+        if self.bug_type not in ('bugzilla', 'redmine'):
```
elyezer: This will fail when importing the test module. Will avoid running the test until the expected bug_type is passed. Previously it was being checked just when running the test. In this case is better to fail earlier.

Ichimonji10: Good call on moving this.

elyezer: Glad you like it!

sthirugn:
ACK

omaciel:
ACK

PR #3166, opened 2015-12-16
Rename abrt tests

sthirugn:
Renamed as per issue #3083

### DIFF
--- a/tests/foreman/cli/test_abrt.py
+++ b/tests/foreman/cli/test_abrt.py
@@ -1,11 +1,13 @@
 """Test for abrt report"""
+from robottelo.decorators import stubbed
 from robottelo.test import CLITestCase
 
 
-class TestAbrt(CLITestCase):
+class AbrtTestCase(CLITestCase):
     """Test class for generating abrt report in CLI."""
 
-    def test_abrt_1(self):
+    @stubbed()
+    def test_positive_create_report(self):
         """@test: a crashed program and abrt reports are send
 
         @Feature: Abrt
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_abrt.py
```diff
@@ -24,7 +24,7 @@ def test_abrt_1(self):
 
         """
 
-    def test_abrt_2(self):
+    def test_positive_create_multiple_reports(self):
```
omaciel: Does this create a ``crash`` report as well as `test_positive_create_crash_report`? If so, perhaps we want to rename this to `test_positive_create_reports`?

tests/foreman/cli/test_abrt.py
```diff
@@ -76,7 +76,7 @@ def test_abrt_4(self):
 
         """
 
-    def test_abrt_5(self):
+    def test_positive_cli_report(self):
```
omaciel: What is the difference between this report generating method and the others? They're all using the CLI so this name doesn't really tell me what is going on. Should this be a `test_positive_search_report_by_id` or something similar?

tests/foreman/cli/test_abrt.py
sthirugn: The test case does not have enough information about it.  But you made a right point.  I am going to call these tests `test_positive_create_report` and `test_positive_create_reports` now.

sthirugn:
@omaciel pr updated

tests/foreman/cli/test_abrt.py
sthirugn: To be honest, I have no idea about it since the test cases are not implemented and have very less info now.

sthirugn: Actually I can update with the name you suggested.

omaciel:
ACK

PR #3168, opened 2015-12-16
Rename cli capsule tests as per issue #3083

sthirugn:


tests/foreman/cli/test_capsule.py
```diff
@@ -20,7 +20,7 @@ def setUp(self):
 
     @run_only_on('sat')
     @tier1
-    def test_verify_redmine_3875(self):
+    def test_positive_create_with_random_url(self):
```
omaciel: I'd remove `random` as we have many tests using random values but we don't explicitly call it out.

sthirugn: :+1:

tests/foreman/cli/test_capsule.py
```diff
@@ -210,7 +210,7 @@ def test_positive_consume_content_through_capsule(self):
         """
 
     @stubbed()
-    def test_positive_unsusbscribe_content_through_capsule(self):
+    def test_positive_unsusbscribe(self):
```
omaciel: Typo from before: s/unsusbscribe/unsubscribe

sthirugn:
Comments updated and rebased

omaciel:
ACK

PR #3174, opened 2015-12-17
Removed test_system_registration.py and moved its tests to test_contenthost.py

sthirugn:
1. Tests from test_system_registration.py are moved to test_contenthost.py
2. test_system_registration is removed


omaciel:
APT

PR #3189, opened 2015-12-21
Rename cli host collection tests as per issue #3083

sthirugn:


sghai:
ACK

tests/foreman/cli/test_host_collection.py
```diff
@@ -119,7 +119,7 @@ def test_positive_create_3(self):
 
     @skip_if_bug_open('bugzilla', 1214675)
     @tier1
-    def test_create_hc_with_unlimited_content_hosts(self):
+    def test_positive_create_with_unlimited_chosts(self):
```
abalakh: `unlimited-content-hosts` is parameter name, imo parts of parameter name shouldn't be shortened

abalakh:
ACK pending 1 comment

tests/foreman/cli/test_host_collection.py
sthirugn: @abalakh Interesting find.  Anyways I did not name it with parameter name in mind and it is not required too.  So I like to keep it as it is. Also I don't recall discussing parameter name in test case names when creating the naming conventions in #3083

PR #3199, opened 2015-12-22
Rename rhai tests as per issue #3083

sthirugn:


tests/foreman/rhai/test_rhai.py
```diff
@@ -80,7 +79,7 @@ def test_client_registration_to_rhai(self):
             vm.get('/var/log/redhat-access-insights/redhat-access'
                    '-insights.log', './insights_client_registration.log')
 
-    def test_org_selection_for_rhai(self):
+    def test_negative_org_not_selected(self):
```
jyejare: s/test_negative_org_not_selected/test_negative_org_selection ??
negative+not = postive :)

tests/foreman/rhai/test_rhai.py
```diff
@@ -103,14 +102,14 @@ def test_org_selection_for_rhai(self):
             self.assertIn("Organization Selection Required", result)
 
     @skip_if_not_set('clients')
-    def test_unregister_system_from_rhai(self):
+    def test_positive_unregister_client(self):
```
jyejare: s/test_positive_unregister_client/test_positive_unregister_client_to_rhat
i think 'rhai' word is very important here because we can register system to satellite/cdn/rhai. Good to add 'rhai' here.

jyejare:
ACK pending comments.

tests/foreman/rhai/test_rhai.py
sthirugn: I agree with you @jyejare.  But I think `test_negative_org_not_selected` more clearly explains the test in my opinion.

tests/foreman/rhai/test_rhai.py
sthirugn: Since this is a `rhai` test module, I thought it would be implied in the test.  But you are right.  I will add `rhai` back.

sthirugn:
Comments updated and rebased.

elyezer:
ACK

PR #3204, opened 2016-01-04
Support around pxe-less discovery and tests

sghai:
Changes includes:
- Support of pxe-less discovery along with pxe-based
- New tests based on both pxe-less and pxe-based.
- User can attach an interface to existing discovered host and see that newly added nic in discovered host via `refresh_fact` feature.
- User can validate custom facts defined while pxe-based as well as pxe-less discovery

Note: For Pxe-less discovery we need to re-build the iso with extra kernel options, so those ISO are already built and placed on external libvirt server. 

Default ISO name: `foreman-discovery-image-3.0.5-3.iso`
Downstream ISOs:
- foreman-discovery-image-3.0.5-3_unattended_down_el7.iso
- foreman-discovery-image-3.0.5-3_unattended_down_el6.iso
- foreman-discovery-image-3.0.5-3_unattended_iso_el6.iso
- foreman-discovery-image-3.0.5-3_unattended_iso_el7.iso

We are here just using iso names via `config/settings.py`.

PR #3210, opened 2016-01-07
Updated properties file for discovery_iso setting

sghai:


sghai:
We are using modified ISO for PXE_Less discovery as mentioned in PR #3204. ISOs are already re-built and placed on external_libvirt server. However user needs to set the ISO name in properties file.
Please note that ISO's are vlan's specific that means each ISO is rebuilt based on the GW and Server_name corresponding to specific vlan. So follow the convention like:

If you'r running test against `down_el6` vlan or `qe-sat6-rhel6` server then ISO name should be like `foreman-discovery-image-3.0.5-3_unattended_down_el6.iso`

sthirugn:
ACK

kbidarkar:
ACK

kbidarkar:
ACK the needed settings part exists in PR #3204.

PR #3212, opened 2016-01-07
increasing the wait time for unregister test

anarang:
I have increased the wait time for the unregister test to 60 sec as we discovered that the tests were faster on Jenkins than run locally and there was a timing error. A particular test passes when run locally but it fails on Jenkins, so I think adding a 60 sec wait time should be okay.

### DIFF
--- a/tests/foreman/rhai/test_rhai.py
+++ b/tests/foreman/rhai/test_rhai.py
@@ -135,10 +135,11 @@ class RHAITestCase(UITestCase):
                     locators['insights.unregister_button']
                 )
                 self.browser.refresh()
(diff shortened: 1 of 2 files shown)


anarang:
@elyezer to verify this on separately, I tried a few things on another branch here -> https://github.com/anarang/robottelo/commit/32e54a1e6e582dbf1c2dc2ccd4a35ba335a45143 and have some observations on another jenkins instance. I hope the wait time increase is justified.

sthirugn:
ACK

kbidarkar:
ACK

PR #3237, opened 2016-01-15
Added `skip_if_not_set` decorator for CLI Docker tests.

omaciel:
Make sure that Docker tests for CLI are skipped if properties file is not properly
configured to support Docker Compute Resources. The same already exists for API and
UI.


elyezer:
APT

sthirugn:
ACK pending test results - to show the required tests are skipped please.

omaciel:
@sthirugn tests won't be skipped using our automation but if you did not have the `[docker]` section enabled, then the tests would be skipped. You can take my word for it :)

sthirugn:
Thank you yeah I understand that tests will be skipped if the section is not provided in config.

PR #3279, opened 2016-02-02
Primary set of CLI stubs for Smart Class Parameter

jyejare:
This is primary set of CLI stubs for feature Smart Class Parameter.
And more to come in next PR.

CLI Usage:
```
# hammer -u <user> -p <passwd> sc-param --help
Usage:
    hammer sc-param [OPTIONS] SUBCOMMAND [ARG] ...

Parameters:
 SUBCOMMAND                    subcommand
 [ARG] ...                     subcommand arguments

Subcommands:
 add-override-value            Create an override value for a specific smart variable
 info                          Show a smart class parameter
 list                          List all smart class parameters
 remove-override-value         Delete an override value for a specific smart variable
 update                        Update a smart class parameter

```

### DIFF
--- a/tests/foreman/cli/test_classparameters.py
+++ b/tests/foreman/cli/test_classparameters.py
@@ -1,0 +1,827 @@
+# -*- encoding: utf-8 -*-
+"""Test class for Smart/Puppet Class Parameter"""
+
+from robottelo.decorators import run_only_on, stubbed, tier1, tier3
+from robottelo.test import CLITestCase
+
+
+class SmartClassParametersTestCase(CLITestCase):
+    """Implements Smart Class Parameter tests in CLI"""
+
+    @run_only_on('sat')
+    @stubbed()
+    @tier3
+    def test_positive_override_check(self):
+        """Override the Default Parameter value.
+
+        @feature: Class Parameters - Override
+
+        @steps:
+
+        1.  Override the parameter.
+        2.  Set the new valid Default Value.
+        3.  Submit the changes.
+
+        @assert: Parameter Value overridden with new value.
+
+        @status: Manual
+        """
+
+    @run_only_on('sat')
+    @stubbed()
+    @tier3
+    def test_negative_override_uncheck(self):
+        """Override the Default Parameter value - override Unchecked.
+
+        @feature: Class Parameters - Override
+
+        @steps:
+
+        1.  Don't override the parameter.
+        2.  Set the new valid Default Value.
+        3.  Attempt to submit the changes.
+
+        @assert: Parameter value not allowed/disabled to override.
+
+        @status: Manual
+        """
+
+    @run_only_on('sat')
+    @stubbed()
+    @tier3
+    def test_positive_puppet_default(self):
+        """On Override, Set Puppet Default Value.
+
+        @feature: Class Parameters - Puppet Default
+
+        @steps:
+
+        1.  Override the parameter.
+        2.  Set puppet default value to 'Use Puppet Default'.
+        3.  Submit the changes.
+
+        @assert: Puppet Default Value applied on parameter.
+
+        @status: Manual
+        """
+
+    @run_only_on('sat')
+    @stubbed()
+    @tier1
+    def test_positive_update_parameter_type(self):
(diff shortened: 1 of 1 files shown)


sthirugn:
All these are cli tests but they mention `checkbox`, `submit` and other UI related terminologies.  Is that intentional?
- Does this feature have a cli counterpart? If so can you please rephrase these tests accordingly?

jyejare:
@sthirugn : submit is intentional but not checkbox i mean to say check there.

jyejare:
@sthirugn , Please re-review the updated PR. Thanks !!

tests/foreman/cli/test_classparameters.py
```diff
@@ -0,0 +1,827 @@
...
+    def test_positive_validate_puppet_default_value(self):
+        """Validation doesn't works on puppet default value.
+
+        @feature: Class Parameters - Optional Validation
+
+        @steps:
+
+        1.  Override the parameter.
+        2.  Set puppet default value to 'Use Puppet Default'.
+        3.  Validate this value under section 'Optional Input Validator'.
```
sthirugn: is this applicable to cli test?

tests/foreman/cli/test_classparameters.py
```diff
@@ -0,0 +1,827 @@
...
+    @tier3
+    def test_negative_validate_matcher_value_with_regex(self):
+        """Error raised for matcher value not matching with regex.
+
+        @feature: Class Parameters - Optional Validation
+
+        @steps:
+
+        1.  Override the parameter.
+        2.  Create a matcher with value that doesn't matches
```
sthirugn: s/matches/match

tests/foreman/cli/test_classparameters.py
```diff
@@ -0,0 +1,827 @@
...
+    def test_positive_override_check(self):
+        """Override the Default Parameter value.
+
+        @feature: Class Parameters - Override
+
+        @steps:
+
+        1.  Override the parameter.
+        2.  Set the new valid Default Value.
+        3.  Submit the changes.
```
sthirugn: `Submit the changes` is implied in cli commands, so this is not required.

tests/foreman/cli/test_classparameters.py
```diff
@@ -0,0 +1,827 @@
...
+    @tier3
+    def test_positive_override_check(self):
+        """Override the Default Parameter value.
+
+        @feature: Class Parameters - Override
+
+        @steps:
+
+        1.  Override the parameter.
+        2.  Set the new valid Default Value.
```
sthirugn: Steps 1 and 2 mean the same thing here right?

tests/foreman/cli/test_classparameters.py
sthirugn: @jyejare for this cli test, I dont think `@steps` section is required at all.  The test name `Override the default parameter value` is sufficient enough to explain the test.  What is your opinion?

sthirugn: if you agree with my comment above, can you also please fix other tests?

tests/foreman/cli/test_classparameters.py
jyejare: Both are different options in CLI command. The intention is to set override to True before setting the value.

tests/foreman/cli/test_classparameters.py
jyejare: See my answer on your questions 'Steps 1 and 2 mean the same thing here right?'.

tests/foreman/cli/test_classparameters.py
jyejare: Yes, As asset says it should not work with puppet default value. The intention is to 'NOT' raise any error for puppet default value.

sthirugn:
TY @jyejare for answering my questions.

sthirugn:
Although the steps explained here is not so intuitive for a cli test (it is for a ui test), but this is not a blocker and can be adjusted/updated when the tests are implemented.  I am okay for now.

sthirugn:
ACK

rplevka:
ACK

PR #3286, opened 2016-02-10
Force the removal of a running container for UI tests

abalakh:
Today automation started failing because of lack of free space. 
`docker ps -a | grep Exited | awk '{print $1 }'` showed there were 149 docker containers not deleted on our automation box for some reason.
As we use 1 container per each ui test, i assume `remove_container()` deletes containers in most cases (as we have 800+ ui tests), but sometimes it doesn't (for example, container was in use or stopped responding).
I'm suggesting using `remove_container()` with `force=True`, which is equal to `docker rm --force`, which uses SIGKILL to stop & remove container.
It may help or maybe not, but, imo, it worth a try.

### DIFF
--- a/robottelo/ui/browser.py
+++ b/robottelo/ui/browser.py
@@ -163,7 +163,7 @@ class DockerBrowser(object):
             return
         self._client.stop(self.container['Id'])
         self._client.wait(self.container['Id'])
-        self._client.remove_container(self.container['Id'])
+        self._client.remove_container(self.container['Id'], force=True)
 
     def __enter__(self):
         self.start()


abalakh:
We can merge this PR and monitor few next automation runs.
If `docker ps -a | grep Exited | awk '{print $1 }'` does not return 140+ docker containers - problem solved, otherwise we can revert this change and implement @JacobCallahan's suggestion - to run 
```
docker rm `docker ps -a | grep Exited | awk '{print $1 }'`
```
after each automation run.

JacobCallahan:
APT. Hopefully this does the trick!

rplevka:
'Image of Playing with fire'
kind of risky, isn't it? I propose testing it first, before merging into master

sthirugn:
ACK, force removing a single container in teardown seems harmless to me.

JacobCallahan:
Especially since the test itself should be finished.

rplevka:
@sthirugn ok

ACK

sthirugn:
:+1:

elyezer:
@rplevka I'm gonna talk about this later with you. I think the left containers was from the automation itself since it uses sesame as an external docker compute resource.

With the `force=True` we should be able to validate if those containers are from automation or from running UI tests. Thanks for the changes.

sthirugn:
@elyezer good point.  I think @kbidarkar's clean up script should take care of it then.  May be we need to run it often, he mentioned he is running it weekly now.

PR #3293, opened 2016-02-16
Implement API hostgroup CRUD tests

### LINKED ISSUE #3255: Add CRUD tests for API Host Groups

abalakh:
Closes #3255
```python
py.test tests/foreman/api/test_hostgroup.py -n auto                         
gw0 [40] / gw1 [40] / gw2 [40] / gw3 [40]
........................................
============================ 40 passed in 288.46 seconds =============================
```

### DIFF
--- a/tests/foreman/api/test_hostgroup.py
+++ b/tests/foreman/api/test_hostgroup.py
@@ -2,12 +2,16 @@
 """Tests for the ``hostgroups`` paths."""
 from fauxfactory import gen_string
 from nailgun import client, entities, entity_fields
+from random import randint
+from requests.exceptions import HTTPError
 from robottelo.api.utils import promote, one_to_one_names
 from robottelo.config import settings
 from robottelo.constants import PUPPET_MODULE_NTP_PUPPETLABS
+from robottelo.datafactory import invalid_values_list, valid_hostgroups_list
 from robottelo.decorators import (
     skip_if_bug_open,
     tier1,
+    tier2,
     tier3,
 )
 from robottelo.helpers import get_data_file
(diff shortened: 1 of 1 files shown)


tests/foreman/api/test_hostgroup.py
```diff
@@ -152,6 +163,826 @@ def test_verify_bugzilla_1107708(self):
...
+
+    @tier2
+    def test_positive_create_with_os(self):
+        """Create a hostgroup with operating system specified
+
+        @feature: HostGroup
+
+        @assert: A hostgroup is created with expected operating system assigned
+        """
+        arch = entities.Architecture().create()
```
elyezer: Seems that this test can be simplified to:

```python
        os = entities.OperatingSystem().create()
        hostgroup = entities.HostGroup(
            architecture=os.architecture,
            location=[self.loc],
            operatingsystem=os,
            organization=[self.org],
            ptable=os.ptable,
        ).create()
        self.assertEqual(hostgroup.operatingsystem.read().name, os.name)
```

I'm not sure if the OS create is going to create those dependent entities. This is a guess since I've been a long time away from writing API tests.

elyezer:
Tests looks good to me. I think we can take more advantage of the related entities that NailGun creates. But as I said, I'm not 100% sure if it is possible.

ACK, my comment is not a block.

tests/foreman/api/test_hostgroup.py
abalakh: Good point, but `EntityCreateMixin.create_missing()` creates only required entities, and for `Operating System` architecture and ptable are optional, so they won't be created.

sthirugn:
ACK

sthirugn:
@abalakh answered @elyezer's question. merging now.

tests/foreman/api/test_hostgroup.py
elyezer: @abalakh thanks for the clarification. I thought they were required and commented. Also is good to know that you are considering these points when writing the tests. Thank you!

PR #3294, opened 2016-02-16
adding try finally block to vm.get() in order to fetch the log files from VM

anarang:
adding try finally block to vm.get() in order to fetch the log files from VM irrespective of pass/fail of the tests.

### DIFF
--- a/tests/foreman/rhai/test_rhai.py
+++ b/tests/foreman/rhai/test_rhai.py
@@ -68,16 +68,21 @@ class RHAITestCase(UITestCase):
         """
         # Register a VM to Access Insights Service
         with VirtualMachine(distro='rhel67') as vm:
-            vm.configure_rhai_client(self.ak_name, self.org_label, 'rhel67')
+            try:
+                vm.configure_rhai_client(self.ak_name, self.org_label,
+                                         'rhel67')
 
-            with Session(self.browser) as session:
-                # view clients registered to Red Hat Access Insights
-                session.nav.go_to_select_org(self.org_name)
-                Navigator(self.browser).go_to_insights_systems()
-                result = self.rhai.view_registered_systems()
-                self.assertIn("1", result, 'Registered clients are not listed')
-            vm.get('/var/log/redhat-access-insights/redhat-access'
-                   '-insights.log', './insights_client_registration.log')
+                with Session(self.browser) as session:
+                    # view clients registered to Red Hat Access Insights
+                    session.nav.go_to_select_org(self.org_name)
+                    Navigator(self.browser).go_to_insights_systems()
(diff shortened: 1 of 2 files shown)


tests/foreman/rhai/test_rhai.py
```diff
@@ -68,16 +68,20 @@ def test_positive_register_client_to_rhai(self):
...
+
+                with Session(self.browser) as session:
+                    # view clients registered to Red Hat Access Insights
+                    session.nav.go_to_select_org(self.org_name)
+                    Navigator(self.browser).go_to_insights_systems()
+                    result = self.rhai.view_registered_systems()
+                    self.assertIn("1", result,
+                                  'Registered clients are not listed')
+            finally:
+                vm.get('/var/log/redhat-access-insights/redhat-access'
```
elyezer: Please improve the formatting:

```python
vm.get(
    '/var/log/redhat-access-insights/redhat-access-insights.log',
     './insights_client_registration.log'
)
```

This applies for the other block.

elyezer:
As per last test run the failure happened on the assertion: `AssertionError: 1 != 0 : --test-connection check was not successful`

Because this I think is better to use the try/finally block just around that assertIn line.

ACK pending comments.

tests/foreman/rhai/test_rhai.py
anarang: @elyezer I changed the formatting to above because it was giving >80 too long error on flak8 check. So, if I change it to what you mentioned, it will not pass the flak8 checks.

anarang:
@elyezer The failure around --test-connection is because of a known bug. 
Here is a reasoning to what you are saying: We require the try finally block around the entire setup as the error might occur at first in vm.configure_rhai_client() which could end up with a failure "Unable to register client to RHAI" and we would like to debug that.

What is the coverage/coverails check all about?

sthirugn:
ACK pending @elyezer's comments

elyezer:
@anarang the coveralls checks how much of the framework is covered in tests. The coverage may have decreased an amount less than 0.1 which is now show. But don't worry about it here.

tests/foreman/rhai/test_rhai.py
elyezer: You can do the following:

```python
vm.get(
    '/var/log/redhat-access-insights/'
    'redhat-access-insights.log',
    './insights_client_registration.log',
)
```

By separating arguments in different lines it makes the code more readable even if you break the string in the middle.

sthirugn:
ACK pending rebase

anarang:
@sthirugn @elyezer rebase done.

elyezer:
ACK

PR #3295, opened 2016-02-17
Added test-stubs for Capsule Management UX feature.

kbidarkar:
*) These are the initial test scenarios for the capsule mgmt ux.
*) Currently the tests scenarios are related to overiview, services
   puppet, puppet-ca and pulp.

### DIFF
--- a/tests/foreman/ui/test_capsule.py
+++ b/tests/foreman/ui/test_capsule.py
@@ -1,7 +1,7 @@
 # -*- encoding: utf-8 -*-
 """Test class for UI functions against an isolated capsule"""
 
-from robottelo.decorators import stubbed, tier3
+from robottelo.decorators import run_only_on, stubbed, tier1, tier3, tier4
 from robottelo.test import UITestCase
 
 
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_capsule.py
```diff
@@ -93,3 +93,349 @@ def test_positive_chost_selector(self):
...
+
+        @Status: Manual
+        """
+
+    @run_only_on('sat')
+    @stubbed
+    @tier1
+    def test_positive_capsule_context(self):
+        """Check Capsule has Location and Organization
+           coloumn in the Index Page.
```
elyezer: Line continuation should be aligned with the opening """, otherwise it is going to generate a bunch of spaces between last word of first line and the first word of the second line.

Please fix this for the remaining of the file.

tests/foreman/ui/test_capsule.py
```diff
@@ -93,3 +93,349 @@ def test_positive_chost_selector(self):
...
+        """Check Capsule Status in Index Page.
+
+        @Feature: Capsule Management UX
+
+        @Steps:
+
+        1. Navigate to the Capsules Index Page
+        2. Check the status of the Proxy.
+
+        @Assert: Whether the status of the Capsules
```
elyezer: Line continuation for assertion should continue aligned to the `@Assert`.

tests/foreman/ui/test_capsule.py
```diff
@@ -93,3 +93,349 @@ def test_positive_chost_selector(self):
...
         @Status: Manual
         """
+
+    @run_only_on('sat')
+    @stubbed
+    @tier1
+    def test_positive_capsule_version(self):
+        """Check Capsule Version in About Page.
+
+        @Feature: Capsule Management UX
```
elyezer: We use to use UI instead of UX. I think we should continue using UI and avoid introducing a new nomenclature.

tests/foreman/ui/test_capsule.py
```diff
@@ -93,3 +93,349 @@ def test_positive_chost_selector(self):
...
+
+        @Steps:
+
+        1. Navigate to the About Page
+        2. Check the versions coloumn of the Capsules.
+
+        @Assert: Whether the version coloumn of the Capsules exist.
```
elyezer: Shouldn't assert if the version is the expected one?

tests/foreman/ui/test_capsule.py
elyezer: Actually this may not impact on docs generation. In that case the alignment here can be left as is. But for the code style sake, all other code align on the start instead of this way.

elyezer:
ACK, pending comments and fixing Travis.

tests/foreman/ui/test_capsule.py
```diff
@@ -93,3 +93,317 @@ def test_positive_chost_selector(self):
...
+    @tier1
+    def test_positive_capsule_version(self):
+        """Check Capsule Version in About Page.
+
+        @Feature: Capsule Management UI
+
+        @Steps:
+
+        1. Navigate to the About Page
+        2. Check the versions coloumn of the Capsules.
```
sthirugn: s/coloumn/column

tests/foreman/ui/test_capsule.py
```diff
@@ -93,3 +93,317 @@ def test_positive_chost_selector(self):
...
+        for Isolated Capsule in the Overview tab.
+
+        @Feature: Capsule Management UI
+
+        @Steps:
+
+        1. Navigate to the Capsules Show Page
+        2. Check the capsule's overview tab in the show Page.
+
+        @Assert: Whether the 'Cancel Sync" button is visible
```
sthirugn: use single quote only?

tests/foreman/ui/test_capsule.py
```diff
@@ -93,3 +93,317 @@ def test_positive_chost_selector(self):
...
+        """Check Capsule Version in About Page.
+
+        @Feature: Capsule Management UI
+
+        @Steps:
+
+        1. Navigate to the About Page
+        2. Check the versions coloumn of the Capsules.
+
+        @Assert: Whether the version of the Capsules exists in the about page.
```
sthirugn: I think you can remove `Whether` and write something like:
`The version of the Capsules is displayed in the About page`

Please fix other asserts in this way to improve readability.

sthirugn:
Pending comments

sthirugn:
ACK

elyezer:
ACK, I will let @sthirugn take a second look before merging.

PR #3300, opened 2016-02-17
[WIP] Adding support for SauceLabs execution.

omaciel:
A while back we removed support for running UI tests via SauceLabs.

This Pull Request adds that support back.

PR #3313, opened 2016-02-24
Add SauceLabs browser

elyezer:
A while back we removed support for running UI tests via SauceLabs. This
is time to get it back with extra improvements. Now one is able to
select which browser is gonna be created selenium, docker or saucelabs.

Also when choosing to run on SauceLabs, setting the webdriver will also
choose the browser that is gonna be created on SauceLabs. Valid values
are firefox, chrome and ie.

Thanks @omaciel for the initial code added on #3300.


elyezer:
I ran tests the UI test_login using each browser and they are working fine.

omaciel:
ACK

JacobCallahan:
APT

PR #3387, opened 2016-04-13
ISS stubs for Redhat Contents

jyejare:
This PR includes stubs for Redhat and mixed(Redhat+custom) contents import and export testcases.

Earlier added testcases/stubs were custom contents related stubs.

### DIFF
--- a/tests/foreman/cli/test_satellitesync.py
+++ b/tests/foreman/cli/test_satellitesync.py
@@ -830,3 +830,289 @@ class InterSatelliteSyncTestCase(CLITestCase):
 
         @status: Manual
         """
+
+# Red Hat Repositories Export and Import
+
+    @run_only_on('sat')
+    @stubbed()
+    @tier3
+    def test_positive_export_redhat_yum_repo(self):
+        """Export Red Hat YUM repo in directory.
+
+        @feature: ISS - Export
+
+        @steps:
+
+        1. Export whole Red Hat YUM repo to some path.
+
+        @assert: Whole YUM repo contents has been exported to directory
+        specified in settings.
+
+        @status: Manual
+        """
+
+    @run_only_on('sat')
+    @stubbed()
+    @tier3
+    def test_positive_export_import_redhat_yum_repo(self):
+        """Import the exported YUM repo contents.
+
+        @feature: ISS - Import
+
+        @steps:
+
+        1. Export Red Hat YUM repo to path which will be accessible over HTTP.
+        2. Import the repository by defining the CDN URL the same as the
+        exported HTTP URL.
+
+        @assert: All the exported YUM repo contents are imported successfully.
+
+        @status: Manual
+        """
+
+    @run_only_on('sat')
+    @stubbed()
+    @tier3
+    def test_positive_export_redhat_incremental_yum_repo(self):
+        """Export Red Hat YUM repo in directory incrementally.
+
+        @feature: ISS - Export
+
+        @steps:
+
+        1. Export whole Red Hat YUM repo.
+        2. Add some packages to the earlier exported yum repo.
+        3. Incrementally export the yum repo from last exported date.
+
+        @assert: Red Hat YUM repo contents have been exported incrementally in
+        separate directory.
+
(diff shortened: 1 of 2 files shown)


elyezer:
I don't agree on having multiple steps/items on the assertion.

Taking this test method:

```
   def test_positive_export_import_redhat_yum_repo(self):
        """Export redhat yum repo in directory and Import.

        @feature: ISS - Export Import

        @steps:

        1. Export whole redhat yum repo to a directory specified in settings.
        2. Explore exported contents to /var/www/html/pub/export directory.
        3. Set the step 2 location path as CDN URL in some org/satellite to
        import the exported yum repo contents.

        @assert:

        1. Whole repo contents has been exported to directory
        specified in settings.
        2. All The exported repo contents has been imported in org/satellite.
        3. Repos rather than imported are restricted to enable in Redhat
        Repositories.

        @status: Manual
        """
```

I understand of the information to be on the assert should express the conclusion of the test case. I will try to express what I'm trying to say by refactoring the above test stub.

I would like to test if I can export a Red Hat YUM repository to a directory and import it. In order to do that I need to:

1. Export the Red Hat YUM repository to _some path (this path should be available through http)_.
2. Import the repository by defining the CDN URL the same as the exported HTTP URL.

Finally I would like to assert that some package available on the exported directory is present on the imported one.

Translating this to a test stub it would be:


```python
def test_positive_export_import_redhat_yum_repo(self):
    """Export a Red Hat YUM repository and import it.

    @steps:
    
    1. Export the Red Hat YUM repository to _some path (this path should be available through http)_.
    2. Import the repository by defining the CDN URL the same as the exported HTTP URL.

    @assert: Some package present on the exported directory is present on the imported one
```

I tried to work using one added test stub as an example but I don't know much about the feature. That said if in the future I get this to automate I will not be able to figure out, for example, if `/var/www/html/pub/export` is the same as the mentioned directory specified on settings, also I will not know what specific settings it is about (to reinforce I don't know much about the Satellite 6 settings in general).

What I am trying to expose here is that we should try to be more brief and direct while being clear on what we are trying to accomplish with some specific test method. Also I'm not being picky with you @jyejare specifically, I'm just taking the opportunity on this PR to try to improve on the test stubs we have.

All that said, let me know if you agree with all I wrote (I know it is very long!).

My only concern here for this PR is that the `@assert` should have just one item which is the conclusion of the test. There is no need to list prerequisites or steps on the assertion because if the final goal is failed to accomplish then the previous steps will likely fail as well or will be the reason of the failure.

sthirugn:
I agree with @elyezer.  Assert can just mention what a test case is exactly trying to test - it doesn't need to assert the pre-requisite steps for a test case.

cswiii:
Agree with above as well. We should assert one single expected result for any test case, for the reasons referenced above - you don't want to get stuck with assertion dependencies.

jyejare:
@elyezer, @sthirugn and @cswiii I have done necessary changes in PR as per your guidance and comments.
Please have a look over it again.

sthirugn:
Right now all the `export` and `import` tests are separate.  This means the `export` part should be done both for the `export` and the `import` tests.  Can we  combine the export and import tests?  By this way:
- the number of tests will be less
- running time will be less
- avoids running the same export steps twic

jyejare:
@sthirugn : If I combine both import and export, then I have to have multiple assert(Expecting) statements which we are not following. And as per @cswiii comment if we have multiple assert statements then its better to separate the TC's.
Earlier I was having the combine Tests but I separate export and import after first review as I said above. So if everyone is OK here to remove duplicate export I can easily just delete 'export' stubs as we are covering export steps in import stubs. And I will rename import as 'export-import'.

sthirugn:
@jyejare thank you for the clarification.  I believe we can make test cases atomic and still avoid duplication by combining tests.  But not a  blocker for this PR.  I am okay with leaving this as it is.

sthirugn:
ACK

elyezer:
ACK, thanks @jyejare for improving the test stubs.

PR #3422, opened 2016-04-29
Pinned paramiko version due to missing dependencies in latest.

rochacbruno:
* Paramiko upgraded its crypto dependencies, we should upgrade also, but before we need to make some further investigation and test
* For now pinning the paramiko version to the older
* Also we should pin all other deps

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ setup(
         'inflector',
         'nailgun',
         'numpy',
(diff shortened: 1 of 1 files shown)


JacobCallahan:
ACK

cswiii:
ACK

sthirugn:
:+1:

PR #3423, opened 2016-04-29
Pinned all third party libs version

rochacbruno:
To avoid what happened with **paramiko** being backwards incompatible (see 2.0.0 statement here: http://www.paramiko.org/changelog.html)

I think it is a good practice to pin all the version of third party packages.

A raised question is that we should always use the latest versions, but more important is keeping our automation working without breaks.

So we can use requires.io to notify us if we are using a very older version or insecure package
example here: https://requires.io/github/rochacbruno/robottelo/requirements/?branch=pinned_external_dependencies

We are always in risk of e.g: requests or unittest2 package having a new broken release which will silently break our automation.

In the cases of fauxfactory and nailgun we can leave unpinned since that packages are under our control.

What do you think?


sthirugn:
ACK

JacobCallahan:
@rochacbruno would you mind creating an issue for the inclusion of requires.io? While I agree that it is good to pin down our requirements for stability's sake. We would be remiss to not want to keep up with the latest versions whenever possible.

elyezer:
I think robottelo master should always follow the latest available version. So far pinned versions was due to the lack of time in order to investigate and fix the dependencies. That said, if others agree with this I am ok, but I prefer to continue making robottelo always compatible with the latest versions of the required packages.

We already have the practice of pinning the packages whenever we tag robottelo.

JacobCallahan:
@elyezer that is a good point as well. i am divided on this. One one hand it is great to have stability without needing to scramble to diagnose errors with dependencies. On the other, it is a great practice to stay in date.

rochacbruno:
@elyezer automation on Jenkins is using tagged releases or master upstream?

elyezer:
@rochacbruno it depends on the version of the satellite we are running. Currently all 6.1.z are running on tagged version since we have created the snapshot to freeze tests and requirements.

Since master is our development branch I don't want to pin dependencies because it could lead to the need of many changes when we decide to update the dependencies.

rochacbruno:
@elyezer Ok, so we have to assume to risk of having a failure caused by 3rd party modules on 6.2 automation. 

Also we need to remove the pinned paramiko version and make sure that it will be upgraded in the next run to include the newest **cryptography** dependency.


BTW: I would agree more with you if we have **development** and **master** branches, **development** keeps unstable and **master** is tested on our CI with pinned versions.

elyezer:
@rochacbruno installing latest paramiko is already possible on Jenkins and the changes describing what is needed will be introduced when the WIP branch [1] gets merged.

Regarding the possible failures by experience we did not got much failures caused by third party libraries so far that is not a big risk in my opinion. Until we stabilize the automation to the final 6.2 version with most or all tests passing we may find the need of updating robottelo code to match the updates on its dependencies, which I think is fine until we have a stable version and decide to tag 6.2.

Having development and master branch by just possible dependency issue don't makes 100% sense to me. I mean just because that since as I said the dependency issue just happened once or twice so far (about 2 years+ of robottelo project).

[1] https://github.com/SatelliteQE/robottelo/compare/master...elyezer:use-latest-paramiko

JacobCallahan:
@elyezer @rochacbruno perhaps we should hold off on this for now and discuss it this Wednesday during our meeting.

PR #3425, opened 2016-05-02
Fix encoding declaration error

Ichimonji10:
File encoding declarations must be on the first line, or the second if a
shebang is present. Before this commit:

    $ python -c 'from tests.foreman.api import test_bookmarks'
      File "tests/foreman/api/test_bookmarks.py", line 198
    SyntaxError: Non-ASCII character '\xe2' in file tests/foreman/api/test_bookmarks.py on line 199, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details

After this commit:

    $ python -c 'from tests.foreman.api import test_bookmarks'


sthirugn:
ACK

elyezer:
ACK

PR #3447, opened 2016-05-11
Update get_value for UI settings

elyezer:
Make sure to return a blank value if `Click to edit..` is the text of the
settings because that text will be present just for blank setting value.

### DIFF
--- a/robottelo/ui/settings.py
+++ b/robottelo/ui/settings.py
@@ -61,6 +61,8 @@ class Settings(Base):
             strategy, value = locators['settings.edit_param']
             element = self.wait_until_element((strategy, value % param_name))
             if element:
+                if element.text.lower().startswith('click to edit'):
+                    return ''
                 return element.text
             else:
                 raise UINoSuchElementError(


sthirugn:
ACK pending test result

omaciel:
Background: @elyezer and I were looking at a very particular error related to the `test_positive_update_login_delegation_logout_url_param` where the original value for a setting was `Click to edit.` and the teardown was trying to revert the value to it. :) It should be reverted to a blank value instead.

robottelo/ui/settings.py
```diff
@@ -61,6 +61,8 @@ def get_saved_value(self, tab_locator, param_name):
             strategy, value = locators['settings.edit_param']
             element = self.wait_until_element((strategy, value % param_name))
             if element:
+                if element.text.lower() != 'click to edit..':
```
sthirugn: shouldn't this be ==

robottelo/ui/settings.py
```diff
@@ -61,6 +61,8 @@ def get_saved_value(self, tab_locator, param_name):
             strategy, value = locators['settings.edit_param']
             element = self.wait_until_element((strategy, value % param_name))
             if element:
+                if element.text.lower() == 'click to edit..':
```
omaciel: How about `if element.text.lower().startswith('click to edit'):`?

omaciel:
ACK

```shell
py.test tests/foreman/ui/test_setting.py::SettingTestCase -k test_positive_update_login_delegation_logout_url_param
================================================== test session starts ==================================================
platform darwin -- Python 2.7.10, pytest-2.9.1, py-1.4.31, pluggy-0.3.1
rootdir: /Users/omaciel/hacking/robottelo, inifile:
plugins: cov-2.2.1, xdist-1.14
collected 28 items

tests/foreman/ui/test_setting.py .

=================== 27 tests deselected by '-ktest_positive_update_login_delegation_logout_url_param' ===================
======================================= 1 passed, 27 deselected in 32.71 seconds ========================================
```

sthirugn:
ACK

rplevka:
...guys, guys, guys...how about comparing it to `data-original-title` element attribute? Would save us from trouble with various locales.

scr

omaciel:
@rplevka PR it and it will be merged :)

elyezer:
@rplevka that is even better. Please send a PR with that change.

PR #3466, opened 2016-05-18
Update API & UI sync plan start date tests according to Sat6.2 changes

abalakh:
1) Server now returns date and time in US style (`%m/%d/%Y %I:%M:%p` instead of `%Y-%m-%d %H:%M:%S`)
2) As per [BZ1335133](https://bugzilla.redhat.com/show_bug.cgi?id=1335133), server returns start date in its own time zone, not in user's (that's especially wrong for UI tests), skipping the tests


sthirugn:
ACK

Issue #3467, opened 2016-05-19
Fix oscap tests since default rhel contents are not created anymore


sthirugn:
test: tests.foreman.ui.test_oscapcontent.OpenScapContentTestCase.test_positive_default 
bug: https://bugzilla.redhat.com/show_bug.cgi?id=1336374

This test will be skipped until a decision is made for a permanent fix for this test.

sghai:
@kbidarkar : is this still valid ?

kbidarkar:
This is fixed in automation-tools and hence can be closed, we now run a command which will create default rhel oscap contents.

kbidarkar:
https://github.com/SatelliteQE/automation-tools/blob/master/automation_tools/__init__.py#L629

PR #3468, opened 2016-05-19
skipping oscap default test due to oscap behavior change

sthirugn:
this test should be rewritten or removed.  This will be tracked in https://github.com/SatelliteQE/robottelo/issues/3467


sthirugn:
This addresses the 39 times failed jenkins test

PR #3476, opened 2016-05-23
Skip CLI content host update name test due to open BZ

abalakh:
[BZ1338780](https://bugzilla.redhat.com/show_bug.cgi?id=1338780) - content host name can't be updated with hammer


sthirugn:
ACK

PR #3482, opened 2016-05-24
Skip CLI activation key subscription remove test due to open BZ

abalakh:
[BZ1339211](https://bugzilla.redhat.com/show_bug.cgi?id=1339211) - 'hammer activation-key subscriptions' shows already removed subscriptions


sthirugn:
ACK

PR #3485, opened 2016-05-24
Improving performance of test_bookmark tests:

sthirugn:
Before this PR, the bookmarks UI tests took 3 hr 15 minutes to complete.

1. Making all the tests use only one entity to test entities except
test_positive_delete_bookmark. Why was this approach chosen?
 - It is very essential to test creation/deletion of bookmarks for all listed
entities which is covered fully by test_positive_delete_bookmark test case.  Every other
test uses the standard bookmark features like rename, updating same name, etc and all of
these tests use the same bookmark UI.

2. Reusing the session for each test - previously each subTest created its own
UI session with associated browser launch, login, logout


sthirugn:
Tests in progress

omaciel:
**3 hours 15 minutes**??? For one test??? Wow, I'll have to see this tomorrow 😀

sthirugn:
@omaciel - 3 hours 15 minutes for the entire bookmark ui module

sthirugn:
Test results:
Before change - total time - 3 hours 15 minutes
After change - total time - 12 minutes

without test_positive_delete_bookmark tests:
```

py.test tests/foreman/ui/test_bookmark.py
======================================================= test session starts =======================================================
platform linux2 -- Python 2.7.11, pytest-2.9.1, py-1.4.31, pluggy-0.3.1
rootdir: /home/suresh/hacking/robottelo, inifile: 
plugins: cov-2.2.1, xdist-1.14
collected 13 items 

tests/foreman/ui/test_bookmark.py .........s.s.

============================================= 11 passed, 2 skipped in 179.73 seconds ==============================================
```

```
py.test tests/foreman/ui/test_bookmark.py -k test_positive_delete_bookmark
======================================================= test session starts =======================================================
platform linux2 -- Python 2.7.11, pytest-2.9.1, py-1.4.31, pluggy-0.3.1
rootdir: /home/suresh/hacking/robottelo, inifile: 
plugins: cov-2.2.1, xdist-1.14
collected 14 items 

tests/foreman/ui/test_bookmark.py .

==================================== 13 tests deselected by '-ktest_positive_delete_bookmark' =====================================
============================================ 1 passed, 13 deselected in 568.82 seconds ============================================
```

robottelo/constants.py
```diff
@@ -837,7 +837,7 @@
         'skip_for_ui': True},
     {'name': 'HostCollection', 'controller': 'katello_host_collections'},
     {'name': 'Architecture', 'controller': 'architectures'},
-    {'name': 'HardwareModel', 'controller': 'models'},
+    {'name': 'HardwareModel', 'controller': 'models', 'skip_for_ui': True},
```
sthirugn: @rplevka I see that hardware model page is empty by default, so search button is not present.  Is this a bug?

robottelo/constants.py
```diff
@@ -861,7 +861,8 @@
     {'name': 'SmartProxy', 'controller': 'smart_proxies', 'skip_for_ui': True},
     {'name': 'Compute_Resource', 'controller': 'compute_resources',
         'setup': entities.DockerComputeResource},
-    {'name': 'Compute_Profile', 'controller': 'compute_profiles'},
+    {'name': 'Compute_Profile', 'controller': 'compute_profiles',
+        'skip_for_ui': True},
```
sthirugn: @rplevka I also see that Compute profile page is empty by default, so search button is not present.  Is this a bug?  I am not sure if it is a bug so I skipped this too for now. Please let me know.

abalakh:
Just wanted to start working on this module and here's a PR :smile: 
But i'm afraid i have to put NACK on this.
3 points:
- if some pages are empty by default, you shouldn't skip them, use setup instead, that's how it was supposed to be implemented. For example: `{'name': 'HardwareModel', 'controller': 'models', 'setup': entities.Model},` this will create 1 hardware model before executing the test so search bar will be present
- (nitpick) `for entity in self.getOneEntity():` is something that is redundant for your approach - no point in a loop for 1 element. But you may want to leave it because of next point:
- personally i don't like this approach with `getOneEntity()` - it's some kind of new workaround, when we have pretty common approach with `run_one_datapoint`. That's how i see  it: move `BOOKMARK_ENTITIES ` from constants to datafactory (and transform into a function returning corresponding list) and decorate it with `@datacheck`. This way we will have 1 entity per test without any extra helpers/workarounds and in addition we won't loose ability to verify all the possible combinations if needed (with `run_one_datapoint=false`)

I know i'm basically asking to completely rewrite this PR, but I can work on that if you have other things to do, just tell me :)

sthirugn:
Thank you @abalakh 

* Answer 1: Yes I don't know how the setup for these entities work.  I thought @rplevka did these tests and so I asked @rplevka for his feedback for both these entities.  Depending on his feedback, I was planning to work on either creating a new bugzilla and skipping the entities or remove them altogether as it shows now.
* Answer 2 & Answer 3: Yes I did know about `@datacheck` approach and I thought about implementing it.  I wanted to get this PR out soon before Snap 13 build (to be delivered today) is triggered, so we could get this merged to see some improvement and improvise on that as follow up, as needed.  I agree with you that there is a better way to do.  But, my question to you is:
    - Do you like to get this quickly in for Snap 13 to realize some improvement and improvise later or should we wait more time to completely rewrite this?   I am ok with either approach.

abalakh:
@sthirugn Roman wrote the stubs, i was implementing them. So, yeah, we have some entities (around 15 of them) that require at least 1 item in a table for search bar to appear, therefore no BZ is needed, we already predicted such behavior and added `setup` key where you can specify which entity to create before test execution. Therefore you should update those entities like that:
```python
    {'name': 'HardwareModel', 'controller': 'models', 'setup': entities.Model},
    {'name': 'Compute_Profile', 'controller': 'compute_profiles',
        'setup': entities.ComputeProfile},
```
Skip should be used only when either we don't have required infrastructure ready (module is not created in robottelo / impossible to create entity via nailgun atm) or when there's an open bz. And in case of open bz, better to use it like `'skip_for_ui': 12345` where 12345 - bz id, this way we'll automatically start using entity when bz is closed.

Re your question - personally i'm against merging it. As @oshtaier mentioned few PRs ago, we've been introducing way too much workarounds lately, and this specific one is not even required as there's a better way to implement it. I'd agree if the tests were broken and we needed better results asap, but tests are ok, it's just an improvement, so let's do it proper way. 
Oh, and another point for datafactory approach - we have api tests too, which are affected as well. With `getOneEntity()` approach we should add similar func to api tests too, with datafactory - no need to.

P.S. My proposal is still actual - if you're busy at the moment but you'd like to get this fixed sooner than later, lemme grab it, it shouldn't take much time from me. I'd rather make some extra efforts than merge another workaround, we definitely overused them.

sthirugn:
@abalakh lets talk about this offline - we have 90+ failures which needs work/fix other than this asap.

robottelo/constants.py
elyezer: I think the skip logic should be implemented on the uitests themselves, something like create a copy of this constant and then pop the value for HardwareModel in order to not pick it up.

robottelo/constants.py
elyezer: Same as previous comment. This is just data should not contain logic.

sthirugn:
As per @abalakh suggestion, reverting the session changes and added entity setup for hardware model and compute profiles

abalakh:
ACK, as datafactory suggestion should be resolved in #3487

rplevka:
@sthirugn sorry for a late reply - fortunately all needed has been said by @abalakh (regarding the `setup` key)

PR #3486, opened 2016-05-25
Mark some skipped tests with run_in_one_thread decorator

abalakh:


sthirugn:
what is the criteria to move these threads to single thread?

abalakh:
@sthirugn latest automation run :) i've just looked through the failures and saw some tests failed because of race condition with importing subscriptions. It was hard to cover every single sequential candidate among 1800  tests from the first attempt, so another PR was expected.

sthirugn:
I spoke to @abalakh and as per his comment - all the affected tests here are manifest related.

sthirugn:
ACK

PR #3492, opened 2016-05-27
Org and Host cleanup using nailgun.signals for API testcases

rochacbruno:
Still waiting for standalone-automation results.

- Using nailgun from: https://github.com/rochacbruno/nailgun/tree/using_signals_emit_decorator 
   (to be merged)
- Delete all hosts and then all orgs
- implemented only for nailgun entities 

Locally with 8 threads api/test_host succeed

```
py.test -n 8 -v tests/foreman/api/test_host.py
=== 45 passed in 202.96 seconds ===
```

Will start standalone-automation to test the whole api test module


rochacbruno:
I observed that assigned hosts is not the only nested entity to be deleted prior to Organization

Got some of tasks failing with this error:

```
PG::Error: ERROR:  update or delete on table "katello_repositories" violates foreign key constraint "filters_repositories_repository_id_fk" on table "katello_content_view_filters_repositories"
DETAIL:  Key (id)=(6) is still referenced from table "katello_content_view_filters_repositories".

Actions::Katello::Organization::Destroy
{"organization"=>{"id"=>622, "name"=>"GbZQzUjLaZ", "label"=>"GbZQzUjLaZ"},
 "services_checked"=>["pulp", "pulp_auth", "candlepin", "candlepin_auth"]}

ActiveRecord::InvalidForeignKey: PG::Error: ERROR:  update or delete on table "katello_repositories" violates foreign key constraint "filters_repositories_repository_id_fk" on table "katello_content_view_filters_repositories"
DETAIL:  Key (id)=(6) is still referenced from table "katello_content_view_filters_repositories".
: DELETE FROM "katello_repositories" WHERE "katello_repositories"."id" = $1
```

omaciel:
@rochacbruno I want to say that you found a bug :) Seems that deleting an org is not cascading down to `content views` and this is why you're seeing this? Did you have a content view for the test that failed? Was it populated with content? Was it published and/or promoted? Depending on your answer we may be able to find what exactly blocks you from deleting an org with a content view

rochacbruno:
@omaciel I checked all the task with result = error and there is no content view for the failing organizations, so maybe it was deleted or a race condition happened. If someone want to take a look I give the box url.

rochacbruno:
S-A build 106 finished, I ran only api tests
```python
py.test -m 'tier1 and not run_in_one_thread and not stubbed' -n 4 tests/foreman/api/
============= 6 failed, 339 passed, 15 skipped in 1150.98 seconds ==============
```
All the created organization has been cleaned, except the ones created on failed tests

rochacbruno:
S-A build 110 ran for ui and all the organizations created by the API has been cleaned.

```
-n 4 foreman/ui  -m "tier1 and not run_in_one_thread and not stubbed"
=== 7 failed, 305 passed, 15 skipped in 5170.79 seconds ===
```

rochacbruno:
Before merging this we should discuss if this approach will be taken or if we are going to stick only with **installer --reset** approach.

*Also before merging nailgun PR should be merged and requiremets here reverted to original.

elyezer:
Both approaches should be done, the reset will help between tier jobs and the org cleanup will be done to maintain the health during a tier run.

setup.py
```diff
@@ -19,7 +19,8 @@
         'cryptography',
         'fauxfactory',
         'inflector',
-        'nailgun',
+        # 'nailgun',
```
sthirugn: please uncomment?

sthirugn:
ACK

elyezer:
ACK

PR #3505, opened 2016-05-31
Set default organization to admin user, removed go_to_org calls

rochacbruno:
Set default_organization to the user created in bookmark tests makes the user to login with that organization context by default.

Also we can remove the  extra 11 calls to ``session.nav.go_to_select_org`` 

```python
py.test -v tests/foreman/ui/test_bookmark.py
=== 1 failed, 9 passed, 4 skipped, 1 error in 3593.26 seconds ===
```

the failure is: test_positive_delete_bookmark (tests.foreman.ui.test_bookmark.BookmarkTestCase) because notification message stayed above the logout button in saucelsbs session.


sthirugn:
ACK

Issue #3519, opened 2016-06-03
Use session_org and session_user in UI test cases instead of creating custom users and orgs.


rochacbruno:
#3515 introduces dynamic way to define **session_user** and **session_org** 

So once it is merged we should look at every UI test_case that creates a custom org or a custom user for the session and replace with the new approach

rochacbruno:
It is covered in the most important parts of UI (and currently the slowly dashboard when "default organization" user logs in has been fixed)

PR #3524, opened 2016-06-07
Using session_org in ui/test_subscription to address #3519

rochacbruno:
Created organization in **set_session_org** makes the user to login with that organization context and avoids 1 call to **go_to_select_org**

```bash
Session set with:
	User: 8:odGMQDoocW
	Organization: 6:IEeJMD
.....
2016-06-07 10:44:00 - robottelo - DEBUG - Cleanup deleted 1 entities
2016-06-07 10:44:00 - robottelo - DEBUG - Deleting session_user
....
=== 1 passed in 84.11 seconds ===
```


sthirugn:
ACK

PR #3572, opened 2016-06-29
Update makefile with endtoend tests instead of smoke

rplevka:
Since `smoke` tests are no longer cool and they have been renamed to `endtoend` tests, we need to update the targets that contain references to them.

The `satellite6-standalone-automation` job template needs to be updated too in order to reflect these changes:
https://github.com/SatelliteQE/robottelo-ci/pull/316


elyezer:
ACK

abalakh:
ACK

PR #3604, opened 2016-07-14
Add hostcolletion.execute_bulk_package_action closes #3598

### LINKED ISSUE #3598: Add execute_package_action() helper for Host Collection
For content hosts there's a UI helper [execute_package_action()](https://github.com/SatelliteQE/robottelo/blob/master/robottelo/ui/contenthost.py#L112) which schedules package/package group installation/update/removal.
Similar helper is required for host collection to be able to test content installation on host collection.
Pages and element placement are different for host collection, but idea should be the same - it should accept host collection name, action name (package installation/removal etc.) and action value (package or package group name to work with).
Another difference is content-host helper checks whether task finished and returns task status, which can't be done for host collection - package operations just schedule task and do not navigate to that task, so this part of helper should be just skipped.

rochacbruno:
closes #3598

Added `hostcollection.execute_bulk_package_action` tested using the @abalakh  approach and worked locally with selenium + firefox for each one of the cases:

- update all
- install a package
- update a package
- remove a package

All **action_names** tested for **package** and **package_group** using **action_via=katello_agent** because **remote_execution** cannot be tested locally (or I missed something about it)

### DIFF
--- a/robottelo/ui/hostcollection.py
+++ b/robottelo/ui/hostcollection.py
@@ -130,3 +130,62 @@ class HostCollection(Base):
             (strategy, value % host_name), timeout=8)
         if element is None:
             raise UIError("Adding host {0} is failed".format(host_name))
+
+    def execute_bulk_package_action(self,
+                                    name,
+                                    action_name,
+                                    action_type='package',
+                                    action_value=None,
+                                    action_via='katello_agent',
+                                    org_name=None):
+        """Execute remote package action on a host collection
+
+        :param name: hostcollection name to remotely execute package action on
+        :param action_name: remote action to execute. Can be one of 4:
+            'update all', 'install', 'update', 'remove'
+        :param action_type: type of package. can be one of 2:
+            'package' or 'package_group'
+        :param action_value: Package or package group name to remotely
+            install/update/remove (depending on `action_name`)
+        :param action_via: the way to perform the action. Can be one of 3:
+            'katello_agent', 'remote_execution', 'remote_execution_custom'
+        :param org_name: The name of the organization for context, it is an
+            optional param since org can be previously selected on session
+        :raise: UIError if remote task finished by timeout
+
+        :return: Returns a string containing task status
+        """
+
+        if action_name != 'update all' and not action_value:
+            raise AttributeError('package or group name is required')
+
+        if org_name:
+            Navigator(self.browser).go_to_select_org(org_name, force=False)
+
+        self.click(self.search(name))
+        self.click(tab_locators['hostcollection.collection_actions'])
+        self.click(locators['hostcollection.collection_actions.packages'])
+
+        if action_value:
+            type_loc = "contenthost.bulk_actions.{0}_type".format(action_type)
+            self.click(locators[type_loc])
+            self.assign_value(
+                locators['contenthost.bulk_actions.package_name_input'],
+                action_value
+            )
+
+        strategy, action_link_locator_path = locators[
+            'contenthost.bulk_actions.via_{0}'.format(action_via)]
+        action_link_locator_path = action_link_locator_path % action_name
+        action_link_locator = (strategy, action_link_locator_path)
+
+        strategy, value = locators['contenthost.bulk_actions.action_dropdown']
+        self.click((strategy, value % action_link_locator_path))
+        self.click(action_link_locator)
+
+        result = self.wait_until_element(
+            locators['contenthost.bulk_actions.remote_action_scheduled']
+        )  # this alert box fades after 2 seconds
+
+        if result is None:
+            raise UIError('Timeout waiting for package action to schedule')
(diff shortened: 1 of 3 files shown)


robottelo/ui/hostcollection.py
```diff
@@ -130,3 +130,72 @@ def add_host(self, name, host_name):
...
+        # dynamically get locators for dropdown button and action link
+        # based in action_name and action_via
+        action_loc_key = "contenthost.bulk_actions.{action}_via_{via}".format(
+            action=action_name, via=action_via)
+        action_link = locators[action_loc_key]
+
+        # dropdown locator built here instead of locators module
+        # because it is dynamically named
+        dropdown = "/../../../button[contains(@data-toggle,'dropdown')]"
+        action_dropdown_locator = (action_link[0], action_link[1] + dropdown)
```
abalakh: We never define locators inside code. If locator has dynamic part - just use `%s` in its definition, and complement it with `% dynamic_part` in code. Take a look at literally every locator which is set to `strategy, value` variables. For example:
```python
# "discoveredhosts.delete_from_facts": (
#        By.XPATH, ("//a[contains(@href,'%s') and contains(.,'Delete')]")),
strategy, value = locators['discoveredhosts.delete_from_facts']
self.click((strategy, value % hostname), wait_for_ajax=False)
```

robottelo/ui/hostcollection.py
```diff
@@ -130,3 +130,72 @@ def add_host(self, name, host_name):
...
+        self.click(action_dropdown_locator)  # because element must be visible
+        self.click(action_link)
+
+        result = self.wait_until_element(
+            locators['contenthost.bulk_actions.remote_action_finished'],
+            timeout=timeout,
+        )  # this alert box fades after 2 seconds
+
+        if result is None:
+            raise UIError('Timeout waiting for package action to finish')
```
abalakh: Unlike content-hosts, for host collections no navigation to created task is performed after task is scheduled. Therefore we can't check whether the task has finished.
What are you doing here is just verifying that task was scheduled and waiting for notification about that, but not for task completion. So please update this message, docstring and `locators['contenthost.bulk_actions.remote_action_finished']` (to smth like `locators['contenthost.bulk_actions.remote_action_scheduled']`) accordingly.

robottelo/ui/hostcollection.py
```diff
@@ -130,3 +130,72 @@ def add_host(self, name, host_name):
...
+
+        return result.get_attribute('bst-alert')  # u'success|warning|error'
```
abalakh: `result` is set to `self.wait_until_element(locators['contenthost.bulk_actions.remote_action_finished'])`. Locator is defined by `'//div[@bst-alert="success"]'`. So `result.get_attribute('bst-alert')` will always return `success`. Therefore this `return` makes no sense. You should either use another xpath to catch all types of notifications or simply drop this return (afaik in most cases helpers don't check notification messages, we usually check them inside the tests. @oshtaier pls correct me if i'm wrong).

abalakh:
Pending comments. Good job btw!

robottelo/ui/hostcollection.py
sghai: +1 to what @abalakh said. We define all locators here: https://github.com/SatelliteQE/robottelo/blob/master/robottelo/ui/locators.py

robottelo/ui/hostcollection.py
```diff
@@ -130,3 +130,72 @@ def add_host(self, name, host_name):
             (strategy, value % host_name), timeout=8)
         if element is None:
             raise UIError("Adding host {0} is failed".format(host_name))
+
+    def execute_bulk_package_action(self,
+                                    name,
+                                    action_name,
+                                    action_value=None,
+                                    action_via='katello_agent',
+                                    action_type='package',
```
abalakh: nitpick: i would place `action_type` before `action_value`, so the call would look like `execute_bulk_package_action(host, 'install', 'package', package_name)`, not `host, 'install', package_name, 'package')`, imo 'install package' sounds more naturally :)

robottelo/ui/hostcollection.py
```diff
@@ -130,3 +130,72 @@ def add_host(self, name, host_name):
...
+        dropdown = "/../../../button[contains(@data-toggle,'dropdown')]"
+        action_dropdown_locator = (action_link[0], action_link[1] + dropdown)
+
+        self.click(action_dropdown_locator)  # because element must be visible
+        self.click(action_link)
+
+        result = self.wait_until_element(
+            locators['contenthost.bulk_actions.remote_action_finished'],
+            timeout=timeout,
+        )  # this alert box fades after 2 seconds
```
abalakh: right, alert appears just after a click and disappears in 2 seconds, no need in custom timeout here (especially for 120 seconds)

robottelo/ui/hostcollection.py
rochacbruno: @abalakh I think it makes sense to wait_until_element with **bst-alert="success"** because this element is available on DOM only after the action button is clicked, if there is something wrong in performing the action that element will never shown in the page and will end in UIError

rochacbruno:
@abalakh @sghai all comments addressed!

sghai:
ACK.

abalakh:
ACK

PR #3662, opened 2016-07-27
Test creation for adding a CR of type rhev

san7ket:
Addition of rhev test class to incorporate rhev CR testcases.
Addition of new section to properties file, to get the rhev compute resource settings.
Addition of new test case to create a CR of type rhev.


elyezer:
@kbidarkar what was the reason to close this?

PR #3668, opened 2016-07-28
Pass the token-prefix option to Testimony

elyezer:
Configure Testimony to use the Robottelo's docstrings token prefix.


elyezer:
This requires a new version of Testimony which is not released yet.

sthirugn:
ACK and merge to trigger automation.

PR #3691, opened 2016-08-09
Added verification if passed to hammer options supported by CLI factory

svtkachenko:
Personally I found it very frustrating to spend time playing around with `hammer` and `make_${entity_name}` methods just to find that passed to method options weren't declared in its body and as result were silently dropped.

Added quick check and logging to `create_object` method that is called by every `make_$entity` method.

Not supported option added:
```
make_architecture({'name': "testme", 'missingopt': "missing"})
2016-08-09 15:41:20 - robottelo.cli.factory - DEBUG - Option(s) set(['missingopt']) not supported by CLI factory. Please check for a typo or update default options
2016-08-09 15:41:22 - robottelo.ssh - INFO - Instantiated Paramiko client 0x7f3a98a58950
2016-08-09 15:41:22 - robottelo.ssh - DEBUG - Connected to [example.com]
2016-08-09 15:41:22 - robottelo.ssh - DEBUG - >>> LANG=en_US.UTF-8  hammer -v -u admin -p changeme --output=csv architecture create --name="testme"
2016-08-09 15:41:24 - robottelo.ssh - DEBUG - <<< stdout
Message,Id,Name
Architecture created,20,testme
```

Only supported options
```
make_architecture({'name': "test"})
2016-08-09 15:38:23 - robottelo.ssh - INFO - Instantiated Paramiko client 0x7f180f13add0
2016-08-09 15:38:23 - robottelo.ssh - DEBUG - Connected to [example.com]
2016-08-09 15:38:23 - robottelo.ssh - DEBUG - >>> LANG=en_US.UTF-8  hammer -v -u admin -p changeme --output=csv architecture create --name="test"
2016-08-09 15:38:25 - robottelo.ssh - DEBUG - <<< stdout
Message,Id,Name
Architecture created,19,test
```

### DIFF
--- a/robottelo/cli/factory.py
+++ b/robottelo/cli/factory.py
@@ -92,6 +92,13 @@ def create_object(cli_object, options, values):
     :return: A dictionary representing the newly created resource.
 
     """
+    if values:
+        diff = set(values.keys()).difference(set(options.keys()))
+        if diff:
+            logger.debug(
+                "Option(s) {0} not supported by CLI factory. Please check for "
+                "a typo or update default options".format(diff)
+            )
     update_dictionary(options, values)
     try:
         result = cli_object.create(options)


robottelo/cli/factory.py
```diff
@@ -92,6 +92,12 @@ def create_object(cli_object, options, values):
     :return: A dictionary representing the newly created resource.
 
     """
+    if values:
+        diff = set(values.keys()).difference(set(options.keys()))
+        logger.debug(
```
abalakh: This way you'll always receive the warning, each call. You should only rise a warning if diff is not empty.

abalakh:
Great idea. For anyone interested - cli factory works only with options defined in specific make_* method as `args` variable, e.g. for [make_activation_key](https://github.com/SatelliteQE/robottelo/blob/master/robottelo/cli/factory.py#L148). If some option is passed and it's not in `args` variable of method (for example when new option was added to hammer and factory became outdated) - it will be simply ignored without any kind of warning or error, making it hard to find the root cause of why is it not working. Such warning would be very handy, thanks for adding it.

abalakh:
Pending comment + logs from both cases:
1) when only supported options were passed 
2) both supported and not supported options passed

abalakh:
ACK

rochacbruno:
ACK.

PR #3692, opened 2016-08-09
Added documentation about ssh module

### LINKED ISSUE #3681: Write documentation for ssh module
robottelo.ssh needs documentation with usage examples

renzon:
Closes #3681

### DIFF
--- a/docs/features/ssh.rst
+++ b/docs/features/ssh.rst
@@ -1,0 +1,80 @@
+SSH (Secure SHell)
+==================
+
+This section explains Robottelo's ssh module.
+
+.. contents::
+
+Introduction
+------------
+
+Robottelo uses ssh extensively to access remote servers.
+Functions from ``robottello.ssh`` make ssh access easier and are
+explained on next sections.
+
+SSHCommandResult
+----------------
+
+``SSHCommandResult`` represents the result of a ssh command.
+It holds ``stdout`` on attribute with same name.
+``stderr`` and ``return_code`` are available the same way.
+An example of typical result is presented bellow::
+
+    SSHCommandResult(
+        stdout=[u'Red Hat Enterprise Linux Server release 7.2 (Maipo)', u''],
+        stderr='',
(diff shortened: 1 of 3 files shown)


renzon:
I felt tempted do add documentation about base.py from cli package but left it for its own section

rochacbruno:
In `ssh` module there are **command** and **execute_command** the only difference is that one creates a new connection since the other uses existing connection.

So it would be nice to add this to the docs.

---

## Executing commands remotelly

There are two functions in `ssh` module to perform commands in remote machine, **command** and **execute_command**.

If you are going to execute only a single command line and you don't have an existing connection you can use **command** example:

```console
output = ssh.command("ls -la /tmp")
```

But if you have some commands to be executed in a sequence and wishes to reuse a single connection you can do

```console
with ssh._get_connection(..) as connection:
    ssh.execute_command('cp /orign /destiny', connection)
    ssh.execute_command('chmod 0777 /destiny', connection)
    ssh.execute_command("echo 'foo' > /destiny/bar")
```

> NOTE: maybe we can rename the **_get_connection** to **get_connection**

renzon:
@rochacbruno didn't know execute_command works this way. I added to the doc.. I put the name without underline on doc and opened this #3693

rochacbruno:
ACK

(I also opened a new issue #3694 )

docs/features/ssh.rst
```diff
@@ -0,0 +1,80 @@
...
+Functions from ``robottello.ssh`` make ssh access easier and are
+explained on next sections.
+
+SSHCommandResult
+----------------
+
+``SSHCommandResult`` represents the result of a ssh command.
+It holds ``stdout`` on attribute with same name.
+``stderr`` and ``return_code`` are available the same way.
+An example of typical result is presented bellow::
```
sghai: s/bellow/below

sghai:
Looks good. Merging.

PR #3731, opened 2016-08-19
Automated Positive Add Roles CLI

### LINKED ISSUE #379: Create a job to create and deploy Katello on an EC2 instance
Will be better to have a job that creates and deploys Katello to an EC2 instance.

The new job should integrates the instance creator with the EC2 katello deploy jobs.

renzon:
close #379

Refactored test which depends on only one user to
UserTestWithCleaningUp so user got deleted on tearDown
 Refactored also Roles so they are created and destroyed
 on once for all tests

### DIFF
--- a/tests/foreman/cli/test_user.py
+++ b/tests/foreman/cli/test_user.py
@@ -22,10 +22,13 @@ When testing email validation [1] and [2] should be taken into consideration.
 @Upstream: No
 """
 import random
+
 from fauxfactory import gen_string
 from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.factory import make_location, make_org, make_role, make_user
+from robottelo.cli.role import Role
 from robottelo.cli.user import User
+from robottelo.config import settings
 from robottelo.datafactory import (
     invalid_emails_list,
     invalid_names_list,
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_user.py
```diff
@@ -22,10 +22,14 @@
 @Upstream: No
 """
 import random
+
 from fauxfactory import gen_string
+
```
sghai: this blank line can be removed.

tests/foreman/cli/test_user.py
```diff
@@ -295,20 +214,22 @@ def test_negative_create_with_invalid_username(self):
...
-                             gen_string('html')):
-            with self.subTest(invalid_name):
+        invalid_names = (
+            '', 'space {0}'.format(gen_string('alpha')),
+            gen_string('alpha', 101),
+            gen_string('html')
+        )
+        for invalid_name in invalid_names:
+            with self.subTest(invalid_name), self.assertRaises(
+                    CLIReturnCodeError):
```
rochacbruno: I am not sure if using this nested context managers in this way will or not mask errors.
if in example `gen_string` fails with `FooError` the assertRaises will catch it? whould not be better to use assertRaises(CLIError) specific on the context of `Use.create` as it was before?

elyezer: assertRaises will just look for the specified exception and will raise AssertionError depending if the exception was not raised. So it won't catch any other exception.

I prefer to just wrap with the assertRaises just the statements we are expecting to fail, that makes the code more clear on where we expect the exception to happen. Using nested context manager is good for decreasing the indentation level, but I think having the meaning of where we expect a failure wins.

elyezer: If just the create call was within the nested context manager then it would be fine.

renzon: It woud be a problem only if FooError is subclass of CLIError. See:

```python
    def test_raise(self):
        with self.subTest('foo'):
            raise IndexError()
            with self.assertRaises(ValueError):
                pass

#Result:
self = <tests.robottelo.test_api_utils.UtilsTestCase testMethod=test_raise>

    def test_raise(self):
        with self.subTest('foo'):
>           raise IndexError()
E           IndexError

test_api_utils.py:25: IndexError
```

```python

    def test_raise(self):
        with self.subTest('foo'), self.assertRaises(ValueError):
            raise IndexError()


self = <tests.robottelo.test_api_utils.UtilsTestCase testMethod=test_raise>

    def test_raise(self):
        with self.subTest('foo'), self.assertRaises(ValueError):
>           raise IndexError()
E           IndexError

test_api_utils.py:25: IndexError
```

Probably assertRaises create the context manage and make the Exception assertion on its __exit__ method.

rochacbruno: > but I think having the meaning of where we expect a failure wins.
I agree with @elyezer, using the `assertRaises` only in the context where it is expected is better for meaning. So I vote to leave this only in the line where `.create` is performed.

renzon:
Closing because I am going to add this in #3733

PR #3802, opened 2016-09-09
Created different classes of CLI errors

### LINKED ISSUE #3790: Check stderr for db errors presence
When hammer command fails, it should show error message explaining the root cause (e.g. `Error: Request Timeout`, `Error: Could not find organization, please set one of options --organization, --organization-label, --organization-id.` etc.), but sometimes instead of error message it shows DB error which is a bug that we should capture.
As per our friday's discussion, i18n testing is not our priority right now, so instead of comparing error message of each hammer command that should fail with some expected message (and creating a huge list of expected error messages, plus reworking all our negative tests), we can try to check db error presence in error message - for example, by forming up some kind of keywords or regular expressions list to compare with. E.g.  `INSERT INTO`, `LEFT OUTER JOIN`, `SELECT(.*)FROM(.*)WHERE(.*)` etc.
All the executed hammer commands are getting analyzed in `cli.Base._handle_response()` method so this should be a good place to interfere. Also negative tests are expecting `CLIReturnCodeError` to be risen, so in case db error was found in stderr - any another exception should be risen and then we won't miss it.
Few examples of db errors in stderr:
```
# hammer role filters
PGError: ERROR:  invalid input syntax for integer: ""
LINE 1: ..."id" = "filters"."role_id" WHERE (("roles"."id" = '')) ORDER...
                                                             ^
: SELECT  "filters"."id" AS t0_r0, "filters"."search" AS t0_r1, "filters"."role_id" AS t0_r2, "filters"."created_at" AS t0_r3, "filters"."updated_at" AS t0_r4, "filters"."taxonomy_search" AS t0_r5, "roles"."id" AS t1_r0, "roles"."name" AS t1_r1, "roles"."builtin" AS t1_r2, "roles"."permissions" AS t1_r3 FROM "filters" LEFT OUTER JOIN "roles" ON "roles"."id" = "filters"."role_id" WHERE (("roles"."id" = '')) ORDER BY role_id, filters.id LIMIT 20 OFFSET 0
```
```
\Error: PG::Error: ERROR:  duplicate key value violates unique constraint "index_katello_subscriptions_on_cp_id"
DETAIL:  Key (cp_id)=(ff80808156a722610156b7811da51b17) already exists.
: INSERT INTO "katello_subscriptions" ("cp_id", "created_at", "updated_at") VALUES ($1, $2, $3) RETURNING "id"
```

renzon:
close #3790

### DIFF
--- a/robottelo/cli/base.py
+++ b/robottelo/cli/base.py
@@ -1,6 +1,7 @@
 # -*- encoding: utf-8 -*-
 """Generic base class for cli hammer commands."""
 import logging
+import re
 
 from robottelo import ssh
 from robottelo.cli import hammer
@@ -11,8 +12,8 @@ class CLIError(Exception):
     """Indicates that a CLI command could not be run."""
 
 
-class CLIReturnCodeError(Exception):
-    """Indicates that a CLI command has finished with return code, different
+class CLIBaseError(Exception):
+    """Indicates that a CLI command has finished with return code different
     from zero.
 
     :param return_code: CLI command return code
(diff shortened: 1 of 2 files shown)


tests/robottelo/test_cli.py
```diff
@@ -1,7 +1,8 @@
 import six
 import unittest2
 
-from robottelo.cli.base import Base, CLIReturnCodeError, CLIError
+from robottelo.cli.base import (Base, CLIReturnCodeError, CLIError,
+                                CLIBaseError, CLIDataBaseError)
```
rochacbruno: I prefer to use the style:

```python
from robottelo.cli.base import (
    Base,
    CLIError,
    ...
)
```

It is just a matter of style, but also helps when we need to add/remove imports and avoid long line indentation.
```

**Comment by rochacbruno** [comment]
ACK

File: tests/robottelo/test_cli.py (hunk 1 above)
Comment (renzon) [inline tests/robottelo/test_cli.py]: I prefer that style too. Pycharm doesn't so have to manualy fix that.

File: robottelo/cli/base.py [hunk 2]
```diff
@@ -37,8 +38,26 @@ def __str__(self):
 
     def __repr__(self):
         """Include return_code, stderr and msg to improve logging"""
-        tmpl = u'CLIReturnCodeError(return_code={!r}, stderr={!r}, msg={!r}'
-        return tmpl.format(self.return_code, self.stderr, self.msg)
+        return u'{}(return_code={!r}, stderr={!r}, msg={!r}'.format(
```
omaciel: Since the returned artifact has included one more element (`type(self).__name__`), the docstring needs to be updated :)

sghai:
ACK.. LGTM

PR #3820, opened 2016-09-14
Implemented tests for Smart Variables

svtkachenko:
Depends on SatelliteQE/nailgun#315
```
% nosetests tests/foreman/api/test_variables.py
.....S..S..SSSSSS........S............SSSSS
Ran 43 tests in 114.762s

OK (SKIP=14)
```

### DIFF
--- a/tests/foreman/api/test_variables.py
+++ b/tests/foreman/api/test_variables.py
@@ -15,16 +15,50 @@
 
 @Upstream: No
 """
+from random import choice
 
-from robottelo.decorators import run_only_on, stubbed, tier1, tier2, tier3
+from fauxfactory import gen_integer, gen_string
+from nailgun import entities
+from requests import HTTPError
+
+from robottelo import ssh
+from robottelo.config import settings
+from robottelo.datafactory import invalid_values_list, valid_data_list
+from robottelo.decorators import (
+    run_only_on,
+    skip_if_bug_open,
+    stubbed,
+    tier1,
+    tier2,
+)
 from robottelo.test import APITestCase
 
 
 class SmartVariablesTestCase(APITestCase):
     """Implements Smart Variables tests in API"""
 
+    @classmethod
+    def setUpClass(cls):
+        """Import some parametrized puppet classes. This is required to make
+        sure that we have data to be able to perform interactions with smart
+        class variables.
+        """
+        super(SmartVariablesTestCase, cls).setUpClass()
+        cls.host_name = settings.server.hostname
+        ssh.command('puppet module install --force puppetlabs/ntp')
+        cls.env = entities.Environment().search(
+            query={'search': 'name="production"'}
+        )
+        if len(cls.env) == 0:
+            raise Exception("Environment not found")
+        cls.env = cls.env[0]
+        cls.proxy = entities.SmartProxy(name=cls.host_name).search()[0]
+        cls.proxy.import_puppetclasses(environment=cls.env)
+        cls.puppet = entities.PuppetClass().search(
+            query={'search': 'name="ntp"'}
+        )[0]
+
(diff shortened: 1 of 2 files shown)


tests/foreman/api/test_variables.py
```diff
@@ -849,31 +1269,16 @@ def test_positive_update_hidden_value_in_variable(self):
...
-        @steps:
-
-        1. Create variable with valid type and empty value
-        2. Set 'Hidden Value' to true
-
-        @assert:
-
-        1. The 'hidden value' flag set to True
-        2. The default value is empty after hide
-
```
svtkachenko: Point 1 is covered in test_positive_hide_variable_default_value
Point 2 is not relevant for API as default_value always available.
So stub was removed.

tests/foreman/api/test_variables.py
```diff
@@ -65,69 +107,142 @@ def test_positive_delete_smart_variable_by_id(self):
         @steps: Delete a smart Variable by id
 
         @assert: The smart Variable is deleted successfully
+        """
+        smart_variable = entities.SmartVariable(
+            puppetclass=self.puppet).create()
```
sghai:
```python
smart_variable = entities.SmartVariable(
        puppetclass=self.puppet
).create()
```

sghai:
ACK. very nicely done.

Only question I've about matching the exact error message from raised exception, are we going to use this approach for api tests as well? Not sure if I'm missing something.

svtkachenko:
@sghai I've found that negative API tests also prone to false positives because we can get DB errors. Looks like that hammer is just a wrapper for API calls (like nailgun) so filtering DB messages for hammer is partial solution.

sghai:
IRC chat:
```
stkachen> sghai: as I understand we decided to filter DB messages for CLI instead of matching error messages, however API also affected.
<sghai> stkachen, yeah.. agreed. but we wanted to avoid to use exact error messages.. as if anything change in message then test may fail..
<sghai> stkachen, for now I'm fine with your approach.. we will see if we can comeup with same implementation for api test as we found for hammer like matching db error etc
```

sghai:
ACK

tests/foreman/api/test_variables.py
svtkachenko: Updated.
Not sure why it is not hidden after re-commit.

sghai:
@abalakh @oshtaier @rochacbruno @renzon @jyejare : ping guys.. need one more reviewer.

tests/foreman/api/test_variables.py
```diff
@@ -15,16 +15,51 @@
...
+        sure that we have data to be able to perform interactions with smart
+        class variables.
+        """
+        super(SmartVariablesTestCase, cls).setUpClass()
+        cls.host_name = settings.server.hostname
+        ssh.command('puppet module install --force puppetlabs/ntp')
+        cls.env = entities.Environment().search(
+            query={'search': 'name="production"'}
+        )
+        if len(cls.env) < 0:
```
renzon: Would it be len(cls.env) == 0? If env is a list it will never have negative len.

tests/foreman/api/test_variables.py
```diff
@@ -15,16 +15,51 @@
...
+            raise Exception("Environment not found")
+        else:
```
renzon: If exception above is raised the code here will never be executed, so you could remove the else ;)

tests/foreman/api/test_variables.py
svtkachenko: Shame on me :)
Updated.

renzon: @svtkachenko it happens on big PR's like this, don't worry. At least review process is working ;)

PR #3848, opened 2016-09-21
Added "make can-i-push?" and "make install-commit-hook"

rochacbruno:
# before commiting run

`make can-i-push?`

It will check flake8, docstrings, uuids and also run test-robottelo

# automating with pre-commit hook

run

`make install-commit-hook`  

> NOTE:  run the above this only once! it will write contents to `.git/hooks/pre-commit` locally

And then the checks will be automatically fired before each commit and you dont have to run it manually.


sghai:
ACK

PR #3879, opened 2016-10-03
test for hammer proxy import-classes, fixes #1027

### LINKED ISSUE #1027: hammer proxy import-classes
hammer proxy import-classes --help
Usage:
    hammer proxy import-classes [OPTIONS]

Options:
    --dryrun                      Do not run the import
    --environment ENVIRONMENT_NAME  
    --environment-id ENVIRONMENT_ID  
    --except EXCEPT               Optional comma-deliminated string containing either 'new,updated,obsolete'
                                  used to limit the import_puppetclasses actions
    --id ID                        
    --name NAME                   Name to search by
    -h, --help                    print help

pondrejk:
test result:
```
nosetests tests/foreman/cli/test_capsule.py:CapsuleTestCase.test_positive_import_puppet_classes -v
Import puppet classes from proxy ... ok

Ran 1 test in 26.937s

OK
```

### DIFF
--- a/tests/foreman/cli/test_capsule.py
+++ b/tests/foreman/cli/test_capsule.py
@@ -166,6 +166,24 @@ class CapsuleTestCase(CLITestCase):
         # Add capsule id to cleanup list
         self.addCleanup(capsule_cleanup, proxy['id'])
 
+    @skip_if_not_set('fake_capsules')
+    @run_only_on('sat')
+    @tier1
+    def test_positive_import_puppet_classes(self):
+        """Import puppet classes from proxy
+
+        @id: 42e3a9c0-62e1-4049-9667-f3c0cdfe0b04
+
+        @Assert: Puppet classes are imported from proxy
+        """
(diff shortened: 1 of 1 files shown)


pondrejk:
@svtkachenko good point, but it turns out I can't do that with the fake capsule. I guess this would require a real capsule on some compute resource, probably we should discuss some global solution with other folks @SatelliteQE, because other tests will need this as well (e.g. capsule provisioning)

abalakh:
@svtkachenko ping?

pondrejk:
@abalakh, @svtkachenko, There's an open question about incorporating capsules to tests at #3893, I didn't get to setting up a vm capsule yet, as that's something that would probably affect some future tests, so I wanted to gather ideas there

svtkachenko:
Well, if we agreed that this test can be improved later I'd suggest to merge it now and create issue to track this improvent.
@pondrejk Can you please create the issue?

PR #3883, opened 2016-10-04
Fix lru_cache import path; fix travis failure

abalakh:
Latest sphinx release 1.4.8 (altogether with new cachetools release 2.0.0, both released 1-3 days ago) uncovered issue with wrong import path - `lru_cache` from `cachetools` is located under `cachetools.func` path, not just `cachetools`. Fixing this to make green travis builds

### DIFF
--- a/robottelo/helpers.py
+++ b/robottelo/helpers.py
@@ -18,7 +18,7 @@ from robottelo.config import settings
 if six.PY3:  # pragma: no cover
     from functools import lru_cache  # noqa
 else:  # pragma: no cover
-    from cachetools import lru_cache  # noqa
+    from cachetools.func import lru_cache  # noqa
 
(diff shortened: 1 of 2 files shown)


abalakh:
Note that this failure was uncovered today and has nothing to do with previous one, which we faced yesterday - yesterday some broken `pycparser` was released (or i'd say uploaded by accident, as it's already removed) which is dependency for `pycrypto`.
More info [here](https://github.com/eliben/pycparser/issues/148) and [here](https://github.com/pyca/cryptography/issues/3187).

abalakh:
Retriggered all failed travis builds for PRs, let's see how it goes.

rplevka:
needs cherrypick to 6.2.z !

PR #3888, opened 2016-10-04
Fix end_to_end tests - GOOGLE_CHROME_REPO doesn't exists anymore.

rochacbruno:
Google Chrome RPM repo is returning 404, Google may shutdown the or changed the URL.

Replacing with Pulp/fixture/rpm/ repo as CUSTOM_RPM_REPO for endtoend tests

### DIFF
--- a/robottelo/constants.py
+++ b/robottelo/constants.py
@@ -318,7 +318,9 @@ FILTER_TYPE = {
 }
 
 DOCKER_REGISTRY_HUB = u'https://registry-1.docker.io'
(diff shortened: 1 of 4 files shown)


rochacbruno:
@omaciel this one needs cherrypick to 6.2.z no? (the GOOGLE_REPO is used there)

rochacbruno:
@omaciel that Google Repo is also used in 6.1.z but we cant cherrypick as the module name has been changed form `smoke`to `end_to_end`

Should we fix that url on all branches?

omaciel:
@rochacbruno for now let's leave `6.1.z` alone as I don't think folks are running tests against it before giving us builds.

PR #3889, opened 2016-10-04
Fix lru_cache import path

omaciel:
Newer version of `cachetools` requires this change for the `6.2.z` branch.


omaciel:
Talked to @rochacbruno and he pointed out that our `.travis.yaml` file for the `6.2.z` branch is not using the 'frozen' requirements file. Fixing that, imho, would be the best solution. We should also do the same for the `6.1.z` branch

rochacbruno:
Closing this one in favor of #3890

PR #3890, opened 2016-10-04
Travis should use requirements-freeze.txt

rochacbruno:


.travis.yml
```diff
@@ -4,7 +4,7 @@ env:
     - TOXENV=py27
     - TOXENV=py34
 install:
-    - pip install -r requirements.txt coveralls flake8 sphinx testimony tox
+    - pip install -r requirements-freeze.txt coveralls flake8 sphinx testimony tox
```
omaciel: I think all other packages can be removed since they should be included in `requirements-freeze.txt`?

omaciel:
@rochacbruno btw, we should do the same for `6.1.z`

rochacbruno:
@omaciel I figured out that tox.ini also needs the change, lets wait the build goes green and then I send it to 6.1.z

tox.ini
```diff
@@ -2,7 +2,7 @@
 envlist = py27,py34
 [testenv]
 deps=
-    -rrequirements.txt
+    -rrequirements-freeze.txt
     pytest-cov
     py27: mock
```
omaciel: Looks like we can remove `mock` here?

rochacbruno:
@omaciel BTW: I don't get why we have `-freeze.txt` file instead of having just `requirements.txt` with different versions across branches.

omaciel:
@rochacbruno I honestly don't know... I didn't name them this way, but feel free to fix them

PR #3895, opened 2016-10-05
removing hook for BZ1325995 due to unsupported BZ

rplevka:


tests/foreman/endtoend/test_api_endtoend.py
```diff
@@ -923,8 +923,8 @@ def test_positive_ping(self):
         #    …
         # }, u'status': u'ok'}
         services = response['services']
-        if bz_bug_is_open('1325995'):
-            services.pop('foreman_auth')
+        # remove foreman_auth service - BZ1325995
```
rplevka: let's keep a reference to the BZ for future generations :)

jyejare:
ACK

Issue #3907, opened 2016-10-10
repo download policy tests (api/cli/ui) need cherry pick to 6.2.z branch


pondrejk:
Applies to commits:
https://github.com/SatelliteQE/robottelo/pull/3771/commits
https://github.com/SatelliteQE/robottelo/pull/3737/commits
https://github.com/SatelliteQE/robottelo/pull/3740/commits
https://github.com/SatelliteQE/robottelo/pull/3734/commits

pondrejk:
I created a PR for cli and api tests https://github.com/SatelliteQE/robottelo/pull/3908
I'm not sure about the ui tests, since the commit (https://github.com/SatelliteQE/robottelo/pull/3771/commits/862b64523cac17fe85c22562c4e9bc007b294f1d) is mixed with other changes, @rochacbruno can you please check if this can be cherry-picked?

rochacbruno:
@pondrejk the UI commit can't be cherry picked because `locators.py` have been promoted to a module `locators/*`

I think I can copy/paste the test cases if needed. @JacobCallahan is it needed to backport the UI tests to 6.2.z?

JacobCallahan:
@rochacbruno yes. as of right now, this functionality is in the 6.2.z branch.

rochacbruno:
covered in #3910

sghai:
@rochacbruno : I think this issue can be closed now ?

PR #3908, opened 2016-10-10
cherry picked repo download policy tests (api/cli), related to #3907

pondrejk:


pondrejk:
For tests see the standalone automation 231 (cli) and 232 (api), all dw policy related tests passed.

rplevka:
ACK

lpramuk:
if we are cherrypicking can we have branch in brackets at start: 
```[6.2.z] (cherrypick) We need this``` 

- so we can quickly see what is what based on titles ;-)
- cherrypick can be optional word as we mainly cherrypick to older branches

PR #3913, opened 2016-10-11
Added another UI errata test and fixed compatibility with Python 3

### LINKED ISSUE #3726: Make `tests/foreman/ui/test_errata.py` compatible with both Python 2 and 3
The `nested` from `contextlib` is not available anymore on Python 3. That is causing an error when running the tests using Python 3.

```console
$ python --version
Python 3.4.3
$ python -c 'from contextlib import nested'
ImportError: cannot import name 'nested'
```

abalakh:
* The `nested` from `contextlib` is not available anymore on Python 3. And, basically, as we do not support python2.6 we don't need `nested`at all so i've removed it.
Closes #3726
* Implemented another stub, although it appears to be blocked by a BZ.

### DIFF
--- a/tests/foreman/ui/test_errata.py
+++ b/tests/foreman/ui/test_errata.py
@@ -15,7 +15,6 @@
 @Upstream: No
 """
 
-from contextlib import nested
 from fauxfactory import gen_string
 from nailgun import entities
 from robottelo.api.utils import enable_rhrepo_and_fetchid
@@ -46,6 +45,7 @@ from robottelo.constants import (
 )
 from robottelo.decorators import (
     run_in_one_thread,
+    skip_if_bug_open,
     skip_if_not_set,
     stubbed,
     tier2,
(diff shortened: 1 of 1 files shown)


abalakh:
No test results as the new test is skipped and another affected test is failing on jenkins for a different reason.

tests/foreman/ui/test_errata.py
```diff
@@ -260,10 +260,8 @@ def test_positive_apply_for_all_hosts(self):
 
         @CaseLevel: System
         """
-        with nested(
-                VirtualMachine(distro=DISTRO_RHEL7),
-                VirtualMachine(distro=DISTRO_RHEL7)
-        ) as (client1, client2):
+        with VirtualMachine(distro=DISTRO_RHEL7) as client1, VirtualMachine(
+                distro=DISTRO_RHEL7) as client2:
```
sghai: :+1:

sghai:
ACK

svtkachenko:
ACK with small nitpick.

Issue #3868, opened 2016-09-29
@skip_if_bz_bug_open logic should involve version check


rplevka:
since we support multiple versions of katello (satellite), and not all bug fixes are being backported to all versions, the following scenario might (and most probably **will**) occur:

1.) A nasty bug is discovered and reported on `sat-6.2.2`
2.) An appropriate test with `@skip_if_bz_bug_open('123456')` is being created and pushed to `master`, and `6.2.z` branches.
3.) Devels mark the target version as `sat-6.3`, deliver a fix and close the BZ.
4.) `@skip_if_bz_bug_open` no longer skips the bug.
5.) test passes on `6.3.*` but starts to fail on `6.2.z` as the fix hasn't been (and is never going to be) backported.

Shall we start deleteing such test cases from version branches which are not supported, or at least start marking them as `@stubbed` or shall we implement some version check mechanism (check the BZ flags, etc.) to the BZ decorators?

discuss ;)

omaciel:
Great point @rplevka ! I need to think a bit now :)

elyezer:
That is a really good point!

lpramuk:
Blocking issue is #3896 as there is wrong logic in ```bz_bug_is_open()```

rplevka:
i'll keep the list of tests (which should be updated once version-check feature is implemented) here:

branch | BZ# | test case
--- | --- | ---
6.1.z | 1138553 | cli.test_user.UserTestCase.test_positive_remove_role

abalakh:
Bumping priority. Another bug which landed 6.3 is 1324508, and as it's not in 6.2 we've got 22 extra failures (11 per rhel version) for 6.2.z - basically all ui bookmarks tests are failing as the bug is striking at test setup.

rplevka:
I was also thinking of implementing a new kind of decorator: `skip_wontfix`.

elyezer:
One of the configuration Pulp Smash receive is the expected version of the system, that allows:

* Make sure the system under testing reports the right version number
* Skip bugs based on the version
* Run or skip tests based on the version

Worth noting that Pulp Smash does not have helper classes like Robottelo does, for example there is no CLI class wrapper based on entity. All commands are run on the test directory (we do have functions to avoid repeating same chunks of code over and over again). Also we didn't faced any API or CLI changes yet.

With all that said I think one thing that robottelo could have is a config with the version and then make the bug skipping logic to consider that information when checking if the test should be skipped or not.

[...]

PR #3957, opened 2016-11-02
[6.2.z] Fix Smart Class Parameters - use custom puppet repo

svtkachenko:
Use custom puppet modules repo for puppet related tests instead of puppet forge.
Should fix test failures for CLI/Smart Class Parameters.

One error is because of bug that was fixed in 6.3 but not in 6.2 and `skip_if_bug_open` does not skip that test. What should I do with this test?
```python
% nosetests tests/foreman/cli/test_classparameters.py
.SSSS..........SSSSSS..........E...........
Ran 43 tests in 901.328s
FAILED (SKIP=10, errors=1)
```


tests/foreman/cli/test_classparameters.py
```diff
@@ -39,6 +46,8 @@
 )
 from robottelo.test import CLITestCase
 
+from nailgun import entities
```
abalakh: please sort alphabetically

tests/foreman/cli/test_classparameters.py
```diff
@@ -123,27 +132,50 @@ def setUpClass(cls):
...
-        be able to work with unique entity for each specific test. Raise an
-        exception and skip all tests in case not enough parameters returned.
+        be able to work with unique entity for each specific test.
         """
         super(SmartClassParametersTestCase, cls).setUpClass()
-        cls.host_name = settings.server.hostname
-        ssh.command('puppet module install --force puppetlabs/ntp')
-        cls.env = Environment.info({u'name': 'production'})
-        Proxy.importclasses({
+        cls.puppet_module = "robottelo/cli_test_classparameters"
```
abalakh: You never use this variable directly, each call splits the value and uses only part of it. Maybe it's better to define separate variables?

tests/foreman/cli/test_classparameters.py
```diff
@@ -123,27 +132,50 @@ def setUpClass(cls):
...
         cls.sc_params_ids_list = [
-            sc_param['id'] for sc_param in sc_params_list]
+            sc_param['id'] for sc_param in cls.sc_params_list]
+
+    def setUp(self):
+        """Checks that there is at least one not overridden
+        smart class parameter before executing test.
+        """
+        if len(self.sc_params_list) == 0:
+            raise Exception("Not enough smart class parameters. Please"
```
abalakh: nitpick: missing space after word `Please`. Your string will result in `Pleaseupdate puppet module.`

tests/foreman/cli/test_classparameters.py
```diff
@@ -166,11 +198,14 @@ def test_positive_list_parameters_by_host_name(self):
             'id': sc_param_id,
         })
         self.assertEqual(sc_param['override'], True)
+        organization = entities.Organization(id=self.org['id']).search()[0]
```
abalakh: Why do you use searching here? I bet you meant 
```python
organization = entities.Organization(id=self.org['id']).read()
```
But that shouldn't be the issue - you don't even need to read an organization, according to tests the only usage of organization is during Host initialization. nailgun works with entity fields in 2 ways - you can either set them to entity or with entity id. As you already have entity id and there's no other usages of entity - id should be enough:
```python
host = entities.Host(organization=self.org['id']).create()
```

tests/foreman/cli/test_classparameters.py
abalakh: Or as an option - a dictionary. With keys `name` and `author`.

tests/foreman/cli/test_classparameters.py
abalakh: Comment is applicable to other tests too

tests/foreman/cli/test_classparameters.py
```diff
@@ -123,27 +132,50 @@ def setUpClass(cls):
...
-                               ' work with in provided puppet class')
         cls.sc_params_ids_list = [
-            sc_param['id'] for sc_param in sc_params_list]
+            sc_param['id'] for sc_param in cls.sc_params_list]
+
+    def setUp(self):
+        """Checks that there is at least one not overridden
+        smart class parameter before executing test.
+        """
+        if len(self.sc_params_list) == 0:
```
abalakh: call `super`

rplevka:
@svtkachenko, first of all, thanks for looking into the failures! 

>One error is because of bug that was fixed in 6.3 but not in 6.2 and skip_if_bug_open does not skip that test. What should I do with this test?

Please add it to my list (if it's not yet there): https://github.com/SatelliteQE/robottelo/issues/3868#issuecomment-252241893

Or, if you're brave enough, clone the Bugzilla, for each version (6.1.z, 6.2.z) and update the ID of the bug in the code.

abalakh:
ACK

sghai:
ACK

PR #3982, opened 2016-11-10
[6.2.z] cherrypick -- added hostgroup clone api test, fixed lru_cache path

pondrejk:
Cherry picked 2 commits adding a test for hostgroup clone API call. Test result:

``` 
nosetests tests/foreman/api/test_hostgroup.py:HostGroupTestCase.test_positive_clone
.
Ran 1 test in 7.942s

OK
```

Also fixed the lru_cache path in a separate commit, this was fixed in master a while ago in https://github.com/SatelliteQE/robottelo/pull/3883

### DIFF
--- a/tests/foreman/api/test_hostgroup.py
+++ b/tests/foreman/api/test_hostgroup.py
@@ -198,6 +198,24 @@ class HostGroupTestCase(APITestCase):
                 ).create()
                 self.assertEqual(name, hostgroup.name)
 
+    @tier1
+    def test_positive_clone(self):
+        """Create a hostgroup by cloning an existing one
+
+        @id: 44ac8b3b-9cb0-4a9e-ad9b-2c67b2411922
+
+        @assert: A hostgroup is cloned with new name
+        """
+        hostgroup = entities.HostGroup(
+            location=[self.loc],
+            organization=[self.org],
+        ).create()
+        hostgroup_cloned_name = gen_string('alpha', 6)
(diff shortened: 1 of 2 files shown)


tests/foreman/api/test_hostgroup.py
```diff
@@ -198,6 +198,24 @@ def test_positive_create_with_name(self):
...
+        """
+        hostgroup = entities.HostGroup(
+            location=[self.loc],
+            organization=[self.org],
+        ).create()
+        hostgroup_cloned_name = gen_string('alpha')
+        hostgroup_cloned = entities.HostGroup(
+            id=hostgroup.id
+        ).clone(data={'name': hostgroup_cloned_name})
+        self.assertEqual(hostgroup_cloned['name'], hostgroup_cloned_name)
```
omaciel: This has nothing to do with this PR but... can we file an issue to improve this test by asserting that the entire object is equal to the cloned one? Checking that only the name matches assumes that everything else also matches and I don't think we should make assumptions.

rochacbruno: IT can be done since we merge the PR which adds the `to_json` method, so we can override the `__eq__` making that special method to compare the `set` of two entities values as in `Host(id=1) == Host(id=1) -> True`

rochacbruno:
Created #3984 to address @omaciel comment

rochacbruno:
ACK

PR #3989, opened 2016-11-11
flake8 version bump (to the same version as 6.2.z has)

lpramuk:
We need similar (or the same) version of linter across branches and compared to version what you have locally.

Otherwise builds are failing for some branches while they are green at your machine.


lpramuk:
@rochacbruno I was required to update also ```mccabe=0.5.2``` as it failed with 0.4.0

renzon:
ACK

rochacbruno:
ACK

PR #4000, opened 2016-11-17
[Cherry pick to 6.2.z] Automated host listing its own errata

### LINKED ISSUE #3950: Host can't list its own errata

renzon:
close #3950

(cherry picked from commit 3cf7617)

[Bz](https://bugzilla.redhat.com/show_bug.cgi?id=1351040) is verified only on 6.3, still not working on 6.2.4, besides target milestone is 6.2:

```console
E           CLIReturnCodeError: Command "host errata list" finished with return_code 70
E           stderr contains following message:
E           [ERROR 2016-11-17 11:00:10 Exception] undefined method `installable_errata' for nil:NilClass
```
On 6.3:

```console

============================= 39 tests deselected ==============================
=================== 1 passed, 39 deselected in 16.51 seconds ===================
```

### DIFF
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -16,6 +16,7 @@
 """
 from fauxfactory import gen_mac, gen_string
 from nailgun import entities
+from robottelo import ssh
 from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.factory import (
     make_activation_key,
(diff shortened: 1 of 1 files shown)


abalakh:
So we'll add a test to 6.2.z which will just fail? Is it ok?

renzon:
@abalakh It seems to me we shouldn't as well. With the new process we would clone the BZ. So my main intention with this PR is get more discussion about it.

@omaciel what your thought? Feel free to invite more people or take this conversation on a more appropriate plate (maybe the doc about bz process).

sghai:
I'm in favor of closing this PR since fix is only available on 6.3 and tbh, this test is very simple one which actually doesn't complete the actual scenario because we can't self registered automation servers.

omaciel:
@renzon let's schedule time to talk about this.

PR #4003, opened 2016-11-18
As per consultation: This test case is incorrect or obsolete.  You ca…

ldjebran:
As per consultations: This test case is incorrect or obsolete.  You cannot refresh an errata in a content view.
Test deleted


ldjebran:
@sghai , @abalakh please review

abalakh:
@ldjebran i guess it should be cherry-picked to 6.2.z too, right?

PR #4132, opened 2017-01-04
[ DO NOT MERGE] Added populate module for data population and validation

### LINKED ISSUE #1986: Upgrade Test: Upgraded Satellite loses no data upon upgrade.
@Test: Server: Upgraded Satellite loses no data upon upgrade.
@Setup: Follow procedures to populate a previous GA version with selected content; upgrade this system.
1.  Via $method (manual examination, automated comparison, db forensics/health-check), assure that all populated data remains exposed to user; any newly opened ports that might exist on upgraded instance do in fact get opened on an upgrade.
   @Assert: No lost and/or irretrievable data.
   @Status: Manual

### LINKED ISSUE #4004: Create Populate and Validation Script
Connected to #1986

Create a script to populate a satellite and aldo validates data.

rochacbruno:
Closes #4004 
Closes #1986  

Populate
========

This new module adds tools and commands to **populate** and **validate** entities in the system based in **YAML** file.

Commands
--------
Using `$ manage` in the robottelo environment and root directory
In the subgroup `data` you can find the ``populate`` and ``validate`` commands.
That commands are used to read data description from YAML file and
populate the system or validate populated entities.

Read more in specific documentation in `populate </api/robottelo.populate>`

Having ``test_data.yaml`` with the following content.

  "By default ``tests/foreman/data`` is the path to look for yaml files
  but you can include a leading :code:`/` to the path ``/tmp/file.yaml``"

```yaml

    vars:
      org_label_suffix = inc
    actions:
      - model: Organization
        log: The first organization...
        register: org_1
        data:
          name: MyOrg
          label: MyOrg{{org_label_suffix}}
```

To populate the system

```console
    (robottelo_env)[you@host robottelo]$ manage data populate test_data.yaml -v
    2017-01-04 04:31:17 - robottelo.populate.base - INFO - CREATE: The first organization...
    2017-01-04 04:31:19 - robottelo.populate.base - INFO - search: Organization {'query': {'search': 'name=MyOrg,label=MyOrg'}} found unique item
    2017-01-04 04:31:19 - robottelo.populate.base - INFO - create: Entity already exists: Organization 36
    2017-01-04 04:31:19 - robottelo.populate.base - INFO - registry: org_1 registered
```

To validate the system

```console

    (robottelo_env)[you@host robottelo]$ manage data validate test_data.yaml
    (robottelo_env)[you@host robottelo]$ echo $?
    0  # system validated else 1
```

Use :`$ manage data --help` and :`$ manage populate --help` for more info

Decorator
------------

Other way to use populate is via decorator, it is useful to decorate a test_case
forcing a populate or validate operation to be performed.

```python

  from robottelo.populate.decorators import populate_with

  class TestCase(unittest.TestCase):

      @populate_with('filename.yaml')
      def test_anything_that_needs_data(self):
          """Now write a test"""
```

The YAML data file
------------------

In the YAML data file it is possible to specify 3 sections, ``config``, ``vars`` and ``actions``.

## config

The config may be used to define special behavior of populator and its keys are:

- populator

  The name of the populator defined in ``populators``
- populators

  The specification of populator modules to be loaded
- verbose

  The verbosity of logging 0, 1 or 2, it can be overwritten with -vvv in commands.

example:

```yaml

    config:
      verbose: 3
      populator: api
      populators:
        api:
          module: robottelo.populate.api.APIPopulator
        cli:
          module: robottelo.populate.cli.CLIPopulator
```

## vars

Variables to be available in the rendering context of the YAML data
every var defined here is available to be referenced using ``Jinja`` syntax in
any action.

```yaml

      vars:
        admin_username: admin
        admin_password: changeme
        org_name_list:
          - company7
          - company8
        prefix: aaaa
        suffix: bbbb
        my_name: me
```

## actions

The actions is the most important section of the YAML, it is a list of actions
being each action a dictionary containing special keys depending on the action type.

The action type is defined in ``action`` key and available actions are:

Actions are executed in the defined order and order is very important because
each action can ``register`` its result to the internal registry to be referenced
later in any other action.

**CRUD ACTIONS**

Crud actions takes a ``model`` argument, any from ``nailgun.entities`` is valid.

- create (the default)

  Creates a new entity if not exists, else gets existing.
- update

  Updates entity with provided ``data`` by ``id`` or unique search
- delete

  deleted entity with ``id`` or unique search

**SPECIAL ACTIONS**

- echo

  Logs and print output to the console
- register

  Register a variable in the internal registry
- unregister

  removes a variable from register
- assertion

  perform assertion operations, if any fails returns exit code 1

## Dynamic Data

There are some ways to fetch dynamic data in action definitions, it depends
on the action type.

For any key you can use ``Jinja`` to provide a dynamic value as in::

```yaml
  value: "{{ get_something }}"
  value: "{{ fauxfactory.gen_string('alpha')}}"
  value: user_{{ item }}
```

For some actions you can provide a ``data`` key, that data is used to create
new entities and also to perform searches or build the action function.

Every ``data`` key accepts 4 special reference directives in its sub-keys.

- from_registry

  Gets anything from registry::
    ```yaml
    data:
      organization:
        from_registry: default_org
      name:
        from_registry: my_name
    ```

- from_object

  Gets any Python object available in the environment::
    ```yaml
    data:
      url:
        from_object:
          name: robottelo.constants.FAKE_0_YUM_REPO
     ```

- from_search

  Perform a search and return its result::
    ```yaml
    data:
      organization:
        from_search:
          model: Organization
          data:
            name: Default Organization
    ```

- from_read

  Perform a read operation, which is useful when we have unique data or id::
   ```yaml
    data:
      organization:
        from_read:
          model: Organization
          data:
            id: 1
    ```

## The internal registry

Every action which returns a result can write its result to the registry, so
it is available to be accessed by other actions.

Provide a ``register`` unique name in ``action`` definition.

The actions that support ``register`` are:

- create
- update
- register
- assertion

All dynamic directives ``from_*`` supports the use of ``register``

Example::

```yaml
  - action: create
    model: Organization
    register: my_org
    data:
      name: my_org

  - model: User
    log: Creating user under {{ register.my_org.name }}
    data:
      organization:
        from_registry: my_org
```

Documentation yet to be updated
https://github.com/rochacbruno/robottelo/blob/data_populate/tests/foreman/data/test_data.yaml provided as reference

##  Writing a complete validation test case in YAML

> Not the purpose of this module, but possible.

```yaml
  # A TEST CASE USING SPECIAL ACTIONS
  # Create a plain vanilla activation key
  # Check that activation key is created and its "unlimited_hosts" attribute defaults to true
  # in the end of the validation the return code will be 0
  # and validation_errors will have the data to report

  - action: create
    log: Create a plain vanilla activation key
    model: ActivationKey
    register: vanilla_key
    data:
       name: vanilla
       organization:
         from_registry: default_orgs[0]

  # if this assertion fails, return code is 1
  - action: assertion
    log: |
      Check that activation key is created and its "unlimited_hosts"
      attribute defaults to true
    operation: eq
    register: vanilla_key_unlimited_hosts
    data:
      - from_registry: vanilla_key.unlimited_hosts
      - true

  - action: echo
    log: Vanilla Key Unlimited Host is False!!!!
    level: error
    print: true
    when: vanilla_key_unlimited_hosts == False

  - action: echo
    log: Vanilla Key Unlimited Host is True!!!!
    level: info
    print: true
    when: vanilla_key_unlimited_hosts

  - action: register
    data:
      you_must_update_vanilla_key: true
    when: vanilla_key_unlimited_hosts == False
```


rochacbruno:
As this new module does not interfere with other functionalities, once reviewed it can be merged so we can start playing with this to get suggestions, bugs and improvements.

Some new features are already planned to be developed.

- Finish documentation detailing YAML actions
- Export data using hammer csv and transform in to the YAML data
- Write unitest style reports
- implement CLI populator
- break base.py in to more modules

rochacbruno:
A question raised is if we should put this code in a separate repository called `foreman-populate` so it can be useful for users without the need of robottelo and it would be easy to turn this into Ansible compatible module.

docs/features/populate.rst
```diff
@@ -0,0 +1,296 @@
...
+
+- populator
+
+  The name of the populator defined in ``populators``
+- populators
+
+  The specification of populator modules to be loaded
+- verbose
+
+  The verbosity of logging 0, 1 or 2, it can be overwritten with -vvv in commands.
```
elyezer: Here says that verbosity is 0, 1 or 2 but in the example below it is 3. Not sure what is the correct information.

tests/foreman/data/test_data.yaml
```diff
@@ -0,0 +1,256 @@
...
+          data:
+            name: Orgcompany7
+          attr: name
+    with_items:
+      - a
+      - b
+
+# DANGER:
+
+# update and delete actions can break validation
```
elyezer: Can I expect that in future it populate will be able to identify update and delete to properly validate the system?

docs/features/populate.rst
```diff
@@ -0,0 +1,296 @@
...
+
+        actions:
+          - model: Organization
+            register: organization_1
+            data:
+              name: My Org
+
+    Then you can use in decorators::
+
+        @populate_with('file.yaml')
```
elyezer: What if I want to have this on setUp or setUpClass? I think applying to either setUp or setUpClass gives more flexibility and allows one time setup per test or per class.

That said I think the docs can mention about it and recommend decorating those methods instead of trying to decorate a class.

docs/features/populate.rst
```diff
@@ -0,0 +1,296 @@
...
+Dynamic Data
+++++++++++++
+
+There are some ways to fetch dynamic data in action definitions, it depends
+on the action type.
+
+For any key you can use ``Jinja`` to provide a dynamic value as in::
+
+  value: "{{ get_something }}"
+  value: "{{ fauxfactory.gen_string('alpha')}}"
```
elyezer: Add a space before `}}`?

docs/features/populate.rst
```diff
@@ -0,0 +1,296 @@
...
+  value: user_{{ item }}
+
+For some actions you can provide a ``data`` key, that data is used to create
+new entities and also to perform searches or build the action function.
+
+Every ``data`` key accepts 4 special reference directives in its sub-keys.
+
+- from_registry
```
elyezer: I see three behaviors for from_registry, from_object, from_search and from_read. For example from_registry accept the registry variable name directly while from_object expect an dict with a `name` key, also from_search and from_read expects a model and data.

To me from_registry and from_object could accept the name  the same way (no need to add a name for from_object) while from_search and from_read could accept the `model` and `data` attributes (the way it is right now).

docs/features/populate.rst
```diff
@@ -0,0 +1,296 @@
...
+Provide a ``register`` unique name in ``action`` definition.
+
+The actions that support ``register`` are:
+
+- create
+- update
+- register
+- assertion
+
+All dynamic directives ``from_*`` supports the use of ``register``
```
elyezer: This is not clear to me what it means. It means that I can have something like:

```
from_registry: some_registry_var
    register: my_reg
```

and that will copy what is in some_registry_var to my_reg?

It would be great to expand a little bit about this or add an example.

robottelo/populate/main.py
```diff
@@ -0,0 +1,58 @@
...
+def validate(data, **extra_options):
+    """Loads and execute populator in validate mode"""
+    verbose = extra_options.get('verbose')
+    populator = get_populator(data, verbose)
+    populator.execute(mode='validate')
+    populator.logger.info("Validator finished!")
+    return populator
+
+
+def wrap_context(context):
```
elyezer: Why this is needed? It is just returning context why not use that directly?

robottelo/populate/decorators.py
```diff
@@ -0,0 +1,53 @@
...
+        """Wrap test method"""
+
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            """decorator wrapper"""
+
+            result = populate(data, **extra_options)
+            if context:
+                context_name = context if context is not True else 'context'
+                kwargs[context_name] = wrap_context(result)
```
elyezer: This is the same as `kwargs[context_name] = result` since `wrap_context` just return `result`.

docs/features/populate.rst
rochacbruno: sure! you can decorate setUp

```
@populate_with(data, context=True)
def setUp(self, context):
    self.context = context

def test_something(self):
    assert self.context.organization_1.name == "Bla"
```

It can be done and I will add an example to the docs.

(although  it is not recommended, because each test_case should have its isolate scope)

docs/features/populate.rst
rochacbruno: I'll fix the comment and example, thanks. currently there are 3 levels of verbosity, but I am not sure if it is right, open to changes.

0 - log nothing
1 - (-v) log only populate INFO
2 - (-vv) log populate DEBUG and nailgun DEBUG
3 - (-vvv) log everything including ssh (which will be added when CLI support impkemented)

docs/features/populate.rst
rochacbruno: That has a reason. 

`from_registry` accepts pure string and you can write `var_name[0].property.object.function(a=1, b=2)` etc.. you are not limited, you can write Python code in there and will get the result.

`from_object` expects a module name, can be anything `polarion.foo.bar.zaz` and this name is imported dynamically, but if you need specific key or attribute or need to call a function you can pass `key, attr` parameters to get that.

docs/features/populate.rst
rochacbruno: yes you are right, you can nest as many `register` you want as it is recursive.

tests/foreman/data/test_data.yaml
rochacbruno: Yeah, that needs to be done, the main problem is that it is stateless, so if we save an artifact file with the results of last run, so we can use that file to validate. Need more ideas to solve that.
(it also applies to the use of gen_string(), because in the validate run string are generated again and invalidate... maybe a result file has to be created)

robottelo/populate/main.py
rochacbruno: I forgot to write the code for `wrap_context` that function existis for a reason, it should wrap the result in an object accessible via keys and attributes (a dict with getattr implemented)

rochacbruno:
@elyezer I'll implement your comments, thanks for the review and ideas. After that this code will be moved to the new repo called `satellite-populate`

docs/features/populate.rst
elyezer: The current definition is reasonable to me.

docs/features/populate.rst
elyezer: Interesting, since they support different approaches then it makes sense to be the way they are. Thanks for the clarification.

elyezer: Maybe it worth a couple more examples?

docs/features/populate.rst
elyezer: Good

tests/foreman/data/test_data.yaml
elyezer: Having dynamic data will complicate. Maybe marking the operation for the same entity with a common register so the tool can track?

I will try to think on something and will comment when implementing that. No need to do it now.

robottelo/populate/decorators.py
ldjebran: Think we should rename wrap_context to default_context_wrapper   
create a function with same name wrap_context with arg context_wrapper with default value None,
that way we can provide a customized context wrapper in decorator kwarg
```python
def warp_context(result, wrapper=None):
    if wrapper:
        return wrapper(result)
    return default_context_wrapper(result)
```
add a kwarg context_wrapper=None to the decorator so that we can use as:
```python 
def my_customized_wrapper(result):
   context = ...
  return context

@populate_with(data, context_wrapper=my_customized_wrapper)
def test_with_populator(self, context):
    # use context

```

rochacbruno:
This tool has moved to https://github.com/SatelliteQE/satellite-populate

PR #4142, opened 2017-01-05
Skip tests because of BZ 1395229

svtkachenko:
A lot of tests failing because of https://bugzilla.redhat.com/show_bug.cgi?id=1395229
Including CLI tests though BZ desribes API problem. As in most cases hammer sends `[]` as value for `remove` commands.

```
# hammer --debug -v -u admin -p changeme  organization remove-domain --domain="oikseyik8f" --name="YtRJH8"

[ INFO 2017-01-05 11:54:39 API] PUT /katello/api/organizations/61
[DEBUG 2017-01-05 11:54:39 API] Params: {
    "organization" => {
        "domain_ids" => []
    }
}
```

@oshtaier Though you suggested to open more relevant BZs, devs close them as duplicates because `1395229` describes the root cause.

### DIFF
--- a/tests/foreman/api/test_organization.py
+++ b/tests/foreman/api/test_organization.py
@@ -319,6 +319,7 @@ class OrganizationUpdateTestCase(APITestCase):
         self.assertEqual(len(org.hostgroup), 1)
         self.assertEqual(org.hostgroup[0].id, hostgroup.id)
 
+    @skip_if_bug_open('bugzilla', 1395229)
     @tier2
     def test_positive_remove_hostgroup(self):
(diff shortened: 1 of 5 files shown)


rochacbruno:
@svtkachenko what if we just send `None` or `""` (empty str) instead of empty list?

We can create a special constant `EMPTY` and use that

svtkachenko:
@rochacbruno, @rplevka found a workaround: send `[""]` instead of `[]`. 
However,
1. It requires changes in nailgun
2. It is applicable only for API tests as we can't control API requests performed by hammer.
And in this change only 3 tests from 24 are API, most of them CLI.

PR #4181, opened 2017-01-13
Added config for mention_bot to skip some people

rochacbruno:
We are going to test the use of mention bot https://github.com/facebook/mention-bot

This config file specifies people who should never be mentioned or assigned as reviewer.

### DIFF
--- a/.mention-bot
+++ b/.mention-bot
@@ -1,0 +1,22 @@
+{
+    "userBlacklist": [
+        "elyezer",
+        "omaciel",
+        "cswiii",
+        "Ichimonji10",
+        "AdamSaleh",
+        "tkolhar",
+        "alda519",
+        "anarang",
+        "jhutar",
+        "danuzclaudes",
+        "helgie",
+        "lhellebr",
+        "ares",
+        "eanxgeek",
+        "apagac"
+    ],
+    "createReviewRequest": true,
+    "delayed": true,
+    "delayedUntil": "5 minutes"
+}


rochacbruno:
@elyezer I added you as reviewer just because you helped to install this and you know the history of contributors. But look at your name in BlackList so you should not be mentioned as reviewer again :)

abalakh:
To 1) : ok, so it doesn't assign people directly, just mentions them. In such case pinging 3 people by default may be ok.

abalakh:
To 5) hm, curious what do they mean by `already assigned PR's`. If it works with `reviewers` field then it's ok, but if `Assignees` - we should leave it with `false` value as we use this field for different purpose

rochacbruno:
@abalakh I implemented some of your comments

1) I think 3 is a good number, as our minimum is 2 reviewers, but it is best if we have more, so even if we have 3 reviewers assigned, the second to approve is the one to merge.

2) I included 5 minutes delay

3) findPotentialReviewers is already true by default

4) The bot looks only on [contributors history](https://github.com/SatelliteQE/robottelo/graphs/contributors) and there are people that are not member of our org, so I think it is better to have explicit Blacklist.

5) This relates to skip the mention if PR is assigned, not if reviewers are assigned, this is something still confuse in github and I think we should not touch this now.

> NOTE: I activated the  `createReviewRequest` so the bot may mention and also assign reviewers.

```
{
    "userBlacklist": [
        # team members that are not included in the review process
        "elyezer",
        "omaciel",
        "cswiii",
        "Ichimonji10",
        # Other contributors
        "AdamSaleh",
        "tkolhar",
        "alda519",
        "anarang",
        "jhutar",
        "danuzclaudes",
        "helgie",
        "lhellebr",
        "ares",
        "eanxgeek",
        "apagac"
    ],
    "createReviewRequest": true,
    "delayed": true,
    "delayedUntil": "5 minutes"
}
```

rochacbruno:
2 ACKs and this PR does not touch code, so safe to merge :)

Issue #4185, opened 2017-01-16
Unsupported MAC addresses being used by LibvirtGuest


rplevka:
There are occurrences of usage of reserved MAC addresses having a prefix of 'FE:'. We generate them using `fauxfactory` module. Such addresses are valid, but reserved in libvirt, thus we might need to exclude them.
There are 2 options we might approach this:

~~1. implement a `libvirt` flag to `fauxfactory`'s `gen_mac` method~~
2. implement the excluding logic into the libvirtGuest class

```
E           LibvirtGuestError: Failed to run virt-install: ERROR    unsupported configuration: Unable to use MAC address starting with reserved value 0xFE - 'fe:44:d2:91:e6:bb' - 
E           Domain installation does not appear to have been successful.
```

PR #4189, opened 2017-01-17
[DNM] fixes 4185 - excluding fe:* MAC addresses on creating libvirt host

rplevka:
Addresses: https://github.com/SatelliteQE/robottelo/issues/4185

requires cherry-pick

this could be considered a proof of the concept (note that 'not' is being used in the actual code, so the 'fe:*' address is __not__ going to be returned)
```python
In [103]: while True:
   .....:     mac = gen_mac(multicast=False, locally=True)
   .....:     if mac.startswith(u'fe'):
   .....:         break
   .....:     

In [104]: mac
Out[104]: u'fe:05:35:d1:9b:41'

In [105]: while True:
    mac = gen_mac(multicast=False, locally=True)
    if mac.startswith(u'fe'):
        break
   .....:     

In [106]: mac
Out[106]: u'fe:4b:12:fb:7f:dc'

In [107]: while True:
    mac = gen_mac(multicast=False, locally=True)
    if mac.startswith(u'fe'):
        break
   .....:     

In [108]: mac
Out[108]: u'fe:c2:cc:6a:ab:2a'
```

### DIFF
--- a/robottelo/libvirt_discovery.py
+++ b/robottelo/libvirt_discovery.py
@@ -56,7 +56,16 @@ class LibvirtGuest(object):
         else:
             self.image_dir = image_dir
         if mac is None:
-            self.mac = gen_mac(multicast=False, locally=True)
+            # fe:* MAC range is considered reserved in libvirt
+            for _ in range(0, 10):
+                mac = gen_mac(multicast=False, locally=True)
+                if not mac.startswith(u'fe'):
+                    self.mac = mac
+                    break
+                mac = None
+            if not mac:
+                raise ValueError('Unable to generate a valid MAC address')
+
         else:
             self.mac = mac
         if bridge is None:


robottelo/libvirt_discovery.py
```diff
@@ -56,7 +56,13 @@ def __init__(
         else:
             self.image_dir = image_dir
         if mac is None:
-            self.mac = gen_mac(multicast=False, locally=True)
+            # fe:* MAC range is considered reserved in libvirt
+            while True:
+                mac = gen_mac(multicast=False, locally=True)
+                if not mac.startswith(u'fe'):
```
rochacbruno: is that `if not` correct? in your comment seems to be only `if`.

Also, what happens if a mac starting with `fe` is never generated?

rochacbruno: OK, I just read the related issue, we cannot generate macs starting with `fe` so maybe the better is to address that problem in `fauxfactory`?

rochacbruno:
```
fauxfactory.gen_mac(....., exclude_pattern="^fe")
```

Or something like the above... so fauxfactory may filter out the invalid values?

rplevka: @rochacbruno but the purpose of the fauxfactory is to generate a mac addresses. Those with FE: are perfectly valid. I decided to go this way so we keep the fauxfactory lightweight and not overweight its methods by further options. This way we might go with other methods as well.

sghai: Yeah.. agreed what @rochacbruno said. fix should go to fauxfactory `gen_mac`

rochacbruno: @rplevka I understand, but what if for some reason this loops forever, lets say, if gen_mac gives us only `fe` values? very unexpected behavior, but who knows....

rochacbruno: can we use `multicast=True` ?

```
In [3]: while True:
   ...:     mac = gen_mac(multicast=True, locally=True)
   ...:     if mac.startswith('fe'):
   ...:         break
   # never breaks, because faux_factory ensure first octet to be started with 1
```

https://github.com/omaciel/fauxfactory/blob/master/fauxfactory/__init__.py#L630

rplevka:
@rochacbruno 
>I understand, but what if for some reason this loops forever

I got your point, but I just consider the amount of effort needed to rule out the corner case as not worth it.

rochacbruno:
@rplevka I am very afraid of `while True`, have you seem my comment about using `multicast=True` ?

rplevka:
after a discussion on IRC we agreed on proceeding with a limited loop with a potential `ValueError` raised.

robottelo/libvirt_discovery.py
```diff
@@ -56,7 +56,16 @@ def __init__(
...
-            self.mac = gen_mac(multicast=False, locally=True)
+            # fe:* MAC range is considered reserved in libvirt
+            for _ in range(0, 10):
+                mac = gen_mac(multicast=False, locally=True)
+                if not mac.startswith(u'fe'):
+                    self.mac = mac
+                    break
+                mac = None
+            if not mac:
+                raise ValueError('Unable to generate a valid MAC address')
```
sghai: 👍

Issue #3921, opened 2016-10-17
Need a more roboust method to handle custom user role creation


san7ket:
Currently, we use complete permissions dict , and permissions according to resource type.
While, some bugs requires , certain subset of permissions, so rather than re-defining the permissions list for a custom role , there should be some technique to accomplish it. 

As suggested by @rplevka [this](https://github.com/SatelliteQE/robottelo/pull/3882#discussion_r82351675)  could be done to achieve this.

san7ket:
PR was merged closing this

PR #4215, opened 2017-01-23
Fixes issue #3921, added helper to create custom user role for bz automation

san7ket:
Defined a new data structure, to associate bz id with permissions needed to create role.
Created Helper to fetch these associations from the constants file and create appropriate filters and the return the custom role.
Automated a Bug test cases, to incorporate the new structure for creating user roles.
Implemented the suggestion @rplevka https://github.com/SatelliteQE/robottelo/pull/3882#discussion_r82351675

nosetests


robottelo/api/utils.py
```diff
@@ -437,3 +438,34 @@ def configure_provisioning(org=None, loc=None):
...
+     :param bz_id: This is the bugzilla id that is specified in the
+        PERMISSIONS_WITH_BZ list, all the permissions associated with the bz_id
+        will be fetched and filters will be created
+     :return: A single role entity will be created from all the created filters
+     """
+    perms_dict = {}
+    role = entities.Role().create()
+    for res_type in PERMISSIONS_WITH_BZ.keys():
+        for value in PERMISSIONS_WITH_BZ[res_type]:
+            if bz_id in value['bz']:
```
rplevka: if value.get('bz') is not None and bz_id in value['bz']

robottelo/constants.py
```diff
@@ -1224,3 +1117,414 @@
...
     '7/27/16'
 ]
+
+# To create custom role for bugzilla's
+# Append bugzilla id to dict, also see utils.get_perms_by_bz()
+PERMISSIONS_WITH_BZ = {
+    None: [
+        {'name': 'access_dashboard', 'bz': [1306359]},
+        {'name': 'access_settings', 'bz': [1306359]},
+        {'name': 'app_root', 'bz': []},
```
rplevka: implementing https://github.com/SatelliteQE/robottelo/pull/4215/files#r97305883 would make 'bz' key with empty list unnecessary (dict.get('key') would return None if the 'key' does not exist so you wouldn't get an exception)

renzon: @rplevka I prefer not having to deal with None values if possible, just like @san7ket did. More than that, all permissions are explicitly listed. But I agree empty lists all around the code are "ugly". So defaultdict can be used to avoiding definening a bz key where there is no bz id:

```python
>>> from collections import defaultdict
>>> dct=defaultdict(list)
>>> dct['bz']
[]
```

What you guys think?

robottelo/api/utils.py
```diff
@@ -437,3 +438,34 @@ def configure_provisioning(org=None, loc=None):
...
+
+     :param bz_id: This is the bugzilla id that is specified in the
+        PERMISSIONS_WITH_BZ list, all the permissions associated with the bz_id
+        will be fetched and filters will be created
+     :return: A single role entity will be created from all the created filters
+     """
+    perms_dict = {}
+    role = entities.Role().create()
+    for res_type in PERMISSIONS_WITH_BZ.keys():
+        for value in PERMISSIONS_WITH_BZ[res_type]:
```
renzon: Just a nitpick that you @san7ket can decide to change or not:

For this kind of Cartesian product I prefer using the function to avoid nested loops whyle typing less boilerplate:

```python
>>> from itertools import product
>>> product([1,2], 'AB')
<itertools.product object at 0x7f7ad96bbeb0>
>>> list(_)
[(1, 'A'), (1, 'B'), (2, 'A'), (2, 'B')]
>>> for n, c in product([1,2], 'AB'): print(n, c)
... 
(1, 'A')
(1, 'B')
(2, 'A')
(2, 'B')
```
This function accepts accepts any iterable ;)

robottelo/api/utils.py
rochacbruno: For that case I prefer to have it in dict set as None

```python
{'name': 'app_root', 'bz': None}
```

and then in code to keep the iteration syntax clean I always do:

```python
if bz_id in value.get('bz', []):
   # do something
```

So if 'bz' is None it always get an empty list without he need of external lib, and it is short circuit, no overhead at all as empty list will not be iterated.

@rplevka @renzon

renzon: We discussed on irc and @plevka and I agreed on @rochacbruno 's solution

san7ket: will implement this , by @rochacbruno suggestion.

robottelo/api/utils.py
san7ket: @renzon trying to do this, facing issues like , here [1,2] can be fetched from PERMISSIONS_WITH_BZ.keys() , but the "AB" , are inside the PERMISSIONS_WITH_BZ.keys() , so we need to loop to get to the values inside keys. The AB are dynamic and are not known, until we loop through each key.

robottelo/api/utils.py
san7ket: I have to iterate through the dictionary, when I do this, I get , 
 if bz_id in value.get('bz', []):
TypeError: argument of type 'NoneType' is not iterable
 
Did i miss something ??

rochacbruno: @san7ket you are right ans that was my fault:

if you use a `{'name': 'app_root', 'bz': None}` the code must be:

```python
if bz_id in value.get('bz') or []:
   # ...
```

only If you use `{'name': 'app_root'}` omitting the 'bz' key, then you can use the `.get('bz', [])`

san7ket: Ommiting the BZ key where ever it's not set , for cleaner code.

san7ket:
Changes done , please review

robottelo/api/utils.py
rochacbruno: If the current solution is working fine, so we can stick with the current implementation for now.

robottelo/api/utils.py
```diff
@@ -437,3 +438,34 @@ def configure_provisioning(org=None, loc=None):
...
+                    perms_dict[res_type].append(value['name'])
+                else:
+                    perms_dict[res_type] = [value['name']]
+    for resource_type in perms_dict.keys():
+        for value in perms_dict[resource_type]:
+            permissions = entities.Permission(
+                name=value,
+                resource_type=resource_type,
+            ).search()
+        entities.Filter(permission=permissions, role=role).create()
```
abalakh: There's something wrong in logic here. For each permission in a list you're overriding the same `permissions` variable, which means only the last permission will be saved and then passed to a filter. Did you mean to form a list of permissions instead?

abalakh: Also, how does your search work? `.search()` returns a list of results, so to take exact entity you need to use it like `.search()[0]` as otherwise it will appear in extra list (or you may even receive multiple entries with similar names if any)

robottelo/api/utils.py
```diff
@@ -437,3 +438,34 @@ def configure_provisioning(org=None, loc=None):
...
+    role = entities.Role().create()
+    for res_type in PERMISSIONS_WITH_BZ.keys():
+        for value in PERMISSIONS_WITH_BZ[res_type]:
+            if bz_id in value.get('bz', []):
+                if res_type in perms_dict:
+                    perms_dict[res_type].append(value['name'])
+                else:
+                    perms_dict[res_type] = [value['name']]
+    for resource_type in perms_dict.keys():
+        for value in perms_dict[resource_type]:
```
abalakh: I don't think there's a need in nested loops, and especially iterating through nested loops once to form a list of items to actually work with in another 2 nested loops, that's too overkill. You can combine these blocks into one (without forming temporary dict). And regarding nested loops - i'd rather use `.items()`, `filter` and list comprehension to avoid them.
Take a look on this implementation instead:
```python
def get_role_by_bz(bz_id):
    role = entities.Role().create()
    for res_type, perms in PERMISSIONS_WITH_BZ.items():
        perms_with_bz = filter(lambda x: x.get('bz'), perms)
        if perms_with_bz:
            permissions = [
                entities.Permission(
                    name=perm['name'],
                    resource_type=res_type,
                ).search()[0]
                for perm in perms
            ]
            entities.Filter(permission=permissions, role=role).create()
    return role.read()
```
This snippet above already uses my assumptions in comments below.
@rochacbruno your thoughts?

robottelo/api/utils.py
```diff
@@ -437,3 +438,34 @@ def configure_provisioning(org=None, loc=None):
...
+            permissions = entities.Permission(
+                name=value,
+                resource_type=resource_type,
+            ).search()
+        entities.Filter(permission=permissions, role=role).create()
+    return role
```
abalakh: Variable `role` doesn't reflect role's actual state, as after assigning you made some changes (created filters) which won't be reflected in outdated variable of course. Please use `return role.read()` instead to return actual role.

robottelo/api/utils.py
rochacbruno: @abalakh I like your implementation, a little note is that 

```python
perms_with_bz = filter(lambda x: x.get('bz'), perms)
``` 
In Python 3 (which we may move soon) ``filter`` now returns an iterator, and doing ``if perms_with_bz`` will make the iterator to be evaluated as list anyway, so there is no advantage of using.

so for clarity and for when we move to Python 3, this can be replaced with: 

```python
perms_with_bz = [x for x in perms if x.get('bz') is not None]
```

rochacbruno: The general recommendation for Python 3 is to use comprehension and expressions over ``map`` and ``filter``

san7ket:
Addressed @abalakh comments.

san7ket:
Attaching test run, 

pytest tests/foreman/ui/test_contentview.py -k test_positive_promote_CV_with_custom_user -vv
tests/foreman/ui/test_contentview.py::ContentViewTestCase::test_positive_promote_CV_with_custom_user_role_and_filters <- robottelo/decorators/__init__.py PASSED

robottelo/api/utils.py
```diff
@@ -437,3 +438,30 @@ def configure_provisioning(org=None, loc=None):
...
+
+     :param bz_id: This is the bugzilla id that is specified in the
+        PERMISSIONS_WITH_BZ list, all the permissions associated with the bz_id
+        will be fetched and filters will be created
+     :return: A single role entity will be created from all the created filters
+     """
+    role = entities.Role().create()
+    for res_type, perms in PERMISSIONS_WITH_BZ.items():
+        perms_with_bz = [x for x in perms if x.get('bz') is
+                         not None and bz_id in x.get('bz')]
```
abalakh: nitpick: how about `[x for x in perms if bz_id in x.get('bz', [])]`?

robottelo/api/utils.py
```diff
@@ -437,3 +438,30 @@ def configure_provisioning(org=None, loc=None):
...
+        if perms_with_bz:
+            permissions = [
+                entities.Permission(
+                    name=perm['name'],
+                ).search()[0]
```
abalakh: Your original search command included resource type (`resource_type=resource_type,`), what happened to it? If Permission search does not support specifying resource type on entity initialization, you can still pass it in raw query like that:
```python
entities.Permission(name=perm['name']).search(
    query={'search': 'resource_type = {}'.format(res_type)}
)[0]
```
That's the idea in case resource type is useful. If it's redundant - couple of nitpicks:
1) As `res_type` is not used anywhere, drop it and use `dict.values()` to iterate through dict:
```python
for perms in PERMISSIONS_WITH_BZ.values():
```
2) Your search command seems to fit in one line:
```python
                entities.Permission(name=perm['name']).search()[0]
```

robottelo/api/utils.py
```diff
@@ -13,6 +13,7 @@
     DEFAULT_TEMPLATE,
     RHEL_6_MAJOR_VERSION,
     RHEL_7_MAJOR_VERSION,
+    PERMISSIONS_WITH_BZ,
```
abalakh: nitpick: please sort it in alphabetical order

robottelo/constants.py
abalakh: `dict.get()` also accepts optional argument with default value, so you can use it like `dict.get('key', [])` and you won't be hit by None, don't have to store empty dicts and no need to import any custom dict implementation.

robottelo/api/utils.py
abalakh: Opened old comments threads and can see this suggestion was already there but it wasn't working as the design of your dict was different at that moment. Now it should :)

san7ket: @abalakh yes, now it's working

robottelo/api/utils.py
san7ket: The problem when I used the res_type to search was , it returned me with a list the all filters available for that res_type, rather than just one permission, on eliminating the res_type we get expected result.

abalakh: Please try option with query explicitly specified, i think that should be the issue. But in case it's not - i'm ok to drop it.

san7ket:
Addressed the comments @abalakh

PR #4224, opened 2017-01-23
[6.2.z] Fix pip install error, as pip itself is requiring packaging>=16.8

rochacbruno:
Travis is failing with Version Conflict

- Pinning versions on setup.py to avoid version conflict
- Unpinning `packaging`

Reason:

We do not need to freeze that specific package.
we should do an overview of frozen requirements because
not need to freeze all of then, only the ones we directly import in the code
there are python tools to generate the freeze file for us.


rochacbruno:
This should be cherry picked to 6.1.z

setup.py
```diff
@@ -16,22 +16,22 @@
...
     include_package_data=True,
     install_requires=[
-        'cachetools',
-        'cryptography',
-        'fauxfactory',
-        'inflector',
+        'cachetools==1.1.6',
+        'cryptography==1.4',
+        'fauxfactory==2.0.9',
+        'inflection==0.3.1',
```
ldjebran: in robottelo we are using inflector, was not able inflection usage (inflection is a requirement   and used by nailgun)

rochacbruno: Good catch! fixed it!

Issue #3769, opened 2016-08-30
Freeze dependencies


rochacbruno:
We have discussed that in the past in #3423 

But I am raising this question again because today we faced the same issue again in automation.

External dependency upgraded and broke our automation.

So I am raising the idea of pinning all deps versions and then use https://pyup.io to notify us of upgraded deps (it includes changelog and we can test the dep before merging)

renzon:
+1 for dependencies pinning. Manually upgrading dependencies allow us finding regression because of lib upgrade. More then that, it keep us from to figure it out on builds that de dependency was the problem and not our code.

elyezer:
I continue thinking that master should always use the latest. But we can pin whenever we come to the point we have a release to deliver, that will help stabilize the dependencies and all failures be automation failures and not framework failures.

The main concern is to make sure we adapt our code from time to time to get all updates, if pyup.io can help on not forgetting to do that then we can use it and pin versions.

rochacbruno:
@elyezer I'll work on this, maybe your help will be needed to setup pyup hooks.

elyezer:
@rochacbruno sure, please let me know what I need to do.

omaciel:
Speaking of pinning, can we update the version of `selenium`? Running UI tests locally fail with newest `firefox` so I always bump it... we would have to check our docker images too I guess, to make sure that our auto can continue to work with it...

rochacbruno:
@omaciel I remember we had one issue with scrolling in the new selenium, I need to check it

rochacbruno:
implemented

PR #4234, opened 2017-01-24
Freezed only imported requirements using pipreqs

rochacbruno:
## Why not `pip-freeze`

- pip freeze only saves the packages that are installed with pip install in your environment.
- pip freeze saves all packages in the environment including those that you don't use in your current project. (if you don't have virtualenv)

## pipreqs

- Used `pipreqs` to freeze only the packages we really import direct in our code
- https://github.com/bndr/pipreqs

## Removed reqs from setup.py

- Robottelo is not installed via PyPI so no need to repeat requirements on setup.py, every installation always runs `pip install -r requirements.txt | -freeze.txt

## AFTER MERGING
- Proposing the pinning of versions in master (to avoid unexpected broken builds)
- related to #3769
- So we need to setup `pyup` (only for master branch)
- [x] Someone with admin access in this repo should setup https://pyup.io/ to check `requirements.txt` file


omaciel:
Added `pyup` service with the configuration displayed below:

screenshot 2017-01-24 12 01 20

rochacbruno:
I am going to include `.pyup` config file in this PR...

PR #4237, opened 2017-01-24
Fix travis mock error and added .pyup.yml config specifing reqs file as requirements.txt

rochacbruno:
- pin: False = If we add a requirement without version, bot should not automatically pin it.
- requirements: Specify the bot to check only requirements.txt and never other req files.

plus: mock is already on requirements, new setuptools complains if installed again.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,5 +4,4 @@ envlist = py27,py34
 deps=
     -rrequirements.txt
     pytest-cov
-    py27: mock
 commands=py.test --cov --cov-config=.coveragerc tests/robottelo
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -4,5 +4,4 @@ envlist = py27,py34
 deps=
     -rrequirements.txt
     pytest-cov
-    py27: mock
 commands=py.test --cov --cov-config=.coveragerc tests/robottelo
```
rochacbruno: MOck is already on requirements.txt, Travis updated setuptools and now the new version complains if we try to install already listed package.

PR #4238, opened 2017-01-24
[6.2.z] Requirements-freeze misses pytest-cov

rochacbruno:
Nit pick I found for travis-ci, pytest-cov and other testing libs is not referred in code so was not added in requirements-freeze

Currently it is inconsistent, z-branches uses only reqs file while master install explicit in travis.yaml


ldjebran:
ACK

rochacbruno:
Need to merge this to fix requirements in CI

PR #4250, opened 2017-01-25
[6.1.z] Renamed requirements-freeze to requirements

rochacbruno:
In Z branches there were no requirements.txt and now even master
has its versions pinned, so it makes no sense to have different filename.
lets use the same pattern.


kbidarkar:
ACK

PR #4319, opened 2017-02-09
Pin testimony version

elyezer:
I have added two **separate** commits (please do not squash) to accomplish the following:

1) Make sure robottelo is being tested on Python 3.5 and 3.6 (this will avoid issues like #4318)
2) Pin testimony version to avoid breaking robottelo when the new version gets released.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
     - "3.5"
     - "3.6"
 install:
-    - pip install -r requirements.txt coveralls flake8 sphinx testimony tox-travis manage
+    - pip install -r requirements.txt coveralls flake8 sphinx tox-travis manage
 script:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,10 +1,12 @@
...
-    - TOXENV=py27
-    - TOXENV=py34
+python:
+    - "2.7"
+    - "3.4"
+    - "3.5"
+    - "3.6"
 install:
-    - pip install -r requirements.txt coveralls flake8 sphinx testimony tox manage
+    - pip install -r requirements.txt coveralls flake8 sphinx tox-travis manage
```
elyezer: Check first commit's message for more information on why tox-travis is being used.

PR #4377, opened 2017-02-23
Move DISTRO_RHEL6 and DISTRO_RHEL7 to robottelo.properties file.

kbidarkar:
a) This PR should be merged against 6.1 and 6.2 branches as well. For now this is for the master branch.

b) Also this PR requires another patch at the robottelo-ci end to dynamically update the `image_el6` and `image_el7` values under the [distro] section in the robottelo.properties file.

Will raise one PR soon to address the b) part as mentioned above.

### DIFF
--- a/robottelo/constants.py
+++ b/robottelo/constants.py
@@ -24,9 +24,6 @@ BZ_CLOSED_STATUSES = [
     'CLOSED'
 ]
 
-DISTRO_RHEL6 = "rhel68"
-DISTRO_RHEL7 = "rhel73"
-
 FOREMAN_PROVIDERS = {
     'libvirt': 'Libvirt',
     'rhev': 'RHEV',
--- a/tests/foreman/api/test_errata.py
+++ b/tests/foreman/api/test_errata.py
@@ -22,11 +22,10 @@ from robottelo.cli.factory import (
     setup_org_for_a_custom_repo,
     setup_org_for_a_rh_repo,
 )
+from robottelo.config import settings
 from robottelo.constants import (
     DEFAULT_ARCHITECTURE,
     DEFAULT_RELEASE_VERSION,
-    DISTRO_RHEL6,
-    DISTRO_RHEL7,
     FAKE_1_CUSTOM_PACKAGE,
     FAKE_2_CUSTOM_PACKAGE,
     FAKE_2_ERRATA_ID,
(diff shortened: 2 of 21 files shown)


kbidarkar:
Oops! looks like I missed to check for travis. Will fix it soon and push the changes.

robottelo/constants.py
```diff
@@ -1,6 +1,7 @@
 # -*- encoding: utf-8 -*-
 """Defines various constants"""
 from nailgun import entities
+from robottelo.config import settings
```
abalakh: This line is most likely redundant

tests/foreman/api/test_errata.py
```diff
@@ -55,6 +54,8 @@
 
 CUSTOM_REPO_URL = FAKE_6_YUM_REPO
 CUSTOM_REPO_ERRATA_ID = FAKE_2_ERRATA_ID
+DISTRO_RHEL6 = settings.distro.image_el6
```
abalakh: Are you sure settings are already initialized at this point? Please, provide test results. Some time ago this change would definitely not work, but maybe things have changed with introducing robozilla and settings are initialized earlier, i'm not sure.

kbidarkar: yeah, sure will provide test results, I already did test for feasibility sake and it worked. But since then did many changes to the code, so feel, I should test it once again.

robottelo/constants.py
kbidarkar: Yup, agree, I just missed that.

tests/foreman/api/test_errata.py
abalakh: Ok, another idea - if settings are already initialized on import level, please take a look whether they're already initialized while importing from `constants.py`. If they're initialized - instead of dropping `DISTRO_RHEL6` and `DISTRO_RHEL7` from constants.py you could set the value from settings file inside `contants.py`. This way you won't need any changes in test modules, distros will be loaded like before.

kbidarkar: I tried to add `from robottelo.config import settings` to constants.py as the first option and it didn't work , I got the DISTRO_RHEL6 and DISTRO_RHEL7 values as `None`.

kbidarkar: Looks like using `DISTRO_RHEL6 = settings.distro.image_el6` as a global variable also does not work, looks like it worked earlier as I missed to remove the value from the constants.py file, but there was no reference to it in the tests.

lpramuk: Just use it as we do with other settings

lpramuk: But you have to have it in setUp

lpramuk: settings.configure() does the trick you need,

lpramuk: robottelo/test.py:130 

TestCase.setUpClass calls settings.configure()  you have to put it in inherited setUpClass
You need class instance for settings to work.

kbidarkar:
guys thanks for the suggestions, will come up with some different solution soon.
Will close this for now.

Issue #4267, opened 2017-01-27
Error handling for negative tests via API (and partially hammer)


svtkachenko:
Based on #3611. 
After discussion it was decided to filter DB errors for hammer. However, there are still 2 problems that leads to false positives:
1. Errors that are not related to negative case not necessary are only DB errors, see example below.
2. API tests check for general `HTTPError` and treats `500` in the same way as `404/422`:
```python
with self.assertRaises(HTTPError):
```

## Examples:
Some time ago we had a problem with `hostgroup` manipulation and `ActiveModel::ForbiddenAttributesError`.

1. `tests.foreman.cli.test_hostgroup/HostGroupTestCase/test_negative_update_name` **passed** with next output:
```
2016-09-23 15:18:39 - robottelo.ssh - DEBUG - >>> LANG=en_US.UTF-8  hammer -v -u admin -p changeme --output=csv hostgroup update --id="34" --new-name="057075282907452832817381666304521049900009272674418549065965163645300983200699571010310829946122665194619921026496739921856071754398297521987525803790039870936758428833398245796914841035399103550707426344808970618629396302172892934291002960149361298917420136522439515492351277430914861338746640671197"
2016-09-23 15:18:41 - robottelo.ssh - DEBUG - <<< stderr
[ERROR 2016-09-23 15:18:41 API] 500 Internal Server Error
[ERROR 2016-09-23 15:18:41 Exception] ActiveModel::ForbiddenAttributesError
Could not update the hostgroup:
  ActiveModel::ForbiddenAttributesError
[ERROR 2016-09-23 15:18:41 Exception] 

RestClient::InternalServerError (500 Internal Server Error):
    /opt/theforeman/tfm/root/usr/share/gems/gems/rest-client-1.8.0/lib/restclient/abstract_response.rb:74:in `return!'
   ... skipped ...
    /usr/bin/hammer:23:in `load'
    /usr/bin/hammer:23:in `<main>'
```

2. While expected result for `tests.foreman.api.test_hostgroup/HostGroupTestCase/test_negative_update_name` was:
```
2017-01-27 20:42:30 - nailgun.client - WARNING - Received HTTP 422 response: {
[...]

PR #4394, opened 2017-03-01
Implemented custom assertNotRaises and assertNotRaisesRegex assertions

abalakh:
### Background

Some of our tests, especially those ones which target negative cases, are validating whether command will succeed and don't raise any exceptions.
As there are no built-in or unittest assertions like `assertNotRaises`, in most cases we were validating such commands by simply executing them and hoping they will pass, and if they won't - failure will tell us about regression. The thing is, `unittest` marks all the tests, which were not failed inside assertions, as 'error', not 'failure', which is confusing causing everyone to think we have an error in our automation code, when the code itself is completely ok.

To workaround this, sometimes we were using something like this:
```python
try:
    do_thing()
except ExpectedException:
    self.fail()
```
which doesn't look great but works for us.

Then #4267 was submitted, pointing to the fact that expecting exception is not enough for api/cli tests. For example when we're expecting 404 error telling us the resource is not found, in case we just catch HTTPError we can hide different failures like server's 500 error.
To workaround this issue, a bit more tricky workaround was introduced:
```python
try:
    entity.read()
except HTTPError as error:
    if error.response.status_code == 404:
        self.fail()
    else:
        raise
```
This workaround is definitely too complicated to use it every time and some user-friendly helper had to be introduced instead.

### What's this PR about

This pull request introduces following 2 helpers:

* assertNotRaises()
* assertNotRaisesRegex()

the purpose of which is to execute some code and in case expected Exception occurred - mark the test as 'failure', not 'error' with no tricky workarounds.

### Usage

The usage is pretty much similar to `assertRaises` assertion.
Let's assume we want to ensure some host exists in satellite. Before, in API test, we had to do something like this:
```python
try:
    host.read()
except HTTPError:
    self.fail()
```

Now we can simply pass the instruction to `self.assertNotRaises()` method:
```python
self.assertNotRaises(HTTPError, host.read)
```

OR use `assertNotRaises` as a context manager:
```python
with self.assertNotRaises(HTTPError):
    host.read()
```
which allows us to execute a block of code and looks better in general.

Both forms are equal and in case no exceptions were risen - the test will proceed, if expected exception was risen - `AssertionError` will be risen and the test will be marked as 'FAIL', if any other exception occurred - exception won't be caught which will lead to test being marked as 'ERROR'.

But what if just checking Exception presence is not enough and we want to specify the exact HTTP status code (or return code for CLI) which we are expecting to happen? For example, in previous example, we're expecting that in case host is not present - 404 error will be shown, but definitely not 500 error or something else.
For such case, `assertRaises` accepts 2 optional arguments - `http_status_code` and `cli_return_code` for API and CLI accordingly.

So instead of
```python
try:
    host.read()
except HTTPError as error:
    if error.response.status_code == 404:
        self.fail()
    else:
        raise
```

We can now use
```python
self.assertNotRaises(HTTPError, host.read, http_status_code=404)
```
for callable object or we can use assertion as a context manager:
```python
with self.assertNotRaises(HTTPError, http_status_code=404):
    host.read()
```

For CLI, syntax is similar:
```python
with self.assertNotRaises(CLIReturnCodeError, cli_return_code=128):
    Host.info({'id': 1})
```

This way, only expected exception will be caught and only in case response code matches the expected one.

And what if that's still not enough and we want to check the exception message to ensure some pattern is there? The answer is `assertNotRaisesRegex` assertion, which is basically extended `assertNotRaises` assertion which should be used similar to `assertRaisesRegex`.

Example usage with exception and pattern:
```python
# with callable:
self.assertNotRaisesRegex(HTTPError, '404 Client Error', host.read)

# as a context manager:
with self.assertNotRaisesRegex(HTTPError, '404 Client Error'):
    host.read()
```
Example usage with exception, response code and pattern:
```python
# with callable:
self.assertNotRaisesRegex(
    HTTPError, '404 Client Error', host.read, http_status_code=404)

# as a context manager:
with self.assertNotRaisesRegex(
        HTTPError, '404 Client Error', http_status_code=404):
    host.read()
```

And, of course, valid regular expression patterns are supported.

Similar to previous examples, if the exception is risen - only in case exception, pattern and response code (if specified) match the expected ones - test will be marked with 'FAIL', all others exceptions won't be handled and will mark the test with 'ERROR' as usual.

### DIFF
--- a/robottelo/test.py
+++ b/robottelo/test.py
@@ -8,6 +8,7 @@ import csv
 import logging
 import os
 import pytest
+import re
 import unittest2
 
 try:
@@ -101,6 +102,99 @@ from robottelo.ui.user import User
 LOGGER = logging.getLogger(__name__)
 
 
+class _AssertNotRaisesContext(object):
+    """A context manager used to implement :meth:`TestCase.assertNotRaises`.
+    """
+
+    def __init__(self, expected, test_case, expected_regex=None,
+                 cli_return_code=None, http_status_code=None):
+        self.expected = expected
+        self.expected_regex = expected_regex
+        if cli_return_code is not None:
+            self.expected_response_code = cli_return_code
+            self.interface = 'cli'
+        elif http_status_code is not None:
+            self.expected_response_code = http_status_code
+            self.interface = 'api'
+        else:
+            self.expected_response_code = None
+            self.interface = None
+        self.failureException = test_case.failureException
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, tb):
+        if exc_type is None:
+            return True
+
+        try:
+            exc_name = self.expected.__name__
+        except AttributeError:
+            exc_name = str(self.expected)
+
+        self.exception = exc_value  # store for later retrieval
+
+        if issubclass(exc_type, self.expected):
+            if not any((self.expected_response_code, self.expected_regex)):
+                raise self.failureException(
+                    "{0} raised".format(exc_name))
+            regex = self.expected_regex
+            response_code = self.expected_response_code
+            if response_code:
+                if self.interface == 'api':
+                    response_code = (
+                        True if response_code == exc_value.response.status_code
+                        else False
+                    )
+                else:
+                    response_code = (
+                        True if response_code == exc_value.return_code
+                        else False
+                    )
+            if regex:
+                regex = True if regex.search(str(exc_value)) else False
+
+            if response_code and regex:
+                raise self.failureException(
+                    "{0} raised with {1} {2} and {3} found in {4}"
+                    .format(
+                        exc_name,
+                        ('status_code'
+                         if self.interface == 'api'
+                         else 'return_code'),
+                        self.expected_response_code,
+                        self.expected_regex.pattern,
+                        str(exc_value),
+                    )
+                )
+            elif response_code and regex is None:
+                raise self.failureException(
+                    "{0} raised with {1} {2}".format(
+                        exc_name,
+                        ('status_code'
+                         if self.interface == 'api'
+                         else 'return_code'),
+                        self.expected_response_code,
+                    )
+                )
+            elif regex and response_code is None:
+                raise self.failureException(
+                    "{0} raised and {1} found in {2}".format(
+                        exc_name,
+                        self.expected_regex.pattern,
+                        str(exc_value),
+                    )
+                )
+            else:
+                # pass through
+                return False
+
+        else:
+            # pass through
+            return False
+
+
 class TestCase(unittest2.TestCase):
     """Robottelo test case"""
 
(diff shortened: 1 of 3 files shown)


abalakh:
As a bonus updated 2 API host interface tests with new assertions, just to show sample usage. Will provide test results soon.

robottelo/test.py
```diff
@@ -101,6 +102,99 @@
 LOGGER = logging.getLogger(__name__)
 
 
+class _AssertNotRaisesContext(object):
+    """A context manager used to implement :meth:`TestCase.assertNotRaises`.
+    """
+
+    def __init__(self, expected, test_case, expected_regex=None,
+                 cli_return_code=None, http_status_code=None):
```
ldjebran: @abalakht, as understood  we have a general usage for  expected_exception, but for callable value  we stick to only cli_status_code, http_status.

waht do you think if we change this to a value handler  for example 
```python
class APIRaisesHandler:
     __init__(self, value):
     self.value = value

    def response_code(self, exc_value):
         return (
                      True self.value == exc_value.response.status_code
                        else False
                    )

class CLIRaisesHandler:
     __init__(self, value):
     self.value = value

    def response_code(self, exc_value):
         return (
                      True self.value == exc_value.return_code
                        else False
                    )
```

```python 
 with self.assertNotRaises(HTTPError, value_handler=APIRaisesHandler(404)):
             host.read()
```
and this should lead to a more general usage, perhaps we should keep   cli_return_code=None, http_status_code=None, but intead of ceating an interface=api or = cli , add the value_handler in args and if it's not available look at the special cli_return_code and http_status_code and recreate the value handler.
and at __exit__ use the response_code of the handler 
what do you think?

abalakh: @ldjebran frankly, i'd rather want to not decide by my own, but to hear others thoughts, as this isn't something i'll use exclusively, but a tool for everyone so the option which fits all of us should win instead of one's personal preference :)
But ok, my imo - from design perspective - no doubt, your suggestion is better, from usage perspective - it's far more complicated, user should remember to pass some `value_handler` arg, remember to import one of correct value helpers and pass it as a value. I'd prefer to keep interface as simple as possible, as this assertion itself isn't the simplest one, let's not overcomplicate its usage.

rochacbruno: @ldjebran @abalakh usability is penalized but turns this in to more flexible solution, so I give +1 to @ldjebran suggestion.

If `value_handlers` are not passed we can have a `DefaultValueHandler` which automatically is `status_code` or `return_code` by inspecting if TestCase is a subclass of API or CLI, or maybe just raises an `AttributeError("value_handler is required")`

ldjebran: @rochacbruno did you mean this ?

```python
class TestCase(unittest2.TestCase)

    _default_value_handlers_classes = []
    def self.assertNotRaise(self, ...):
        #use self._default_value_handlers_classes if no values_handlers supplied


class APITestCase(TestCase):
    """Test case for API tests."""
    _multiprocess_can_split_ = True
    _default_value_handlers_classes =  [APIRaisesValueHandler] 

class CLITestCase(TestCase):
    """Test case for API tests."""
    _multiprocess_can_split_ = True
    _default_value_handlers_classes =  [CLIRaisesValueHandler] 
```






```

Comment (rochacbruno) [inline robottelo/test.py]: @ldjebran yes! good implementation!

**Comment by abalakh** [comment]
@ldjebran @rochacbruno updated PR with your suggestion with few changed to keep usage as simple as possible. Please take another look.

**Comment by ldjebran** [comment]
ACK, excellent

File: robottelo/test.py [hunk 2]
```diff
@@ -101,9 +102,148 @@
...
+                )
+            else:
+                # pass through
+                return False
+
+        else:
+            # pass through
+            return False
+
+
```
rochacbruno: What if we put all the new code above in a separated module `value_handlers.py` and then import here?

robottelo/test.py
```diff
@@ -169,14 +309,78 @@ def log_method_name(self, prefix=None):
             self._testMethodName,
         ))
 
+    def assertNotRaises(self, expected_exception, callableObj=None,
```
rochacbruno: `/s/callableObj/callable_obj` ?

robottelo/test.py
```diff
@@ -169,14 +309,78 @@ def log_method_name(self, prefix=None):
...
+            expected_value=expected_value,
+            value_handler=value_handler,
+        )
+        if callableObj is None:
+            return context
+        with context:
+            callableObj(*args, **kwargs)
+
+    def assertNotRaisesRegex(self, expected_exception, expected_regex,
+                             callableObj=None, expected_value=None,
```
rochacbruno: `s/callableObj/callable_obj` ?

robottelo/test.py
abalakh: `_AssertNotRaisesContext` is more related to `assertNotRaises` rather to value helpers, and without `_AssertNotRaisesContext` - it's just two small classes (which are closely connected to APITestCase and CLITestCase here btw). I'm not sure if they worth separate module (it's just ~50 lines) but i can move them if you think it's worth it.

robottelo/test.py
abalakh: I was following unittest style - for assertRaises it's called `callableObj`, as both assertions are pretty similar i wanted to keep interface as similar as possible. Any advantages in following different style?

Issue #4088, opened 2016-12-13
Destructive tests


ldjebran:
Destructive tests are tests, that cannot run in parallel with other tests as to achieve their test functionality they have to stop, start, restart the server application, some of them have to destroy and restore the existing data or to achieve an upgrade from version x to version y.

Even if the test will run sequentially and if something goes wrong (in case of test failing or for example restoring the data or bringing the service back online) all the other remaining tests have no chance to even start running.

Classification of destructive tests:     
1. tests that need a service restart
    - UI/CLI/API test_positive_restart_promote_via_dynflow
    - UI/CLI/API test_positive_restart_dynflow_publish
2. tests that need a service restart + data destroy + data recovery
    - HotBackup test pipeline #3905 
3. tests that need a service restart + service version modification + data modifications

We needed to find a well designed running scenario solution to these tests categories.
Waiting you comments,
Please do not hesitate to add the tests or category you found it missing.

rochacbruno:
I asked the same question on #3905 about how to deal with backup tests as it demands stopping the system to perform the restore.

ldjebran:
@rochacbruno yes I know but we have a collection of such tests raising and we have to find a general solution for them all, it seems to me running at the end of the pipe line is very temporary solution.

ldjebran:
@rochacbruno or we should have two or three sat instances running to be used in case of instance failure

sghai:
I think these tests can be run once all tiers are finished. There are manly tests that we may need to automate around installer options. All of them can be covered this way.

We can run them after long running tests or after rhai tests and call them as `system_tests`.

ldjebran:
@sghai, that way in case of services that restart the system, we can try to test if the service is alive and start it if down, and let backup restore to the end of the queue
how we should mark the tests in your case? - this the main question
and how about the upgrade tests? - for the moment I have no task for this

sghai:
@ldjebran : I think we can anytime check the katello-service status or any individual service status before/after  running the test. On other thought, this could be part of teardown to revert the service back once test is done, If I understood your question correctly.

Also, about upgrade tests, upgrade is completely being handled via [automation-tools](https://github.com/SatelliteQE/automation-tools/blob/master/automation_tools/satellite6/upgrade/satellite.py)

ntkathole:
Please feel free to re-open, if this still needs fixing.

Issue #3905, opened 2016-10-07
Question: HotBackup test pipeline


rochacbruno:
For Hot Backup feature we have test cases like this:

```
        @Setup:
        Populate the system with hosts/repos/contents
        @Steps:
        1. Run full backup (base)
        2. Make config change c1
        3. Run incremental backup ib1
        4. Restore base backup, verify c1 config doesnt not exist
        5. restore ib1, veirfy c1 config does exist
        6. Make config change c2
        7. run katello-backup <path_to_ib1> --incremental
        @Assert: Backup "ib1" is backed up.
```

Online backup can run along with the system (no need to shutdown or restart the system) but it is time consuming, also we can have race conditions if backup occurs along with other automation tests.

So I think we need a separated pipeline for sys/backup tests, reasons:
1. we do not want it to take longtime so we are not going to backup a heavily full system
2. we want to assert incremental backups so ensure parallel tests will not compromise the assertions
3. if something went wrong we want to have a clean debug and log

> NOTE: Backup tests are all `run_in_one_thread`  
## **Solutions**:

**Put at the end of existing pipeline.**
1. install pytest-order plugin
2. make backup tests to be the latest to run after some tier test ends
3. record separate artifacts on backup class teardown (system logs)

Pros: we have a system populated by previous tests. 
Cons: Backup errors will fail that tier, we can have more entities than needed taking a longtime to backup.

---

**Create a new pipeline only for sys/backup tests**
1. It can be triggered after tier4, we can have a `pytest.mark.sys`
2. Starts with clean system
3. Populate the system only with sufficient number of important entities
4. Artifacts will be separated anyway
5. It is a ONE THREAD pipeline

Pros: all the above
Cons: One more pipeline to maintain

---

Ideas?

rochacbruno:
@kbidarkar @sghai @omaciel I am writing the tests but we need to discuss the pipeline in the next mtg, any ideas?

ldjebran:
As a general idea we can create a service resource locker when a test case timeout waiting to lock some  resource the test runner will skip that test and send it to a a retry queue. and when all the tests pass, the test runner run the retry tests queue (until the retry number of all functions will be achieved).

example:
```python 
@lock_resource(resource='some_resource_id', timeout=60, retry=2)
def test_some_test _that_need_resource(self):
     pass
```
or in a function :
```python
def test_some_test _that_need_resource(self):
    ...
    with locking_resource(resource='some_resource_id', timeout=60)):
[...]

PR #4465, opened 2017-03-16
Replaced tier3 with backup mark (until we find a solution for  #3905)

rochacbruno:
`katello-restore` is a destructive test and we cannot run it now in our automation

So while we do not have a decision about #3905  and #4088 lets mark only as `@backup`  and not run those tests until pipeline is solved.


JacobCallahan:
ACK

PR #4481, opened 2017-03-22
DockerBrowser - name container by test_case for better readibility

rplevka:
This names container after a test case + test which utilizes it. This make debugging much easier and sat-qe readable
It also adds a random 3-char suffix just to prevent name collision

```
          # docker ps
          CONTAINER ID    IMAGE                         COMMAND  CREATED        STATUS              PORTS                     NAMES
before -> 96f56cb4b156    selenium/standalone-firefox   "/..."   44 seconds ago Up 43 seconds       0.0.0.0:33646->4444/tcp   big_torvalds 
after  -> 485f30c7f16f    selenium/standalone-firefox   "/..."   12 seconds ago Up 11 seconds       0.0.0.0:33647->4444/tcp   RplevkaTestCase.test_rplevka_foo_test_bBv
```

### DIFF
--- a/robottelo/ui/browser.py
+++ b/robottelo/ui/browser.py
@@ -1,4 +1,5 @@
 """Tools to help getting a browser instance to run UI tests."""
+from fauxfactory import gen_string
 import logging
 import six
 import time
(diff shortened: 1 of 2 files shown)


robottelo/ui/browser.py
```diff
@@ -159,6 +160,7 @@ def __init__(self):
         self.webdriver = None
         self.container = None
         self._client = None
+        self.name = name
```
lpramuk: or can be ```self._name = name```

rplevka:
done, plus I made the name splitting part a little shorter and nicer

PR #4482, opened 2017-03-22
add support for multiple manifests

rplevka:
This extends the `manifests` module to support multiple manifests (for different scenarios for future need) instead of a single one.

Note that this requires a change to robottelo.properties so some orchestration needs to be done on CI server while merging this.

### DIFF
--- a/robottelo.properties.sample
+++ b/robottelo.properties.sample
@@ -122,8 +122,11 @@ ssh_key=
 
 # Fake manifest testing.
 # [fake_manifest]
-# URL of the base manifest
-# url=http://example.org/valid-redhat-manifest.zip
+# URLs of the base manifests, comma-separated key-value pairs
+# urls=
+#    manifest1=http://example.org/valid-redhat-manifest.zip,
+#    manifest2=http://example.org/another-valid-redhat-manifest.zip,
+#    ...
 # URL of the key file
 # key_url=http://example.org/fake_manifest.key
 # URL of the certificate file
--- a/robottelo/manifests.py
+++ b/robottelo/manifests.py
@@ -14,13 +14,15 @@ from robottelo.config import settings
 
 class ManifestCloner(object):
     """Manifest clonning utility class."""
-    def __init__(self, template=None, signing_key=None):
-        self.template = template
+    def __init__(self, template=None, signing_key=None, manifest="default"):
+        self.manifest = manifest
         self.signing_key = signing_key
+        self.template = template
 
     def _download_manifest_info(self):
         """Download and cache the manifest information."""
-        self.template = requests.get(settings.fake_manifest.url).content
+        self.template = requests.get(
+            settings.fake_manifest.urls.get(self.manifest)).content
         self.signing_key = requests.get(settings.fake_manifest.key_url).content
         self.private_key = serialization.load_pem_private_key(
             self.signing_key,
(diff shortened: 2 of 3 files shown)


robottelo.properties.sample
```diff
@@ -122,8 +122,11 @@ ssh_key=
 
 # Fake manifest testing.
 # [fake_manifest]
```
rplevka: this is now a list of key=value pairs and is being parsed by dict cast function, which was already in place

robottelo/manifests.py
```diff
@@ -103,10 +105,14 @@ def original(self):
             self._download_manifest_info()
         return six.BytesIO(self.template)
 
-
+if not settings.configured:
+    settings.configure()
 # Cache the ManifestCloner in order to avoid downloading the manifest template
 # every single time.
-_manifest_cloner = ManifestCloner()
+_manifest_cloner = {
```
rplevka:
```
In [2]: manifests._manifest_cloner
Out[2]: 
{'default': <robottelo.manifests.ManifestCloner at 0x7f182b12a590>,
 'lite': <robottelo.manifests.ManifestCloner at 0x7f182b12a5d0>}
```

robottelo/manifests.py
```diff
@@ -166,5 +174,6 @@ def original_manifest():
 
         with original_manifest() as manifest:
             # my fancy stuff
+    :parameter manifest: a key from settings.fake_manifest.urls
     """
-    return Manifest(_manifest_cloner.original())
+    return Manifest(_manifest_cloner.original(manifest))
```
elyezer: Does this work? `_manifest_cloner` is a dict not an object. I think this would be something like `_manifest_cloner[menifest].original()`

robottelo/manifests.py
elyezer: I would use a different approach. I would make `ManifestClone` to have an internal dict where users can call clone passing a manifest key (the same key present on properties file) then it would check for the caching dict if the required manifest was already downloaded and then download if necessary. It also would require that  the original method receive the manifest key (or name) so it could return the proper original manifest.

robottelo/manifests.py
```diff
@@ -119,12 +125,13 @@ class Manifest(object):
...
     """
-    def __init__(self, content=None, filename=None):
+    def __init__(self, content=None, filename=None, manifest='default'):
         self._content = content
         self.filename = filename
+        self.manifest = manifest
 
         if self._content is None:
-            self._content = _manifest_cloner.clone()
+            self._content = _manifest_cloner[self.manifest].clone()
```
elyezer: Doing as I suggested on the previous comment this call would be something like `_manifest_cloner.clone(self.manifest)`

robottelo/manifests.py
```diff
@@ -103,10 +105,14 @@ def original(self):
             self._download_manifest_info()
         return six.BytesIO(self.template)
 
-
+if not settings.configured:
```
elyezer: Missed this, but avoid calling settings.configure at import time. Configuring at import time will break documentation generation.

robottelo/manifests.py
rplevka: i missed this bit

ldjebran:
1- Think this module already need a naming convention
    manifest should an instance of Manifest class
    inside Manifest and ManifestCloner classes and in their respective \__init\__ kwars the manifest name should be renamed to name or key or manifest_name
    we simply have too manifest names with different contexts

2- Please run some tests using the default and custom one, 
    even in python console to show that the manifests can be uploaded

what do you think?

Issue #4349, opened 2017-02-15
Review the tests decorated with DUPLICATE BUGS


rochacbruno:
robozilla in 6.2.z  shows us the output containg some **CLOSED_DUPLICATE** BZs used in decorator, and bugs on that resolution should not be used in decorator as it will not have proper resolution.

We must implement the #4263 to follow clones and duplicates, but in some cases we need to decide if the test is going to be removed as in #4348 or if we need to clone the original bug with 6.2.z flag set.

So we need to review this list.

```

brocha@localhost(rob62) :~/P/robottelo|6.2.z✓
➤ robozilla tests/foreman/ --filters decorator
scanning: ../robottelo/tests/foreman/
warn: True
duplicates: False
clones: False
depends: False
filters: decorator
connecting to bugzilla without credentials
found 83 bugs usage in 58 files (occurrences 134) in 0.11 seconds
getting bugs info ...
generating report ...
Handler           | BZ         | State                  | Flags                  | Line -> File
skip_if_bug_open  | 1156555    | CLOSED_WONTFIX         | sat-backlog+           | 127 -> ../robottelo/tests/foreman/api/test_activationkey.py
skip_if_bug_open  | 1302725    | VERIFIED               | sat-6.3.0+             | 199 -> ../robottelo/tests/foreman/api/test_bookmarks.py
skip_if_bug_open  | 1297308    | CLOSED_ERRATA          | sat-6.1.z+             | 176 -> ../robottelo/tests/foreman/api/test_contentview.py
skip_if_bug_open  | 1147100    | ASSIGNED               | sat-backlog+           | 991 -> ../robottelo/tests/foreman/api/test_contentview.py
skip_if_bug_open  | 1242534    | CLOSED_ERRATA          | sat-6.2.0+             | 647 -> ../robottelo/tests/foreman/api/test_contentviewfilter.py
skip_if_bug_open  | 1349364    | VERIFIED               | sat-6.3.0+, sat-6.2.z+ | 129 -> ../robottelo/tests/foreman/api/test_discoveredhost.py
skip_if_bug_open  | 1217635    | CLOSED_WONTFIX         | sat-backlog?           | 610 -> ../robottelo/tests/foreman/api/test_docker.py
skip_if_bug_open  | 1282431    | CLOSED_ERRATA          | sat-6.1.z+             | 1233 -> ../robottelo/tests/foreman/api/test_docker.py
[...]

PR #4485, opened 2017-03-23
Update BZ ids for AK product content tests

abalakh:
Original BZ is flagged for 6.2.z only, updating BZ id to corresponding 6.3 clone

Per https://github.com/SatelliteQE/robottelo/pull/4452#issuecomment-286770741

### DIFF
--- a/tests/foreman/cli/test_activationkey.py
+++ b/tests/foreman/cli/test_activationkey.py
@@ -679,7 +679,7 @@ class ActivationKeyTestCase(CLITestCase):
 
         :CaseLevel: System
 
-        :BZ: 1360239
+        :BZ: 1426386
         """
         result = setup_org_for_a_custom_repo({
             u'url': FAKE_0_YUM_REPO,
@@ -711,7 +711,7 @@ class ActivationKeyTestCase(CLITestCase):
 
         :CaseLevel: System
 
-        :BZ: 1360239
+        :BZ: 1426386
         """
         org = make_org()
         result = setup_org_for_a_rh_repo({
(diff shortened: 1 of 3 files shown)


tests/foreman/cli/test_activationkey.py
```diff
@@ -1222,7 +1222,7 @@ def test_negative_update_autoattach(self):
         self.assertIn(
             u"'--auto-attach': value must be one of", exe.exception.stderr)
 
-    @skip_if_bug_open('bugzilla', 1360239)
+    @skip_if_bug_open('bugzilla', 1435286)
```
abalakh: Seems like another bug is blocking this test for 6.3 :|

tests/foreman/cli/test_activationkey.py
```diff
@@ -1248,20 +1248,21 @@ def test_positive_content_override(self):
...
+                    u'value': int(override_value),
                 })
                 # Retrieve the product content enabled flag
                 content = ActivationKey.product_content({
                     u'id': result['activationkey-id'],
                     u'organization-id': self.org['id'],
                 })
-                self.assertEqual(content[0]['enabled?'], override_value)
+                self.assertEqual(
+                    content[0]['enabled?'], str(override_value).lower())
```
abalakh: Currently returned state is `'true'` or `'false'`, not `'1'` or `'0'` anymore, updating it accordingly.

rochacbruno:
we may use robozzila to get another cases like this. 
I have a pending issue #4349 and I'll include this check there.

PR #4497, opened 2017-03-24
[6.1.z] DockerBrowser - name container by test_case for better readibility

rplevka:
cherry-picking ccd00d3d12cd1a7f4900ab32030e1f48844b53ab

### DIFF
--- a/robottelo/ui/browser.py
+++ b/robottelo/ui/browser.py
@@ -2,6 +2,7 @@
 import six
 import time
 
+from fauxfactory import gen_string
 from robottelo.config import settings
 from selenium import webdriver
 
@@ -64,7 +65,7 @@ def browser():
 
 class DockerBrowser(object):
     """Provide a browser instance running inside a docker container."""
-    def __init__(self):
+    def __init__(self, name=None):
         if docker is None:
             raise DockerBrowserError(
                 'Package docker-py is not installed. Install it in order to '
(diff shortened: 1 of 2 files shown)


robottelo/ui/browser.py
```diff
@@ -173,6 +176,8 @@ def _create_container(self):
             host_config=self._client.create_host_config(
                 publish_all_ports=True),
             image='selenium/standalone-firefox',
+            name=self._name.split('.', 4)[-1] + '_{0}'.format(
```
elyezer: If name is None then this will raise an exception. Better to check if the name is set before calling split.

rplevka:
@elyezer, how about now?

robottelo/ui/browser.py
```diff
@@ -1,4 +1,6 @@
 """Tools to help getting a browser instance to run UI tests."""
+from fauxfactory import gen_string
```
elyezer: Move the `from .. import ..` after the `import` lines

rplevka: addressed

elyezer:
Travis is not happy because ``/robottelo/ui/browser.py:2:1: F401 'logging' imported but unused``

renzon:
@rplevka code is ok, just merge it after removing unused import Travis is complaining about.

rplevka:
@renzon sorry for the delay. addressed.

PR #4516, opened 2017-03-30
[6.2.z] add vm __enter__ exception handling

ldjebran:
no exception handing in VirtualMachine __enter__
python console log with this code : http://pastebin.test.redhat.com/470225

python console log with this code and put  a = 1/0  at the end of create to simulate a ZeroDivisionError : http://pastebin.test.redhat.com/470227


rplevka:
ACK

Issue #4501, opened 2017-03-27
RemoteExecution tests - REX needs to be enabled for capsule


rplevka:
Apparently, the capsule is not set to support REX. The REX tests currently fail with:

```LANG=en_US.UTF-8  hammer -v -u admin -p changeme  job-invocation output --host="139624527552784.foo.com" --id="1"
2017-03-23 03:14:17 - robottelo.ssh - DEBUG - <<< stdout
Failed to initialize command: Dynflow::ExecutionPlan::Steps::Error - Could not use any Capsule. Consider configuring remote_execution_global_proxy or remote_execution_fallback_proxy in settings```

PR #4528, opened 2017-03-31
add rex capsule to subnet

pondrejk:
Related to issue #4501

### DIFF
--- a/tests/foreman/cli/test_remoteexecution.py
+++ b/tests/foreman/cli/test_remoteexecution.py
@@ -17,6 +17,7 @@
 """
 from datetime import datetime, timedelta
 from fauxfactory import gen_string
+from nailgun import entities
 from robottelo import ssh
 from robottelo.cleanup import vm_cleanup
 from robottelo.cli.base import CLIReturnCodeError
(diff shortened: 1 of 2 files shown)


robottelo/helpers.py
```diff
@@ -280,6 +281,20 @@ def add_remote_execution_ssh_key(hostname, key_path=None, **kwargs):
     ssh.add_authorized_key(server_key, hostname=hostname, **kwargs)
 
 
+def add_remote_execution_capsule_to_subnet(hostname, capsule_id=1):
```
rplevka: Honestly, by name `add_remote_execution_capsule_to_subnet` I would understand a method that accepts `subnet` and `capsule` as parameters. This is a little misleading as it actually takes a host, and parses a subnet assigned to it.

This makes the method too specific.

rplevka: Honestly, by name add_remote_execution_capsule_to_subnet I would understand a method that accepts subnet and capsule as parameters. This is a little misleading as it actually takes a host, and parses a subnet assigned to it.

This makes the method too specific.

rplevka: btw, I personally wouldn't create a helper for every single 'scenario', i don't see a problem in having the following 4 lines of code directly in a test (at least i know what is really going on). But I guess this is very subjective - not a blocker

renzon: I like helpers because you read them once and can continue reading other testing with the high level understand if it. Besides I agree on better naming it. I added a discussion about it on Friday mtg.

pondrejk:
@rplevka addressed, it turned out there was no existing subnet associated with the vm hosts, so I had to first create it, than add the rex capsule to it

rplevka:
thanks, please cherry-pick this

PR #4537, opened 2017-04-03
Add missing `edge` support to saucelabs

rochacbruno:
I tested the new support of Microsoft Edge and it works on SauceLabs

So this PR adds missing `edge` support to saucelabs


rochacbruno:
Also matches names from https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/common/desired_capabilities.py

Issue #4499, opened 2017-03-24
Test repo sync, CV publish, CV promote, capsule sync workflow


dkliban:
1. Create a repository repoFoo that contains a-0.1.rpm and b-1.0.rpm. 
2. Register capsule with satellite
3. Create custom productFoo in Satellite and associate repoFoo with it.
4. Create lifecycle environment lceFoo.
5. Associate lceFoo with the capsule.
6. Create Content View cvFoo.
7. Associate productFoo repository with cvFoo.
8. Sync productFoo repository.
9. Publish new version of cvFoo.
10. Promote cvFoo to lceFoo.
11. Assert that a task to sync lceFoo to the capsule is started.
12. Assert that the content of the published cvFoo in lceFoo is exactly the same as content of repoFoo.
13. Once capsule sync finishes assert that the content published on the capsule for cvFoo in lceFoo is exactly the same as repoFoo.
14. Sync productFoo repo a second time. 
15. Assert that the publish task from previous step contains a summary that says the publish was skipped because content had not changed.

omaciel:
@abalakh mind taking this and prioritizing it for automation?

abalakh:
@omaciel sure thing :)

dkliban:
@abalakh and I decided that this should be automated using Katello API.

pgagne:
@abalakh Flag me as a reviewer when you're ready to do a pull request on this.

rochacbruno:
@dkliban please include any BZ link if you have related to this test case

PR #4552, opened 2017-04-07
[6.2.z] Implemented initial test for content management

abalakh:
Depends on SatelliteQE/nailgun#396
Depends on SatelliteQE/robottelo-ci#581

* Updated capsule deployment helper to use correct capsule repo (our internal for downstream, cdn for cdn)
* Updated rhel repo download policy in capsule deployment helper to 'on_demand' (significant performance inrease, syncing time decreased from 40 min to 15 min for me locally)
* Added helper to form local path to published repository (some of our existing tests can use that)
* Added helper to create a repository (not satellite's repository entity, but a group of rpms with repodata)
* Added helper to fetch specific repository's revision
* Added new dir for pulp/content management related tests
* Implemented the first content management test described in #4499

### DIFF
--- a/robottelo/cli/factory.py
+++ b/robottelo/cli/factory.py
@@ -2891,6 +2891,17 @@ def setup_capsule_virtual_machine(capsule_vm, org_id=None, lce_id=None,
         raise CLIFactoryError(result.return_code, result.stderr,
                               u'foreman installer failed at capsule host')
 
+    # manually start pulp_celerybeat service if BZ1446930 is open
+    if bz_bug_is_open(1446930):
+        result = capsule_vm.run('systemctl status pulp_celerybeat.service')
+        if 'inactive (dead)' in result.stdout[2]:
+            result = capsule_vm.run('systemctl start pulp_celerybeat.service')
+            if result.return_code != 0:
+                raise CLIFactoryError(
+                    'Failed to start pulp_celerybeat service\n{}'.format(
+                        result.stderr)
+                )
+
     capsule = Capsule.info({'name': capsule_vm.hostname})
 
     if organization_ids:
--- a/robottelo/helpers.py
+++ b/robottelo/helpers.py
@@ -15,7 +15,11 @@ from robottelo import ssh
 from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.proxy import CapsuleTunnelError
 from robottelo.config import settings
-from robottelo.constants import RHEL_6_MAJOR_VERSION, RHEL_7_MAJOR_VERSION
+from robottelo.constants import (
+    PULP_PUBLISHED_YUM_REPOS_PATH,
+    RHEL_6_MAJOR_VERSION,
+    RHEL_7_MAJOR_VERSION,
+)
 
 # This conditional is here to centralize use of lru_cache and urljoin
 if six.PY3:  # pragma: no cover
@@ -475,14 +479,13 @@ def form_repo_path(org=None, lce=None, cv=None, cvv=None, prod=None,
     if not any([lce, cvv]):
         raise ValueError('Either `lce` or `cvv` is required')
 
-    base_path = '/var/lib/pulp/published/yum/http/repos'
     if lce:
         repo_path = '{}/{}/{}/custom/{}/{}'.format(org, lce, cv, prod, repo)
     elif cvv:
         repo_path = '{}/content_views/{}/{}/custom/{}/{}'.format(
             org, cv, cvv, prod, repo)
 
-    return os.path.join(base_path, repo_path)
+    return os.path.join(PULP_PUBLISHED_YUM_REPOS_PATH, repo_path)
 
 
 def create_repo(repo_fetch_url, packages, repo_name, hostname=None):
@@ -498,8 +501,7 @@ def create_repo(repo_fetch_url, packages, repo_name, hostname=None):
     :return: URL where the repository can be accessed
     :rtype: str
     """
-    base_path = '/var/lib/pulp/published/yum/http/repos'
-    repo_path = os.path.join(base_path, repo_name)
+    repo_path = os.path.join(PULP_PUBLISHED_YUM_REPOS_PATH, repo_name)
     result = ssh.command(
         'sudo -u apache mkdir -p {}'.format(repo_path), hostname=hostname)
     if result.return_code != 0:
--- a/robottelo/helpers.py
+++ b/robottelo/helpers.py
@@ -12,15 +12,18 @@ from tempfile import mkstemp
 from time import sleep
 from nailgun.config import ServerConfig
 from robottelo import ssh
+from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.proxy import CapsuleTunnelError
 from robottelo.config import settings
 from robottelo.constants import RHEL_6_MAJOR_VERSION, RHEL_7_MAJOR_VERSION
 
-# This conditional is here to centralize use of lru_cache
+# This conditional is here to centralize use of lru_cache and urljoin
 if six.PY3:  # pragma: no cover
     from functools import lru_cache  # noqa
+    from urllib.parse import urljoin  # noqa
 else:  # pragma: no cover
     from cachetools.func import lru_cache  # noqa
+    from urlparse import urljoin  # noqa
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


robottelo/vm.py
```diff
@@ -249,6 +249,9 @@ def install_katello_agent(self):
             raise VirtualMachineError('Failed to install katello-agent')
         gofer_check = self.run('service goferd status')
         if gofer_check.return_code != 0:
+            self.run('service goferd start')
```
ldjebran: perhaps we should add a function parameter to try to start goferd service with False default value, to not break the current tests and in capsule helper call this function with true,

```python
    def install_katello_agent(self, start_goferd=False):
        ...
        if gofer_check.return_code != 0 and start_goferd:
            self.run('service goferd start')
        ...


```

abalakh: Frankly i'm hoping while this PR gets 2 acks and all the issues resolved, katello-agent BZ will become fixed (as it's in progress right now) and i'll just remove these lines 😄 
Otherwise i'd like to insist on adding a workaround here in any acceptable way, as it blocks too much tests for us, even those which aren't directly related (like capsule or content management tests).

abalakh:
Just for the record, to remove `[WIP]` tag, 3 issues have to be resolved:
- `pulp_celerybeat.service` is refusing to start automatically for me. It's most likely env issue as it's working ok for @ldjebran , still, i have to make sure (and remove corresponding workaround when it's resolved)
- katello-agent is not starting automatically and we've got NACK on workaround for it. If we don't change the approach - this PR will be blocked while BZ1431747 is not fixed yet
- for some reason for me repo.sync() doesn't fetch a new repo (after adding it with repocreate) from the first attempt, only from second sync. Need to investigate more on it and either disregard, or fill a bz and add a workaround.

robottelo/helpers.py
```diff
@@ -413,3 +414,86 @@ def default_url_on_new_port(oldport, newport):
...
+    :param str repo_name: repository name - name of a directory with packages
+        and repodata
+    :param str optional hostname: hostname or IP address of the remote host. If
+        ``None`` the hostname will be get from ``main.server.hostname`` config.
+    :return: URL where the repository can be accessed
+    :rtype: str
+    """
+    repo_path = '/var/lib/pulp/published/yum/http/repos'
+    result = ssh.command(
+        'mkdir -p {}/{}'.format(repo_path, repo_name), hostname=hostname)
```
ldjebran: change naming:
```python
# this can be moved to constants as PULP_PUBLISHED_REPOS_PATH
repos_path = '/var/lib/pulp/published/yum/http/repos'
repo_path = os.path.join(repos_path, repo_name)
result = ssh.command(
        'mkdir -p {}}'.format(repo_path), hostname=hostname)

```

robottelo/helpers.py
```diff
@@ -413,3 +414,86 @@ def default_url_on_new_port(oldport, newport):
...
+    if result.return_code != 0:
+        raise CLIReturnCodeError(
+            result.return_code, result.stderr, 'Unable to create repo dir')
+    for package in packages:
+        result = ssh.command(
+            '(cd {}/{} && wget {}{})'.format(
+                repo_path, repo_name, repo_fetch_url, package),
```
ldjebran: same as above use os.path.join
and use urljoin for repo_fetch_url and  package

robottelo/cli/factory.py
```diff
@@ -2679,14 +2681,30 @@ def setup_capsule_virtual_machine(capsule_vm, org_id=None, lce_id=None,
         except CLIReturnCodeError as err:
             raise CLIFactoryError(
                 u'Failed to fetch repository info\n{0}'.format(err.msg))
-    # Synchronize the RH repositories
-    for rh_repo in rh_repos:
+    # If we aren't working with CDN, create custom repo with latest capsule
+    # repo available
+    if not settings.cdn and settings.capsule_repo:
```
ldjebran: think we should raise exception when not settings.cdn and not settings.capsule_repo

abalakh: i thought about this, it won't be user-friendly - i suspect some of our folks don't have `capsule_repo` and `sattools_repo` settings set, by introducing exception we'll block them till they update properties file. And also for robottelo users outside our team this change will basically mean 'if you don't want to run the tests against cdn, you have to have downstream repos for satellite tools and satellite capsule' which is probably not the best approach.
Basically, these lines mean 'if we're in downstream then try to use internal downstream build if available, if not - at least use latest released one' (which will work in most cases btw).

ldjebran: @abalakh , so by this logic you will end without exception and without the necessary repos in content view

abalakh: @ldjebran good catch, i've messed with commits. The checks above should be `if setting.cdn or not settings.capsule_repo`. This way repos will be included.

abalakh: @ldjebran i've updated #4555 as that's a better place to introduce this change. Let's continue our discussion there.

robottelo/helpers.py
abalakh: No objections on using os.path.join, but what's the benefit of using urljoin in this particular case? I'm not constructing url from different parts here, i'm building a new one using scheme, hostname, port and static path. Can't see any use in urljoin here, maybe you meant urlunparse or urlunsplit (which are overkill here too, imo)?

abalakh: Just curious - what's the benefit of using urlparse here? It has the same drawback:
```python
In [20]: urljoin('http://example.com/some/repo', 'package.rpm')
Out[20]: 'http://example.com/some/package.rpm'
```
I'll add extra check on trailing slash, but so far switching to urljoin makes no sense for me

abalakh:
Depends on #4555

abalakh:
Recent changes:
* Removed invalid assertions (revision id on satellite will always change after repo sync)
* Fixed bug with wrong owner for repo folder on satellite
* Added workaround for BZ1446930
* Updated test to be able to re-use existing capsule
* Moved capsule creation/deletion to setUpClass/tearDownClass accordingly

abalakh:
Made a squash of all the commits, rebase and basically PR is ready for review. Test results with reusing existing capsule:
```python
py.test -v tests/foreman/api/test_contentmanagement.py
platform darwin -- Python 2.7.13, pytest-3.0.7, py-1.4.33, pluggy-0.4.0 -- /Users/andrii/workspace/env62z/bin/python
cachedir: .cache
plugins: xdist-1.15.0, services-1.1.14, cov-2.3.1
collected 1 items
2017-05-05 14:33:07 - conftest - DEBUG - Found WONTFIX in decorated tests ['1110476', '1156555', '1269196', '1402826', '1245334', '1221971', '1217635', '1226425', '1199150', '1204686', '1267224', '1103157', '1230902', '1214312', '1079482']

2017-05-05 14:33:07 - conftest - DEBUG - Collected 1 test cases


Implemented initial test for content management


tests/foreman/api/test_contentmanagement.py::ContentManagementTestCase::test_positive_capsule_sync PASSED

========================== 1 passed in 336.04 seconds ==========================
```
Will post test results without reusing capsule in ~1h.

abalakh:
Test results:
```python
py.test -v tests/foreman/api/test_contentmanagement.py
platform darwin -- Python 2.7.13, pytest-3.0.7, py-1.4.33, pluggy-0.4.0 -- /Users/andrii/workspace/env/bin/python
cachedir: .cache
plugins: xdist-1.15.0, services-1.1.14, cov-2.3.1
collected 1 items
2017-05-05 14:57:45 - conftest - DEBUG - Found WONTFIX in decorated tests ['1110476', '1156555', '1269196', '1402826', '1245334', '1221971', '1217635', '1226425', '1199150', '1204686', '1267224', '1103157', '1230902', '1214312', '1079482']

2017-05-05 14:57:45 - conftest - DEBUG - Collected 1 test cases


tests/foreman/api/test_contentmanagement.py::ContentManagementTestCase::test_positive_capsule_sync PASSED

========================= 1 passed in 2172.58 seconds ==========================
```

robottelo/helpers.py
```diff
@@ -448,3 +451,89 @@ def get_services_status():
...
+    :param str repo_fetch_url: URL to fetch packages from
+    :param packages: list of packages to fetch (with extension)
+    :param str repo_name: repository name - name of a directory with packages
+        and repodata
+    :param str optional hostname: hostname or IP address of the remote host. If
+        ``None`` the hostname will be get from ``main.server.hostname`` config.
+    :return: URL where the repository can be accessed
+    :rtype: str
+    """
+    base_path = '/var/lib/pulp/published/yum/http/repos'
```
ldjebran: think this can be a subject of a constant variable
```python
PULP_PUBLISHED_YUM_REPOS_PATH = '/var/lib/pulp/published/yum/http/repos'
```
what do you think?

PR #4561, opened 2017-04-12
Added stubs/testcases for compute resource of type EC2

san7ket:
Added stubs for tier1 , tier2 and tier3 test cases for EC2 Compute resource

### DIFF
--- a/tests/foreman/ui/test_computeresource_ec2.py
+++ b/tests/foreman/ui/test_computeresource_ec2.py
@@ -1,0 +1,413 @@
+"""Test for compute resource UI
+
+:Requirement: Computeresource
+
+:CaseLevel: Acceptance
+
+:CaseComponent: UI
+
+:TestType: Functional
+
+:CaseImportance: High
+
+:Upstream: No
+"""
+
+from robottelo.decorators import run_only_on, stubbed, tier1, tier2, tier3
+from robottelo.test import UITestCase
+
+
+class Ec2ComputeResourceTestCase(UITestCase):
+    """Implements EC2 compute resource tests in UI"""
+
+    @run_only_on('sat')
+    @stubbed()
+    @tier1
+    def test_positive_create_ec2_with_name(self):
+        """Create a new ec2 compute resource with valid name
+
+        :id: 4c74d04a-a276-4d6a-b080-77b2b64942ef
+
+        :setup: ec2 hostname and credentials.
+
+        :steps:
+            1. Create a compute resource of type ec2.
+            2. Provide a valid Access Key and Secret Key.
+            3. Provide a valid name to ec2 compute resource.
+            4. Test the connection using Load Regions and submit.
+
+        :expectedresults: An ec2 compute resource is created
+            successfully.
+
+        :Caseautomation: notautomated
+
+        :CaseImportance: Critical
+        """
+
+    @run_only_on('sat')
+    @stubbed()
+    @tier1
+    def test_positive_create_ec2_with_description(self):
+        """Create ec2 compute resource with description.
+
+        :id: bca54de0-ac53-432c-85af-6f5b842cd10b
+
+        :setup: ec2 hostname and credentials.
+
+        :steps:
+            1. Create a compute resource of type ec2.
+            2. Provide a valid Access Key and Secret Key.
+            3. Provide a valid description to ec2 compute resource.
+            4. Test the connection using Load Regions and submit.
+
+        :expectedresults: An ec2 compute resource is created successfully
+
+        :Caseautomation: notautomated
+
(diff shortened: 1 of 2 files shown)


tests/foreman/ui/test_computeresource_ec2.py
```diff
@@ -0,0 +1,382 @@
...
+
+        :Caseautomation: notautomated
+        """
+
+    @run_only_on('sat')
+    @stubbed()
+    @tier3
+    def test_positive_provision_ec2_with_custom_compute_settings(self):
+        """ Provision a host on ec2 compute resource with
+         custom flavour and disk.
```
kbidarkar: I doubt, if we can update the disk to a custom value, can you please check this.

tests/foreman/ui/test_computeresource_ec2.py
```diff
@@ -0,0 +1,382 @@
...
+               to add ssh-key parameter.
+
+        :steps:
+            1. Populate images into satellite from ec2 images.
+            2. Associate Activation key and CV to the host.
+            3. Edit the required Operating system to use the custom
+               provisioning template.
+            4. Go to "Hosts --> New host".
+            5. Fill in the required details.(eg name,loc, org)(hostgroup)
+            6. Select ec2 compute resource from "Deploy on" drop down.
```
kbidarkar: I think, we should be expanding this a bit more, for example:
1) Interfaces Tab: Deselect subnet option
2) Virtual Machine Tab: Mention the option to select
3) Parameters Tab: Ensure the sshkey parameter is added by visible depending upon whether we are using HG.

Can you please update this also at other relevant places?

tests/foreman/ui/test_computeresource_ec2.py
```diff
@@ -0,0 +1,382 @@
...
+        :id: 441ba03a-98e7-4475-a0d5-56bad23e645b
+
+        :setup: ec2 hostname, credentials and provisioning setup.
+
+        :steps:
+            1. Go to "Hosts --> New host".
+            2. Fill in the required details.(eg name,loc, org).
+            3. Select ec2 custom compute resource from "Deploy on" drop
+               down.
+            4. Select the custom compute profile" with custom disk size and
```
kbidarkar: the opening or closing double quotes are missing here.

tests/foreman/ui/test_computeresource_ec2.py
```diff
@@ -0,0 +1,382 @@
...
+
+        :setup: ec2 hostname, credentials and images in ec2.
+
+        :steps:
+
+            1. Create a compute resource of type ec2.
+            2. Provide a valid Access Key and Secret Key.
+            3. Select the created ec2 CR and click images tab.
+            4. Select "New image" , provide it valid name and information.
+            5. Select the desired Amazon Machine Image (ami) to create image
```
kbidarkar: also, let's mention here that the user which we need to select needs to be "ec2-user" only, for any template commands to get executed on the host.

tests/foreman/ui/test_computeresource_ec2.py
san7ket: Will do, but shouldn't step 3 go in setup phase of the test case, just wondering.

kbidarkar: I feel 3 will be part of the test-case and hence, as we have to ensure it exists for every host, so that we can login, with our own keys, as foreman uses it's own keys during configuring the instance.

tests/foreman/ui/test_computeresource_ec2.py
```diff
@@ -0,0 +1,381 @@
...
+    @stubbed()
+    @tier3
+    def test_positive_provision_ec2_host_with_image(self):
+        """ Provision a host on ec2 compute resource with image based
+
+        :id: 5b68bfb8-6c15-4629-b25a-8928afdc7ca9
+
+        :setup: ec2 hostname and credentials.
+            1. Addition of custom provisioning template
+               to add ssh-key parameter.
```
kbidarkar: This is perfectly fine here as part of setup, below in the steps, we just need to make sure the parameter is added for each host and I feel it depends differently upon whether we are using hostgroup or not.

san7ket:
Comments addressed.

tests/foreman/ui/test_computeresource_ec2.py
kbidarkar: Virtual Machine Tab: Mention the actual "options to select". :)
Also the same for, Parameters Tab:  Those are just suggestions to update as an example. I mean if HG exists no need to update , if creating host without HG, then how to add and so on.

kbidarkar:
ACK

PR #4449, opened 2017-03-15
[6.2.z] KatelloAgent: Use internal/CDN Tools appropriately

lpramuk:
- add new settings: cdn, capsule_repo, sattools_repo
- use internal/CDN Tools according to `cdn` setting

(addresses #4429 for 6.2.z)

PR #4439, opened 2017-03-14
KatelloAgent: Use internal/CDN Tools appropriately

lpramuk:
- add new settings: cdn, capsule_repo, sattools_repo
- use internal/CDN Tools according to `cdn` setting

(addresses #4429 for master)

Issue #4429, opened 2017-03-13
KatelloAgentTestCase (and generally all client testing) uses released bits


lpramuk:
In this test case we use packages out of CDN, thus we test already released packages. But we should test the packages that are going to be released (i.e. out of internal repositories).

Specifically I spotted it in KatelloAgentTestCase but generally all client testing is performed that way, i.e. out of CDN (we hit client issues together with CUs)

Lets use internal candidate repos instead.

lpramuk:
Unfortunately it brought us unwanted fruits already:
- https://bugzilla.redhat.com/show_bug.cgi?id=1431747

kbidarkar:
The infra required for this already exists in robottelo.properties file and the necessary stuff have been put in place, this requires only changes in the robottelo repos. I thought this was fixed.

I would suggest, this needs to be fixed as below,

```
If cdn == true;
   use repos from GA  ( the current existing method )
else:
   repo == settings.robottelo.sattools_repo ( use the dogfood link )
```
robottelo-ci already updates the right link in the robottelo.properties file.


https://gitlab.redhat.com/satelliteqe/jenkins-configs/blob/master/robottelo.properties#L82

kbidarkar:
This was the issue which was raised in 2015, https://github.com/SatelliteQE/robottelo/issues/3093

sghai:
Guys, we need to fix this on priority..

kbidarkar:
Some more details, for folks, who would be working on this, https://github.com/SatelliteQE/robottelo-ci/blob/master/scripts/satellite6-automation.sh#L47

The above script mentions how things are populated in the robottelo.properties file.

kbidarkar:
- [ ] api/test_errata.py:    setup_org_for_a_rh_repo
- [ ] cli/test_activationkey.py:    setup_org_for_a_rh_repo
- [ ] cli/test_host.py:    setup_org_for_a_rh_repo
- [ ] ui/test_contenthost.py:    setup_org_for_a_rh_repo
- [ ] ui/test_hostunification.py:        setup_org_for_a_rh_repo
- [ ] ui/test_activationkey.py:        setup_org_for_a_rh_repo
- [ ] ui/test_hostcollection.py:    setup_org_for_a_rh_repo
- [ ] ui/test_errata.py:    setup_org_for_a_rh_repo

PR #4555, opened 2017-04-11
[6.2.z] Refactor katello-agent related tests to use cdn/downstream repos accordingly

abalakh:
Basically reverts all the changes done with tests in #4439 #4449 #4462 #4463 and introduces another approach instead - helpers, not tests are updated to use the proper repo, this way tests require from little to no changes.
PR covers all places mentioned in #4429 so basically fully resolves it for 6.2.z branch
It also resolves issue #4153 as it's a blocker for downstream repo usage (assertions mentioned there are always failing for downstream repo, it was working for us only in cdn)

PR #4572, opened 2017-04-18
[6.3] Refactor katello-agent related tests to use cdn/downstream repos accordingly

abalakh:
Basically reverts all the changes done with tests in #4439 #4449 #4462 #4463 and introduces another approach instead - helpers, not tests are updated to use the proper repo, this way tests require from little to no changes.
PR covers all places mentioned in #4429 so basically fully resolves it for master branch
It also resolves issue #4153 as it's a blocker for downstream repo usage (assertions mentioned there are always failing for downstream repo, it was working for us only in cdn)

### DIFF
--- a/robottelo/cli/factory.py
+++ b/robottelo/cli/factory.py
@@ -2514,7 +2514,7 @@ def setup_org_for_a_custom_repo(options=None):
     }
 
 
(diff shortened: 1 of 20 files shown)


abalakh:
Please note that this PR is *not* a cherry-pick of #4555 as branches and tests differ.

PR #4554, opened 2017-04-11
[cherry-pick] additional katello-backup tests

pondrejk:
Cherry pick from #4473
Related to issue #4456

Test results (from standalone automation job number 482):
```

tests/foreman/sys/test_hot_backup.py::HotBackupTestCase::test_positive_incremental PASSED

tests/foreman/sys/test_hot_backup.py::HotBackupTestCase::test_positive_incremental_skip_pulp FAILED

tests/foreman/sys/test_hot_backup.py::HotBackupTestCase::test_positive_online_backup_with_directory_created PASSED

tests/foreman/sys/test_hot_backup.py::HotBackupTestCase::test_positive_online_backup_with_existing_directory PASSED
tests/foreman/sys/test_hot_backup.py::HotBackupTestCase::test_positive_online_incremental <- robottelo/decorators/__init__.py SKIPPED

tests/foreman/sys/test_hot_backup.py::HotBackupTestCase::test_positive_online_incremental_skip_pulp FAILED

tests/foreman/sys/test_hot_backup.py::HotBackupTestCase::test_positive_online_skip_pulp PASSED

tests/foreman/sys/test_hot_backup.py::HotBackupTestCase::test_positive_skip_pulp PASSED
```

note: 2 tests fail because of .pulp.snar is being created when creating incremental backup without pulp content. I'll investigate with devel if this is a bug or intended behavior (not happening in 6.2).

PR #4610, opened 2017-04-25
katello-backup added code handling bz#1445224

pondrejk:
Made a change to handle https://bugzilla.redhat.com/show_bug.cgi?id=1445224
For master, the change is icluded as a separate commit within PR: #4554 

```
nosetests -v tests/foreman/sys/test_hot_backup.py:HotBackupTestCase.test_positive_online_incremental_skip_pulp
Katello-backup with --online --skip-pulp-content and ... 
Ran 1 test in 38.102s

OK

nosetests -v tests/foreman/sys/test_hot_backup.py:HotBackupTestCase.test_positive_incremental_skip_pulp
Katello-backup with --skip-pulp-content and --incremental ... ok

Ran 1 test in 137.489s

OK
```

### DIFF
--- a/tests/foreman/sys/test_hot_backup.py
+++ b/tests/foreman/sys/test_hot_backup.py
@@ -20,7 +20,11 @@
 from fauxfactory import gen_string
 from nailgun import entities
 from robottelo.constants import BACKUP_FILES, HOT_BACKUP_FILES
-from robottelo.decorators import stubbed, destructive, skip_if_bug_open
+from robottelo.decorators import (
+        destructive,
+        skip_if_bug_open,
+        stubbed,
+)
 from robottelo.helpers import get_services_status
 from robottelo.ssh import _get_connection
 from robottelo.test import TestCase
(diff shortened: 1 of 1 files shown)


ldjebran:
Think we have to find a better approach as the test goal is to verify that pulp incremental backup is skipped, the test fail to do that as the incremental file is there, and we are skipping the assertion on that,
 it will give us the impression that the test pass and this is not true.
But this arrive only on rh7, perhaps we have to introduce a new functionality skip_if bug_open to verify the platform too?

perhaps  skipping the whole test or let it fail on the relevant platform, and simply mark the test in doc string with this BZ (and mark the bug qe_test_coverage + ) will be the best approaches for this test,  what do you think?

pondrejk:
@ldjebran thanks for the feedback.

"perhaps we have to introduce a new functionality skip_if bug_open to verify the platform too?" -- yes, that would be useful in cases like this

"perhaps skipping the whole test or let it fail on the relevant platform, and simply mark the test in doc string" -- went with the second option, I don't want to skip the whole test since this bug should not really affect the backup.

Issue #4578, opened 2017-04-19
remote exectution tests failing with SocketError getaddrinfo: Name or service not known


pondrejk:
Appears both in UI and CLI tests:

```
E               AssertionError: False is not true : host output: Error initializing command #<Proxy::RemoteExecution::Ssh::Dispatcher::Command:0x007fb55408d4f8> SocketError getaddrinfo: Name or service not known Exit status: EXCEPTION
```

pondrejk:
This is given by the current infrastructure for creating hosts, sat can not reach vms from other machine by hostname atm. Looking for a solution, a workaround is to enable the global setting remote_execution_connect_by_ip, executing remote jobs by ip address works all right in this scenario.

pondrejk:
The problem should be fixed by creating the vm host with the right bridge interface. Filed #4624 to some testing for remote_execution_connect_by_ip

sghai:
At first place, this is the issue w/ hostname resolution. Just a give a brief background, we are trying to create a VM through `robottelo.vm` that provides the hostname like `140167802457360.DOMAIN` and later we run the REX job on that VM and it failed w/

```
2017-04-27 08:46:51 - robottelo.ssh - INFO - >>> LANG=en_US.UTF-8  hammer -v -u admin -p changeme  job-invocation output --host="140167802457360.lab.eng.rdu2.redhat.com" --id="3"
2017-04-27 08:46:53 - robottelo.ssh - INFO - <<< stdout
Error initializing command #<Proxy::RemoteExecution::Ssh::Dispatcher::Command:0x007f23a0191a20>
SocketError getaddrinfo: Name or service not known
Exit status: EXCEPTION
```

sghai:
This comment clearly explains everything and w/ a workaround of DDNS..
https://github.com/SatelliteQE/robottelo/issues/4101#issuecomment-272400033

I can't think of any other solution. Please respond if anyone has any other suggestion/solution.

sghai:
Complete implementation of DDNS is defined here though not sure how flexible it would be w/ no of hosts..

https://github.com/SatelliteQE/robottelo/blob/6.2.z/robottelo/vm_capsule.py

san7ket:
vm.py will give us the ip and hostname of created host , we can always add those entry in satellite's /etc/hosts file , a ugly solution , but it should be inexpensive. 
We can come up with some module to handle this, along with the "clean-up".

rplevka:
@sghai @san7ket,
[...]

PR #4623, opened 2017-04-26
pass bridge to snap-guest vm

pondrejk:
Required in  #4578, pending tests

### DIFF
--- a/robottelo/vm.py
+++ b/robottelo/vm.py
@@ -50,7 +50,7 @@ class VirtualMachine(object):
     def __init__(
             self, cpu=1, ram=512, distro=None, provisioning_server=None,
             image_dir=None, tag=None, hostname=None, domain=None,
-            target_image=None):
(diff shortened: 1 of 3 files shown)


kbidarkar:
ACK

pondrejk:
see the standalone automation job number 542 to see the test results, basically the bridge is now correctly propagated from robottelo.settings and set to default value if unset. The snap-guest now fails on:

```
ERROR    Cannot get interface MTU on 'sfeature_el6': No such device
Domain installation does not appear to have been successful.
```

This is given by infrastructure as the hypervisor for generated host doesnt have the same interfaces enables as the Sat server does

pondrejk:
Putting back do not merge as this probably doesn't fix the original isssue, see the discussion in #4578 Will keep this open until we come to a decision on how to tackle the problem

rochacbruno:
@pondrejk pending CP to 6.2?

PR #4646, opened 2017-05-03
Add both RHELs' Satellite Tools repos

lpramuk:
Depends on:
- (ready) https://github.com/SatelliteQE/robottelo-ci/pull/612
- (ready) satelliteqe/jenkins-configs/merge_requests/65

Fixes https://github.com/SatelliteQE/robottelo-ci/issues/563 within robottelo on master branch

### DIFF
--- a/robottelo.properties.sample
+++ b/robottelo.properties.sample
@@ -106,7 +106,9 @@ ssh_key=
 # be pulled from there instead of using the CDN channel. These information is
 # more suited to be used for downstream, downstream-iso and zstream builds.
 # capsule_repo=http://capsule/repo
-# sattools_repo=http://sattools/repo
+# sattools_repo=
+#   rhel6=http://sattools/repo/el6,
+#   rhel7=http://sattools/repo/el7,
 
 # Bugzilla data to authenticate API calls
 # [bugzilla]
--- a/robottelo/vm.py
+++ b/robottelo/vm.py
@@ -243,8 +243,10 @@ class VirtualMachine(object):
 
         """
         downstream_repo = None
-        if repo in (REPOS['rhst6']['id'], REPOS['rhst7']['id']):
-            downstream_repo = settings.sattools_repo
+        if repo == REPOS['rhst6']['id']:
+            downstream_repo = settings.sattools_repo['rhel6']
+        elif repo == REPOS['rhst7']['id']:
+            downstream_repo = settings.sattools_repo['rhel7']
         elif repo in (REPOS['rhsc6']['id'], REPOS['rhsc7']['id']):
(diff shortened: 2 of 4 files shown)


abalakh:
Frankly i don't like approach with dicts - it's worse both from user perspective (custom multiline format, extra spaces etc in settings file) and from dev perspective (introduces regex searches for trivial things), that's why i'm suggesting to move it's discussion into separate issue not to put on hold support of rhel6/7 for sat tools repo. But if other folks are fine with dict settings then i won't insist or block PR, it's just imo.

lpramuk:
do you guys like dict approach in settings?

robottelo.properties.sample
```diff
@@ -106,7 +106,9 @@ ssh_key=
 # be pulled from there instead of using the CDN channel. These information is
 # more suited to be used for downstream, downstream-iso and zstream builds.
 # capsule_repo=http://capsule/repo
-# sattools_repo=http://sattools/repo
+# sattools_repo=
```
elyezer: Maybe this should be two separated options, like we have above the rhel6_os and rhel7_os. Mostly all the options that needs both systems they are separated so I was wondering if we should follow that for these too.

lpramuk: The same I can say about rhel6_os, rhel7_os, we can convert them to dict :)

lpramuk: come on guys, using dict in python is like breathing air ;)

robottelo/vm.py
```diff
@@ -244,7 +244,7 @@ def enable_repo(self, repo, force=False):
         """
         downstream_repo = None
         if repo in (REPOS['rhst6']['id'], REPOS['rhst7']['id']):
-            downstream_repo = settings.sattools_repo
+            downstream_repo = settings.sattools_repo[self.distro[:5]]
```
elyezer: I am wondering if we should pick rhel6 when `rhst6` repo is the repo id (same for rhel7) instead of basing the decision on the distro.

robottelo.properties.sample
elyezer: I was just reading the comment you said about using dicts and was writing never mind here.

robottelo/vm.py
lpramuk: sure we can, I refactor it

robottelo.properties.sample
lpramuk: Thanks for your feedback

lpramuk:
@elyezer Thanks. Based on your feedback I refactored it the way @abalakh possibly wanted :-)
- avoided re.search and feeling happy (win - win)

sghai:
ACK

PR #4649, opened 2017-05-05
unskip backup online incremental test

pondrejk:
This skip had wrong bz number attached, retesting shows it is no longer necessary:

```
nosetests -v tests/foreman/sys/test_hot_backup.py:HotBackupTestCase.test_positive_online_incremental
Make an incremental online backup ... ok

Ran 1 test in 674.902s

OK
```


renzon:
@pondrejk could you add the BZ id on @BZ token on docstring? This way we keep a reference for it so we can search latter. Besides that, ACK

pondrejk:
@renzon, I think the correct bz should have been 1445871, updated the docstring accordingly. Cheers

PR #4675, opened 2017-05-11
Renamed bz_bug_is_open to affected_by_bz

### LINKED ISSUE #4262: Rename bz_bug_open to affected_by_bz and improve its logic
That function **NOT THE DECORATOR** will be renamed to `affected_by_bz` and the logic is:

- Return True if:
  - OPEN - Current used approach
  - CLOSED - WONTFIX
  - CLOSED - CANTFIX

- Return False if:
  - CLOSED (verified, onqa) - Current used approach

renzon:
fix #4262

### DIFF
--- a/robottelo/api/utils.py
+++ b/robottelo/api/utils.py
@@ -22,7 +22,7 @@ from robottelo.constants import (
     RHEL_6_MAJOR_VERSION,
     RHEL_7_MAJOR_VERSION,
 )
-from robottelo.decorators import bz_bug_is_open
+from robottelo.decorators import affected_by_bz
 
 
 def enable_rhrepo_and_fetchid(basearch, org_id, product, repo,
@@ -49,7 +49,7 @@ def enable_rhrepo_and_fetchid(basearch, org_id, product, repo,
     r_set.enable(data=payload)
     result = entities.Repository(name=repo).search(
         query={'organization_id': org_id})
-    if bz_bug_is_open(1252101):
+    if affected_by_bz(1252101):
         for _ in range(5):
             if len(result) > 0:
                 break
--- a/robottelo/cleanup.py
+++ b/robottelo/cleanup.py
@@ -6,7 +6,7 @@ from nailgun import entities, signals
 from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.proxy import Proxy
 from robottelo.constants import DEFAULT_ORG_ID
-from robottelo.decorators import bz_bug_is_open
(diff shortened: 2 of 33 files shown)


svtkachenko:
There is another function, `rm_bug_is_open`. I'm not sure that it is used anywhere but perhaps it worth also to rename it to match the new name of `bz_bug_is_open`?

renzon:
@svtkachenko the [rm functon](https://github.com/renzon/robottelo/blob/cee0df495bbae5354fe5d9f15a29e9da05927584/robottelo/decorators/__init__.py#L509-L509) is actually only checking bug status. But on bz we are checking status, clones and other fields. That is what motivated this change, to be correct on term of semantic. So I fell rm should stay like it is know. What you think after this new info?

renzon:
fun fact: coveralls decreased coverage due to a function renaming? really? I tend to agree with @rochacbruno to change to codecov ;)

svtkachenko:
@renzon As these functions are used for the same purpose, I'd like to see uniform names, but what you've said makes sense. So... I don't know :) I do not insist, it was rather a question.

renzon:
@svtkachenko: to be honest, I am not completly sure as well. I am more trying to figure it out than to convince you ;)

I found only [one direct](https://github.com/renzon/robottelo/blob/dba9124b09bbe5a5254b987598d35941b1d544d2/tests/foreman/api/test_activationkey.py#L528-L528) use in our test case. An after looking into [it on Redmine](http://projects.theforeman.org/issues/4638), it's a old issue which sees to be already fixed.

Keeping investigation I found no indirectly use of it , e.g., there is no decorator '@skip_if_bug_open('redmine', 1156555)' on our entire code base.

So it seems that we prepared ourselves to handle multiple but trackers but in fact we are using only bugzilla, appart from the only one redmine issue I mentioned.

So ask people who has more experience (@lpramuk, @elyezer @oshtaier): keeping  redmine support still make sense? Maybe there is something I and @svtkachenko are missing here ;)

If not, we could get rid of redmine and make our @skip_if_bug_decorator having 'bugzilla' as default, so we could use it as '@skip_if_bug_open(bug_id=123)' instead of  '@skip_if_bug_open("bugzila", 123)'

renzon:
In fact after reading this I decide to bring the discussion up on our weekly mtg ;)

elyezer:
@renzon upstream uses redmine, we might want to skip based on that on cases like:

1. A bug is only affecting upstream
2. There is no BZ for that bug (which might get cloned into BZ sometime)

PR #4707, opened 2017-05-19
Initial work to increase coverage on robottelo.cli module

rochacbruno:
using pytest parametrize and pytest-mock
added some tests to assert the call of cli methods
more tests will be added later

NOTE: Using pytest style for robottelo tests, while for `tests/forreman` we keep using `unittest` for compatibility with Junit reports and `subTest`. For `tests/robottelo` we can use `pytest` or even mix it with `unittest`.

### DIFF
--- a/robottelo/api/utils.py
+++ b/robottelo/api/utils.py
@@ -566,7 +566,7 @@ def configure_puppet_test():
     puppet_env.location.append(loc)
     puppet_env.organization.append(org)
     puppet_env = puppet_env.update(['location', 'organization'])
-    Proxy.importclasses({
+    Proxy.import_classes({
         u'environment': puppet_env.name,
         u'name': sat6_hostname,
     })
--- a/robottelo/cli/hammer.py
+++ b/robottelo/cli/hammer.py
@@ -33,7 +33,7 @@ def _csv_reader(output):
         data = data.encode('utf8')
     handler = StringIO(data)
 
-    for row in csv.reader(handler):
+    for row in csv.reader(handler):  # pragma: no cover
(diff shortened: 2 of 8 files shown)


robottelo/api/utils.py
```diff
@@ -566,7 +566,7 @@ def configure_puppet_test():
     puppet_env.location.append(loc)
     puppet_env.organization.append(org)
     puppet_env = puppet_env.update(['location', 'organization'])
-    Proxy.importclasses({
+    Proxy.import_classes({
```
rochacbruno: renamed this method to keep the standard with all other method names. (grepped and renamed all references)

robottelo/cli/hammer.py
```diff
@@ -33,7 +33,7 @@ def _csv_reader(output):
         data = data.encode('utf8')
     handler = StringIO(data)
 
-    for row in csv.reader(handler):
+    for row in csv.reader(handler):  # pragma: no cover
```
rochacbruno: Added `no cover` here because it is difficult to merge coverage reports for python 2 and 3 in final report

tests/robottelo/test_hammer.py
```diff
@@ -367,6 +367,8 @@ def test_parse_info(self):
             '    Repo ID:   10',
             ' 2) Repo Name: repo2',
             '    Repo ID:   20',
+            ' 3) Repo Name => repo3',
+            '    Repo ID =>   30',
```
rochacbruno: included `=>` to test this alternative format and hit the parser.

PR #4730, opened 2017-05-24
[6.3] API FIX docker cv test_positive_publish_with_docker_repo

ldjebran:
The was an was an assert error '1.0' != 1.0

```console
(sat-6.3.0) dlezz@elysion:~/projects/robottelo-fork$ py.test -v tests/foreman/api/test_docker.py::DockerContentViewTestCase::test_positive_publish_with_docker_repo
platform linux2 -- Python 2.7.13, pytest-3.0.7, py-1.4.32, pluggy-0.4.0 -- /home/dlezz/.pyenv/versions/sat-6.3.0/bin/python2.7
cachedir: .cache
plugins: xdist-1.15.0, services-1.2.1, mock-1.6.0, cov-2.4.0
collected 14 items 
2017-05-24 17:00:54 - conftest - DEBUG - Deselect of WONTFIX BZs is disabled in settings


tests/foreman/api/test_docker.py::DockerContentViewTestCase::test_positive_publish_with_docker_repo <- robottelo/decorators/__init__.py PASSED

============================================== 1 passed in 26.82 seconds ===============================================
(sat-6.3.0) dlezz@elysion:~/projects/robottelo-fork$ 

```

### DIFF
--- a/tests/foreman/api/test_docker.py
+++ b/tests/foreman/api/test_docker.py
@@ -612,13 +612,13 @@ class DockerContentViewTestCase(APITestCase):
         # Not published yet?
         content_view = content_view.read()
         self.assertIsNone(content_view.last_published)
-        self.assertEqual(content_view.next_version, 1)
+        self.assertEqual(float(content_view.next_version), 1.0)
 
         # Publish it and check that it was indeed published.
         content_view.publish()
         content_view = content_view.read()
         self.assertIsNotNone(content_view.last_published)
-        self.assertGreater(content_view.next_version, 1)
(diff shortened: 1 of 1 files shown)


ldjebran:
@oshtaier, The purpose of the asserts is to check the version number and not the format of the version number, as the test exist I can understand that the format was float and now the format is string. we have no guaranty that the version will not be float again, as we have no restrictions on this.
This  PR fix the issue and stop relying on the format returned (int, float, string) , as this is not the goal of the test.
I simply want this test to stop failing because of the format returned.

PR #4753, opened 2017-05-29
[6.2.z] fix capsule setup on rhel7

ldjebran:
capsule setup fail due to of rh repo 7.2 usage and to vm changing distro values
fail log illustrated here http://pastebin.test.redhat.com/488624
current test
tests/foreman/cli/test_contentview.py::ContentViewTestCase::test_positive_remove_cv_version_from_multi_env_capsule_scenario
The current PR fix the issue but faced bug: https://bugzilla.redhat.com/show_bug.cgi?id=1456446
test log file: http://pastebin.test.redhat.com/488728
after applying the fix in that bug faced/created an other bug 
https://bugzilla.redhat.com/show_bug.cgi?id=1456559
test log
http://pastebin.test.redhat.com/488769

in any cases as per the logs the capsule is successfully created but the failures affected content view publish and capsule content synchronize that are not subject of this PR

### DIFF
--- a/robottelo/cli/factory.py
+++ b/robottelo/cli/factory.py
@@ -2585,13 +2585,13 @@ def _get_capsule_vm_distro_repos(distro):
     rh_repos = []
     if distro == DISTRO_RHEL7:
         # Red Hat Enterprise Linux 7 Server
-        rh_product_arch = PRD_SETS['rhel_72']['arch']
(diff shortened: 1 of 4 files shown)


sghai:
Nice to see that we are catching capsule issues too through automation.. @ldjebran 👍

PR #4837, opened 2017-06-16
ui-rex-connect-by-ip

pondrejk:
adding by ip versions of the ui remote execution tests and changing the provisioning server
related tp #4624

Issue #4624, opened 2017-04-26
add tests for remote_execution_connect_by_ip


pondrejk:
By default, remote jobs are executed on hosts using the hostname -- the remote_execution_connect_by_ip setting is disabled. Automation should test also the scenario when hosts are reached via ip address.

ntkathole:
Please feel free to re-open, if this still needs fixing.

PR #4838, opened 2017-06-16
[6.2.z] cherry pick ui-rex-connect-by-ip

pondrejk:
adding by ip versions of the ui remote execution tests and changing the provisioning server
cherry pick of #4837 
related to #4624


rplevka:
i think this PR is overall good to go, I would just stub the 'non-ip' tests the same way as in master

PR #4875, opened 2017-06-22
[6.3] FIX discoveryrule test_positive_update_org_loc

ldjebran:
the error on current sat 6.3 build 
```
ERROR 2017-06-22 07:48:24 API] 422 Unprocessable Entity
[ERROR 2017-06-22 07:48:24 Exception] Organizations Host group organization sy0aJw must also be associated to the discovery rule
Locations Host group location 103znYRliN must also be associated to the discovery rule
Could not update the rule:
  Organizations Host group organization sy0aJw must also be associated to the discovery rule
  Locations Host group location 103znYRliN must also be associated to the discovery rule
```
The test still affected by the bug even state ON_QA : https://bugzilla.redhat.com/show_bug.cgi?id=1377990
but now it fail as per the bug description
```
>       self.assertIn(new_org['name'], rule['organizations'])
E       KeyError: 'organizations'
```

### DIFF
--- a/tests/foreman/cli/test_discoveryrule.py
+++ b/tests/foreman/cli/test_discoveryrule.py
@@ -345,19 +345,26 @@ class DiscoveryRuleTestCase(CLITestCase):
 
         :expectedresults: Rule was updated and with given org & location.
 
+        :BZ: 1377990
+
         :CaseLevel: Integration
         """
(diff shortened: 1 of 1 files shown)


sghai:
ACK

Issue #4864, opened 2017-06-21
Automate https://bugzilla.redhat.com/show_bug.cgi?id=1418412


chbrown13:
Checked through CLI tests and couldn't find a test that covers this case.

PR #4896, opened 2017-06-28
Added test to automate https://bugzilla.redhat.com/show_bug.cgi?id=1418412

chbrown13:
Pull request for issue #4864 to automatically test that multibyte and latin1 org names are displayed with the same width.

### DIFF
--- a/tests/foreman/cli/test_organization.py
+++ b/tests/foreman/cli/test_organization.py
@@ -1714,13 +1714,12 @@ class OrganizationTestCase(CLITestCase):
         for org in org_names:
             Org.delete({'name': org})
         if org_list:
-            width = len(org_list[0])
             for org_str in org_list:
-                n = 0
-                for c in org_str:
-                    eaw = unicodedata.east_asian_width(c)
+                width = 0
+                for char in org_str:
+                    eaw = unicodedata.east_asian_width(char)
                     if eaw in ["Na", "N", "A", "H"]:  # Narrow, neutral,...
-                        n += 1
+                        width += 1
                     else:  # Wide
-                        n += 2
-                self.assertEqual(n, width)
+                        width += 2
+                self.assertEqual(len(org_list[0]), width)
--- a/tests/foreman/cli/test_organization.py
+++ b/tests/foreman/cli/test_organization.py
@@ -1712,7 +1712,6 @@ class OrganizationTestCase(CLITestCase):
         org_list = [line for line in Org.list(output_format='table') if line]
         self.assertGreaterEqual(len(org_list), len(org_names))
         for org_str in org_list:
-            width = 0
             for char in org_str:
                 width = sum(
                     1 if unicodedata.east_asian_width(char)
--- a/tests/foreman/cli/test_organization.py
+++ b/tests/foreman/cli/test_organization.py
@@ -1707,9 +1707,9 @@ class OrganizationTestCase(CLITestCase):
             displayed with consistent spacing
         """
         org_names = [
-            gen_string('alpha', random.randint(1, 50)),
-            gen_string('latin1', random.randint(1, 50)),
-            gen_string('utf8', random.randint(1, 50))
+            gen_string('alpha', random.randint(1, 30)),
+            gen_string('latin1', random.randint(1, 30)),
+            gen_string('utf8', random.randint(1, 30))
         ]
         for org in org_names:
             make_org({'name': org})
--- a/tests/foreman/cli/test_organization.py
+++ b/tests/foreman/cli/test_organization.py
@@ -17,6 +17,7 @@
 :Upstream: No
 """
 import random
+import unicodedata
 
 from fauxfactory import gen_string
 from itertools import cycle
@@ -1695,32 +1696,34 @@ class OrganizationTestCase(CLITestCase):
         self.assertEqual(org['id'], result['id'])
 
     @tier1
+    @skip_if_bug_open('bugzilla', 1418412)
     def test_positive_multibyte_latin1_org_names(self):
         """Hammer Multibyte and Latin-1 Org names break list pagination
 
-        :id: 35840da7-668e-4f78-990a-738aa688d586
+        :id: 4fa0afe7-6d0a-4c3e-a0fc-4ecb95c50fc9
 
         :BZ: 1418412
 
-        :expectedresults: List of organization names should have consistent spacing
+        :expectedresults: Multibyte and latin1 names need to be
+            displayed with consistent spacing
 
         :CaseImportance: Critical
         """
-        lat = gen_string('latin1', random.randint(1, 100))
-        utf = gen_string('utf8', random.randint(1, 100))
-        num = gen_string('numeric', random.randint(1, 100))
-        abc = gen_string('alpha', random.randint(1, 100))
-        cjk = gen_string('cjk', random.randint(1, 100))
-        make_org({'name':lat})
-        make_org({'name':utf})
-        make_org({'name':num})
-        make_org({'name':abc})
-        make_org({'name':cjk})
+        org_names = valid_org_names_list()
+        for org in org_names:
+            make_org({'name': org})
         org_list = Org.list(output_format='table')
-        Org.delete({'name':lat})
-        Org.delete({'name':utf})
-        Org.delete({'name':num})
-        Org.delete({'name':abc})
-        Org.delete({'name':cjk})
-        for o in org_list:
-            print o, len(o.decode('ascii'))
+        for org in org_names:
+            Org.delete({'name': org})
+        if org_list:
+            width = len(org_list[0])
+            for org_str in org_list:
+                if org_str:
+                    n = 0
+                    for c in org_str:
+                        eaw = unicodedata.east_asian_width(c)
+                        if eaw in ["Na", "N", "A", "H"]:  # Narrow, neutral,...
+                            n += 1
+                        else:  # Wide
+                            n += 2
+                    self.assertEqual(n, width)
--- a/tests/foreman/cli/test_organization.py
+++ b/tests/foreman/cli/test_organization.py
@@ -1706,11 +1706,17 @@ class OrganizationTestCase(CLITestCase):
         :expectedresults: Multibyte and latin1 names need to be
             displayed with consistent spacing
         """
-        org_names = valid_org_names_list()
+        org_names = [
+            gen_string('alpha', random.randint(1, 50)),
+            gen_string('latin1', random.randint(1, 50)),
+            gen_string('utf8', random.randint(1, 50))
+        ]
         for org in org_names:
             make_org({'name': org})
         org_list = [line for line in Org.list(output_format='table') if line]
         self.assertGreaterEqual(len(org_list), len(org_names))
+        for name in org_names:
+            self.assertTrue(any(name in line for line in org_list))
         for org_str in org_list:
             width = sum(
                 1 if unicodedata.east_asian_width(char)
--- a/tests/foreman/cli/test_organization.py
+++ b/tests/foreman/cli/test_organization.py
@@ -1693,3 +1693,34 @@ class OrganizationTestCase(CLITestCase):
         org = make_org()
         result = Org.info({'name': org['name']})
         self.assertEqual(org['id'], result['id'])
+
+    @tier1
+    def test_positive_multibyte_latin1_org_names(self):
+        """Hammer Multibyte and Latin-1 Org names break list pagination
+
+        :id: 35840da7-668e-4f78-990a-738aa688d586
+
+        :BZ: 1418412
+
+        :expectedresults: List of organization names should have consistent spacing
+
+        :CaseImportance: Critical
+        """
+        lat = gen_string('latin1', random.randint(1, 100))
+        utf = gen_string('utf8', random.randint(1, 100))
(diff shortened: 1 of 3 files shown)


chbrown13:
Test results

[root@localhost robottelo-fork]# pytest tests/foreman/cli/test_organization.py -k 'test_positive_multibyte_latin1_org_names'
==================================================================================================================================== test session starts ====================================================================================================================================
rootdir: /home/chris/Documents/robottelo-fork, inifile:
plugins: cov-2.5.1, mock-1.6.0, xdist-1.17.1, services-1.2.1
collected 70 items 
2017-06-28 14:39:02 - conftest - DEBUG - Deselect of WONTFIX BZs is disabled in settings


tests/foreman/cli/test_organization.py .

==================================================================================================================================== 69 tests deselected ====================================================================================================================================
========================================================================================================================= 1 passed, 69 deselected in 137.14 seconds =========================================================================================================================

tests/foreman/cli/test_organization.py
```diff
@@ -1693,3 +1694,36 @@ def test_positive_info_by_name(self):
...
+        """Hammer Multibyte and Latin-1 Org names break list pagination
+
+        :id: 4fa0afe7-6d0a-4c3e-a0fc-4ecb95c50fc9
+
+        :BZ: 1418412
+
+        :expectedresults: Multibyte and latin1 names need to be
+            displayed with consistent spacing
+
+        :CaseImportance: Critical
```
JacobCallahan: why is this critical?

tests/foreman/cli/test_organization.py
```diff
@@ -1693,3 +1694,36 @@ def test_positive_info_by_name(self):
...
+            make_org({'name': org})
+        org_list = Org.list(output_format='table')
+        for org in org_names:
+            Org.delete({'name': org})
+        if org_list:
+            width = len(org_list[0])
+            for org_str in org_list:
+                if org_str:
+                    n = 0
+                    for c in org_str:
```
JacobCallahan: this line makes the ```if org_str:``` line unnecessary

tests/foreman/cli/test_organization.py
chbrown13: My bad, copy-paste error

tests/foreman/cli/test_organization.py
chbrown13: I think we need to have a check here because the last item in the list is empty string and will always fail the assertion

tests/foreman/cli/test_organization.py
```diff
@@ -1693,3 +1694,32 @@ def test_positive_info_by_name(self):
         org = make_org()
         result = Org.info({'name': org['name']})
         self.assertEqual(org['id'], result['id'])
+
+    @tier1
+    @skip_if_bug_open('bugzilla', 1418412)
```
oshtaier: Redundant statement as defect is already in `VERIFIED` state

oshtaier:
Please, make sure that you have one commit in PR. It is not a strict rule, but we try to follow it.
In the other hand, it is fine to have more commits when you addressing other people review comments to make the process easier

tests/foreman/cli/test_organization.py
```diff
@@ -1693,3 +1694,32 @@ def test_positive_info_by_name(self):
...
+        :BZ: 1418412
+
+        :expectedresults: Multibyte and latin1 names need to be
+            displayed with consistent spacing
+        """
+        org_names = valid_org_names_list()
+        for org in org_names:
+            make_org({'name': org})
+        org_list = Org.list(output_format='table').remove('')
+        for org in org_names:
```
oshtaier: Do we make cleanup procedure here? Or what the purpose to remove all created organizations?

tests/foreman/cli/test_organization.py
```diff
@@ -1693,3 +1694,32 @@ def test_positive_info_by_name(self):
...
+
+        :BZ: 1418412
+
+        :expectedresults: Multibyte and latin1 names need to be
+            displayed with consistent spacing
+        """
+        org_names = valid_org_names_list()
+        for org in org_names:
+            make_org({'name': org})
+        org_list = Org.list(output_format='table').remove('')
```
abalakh: `remove()` returns None, your org_list will always appear empty.

abalakh: In addition, `remove()` will delete only the first found element (empty string in your case), so in case there're few empty lines, this approach won't work. Try
```python
org_list = [line for line in Org.list(output_format='table') if line]
```
instead.

tests/foreman/cli/test_organization.py
abalakh: Same question. Cleanup started too early, in case organization deletion is broken, the test won't even test anything. I'd rather either left orgs untouched or removed them as the last step, in some `finally:` block.

tests/foreman/cli/test_organization.py
```diff
@@ -1693,3 +1694,32 @@ def test_positive_info_by_name(self):
...
+        for org in org_names:
+            Org.delete({'name': org})
+        if org_list:
```
abalakh: This check is basically wrong. In case org_list is empty - no steps  will be performed and test will show us false positive. And, unfortunately, that's what really happened - `.remove('')` returned you `None`, `if org_list:` was skipped as org_list is None, test showed you 'PASSED' but it verified nothing.

Use assertions instead, `self.assertGreaterEqual(len(org_list), len(org_names))` in this case - this way you'll ensure we have enough orgs in output and fail the test otherwise. And, as a bonus, it will reduce unnecessary indentation :)

I'd also think of `self.assertIn` to ensure concrete orgs are present in output - as otherwise (lets say orgs with special characters appeared on the next page) the test won't really verify bugzilla's bug.

tests/foreman/cli/test_organization.py
```diff
@@ -1693,3 +1694,32 @@ def test_positive_info_by_name(self):
...
+            for org_str in org_list:
+                width = 0
+                for char in org_str:
+                    eaw = unicodedata.east_asian_width(char)
+                    if eaw in ["Na", "N", "A", "H"]:  # Narrow, neutral,...
+                        width += 1
+                    else:  # Wide
+                        width += 2
```
abalakh: Nitpick: Multiple checks and multiple loops made a huge indentation. How about a bit different style to reduce it: for each char add +1 width, and, if it's not in `["Na", "N", "A", "H"]`, add extra +1? Smth like that:
```python
for char in org_str:
    eaw = unicodedata.east_asian_width(char)
    width += 1
    if eaw not in ["Na", "N", "A", "H"]:
        width += 1
```

abalakh: Personally i'd use different format instead:
```python
for org_str in org_list:
    width = sum(
        1 if unicodedata.east_asian_width(char) in ["Na", "N", "A", "H"]
        else 2
        for char in org_str
    )
    self.assertEqual(len(org_list[0]), width)
```

chbrown13:
Thanks all, changes have been pushed

tests/foreman/cli/test_organization.py
```diff
@@ -1693,3 +1694,29 @@ def test_positive_info_by_name(self):
...
+            make_org({'name': org})
+        org_list = [line for line in Org.list(output_format='table') if line]
+        self.assertGreaterEqual(len(org_list), len(org_names))
+        for org_str in org_list:
+            width = 0
+            for char in org_str:
+                width = sum(
+                    1 if unicodedata.east_asian_width(char)
+                    in ["Na", "N", "A", "H"]
+                    else 2 for char in org_str
```
abalakh: This `sum()` is what's gonna calculate `width` of `org_str` for you, you don't need `width = 0`and `for char in org_str:` anymore

chbrown13: Fixed

abalakh: Not exactly :) The comprehension inside `sum()` contains a loop already which will navigate through each character in line. This way you don't need extra loop `for char in org_str:` outside, it basically breaks all logic which currently sounds like `for each char in line calculate sum of all characters inside that char ` :)

chbrown13: Oops, sorry my bad. Removed

abalakh:
@chbrown13 test results, please? Make sure to provide test results each time you change test logic, it's both preventing mistakes in test logic and proving your changes are valid.

chbrown13:
Latest test results:

[root@localhost robottelo-fork]# pytest -vv tests/foreman/cli/test_organization.py -k 'test_positive_multibyte_latin1_org_names'
==================================================================== test session starts =====================================================================
platform linux2 -- Python 2.7.13, pytest-3.0.7, py-1.4.34, pluggy-0.4.0 -- /usr/bin/python
cachedir: .cache
rootdir: /home/chris/Documents/robottelo-fork, inifile:
plugins: cov-2.5.1, mock-1.6.0, xdist-1.17.1, services-1.2.1
collected 70 items 
2017-06-29 10:50:40 - conftest - DEBUG - Deselect of WONTFIX BZs is disabled in settings


tests/foreman/cli/test_organization.py::OrganizationTestCase::test_positive_multibyte_latin1_org_names PASSED

==================================================================== 69 tests deselected =====================================================================
========================================================== 1 passed, 69 deselected in 85.80 seconds ==========================================================

chbrown13:
@abalakh Is just the test result enough or is the entire verbose test output preferred?

abalakh:
@chbrown13 test results is enough, test output is too big for a comment and basically contains sensitive information (like internal hostnames)

tests/foreman/cli/test_organization.py
```diff
@@ -1693,3 +1694,27 @@ def test_positive_info_by_name(self):
...
+        """Hammer Multibyte and Latin-1 Org names break list pagination
+
+        :id: 4fa0afe7-6d0a-4c3e-a0fc-4ecb95c50fc9
+
+        :BZ: 1418412
+
+        :expectedresults: Multibyte and latin1 names need to be
+            displayed with consistent spacing
+        """
+        org_names = valid_org_names_list()
```
abalakh: `valid_org_names_list()` is decorated with `@filtered_datapoint` which means it returns only 1 element if `run_one_datapoint` is set to `true` in `robottelo.properties`.
It may work ok for you locally (if `run_one_datapoint=false`), but `run_one_datapoint` is set to `true` on our CI, so instead of entire list you'll actually get only 1 random item (which may be something different from latin1/utf8).
In addition, according to test name and BZ description issue is not about alpha/alphanumeric names, so instead of reusing `valid_org_names_list()` which won't work anyway, consider forming your own list with cjk/latin1/utf8 names.

tests/foreman/cli/test_organization.py
```diff
@@ -1693,3 +1694,27 @@ def test_positive_info_by_name(self):
...
+        for org in org_names:
+            make_org({'name': org})
+        org_list = [line for line in Org.list(output_format='table') if line]
```
abalakh: What if Org.list() hasn't returned our latin1 names at all (because of a bug or pagination)? Imo it's worth asserting `org_names` elements are present inside strings of `org_list`.

chbrown13: Ok, how should we handle the case where the name is not in the list? Just fail the test?

abalakh: Yep, just use assertions, in case assertion won't pass it will fail the test.

abalakh: Quick example off the top of my head:
```python
for name in org_names:
    self.assertTrue(any(name in line for line in org_list))
```

chbrown13: Ok cool, thanks. I kept running into an error and realized when long org names are cut off with "...". Just fixed it and pushed.

chbrown13:
Test results:

[root@localhost robottelo-fork]# pytest -vv tests/foreman/cli/test_organization.py -k 'test_positive_multibyte_latin1_org_names'
==================================================================================================================================== test session starts ====================================================================================================================================
platform linux2 -- Python 2.7.13, pytest-3.0.7, py-1.4.34, pluggy-0.4.0 -- /usr/bin/python
cachedir: .cache
rootdir: /home/chris/Documents/robottelo-fork, inifile:
plugins: cov-2.5.1, mock-1.6.0, xdist-1.17.1, services-1.2.1
collected 70 items 
2017-06-29 12:13:59 - conftest - DEBUG - Deselect of WONTFIX BZs is disabled in settings


tests/foreman/cli/test_organization.py::OrganizationTestCase::test_positive_multibyte_latin1_org_names PASSED

==================================================================================================================================== 69 tests deselected ====================================================================================================================================
========================================================================================================================= 1 passed, 69 deselected in 40.14 seconds ==========================================================================================================================

tests/foreman/cli/test_organization.py
abalakh: What's the max name length which is not shortened? I mean, have you tried 50 chars explicitly? `random.randint(1, 50)` could return you 10 chars and you could miss the fact it will actually fail for 50 chars 'cause boundary value is, let's say, 30 chars. Better to double check as otherwise test can fail randomly :)

chbrown13:
Makes sense, 30 will definitely work. Pushed

On Thu, Jun 29, 2017 at 12:18 PM, Andrii Balakhtar <notifications@github.com

sghai:
@oshtaier : Pending on you for final comment and merge :)

Issue #4889, opened 2017-06-27
datafactory valid_emails_list actually produces invalid emails


pondrejk:
The way valid_emails_list() produces emails forces user tests to do additional treatment like

```escaped_email = email.replace('"', r'\"').replace('`', r'\`')```

This should be done in the function itself so that the generated emails are really valid

PR #4905, opened 2017-06-29
fixed valid emails list

pondrejk:
Related to issue #4889

```
 py.test tests/foreman/cli/test_user.py -k test_positive_update_email
=================================== test session starts ====================================
platform linux2 -- Python 2.7.11, pytest-3.0.7, py-1.4.34, pluggy-0.4.0
plugins: xdist-1.17.1, services-1.2.1, mock-1.6.0, html-1.10.0, cov-2.3.1
collected 49 items 
2017-06-29 17:54:36 - conftest - DEBUG - Deselect of WONTFIX BZs is disabled in settings


tests/foreman/cli/test_user.py .

=================================== 48 tests deselected ====================================
======================== 1 passed, 48 deselected in 186.12 seconds =========================
```

### DIFF
--- a/robottelo/datafactory.py
+++ b/robottelo/datafactory.py
@@ -166,7 +166,7 @@ def valid_data_list():
 @filtered_datapoint
 def valid_emails_list():
     """Returns a list of valid emails."""
-    return [
+    email_list = [
         u'{0}@example.com'.format(gen_string('alpha')),
         u'{0}@example.com'.format(gen_string('alphanumeric')),
         u'{0}@example.com'.format(gen_string('numeric')),
@@ -179,9 +179,9 @@ def valid_emails_list():
             gen_string('alphanumeric'),
             gen_string('alphanumeric'),
         ),
-        u'"():;"@example.com',
         u'!#$%&*+-/=?^`{|}~@example.com',
     ]
+    return [email.replace('"', '').replace('`', '') for email in email_list]
 
 
 @filtered_datapoint
--- a/tests/foreman/cli/test_user.py
+++ b/tests/foreman/cli/test_user.py
@@ -104,10 +104,7 @@ class UserTestCase(CLITestCase):
         """
         for email in valid_emails_list():
             with self.subTest(email):
-                # The email must be escaped because some characters to not fail
-                # the parsing of the generated shell command
-                escaped_email = email.replace('"', r'\"').replace('`', r'\`')
-                user = make_user({'mail': escaped_email})
+                user = make_user({'mail': email})
                 self.assertEqual(user['email'], email)
 
     @tier1
@@ -806,8 +803,7 @@ class UserWithCleanUpTestCase(CLITestCase):
             with self.subTest(email):
                 User.update({
                     'id': user['id'],
-                    # escape to avoid bash syntax error
-                    'mail': email.replace('"', r'\"').replace('`', r'\`'),
+                    'mail': email,
                 })
                 result = User.info({'id': user['id']})
                 self.assertEqual(result['email'], email)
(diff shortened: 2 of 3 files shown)


robottelo/datafactory.py
```diff
@@ -179,9 +179,9 @@ def valid_emails_list():
             gen_string('alphanumeric'),
             gen_string('alphanumeric'),
         ),
-        u'"():;"@example.com',
```
svtkachenko: Why this line was removed?
Also it cause unit-test failing as now this function return less results.

robottelo/datafactory.py
```diff
@@ -179,9 +179,9 @@ def valid_emails_list():
...
         u'!#$%&*+-/=?^`{|}~@example.com',
     ]
+    return [email.replace('"', '').replace('`', '') for email in email_list]
```
svtkachenko: So if these characters always will be removed what the point to have them in the `email_list` at all?

pondrejk:
adding requested results from the related ui and api tests:

```
py.test tests/foreman/ui/test_user.py -k test_positive_create_with_email
================================ test session starts ================================
platform linux2 -- Python 2.7.11, pytest-3.0.7, py-1.4.34, pluggy-0.4.0
plugins: xdist-1.17.1, services-1.2.1, mock-1.6.0, html-1.10.0, cov-2.3.1
collected 53 items 
2017-06-30 08:58:58 - conftest - DEBUG - Deselect of WONTFIX BZs is disabled in settings


tests/foreman/ui/test_user.py .

================================ 52 tests deselected ================================
===================== 1 passed, 52 deselected in 222.45 seconds =====================
 

py.test tests/foreman/api/test_user.py -k test_positive_create_with_email
================================ test session starts ================================
platform linux2 -- Python 2.7.11, pytest-3.0.7, py-1.4.34, pluggy-0.4.0
plugins: xdist-1.17.1, services-1.2.1, mock-1.6.0, html-1.10.0, cov-2.3.1
collected 13 items 
2017-06-30 09:03:54 - conftest - DEBUG - Deselect of WONTFIX BZs is disabled in settings


tests/foreman/api/test_user.py .

================================ 12 tests deselected ================================
====================== 1 passed, 12 deselected in 9.36 seconds ======================
```

robottelo/datafactory.py
pondrejk: Updated the unit test, thanks. I removed that because it was breaking on a cli test, another case that needed escaping, this time the brackets

robottelo/datafactory.py
pondrejk: I think there is no point having them, but I can't force gen_string not to create them, or can I? Basically, I just thought that if something's making trouble, it shouldn't be on the valid data list, so that tests don't have to tweak data afterwards to make them actually valid.

svtkachenko: 1. genstring here used only with `alpha`, `numeric` and `alphanumeric` parameters, hence it generates only [a-zA-Z0-9], no unicode or special characters. Double quotes and grace accent present only in the lines ``` u'"():;"@example.com'``` and ```u'!#$%&*+-/=?^`{|}~@example.com'```. So it would be enough to edit that lines.

2. But. It is valid characters according to RFC5322 [0] [1] and I doubt about removing them. Shell (or CLI) problem it is a shell problem, the fact that bash requires escaping some chars should not affect proper email validation and API/UI tests.

[0] https://en.wikipedia.org/wiki/Email_address#Local-part
[1] https://tools.ietf.org/html/rfc5322

abalakh: +1 to what @svtkachenko said.

pondrejk: Well the question is if such emails will ever be used by anyone, though if you guys feel strongly about this, I'll pull it back, no big deal

PR #4921, opened 2017-07-04
Removed duplicated test & updated associated BZs

abalakh:
`test_positive_host_associations` and `test_positive_list_hosts` are testing basically the same. The confusion was introduced by 3 (!) BZs describing the same issue: BZ1344033, BZ1372826, BZ1394388.
Removing `test_positive_list_hosts` as it's more complicated, slower and is not cherry-picked to 6.2.z (`test_positive_host_associations` is present in both branches)


sghai:
ACK

PR #4922, opened 2017-07-04
Updated docstring of UI test_positive_host_associations

abalakh:
As there're 3 BZ describing the issue already, adding all of them not to forget about them. Also this change will update 6.2.z according to master's PR #4921


sghai:
ack

PR #4941, opened 2017-07-10
[6.3 Feature] Parametrized Subnets CLI stubs

jyejare:
Note:
There is no provision to list and delete the subnet parameters from CLI hammer, and hence the bugs are raised for same. But all together my bugs have been closed as duplicate due to wrong option to create the subnet parameters is existing in current satellite hammer. The bugzilla that corrects the fixes the wrong option and missing options such as listing, deleting parameters in subnet has been added with BZ : 1426612

Also there was no way to list the global params from CLI hammer and hence the RFE is raised for same : 1470014.

All the tests have been skipped with these bugs and decorator skip_if_bug_open().

### DIFF
--- a/tests/foreman/cli/test_subnet.py
+++ b/tests/foreman/cli/test_subnet.py
@@ -25,8 +25,9 @@ from robottelo.cli.factory import make_domain, make_subnet, CLIFactoryError
 from robottelo.cli.subnet import Subnet
 from robottelo.constants import SUBNET_IPAM_TYPES
 from robottelo.datafactory import filtered_datapoint, valid_data_list
-from robottelo.decorators import run_only_on, tier1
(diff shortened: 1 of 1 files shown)


jyejare:
@SatelliteQE/quality-engineers , The 6.3 feature stubs here are pending from long time to review. Folks please mind reviewing them.

ldjebran:
if bug is affect all the tests of the tests case  put it on setupClass

renzon:
@ldjebran I read it again and thought that putting BZ on setUp could lead to one problem: Someone can add a test on the class not related to BZ and not noting it on setupClass. In that case would be better having it explicit on each test. What you think?

ldjebran:
@renzon also true, so ACK and merge

PR #4946, opened 2017-07-11
Drop some data from invalid_emails datafactory.

svtkachenko:
Though successive dots violates RFC, it was decided not to be more restrictive than gmail (and other mail services) by both foreman devs and by `mail` ruby gem devs as such emails are common on the wild.

BZ [1455501](https://bugzilla.redhat.com/show_bug.cgi?id=1455501) closed as NOTABUG.
More information: https://github.com/theforeman/foreman/pull/4587#pullrequestreview-45746763

Though I'm not sure whether these cases should be just dropped from `invalid_emails_list` or moved to `valid_emails_list`, but I've chose the first option.

### DIFF
--- a/robottelo/datafactory.py
+++ b/robottelo/datafactory.py
@@ -69,7 +69,16 @@ def generate_strings_list(length=None, exclude_types=None, bug_id=None,
 
 @filtered_datapoint
 def invalid_emails_list():
-    """Returns a list of invalid emails."""
+    """
+    Returns a list of invalid emails.
+
+    Based on RFC 5321 and 5322, however consecutive dots are removed from
+    the list, as such emails, e.g. `email@example..c` or `dot..dot@example.com`
+    are common on the wild and it was decided to treat them as valid.
+
+    For more information, see `Bugzilla #1455501:
+    <https://bugzilla.redhat.com/show_bug.cgi?id=1455501>`_.
+    """
     return [
         u'foreman@',
         u'@foreman',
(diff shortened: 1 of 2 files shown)


robottelo/datafactory.py
```diff
@@ -70,24 +70,17 @@ def generate_strings_list(length=None, exclude_types=None, bug_id=None,
...
         u'@',
         u'Abc.example.com',
         u'A@b@c@example.com',
-        u'email@example..c',
         # total length 255:
         u'{0}@example.com'.format(gen_string('alpha', 243)),
         u'{0}@example.com'.format(gen_string('html')),
         u's p a c e s@example.com',
-        u'dot..dot@example.com'
     ]
```
ldjebran: we can add a comment saying "emails with successive dots was removed as per bug: 1455501 CLOSED NOTABUG"

tests/robottelo/test_datafactory.py
```diff
@@ -47,7 +47,7 @@ def test_filtered_datapoint_True(self):
                         return_value=False):
```
abalakh: Is mocking bz_bug_is_open still needed for any other helper?

svtkachenko: Not needed, but does not make any harm either and can be used in case other datafactory function would be altered by bz :)
But I think I'll remove that to keep code cleaner.

svtkachenko:
All comments addressed.

Issue #4957, opened 2017-07-12
Investigate why Saucelabs unable to capture PASS/FAIL status of the tests being run for Sat6.3


kbidarkar:
It looks like, Saucelabs is unable to capture PASS/FAIL status of the tests being run for Sat6.3 Jobs.

But it did work fine with Sat6.2 Jobs.

Investigate as to why for Sat6.3 Jobs, Saucelabs is unable to capture PASS/FAIL status and only mention COMPLETED status.

May be a good place to start would be https://wiki.saucelabs.com/display/DOCS/Setting+Test+Status+to+Pass+or+Fail

kbidarkar:
Was investigating into this for the past some time and realized that for Sat6.3 we do not have `sauceclient` installed and may be hence the issue.

For Sat6.2 , it's added to the `requirement.txt` file.

Will raise a PR shortly to fix this.

kbidarkar:
SauceLabs started to capture PASS/FAIL status after this issue was fixed.

PR #4960, opened 2017-07-12
Added sauceclient so that pass/fail status is known to saucelabs

kbidarkar:
Fixes the issue #4957

### DIFF
--- a/requirements-optional.txt
+++ b/requirements-optional.txt
@@ -15,9 +15,6 @@ sphinx
 # For running UI tests within a docker browser
 docker-py
 
-# For reporting test results on SauceLabs
-sauceclient
-
 # For 'manage' interactive shell
 manage>=0.1.13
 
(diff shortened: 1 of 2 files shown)


kbidarkar:
@rochacbruno , wondering if we should move `sauceclient` from `requirements-optional.txt` to `requirements.txt`. What do you think ?

kbidarkar:
I mean to say, probably remove it from requirements-optional.txt file.

abalakh:
I would rather nack such change. For local runs there's no need in `sauceclient` and most of us won't use it as we don't use saucelabs. If some lib is optional, not needed for tests execution and its absence introduces no issues for generic workflow - then it should be placed to `requirements-optional`. Our CI boxes should be updated to install `sauceclient` from `requirements-optional` instead.

kbidarkar:
Ok, I agree, let me make this part of the `robottelo-ci` project then, will move `sauceclient` to `requirements-optional`, but would also ensure `sauceclient` would get installed separately after requriements.txt is installed. 

Let me file an issue in robottelo-ci

kbidarkar:
Something like this,

```
pip install -r requirements.txt sauceclient
```

Issue #4983, opened 2017-07-18
Fix compute resource add image test case


san7ket:
The image add test cases are failing due , to no operating system RedHat 7.3 present when the test are run. 
Add the os in the test , if not present.

-  tests.foreman.ui.test_computeresource_rhev.RhevComputeResourceTestCase.test_negative_add_image_rhev_with_invalid_name
- tests.foreman.ui.test_computeresource_rhev.RhevComputeResourceTestCase.test_positive_add_image_rhev_with_name
- tests.foreman.ui.test_computeresource_vmware.VmwareComputeResourceTestCase.test_negative_add_image_vmware_with_invalid_name

rplevka:
this is because it searches for `{0}{1}.{2}.format(os.name, os.major, os.minor` however, that particular field uses `os.description` for listing/searching the os.
The result is:
```
'RedHat7.3' != 'RHEL Server 7.3'
```

san7ket:
Looking on the passed test case for vmware[1] , it doesn't look like it , the OS is searched using the title parameter , which has to be created prior adding image to CR. 
In this test particularly the OS was created, maybe because of some other test and this test-case was passed.

- tests.foreman.ui.test_computeresource_vmware.VmwareComputeResourceTestCase.test_positive_add_image_vmware_with_name

```
2017-07-15 08:53:43 - robottelo.ui.browser - DEBUG - findElement: id: 28 {'using': 'xpath', 'value': u"//div[@id='select2-drop']//input"}
2017-07-15 08:53:45 - robottelo.ui.browser - DEBUG - findElement: id: 28 {'using': 'xpath', 'value': u"//div[@id='select2-drop']//input"}
2017-07-15 08:53:47 - robottelo.ui.browser - DEBUG - sendKeysToElement:  {'id': u'28', 'value': 'RedHat 7.3'}
2017-07-15 08:53:48 - robottelo.ui.base - DEBUG - Assigned value RedHat 7.3 to <|strategy=xpath|value=//div[@id='select2-drop']//input|>
2017-07-15 08:53:48 - robottelo.ui.browser - DEBUG - findElement: id: 29 {'using': 'xpath', 'value': u"//ul/li/div[normalize-space(.)='RedHat 7.3']"}
2017-07-15 08:53:50 - robottelo.ui.browser - DEBUG - clickElement:  {'id': u'29'}
[...]

PR #4998, opened 2017-07-20
Fixes the Image Creation tests for rhev and vmware CR

san7ket:
Issue: https://github.com/SatelliteQE/robottelo/issues/4983

```
 pytest tests/foreman/ui/test_computeresource_vmware.py -k add_image_vmware_with
tests/foreman/ui/test_computeresource_vmware.py ..
 2 passed, 14 deselected in 153.42 seconds 

/home/sjagtap/PycharmProjects/robottelo  pytest tests/foreman/ui/test_computeresource_rhev.py -k add_image_rhev_with
tests/foreman/ui/test_computeresource_rhev.py ..
 2 passed, 16 deselected in 152.15 seconds 
```

### DIFF
--- a/robottelo/ui/computeresource.py
+++ b/robottelo/ui/computeresource.py
@@ -1,6 +1,7 @@
 # -*- encoding: utf-8 -*-
 from robottelo import ssh
 from robottelo.constants import FILTER, FOREMAN_PROVIDERS
+from nailgun import entities
 from robottelo.ui.base import Base, UINoSuchElementError, UIError
 from robottelo.helpers import ProvisioningCheckError
 from robottelo.ui.locators import common_locators, locators, tab_locators
@@ -444,23 +445,43 @@ class ComputeResource(Base):
         resource_profile_form.set_values(**kwargs)
         resource_profile_form.submit()
 
+    def host_provisioning_check(self, ip_addr):
+        """Check the provisioned host status by pinging the ip of host and check
+        to connect to ssh port
 
-def host_provisioning_check(self, ip_addr):
-    """Check the provisioned host status by pinging the ip of host and check
-    to connect to ssh port
+        :param ip_addr: IP address of the provisioned host
+        :return: ssh command return code and stdout
+        """
+        result = ssh.command(
+            u'for i in {{1..60}}; do ping -c1 {0} && exit 0; sleep 20;'
+            u' done; exit 1'.format(ip_addr))
+        if result.return_code != 0:
+            raise ProvisioningCheckError(
+                'Failed to ping virtual machine Error:{0}'.format(
+                    result.stdout))
+        ssh_check = ssh.command(
+            u'for i in {{1..60}}; do nc -vn {0} 22 <<< "" && exit 0; sleep 20;'
+            u' done; exit 1'.format(ip_addr))
+        if ssh_check.return_code != 0:
+            raise ProvisioningCheckError(
+                'Failed to connect to SSH port of the virtual machine')
 
-    :param ip_addr: IP address of the provisioned host
-    :return: ssh command return code and stdout
-    """
-    result = ssh.command(
-        u'for i in {{1..60}}; do ping -c1 {0} && exit 0; sleep 20;'
-        u' done; exit 1'.format(ip_addr))
-    if result.return_code != 0:
-        raise ProvisioningCheckError(
-            'Failed to ping virtual machine Error:{0}'.format(result.stdout))
-    ssh_check = ssh.command(
-        u'for i in {{1..60}}; do nc -vn {0} 22 <<< "" && exit 0; sleep 20;'
-        u' done; exit 1'.format(ip_addr))
-    if ssh_check.return_code != 0:
-        raise ProvisioningCheckError(
-            'Failed to connect to SSH port of the virtual machine')
+    def check_image_os(self, os_name):
+        """Check if the OS is present, if not create the required OS
+
+        :param os_name: OS name to check, and create
+        :return: Created os
+        """
+        # Check if OS that image needs is present or no, If not create the OS
+        result = entities.OperatingSystem().search(query={
+            u'search': u'title="{0}"'.format(os_name)
+        })
+        if result:
+            os = result[0]
+        else:
+            os = entities.OperatingSystem(
+                name=os_name.split(' ')[0],
+                major=os_name.split(' ')[1].split('.')[0],
+                minor=os_name.split(' ')[1].split('.')[1],
+            ).create()
+        return os
--- a/tests/foreman/ui/test_computeresource_vmware.py
+++ b/tests/foreman/ui/test_computeresource_vmware.py
@@ -66,7 +66,7 @@ class VmwareComputeResourceTestCase(UITestCase):
 
         :expectedresults: A vmware compute resource is created successfully.
 
-        :Caseautomation: notautomated
+        :Caseautomation: Automated
 
         :CaseImportance: Critical
         """
@@ -104,7 +104,7 @@ class VmwareComputeResourceTestCase(UITestCase):
 
         :expectedresults: A vmware compute resource is created successfully
 
-        :Caseautomation: notautomated
+        :Caseautomation: Automated
 
         :CaseImportance: Critical
         """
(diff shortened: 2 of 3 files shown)


robottelo/ui/computeresource.py
```diff
@@ -441,3 +442,23 @@ def set_profile_values(self, resource_name, profile_name, **kwargs):
...
+        })
+        os = entities.OperatingSystem().create()
+        if len(result) == 0 or (os_name != result[0].title):
+            os = entities.OperatingSystem(
+                id=os.id,
+                name=os_name.split(' ')[0],
+                major=os_name.split(' ')[1].split('.')[0],
+                minor=os_name.split(' ')[1].split('.')[1],
+            ).update(['name', 'major', 'minor'])
+        return os
```
ldjebran: we are creating an os in any case, as if we are searching by title we do not need to update it,
think this has to be rewritten approximately like this:

```python
 def check_image_os(self, os_name):
        """Check if the OS is present, if not create the required OS

        :param os_name: OS name to check, and create
        :return Created os
        """
        # Check if OS that image needs is present or no, If not create the OS
        result = entities.OperatingSystem().search(query={
            u'search': u'title="{0}"'.format(os_name)
        })
        if result:
            os = result[0]
        else:
            os = entities.OperatingSystem(
                name=os_name.split(' ')[0],
                major=os_name.split(' ')[1].split('.')[0],
                minor=os_name.split(' ')[1].split('.')[1],
            ).create()
        return os

```

san7ket: This will create a issue, that we are tackling right now, as if there will be a default os we will end up deleting at the end of the test and that could be problematic as we have not created it , so we should not delete it. The test will try to delete the used os.i.e Redhat 7.3. To avoid interfering with Default OS , we are creating new os, updating it (if no default os), deleting it. This way we will avoid deleting the Default os.

```
2017-07-24 10:12:46 - nailgun.client - DEBUG - Making HTTP DELETE request to https://qe-sat6-feature-rhel6/api/v2/operatingsystems/34 with options {'verify': Fal
se, 'auth': ('admin', 'changeme'), 'headers': {'content-type': 'application/json'}} and no data.
2017-07-24 10:12:47 - nailgun.client - WARNING - Received HTTP 422 response: {
  "error": {"id":34,"errors":{"base":["RedHat 7.3 is used by david-blasi.","RedHat 7.3 is used by petra-drossos.","RedHat 7.3 is
 used by qe-sat6-feature-rhel6","RedHat 7.3 is used by aimee-vietzke.host.centralci.eng.rdu2.redhat.com","RedHat 7.3 is used by ethan-molinari."]},"full_messages":["RedHat 7.3 is used by david-blasi.","RedHat 7.3 is used by petra-drossos.","RedHat 7.3 is
used by qe-sat6-feature-rhel6t.com","RedHat 7.3 is used by aimee-vietzke..com","RedHat 7.3 is used by ethan-molinari.com"]}
}

2017-07-24 10:12:47 - robottelo - DEBUG - Finished Test: RhevComputeResourceTestCase/test_positive_add_image_rhev_with_name
2017-07-24 10:12:47 - robottelo.test - DEBUG - Saving screenshot /tmp/robottelo/screenshots
```

ldjebran: @san7ket   we are searching by title and we have  u'search': u'title="{0}"'.format(os_name)
in which case we can have os_name != result[0].title

san7ket: Yes, we can do that,  but we have to be sure of deleting the Default OS(Red Hat 7.3 ) , if we create it. And if we use the approach you have mentioned and if the Default OS is already being used by some hosts or associated with resources , we won't be able to delete this.

> P.S  OS is global and not limited to Org's

Now , Why we need to delete the `created` Default OS (Red Hat 7.3) ?
This is because, the default OS is auto-populated by satellite and has all the associations, like provisioning templates , kickstarts , etc.
The OS that we create will not have those, and if satellite tries to create the default os it will fail as we would have created the os with same name.
The problem, we are facing is, 
1) Some other test is deleting the Default OS  (Red Hat 7.3).
2) There are 4 image based tests of which 3 are run , before the satellite auto-populates the Default OS  (Red Hat 7.3).
3) The 4th test is a success, because by the time the 4th test is run, Satellite auto-populates the required Default OS  (Red Hat 7.3).

So, the approach,
1) Create a random OS
2) Check whether we have a default OS. If yes , do nothing
3) If no, update the random os to default os
4) Delete the os  (In both if and else we are able to get and delete a OS)

san7ket: Using the above approach, we will need two returns

```
 def check_image_os(self, os_name):
        """Check if the OS is present, if not create the required OS

        :param os_name: OS name to check, and create
        :return Created os
        """
        # Check if OS that image needs is present or no, If not create the OS
        result = entities.OperatingSystem().search(query={
            u'search': u'title="{0}"'.format(os_name)
        })
        if result:
            os = result[0].title
            default = True
        else:
            os = entities.OperatingSystem(
                name=os_name.split(' ')[0],
                major=os_name.split(' ')[1].split('.')[0],
                minor=os_name.split(' ')[1].split('.')[1],
            ).create()
        return os, default
```
This is to, avoid deleting the default os that is auto populated by satellite, and then we will have to check the `default=True `in the test . So, i prefer the updation logic

abalakh: `os.delete()` at the end of test case is a bad idea as it won't be executed in case of failure. `try: finally:` isn't the best way either as it will make tests look ugly. cleanup func is a way to go. And as you'll need to define a cleanup func, why not checking whether it's ok to delete the os or not inside it's body?

abalakh: Take a look at org_cleanup, capsule_cleanup, vm_cleanup etc.

san7ket: Too much over-head for 3 tests having a cleanup fucn, will into @ldjebran approach without deleting the os.

tests/foreman/ui/test_computeresource_vmware.py
```diff
@@ -321,6 +321,7 @@ def test_positive_add_image_vmware_with_name(self):
```
abalakh: :CaseAutomation: Automated

tests/foreman/ui/test_computeresource_vmware.py
```diff
@@ -376,6 +379,7 @@ def test_negative_add_image_vmware_with_invalid_name(self):
```
abalakh: :CaseAutomation: Automated

robottelo/ui/computeresource.py
abalakh: Why overhead? cleanup funcs aren't hard to implement nor time-consuming, i don't see real issues with them. If there's something you wanna do after executing test steps - there's only two options: teardown or cleanup funcs. I'm not sure os deletion is that mandatory here though.

tests/foreman/ui/test_computeresource_vmware.py
abalakh: Pls also check till the end of the module, i can't leave comments to lines not affected by this PR :) However we definitely need to update that status.

san7ket: I don't understand , why do I keep missing this. :) , will update

robottelo/ui/computeresource.py
san7ket: Had a discussion with @ldjebran , yes will OS deletion is not mandatory. If we create a OS, with associating all the parameters like provisioning templates,kickstart etc. <-- Will do this

san7ket:
```
pytest tests/foreman/ui/test_computeresource_vmware.py -k add_image_vmware_with
test session starts 
2017-07-31 12:38:57 - conftest - DEBUG - Deselect of WONTFIX BZs is disabled in settings
tests/foreman/ui/test_computeresource_vmware.py ..
14 tests deselected 2 passed, 14 deselected in 135.76 seconds 

 pytest tests/foreman/ui/test_computeresource_rhev.py -k add_image_rhev_with
test session starts 
tests/foreman/ui/test_computeresource_rhev.py ..
16 tests deselected 2 passed, 16 deselected in 154.84 seconds 
```

san7ket:
@abalakh @ldjebran Please re-visit.

PR #5005, opened 2017-07-20
Move sauceclient to requriements-optional.txt

kbidarkar:
Let's install the sauceclient along with requirements.txt to
allow the capture of PASS/FAIL by saucelabs.


kbidarkar:
Will also raise a PR for sat6.2 branch soon.

PR #5032, opened 2017-07-26
vm.py - create() - implement proper cleanup if unable to reach the VM

rplevka:
There was a bug in `vm.create()` where we bailed out without destroying & undefining the libvirt domain, if the `create()` failed during obtaining the VM IP or testing the SSH connectivity.

This led to 'dead souls' running on the libvirt CR.

### DIFF
--- a/robottelo/vm.py
+++ b/robottelo/vm.py
@@ -172,6 +172,8 @@ class VirtualMachine(object):
         if result.return_code != 0:
             raise VirtualMachineError(
                 u'Failed to run snap-guest: {0}'.format(result.stderr))
+        else:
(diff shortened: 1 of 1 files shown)


robottelo/vm.py
```diff
@@ -172,6 +172,8 @@ def create(self):
         if result.return_code != 0:
             raise VirtualMachineError(
                 u'Failed to run snap-guest: {0}'.format(result.stderr))
+        else:
+            self._created = True
```
rplevka: the `_created` attribute should be set immediately after we manage to run `snap-guest` successfully.

Issue #5035, opened 2017-07-26
cli end-to-end test deal with 1326101


pondrejk:
At the moment there is a skipping set in https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/endtoend/test_cli_endtoend.py#L421 -- the bug was closed as deferred and consequently the test fails before the provisioning step. We should find some solution to work around the problem.

pondrejk:
apparently fixed in https://github.com/SatelliteQE/robottelo/commit/a1f8c9d98504221a11104248cacbea028f911fb6

PR #5036, opened 2017-07-26
fixing asserts in cli end to end test

pondrejk:
Cli end to end test fails on key error (see standalone automation job 725)

```      
self.assertEqual(
>           content_host['content-view'],
            content_view['name']
        )
E       KeyError: 'content-view'
```

This PR fixes the comparison, see standalone automation job 727. Though there is another issue in that testcase, see #5035


pondrejk:
FYI does not apply to 6.2 version

PR #5059, opened 2017-07-28
[6.3]  fix ssh client settings to have default values in any case

ldjebran:
with config:
```
# [ssh_client]
# Time to wait for the ssh command to finish
# command_timeout=300
# Time to wait for establishing the connection
# connection_timeout=10
```
```python
print settings.ssh_client.command_timeout
300
print settings.ssh_client.connection_timeout
10
```

with config:
```text
[ssh_client]
# Time to wait for the ssh command to finish
command_timeout=500
# Time to wait for establishing the connection
connection_timeout=20
```
```python
settings.configure()
print settings.ssh_client.command_timeout
500
print settings.ssh_client.connection_timeout
20
```

### DIFF
--- a/robottelo/config/base.py
+++ b/robottelo/config/base.py
@@ -762,14 +762,24 @@ class SSHClientSettings(FeatureSettings):
     """SSHClient settings definitions."""
     def __init__(self, *args, **kwargs):
         super(SSHClientSettings, self).__init__(*args, **kwargs)
-        self.command_timeout = None
-        self.connection_timeout = None
+        self._command_timeout = None
+        self._connection_timeout = None
+
+    @property
+    def command_timeout(self):
+        return self._command_timeout if (
+            self._command_timeout is not None) else 300
+
+    @property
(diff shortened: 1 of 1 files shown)


ldjebran:
@abalakh the solution here is more transparent, and does not need any intervention from CI (backward compatible), and work without any settings of these values, in the shell and cmds you will have always the default values, + the changes are in the same place as SSHClient settings as this functionality concern only ssh client

ldjebran:
**Note** we have removed the skip if bug open  decorator from KatelloAgentTestCase class where we have blocking hammer commands, please review this PR as this will block the automation without this changes

abalakh:
@ldjebran 
>the solution here is more transparent

that's arguable
>and does not need any intervention from CI (backward compatible)

they both are backwards compatible

> in the shell and cmds you will have always the default values

`manage shell` always configures settings for us, so not an issue

But ok, don't wanna block it as the fix itself is very important and your approach isn't bad at all.

PR #5063, opened 2017-07-28
Added support to run only sys tests to Makefile

kbidarkar:
This support is required so that standalone-automation can only
run these tests.


sghai:
ACK

PR #5075, opened 2017-08-01
Fix UI/packages / Filter search by repo

svtkachenko:
Test results:
```
% pytest tests/foreman/ui/test_packages.py -k 'test_positive_search_in_multiple_repos'
========================================================= test session starts ==========================================================
plugins: xdist-1.18.1, services-1.2.1, mock-1.6.0, html-1.12.0, cov-2.5.1
collected 6 items 
2017-08-01 19:22:56 - conftest - DEBUG - Deselect of WONTFIX BZs is disabled in settings


tests/foreman/ui/test_packages.py .

========================================================== 5 tests deselected ==========================================================
=============================================== 1 passed, 5 deselected in 128.60 seconds ===============================================
```

### DIFF
--- a/robottelo/ui/locators/common.py
+++ b/robottelo/ui/locators/common.py
@@ -144,7 +144,7 @@ common_locators = LocatorDict({
     "kt_clear_search": (
         By.XPATH, "//button[contains(@ng-click, 'searchCompleted = false')]"),
     "kt_search_no_results": (
-        By.XPATH, "//span[@data-block='no-search-results-message']"),
+        By.XPATH, "//table//span[@data-block='no-search-results-message']"),
     "kt_search_button": (
         By.XPATH,
         "//button[@ng-click='table.search(table.searchTerm)']"),
--- a/robottelo/ui/packages.py
+++ b/robottelo/ui/packages.py
@@ -2,7 +2,7 @@
 """Implements Packages UI"""
 
 from robottelo.ui.base import Base, UIError
-from robottelo.ui.locators import locators, tab_locators
+from robottelo.ui.locators import common_locators, locators, tab_locators
 from robottelo.ui.navigator import Navigator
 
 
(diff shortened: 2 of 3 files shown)


robottelo/ui/locators/common.py
```diff
@@ -144,7 +144,7 @@
     "kt_clear_search": (
         By.XPATH, "//button[contains(@ng-click, 'searchCompleted = false')]"),
     "kt_search_no_results": (
-        By.XPATH, "//span[@data-block='no-search-results-message']"),
+        By.XPATH, "//table//span[@data-block='no-search-results-message']"),
```
svtkachenko: There 2 elements with such `@data-block` but only one is visible, the one in the results table.

robottelo/ui/packages.py
```diff
@@ -51,3 +51,13 @@ def check_file_list(self, name, file_list):
...
                 locators['package.content_file'] % package_file)
+
+    def search_in_repo(self, name, repo_name, expecting_results=True):
+        """Search for package and filter results by repository"""
+        self.search(name)
+        self.select_repo(repo_name)
+        # In case we expecting that search should not find any entity
+        if expecting_results is False:
+            return self.wait_until_element(
+                common_locators['kt_search_no_results'])
```
ldjebran: please decrease the timeout to 2 seconds

robottelo/ui/packages.py
```diff
@@ -51,3 +51,13 @@ def check_file_list(self, name, file_list):
         for package_file in file_list:
             self.wait_until_element(
                 locators['package.content_file'] % package_file)
+
+    def search_in_repo(self, name, repo_name, expecting_results=True):
```
ldjebran: This will work if we want to search package in all repositories?

svtkachenko: In that case we should use simple search. Actually, it is possible to pass `All repositories` as `repo_name` but it just has no sense :)

PR #5112, opened 2017-08-09
Update ui.docker/test_positive_create_with_external_registry

svtkachenko:
Depends on SatelliteQE/nailgun/pull/434.

Created registry had no assigned organization hence it was not available for select on UI where context was set to specific organization.
```
pytestrunner.py -p pytest_teamcity /home/qui/code/robottelo/tests/foreman/ui/test_docker.py "-k DockerContainerTestCase and test_positive_create_with_external_registry"
============================= 44 tests deselected ==============================
================== 1 passed, 44 deselected in 269.36 seconds ===================

### DIFF
--- a/tests/foreman/ui/test_docker.py
+++ b/tests/foreman/ui/test_docker.py
@@ -1399,9 +1399,11 @@ class DockerContainerTestCase(UITestCase):
         :CaseLevel: Integration
         """
         repo_name = 'rhel'
-        container_name = gen_string('alphanumeric')
+        container_name = gen_string('alphanumeric').lower()
         registry = entities.Registry(
-            url=settings.docker.external_registry_1).create()
+            organization=[self.organization],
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/foreman/ui/test_docker.py [hunk 1]
```diff
@@ -1399,9 +1399,11 @@ def test_positive_create_with_external_registry(self):
         :CaseLevel: Integration
         """
         repo_name = 'rhel'
-        container_name = gen_string('alphanumeric')
+        container_name = gen_string('alphanumeric').lower()
```
ldjebran: is there a special reason why we transform to lower?

svtkachenko: Sorry, forgot to mention that. There is a change for UI that makes docker containers names looking more nicely -- it splits names with a space by capital letter and search fails. With lowercase there are no changes in the name.
scr

PR #5151, opened 2017-08-16
Fix UI.dashboard.test_positive_content_host_subscription_status

svtkachenko:
```
pytestrunner.py -p pytest_teamcity /home/qui/code/robottelo/tests/foreman/ui/test_dashboard.py "-k DashboardTestCase and test_positive_content_host_subscription_status"

============================= 24 tests deselected ==============================
================== 1 passed, 24 deselected in 593.30 seconds ===================
```

### DIFF
--- a/tests/foreman/ui/test_dashboard.py
+++ b/tests/foreman/ui/test_dashboard.py
@@ -29,6 +29,7 @@ from robottelo.constants import (
 )
 from robottelo.datafactory import gen_string
 from robottelo.decorators import (
+    bz_bug_is_open,
     run_in_one_thread,
     skip_if_bug_open,
     skip_if_not_set,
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_dashboard.py
```diff
@@ -631,12 +632,22 @@ def test_positive_content_host_subscription_status(self):
...
-                    common_locators['kt_search_no_results']))
+                if not bz_bug_is_open(1482150):
+                    self.assertIsNotNone(self.dashboard.wait_until_element(
+                        common_locators['kt_search_no_results']))
+                    self.assertTrue(self.dashboard.validate_chss_navigation(
+                        'Partial', u'subscription_status = partial'))
+                    self.assertIsNotNone(self.dashboard.wait_until_element(
+                        common_locators['kt_search_no_results']))
+                else:
+                    hosts = self.dashboard.find_elements(
```
svtkachenko: @oshtaier I know that using of disposable variables contradicts best practices but three nested function calls looks too ugly :)

tests/foreman/ui/test_dashboard.py
```diff
@@ -631,12 +632,22 @@ def test_positive_content_host_subscription_status(self):
                 set_context(session, org=org.name)
                 self.assertTrue(self.dashboard.validate_chss_navigation(
                     'Invalid', u'subscription_status = invalid'))
-                self.assertIsNotNone(self.dashboard.wait_until_element(
-                    common_locators['kt_search_no_results']))
-                self.assertTrue(self.dashboard.validate_chss_navigation(
-                    'Partial', u'subscription_status = partial'))
-                self.assertIsNotNone(self.dashboard.wait_until_element(
-                    common_locators['kt_search_no_results']))
+                if not bz_bug_is_open(1482150):
```
ldjebran: prefer the inverse, what do you think ? 
```
if bz_bug_is_open(1482150):
    # BZ work-arround 
    ...
else:
   # the normal test situation
   ...
```

tests/foreman/ui/test_dashboard.py
ldjebran: In normal situation when there is a content hots self.dashboard.find_elements(self.contenthost._search_locator() % "") will return all content hosts web elements?

svtkachenko: Yes.

tests/foreman/ui/test_dashboard.py
svtkachenko: Dunno. Currently while you reading the code you'll see more clear code first and only then ugly workaround :)

PR #5242, opened 2017-09-07
Added test for UI content host's errata search as non-admin user (BZ1255515)

abalakh:
https://bugzilla.redhat.com/show_bug.cgi?id=1255515
```python
py.test -v tests/foreman/ui/test_contenthost.py -k test_positive_search_errata_non_admin
cachedir: .cache
collected 12 items
2017-09-07 12:20:00 - conftest - DEBUG - Deselect of WONTFIX BZs is disabled in settings


tests/foreman/ui/test_contenthost.py::ContentHostTestCase::test_positive_search_errata_non_admin PASSED

============================= 11 tests deselected ==============================
================== 1 passed, 11 deselected in 433.05 seconds ===================
```

### DIFF
--- a/tests/foreman/ui/test_contenthost.py
+++ b/tests/foreman/ui/test_contenthost.py
@@ -293,7 +293,14 @@ class ContentHostTestCase(UITestCase):
         user_password = gen_string('alpha')
         default_loc = entities.Location().search(
             query={'search': 'name="{0}"'.format(DEFAULT_LOC)})[0]
-        role = entities.Role().search(query={'search': 'name="Manager"'})[0]
+        role = entities.Role().create()
+        for permission_name in (
+                'view_hosts', 'view_lifecycle_environments',
+                'view_content_views', 'view_organizations'):
+            entities.Filter(
+                permission=entities.Permission(name=permission_name).search(),
+                role=role,
+            ).create()
         entities.User(
             role=[role],
             admin=False,
--- a/tests/foreman/ui/test_contenthost.py
+++ b/tests/foreman/ui/test_contenthost.py
@@ -15,6 +15,7 @@
 
 :Upstream: No
 """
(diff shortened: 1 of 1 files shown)


abalakh:
@ldjebran bug title says `Unable to access errata page of content host with non admin user` and bug description basically states the same `Create non-admin account with Role have all filters`, that's what i was covering. Basically both scenarios have some value - as it's very likely that customer will use default 'Manager' role so we should ensure it works properly, and at the same time minimal set of permissions should be enough.
@oshtaier thoughts? Should i only verify second scenario or preserve this test as well?

svtkachenko:
@abalakh If we'll finally start using Manager role by default for all tests, current test-case would be redundant, isn't it?

abalakh:
@svtkachenko like a lot of other tests :) Lots of bugs will become covered automatically.
But that's a good point anyway, let's use minimal set of permissions where it's possible to get ready for switch to Manager role.

ldjebran:
@abalakh you are covering a bug, that bug was verified by a defined set of permissions, if the test fail with that set of permissions this is a regression but your test may not fail, and no body will know that we have a regression of that bug.
conclusion eg: we cannot cover that bug with Manager role

abalakh:
@ldjebran @svtkachenko addressed.
```python
py.test -v tests/foreman/ui/test_contenthost.py -k test_positive_search_errata_non_admin
cachedir: .cache
collected 12 items
2017-09-08 16:57:42 - conftest - DEBUG - Deselect of WONTFIX BZs is disabled in settings


tests/foreman/ui/test_contenthost.py::ContentHostTestCase::test_positive_search_errata_non_admin PASSED

============================= 11 tests deselected ==============================
================== 1 passed, 11 deselected in 388.25 seconds ===================
```

PR #5255, opened 2017-09-09
Using `root-pass` for host creation is deprecated.

omaciel:
Updated all tests related to host creation to use `root-password` instead of
`root-pass` as the latter argument is deprecated.


omaciel:
We will need to update `tests/foreman/data/hammer_commands.json` as well.

svtkachenko:
APT

abalakh:
```
20:56 abalakht: omaciel++
20:56 redqueen: omaciel now has 107 points of karma.
```
🙂

Issue #4190, opened 2017-01-17
Replace endless loop with a retry/counter


rochacbruno:
Currently we have 3 occurrences of a `while True` in tests

https://github.com/SatelliteQE/robottelo/blob/edd603c49b1025cea8ad18ac21abbee9da59bcc8/tests/foreman/ui/test_discoveredhost.py#L74

and more in https://github.com/SatelliteQE/robottelo/search?utf8=%E2%9C%93&q=%22while+True%22&type=

My suggestions is to replace that endless loop with a retry counter.

```
with retry(times=30):
    # do something, if fails retry...
```

So we avoid ending up with an infinite loop in our jenkins build.

ldjebran:
propose to update like this
```python 

import time


class RetryTimeoutException(Exception):
    """Timeout exception"""


class RetryTimer(object):

    def __init__(self, timeout, start_time=time.time(), sleep_time=1,
                 correct_sleep_time=False, min_sleep_time=0.1,
                 raise_error=False):
        self._raise_error = raise_error
        self._timeout = timeout
        self._start_time = start_time
        self._sleep_time = sleep_time
        self._last_call_time = 0
        self._correct_sleep_time = correct_sleep_time
        self._min_sleep_time = min_sleep_time

    @property
    def elapsed(self):
        return time.time() - self._start_time

    def sleep(self):
        sleep_time = self._sleep_time
        if self._correct_sleep_time and self._last_call_time:
            # make a correction
            sleep_time -= (time.time() - self._last_call_time)
            if sleep_time < self._min_sleep_time:
                sleep_time = self._min_sleep_time
        
        return time.sleep(sleep_time)

    def can_loop(self, timeout=None):
        self._last_call_time = time.time()
        if timeout is None:
            timeout = self._timeout
        if time.time() > (self._start_time + timeout):
            if self._raise_error:
                raise RetryTimeoutException(
                    'Timeout error, time elapsed: {0}'.format(self.elapsed))
            else:
                return False

        return True

    def __call__(self, timeout=None):
        return self.can_loop(timeout=timeout)
```
Ping host became like: 

```python
import subprocess


def _ping_host(host, timeout=60):

    retry = RetryTimer(timeout, sleep_time=5)
    while retry():
        command = subprocess.Popen(
            'ping -c1 {0}; echo $?'.format(host),
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE,
            shell=True
        )
        output = command.communicate()[0]
[...]

PR #5290, opened 2017-09-15
Implemented CLI/UI test for cvv errata exclusion filter for RH content (BZ1455990)

abalakh:
https://bugzilla.redhat.com/show_bug.cgi?id=1455990
Also discovered following blocking BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1492114
```python
py.test tests/foreman/ui/test_contentview.py -k test_positive_publish_rh_content_with_errata_by_date_filter
collected 95 items
2017-09-15 17:47:35 - conftest - DEBUG - Deselect of WONTFIX BZs is disabled in settings


tests/foreman/ui/test_contentview.py s

============================= 94 tests deselected ==============================
```
Test results won't make much sense here, i've used some workarounds like passing different repo, commenting out some sections and few manual steps to make sure code is valid.

### DIFF
--- a/robottelo/ui/contentviews.py
+++ b/robottelo/ui/contentviews.py
@@ -646,6 +646,50 @@ class ContentViews(Base):
         return self.wait_until_element(
             locators['contentviews.version.package_name'] % package_name)
 
+    def fetch_version_packages(self, name, version_name):
+        """Return a list of all the packages inside specific content view
+        version"""
+        self.click(self.version_search(name, version_name))
+        self.click(tab_locators['contentviews.tab_version_packages'])
+        packages = []
+        while True:
+            names = self.find_elements(
+                locators['contentviews.version.package_name'] % '')
+            versions = self.find_elements(
+                locators['contentviews.version.package_version'] % '')
+            releases = self.find_elements(
+                locators['contentviews.version.package_release'] % '')
+            for name, version, release in zip(names, versions, releases):
+                packages.append((name.text, version.text, release.text))
+            next_ = self.find_element(
+                locators['contentviews.version.content_next_page'])
+            if next_ is None:
+                break
+            self.click(next_)
+        return packages
+
+    def fetch_version_errata(self, name, version_name):
+        """Return a list of all the errata inside specific content view
+        version"""
+        self.click(self.version_search(name, version_name))
+        self.click(tab_locators['contentviews.tab_version_errata'])
+        errata = []
+        while True:
+            ids = self.find_elements(
+                locators['contentviews.version.errata_id'] % '')
+            titles = self.find_elements(
+                locators['contentviews.version.errata_title'] % '')
+            types = self.find_elements(
+                locators['contentviews.version.errata_type'] % '')
+            for id_, title, type_ in zip(ids, titles, types):
+                errata.append((id_.text, title.text, type_.text))
+            next_ = self.find_element(
+                locators['contentviews.version.content_next_page'])
+            if next_ is None:
+                break
+            self.click(next_)
+        return errata
+
     def puppet_module_search(self, name, version, module_name):
         """Search for puppet module element in content view version"""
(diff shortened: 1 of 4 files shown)


abalakh:
@ldjebran your suggestions? Unfortunately python doesn't have `do .. while` or `repeat .. until`, this is basically equivalent.

ldjebran:
@abalakh I understand that it has and should work, but they are not equivalent, it's better to have a real condition in there.
suppose there is a bug in next button , and it always there what will happen

robottelo/ui/contentviews.py
```diff
@@ -646,6 +646,50 @@ def package_search(self, name, version_name, package_name,
         return self.wait_until_element(
             locators['contentviews.version.package_name'] % package_name)
 
+    def fetch_version_packages(self, name, version_name):
+        """Return a list of all the packages inside specific content view
+        version"""
+        self.click(self.version_search(name, version_name))
+        self.click(tab_locators['contentviews.tab_version_packages'])
+        packages = []
+        while True:
```
ldjebran: try this:
```python
next_ = True
while next_:
    # do what you want here
    ...
     next_ = self.find_element(
                   locators['contentviews.version.content_next_page'])
     if next_:
         self.click(next_)
```

abalakh: And what's the core difference? How will it help to handle your early mentioned case when we have next button 'broken' and always available?
Imo it will end up in a similar flow with the same drawbacks, the only thing which really differs - few extra checks which decrease readability.

ldjebran: @abalakh I do not agree with you, please look at code, 

will ack and will not block this PR , let other other reviewers decide.

abalakh: I'm not against such approach, i just can't see real benefits in it. @rochacbruno @renzon what'd you suggest? I mean the best pythonic equivalent of `do .. while` or `repeat .. until`

rochacbruno: @abalakh @ldjebran @renzon 

I am against using an endless loop as I pointed in https://github.com/SatelliteQE/robottelo/issues/4190 

For this case I think we should use a `retry` with a limited counter

```python
with retry(times=30, sleep_interval=..., exceptions=...):
    # do things
    next_ = self.find_element(
                   locators['contentviews.version.content_next_page'])
    self.click(next_)  # if this operation fails and raise an error the operation will be retried for more 30 times
```
as you can see in #4190 we need to choose an external package that implements this (there are many of them)

If not using the above approach, I prefer to use @abalakh  version using the `while` with a `break` I think it is more readable.

```python
while True:
    if cond:
        break
```

is the Pythonic version of `repeat....until`

But I would prefer to remove all `while True` from our code and start using a `retry` context.

ldjebran: @rochacbruno even if it's known that this is the replacement of repeat....until, even if I know that this is a working scenario. I will never recommend it's usage (this is a very bad programming habit let it without a condition).

renzon: @ldjebran When I was a Java programmer that community really avoid such king of loop. But there since there we have the mentioned construction (do while), code readability isn't compromised most of times.

Once I started to code in Python I realized people are ok with infinite loop with some break condition inside it to improve readbility, not having to create a flag only to control flux.

So in this case I agree with @rochacbruno. Using a a retry function would be the best, otherwise I would go with while True: break.

ldjebran: @renzon we have to write to Guido to change that and implement do while :)

abalakh: In the meantime - we have a separate issue to decide which `retry` implementation is the best for us, this place is yet another one which should be replaced in scope of that issue. Let's not block this PR which addresses different issue and uses preferred approach (3 votes for it and 1 against 🙂 )

PR #5314, opened 2017-09-22
Implemented API and UI test for host's inherited params (BZ1287223)

abalakh:
Depends on SatelliteQE/nailgun#446
https://bugzilla.redhat.com/show_bug.cgi?id=1287223
```python
 py.test -v tests/foreman/ui/test_host.py -k test_positive_create_with_inherited_params
cachedir: .cache
collected 29 items
2017-09-21 21:10:10 - conftest - DEBUG - Deselect of WONTFIX BZs is disabled in settings


tests/foreman/ui/test_host.py::HostTestCase::test_positive_create_with_inherited_params PASSED

============================= 28 tests deselected ==============================
================== 1 passed, 28 deselected in 101.17 seconds ===================
 py.test -v tests/foreman/api/test_host.py -k test_positive_create_with_inherited_params
cachedir: .cache
collected 75 items
2017-09-21 21:12:24 - conftest - DEBUG - Deselect of WONTFIX BZs is disabled in settings


tests/foreman/api/test_host.py::HostTestCase::test_positive_create_with_inherited_params PASSED

============================= 74 tests deselected ==============================
=================== 1 passed, 74 deselected in 15.64 seconds ===================
```

### DIFF
--- a/tests/foreman/api/test_host.py
+++ b/tests/foreman/api/test_host.py
@@ -266,6 +266,36 @@ class HostTestCase(APITestCase):
             hostgroup.content_view.id
         )
 
+    @tier2
+    def test_positive_create_with_inherited_params(self):
+        """Create a new Host in organization and location with parameters
+
+        :BZ: 1287223
+
+        :id: 5e17e968-7fe2-4e5b-90ca-ae66f4e5307a
+
+        :expectedresults: Host has inherited parameters from organization and
+            location
+
+        :CaseImportance: High
+        """
+        org = entities.Organization().create()
+        org_param = entities.Parameter(organization=org).create()
+        loc = entities.Location().create()
+        loc_param = entities.Parameter(location=loc).create()
+        host = entities.Host(
+            location=loc,
+            organization=org,
+        ).create()
+        self.assertEqual(
+            {org_param.name, loc_param.name},
+            {param['name'] for param in host.all_parameters}
+        )
+        self.assertEqual(
+            {org_param.value, loc_param.value},
+            {param['value'] for param in host.all_parameters}
+        )
+
     @run_only_on('sat')
     @tier1
     def test_positive_create_with_puppet_proxy(self):
--- a/tests/foreman/api/test_host.py
+++ b/tests/foreman/api/test_host.py
@@ -287,13 +287,11 @@ class HostTestCase(APITestCase):
             location=loc,
             organization=org,
         ).create()
+        self.assertEqual(len(host.all_parameters), 2)
         self.assertEqual(
-            {org_param.name, loc_param.name},
-            {param['name'] for param in host.all_parameters}
-        )
-        self.assertEqual(
-            {org_param.value, loc_param.value},
-            {param['value'] for param in host.all_parameters}
+            {(org_param.name, org_param.value),
+             (loc_param.name, loc_param.value)},
+            {(param['name'], param['value']) for param in host.all_parameters}
         )
 
     @run_only_on('sat')
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -585,16 +585,11 @@ class HostTestCase(UITestCase):
             )
             actual_params = self.hosts.fetch_global_parameters(
                 host.name, host.domain.name)
-            actual_names = set()
-            actual_values = set()
-            for name, value in actual_params:
-                actual_names.add(name)
-                actual_values.add(value)
-            self.assertEqual({org_param_name, loc_param_name}, actual_names)
+            actual_names, actual_values = zip(*actual_params)
             self.assertEqual(
-                {org_param_value, loc_param_value},
-                actual_values
-            )
+                {org_param_name, loc_param_name}, set(actual_names))
+            self.assertEqual(
+                {org_param_value, loc_param_value}, set(actual_values))
 
     @run_only_on('sat')
     @tier3
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -54,6 +54,8 @@ from robottelo.test import UITestCase
 from robottelo.ui.factory import (
     make_host,
     make_hostgroup,
+    make_loc,
+    make_org,
     set_context,
 )
 from robottelo.ui.base import UINoSuchElementError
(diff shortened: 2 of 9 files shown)


tests/foreman/api/test_host.py
```diff
@@ -266,6 +266,36 @@ def test_positive_create_inherit_lce_cv(self):
...
+        ).create()
+        self.assertEqual(
+            {org_param.name, loc_param.name},
+            {param['name'] for param in host.all_parameters}
+        )
+        self.assertEqual(
+            {org_param.value, loc_param.value},
+            {param['value'] for param in host.all_parameters}
+        )
+
```
ldjebran: think more accurate assert
```python
all_parameters = {param['name']: param['value']  param in host.all_parameters}
org_loc_params = {org_param.name: org_param.value, loc_param.name: loc_param.value}
self.assertEqual(all_parameters, org_loc_params)
```
what do you think?

tests/foreman/ui/test_host.py
```diff
@@ -507,6 +509,84 @@ def test_positive_create(self):
             )
             self.assertIsNotNone(search)
 
+    @tier3
+    def test_positive_create_with_inherited_params(self):
```
ldjebran: why we are covering the UI , should we cover all bugs with UI too if they CLI or API (this should lead to too much efforts to cover bugs)?

ldjebran: the bug description was in CLI, perhaps it was natural to add a CLI test instead of UI?

tests/foreman/ui/test_host.py
```diff
@@ -507,6 +509,84 @@ def test_positive_create(self):
...
+        loc_param_name = gen_string('alphanumeric')
+        org_param_value = gen_string('alphanumeric')
+        loc_param_value = gen_string('alphanumeric')
+        with Session(self) as session:
+            make_org(
+                session,
+                org_name=org_name,
+                params=[(org_param_name, org_param_value)],
+            )
+            session.nav.go_to_select_org(org_name)
```
ldjebran: why not use set_content from factory?

ldjebran: why not use set_content from factory?

tests/foreman/ui/test_host.py
```diff
@@ -507,6 +509,84 @@ def test_positive_create(self):
...
+                ],
+            )
+            actual_params = self.hosts.fetch_global_parameters(
+                host.name, host.domain.name)
+            actual_names, actual_values = zip(*actual_params)
+            self.assertEqual(
+                {org_param_name, loc_param_name}, set(actual_names))
+            self.assertEqual(
+                {org_param_value, loc_param_value}, set(actual_values))
+
```
ldjebran: same comment as for API asserts

tests/foreman/api/test_host.py
```diff
@@ -266,6 +266,36 @@ def test_positive_create_inherit_lce_cv(self):
...
+            location=loc,
+            organization=org,
+        ).create()
+        self.assertEqual(
+            {org_param.name, loc_param.name},
+            {param['name'] for param in host.all_parameters}
+        )
+        self.assertEqual(
+            {org_param.value, loc_param.value},
+            {param['value'] for param in host.all_parameters}
```
ldjebran: all_params is a list of dicts or dict?

tests/foreman/api/test_host.py
ldjebran: if all_params is a dict
```python
org_loc_params = {org_param.name: org_param.value, loc_param.name: loc_param.value}
self.assertEqual(host.all_parameters, org_loc_params)

```

abalakh: I was thinking about that and decided to keep things simple. But since you've noticed that too, i should probably update assertions, thanks for suggestion 🙂

tests/foreman/ui/test_host.py
abalakh: There's no way to verify changes in CLI as of now (even verification steps are using API), that's why i dropped it. If i remember it right - functionally is fixed and works ok, but there is no command to fetch inherited params or smth, so the tests won't be able to use assertions and will be completely useless.
Regarding UI - i had difficulties with automating API part at first and was not sure whether it's possible at all. On the contrary, UI was pretty straight forward so i started with it first. UI part is not required for BZ coverage, but since it's already automated - why not?) Consider it as a bonus.

tests/foreman/ui/test_host.py
abalakh: Hm... Same question to you - why `set_context` instead of `go_to_select_org`? 🙂 `git grep` shows  `go_to_select_org` is much more preferable way; they both are doing pretty similar things though and `set_context` even has it's own implementation of `force_context` (imo that should be refactored and `set_context` should reuse go_to helpers' force arg). Any real benefits from using it? For me it like 2 helpers doing pretty much the same.

abalakh: Maybe originally go_to helpers had no force_context arg and that's why `set_context` was introduced, but right now when both helpers support forcing context - i can't see much need in `set_context`, it's a duplication.

tests/foreman/api/test_host.py
abalakh: List of dicts

tests/foreman/api/test_host.py
abalakh: Hm.. If not just param name but param value is duplicated too, sets won't show us that. We should probably check the `len()` as well, since list won't work for us as order can be easily damaged.

abalakh:
```python
 py.test -v tests/foreman/{api,ui}/test_host.py -k test_positive_create_with_inherited_params
cachedir: .cache
collected 111 items
2017-10-03 13:04:09 - conftest - DEBUG - Deselect of WONTFIX BZs is disabled in settings


tests/foreman/api/test_host.py::HostTestCase::test_positive_create_with_inherited_params PASSED
tests/foreman/ui/test_host.py::HostTestCase::test_positive_create_with_inherited_params PASSED

============================= 109 tests deselected =============================
================== 2 passed, 109 deselected in 163.27 seconds ==================
```

tests/foreman/ui/test_host.py
ldjebran: you omitted to say that set_content can also set the location, so we have one function that set org and loc when indicated, I was thinking about the uniformity of the code not more not less.

ldjebran: this is not that important, as already many tests are using go_to_select_org.
But we have to make some standard way to set that context and prefer the set context one because it verify if org or loc are already selected before trying to select, and stick well with the concept of context org,loc that we are working in.

PR #5325, opened 2017-09-27
vm.create() - tweak connection_timeout

rplevka:
this was falling back to 10s since the setting is not being set in `robottelo.properties` which was sometimes just too little for our overloaded ginger.

- This should fix many failing tests, using ginger and claimed as `infra_error`


omaciel:
@ldjebran @abalakh @oshtaier can you guys review it? :)

omaciel:
Thank you @abalakh and @oshtaier  💯

PR #5334, opened 2017-10-03
Fix fix_uuids script to actually find duplicates

elyezer:
The -n option to sort says that it should compare a numeric value, but
UUID is an alphanumeric value. Fix that so finding duplicates will work
as expected.

This will break travis so someone needs to fix the duplicates before getting this merged. Feel free to add another commit on top of this PR's commit to fix that issue.

### DIFF
--- a/scripts/fix_uuids.sh
+++ b/scripts/fix_uuids.sh
@@ -49,12 +49,12 @@ done
 # sort the output and filters only the duplicated
 # then looks for existence of ":id:" in final output
 # NOTE: can't print the line number -n here because of uniq -d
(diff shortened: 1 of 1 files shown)


rochacbruno:
@elyezer travis CI is failing because with the new code lots of dups were found.. dont you want to send the fix for it in the same PR? How will this ID change impact polarion?

elyezer:
I sent the PR on this repositories branch so anyone can work on getting the IDs fixed. I have a lot of other tasks to do so o won't be able to handle the deduplication. I sent the PR with the command that will be run to validate camayoc's tests.

On polarion it will create a new tests when new IDs are added. That shouldn't cause any issue.

rochacbruno:
ok! I'll ACK and once merged I send another PR for de-duplication

sghai:
ACK

PR #5337, opened 2017-10-03
Fix uuid-fix script and also replacing existing duplicates

rochacbruno:
already checked with @elyezer and Polarion will be god with this
change by creating a new test for the duplicated IDs.


rochacbruno:
This PR is blocking all other PRs to pass in Travis CI

PR #5338, opened 2017-10-03
Fix flake8 update do not accept empty argument anymore

rochacbruno:
in endless loop (maximum recursion) if called without arguments
sometimes the `git diff` returns empty result, so I added `robottelo/__init__.py`
just to address that case. (it is an empty file no overhead added).

PS: this `gitflake8` is used by `make can-i-push?` target


rochacbruno:
Tavis CI is depending on #5337

PR #5388, opened 2017-10-19
[6.3] Reinstall capsule on the same node after remove (BZ: 1327442)

ldjebran:
cover https://bugzilla.redhat.com/show_bug.cgi?id=1327442
```console
(sat-6.3.0) dlezz@elysion:~/projects/robottelo-fork$ pytest -v tests/foreman/cli/test_capsule_installer.py::CapsuleInstallerTestCase::test_positive_reinstall_on_same_node_after_remove
platform linux2 -- Python 2.7.13, pytest-3.1.3, py-1.4.34, pluggy-0.4.0 -- /home/dlezz/.pyenv/versions/sat-6.3.0/bin/python2.7
cachedir: .cache
plugins: xdist-1.20.0, services-1.2.1, mock-1.6.2, forked-0.2, cov-2.5.1
collected 1 item 
2017-10-19 14:34:08 - conftest - DEBUG - Deselect of WONTFIX BZs is disabled in settings


tests/foreman/cli/test_capsule_installer.py::CapsuleInstallerTestCase::test_positive_reinstall_on_same_node_after_remove <- robottelo/decorators/__init__.py PASSED

============================================= 1 passed in 5900.26 seconds ==============================================
```

### DIFF
--- a/robottelo/cli/factory.py
+++ b/robottelo/cli/factory.py
@@ -3226,7 +3226,10 @@ def publish_puppet_module(puppet_modules, repo_url, organization_id=None):
     return ContentView.info({u'id': cv['id']})
 
 
-def _extract_capsule_satellite_installer_command(text):
+def extract_capsule_satellite_installer_command(text):
+    """Extract satellite installer command from capsule-certs-generate command
+    output
+    """
     cmd_start_with = 'satellite-installer'
     cmd_lines = []
     if text:
@@ -3502,7 +3505,7 @@ def setup_capsule_virtual_machine(capsule_vm, org_id=None, lce_id=None,
             u'was unable to generate certificate\n{}'.format(result.stderr))
 
     # retrieve the installer command from the result output
-    satellite_installer_cmd = _extract_capsule_satellite_installer_command(
+    satellite_installer_cmd = extract_capsule_satellite_installer_command(
         result.stdout
     )
     # copy the certificate to capsule vm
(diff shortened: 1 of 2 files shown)


pondrejk:
Or differently, we could assert that configs were removed after katello-remove finishes, this way if something slips, we wouldn't have to wait for failing capsule-installer to find out, does it make sense?

ldjebran:
the bz say that the install failed because of existing configs and certs, the BZ complain about install failure after remore, 
so intall after remore is sufficient

@abalakh @pondrejk this is not a test coverage of katello-remove   
if it was the case we have to have a list of certs intall dirs all services checked start/stop/existence  checked, but this is not the case

ldjebran:
@abalakh @pondrejk even if we check that all the configs where removed, we have no guaranty that the capsule will  reinstall successfully , so the only way to ensure that, is to reinstall it for real

abalakh:
@ldjebran you probably didn't get me right. I'm not asking about removing reinstall part of the test (actually i'm against it, that's a nice coverage which we hadn't before), but proposing to add one extra assertion. If configs were not removed then reinstall won't succeed so there's no point in proceeding with scenario. Consider it smth like our `self.assertEqual(result.return_code, 0)` asserts - which aren't the things we need to verify in our tests but are saving us time since there's no point in proceeding when we know already that one of key commands has failed.

PR #5400, opened 2017-10-23
Fixed flake8-3.5.0 compatibility

abalakh:
Recently flake updated to 3.5.0 and caught a new nitpick:
```python
> pip list installed | grep flake
flake8 (3.3.0)
pyflakes (1.5.0)
> flake8 .
>
```
```python
> pip list installed | grep flake
flake8 (3.5.0)
pyflakes (1.6.0)
> flake8 .
./robottelo/vm_capsule.py:245:9: E722 do not use bare except'
```

Unblocking travis for recent PRs with least changes to framework logic

### DIFF
--- a/robottelo/vm_capsule.py
+++ b/robottelo/vm_capsule.py
@@ -242,7 +242,7 @@ class CapsuleVirtualMachine(VirtualMachine):
         super(CapsuleVirtualMachine, self).create()
         try:
             self._setup_capsule()
-        except:
+        except Exception:
             # handle exception as VirtualMachine has no exception handling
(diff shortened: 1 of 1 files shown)


abalakh:
@ldjebran  did you have really good reasons to catch everything, like `SystemExit` and `KeyboardInterrupt`? If that's the case, `except BaseException` would be a better option, but its use is strongly discouraged, as it'll be hard to exit any capsule test 🙂

pondrejk:
I'd merge but I see there is a pending question above, feel free to merge when resolved

Issue #5324, opened 2017-09-27
CLI - hammer user list now contains more fields


rplevka:
The assertions for at least these tests need to be updated:

`test_positive_list_email`
`test_positive_list_firstname`
`test_positive_list_surname`
`test_positive_list_username`
```
Id,Login,Name,Email
87,6nh1Ds,𢱲𨷿𠹊𥟀μߢ𦬇앣𥣭𓃦딿𡠳㳘铪𫗾ΰϛ韃𤒮戒ᚧ𤣣쀓輊충쯅䢿𢇾됢嚘𤿗럈𥺻𓉜Å痍𥻐 sees0k0Mpy,6nh1Ds@example.com
```
```
Id,Login,Name,Email,Admin,Last login,Authorized by
87,K5WIB7,872555310951844493906825832510985336 xf9eMDdTHc,K5WIB7@example.com,false,"",Internal
```

ntkathole:
Closing as PR merged!

rplevka:
there's now even more:
```
Id,Login,Name,Email,Admin,Last login,Authorized by
73,13KycK,Wa2RXh79BE 0rCxqWr3jw,íÔÙÊÏêÇØïÐ@somemail.com,false,"",Internal
```

let's not try to catch 'em all, and just assert on an existence and values of specific fields

ntkathole:
@rplevka could you please point me which tests failed?

abalakh:
@ntkathole failure happened in test `test_positive_create_with_email_utf8_latin` which wasn't updated in #5422 btw. I'd suggest to take a careful look at every single test in module and update all the outdated assertions.

rplevka:
thanks @abalakh, yes, that is the one.
I can see @ntkathole actually used the assertion exactly as i wanted to suggest, it's indeed just this test missed from that PR.
- since no other test fails, i guess this will be all. however, who knows :)

ntkathole:
@rplevka thanks for info. I will raise PR for this test and will confirm whether some other tests from cli.test_user fail due to same reason.

ntkathole:
All the test_user tests in CLI passed for snap 28 after fix.  Closing this issue!

PR #5422, opened 2017-10-31
Fixes CLI - hammer user list now contains more fields

ntkathole:
Issue #5324 

Test results

```
# pytest tests/foreman/cli/test_user.py -k test_positive_list
collected 51 items

tests/foreman/cli/test_user.py ....

========================================================== 47 tests deselected ==========================================================
=============================================== 4 passed, 47 deselected in 397.97 seconds ===========================================
```

### DIFF
--- a/tests/foreman/cli/test_user.py
+++ b/tests/foreman/cli/test_user.py
@@ -526,12 +526,10 @@ class UserTestCase(CLITestCase):
                 })
                 self.assertEqual(len(result), 1)
                 # make sure user is in list result
-                self.assertEqual({
-                    'email': user['email'],
-                    'id': user['id'],
-                    'login': user['login'],
-                    'name': user['name'],
-                }, result[0])
+                self.assertEqual(
+                    {user['id'], user['login']},
+                    {result[0]['id'], result[0]['login']}
+                )
 
     @tier1
     def test_positive_list_firstname(self):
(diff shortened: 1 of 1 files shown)


ldjebran:
Do not think that the purpose of this tests, to test the content of all listed fields, but perhaps to verify the only needed ones. 

If we need to verify that all fields are returned we have to create one test for that purpose.

@oshtaier @abalakh what do you think?

* We will face this tests to fail every time a field is added/removed

abalakh:
@ldjebran completely agree with you. Even inline comment says `# make sure user is in list result`.
We should only check result[0] is our user, and should not compare every possible field. I believe checking id and login should be enough.

ntkathole:
@abalakh and @ldjebran Thanks...I am also agree with your comments...will update the PR accordingly.

ntkathole:
@ldjebran and @abalakh Changed some logic during assert...I found this shortest solution...please suggest if this could be more better.Thanks :)

tests/foreman/cli/test_user.py
```diff
@@ -527,11 +527,10 @@ def test_positive_list_username(self):
                 self.assertEqual(len(result), 1)
                 # make sure user is in list result
                 self.assertEqual({
-                    'email': user['email'],
                     'id': user['id'],
                     'login': user['login'],
-                    'name': user['name'],
-                }, result[0])
+                }, {k: v for k, v in result[0].items()
+                    if k in ['id', 'login']})
```
abalakh: Why not simply
```python
self.assertEqual(
    {user['id'], user['login']},
    {result[0]['id'], result[0]['login']}
)
```
we're interested in values, not keys.

ntkathole:
@abalakh @ldjebran comment addressed!

PR #5423, opened 2017-11-01
Remove dead upgrade test code in Robotello

jyejare:
There is separate repository under SatelliteQE to handle all upgrade tests and hence we are removing it from here.


sghai:
APT

PR #5576, opened 2017-11-14
Update the test-foreman-sys to not just include sys path

kbidarkar:

PR #5577, opened 2017-11-14
Update the test-foreman-sys to not just include sys path

kbidarkar:


kbidarkar:
This is just the same as PR #5576 cherry-picked to sat6.2.z

PR #5580, opened 2017-11-15
[6.3] Automated API tests for settings

ntkathole:


pondrejk:
Hi @ntkathole could you please paste some test results?

tests/foreman/api/test_setting.py
```diff
@@ -31,23 +32,35 @@ def test_positive_update_login_page_footer_text(self):
...
-        :caseautomation: notautomated
         """
+        login_text_id = [ele.id for ele in entities.Setting().search(query={
+            "per_page": 200,
+            'search': 'name="login_text"'
+        })][0]
+        login = entities.Setting(id=login_text_id).read()
+        login.value = ""
+        login = login.update(['value'])
+        self.assertEqual(login.value, "")
```
pondrejk: Is there some default value for the footer text? Maybe you should revert the setting back to default at the end of the test so that you don't infuence your other current/future tests. Also I wonder if those tests could be run in multiple threads, if there's a possibility o race condition, pls add  @run_in_one_thread()

ntkathole:
@pondrejk Test results 

```
pytest tests/foreman/api/test_setting.py
collected 3 items

tests/foreman/api/test_setting.py ...

======================================================= 3 passed in 25.99 seconds ======================================================
```

tests/foreman/api/test_setting.py
ntkathole: As this is the message text field which is shown on login page and no other entity is dependant on this parameter, so  it won't influence any other current/future tests and yes, I agree with you for @run_in_one_thread(), there may be chances, will update.

san7ket:
Ack

PR #5602, opened 2017-11-21
[6.3] Added API tests for discovery hostname setting

ntkathole:
Test Results

```
pytest tests/foreman/api/test_setting.py

collected 8 items

tests/foreman/api/test_setting.py sss.....

================================================= 5 passed, 3 skipped ===========================================
```

### DIFF
--- a/tests/foreman/api/test_setting.py
+++ b/tests/foreman/api/test_setting.py
@@ -16,8 +16,9 @@
 :Upstream: No
 """
 from nailgun import entities
+from robottelo.cleanup import setting_cleanup
 from robottelo.datafactory import generate_strings_list, valid_data_list
-from robottelo.decorators import run_in_one_thread, tier1, upgrade
+from robottelo.decorators import run_in_one_thread, stubbed, tier1, upgrade
 from robottelo.test import APITestCase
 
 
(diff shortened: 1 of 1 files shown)


tests/foreman/api/test_setting.py
```diff
@@ -17,13 +17,18 @@
...
 from robottelo.test import APITestCase
 
 
 class SettingTestCase(APITestCase):
     """Implements tests for Settings for API"""
 
+    def setUp(self):
+        """Initialize test variables"""
+        super(SettingTestCase, self).setUp()
+        self.original_value = None
```
renzon: Why do you need create a None var on setup?

ntkathole: @renzon original_value is used for reverting the value of parameter so that it should not affect other tests. It will be none in setup and before test starts, original value will be set.

ntkathole:
@renzon could you please ACK if it looks fine to you?

tests/foreman/api/test_setting.py
```diff
@@ -83,3 +88,81 @@ def test_positive_update_login_page_footer_text_with_long_string(self):
...
+        """
+        hostname_prefix_id = [ele.id for ele in entities.Setting().search(
+            query={"per_page": 200, 'search': 'name="discovery_prefix"'})][0]
+        prefix = entities.Setting(id=hostname_prefix_id).read()
+        self.original_value = prefix.name
+        prefix.value = ""
+        prefix = prefix.update(['value'])
+        self.assertEqual(prefix.value, "")
+        prefix.value = self.original_value
+        prefix.update(['value'])
```
pondrejk: I'm a bit confused by the last two lines, so we are setting the original_value to prefix *name* and after assert we are assigning it to prefix *value*. Also I'd use try - finally block to make sure the prefix is changed back also in case of unexpected failures e.g. when updating prefix

ntkathole: @pondrejk this implementation is same as we do teardown() in tests. Please see similar implementation that we do in UI tests https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/ui/test_setting.py#L147 .

ntkathole: @pondrejk good catch! updated the PR. Please revisit

pondrejk:
Just noting Nikhil did retest after making those changes

PR #5699, opened 2017-12-15
Make generic method for sending keys to browser.

odovzhenko:
Make generic method from #5683

```
(sat-6.3.0) odovz@bueno:~/projects/robottelo$ py.test -v tests/foreman/ui/test_host.py::HostTestCase::test_positive_check_permissions_affect_create_procedure
=========================================================================================== test session starts ============================================================================================
platform linux2 -- Python 2.7.12, pytest-3.2.3, py-1.4.34, pluggy-0.4.0 -- /home/odovz/venv/sat-6.3.0/bin/python2.7
cachedir: .cache
rootdir: /home/odovz/projects/robottelo, inifile:
collected 1 item                                                                                                                                                                                            
2017-12-15 16:10:38 - conftest - DEBUG - BZ deselect is disabled in settings


tests/foreman/ui/test_host.py::HostTestCase::test_positive_check_permissions_affect_create_procedure PASSED

======================================================================================== 1 passed in 158.55 seconds ========================================================================================
```

### DIFF
--- a/robottelo/ui/base.py
+++ b/robottelo/ui/base.py
@@ -886,13 +886,8 @@ class Base(object):
             including special, like ``control+shift+q``, single special or
             simple key, e.g. ``escape``, ``q``.
         """
-        if '+' in keys:
-            keys_to_send = ''.join(
-                getattr(Keys, key.upper()) if key.upper() in dir(Keys) else
-                key for key in keys.split('+'))
-        elif keys.upper() in dir(Keys):
-            keys_to_send = getattr(Keys, keys.upper())
-        else:
-            keys_to_send = keys
+        keys_to_send = ''.join(
+            getattr(Keys, key.upper()) if key.upper() in dir(Keys) else
+            key for key in keys.split('+'))
         ActionChains(self.browser).send_keys(keys_to_send).perform()
         self.wait_for_ajax()
--- a/robottelo/ui/base.py
+++ b/robottelo/ui/base.py
@@ -15,6 +15,7 @@ from selenium.common.exceptions import NoSuchElementException
 from selenium.common.exceptions import TimeoutException
 from selenium.common.exceptions import WebDriverException
 from selenium.webdriver.common.action_chains import ActionChains
+from selenium.webdriver.common.keys import Keys
 from selenium.webdriver.support import expected_conditions
 from selenium.webdriver.support.select import Select
 from selenium.webdriver.support.ui import WebDriverWait
@@ -878,10 +879,20 @@ class Base(object):
         ]
         return cell_values
 
-    def perform_action_send_key_to_browser(self, key):
+    def perform_action_send_keys_to_browser(self, keys):
         """Send some key(s) to browser.
 
-        :param key: Key(s) to send to browser.
+        :param keys: Key(s) to send to browser. Keys may be several keys
+            including special, like ``control+shift+q``, single special or
+            simple key, e.g. ``escape``, ``q``.
         """
-        ActionChains(self.browser).send_keys(key).perform()
+        if '+' in keys:
+            keys_to_send = ''.join(
+                getattr(Keys, key.upper()) if key.upper() in dir(Keys) else
+                key for key in keys.split('+'))
+        elif keys.upper() in dir(Keys):
+            keys_to_send = getattr(Keys, keys.upper())
+        else:
+            keys_to_send = keys
+        ActionChains(self.browser).send_keys(keys_to_send).perform()
         self.wait_for_ajax()
--- a/robottelo/ui/base.py
+++ b/robottelo/ui/base.py
@@ -15,7 +15,6 @@ from selenium.common.exceptions import NoSuchElementException
 from selenium.common.exceptions import TimeoutException
 from selenium.common.exceptions import WebDriverException
 from selenium.webdriver.common.action_chains import ActionChains
-from selenium.webdriver.common.keys import Keys
 from selenium.webdriver.support import expected_conditions
 from selenium.webdriver.support.select import Select
 from selenium.webdriver.support.ui import WebDriverWait
@@ -879,7 +878,10 @@ class Base(object):
         ]
         return cell_values
 
-    def perform_action_send_escape_key(self):
-        """Send escape key to browser"""
-        ActionChains(self.browser).send_keys(Keys.ESCAPE).perform()
+    def perform_action_send_key_to_browser(self, key):
+        """Send some key(s) to browser.
+
+        :param key: Key(s) to send to browser.
+        """
+        ActionChains(self.browser).send_keys(key).perform()
         self.wait_for_ajax()
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -15,7 +15,6 @@
 
 :Upstream: No
 """
-from selenium.webdriver.common.keys import Keys
 import six
 import yaml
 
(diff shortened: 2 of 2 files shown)


oshtaier:
no-no. That solution is not good at all as you will need to put `from selenium.webdriver.common.keys import Keys` in all modules where you need to use the method. It is not acceptible

tests/foreman/ui/test_host.py
```diff
@@ -976,7 +977,7 @@ def test_positive_check_permissions_affect_create_procedure(self):
                 with self.assertRaises(UINoSuchElementError):
                     self.hosts.assign_value(
                         entity['locator'], entity['unexpected_entity'].name)
-                self.hosts.perform_action_send_escape_key()
+                self.hosts.perform_action_send_key_to_browser(Keys.ESCAPE)
```
abalakh: Let's discuss. I can't really see a value of `self.hosts.perform_action_send_key_to_browser` instead of `ActionChains(self.browser).send_keys(Keys.ESCAPE).perform()`.
Such "helper" isn't really helpful - still takes the same amount of lines and needs extra imports. 
Maybe we should think of some other implementation - let's say passing string equivalent of desired key (e.g. 'esc' instead `Keys.ESCAPE`), that way we'll at least get rid of extra import. But even with such improvement i'm not really sure it's worth separate helper to save 1 line.

ldjebran: @abalakh and create a constant file of keys that has been already done by sellenium???

abalakh: That's why i'm asking for suggestions 🙂 I'm not saying we should go for it, that was one of ideas off the top of my head.
Yep, we could create a dictionary with all the keys, or, for example, take the input string and grab the webdriver's `Key` class attribute by string (getattr), this way we don't need a duplicated dict. Or we could go with something totally different, feel free to suggest.

odovzhenko: We might use:
`self.hosts.perform_action_send_key_to_browser("escape")`
`ActionChains(self.browser).send_keys(getattr(Keys, key.upper())).perform()`

abalakh: @oshtaier how's this idea for you?

oshtaier: yeah, it is better

oshtaier: can it parse, something like ctrl+shift+q?

ldjebran: propose to have the two function, and the the escape one a shot cut of the more general send_keys
EG:
```python
    def perform_action_send_keys(self, **keys):
        """Perform send keys to browser"""
        ActionChains(self.browser).send_keys(**keys).perform()

    def perform_action_send_escape_key(self, wait_for_ajax=True):
        """Send escape key to browser"""
        self.perform_action_send_keys(Keys.ESCAPE)
        if wait_for_ajax:
            self.wait_for_ajax()
```

odovzhenko: It can't. We can use:
```
def perform_action_send_key_to_browser(keys):
    if '+' in keys:
        keys_to_send = ''.join(
            getattr(Keys, key.upper()) if key.upper() in dir(Keys) else key for
            key in keys.split('+'))
    elif keys.upper() in dir(Keys):
        keys_to_send = getattr(Keys, keys.upper())
    else:
        keys_to_send = keys
    return keys_to_send  

perform_action_send_key_to_browser('control+shift+q')
u'\ue009\ue008q'
perform_action_send_key_to_browser('escape')
u'\ue00c'
perform_action_send_key_to_browser('q')
'q'
```

to handle this.

oshtaier: okay, proposed solution
```
self.hosts.perform_action_send_key_to_browser("escape")
def perform_action_send_key_to_browser
    ActionChains(self.browser).send_keys(getattr(Keys, key.upper())).perform()
    wait
```
looks good to me
```
@ldjebran we can't do like you suggested as escape is not something super special to have explicit method

**Comment by odovzhenko** [comment]
Propose to use:
```
def perform_action_send_key_to_browser(keys):
    if '+' in keys:
        keys_to_send = ''.join(
            getattr(Keys, key.upper()) if key.upper() in dir(Keys) else key for
            key in keys.split('+'))
    elif keys.upper() in dir(Keys):
        keys_to_send = getattr(Keys, keys.upper())
    else:
        keys_to_send = keys
    return keys_to_send  

perform_action_send_key_to_browser('control+shift+q')
u'\ue009\ue008q'
perform_action_send_key_to_browser('escape')
u'\ue00c'
perform_action_send_key_to_browser('q')
'q'
perform_action_send_key_to_browser('F+1')
'F1'
perform_action_send_key_to_browser('F1')
u'\ue031'
```
With this approach test is working:
```
(sat-6.3.0) odovz@bueno:~/projects/robottelo$ py.test -v tests/foreman/ui/test_host.py::HostTestCase::test_positive_check_permissions_affect_create_procedure
=========================================================================================== test session starts ============================================================================================
platform linux2 -- Python 2.7.12, pytest-3.2.3, py-1.4.34, pluggy-0.4.0 -- /home/odovz/venv/sat-6.3.0/bin/python2.7
cachedir: .cache
rootdir: /home/odovz/projects/robottelo, inifile:
collected 1 item                                                                                                                                                                                            
2017-12-19 11:59:12 - conftest - DEBUG - BZ deselect is disabled in settings


tests/foreman/ui/test_host.py::HostTestCase::test_positive_check_permissions_affect_create_procedure PASSED

======================================================================================== 1 passed in 138.14 seconds ========================================================================================
```

File: robottelo/ui/base.py [hunk 2]
```diff
@@ -878,10 +879,20 @@ def sort_table_by_column(self, column_name):
...
-        :param key: Key(s) to send to browser.
+        :param keys: Key(s) to send to browser. Keys may be several keys
+            including special, like ``control+shift+q``, single special or
+            simple key, e.g. ``escape``, ``q``.
         """
-        ActionChains(self.browser).send_keys(key).perform()
+        if '+' in keys:
+            keys_to_send = ''.join(
+                getattr(Keys, key.upper()) if key.upper() in dir(Keys) else
+                key for key in keys.split('+'))
```
abalakh: Will this implementation actually handle `control+shift+q` or it will rather press `control`, then `shift`, then `q` one by one independently?

robottelo/ui/base.py
```diff
@@ -879,7 +879,20 @@ def sort_table_by_column(self, column_name):
...
+
+        :param keys: Key(s) to send to browser. Keys may be several keys
+            including special, like ``control+shift+q``, single special or
+            simple key, e.g. ``escape``, ``q``.
+        """
+        if '+' in keys:
+            keys_to_send = ''.join(
+                getattr(Keys, key.upper()) if key.upper() in dir(Keys) else
+                key for key in keys.split('+'))
+        elif keys.upper() in dir(Keys):
```
abalakh: btw that's kinda weird check. Just use `keys.split('+')` and you'll receive a list of keys which you can process in a single loop, no need in extra elif/else with the same implementation inside for a single key.

robottelo/ui/base.py
odovzhenko: It will press control+shift+q simultaneously.

odovzhenko:
Comment addressed.
```
(sat-6.3.0) odovz@bueno:~/projects/robottelo$ py.test -v tests/foreman/ui/test_host.py::HostTestCase::test_positive_check_permissions_affect_create_procedure
=========================================================================================== test session starts ============================================================================================
platform linux2 -- Python 2.7.12, pytest-3.2.3, py-1.4.34, pluggy-0.4.0 -- /home/odovz/venv/sat-6.3.0/bin/python2.7
cachedir: .cache
rootdir: /home/odovz/projects/robottelo, inifile:
collected 1 item                                                                                                                                                                                            
2017-12-19 14:38:29 - conftest - DEBUG - BZ deselect is disabled in settings


tests/foreman/ui/test_host.py::HostTestCase::test_positive_check_permissions_affect_create_procedure PASSED

======================================================================================== 1 passed in 121.73 seconds ========================================================================================
```

odovzhenko:
screenshot from 2017-12-19 16-04-24

`self.hosts.perform_action_send_keys_to_browser("shift+r+r")`

screenshot from 2017-12-19 16-04-40

PR #5729, opened 2017-12-21
[6.3] API Add bulk host package install workaround for BZ 1528275

ldjebran:
These test has been influenced by bug 1374669 that has been marked as duplicate of bug 1108106 that is CLOSED ERRATA
created a new clone  https://bugzilla.redhat.com/show_bug.cgi?id=1528275 of the bug 1108106 as have the same behavior.
The last automation was not affected by that bug as the bug was not marked as duplicate and had status NEW.
For the moment we have a bug in robozilla that follow the clone status instead of the current one
Hope on @abalakh fix :)
```
(sat-6.3) dlezz@dlezz:~/projects/robottelo-fork$ pytest tests/foreman/api/test_errata.py -v -k "test_positive_install_in_host or test_positive_install_in_hc or test_positive_install_multiple_in_host"
=================================================================================================== test session starts ====================================================================================================
platform linux2 -- Python 2.7.14, pytest-3.2.3, py-1.5.2, pluggy-0.4.0 -- /home/dlezz/.pyenv/versions/2.7.14/envs/sat-6.3/bin/python2.7
cachedir: .cache
shared_function enabled - ON - scope: ffff - storage: file
rootdir: /home/dlezz/projects/robottelo-fork, inifile:
plugins: xdist-1.20.1, services-1.2.1, mock-1.6.3, forked-0.2, cov-2.5.1, fauxfactory-1.1.1
collected 14 items                                                                                                                                                                                                          
2017-12-21 14:40:58 - conftest - DEBUG - Collected 14 test cases


tests/foreman/api/test_errata.py::ErrataTestCase::test_positive_install_in_hc PASSED
tests/foreman/api/test_errata.py::ErrataTestCase::test_positive_install_in_host PASSED
tests/foreman/api/test_errata.py::ErrataTestCase::test_positive_install_multiple_in_host PASSED

=================================================================================================== 11 tests deselected ====================================================================================================
======================================================================================== 3 passed, 11 deselected in 1104.18 seconds ========================================================================================
```

PR #5738, opened 2017-12-26
[6.3] add bulk install packages

ldjebran:
This test was separated from PR https://github.com/SatelliteQE/robottelo/pull/5729

The tested is not skipped and fail with :
qe_test_coverage: https://bugzilla.redhat.com/show_bug.cgi?id=1528275
```
E   AssertionError: Package walrus-0.71-1.noarch was not installed on host up0ladwtfoydzjge.lab.eng.rdu2.redhat.com
```
Put do not merge waiting for the bug to be flagged.

### DIFF
--- a/tests/foreman/api/test_errata.py
+++ b/tests/foreman/api/test_errata.py
@@ -28,6 +28,7 @@ from robottelo.constants import (
     DISTRO_RHEL6,
     DISTRO_RHEL7,
     FAKE_1_CUSTOM_PACKAGE,
+    FAKE_1_CUSTOM_PACKAGE_NAME,
     FAKE_2_CUSTOM_PACKAGE,
     FAKE_2_ERRATA_ID,
     FAKE_3_ERRATA_ID,
@@ -45,7 +46,6 @@ from robottelo.constants import (
 )
 from robottelo.decorators import (
     bz_bug_is_open,
-    skip_if_bug_open,
     run_in_one_thread,
     skip_if_not_set,
     stubbed,
@@ -96,7 +96,8 @@ class ErrataTestCase(APITestCase):
             'activationkey-id': cls.activation_key.id,
         })
 
-    def _install_package(self, clients, host_ids, package_name, via_ssh=True):
+    def _install_package(self, clients, host_ids, package_name, via_ssh=True,
+                         rpm_package_name=None):
         """Install package via SSH CLI if via_ssh is True, otherwise
         install via http api: PUT /api/v2/hosts/bulk/install_content
         """
@@ -113,7 +114,7 @@ class ErrataTestCase(APITestCase):
                 'content_type': 'package',
                 'content': [package_name],
             })
-            self._validate_package_installed(clients, package_name)
+            self._validate_package_installed(clients, rpm_package_name)
 
     def _validate_package_installed(self, hosts, package_name,
                                     expected_installed=True, timeout=120):
@@ -174,7 +175,6 @@ class ErrataTestCase(APITestCase):
                 )
             )
 
-    @skip_if_bug_open('bugzilla', 1528275)
     @upgrade
     @tier3
     def test_positive_bulk_install_package(self):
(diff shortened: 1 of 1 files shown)


ldjebran:
This was closed by error, and restored

ldjebran:
Seems Initially the function was wrongly implemented after deep investigations using client log output fixed
```bash
(sat-6.3-py3) dlezz@dlezz:~/projects/robottelo-fork$ pytest -v tests/foreman/api/test_errata.py::ErrataTestCase::test_positive_bulk_install_package
===================================================================================================== test session starts =====================================================================================================
platform linux -- Python 3.6.3, pytest-3.4.0, py-1.5.2, pluggy-0.6.0 -- /home/dlezz/.pyenv/versions/sat-6.3-py3/bin/python
cachedir: .pytest_cache
plugins: wait-for-1.0.9, xdist-1.20.1, services-1.2.1, mock-1.6.3, forked-0.2, cov-2.5.1, fauxfactory-1.1.1
collected 1 item                                                                                                                                                                                                              
2018-05-02 14:34:01 - conftest - DEBUG - BZ deselect is disabled in settings


tests/foreman/api/test_errata.py::ErrataTestCase::test_positive_bulk_install_package PASSED                                                                                                                             [100%]

================================================================================================= 1 passed in 357.64 seconds ==================================================================================================
```

ldjebran:
@abalakh @renzon please revisit

tests/foreman/api/test_errata.py
```diff
@@ -112,7 +114,7 @@ def _install_package(self, clients, host_ids, package_name, via_ssh=True):
                 'content_type': 'package',
                 'content': [package_name],
             })
-            self._validate_package_installed(clients, package_name)
+            self._validate_package_installed(clients, rpm_package_name)
```
abalakh: You should probably use smth like `rpm_package_name if rpm_package_name else package_name` not to introduce regressions.

tests/foreman/api/test_errata.py
```diff
@@ -173,6 +175,38 @@ def _fetch_available_errata(self, host, expected_amount, timeout=120):
...
+            client.enable_repo(REPOS['rhst7']['id'])
+            client.install_katello_agent()
+            host_id = entities.Host().search(query={
+                'search': 'name={0}'.format(client.hostname)})[0].id
+            self._install_package(
+                [client],
+                [host_id],
+                FAKE_1_CUSTOM_PACKAGE_NAME,
+                via_ssh=False,
+                rpm_package_name=FAKE_2_CUSTOM_PACKAGE
```
abalakh: Is this extra kwarg really necessary? Can't you just pass 'walrus' as `package_name`? That will install latest version and ensure it's installed. Or what's the expected flow here?

tests/foreman/api/test_errata.py
ldjebran: in case of ssh install we need to use package_name, in case of REX  we need both

tests/foreman/api/test_errata.py
ldjebran: @abalakh we need to be sure it has install a package with version and arch, for REX bulk install  we pass "walrus" as package name but for verification we need full package name with version and arch "rpm_package_name" that is "walrus-5.21-1.noarch"

tests/foreman/api/test_errata.py
ldjebran: @abalakh adding that param we introduce no regression , as all other tests use ssh

abalakh: My point was - existing tests don't use `rpm_package_name` kwarg, so it'll be `None`. If you leave this check as it is - tests will ensure package `None` is installed. Or am i missing something?

tests/foreman/api/test_errata.py
ldjebran: @abalakh also note that there is no other test that use that part of the function eg: via_ssh=False, except the test implemented here.

Issue #5305, opened 2017-09-19
CLI: `hammer default` tests should use custom user to prevent leakages of the settings to other tests


rplevka:
there are multiple tests which cause this. The recent one that caused the problem was:

ContentViewFilterTestCase/test_positive_create_with_default_taxonomies
similar one (https://github.com/SatelliteQE/robottelo/issues/5071) got fixed by marking it to be ran in a single thread, however, if the defaults are user-account specific, we should use custom users for the tests. Otherwise it is not optimal and still can cause trouble if something goes wrong (the teardown for example) - quite unlikely, but still..

omaciel:
Assigning this to @pondrejk

PR #5774, opened 2018-01-10
Fixed OSTree Cli Tests to use isolated user

### LINKED ISSUE #5769: tests.foreman.cli.test_ostreebranch.OstreeBranchTestCase.test_positive_list_by_repo_id
Reason (thanks @rplevka ):

#5305

Stack Trace:

Error Message
AssertionError: 0 not greater than 0
Stacktrace
args = (<tests.foreman.cli.test_ostreebranch.OstreeBranchTestCase testMethod=test_positive_list_by_repo_id>,)
kwargs = {}, settings_project = 'sat'

    @wraps(func)
    def wrapper(*args, **kwargs):
        """Wrapper that will skip the test if the test method project does
                not match with the settings project.
                """
        # Validate project value
        if settings.project:
            settings_project = settings.project.lower()
        else:
            settings_project = 'sat'
    
        if project not in allowed_project_modes:
            raise ProjectModeError(
                '"{0}" is not a project mode. The allowed project modes '
                'are: {1}'.format(project, allowed_project_modes)
            )
    
        # If robottelo.properties not present or does not specify a project
        # use sat
        if settings_project not in allowed_project_modes:
            raise ProjectModeError(
                '"{0}" is not an acceptable "[robottelo] project" value '
                'in robottelo.properties file. The allowed project modes '
                'are: {1}'.format(settings_project, allowed_project_modes)
            )
    
        # Preconditions PASS.  Now skip the test if modes does not match
        if project != settings_project:
            raise unittest2.SkipTest(
                'Server runs in "{0}" mode and this test will run '
                'only on "{1}" mode.'.format(settings_project, project)
            )
        else:
>           return func(*args, **kwargs)

robottelo/decorators/__init__.py:268: 
tests/foreman/cli/test_ostreebranch.py:88: in test_positive_list_by_repo_id
    self.assertGreater(len(result), 0)
E   AssertionError: 0 not greater than 0

renzon:
close #5769

related to #5305

Test Results:

```console

$ pytest tests/foreman/cli/test_ostreebranch.py
=============================================== test session starts ===============================================
platform linux2 -- Python 2.7.14, pytest-3.3.2, py-1.5.2, pluggy-0.6.0
plugins: xdist-1.15.0, services-1.2.1, mock-1.6.3, cov-2.4.0
collected 6 items                                                                                                 
2018-01-10 15:55:58 - conftest - DEBUG - BZ deselect is disabled in settings


tests/foreman/cli/test_ostreebranch.py ......                                                               [100%]

=========================================== 6 passed in 178.03 seconds ============================================
```

### DIFF
--- a/robottelo/cli/factory.py
+++ b/robottelo/cli/factory.py
@@ -148,6 +148,19 @@ def create_object(cli_object, options, values):
     return result
 
 
+def _entity_with_credentials(credentials, cli_entity_cls):
+    """Create entity class using credentials. If credentials is None will
+    return cli_entity_cls itself
+
+    :param credentials: tuple (login, password)
+    :param cli_entity_cls: Cli Entity Class
+    :return: Cli Entity Class
+    """
+    if credentials is not None:
+        cli_entity_cls = cli_entity_cls.with_user(*credentials)
+    return cli_entity_cls
+
+
 @cacheable
 def make_activation_key(options=None):
     """
@@ -318,6 +331,36 @@ def make_content_view(options=None):
         --name NAME                   Name of the content view
         --organization ORGANIZATION_NAME  Organization name to search by
         --organization-id ORGANIZATION_ID Organization identifier
+        --organization-label ORGANIZATION_LABEL Organization label to
+        search by
+        --product PRODUCT_NAME          Product name to search by
+        --product-id PRODUCT_ID         product numeric identifier
+        --repositories REPOSITORY_NAMES Comma separated list of values.
+        --repository-ids REPOSITORY_IDS List of repository ids
+                                        Comma separated list of values.
+        -h, --help                    print help
+
+    """
+    return make_content_view_with_credentials(options)
+
+
+def make_content_view_with_credentials(options=None, credentials=None):
+    """
+    Usage::
+
+        hammer content-view create [OPTIONS]
+
+    Options::
+
+        --component-ids COMPONENT_IDS List of component content view
+        version ids for composite views
+                                      Comma separated list of values.
+        --composite                   Create a composite content view
+        --description DESCRIPTION     Description for the content view
+        --label LABEL                 Content view label
+        --name NAME                   Name of the content view
+        --organization ORGANIZATION_NAME  Organization name to search by
+        --organization-id ORGANIZATION_ID Organization identifier
         --organization-label ORGANIZATION_LABEL Organization label to search by
         --product PRODUCT_NAME          Product name to search by
         --product-id PRODUCT_ID         product numeric identifier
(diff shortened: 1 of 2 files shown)


renzon:
@rplevka based on your comments regarding cli user I have tried to create an user to all test on cli OS Tree. The only part is dependent on default user is the isolated user creation itself.

@abalakh I tried to find a way to make use of factory methods with different user but haven't found None. Do you know if there is one? Anyway, in case there is no way to that, I have marked you as reviewer so I can validate my approach ;)

robottelo/cli/factory.py
```diff
@@ -892,6 +901,10 @@ def make_registry(options=None):
 
 @cacheable
 def make_repository(options=None):
+    return make_product_with_credentials(options)
+
+
+def make_repository_with_credentials(options=None, credentials=None):
```
ldjebran: Think we can refactor the current function by adding credentials kwarg with default None value, without creating additional functions, what do you think?
```python
def make_repository(options=None, credentials=None):
    ...
```

robottelo/cli/factory.py
```diff
@@ -892,6 +901,10 @@ def make_registry(options=None):
 
 @cacheable
 def make_repository(options=None):
+    return make_product_with_credentials(options)
```
ldjebran: we can put a simple doc string like: """Create a repository"""

robottelo/cli/factory.py
renzon: That is what I tried at first. But that way @cacheable is designed prevent it since. Check it has options and cache as boolean. So it would be kind of hard doing that right now since a lot of test are using this function. But anyway this is an example of complexity that I mentioned when proposed to start a new CLI approach in parallel with current one ;)

robottelo/cli/factory.py
```diff
@@ -148,6 +148,12 @@ def create_object(cli_object, options, values):
     return result
 
 
+def _entity_with_credentials(credentials, cli_entity_cls):
+    if credentials is not None:
```
abalakh: Please add a docstring

robottelo/cli/factory.py
```diff
@@ -302,6 +308,10 @@ def make_container(options=None):
 
 @cacheable
 def make_content_view(options=None):
+    return make_content_view_with_credentials(options)
```
abalakh: And here

jhutar:
Hello @renzon. Do you think this addresses https://github.com/SatelliteQE/robottelo/issues/5305 ? I believe that the goal of 5305 was to run every CLI test that touches Hammer defaults under different user and to use users settings and not system wide settings.

PR #5795, opened 2018-01-19
[6.2.z] update rhel-server repo constants to keep pointing to the recent versions

rplevka:
`[6,7]Server` repos point always to the recent minor version and saves us from maintaining the constants after every release.

I also removed the obsoleted version entries since they are not used anymore

this is a concept backport of https://github.com/SatelliteQE/robottelo/pull/5796


lpramuk:
It was great collaboration. Thanks.

PR #5796, opened 2018-01-19
updated rhel-server repo to always point to the recent released RHEL version

rplevka:
to save us from maintenance duty and failed tests due to missed version bumps

`7Server` always points to the recent released minor RHEL7 version including errata

### DIFF
--- a/robottelo/constants.py
+++ b/robottelo/constants.py
@@ -250,8 +250,8 @@ REPOSET = {
 REPOS = {
     'rhel7': {
         'id': 'rhel-7-server-rpms',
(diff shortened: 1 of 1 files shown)


ldjebran:
@rplevka will suggest to run some tests, especially virt-who contentaccess and capsule tests that patch the VMs to enable yum to point to the correct repo  url, I quite 90% sure that all that tests will fail after this update.
Note: all that tests user that RHEL7 repo.

lpramuk:
@kbidarkar you assumption is not correct. I was mistaken too.

RHEL 7.4 is about 25G and RHEL 7Server approx. the same

PR #5803, opened 2018-01-25
[6.3] Fix API UI errata tests with force upload manifest

ldjebran:
```console

pytest -v tests/foreman/api/test_errata.py::ErrataTestCase::test_positive_get_applicable_for_host tests/foreman/api/test_errata.py::ErrataTestCase::test_positive_get_count_for_host tests/foreman/ui/test_errata.py::ErrataTestCase::test_positive_show_count_on_chost_details_page tests/foreman/ui/test_errata.py::ErrataTestCase::test_positive_show_count_on_chost_page
============================================ test session starts =============================================
platform linux2 -- Python 2.7.14, pytest-3.3.2, py-1.5.2, pluggy-0.6.0 -- /home/dlezz/.pyenv/versions/sat-6.3/bin/python
cachedir: .cache
plugins: xdist-1.20.1, services-1.2.1, mock-1.6.3, forked-0.2, cov-2.5.1, fauxfactory-1.1.1
collected 4 items                                                                                            
2018-01-25 17:15:49 - conftest - DEBUG - Collected 4 test cases


tests/foreman/api/test_errata.py::ErrataTestCase::test_positive_get_applicable_for_host PASSED         [ 25%]
tests/foreman/api/test_errata.py::ErrataTestCase::test_positive_get_count_for_host PASSED              [ 50%]
tests/foreman/ui/test_errata.py::ErrataTestCase::test_positive_show_count_on_chost_details_page PASSED [ 75%]
tests/foreman/ui/test_errata.py::ErrataTestCase::test_positive_show_count_on_chost_page PASSED         [100%]

============================================= 0 tests deselected =============================================
======================================== 4 passed in 3108.37 seconds =========================================
```

### DIFF
--- a/tests/foreman/api/test_errata.py
+++ b/tests/foreman/api/test_errata.py
@@ -522,7 +522,7 @@ class ErrataTestCase(APITestCase):
             'content-view-id': content_view.id,
             'lifecycle-environment-id': env.id,
             'activationkey-id': activation_key.id,
-        }, force_use_cdn=True)
+        }, force_manifest_upload=True)
         setup_org_for_a_custom_repo({
             'url': CUSTOM_REPO_URL,
             'organization-id': org.id,
@@ -602,7 +602,7 @@ class ErrataTestCase(APITestCase):
             'content-view-id': content_view.id,
             'lifecycle-environment-id': env.id,
             'activationkey-id': activation_key.id,
-        }, force_use_cdn=True)
+        }, force_manifest_upload=True)
         setup_org_for_a_custom_repo({
             'url': CUSTOM_REPO_URL,
             'organization-id': org.id,
--- a/tests/foreman/ui/test_errata.py
+++ b/tests/foreman/ui/test_errata.py
@@ -686,7 +686,7 @@ class ErrataTestCase(UITestCase):
             'content-view-id': content_view.id,
             'lifecycle-environment-id': env.id,
             'activationkey-id': activation_key.id,
-        }, force_use_cdn=True)
+        }, force_manifest_upload=True)
         setup_org_for_a_custom_repo({
             'url': CUSTOM_REPO_URL,
             'organization-id': org.id,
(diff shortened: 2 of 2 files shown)


tests/foreman/api/test_errata.py
```diff
@@ -522,7 +522,7 @@ def test_positive_get_count_for_host(self):
             'content-view-id': content_view.id,
             'lifecycle-environment-id': env.id,
             'activationkey-id': activation_key.id,
-        }, force_use_cdn=True)
+        }, force_manifest_upload=True)
```
abalakh: That's a risky change in these specific tests - custom repos may change their contents every week with each new snap, new errata may be added. Our tests rely on exact errata counts in their assertions, meaning if any new errata added - they will fail. Such logic was pretty much ok for 'stable' cdn repos but is risky for custom repos where new errata may be added at any moment. 2 possible actions here:
- Use `force_manifest_upload=True` but update assertions logic to smth like `assertGreaterEqual(errata_counts, 1)`
- Keep `force_use_cdn=True` and update enabling repos with `force=True`

abalakh: Btw if my memory doesn't betray me, at least some of these tests were not forcing cdn before, that `force_use_cdn=True` was introduced because of such failures 🙂

ldjebran: @abalakh, I am aware about the issue you mentioned. That repo is for installing katello-agent only (The same way we are using sat-tools 7) and the errata are for packages installed from other repos, we force upload manifest as needed by other RH repos enabled later, so there is no worries about that.

abalakh: Errata is just metadata for package versions, and since both repos are in the same org - if sattools repo has any new packages there will be new applicable errata for that host.
But i can see your point - most errata from sattools shouldn't be applicable for host as we don't really use anything except katello-agent, and katello-agent won't have applicable errata as we install the latest version, so probability of failure because of new errata introduced is quite low in these tests. Ok, consider my comment as a nitpick, not blocker.

PR #5829, opened 2018-03-01
Add a unique name using uuid for RUBY_COVERAGE_NAME as prefix for hammer cmds

kbidarkar:
a) Pre-fix RUBY_COVERAGE_NAME ENV to every hammer command being run.
b) Fix the `hammer ping` testcase, which fails with the changes made to hammer calls.
c) Fix the `Robottelo Tests`, which fails with the changes made to hammer calls.

### DIFF
--- a/robottelo/cli/base.py
+++ b/robottelo/cli/base.py
@@ -2,6 +2,7 @@
 """Generic base class for cli hammer commands."""
 import logging
 import re
+import uuid
 
 from robottelo import ssh
 from robottelo.cli import hammer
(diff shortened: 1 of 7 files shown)


robottelo/cli/base.py
```diff
@@ -284,15 +285,18 @@ def execute(cls, command, user=None, password=None, output_format=None,
...
-        cmd = u'LANG={0} {1} hammer -v {2} {3} {4} {5}'.format(
-            settings.locale,
-            u'time -p' if time_hammer else '',
-            u'-u {0}'.format(user) if user is not None
-            else u'--interactive no',
-            u'-p {0}'.format(password) if password is not None else '',
-            u'--output={0}'.format(output_format) if output_format else u'',
-            command,
-        )
+        cmd = (u'RUBY_COVERAGE_NAME={0} LANG={1}'
```
kbidarkar: This change would require updating to `mock`  of robottelo tests.

kbidarkar: This is now fixed.

PR #5902, opened 2018-03-29
[6.3.z] fix lce test to be py2-backward compatible

rplevka:
without the fix:
```python
In [23]: ceil(12/5)
Out[23]: 2.0
```
fixed:
```python
In [22]: int(ceil(12/float(5)))
Out[22]: 3
```


lpramuk:
No science here. Merging.

PR #5932, opened 2018-04-17
make use of six.iteritems

kbidarkar:
Signed-off-by: Kedar Bidarkar <kbidarka@redhat.com>

### DIFF
--- a/robottelo/config/base.py
+++ b/robottelo/config/base.py
@@ -16,6 +16,7 @@ from six.moves.configparser import (
     NoSectionError,
     ConfigParser
 )
+from six import iteritems
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


robottelo/config/base.py
```diff
@@ -644,7 +645,7 @@ def read(self, reader):
     def validate(self):
         """Validate rhev settings."""
         validation_errors = []
-        values = [v for k, v in vars(self).iteritems() if k != 'ca_cert']
+        values = [v for k, v in iteritems(vars(self)) if k != 'ca_cert']
```
abalakh: Maybe `import six` and then `six.iteritems(vars(self))` to avoid any confusion with python's iteritems?

kbidarkar: hmm, that module already uses few like `from six.  ` , so thought I would follow the same.

PR #5938, opened 2018-04-26
adding make test-foreman-upgrade

pondrejk:
so that upgrade-marked tests can be collected from robotello-ci

```
make test-foreman-upgrade 
python -m cProfile -o test-foreman-upgrade.pstats $(which py.test) -v --junit-xml=foreman-results.xml -m 'not stubbed' -n auto --boxed -m 'not stubbed and upgrade' tests/foreman/{api,cli,ui}
============================================= test session starts =============================================

...
```


kbidarkar:
@pondrejk You may also want to raise this PR against 6.3.z branch ?

PR #6020, opened 2018-05-23
[6.3] cherry-pick fix test long name CV publish

pondrejk:
cherry pick of #6019

### DIFF
--- a/tests/foreman/api/test_contentview.py
+++ b/tests/foreman/api/test_contentview.py
@@ -464,14 +464,16 @@ class ContentViewPublishPromoteTestCase(APITestCase):
 
         :id: 1c786756-266d-49b2-912f-7808096f5cc0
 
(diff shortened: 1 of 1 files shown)


ldjebran:
@pondrejk @rplevka same comment as here  https://github.com/SatelliteQE/robottelo/pull/6019#issuecomment-392450470
please rebase also :)

ldjebran:
basically ACK as already approved the other but: 
that BZ is for 6.3 or 6.4 as this is the same as in the other PR, perhaps it has to be cloned for 6.3 and put the right flags on BZ.
@pondrejk @rplevka  what do think?

pondrejk:
@ldjebran rebased. To be honest, I'd rather wait to see how and if the bz is resolved, it may as well be wontfix, as the potential of hitting it by users is kind of low -- only appears with 211 to 255 character hostnames

PR #6028, opened 2018-05-25
UI-test improvements foreman-1.16

m-bucher:
Far away from getting a good pass-rate on foreman-1.16, but these fix some of the problems that occured with foreman-1.16.

### DIFF
--- a/robottelo/ui/base.py
+++ b/robottelo/ui/base.py
@@ -455,14 +455,16 @@ class Base(object):
         angular_active = False
 
         try:
-            jquery_active = driver.execute_script('return jQuery.active') > 0
+            val = driver.execute_script('return jQuery.active')
+            jquery_active = isinstance(val, int) and val > 0
         except WebDriverException:
             pass
 
         try:
-            angular_active = driver.execute_script(
+            val = driver.execute_script(
                 u'return angular.element(document).injector().get("$http")'
-                u'.pendingRequests.length') > 0
+                u'.pendingRequests.length')
+            angular_active = isinstance(val, int) and val > 0
         except WebDriverException:
             pass
 
--- a/robottelo/ui/navigator.py
+++ b/robottelo/ui/navigator.py
@@ -450,7 +450,7 @@ class Navigator(Base):
         self.click(menu_locators['menu.current_text'])
         # get to left corner of the browser instance to not have impact on
         # further actions
-        self.perform_action_chain_move_by_offset(-150, -150)
+        self.perform_action_chain_move_by_offset(0, 0)
         return org
 
     def go_to_select_loc(self, loc):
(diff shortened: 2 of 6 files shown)


robottelo/ui/base.py
```diff
@@ -455,14 +455,16 @@ def ajax_complete(self, driver):
         angular_active = False
 
         try:
-            jquery_active = driver.execute_script('return jQuery.active') > 0
+            val = driver.execute_script('return jQuery.active')
+            jquery_active = type(val) == int and val > 0
```
ldjebran: prefer 
```python
jquery_active =  isinstance(val, int) and val > 0
```

robottelo/ui/base.py
```diff
@@ -455,14 +455,16 @@ def ajax_complete(self, driver):
...
         except WebDriverException:
             pass
 
         try:
-            angular_active = driver.execute_script(
+            val = driver.execute_script(
                 u'return angular.element(document).injector().get("$http")'
-                u'.pendingRequests.length') > 0
+                u'.pendingRequests.length')
+            angular_active = type(val) == int and val > 0
```
ldjebran: prefer
```python
angular_active =  isinstance(val, int) and val > 0
```

robottelo/ui/navigator.py
```diff
@@ -12,6 +12,7 @@ def menu_click(self, top_menu_locator, sub_menu_locator,
                    tertiary_menu_locator=None):
         self.perform_action_chain_move(top_menu_locator)
         if not tertiary_menu_locator:
+            self.wait_until_element(sub_menu_locator)
```
ldjebran: do not see why we have to call explicitly  wait_until_element here as already called in click function

ldjebran:
Please supply tests results for the modified tests, (as we are in master they should be run for 6.4 server)

robottelo/ui/navigator.py
m-bucher: Might have been a problem with my test-setup. I tested with firefox, but as I learned recently there seem to be problems with firefox not waiting for elements.
I can remove this line, if you like.

abalakh:
@omaciel frankly i'm not sure what to do with this PR. It partially fixes legacy robottelo UI automation for foreman-1.16. But we do not support legacy UI automation anymore :)
And, basically, it will be a problem to test these changes - latest selenium is set in requirements, which won't work correctly with old FF42-45, not to mention we don't use FF anymore, so in order to test it out one have to roll back a lot of changes.
I'd vote for closing this PR and recommending to use/contribute to airgun instead :)

abalakh:
@m-bucher yeah, if foreman-1.16 is using horizontal navigation, then sat 6.3.z is your choice. But 6.3.z is using older selenium fully compatible with FF42-47, so most of your changes shouldn't be required there. Pls try it

rplevka:
@m-bucher hey. Are you still interested in this? it's been untouched for some time now

m-bucher:
@rplevka No, since master is vertical-nav anyway, there should be no need to merge a fix for the old-navigation into it.
You may close this PR, thanks.

rplevka:
Thx for confirmation

PR #6032, opened 2018-05-28
RSS feed settings CLI tests

rdrazny:
$ pytest tests/foreman/cli/test_setting.py -k "rssfeed"
================== test session starts ===================
platform linux -- Python 3.6.5, pytest-3.4.0, py-1.5.3, pluggy-0.6.0
rootdir:~/PycharmProjects/robottelo, inifile:
plugins: wait-for-1.0.9, xdist-1.22.2, services-1.2.1, mock-1.6.3, forked-0.2, cov-2.5.1
collected 27 items                                       
2018-05-29 11:28:39 - conftest - DEBUG - BZ deselect is disabled in settings

tests/foreman/cli/test_setting.py ..               [100%]

================== 25 tests deselected ===================
======== 2 passed, 25 deselected in 48.97 seconds ========

### DIFF
--- a/tests/foreman/cli/test_setting.py
+++ b/tests/foreman/cli/test_setting.py
@@ -324,8 +324,8 @@ class SettingTestCase(CLITestCase):
         orig_url = Settings.list({'search': 'name=rss_url'})[0]['value']
         for test_url in valid_url_list():
             Settings.set({'name': 'rss_url', 'value': test_url})
-            updated_url = Settings.list({'search': 'name=rss_url'})[0]['value']
-            self.assertEqual(test_url, updated_url)
+            updated_url = Settings.list({'search': 'name=rss_url'})[0]
+            self.assertEqual(test_url, updated_url['value'])
         Settings.set({'name': 'rss_url', 'value': orig_url})
 
 


tests/foreman/cli/test_setting.py
```diff
@@ -285,6 +286,50 @@ def test_positive_update_send_welcome_email(self):
...
+        :steps: Set rss_enable true or false
+
+        :expectedresults: rss_enable is updated
+
+        :caseautomation: automated
+        """
+        for value in ['true', 'false']:
+            Settings.set({'name': 'rss_enable', 'value': value})
+            rss_setting = Settings.list({'search': 'name=rss_enable'})[0]
+            self.assertEqual(value, rss_setting['value'])
```
pondrejk: should we record and restore the default value here? It's done in the second test so if there is a risk of interfering with other tests, I'd do it also here

tests/foreman/cli/test_setting.py
```diff
@@ -285,6 +286,50 @@ def test_positive_update_send_welcome_email(self):
...
+            4. Restore the original feed URL
+
+        :expectedresults: RSS feed URL is updated
+
+        :caseautomation: automated
+        """
+        orig_url = Settings.list({'search': 'name=rss_url'})[0]['value']
+        # The 'valid_url_list()' func  provides list of URLs,
+        # and we need only one.
+        # And there is always at least one URL, so 0 is safe.
```
pondrejk: I'd say the comment is not needed, but why not check both http and https alternative when we have chance?

tests/foreman/cli/test_setting.py
rdrazny: Ok, good point, I have missed this. Added.

tests/foreman/cli/test_setting.py
rdrazny: This test is more about the ability to update the URL setting than what kind of URL schema is used, but we can check both alternatives, no problem.

omaciel:
@rdrazny @jhutar @pondrejk I was wondering, why not use a `tearDown` method to reset the values back to their original values? If something fails during the test, the original value would never be reset... or perhaps use `mock`?

jhutar:
Agree. Created https://github.com/SatelliteQE/robottelo/issues/6039

PR #6074, opened 2018-06-18
Tests timeout on UI, removing them as they are covered in CLI and API

san7ket:


jyejare:
@san7ket , Why dont u add API and CLI counter part of these tests in this PR.

san7ket:
@jyejare they are already covered, these were redundant tests from UI interface

PR #6241, opened 2018-08-30
Fixes discovery rule tests not picking random priority id 

### LINKED ISSUE #6230: discovery rule tests not picking random priority id
- [ ] test_positive_create_with_priority

- [ ] test_positive_update_priority

jameerpathan111:
Fixes #6230 

Test Result:

```
$ pytest tests/foreman/cli/test_discoveryrule.py -k "test_positive_create_with_priority or test_positive_update_priority"
========================================================================= test session starts ==========================================================================
collecting 31 items                                                                                                                                                    
collected 31 items / 29 deselected                                                                                                                                     

tests/foreman/cli/test_discoveryrule.py ..                                                                                                                       [100%]

=============================================================== 2 passed, 29 deselected in 88.94 seconds ===============================================================

```

### DIFF
--- a/tests/foreman/cli/test_discoveryrule.py
+++ b/tests/foreman/cli/test_discoveryrule.py
@@ -15,6 +15,7 @@
 
 :Upstream: No
 """
+import random
 
 from fauxfactory import gen_alphanumeric, gen_choice, gen_string
 from robottelo.cli.base import CLIReturnCodeError
@@ -245,9 +246,11 @@ class DiscoveryRuleTestCase(CLITestCase):
 
         :CaseImportance: Critical
         """
-        rule_priority = '100'
-        rule = self._make_discoveryrule({u'priority': rule_priority})
-        self.assertEqual(rule['priority'], rule_priority)
+        available = set(range(1, 1000)) - set(
+            [r['priority'] for r in DiscoveryRule.list()])
+        rule_priority = random.sample(available, 1)
+        rule = self._make_discoveryrule({u'priority': rule_priority[0]})
+        self.assertEqual(rule['priority'], str(rule_priority[0]))
 
     @run_only_on('sat')
     @tier1
@@ -505,11 +508,17 @@ class DiscoveryRuleTestCase(CLITestCase):
 
         :CaseImportance: Critical
         """
-        rule = self._make_discoveryrule({u'priority': 100})
-        new_priority = '1'
-        DiscoveryRule.update({'id': rule['id'], 'priority': new_priority})
+        available = set(range(1, 1000)) - set(
+            [r['priority'] for r in DiscoveryRule.list()])
+        rule_priority = random.sample(available, 1)
+        rule = self._make_discoveryrule({u'priority': rule_priority[0]})
+        self.assertEqual(rule['priority'], str(rule_priority[0]))
+        available = set(range(1, 1000)) - set(
+            [r['priority'] for r in DiscoveryRule.list()])
+        rule_priority = random.sample(available, 1)
+        DiscoveryRule.update({'id': rule['id'], 'priority': rule_priority[0]})
         rule = DiscoveryRule.info({'id': rule['id']})
-        self.assertEqual(rule['priority'], new_priority)
+        self.assertEqual(rule['priority'], str(rule_priority[0]))
 
     @run_only_on('sat')
     @tier1


tests/foreman/cli/test_discoveryrule.py
```diff
@@ -245,9 +245,16 @@ def test_positive_create_with_priority(self):
 
         :CaseImportance: Critical
         """
-        rule_priority = '100'
+        rule_list = DiscoveryRule.list()
+        p_list = []
+        for rule in rule_list:
+            p_list.append(rule['priority'])
+        while True:
+            rule_priority = gen_integer(1, 100)
```
rplevka: why limiting this to just 100? i'd increase this to at least 1000 to lower the probability of a conflict

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -245,9 +245,16 @@ def test_positive_create_with_priority(self):
 
         :CaseImportance: Critical
         """
-        rule_priority = '100'
+        rule_list = DiscoveryRule.list()
+        p_list = []
+        for rule in rule_list:
+            p_list.append(rule['priority'])
+        while True:
```
rplevka: i'd probably use some upper limit for the max number of iterations as a safety fuse

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -245,9 +245,16 @@ def test_positive_create_with_priority(self):
 
         :CaseImportance: Critical
         """
-        rule_priority = '100'
+        rule_list = DiscoveryRule.list()
+        p_list = []
```
rplevka:
```python
p_list = [r['priority'] for r in DiscoveryRule.list()]
```

tests/foreman/cli/test_discoveryrule.py
jameerpathan111: thanks for pointing that out
i will limit the no. of iterations

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -505,11 +512,24 @@ def test_positive_update_priority(self):
...
+            old_priority = gen_integer(1, 100)
+            if str(old_priority) not in p_list:
+                break
+        rule = self._make_discoveryrule({u'priority': old_priority})
+        self.assertEqual(rule['priority'], str(old_priority))
+        p_list.append(str(old_priority))
+        while True:
+            new_priority = gen_integer(1, 100)
+            if str(new_priority) not in p_list:
+                break
```
rplevka: again, i'd replace all the above added code by this 2-liner:
```python
import random
```

```python
available = set(range(1,1000)) - [r['priority'] for r in DiscoveryRule.list()]
rule_priority = random.sample(available, 1)
```

rplevka: one could use the 1st available value from the set straight away, but i would suggest using the random since otherwise we might hit the conflict with similar tests running in parallel (e.g. 2 tests list the disco rules at the same time, getting the the same p_list would try to create a disco rule with the same priority id again, leading one of them to fail)

tests/foreman/cli/test_discoveryrule.py
jameerpathan111: your right, i will change it as well

rplevka: pls, check my suggestions in the review comments

tests/foreman/cli/test_discoveryrule.py
jameerpathan111: Thanks @rplevka for your comments. I will do the changes you suggested

rplevka:
@jameerpathan111 thanks for the change.
Would you mind sending updated test results so we're sure it doesn't break anything?

ACK pending test results

jameerpathan111:
Test Result:

```
$ pytest tests/foreman/cli/test_discoveryrule.py -k "test_positive_create_with_priority or test_positive_update_priority" 
========================================================================= test session starts ==========================================================================
collecting 31 items                                                                                                                                                    
collected 31 items / 29 deselected                                                                                                                                     

tests/foreman/cli/test_discoveryrule.py ..                                                                                                                       [100%]

=============================================================== 2 passed, 29 deselected in 99.37 seconds ===============================================================
```

PR #6249, opened 2018-09-05
add native report_portal logger support

rplevka:
this conditionally sets up the report_portal native logger.
The plugin is not required, thus it is placed in the `requirements-optional.txt`.
It currently points to our own fork, since the latest released version does not contain the commit needed by our code (commit that fixes the behaviour if the rp_logger configuration is not passed). And it also fixes 1 bug I discovered.

- the conftest.py tries to import the modules from the plugin conditionally, so it won't crash if the plugin is not installed.
- it also handles the scenario where the plugin is installed but the config is missing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,4 @@
 *~
 
 .pytest_cache/
+tests/foreman/pytest.ini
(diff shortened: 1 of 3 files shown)


rplevka:
@jhutar - i addressed this by adding the following code to the plugin:

https://github.com/SatelliteQE/agent-python-pytest/blob/master/pytest_reportportal/plugin.py#L94-L105

PR #6254, opened 2018-09-07
tagging some more tests with @upgrade

rplevka:
depends on: https://github.com/SatelliteQE/robottelo/pull/6255

- otherwise the improper capsule teardowns interfere with the discovery tests (they all fail at setup as they are unable to successfully deploy default PXE templae to all capsules)

### DIFF
--- a/tests/foreman/api/test_contentmanagement.py
+++ b/tests/foreman/api/test_contentmanagement.py
@@ -54,6 +54,7 @@ from robottelo.decorators import (
     skip_if_not_set,
     tier2,
     tier4,
+    upgrade
 )
 from robottelo.helpers import (
     create_repo,
@@ -272,6 +273,7 @@ class CapsuleContentManagementTestCase(APITestCase):
         self.assertEqual(len(capsule_rpms), 1)
         self.assertEqual(capsule_rpms[0], RPM_TO_UPLOAD)
 
+    @upgrade
     @tier4
     def test_positive_checksum_sync(self):
         """Synchronize repository to capsule, update repository's checksum
@@ -391,6 +393,7 @@ class CapsuleContentManagementTestCase(APITestCase):
         self.assertEqual(result.return_code, 0)
         self.assertGreater(len(result.stdout), 0)
 
+    @upgrade
     @tier4
     def test_positive_capsule_sync(self):
         """Create repository, add it to lifecycle environment, assign lifecycle
@@ -606,6 +609,7 @@ class CapsuleContentManagementTestCase(APITestCase):
             get_repo_files(cvv_repo_path)
         )
 
+    @upgrade
     @tier4
     def test_positive_iso_library_sync(self):
         """Ensure RH repo with ISOs after publishing to Library is synchronized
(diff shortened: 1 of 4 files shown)


tests/foreman/api/test_contentmanagement.py
```diff
@@ -272,6 +273,7 @@ def test_positive_uploaded_content_library_sync(self):
         self.assertEqual(len(capsule_rpms), 1)
         self.assertEqual(capsule_rpms[0], RPM_TO_UPLOAD)
 
+    @upgrade
```
ntkathole: This has already covered in upgrade workflow scenarios that's why we are able to install capsule and move forward.

tests/foreman/api/test_contentmanagement.py
```diff
@@ -391,6 +393,7 @@ def test_positive_checksum_sync(self):
         self.assertEqual(result.return_code, 0)
         self.assertGreater(len(result.stdout), 0)
 
+    @upgrade
```
ntkathole: same as above.

tests/foreman/api/test_syncplan.py
```diff
@@ -953,6 +953,7 @@ def test_positive_synchronize_rh_product_future_sync_date(self):
         self.validate_repo_content(
             repo, ['erratum', 'package', 'package_group'])
 
+    @upgrade
```
ntkathole: I think this functionality is already covered by CLI
https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/cli/test_syncplan.py#L767

tests/foreman/cli/test_contentview.py
```diff
@@ -2647,6 +2647,7 @@ def test_positive_auto_update_composite_to_latest_cv_version(self):
             components[0]['version-id'], '{0} (Latest)'.format(version_2_id))
         self.assertEqual(components[0]['current-version'], '2.0')
 
+    @upgrade
```
ntkathole: shouldn't the following tests cover this scenario?
https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/cli/test_contentview.py#L2737
https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/cli/test_contentview.py#L2888

tests/foreman/cli/test_docker.py
```diff
@@ -1280,6 +1280,7 @@ def tearDownClass(cls):
         cls.docker_host.destroy()
 
     @run_only_on('sat')
+    @upgrade
```
ntkathole: We already have docker client scenario https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/cli/test_docker.py#L1348. Not sure we should cover this specially for 'docker pull'

tests/foreman/api/test_contentmanagement.py
rplevka: ok, we might need to come up with another decorator to tag the tests we want to run in z-stream.

tests/foreman/api/test_syncplan.py
rplevka: :+1:

tests/foreman/cli/test_contentview.py
rplevka: :+1:

tests/foreman/cli/test_docker.py
rplevka: you are right, i simply missed that

PR #6304, opened 2018-09-27
remove restore tests in favor of testFM

pondrejk:
related PRs in tesfm:

https://github.com/SatelliteQE/testfm/pull/21
https://github.com/SatelliteQE/testfm/pull/5


san7ket:
Until we are sure, that the testFM, send results to polarion, I think we should not remove this.
@ntkathole @sghai Should we start sending results for testFM, or are we already doing it?

pondrejk:
@san7ket good point, but consider these tests are and will be failing as the utility is no longer working, so I don't know if failed results are better than none :)

san7ket:
@pondrejk @ntkathole I will remove the tests for this from Polarion and then we can merge this

san7ket:
Won't remove tests from polarion as they are still valid for 6.3 and before

PR #6334, opened 2018-10-09
Tokenauth testcases

pgagne:
I am adding these test cases to accomplish 2 things:
- test registries with long passwords (>255) 
- start to be able test against a larger number of registries 

As part of this, I needed to find someway to be able to define (potentially) multiple registry configs, with each config having a number of fields (username, password, etc). I didn't see a clean way of doing this with the existing robotello.properties, so I added a separate config call `container_repo.yaml` that I included a sample file for. I will create a pull request for "real" one to jenkins_configs after we have had a chance to review this. 

Please let me know if you have any thoughts on how I went about implementing the testing this. Is been a while since I have written robotello automation, so I may not be up to date on best practices.


san7ket:
@pgagne Is this a 6.5 feature ?

san7ket:
Is it necessary to have the configs in separate yaml file, can't we use the existing robotello.properties file.
If it absolutely requires a separate config, we will need to make changes in CI code to add the yaml file in our environment. So, that the tests are able to consume it's contents.

pgagne:
On the separate config, it wouldn't be an issue having this in something like `robotello.properties` however its INI makes it really awkward to specify anything more complex then simple key values. I anticipate having to have at least 3 or 4 registries in that config file . 

For example, if i was too have to have this in robotello.properties it would look something like this:

```
...
[container-repo]
registry1_url = https://registry1.example.com
registry1_username = user
registry1_password = password
registry1_repos = repo1, repo2, repo3

registry2_url = https://registry2.example.com
registry2_username = user
registry2_password = password
registry2_repos = repo4 repo5, repo6

registry3_url = https://registry3.example.com
registry3_username = user
registry3_password = password
registry3_repos = repo7, repo8, repo9
````
Then I would have to write code for each individual field, and everytime we wanted to add a registry or set of repos to test against we would have to edit the `ContainerRepositorySettings` object to parse the next `registryN_username`, `registryN_password`, etc. 

With yaml we can structure the config a bit better and it parse it more easily. IMHO we should probably consider long term moving robotello.properties to something like yaml, as it currently is a bit unwieldy. If robotello.properties was in YAML I wouldn't have added a second config file.

san7ket:
Moving to YAML is a broader discussion I suppose, which would be helpful, but the introduction of new config file will add complexity because now we will have two manage two separate config's at the same time. Also, I guess there is a better way to handle this in INI from what you mentioned. Look, for the implementation of  
https://github.com/SatelliteQE/robottelo/blob/master/robottelo.properties.sample#L120
registers=register1_url=adas,user1=,password1=pass,registery2_url=asaa.as,user2 etc.
This way you don't need to add additional code for every time you add a new registry.
And for repos to sync, I feel we should add those in constants file, because if a repo is obsolete we may want to change it.

pgagne:
Are there scripts that need to be updated to copy the file over, I would be happy to update them if you point me in the right direction. 

I am not sure I see how I would do the same thing in INI, can you point to a better example?

I guess just assumed I needed to added it to jenkins-configs and it would get checked out along with robotello.properties. I guess I also need to update robotello.properties in jenkins-configs.

flatfender:
> Is it necessary to have the configs in separate yaml file, can't we use the existing robotello.properties file.
> If it absolutely requires a separate config, we will need to make changes in CI code to add the yaml file in our environment. So, that the tests are able to consume it's contents.

I see no issues with yaml files that hold data for test case scenarios. It makes the code much more flexible.  What would be the objections to yaml files?

san7ket:
We should move everything to yaml, having separate YAML , for every feature, scenario and the maninating those would be too much. My suggestion is to keep this now in robottelo.properties as it still holds the functionality for what's needed here.
And maybe create a issue in robottelo to move the entire properties file to yaml. Maybe, someone can drive that.

san7ket:
@pgagne INI also can support dict, so for the said setting we can have a dict type data structure
Look at https://github.com/SatelliteQE/robottelo/blob/master/robottelo/config/base.py#L1175

pgagne:
@san7ket the config i need is actually a list of dicts, and I can seem to get that to work using the code currently in `INIReader`

flatfender:
> We should move everything to yaml rather than having separate YAML for every feature, scenario and maintaining those would be too much. My suggestion is to keep this now in robottelo.properties as it still holds the functionality for what's needed here.
> And maybe create a issue in robottelo to move the entire properties file to yaml. Maybe, someone can drive that.
I agree that a general place for the majority of configs such as robotello.properties is the right approach and separate yamls for each feature doesn't scale. That being said I do think that once the number of properties grows to significant value that are test case specific, a separate file makes sense.  Given that what's in this yaml is small, it can be in the main file.

So we have two options:
1. Move the contents of the yaml to robotello.properties and table the move to yaml
2. If we find moving robotello.properties to yaml is a valuable and the direction we want to move, then rename the yaml to somthing more generic to replace robotello.properties and then keep what's in the YAML and use it as a starting point to move the rest over to yaml once we determine who can take on that project.

pgagne:
Thanks @flatfender for the feedback. If @san7ket agrees with 2. I can rename the config file to something like `robotello.yaml` and put all my stuff under a container repo item in there. We can then use that as a base for new stuff.

flatfender:
> Thanks @flatfender for the feedback. If @san7ket agrees with 2. I can rename the config file to something like `robotello.yaml` and put all my stuff under a container repo item in there. We can then use that as a base for new stuff.

I think moving robotello.properties to .yaml is a good idea, but needs a wider audience before we make such a larger change.

rplevka:
bump, is this PR going somewhere or can it be closed?

pgagne:
We can probably close it. I will need to add this test back at some point, but it would probably be better as a fresh PR. We can close it and I will use the diffs as reference later

PR #6337, opened 2018-10-11
Add t3 tests for provisioning for RHV and Vmware

san7ket:
1. Adds a new requirement https://github.com/ManageIQ/wrapanapi
2. Added e2e Test for RHV and Vmware 
3. Wrapanapi will help validate the created VM's on RHV and Vmware

### DIFF
--- a/tests/foreman/longrun/test_provisioning_computeresource.py
+++ b/tests/foreman/longrun/test_provisioning_computeresource.py
@@ -1,0 +1,277 @@
+# -*- encoding: utf-8 -*-
+"""
+:CaseLevel: Acceptance
+
+:CaseComponent: CLI
+
+:TestType: Functional
+
+:CaseImportance: Critical
+
+:Upstream: No
+"""
+from fauxfactory import gen_string
+from nailgun import entities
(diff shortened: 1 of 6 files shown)


tests/foreman/longrun/test_provisioning_computeresource.py
```diff
@@ -0,0 +1,273 @@
...
+            'root-password': gen_string('alpha'),
+            'organization': self.org_name,
+            'location': self.loc_name,
+            'hostgroup': self.config_env['host_group'],
+            'pxe-loader': 'PXELinux BIOS',
+            'compute-resource-id': vmware_cr.get('id'),
+            'compute-attributes': "cpus=2,"
+                                  "corespersocket=2,"
+                                  "memory_mb=4028,"
+                                  "cluster=Satellite_Engineering,"
```
sghai: do you think reading cluster, datastore values should be defined out of tests, may be in setup? For now its fine as one test is using them. But later if multiple tests uses them then reading from one place would be good.

sghai:
Looks good to me.. ACK

Issue #6348, opened 2018-10-15
configparser is Needed for Robottelo


latran:
RHEL 7 needs to have configparser in order to run robotello.
```
(pythonenv) -bash-4.2# pytest tests/foreman/api/test_setting.py -v
Traceback (most recent call last):
  File "/root/robottelo/pythonenv/lib/python2.7/site-packages/_pytest/config.py", line 392, in _importconftest
    mod = conftestpath.pyimport()
  File "/root/robottelo/pythonenv/lib/python2.7/site-packages/py/_path/local.py", line 668, in pyimport
    __import__(modname)
  File "/root/robottelo/pythonenv/lib/python2.7/site-packages/_pytest/assertion/rewrite.py", line 216, in load_module
    py.builtin.exec_(co, mod.__dict__)
  File "/root/robottelo/pythonenv/lib/python2.7/site-packages/py/_builtin.py", line 221, in exec_
    exec2(obj, globals, locals)
  File "<string>", line 7, in exec2
  File "/root/robottelo/tests/foreman/conftest.py", line 13, in <module>
    from robottelo.cleanup import EntitiesCleaner
  File "/root/robottelo/robottelo/cleanup.py", line 6, in <module>
    from robottelo.cli.base import CLIReturnCodeError
  File "/root/robottelo/robottelo/cli/base.py", line 6, in <module>
    from robottelo import ssh
  File "/root/robottelo/robottelo/ssh.py", line 14, in <module>
    from robottelo.config import settings
  File "/root/robottelo/robottelo/config/__init__.py", line 1, in <module>
    from robottelo.config.base import Settings
  File "/root/robottelo/robottelo/config/base.py", line 18, in <module>
    import airgun.settings
  File "/root/robottelo/pythonenv/lib/python2.7/site-packages/airgun/__init__.py", line 1, in <module>
    from airgun.settings import Settings
  File "/root/robottelo/pythonenv/lib/python2.7/site-packages/airgun/settings.py", line 4, in <module>
    from configparser import ConfigParser
ImportError: No module named configparser
ERROR: could not load /root/robottelo/tests/foreman/conftest.py
```
Installation doc needs to be updated to include this info.

pgagne:
Looks like `ConfigParser` v `configparser` issue that might be handled by `six`

PR #6349, opened 2018-10-15
Update index.rst

latran:
Updated installer to include configparser for python2.7 on sat6.2.


latran:
Just also wanted to add this is for fix #6348

PR #6410, opened 2018-10-18
Removed redundant and deprecated call to configparser.readpf

### LINKED ISSUE #6339: robottelo/config/base.py:62: DeprecationWarning: This method will be removed in future versions.  Use 'parser.read_file()' instead.
Noticed this whenever I run robotello:
```
robottelo/robottelo/config/base.py:62: DeprecationWarning: This method will be removed in future versions.  Use 'parser.read_file()' instead.
```

We should update it to remove this.

latran:
fixes #6339

### DIFF
--- a/robottelo/config/base.py
+++ b/robottelo/config/base.py
@@ -4,7 +4,7 @@ import logging
 import logging.config
 
 import os
-import sys
+import six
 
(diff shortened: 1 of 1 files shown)


rplevka:
ACK

PR #6415, opened 2018-10-22
Removing upgrade folder from robottelo tests

jyejare:
Removing the upgrade folders which was left while renaming upgrade to upgrades. Also moving one test from test_role.


jyejare:
ACK from @ntkathole ,  he facing network issues. So merging.

PR #6471, opened 2018-11-14
In Satellite 6.2, bug 1427192 was resolved as CLOSED(NEXTRELEASE)

jhutar:
That means it was fixed in 6.3. We do not need these tests because we do not plan any more investment into 6.2 automation now.


pondrejk:
@jhutar but also remove the imported decorator to make travis happy
./tests/foreman/api/test_host.py:37:1: F401 'robottelo.decorators.skip_if_bug_open' imported but unused

jhutar:
@pondrejk Thank you! Should be fixed now.

Issue #6489, opened 2018-11-23
Fix preupgrade scenarios tests


ntkathole:
- tests.upgrades.test_syncplan.ScenarioSyncPlan.test_pre_sync_plan_migration

`AttributeError: 'ScenarioSyncPlan' object has no attribute 'sync_plan_name'`

- tests.upgrades.test_client.Scenario_upgrade_old_client_and_package_installation.test_pre_scenario_preclient_package_installation
`
KeyError: 'module_stream_count'`

sghai:
Is this the only test failure in scenario tests ?

ntkathole:
Yes, in pre upgrade only. Post upgrade not run yet.

ntkathole:
@jyejare just an update here. Both the tests are still failing but this time (snap 7) error changed as below:

- tests.upgrades.test_syncplan.ScenarioSyncPlan.test_pre_sync_plan_migration

`MissingValueError: Cannot find a value for the "foreman_tasks_recurring_logic" field. Searched for keys named ('foreman_tasks_recurring_logic', 'foreman_tasks_recurring_logic_id'), but available keys are [u'description', u'created_at', u'interval', u'enabled', u'updated_at', u'organization_id', u'sync_date', u'next_sync', u'products', u'permissions', u'id', u'name'].`

- tests.upgrades.test_client.Scenario_upgrade_old_client_and_package_installation.test_pre_scenario_preclient_package_installation
`KeyError: 'registry_unauthenticated_pull`

jyejare:
@ntkathole , I think changes are happened in Feature and @omkarkhatavkar  is the best person to handle that.

omkarkhatavkar:
@ntkathole @jyejare : need latest nailgun to solve this

ntkathole:
Update with snap 10, failed tests are:

- tests.upgrades.test_capsule.Scenario_capsule_sync.test_pre_user_scenario_capsule_sync
```
_, env_name = hammer.hammer_determine_cv_and_env_from_ak(
>           self.activation_key, '1')
AttributeError: 'dict' object has no attribute 'iteritems'
```

- tests.upgrades.test_contentview.ScenarioBug1429201.test_pre_user_scenario_bug_1429201

```
 print(hammer.hammer_product_create(self.prd_name, self.org_id))
E           AttributeError: 'dict' object has no attribute 'iteritems'
```

[...]

PR #6491, opened 2018-11-26
Fix 6.5 syncplan scenario failing

jyejare:
Fix SyncPlan Upgrade scenario failing with error :
```
AttributeError: 'ScenarioSyncPlan' object has no attribute 'sync_plan_name'
```


ntkathole:
Issue #6489

PR #6513, opened 2018-11-29
Add API tests around Registry Name Pattern field

mirzal:
Registry Name Pattern is new feature in future Satellite 6.5 release. These tests cover it on API side.

It depends on Nailgun [PR #573](https://github.com/SatelliteQE/nailgun/pull/573).

```
$ pytest -v -k 'name_change or pattern' tests/foreman/api/test_docker.py 
platform linux -- Python 3.6.3, pytest-3.6.1, py-1.7.0, pluggy-0.6.0 
tests/foreman/api/test_docker.py::DockerContentViewTestCase::test_negative_promote_and_set_non_unique_name_pattern PASSED                                                               [ 20%]
tests/foreman/api/test_docker.py::DockerContentViewTestCase::test_negative_set_non_unique_name_pattern_and_promote PASSED                                                               [ 40%]
tests/foreman/api/test_docker.py::DockerContentViewTestCase::test_positive_name_pattern_change PASSED                                                                                   [ 60%]
tests/foreman/api/test_docker.py::DockerContentViewTestCase::test_positive_product_name_change_after_promotion PASSED                                                                   [ 80%]
tests/foreman/api/test_docker.py::DockerContentViewTestCase::test_positive_repo_name_change_after_promotion PASSED
```

### DIFF
--- a/tests/foreman/api/test_docker.py
+++ b/tests/foreman/api/test_docker.py
@@ -857,6 +857,7 @@ class DockerContentViewTestCase(APITestCase):
             self.assertEqual(len(comp_cvv.read().environment), i + 1)
 
     @tier2
+    @upgrade
     @run_only_on('sat')
     def test_positive_name_pattern_change(self):
         """Promote content view with Docker repository to lifecycle environment.
--- a/tests/foreman/api/test_docker.py
+++ b/tests/foreman/api/test_docker.py
@@ -856,6 +856,228 @@ class DockerContentViewTestCase(APITestCase):
             promote(comp_cvv, lce.id)
             self.assertEqual(len(comp_cvv.read().environment), i + 1)
 
+    @tier2
+    @run_only_on('sat')
+    def test_positive_name_pattern_change(self):
+        """Promote content view with Docker repository to lifecycle environment.
+        Change registry name pattern for that environment. Verify that repository
+        name on product changed according to new pattern.
+
+        :id: cc78d82d-027b-4cb7-92c5-dcccf9b592ea
+
+        :expectedresults: Container repository name is changed
+            according to new pattern.
+
+        :CaseLevel: Integration
+        """
+        pattern_prefix = gen_string('alpha', 5)
+        docker_upstream_name = 'hello-world'
+        new_pattern = ("{}-<%= organization.label %>"
+                       "/<%= repository.docker_upstream_name %>").format(
+                pattern_prefix)
+
+        repo = _create_repository(
+                entities.Product(organization=self.org).create(),
+                upstream_name=docker_upstream_name)
+        repo.sync()
+        content_view = entities.ContentView(
+                composite=False,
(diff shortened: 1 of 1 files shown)


ntkathole:
@mirzal Can you please decorate one of the tests with upgrade marker so that it can be picked up by upgrade automation as well.

mirzal:
@ntkathole Added marker to most important positive test in 02e45aec137819555eeb3185d65f10066de7f587

tests/foreman/api/test_docker.py
```diff
@@ -856,6 +856,229 @@ def test_positive_promote_multiple_with_docker_repo_composite(self):
...
+                query={'environment_id': lce.id})
+
+        expected_pattern = "{}-{}/{}".format(pattern_prefix, self.org.label,
+                                             docker_upstream_name).lower()
+        self.assertEqual(lce.registry_name_pattern, new_pattern)
+        self.assertEqual(repos[0].container_repository_name, expected_pattern)
+
+    @tier2
+    @run_only_on('sat')
+    def test_positive_product_name_change_after_promotion(self):
```
rplevka: not a blocker, just a thought (which might sparkle a fruitful discussion) -
isn't this something that might get parametrized and tried on e.g. all types of repos?

Aren't these tests supposed to go to `test_repository.py` anyway?
I understood `test_docker.py` as a place where the actual integration with docker should be tested.

I can't see how the creating/renaming/updating of the repo challenges the docker integration.
Don't take me wrong, it might as well be me who's confused (and i'm certain that i'm being confused increasingly :))

mirzal: > Aren't these tests supposed to go to `test_repository.py` anyway?

Here's my reasoning for putting them here:
1. Container name pattern is part of larger feature introduced in Sat 6.5. Main idea is that users will be able to issue (part of) Docker API against Sat server. At this point, path under which we store container images will be exposed - and we want to have it configurable.\
I haven't yet get down to writing end to end tests, but I'm pretty sure they fit into this file. I wanted to have everything in one place, so tests for first part of that feature landed here.
2. Similar test cases are defined in this file. See e.g. `test_positive_create_with_name`. You could argue that all of them should be moved, but I'm merely following convention here.
3. This file already has helper function `_create_repository` that is useful for me.

Let me know if that makes sense for you, or you would prefer to move these tests anyway.

As for parametrizing tests - I don't want to turn down discussion, but I think separate issue or our new shiny satqe mailing list is better place to host it. I'm just trying to write my first API tests here :) .

rplevka: i'm actually in favor of keeping the tests organized per feature it's just that AFAIK it's currently not being followed so much elsewhere.
thanks for the reasons you provided.

sghai:
@ntkathole @rplevka : is there anything that holding this PR to merge ? Please ACK and merge.

PR #6529, opened 2018-12-05
add global fixture to record start times in junit xml

rplevka:
fixing the scenario where `junit-xml` plugin is not being used.

(sorry for any inconvenience caused)


rplevka:
please, do not merge yet

PR #6537, opened 2018-12-10
decluttering the cli rex testcase

pondrejk:
This PR removes stubbed REX tests that have _by_ip counterpart, which means the functionality is covered by other tests. To unstub these test, infrastructure changes would be required (which was originally expected), though the potential gain compared to what is currently tested would be low. This also aims to declutter the class a bit to make it easier to parametrize the class for other uses (such as testing on suse clients)


JacobCallahan:
Why were these marked as stubs, when they have code in them? Were they not fleshed out correctly?

pondrejk:
@JacobCallahan  I implemented those, then I realized our infrastructure won't allow to run them like that (hostname resolving issue), so I created copies using the by_ip setting, sutubbing the original tests thinking we might want to run also those in the future, but now I think it's not worth it

PR #6548, opened 2018-12-12
Assert there are some tags in `docker manifest list` output

mirzal:
test_positive_read_docker_tags correctly identifies that some rows in `docker manifest list` output do not have tags. But it never checked if at least one of them does have tag. When none did, last loop did not
execute and test finished with implicit pass - without running a single assertion.

Specifically assert manifests list is not empty to catch current breakage of hammer.


sghai:
ACK and roll. small change.

PR #6719, opened 2019-02-05
Updated test ostree Bug 1672491

vijay8451:
- Updated Bugzilla[1] for ostree repo.
[1] https://bugzilla.redhat.com/show_bug.cgi?id=1672491

### DIFF
--- a/tests/foreman/cli/test_repository.py
+++ b/tests/foreman/cli/test_repository.py
@@ -1884,7 +1884,7 @@ class OstreeRepositoryTestCase(CLITestCase):
 
     @tier2
     @upgrade
-    @skip_if_bug_open('bugzilla', 1658444)
+    @skip_if_bug_open('bugzilla', 1672491)
     def test_positive_synchronize_ostree_repo(self):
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_repository.py
```diff
@@ -1884,7 +1884,7 @@ def test_negative_create_unprotected_ostree_repo(self):
 
     @tier2
     @upgrade
-    @skip_if_bug_open('bugzilla', 1658444)
+    @skip_if_bug_open('bugzilla', 1672491)
     def test_positive_synchronize_ostree_repo(self):
         """Synchronize ostree repo
```
pondrejk: I'd keep the record of the bugs also in the docstring, like `bz: 1658444, 1672491` it's done a lot around robottelo and I think it's a good practice as it keeps the history of bzs associated with the test even if we no longer skip by it

vijay8451: @pondrejk agreed , and updated same.

Issue #6717, opened 2019-02-04
ContentViewSync testcases need to be updated


pgagne:
ContentViewSync testcases need to be updated to account for the changes from:
https://bugzilla.redhat.com/show_bug.cgi?id=1652526
```
tests/foreman/cli/test_satellitesync.py::ContentViewSync::test_negative_import_cv_without_replicating_import_part
tests/foreman/cli/test_satellitesync.py::ContentViewSync::test_negative_reimport_cv_with_same_major_minor
tests/foreman/cli/test_satellitesync.py::ContentViewSync::test_positive_export_import_promoted_cv
tests/foreman/cli/test_satellitesync.py::ContentViewSync::test_positive_export_import_cv
tests/foreman/cli/test_satellitesync.py::ContentViewSync::test_negative_import_without_associating_repo_to_cv 
tests/foreman/cli/test_satellitesync.py::ContentViewSync::test_positive_exported_cv_tar_contents
tests/foreman/cli/test_satellitesync.py::ContentViewSync::test_positive_export_import_cv

```

mshriver:
This should have been closed with #6720

PR #6720, opened 2019-02-05
Fix all CV Export tests

jyejare:
CV Export tests were failing due to the recent changes and Bugzilla fixes.

### DIFF
--- a/tests/foreman/cli/test_satellitesync.py
+++ b/tests/foreman/cli/test_satellitesync.py
@@ -51,7 +51,6 @@ from robottelo.decorators import (
     upgrade
 )
 from robottelo.test import CLITestCase
-from robozilla.decorators import skip_if_bug_open
 
 
 class ExportDirectoryNotSet(Exception):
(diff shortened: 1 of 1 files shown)


jyejare:
Fixes #6717

tests/foreman/cli/test_satellitesync.py
```diff
@@ -404,7 +406,6 @@ def test_positive_export_import_cv(self):
         self.assertTrue(len(imported_packages) > 0)
         self.assertEqual(len(exported_packages), len(imported_packages))
 
-    @skip_if_bug_open('bugzilla', 1655239)
```
pondrejk: If the skip is no longer justified I'd suggest just adding the number to the docstring (`:bz: 1655239`) so that we have some history recorded in case something similar reappears

jyejare: Done !

jyejare:
# py.test -q tests/foreman/cli/test_satellitesync.py::ContentViewSync
2019-02-07 14:16:31 - conftest - DEBUG - Collected 9 test cases

.........F
8 passed, 1 Failed in 792.26 seconds

The one failed is due to the manifest timeout issue we are facing from CLI. I will not fix that test until we find a common fix for all manifest upload cli tests.

PR #6721, opened 2019-02-06
some refactoring for organization cli @tier2 tests

pondrejk:
Reviewing org-related automation results I noticed that many tier2 tests are, probably from historical reasons, written in unit-test style, enforcing one action per test. These tests deal predominantly with adding and removing associations with components to organizations. Strong separation lead to in my opinion unjustified overhead, therefore I'm proposing this reshuffling based on the following rules:

- handling just tier2 tests, leaving tier1 untouched
- no assertions removed. There is still a fresh org created per test case, but combining related tests under one testcase allows to reuse the same org (for example `test_positive_add_and_remove_domains` does the work previously done by `test_positive_add_domain_by_id, test_positive_add_domain_by_name, test_positive_remove_domain_by_id`, and `test_positive_remove_domain_by_name` but with one org and reduced number of ssh calls)

Due to the scale of changes I kept the commits separated by topic, test result:

```
pytest tests/foreman/cli/test_organization.py -m tier2
==================================================== test session starts =====================================================
platform linux -- Python 3.7.2, pytest-4.0.2, py-1.7.0, pluggy-0.8.1
collecting ... 2019-02-06 09:45:37 - conftest - DEBUG - BZ deselect is disabled in settings

collected 37 items / 27 deselected                                                                                           

tests/foreman/cli/test_organization.py ..........                                                                      [100%]

=================================== 10 passed, 27 deselected, 6 warnings in 622.08 seconds ===================================
```

Tier2 run leaves 10 organizations on the test system after 622.08 seconds running in sequential run, for comparison without this change we end up with 44 orgs and 1592.94 seconds run.

### DIFF
--- a/tests/foreman/cli/test_organization.py
+++ b/tests/foreman/cli/test_organization.py
@@ -524,129 +524,68 @@ class OrganizationTestCase(CLITestCase):
         org = Org.info({'id': org['id']})
         self.assertNotIn(hostgroup['name'], org['hostgroups'])
 
-    @run_only_on('sat')
-    @skip_if_not_set('compute_resources')
-    @tier2
-    def test_positive_add_compresource_by_name(self):
-        """Add a compute resource to organization by its name
-
-        :id: 4bc1f281-ef8e-450b-8ef6-f8d11da5324f
-
-        :expectedresults: Compute Resource is added to the org
-
-        :CaseLevel: Integration
-        """
-        org = make_org()
-        compute_res = make_compute_resource({
-            'provider': FOREMAN_PROVIDERS['libvirt'],
-            'url': u'qemu+ssh://root@{0}/system'.format(
-                settings.compute_resources.libvirt_hostname
-            )
-        })
-        Org.add_compute_resource({
-            'compute-resource': compute_res['name'],
-            'name': org['name'],
-        })
-        org = Org.info({'id': org['id']})
-        self.assertEqual(org['compute-resources'][0], compute_res['name'])
-
-    @tier2
-    def test_positive_add_compresource_by_id(self):
-        """Add a compute resource to organization by its ID
-
-        :id: 355e20c5-ec04-49f7-a0ae-0864a3fe0f3f
-
-        :expectedresults: Compute Resource is added to the org
-
-        :CaseLevel: Integration
-        """
-        compute_res = make_compute_resource()
-        org = make_org({'compute-resource-ids': compute_res['id']})
-        self.assertEqual(len(org['compute-resources']), 1)
-        self.assertEqual(org['compute-resources'][0], compute_res['name'])
-
-    @tier2
-    def test_positive_add_compresources_by_id(self):
-        """Add multiple compute resources to organization by their IDs
-
-        :id: 65846f05-583b-4914-ad0a-9251ca585af5
-
-        :expectedresults: All compute resources are added to the org
-
-        :CaseLevel: Integration
-        """
-        cr_amount = random.randint(3, 5)
-        resources = [make_compute_resource() for _ in range(cr_amount)]
-        org = make_org({
-            'compute-resource-ids':
-                [resource['id'] for resource in resources],
-        })
-        self.assertEqual(len(org['compute-resources']), cr_amount)
-        for resource in resources:
-            self.assertIn(resource['name'], org['compute-resources'])
-
-    @run_only_on('sat')
     @skip_if_not_set('compute_resources')
     @skip_if_bug_open('bugzilla', 1395229)
     @tier2
     @upgrade
-    def test_positive_remove_compresource_by_id(self):
-        """Remove a compute resource from organization by its ID
+    def test_positive_add_and_remove_compresources(self):
+        """amd and remove a compute resource from organization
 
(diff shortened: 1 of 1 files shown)


omkarkhatavkar:
> Reviewing org-related automation results I noticed that many tier2 tests are, probably from historical reasons, written in unit-test style, enforcing one action per test. These tests deal predominantly with adding and removing associations with components to organizations. Strong separation lead to in my opinion unjustified overhead, therefore I'm proposing this reshuffling based on the following rules:
> 
> * handling just tier2 tests, leaving tier1 untouched
> * no assertions removed. There is still a fresh org created per test case, but combining related tests under one testcase allows to reuse the same org (for example `test_positive_add_and_remove_domains` does the work previously done by `test_positive_add_domain_by_id, test_positive_add_domain_by_name, test_positive_remove_domain_by_id`, and `test_positive_remove_domain_by_name` but with one org and reduced number of ssh calls)
> 
> Due to the scale of changes I kept the commits separated by topic, test result:
> 
> ```
> pytest tests/foreman/cli/test_organization.py -m tier2
> ==================================================== test session starts =====================================================
> platform linux -- Python 3.7.2, pytest-4.0.2, py-1.7.0, pluggy-0.8.1
> shared_function enabled - OFF - scope:  - storage: file
> rootdir: /home/pondrejk/Documents/robottelo, inifile:
> plugins: xdist-1.25.0, services-1.3.1, mock-1.10.0, forked-0.2, cov-2.6.1
> collecting ... 2019-02-06 09:45:37 - conftest - DEBUG - BZ deselect is disabled in settings
> 
> collected 37 items / 27 deselected                                                                                           
> 
> tests/foreman/cli/test_organization.py ..........                                                                      [100%]
> 
> =================================== 10 passed, 27 deselected, 6 warnings in 622.08 seconds ===================================
> ```
> Tier2 run leaves 10 organizations on the test system after 622.08 seconds running in sequential run, for comparison without this change we end up with 44 orgs and 1592.94 seconds run.

@pondrejk I really like the idea, trying to club all individual tests of same intent. Good thing is that we are reducing time without changing the actual test case objective.

tests/foreman/cli/test_organization.py
```diff
@@ -252,710 +253,304 @@ def test_positive_search_scoped(self):
...
-        self.assertEqual(len(org['subnets']), 0)
+        org_info = Org.info({'id': org['id']})
+        self.assertEqual(len(org_info['subnets']), 0,
+                         "Failed to remove subnets")
 
     @tier2
-    def test_positive_add_user_by_name(self):
-        """Add an user to organization by its name
+    def test_positive_add_and_remove_users(self):
+        """Add and remove (admi) user to organization
```
rplevka: nitpick: 'admin'

tests/foreman/cli/test_organization.py
```diff
@@ -252,710 +253,304 @@ def test_positive_search_scoped(self):
...
+                         "Failed to remove hostgroup by id")
 
+    @skip_if_not_set('compute_resources')
+    @skip_if_bug_open('bugzilla', 1395229)
     @tier2
-    def test_positive_add_templates_by_id(self):
-        """Add multiple provisioning templates to organization by their IDs
+    @upgrade
+    def test_positive_add_and_remove_compresources(self):
+        """amd and remove a compute resource from organization
```
rplevka: nitpick: 'add'

tests/foreman/cli/test_organization.py
```diff
@@ -252,710 +253,304 @@ def test_positive_search_scoped(self):
...
+            2. add and remove hostgroup by id
 
         :CaseLevel: Integration
         """
-        conf_templ = make_template()
         org = make_org()
-        Org.add_config_template({
-            'config-template-id': conf_templ['id'],
+        hostgroup_a = make_hostgroup()
+        hostgroup_b = make_hostgroup()
```
rplevka: this will look like a nitpick at first sight, but I would logically divide all the tests to the env. setup (fixture) and the actual test steps.
e.g. the above 3 lines should go to the `setup` part as they are not really object to test.
- and once we agree on this, I would further suggest to avoid using cli factories for the setup and use nailgun instead. This will bring in additional runtime boost. Just realize that every cli command means at least 1 (but usually 2) ssh sessions + some hammer overhead.

- what do you think?

pondrejk: Overall I agree with all, though I think we should approach it as a coordinated project across the cli, something similar to airgun project. So, I support the idea, but I wouldn't block this pr on it, as this is mainly to scratch my current itch with org + I don't want to make it bigger than it is

pondrejk:
Filed #6727 to follow-up on this

tests/foreman/cli/test_organization.py
omkarkhatavkar: I agree with @pondrejk. These things can be refactored later as most of them CLI tests are not using a nailgun, we can create separate issue add tests which need refactoring like this. The PR intention itself big enough to merge interrelated scenarios.

I also even recommend using pytest module fixtures whenever required to use the same as used in airgun tests. But yeah not in this PR :)

PR #6738, opened 2019-02-14
test foreman hooks

### LINKED ISSUE #6067: foreman_hooks end to end test
Some coverage of the foreman_hooks plugin is requested, from an email thread:

Typical workflow would be:

    Install and enable the plugin
    Deploy all possible hooks, all copies (or symlinks) of a simple
    script which simply records the hook JSON into a log file
    Run all tests
    Verify the log file contains expected fired hooks (test for hook presence)
    Compare JSON with expected JSON (only for the most important hook:
    Create Host, Destroy Host)

This is quite reasonable e2e test, you need a small script and a test
case that will grep a log file. Other implementations are possible,
the sky is the limit.

pondrejk:
fixes #6067
Tests create a basic logger script that gets linked in the hook dir structure. As hook events are triggered (for host and hostgroup), logs are checked if hooks are executed at proper times

test result:
```
pytest tests/foreman/sys/test_hooks.py
================================================================== test session starts ==================================================================
platform linux -- Python 3.7.2, pytest-4.0.2, py-1.7.0, pluggy-0.8.1
collecting ... 2019-02-14 13:34:22 - conftest - DEBUG - BZ deselect is disabled in settings

collected 2 items                                                                                                                                       

tests/foreman/sys/test_hooks.py ..                                                                                                                [100%]

============================================================== 2 passed in 189.67 seconds ===================================================
```

### DIFF
--- a/tests/foreman/sys/test_hooks.py
+++ b/tests/foreman/sys/test_hooks.py
@@ -1,0 +1,159 @@
+# -*- encoding: utf-8 -*-
+"""Test class for Foreman Hooks
+
+:Requirement: Foreman Hooks
+
+:CaseAutomation: Automated
+
+:CaseLevel: System
+
+:CaseComponent: SYS
+
+:TestType: Functional
+
+:CaseImportance: High
+
+:Upstream: No
+
+"""
+from fauxfactory import gen_ipaddr
+from nailgun import entities
+from requests.exceptions import HTTPError
+from robottelo import ssh
+from robottelo.decorators import run_in_one_thread, destructive
+from robottelo.test import TestCase
+from robottelo.datafactory import valid_hosts_list, valid_hostgroups_list
+
+HOOKS_DIR = '/usr/share/foreman/config/hooks'
+LOGS_DIR = '/usr/share/foreman/tmp/hooks.log'
+
+
+@run_in_one_thread
+class ForemanHooksTestCase(TestCase):
+    """Test class for testing foreman hooks"""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create logger script to be executed via hooks"""
+        super(ForemanHooksTestCase, cls).setUpClass()
+        cls.org = 1  # using Default_Organization
+        cls.script_path = "{}/logger.sh".format(HOOKS_DIR)
+        ssh.command(
+            '''printf '#!/bin/sh\necho "$(date): Executed $1 hook'''
+            + ''' on object $2" > {0}' > {1}'''.format(
+                LOGS_DIR, cls.script_path)
+        )
+        ssh.command("chmod 774 {}".format(cls.script_path))
+        ssh.command("chown foreman:foreman {}".format(cls.script_path))
+        ssh.command("restorecon -RvF {}".format(HOOKS_DIR))
+
+    @classmethod
+    def tearDownClass(cls):
+        """clean up hook files"""
+        super(ForemanHooksTestCase, cls).tearDownClass()
+        ssh.command("rm -rf {}/*".format(HOOKS_DIR))
+
+    @destructive
+    def test_positive_host_hooks(self):
+        """Create hooks to be executed on host create, update and destroy
+
+        :id: 4fe35fda-1524-44f7-9221-96d1aeafc75c
+
+        :Steps:
+            1. Create hook directories with symlinks to logger script
+            for subset of supported events: create, update, destroy
+            2. Perform trigger actions (create, update and delete host)
+            3. Observe custom logs created by hook script
+
+        :expectedresults: hook scripts are executed at proper time
+        """
+        host_name = valid_hosts_list()[0]
+        expected_msg = "Executed {0} hook on object {1}"
+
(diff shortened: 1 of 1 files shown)


tests/foreman/sys/test_hooks.py
```diff
@@ -0,0 +1,159 @@
...
+from fauxfactory import gen_ipaddr
+from nailgun import entities
+from requests.exceptions import HTTPError
+from robottelo import ssh
+from robottelo.decorators import run_in_one_thread, tier3
+from robottelo.test import TestCase
+from robottelo.datafactory import valid_hosts_list, valid_hostgroups_list
+
+HOOKS_DIR = '/usr/share/foreman/config/hooks'
+LOGS_DIR = '/usr/share/foreman/tmp/hooks.log'
```
omkarkhatavkar: This can be placed in constant.py ?

pondrejk: It could but I'd still prefer to have it in a single file for convenience, as I don't believe these could be useful in any other test than this one

tests/foreman/sys/test_hooks.py
```diff
@@ -0,0 +1,159 @@
...
+
+@run_in_one_thread
+class ForemanHooksTestCase(TestCase):
+    """Test class for testing foreman hooks"""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create logger script to be executed via hooks"""
+        super(ForemanHooksTestCase, cls).setUpClass()
+        cls.org = 1  # using Default_Organization
```
omkarkhatavkar: Any specific reason to run this test against the default organisation?

tests/foreman/sys/test_hooks.py
```diff
@@ -0,0 +1,159 @@
...
+
+        # create hook destination directories with logger script
+        create_event = "create"
+        update_event = "update"
+        destroy_event = "destroy"
+        for event in [create_event, destroy_event, update_event]:
+            hook_dir = "{0}/host/managed/{1}".format(HOOKS_DIR, event)
+            ssh.command('''mkdir -p {}'''.format(hook_dir))
+            ssh.command('''ln -sf {0} {1}/'''.format(self.script_path, hook_dir))
+        result = ssh.command("systemctl restart httpd")
```
omkarkhatavkar: As I see multiple system calls.  Can we make this test as a destructive test?
What do you think?

tests/foreman/sys/test_hooks.py
```diff
@@ -0,0 +1,159 @@
...
+
+        # update host, check logs for hook activity
+        new_ip = gen_ipaddr()
+        host.ip = new_ip
+        host = host.update(['ip'])
+        self.assertEqual(host.ip, new_ip)
+        result = ssh.command('cat {}'.format(LOGS_DIR))
+        self.assertEqual(result.return_code, 0)
+        self.assertIn(expected_msg.format(update_event, host_name), result.stdout[0])
+
```
omkarkhatavkar: rather repeating same steps, can be placed in another function e.g.
``` python
def _check_event_in_logfile(self, event_name, host_name):
  result = ssh.command('cat {}'.format(self.LOGS_DIR))
  self.assertIn(expected_msg.format(event_name, host_name), result.stdout[0])
```
What do you think ?

tests/foreman/sys/test_hooks.py
pondrejk: I thought that running in sequential tier would be safe enough, I don't mind moving it to the destructive tier if there is a call for it

tests/foreman/sys/test_hooks.py
pondrejk: I did't want to create a separate org for this for performance reasons, hooks should react to actions accross orgs, I thought I could use what we already have for these basic trigger actions.

tests/foreman/sys/test_hooks.py
pondrejk: I see what you mean, will update

tests/foreman/sys/test_hooks.py
```diff
@@ -0,0 +1,159 @@
...
+            '''printf '#!/bin/sh\necho "$(date): Executed $1 hook'''
+            + ''' on object $2" > {0}' > {1}'''.format(
+                LOGS_DIR, cls.script_path)
+        )
+        ssh.command("chmod 774 {}".format(cls.script_path))
+        ssh.command("chown foreman:foreman {}".format(cls.script_path))
+        ssh.command("restorecon -RvF {}".format(HOOKS_DIR))
+
+    @classmethod
+    def tierDownClass(cls):
```
rplevka: `s/tierDown/tearDown/`

tests/foreman/sys/test_hooks.py
```diff
@@ -0,0 +1,159 @@
...
+        """clean up hook files"""
+        super(ForemanHooksTestCase, cls).tierDownClass()
```
rplevka: and here

PR #6747, opened 2019-02-18
Proposal for test case on PXE provisioning of a RHEL8 test case

jhutar:
Plan is to automate this (in progress, based on recent @rplevka work on gating tests) and run on both RHEL7 and RHEL8 kickstarts.

Requesting @stbenjam ack here (I know you are not working on "RHEL8 provisioning" Satellite 6.5 feature now, but I have not found anybody who would claim the ownership). I have tested it manually and it looks good, this is just in an effort to automate it. Please, could you take a look and ack?

### DIFF
--- a/tests/foreman/cli/test_provisioning.py
+++ b/tests/foreman/cli/test_provisioning.py
@@ -1,0 +1,42 @@
+"""Provisioning tests
+
+:Requirement: Provisioning
+
+:CaseAutomation: Automated
+
+:CaseLevel: System
+
+:CaseComponent: Provisioning
+
+:TestType: Functional
+
+:CaseImportance: High
+
+:Upstream: No
+"""
+
+from robottelo.decorators import (
+    stubbed,
+    tier3,
+)
+
+
+@stubbed
+@tier3
+def test_rhel_pxe_provisioning_on_libvirt():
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_provisioning.py
```diff
@@ -0,0 +1,42 @@
...
+def test_rhel_pxe_provisioning_on_libvirt():
+    """Provision RHEL system via PXE on libvirt and make sure it behaves
+
+    :id: a272a594-f758-40ef-95ec-813245e44b63
+
+    :steps:
+        1. Provision RHEL system via PXE on libvirt
+        2. Check that resulting host is registered to Satellite
+        3. Check host can install package from Satellite
+
```
omkarkhatavkar: If it is only specific to RHEL8, Can you please also consider installing module streams like this one of the major feature?

jhutar: Exactly what I'm planning: On RHEL8 I want to attempt to install module, on RHEL7 I would skip that check.

jhutar:
Thank you @stbenjam - exactly right. I want to have test case generic, so not mentioning specific setup there.

Issue #6679, opened 2019-01-24
test_positive_update_interval should be  simplified


swadeley:
In "tests/foreman/api/test_syncplan.py", this test:
"def test_positive_update_interval(self):"

has this section that I found hard to understand as a Python newbie:

 "374                 sync_plan.interval = sample(
   375                     set(result.choices) - set([interval]),
   376                     1
   377                 )[0]"

In the same test there is this line:
"sync_plan.cron_expression = gen_choice(valid_cron_expressions())"

@JacobCallahan pointed out that if we use the same "gen_choice" command in the first part of the test, we can maker it easier to maintain the code and drop the "import random" to save load on the test server.

swadeley:
More urgent is that its failing with custom cron.

PR #6754, opened 2019-02-20
Fix test_positive_update_interval (#6679)

swadeley:
fix to handle create with custom cron
simplify to not use sample()

================================

I will remove the "from random import sample" in another commit after test completes

### DIFF
--- a/tests/foreman/api/test_syncplan.py
+++ b/tests/foreman/api/test_syncplan.py
@@ -381,7 +381,7 @@ class SyncPlanUpdateTestCase(APITestCase):
 
             valid_intervals = valid_sync_interval()
             valid_intervals.remove(interval)
-            new_interval = valid_intervals[0]
+            new_interval = gen_choice(valid_intervals)
             sync_plan.interval = new_interval
             if new_interval == SYNC_INTERVAL['custom']:
                 sync_plan.cron_expression = gen_choice(valid_cron_expressions())
--- a/tests/foreman/api/test_syncplan.py
+++ b/tests/foreman/api/test_syncplan.py
@@ -21,7 +21,6 @@ http://www.katello.org/docs/api/apidoc/sync_plans.html
 from datetime import datetime, timedelta
 from fauxfactory import gen_string, gen_choice
 from nailgun import client, entities
-from random import sample
 from robottelo import manifests
 from robottelo.api.utils import (
     enable_rhrepo_and_fetchid,
@@ -371,21 +370,25 @@ class SyncPlanUpdateTestCase(APITestCase):
         :CaseImportance: Critical
         """
         for interval in valid_sync_interval():
-            with self.subTest(interval):
-                sync_plan = entities.SyncPlan(organization=self.org)
-                result = sync_plan.get_fields()['interval']
-                sync_plan.interval = sample(
-                    set(result.choices) - set([interval]),
-                    1
-                )[0]
-                sync_plan = sync_plan.create()
-                sync_plan.interval = interval
-                if (interval == 'custom cron'):
-                    sync_plan.cron_expression = gen_choice(valid_cron_expressions())
-                    sync_plan = sync_plan.update(['interval', 'cron_expression'])
-                else:
-                    sync_plan = sync_plan.update(['interval'])
-                self.assertEqual(sync_plan.interval, interval)
+            sync_plan = entities.SyncPlan(
+                description=gen_string('alpha'),
+                organization=self.org,
+                interval=interval
+            )
+            if interval == SYNC_INTERVAL['custom']:
+                sync_plan.cron_expression = gen_choice(valid_cron_expressions())
+            sync_plan = sync_plan.create()
+
+            valid_intervals = valid_sync_interval()
+            valid_intervals.remove(interval)
+            new_interval = valid_intervals[0]
+            sync_plan.interval = new_interval
+            if new_interval == SYNC_INTERVAL['custom']:
+                sync_plan.cron_expression = gen_choice(valid_cron_expressions())
+                sync_plan = sync_plan.update(['interval', 'cron_expression'])
+            else:
+                sync_plan = sync_plan.update(['interval'])
+            self.assertEqual(sync_plan.interval, new_interval)
 
     @tier1
     @run_only_on('sat')


tests/foreman/api/test_syncplan.py
```diff
@@ -371,21 +371,21 @@ def test_positive_update_interval(self):
...
-                else:
-                    sync_plan = sync_plan.update(['interval'])
-                self.assertEqual(sync_plan.interval, interval)
+            sync_plan = entities.SyncPlan(
+                description=gen_string('alpha'),
+                organization=self.org,
+                interval=interval
+            )
+            if interval == SYNC_INTERVAL['custom']:
+                sync_plan.cron_expression = gen_choice((valid_cron_expressions()))
```
mirzal: Do you need parentheses around `valid_cron_expressions()`? I didn't notice any difference between version with them and without them in Python shell.

tests/foreman/api/test_syncplan.py
```diff
@@ -371,21 +371,21 @@ def test_positive_update_interval(self):
...
+            sync_plan = sync_plan.create()
+
+            if (interval == 'custom cron'):
```
mirzal: I know this is unchanged from old version, but please use `interval == SYNC_INTERVAL['custom']` and drop parentheses here.

tests/foreman/api/test_syncplan.py
swadeley: Re: parentheses around valid_cron_expressions()
I see its used in about half the occurrences of gen_choice. I will remove in my test. Would it be acceptable to remove them throughout the /tests/foreman/api/test_syncplan.py file? or just leave that for another issue?

Re: sync_plan.interval = something
The original idea was to use gen_choice as its already used in this test. The example you gave would not be random would it. But valid_intervals.remove(interval) would use a value not already in use, which is the important point here; its what we want. I will try that. 

Thank you

tests/foreman/api/test_syncplan.py
swadeley: OK

tests/foreman/api/test_syncplan.py
swadeley: My flake 8 test is passing, must be a config issue on my side.

mirzal: I agree, using `gen_choice` sounds like better idea than my proposal above.

Re: parentheses - I only saw your patch and noticed they weren't there originally, but are now. If there are other places in that file, or Robottelo, where they are around `valid_cron_expressions()`, then I think having them here is fine. They aren't really wrong - just unnecessary, as far as I can tell. That's why flake8 won't catch them.

If you want to remove them all from this file, personally I would do it in another PR. I think it's easier for reviewers to focus on one thing at a time. But nobody will complain if you do it in another commit in this PR.

swadeley: I would like to get your method working first, its invaluable experience from me. I just made that commit.

If we use gen_choice, we would still need to ensure we had a new interval.

mirzal: That's what `valid_intervals.remove(interval)` is for. After that line, we are sure that `valid_intervals` does not include interval used during sync_plan.create(). You are safe to use `gen_choice` on that list.

tests/foreman/api/test_syncplan.py
```diff
@@ -371,21 +370,25 @@ def test_positive_update_interval(self):
...
+
+            valid_intervals = valid_sync_interval()
+            valid_intervals.remove(interval)
+            new_interval = valid_intervals[0]
+            sync_plan.interval = new_interval
+            if new_interval == SYNC_INTERVAL['custom']:
+                sync_plan.cron_expression = gen_choice(valid_cron_expressions())
+                sync_plan = sync_plan.update(['interval', 'cron_expression'])
+            else:
+                sync_plan = sync_plan.update(['interval'])
```
pondrejk: Overall it looks good to me, if you share test result I'm ok with merging. iiuc there could be three update situations in this test:

non-custom > non-custom
non-custom > custom
custom > non-custom

I'm a bit concerned about the third one, If I create custom syncplan with cron expression and then change the interval but won't touch the cron experssion, will that succeed? If no, we'll need to clean cron_expression in that else statement. Anyway, test result should show

tests/foreman/api/test_syncplan.py
swadeley: I understand that line `valid_intervals.remove(interval)`, I was just wondering how many lines we save by using gen_choice  if it cannot do that as well. It seems to have other error handling benefits:
https://fauxfactory.readthedocs.io/en/latest/_modules/fauxfactory/factories/choices.html?highlight=gen_choice

tests/foreman/api/test_syncplan.py
swadeley: Hello @pondrejk , thank you for your review. Tests pass locally but Travis build failed. I am going to change the section to use gen_choice, but here is test result before I do that:

 robottelo]$ pytest -pdb tests/foreman/api/test_syncplan.py::SyncPlanUpdateTestCase::test_positive_update_interval
===================== test session starts =================================
platform linux -- Python 3.6.6, pytest-4.0.2, py-1.6.0, pluggy-0.8.1
rootdir: /sat6/robottelo, inifile:
plugins: xdist-1.23.2, services-1.3.1, mock-1.10.0, forked-0.2, cov-2.6.0
collecting ... 2019-02-22 10:31:07 - conftest - DEBUG - BZ deselect is disabled in settings

collected 1 item                                                                                                             

tests/foreman/api/test_syncplan.py .                                                                                   [100%]

======================== 1 passed in 5.05 seconds ==========================

swadeley:
Hello

After changing one line to use gen_choice, and making the commit, I ran tests about five times:

robottelo]$ pytest -pdb tests/foreman/api/test_syncplan.py::SyncPlanUpdateTestCase::test_positive_update_interval
======================== test session starts =================================
platform linux -- Python 3.6.6, pytest-4.0.2, py-1.6.0, pluggy-0.8.1
rootdir: /sat6/robottelo, inifile:
plugins: xdist-1.23.2, services-1.3.1, mock-1.10.0, forked-0.2, cov-2.6.0
collecting ... 2019-02-22 13:53:36 - conftest - DEBUG - BZ deselect is disabled in settings

collected 1 item                                                                                                                                                            

tests/foreman/api/test_syncplan.py .                                                                                                                           [100%]

======================= 1 passed in 4.39 seconds =============================

swadeley:
This is what I was looking for, how to run the same test more than once:

 robottelo]$ py.test --dist=each --tx=3*popen tests/foreman/api/test_syncplan.py::SyncPlanUpdateTestCase::test_positive_update_interval
======================= test session starts ====================================
platform linux -- Python 3.6.6, pytest-4.0.2, py-1.6.0, pluggy-0.8.1
rootdir: /sat6/robottelo, inifile:
plugins: xdist-1.23.2, services-1.3.1, mock-1.10.0, forked-0.2, cov-2.6.0
gw0 [1] / gw1 [1] / gw2 [1]
scheduling tests via EachScheduling
.                                                                                                                                                                     [100%]. [100%]. [100%]
===================== 3 passed in 10.94 seconds ==============================
2019-02-22 14:58:18 - conftest - DEBUG - Registering custom pytest_namespace

swadeley:
@pondrejk thank you for the merge.

PR #6755, opened 2019-02-20
Fixes UI Session hang on saucelabs

ldjebran:
This is the simplest way to resolve this issue as we need a test Session raised in the same process as the test, trying many scenarios with threads ..., but this one with signal alarm is the simplest and best one.
tried in one thread and with xdist , 5 tests 2 that must pass and 3 that fails, one with assertion failure one with error division by zero, and one with timeout error using a time.sleep
all tests had the needed behavior. 

<details>
  <summary>Tests</summary>
<pre lang="python">
import time
def test_aaa(session):
    rule_name = f'test_rule_1_{gen_integer(5, 1000)}'
    search = 'cpu_count'
    hg_name = 'test_hg'
    with session:
        time.sleep(70)
        session.discoveryrule.create({
            'primary.name': rule_name,
            'primary.search': search,
            'primary.host_group': hg_name,
        })
        values = session.discoveryrule.read(rule_name, 'primary')
        assert values['primary']['search'] == search

def test_aaa2(session):
    rule_name = f'test_rule_2_{gen_integer(5, 1000)}'
    search = 'cpu_count'
    hg_name = 'test_hg'
    with session:
        session.discoveryrule.create({
            'primary.name': rule_name,
            'primary.search': search,
            'primary.host_group': hg_name,
        })
        values = session.discoveryrule.read(rule_name, 'primary')
        assert values['primary']['search'] == search

def test_aaa3(session):
    rule_name = f'test_rule_3_{gen_integer(5, 1000)}'
    search = 'cpu_count'
    hg_name = 'test_hg'
    with session:
        session.discoveryrule.create({
            'primary.name': rule_name,
            'primary.search': search,
            'primary.host_group': hg_name,
        })
        a = 1/0
        values = session.discoveryrule.read(rule_name, 'primary')
        assert values['primary']['search'] == search

def test_aaa4(session):
    rule_name = f'test_rule_4_{gen_integer(5, 1000)}'
    search = 'cpu_count'
    hg_name = 'test_hg'
    with session:
        session.discoveryrule.create({
            'primary.name': rule_name,
            'primary.search': search,
            'primary.host_group': hg_name,
        })
        assert False
        values = session.discoveryrule.read(rule_name, 'primary')
        assert values['primary']['search'] == search

def test_aaa5(session):
    rule_name = f'test_rule_5_{gen_integer(5, 1000)}'
    search = 'cpu_count'
    hg_name = 'test_hg'
    with session:
        session.discoveryrule.create({
            'primary.name': rule_name,
            'primary.search': search,
            'primary.host_group': hg_name,
        })
        values = session.discoveryrule.read(rule_name, 'primary')
        assert values['primary']['search'] == search
</pre>
</details>

### DIFF
--- a/tests/foreman/ui_airgun/conftest.py
+++ b/tests/foreman/ui_airgun/conftest.py
@@ -1,4 +1,6 @@
 import logging
+import os
+import signal
 
 import nailgun.entities
 from airgun.session import Session
@@ -9,6 +11,33 @@ from robottelo.decorators import fixture
 
 
 LOGGER = logging.getLogger('robottelo')
+
+SESSION_TIMEOUT = 7000
+
+
+class SessionTimeoutError(Exception):
+    """Raised when UI session timeout"""
+
+
+def _signal_session_timeout(signum, frame):
+    """Raise exception when signal alarm call this function"""
+    LOGGER.error('UI Session reached timeout value "%s" and will be interrupted', SESSION_TIMEOUT)
+    raise SessionTimeoutError('Session timeout')
+
+
+def _signal_session_clear():
+    """Clear the signal alarm"""
+    if os.name == 'posix':
+        LOGGER.info('UI Session clear timeout alarm')
+        signal.alarm(0)
+
+
+def _signal_session_init():
+    """Initiate a signal alarm for UI session"""
+    if os.name == 'posix':
+        signal.signal(signal.SIGALRM, _signal_session_timeout)
+        signal.alarm(SESSION_TIMEOUT)
+        LOGGER.info('UI Session set alarm to timeout after %s', SESSION_TIMEOUT)
 
 
 @fixture(scope='module')
@@ -117,7 +146,12 @@ def session(test_name, module_user):
                 session.architecture.create({'name': 'bar'})
 
     """
-    return Session(test_name, module_user.login, module_user.password)
+    session = Session(test_name, module_user.login, module_user.password)
+    _signal_session_init()
+    try:
+        yield session
+    finally:
+        _signal_session_clear()
 
 
 @fixture()


tests/foreman/ui_airgun/conftest.py
```diff
@@ -10,6 +12,33 @@
 
 LOGGER = logging.getLogger('robottelo')
 
+SESSION_TIMEOUT = 7000
```
abalakh: We do have max test duration defined in settings/robottelo/webdriver_desired_capabilities/maxDuration, why not reusing that constant? Imo that'd be better idea than making up new constant. And btw max test session duration on saucelabs is 90mins (5400secs), so there's no point in waiting longer than that.

tests/foreman/ui_airgun/conftest.py
```diff
@@ -117,7 +146,12 @@ def test_foo(session):
                 session.architecture.create({'name': 'bar'})
 
     """
-    return Session(test_name, module_user.login, module_user.password)
+    session = Session(test_name, module_user.login, module_user.password)
+    _signal_session_init()
```
abalakh: I'd also start that timeout only in case we're working with saucelabs - we're facing the issue with loosing saucelabs connection, local runs are not affected, so we shouldn't forbid long local UI sessions.

ldjebran: do not think that we have tests that work fot 7000 seconds

tests/foreman/ui_airgun/conftest.py
ldjebran: because the value here will be for the overall test run as it's created when entering the test

tests/foreman/ui_airgun/conftest.py
san7ket: 7000 secs was added to wrap up a test with 2 hrs in CI , I had seen some capsule tests taking more than a hour, so this was a approx estimate. We can look into reducing this
@lpramuk @rplevka any inputs here

ldjebran:
put do not merge as we are looking for an other solution with @abalakh

san7ket:
I don't see session hang or timeout anymore after switching to 3.141 . 
@ldjebran do we still need this?

JacobCallahan:
Hopefully, the session hang issue is fixed for good. However, in case it isn't, it would be nice to have this solution in-place.

ldjebran:
let's close this PR seems not needed for the moment as the timeout is passed to python base lib, but will be still here in closed PRs if needed.

PR #6758, opened 2019-02-25
Fixes Pycurl-config issue with robottelo which fails docs building

san7ket:


setup.py
```diff
@@ -1,10 +1,18 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
+from os import system, environ
 from setuptools import find_packages, setup
 
 with open('README.rst', 'r') as f:
     README = f.read()
 
+if system('curl --version | grep NSS 2>/dev/null') != 0:
```
ldjebran: this way if we execute in the shell
```shell
 python setup.py --help
```
we will have this pycurl installed?

san7ket: i don't think so, this is for installing the correct opensssl , when we do 

`pip install -r -requirements.txt`

san7ket: I don't think so, We need to set the variable before we install pycurl
curl should already be present on the box.

ldjebran: Think this is very tricky to use pip inside setup.py as any invocation to that setup.py will try to install the pycurl, perhaps this should be done in automation script apart before invoking pip install -r requirements.txt ...

san7ket: Understood, Will close this, will try to come with something else

PR #6765, opened 2019-02-26
Add coverage around bz1671148

mirzal:
Due to limitation of how foreman managed plugins and apipie, attributes added by first plugin only were included. Bug manifested itself by hammer recognizing arguments from one plugin only. As an example, `hammer host (create|update)` recognized `--openscap-proxy-id` (from foreman_openscap) **or** `--lifecycle-environment-id`, `--autoheal`, `--content-view-id`, `--hypervisor_guest_uuids`and others (from Katello).

This adds two cases designed to fail if bug still occurs. 

First, try to create new host with `openscap-proxy-id`. If bugs exists, one test in suite will fail: either this new case, or one of many existing cases that use Katello arguments.

Second, we check `hammer host update --help` output for flags from both foreman_openscap and Katello. If bug exists, one of flags will not be present. `hammer --help` output depends on Satellite apipie cache file, so checking for string existence is enough. Whether these options actually work is verified by other cases in suite.

-----

As these tests will fail on current downstream builds (fixes are not yet cherry-picked), I am adding DO NOT MEGRE label and skipping pytest output. I will add it when this will be ready to merge. I am requesting reviews based on component ownership.

### DIFF
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -608,6 +608,26 @@ class HostCreateTestCase(CLITestCase):
         )
 
     @tier1
+    def test_positive_create_with_openscap_proxy_id(self):
+        """Check if host can be created with OpenSCAP Proxy id
+
+        :id: 3774ba08-3b18-4e64-b07f-53f6aa0504f3
+
+        :expectedresults: Host is created and has OpenSCAP Proxy assigned
+
+        :CaseImportance: Medium
+        """
+        openscap_proxy = Proxy.list({
+            'search': 'url = https://{0}:9090'.format(settings.server.hostname)
+        })[0]
+
+        host = make_fake_host({
+            'organization-id': self.new_org['id'],
+            'openscap-proxy-id': openscap_proxy['id']
+        })
+        assert host['openscap-proxy'] == openscap_proxy['id']
+
+    @tier1
     def test_negative_create_with_name(self):
         """Check if host can be created with random long names
 
@@ -646,6 +666,30 @@ class HostCreateTestCase(CLITestCase):
                 'lifecycle-environment-id': env,
                 'organization-id': self.new_org['id'],
             })
+
+    @run_only_on('sat')
+    @tier3
+    @upgrade
+    def test_positive_katello_and_openscap_loaded(self):
+        """Verify that command line arguments from both Katello
+        and foreman_openscap plugins are loaded and available
+        at the same time
+
+        :id: 5b5db1d4-50f9-45a0-bb92-4571fc8d729b
+
+        :expectedresults: Command line arguments from both Katello
+            and foreman_openscap are available in help message
+            (note: help is generated dynamically based on apipie cache)
+
+        :CaseImportance: Medium
+
+        :BZ: 1671148
+        """
+        help_output = Host.execute('host update --help')
+        for arg in ['lifecycle-environment-id', 'openscap-proxy-id']:
+            assert any(('--{}'.format(arg) in line for line in help_output)), (
+                "--{} not supported by update subcommand".format(arg)
+            )
 
     @tier3
     @upgrade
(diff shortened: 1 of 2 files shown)


mirzal:
Relevant changes are included in latest snap, so this can be properly reviewed and merged now. New commit only rebases on top of recent master. Test output below.

```
11:50 $ pytest -v -k 'test_positive_create_with_openscap_proxy_id or test_positive_katello_and_openscap_loaded' tests/foreman/cli/test_host.py
platform linux -- Python 3.6.3, pytest-3.6.1, py-1.7.0, pluggy-0.6.0 -- /home/mzalewsk/.virtualenvs/robottelo/bin/python3
cachedir: .pytest_cache
plugins: xdist-1.24.1, services-1.3.0, mock-1.10.0, forked-0.2, cov-2.6.0
collecting 96 items                                                                                                                                                                           2019-03-11 11:51:12 - conftest - DEBUG - BZ deselect is disabled in settings

collected 96 items / 94 deselected                                                                                                                                                            

tests/foreman/cli/test_host.py::HostCreateTestCase::test_positive_create_with_openscap_proxy_id PASSED                                                                                  [ 50%]
tests/foreman/cli/test_host.py::HostCreateTestCase::test_positive_katello_and_openscap_loaded PASSED                                                                                    [100%]

========================================================================== 2 passed, 94 deselected in 159.66 seconds ==========================================================================
```

tests/foreman/cli/test_host.py
```diff
@@ -607,6 +607,26 @@ def test_positive_create_with_puppet_class_name(self):
...
+        :id: 3774ba08-3b18-4e64-b07f-53f6aa0504f3
+
+        :expectedresults: Host is created and has OpenSCAP Proxy assigned
+
+        :CaseImportance: Medium
+        """
+        openscap_proxy = Proxy.list({
+            'search': 'url = https://{0}:9090'.format(settings.server.hostname)
+        })[0]
+
```
omkarkhatavkar: Do we going to have more scenarios for 'openscap-proxy' as the same case we had for puppet proxy earlier and that was placed in a setup function? What do you think?

tests/foreman/cli/test_host.py
```diff
@@ -647,6 +667,30 @@ def test_negative_create_with_unpublished_cv(self):
...
+        :expectedresults: Command line arguments from both Katello
+            and foreman_openscap are available in help message
+            (note: help is generated dynamically based on apipie cache)
+
+        :CaseImportance: Medium
+
+        :BZ: 1671148
+        """
+        help_output = Host.execute('host update --help')
```
omkarkhatavkar: a couple of questions: 
1. I'm ok adding this test here. But an adding as a unit test would be good ? what you think. 
2. ```Host.execute('host update --help')``` is any way put this into CLI command builder, if not i'm ok with this.

tests/foreman/cli/test_host.py
mirzal: I think we do lack coverage around openscap_proxy (given that this bug wasn't discovered earlier), but I'll leave it up to component owners to add more scenarios. To be honest, right now I'm not even sure what else we should cover - I added this scenario only on request of developer.

So I'd say - leave this in test case for now, move to setup (or even it's own class) as we increase coverage in future.

omkarkhatavkar: Sure No Problem

tests/foreman/cli/test_host.py
mirzal: As discussed on IRC, this would fit BATS suite rather well. This is very good idea. I'm only not sure if we can put it in upstream, as it requires two plugins to be installed at the same time. I'll look into this and get back to you, hopefully with new commit that removes this case. Adding do not merge flag for now.

ntkathole:
@mirzal ahh I missed you added do not merge again. Is it ok or you had any work pending on it?

tests/foreman/cli/test_host.py
ntkathole: Worth keeping here for downstream

mirzal:
@ntkathole I think it's alright. My intention was to submit one test upstream and remove it from there (have it run upstream and during downstream Satellite build). But it turned out that luna doesn't work, and upstream isn't prepared for tests that span multiple modules, so it's taking long time. I think it's better to have it in Robottelo for time being and remove one of tests once it is actually ported upstream. I'll send new PR then.

PR #6781, opened 2019-03-05
add ProvisionNetwork properties section

rplevka:
We currently lack the very basic provisoning subnet settings for the host provisioning.
The closest we can get is the `[vlan_networking]` section, which does not corresponds 100%. The ultimate goal is to use more generic provision networking settings and get rid of the `[vlan_networking]` by time

### DIFF
--- a/robottelo.properties.sample
+++ b/robottelo.properties.sample
@@ -223,6 +223,18 @@ ssh_key=
 # insights_client_el6repo=https://www.example.com/insights-client/repo/insights-client-6.repo
 # insights_client_el7repo=https://www.example.com/insights-client/repo/insights-client-7.repo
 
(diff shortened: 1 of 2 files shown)


rplevka:
while i don't use this in any tests inside robottelo so far, I rely on these settings in my smoke test (which will most likely land in the forklift repo)

omkarkhatavkar:
@rplevka Do you have any mock test with the result, which verifies that this code works fine from robottelo?

rplevka:
@omkarkhatavkar 
After a discussion with @lpramuk I'm about to do a major rework on this. The easiest way here is to close the PR and i'll reopen a new one.
Will provide some test results there.

PR #6855, opened 2019-04-10
Increase delay in UI sync plans (#6854)

swadeley:
Changing delay from 5 to 10 minutes to allow more time for slow web UI to complete the set up tasks before the scheduled sync event.

### DIFF
--- a/tests/foreman/ui_airgun/test_syncplan.py
+++ b/tests/foreman/ui_airgun/test_syncplan.py
@@ -231,7 +231,7 @@ def test_positive_synchronize_custom_product_past_sync_date(
     :CaseLevel: System
     """
     interval = 60 * 60  # 'hourly' sync interval in seconds
-    delay = 5 * 60
+    delay = 10 * 60
     plan_name = gen_string('alpha')
     product = entities.Product(organization=module_org).create()
     repo = entities.Repository(product=product).create()
@@ -279,7 +279,7 @@ def test_positive_synchronize_custom_product_future_sync_date(
 
     :CaseLevel: System
     """
-    delay = 5 * 60  # delay for sync date in seconds
+    delay = 10 * 60  # delay for sync date in seconds
     plan_name = gen_string('alpha')
     product = entities.Product(organization=module_org).create()
     repo = entities.Repository(product=product).create()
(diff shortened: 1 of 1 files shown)


vijay8451:
@swadeley Howmuch is the failure rate with `delay = 5 * 60` ?

swadeley:
@vijay8451 I counted eight tests with `test_positive_synchronize` in the file `tests/foreman/ui_airgun/test_syncplan.py` and eight are failing. But I just had `test_positive_synchronize_custom_product_custom_cron_past_sync_date` fail locally after updating the interval to 10 minutes so there ios some other issue there I need to investigate.

vijay8451:
> @vijay8451 I counted eight tests with `test_positive_synchronize` in the file `tests/foreman/ui_airgun/test_syncplan.py` and eight are failing. But I just had `test_positive_synchronize_custom_product_custom_cron_past_sync_date` fail locally after updating the interval to 10 minutes so there is some other issue there I need to investigate.

Okay , Good to catch the root cause as that would give more clear picture here.

swadeley:
> > @vijay8451 I counted eight tests with `test_positive_synchronize` in the file `tests/foreman/ui_airgun/test_syncplan.py` and eight are failing. But I just had `test_positive_synchronize_custom_product_custom_cron_past_sync_date` fail locally after updating the interval to 10 minutes so there is some other issue there I need to investigate.
> 
> Okay , Good to catch the root cause as that would give more clear picture here.

Like all the sync plan tests this creates a start date minus an interval but less a few minutes standard delay, 4 mins in API tests but in the web UI tests I see 5 mins delay.

The test has this line:
'cron_expression': '*/5 * * * *',
that means every 5 mins starting on the hour, not from "now".

This means the time from start to next sync could be anything 1 to 5 mins, not the standard delay. The `validate_task_status` check will run for 10 mins if I understand correctly, so that should cope with a 5 min delay to the next cron sync time, provided someone like me does not increase the delay from 4 mins to 10 mins to cope with slow web UI response on test servers.

swadeley:
Re https://github.com/SatelliteQE/robottelo/pull/6855#issuecomment-482522706
I created new issue for that as it completely different to the issue that this pull request relates to:

https://github.com/SatelliteQE/robottelo/issues/6857

swadeley:
> To be honest - I don't think that is a solution. Test automation script should work stable no matter what environment you have. If tomorrow we have slow down - you get fail anyway. 

I agree this pull request contains a work around not an ideal solution and might still fail if the system slows down even further (as would other tests). On the other hand, I have as yet no way to improve the responsiveness of the test infrastructure. I will work on that and just run the tests locally every time they fail until the infrastructure is improved.

PR #6860, opened 2019-04-16
6.6 Reporting testcases

lhellebr:


lhellebr:
@ares

tests/foreman/ui_airgun/test_reporttemplates.py
```diff
@@ -256,7 +260,7 @@ def test_positive_generate_report_hoststatus_filter(session):
 @tier2
 @stubbed()
 def test_positive_generate_report_hoststatus_sanitized(session):
-    """ Edit a report template
+    """ Generate a report for some host with comma in some field
```
ares: what is this testing? what field the comma put to?

tests/foreman/ui_airgun/test_reporttemplates.py
```diff
@@ -416,3 +422,123 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+
+    :id: cd19b90d-836f-4efd-c4bc-d5e09a909a67
+
+    :setup: User with reporting access rights, some Host, some report with host input
+
+    :steps:
+
+        1. Monitor -> Report Templates
+        2. Registered Hosts -> Generate
+        3. Check that the result shoudl be sent via e-mail
```
ares: s/shoudl/should/

tests/foreman/ui_airgun/test_reporttemplates.py
```diff
@@ -416,3 +422,123 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+        4. Submit
+        5. Receive the e-mail
```
ares: if the email address is wrongly formatted, I guess it should be sent via email right? instead, user should see the error

tests/foreman/ui_airgun/test_reporttemplates.py
```diff
@@ -416,3 +422,123 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+    :id: cd19b90d-836f-4efd-c3bc-d5e09a909a67
+
+    :setup: User with reporting access rights, some Host
+
+    :steps:
+
+        1. Monitor -> Report Templates
+        2. Registered Hosts -> Generate
+        3. Set schedule to current time + 1 minute
+        4. Check that the result should be sent via e-mail
```
ares: and is compressed (always)

lhellebr:
I know there are API and CLI changes. In this pull request, there is only a WebUI testcases commit so far.

The cleanup test would take 24 hours to run (or require meddling with time). I don't think it's worth it.

I will add the test that only the report author can download a report, good point.

Why would the report generation be cancelled?

tests/foreman/ui_airgun/test_reporttemplates.py
lhellebr: This tests if the comma in some field (in steps, it is specified it's gonna be a OS field but it doesn't matter much) doesn't break the resulting CSV.

This is not related to the new feature, there was bad description and I just fixed it as part of the commit.

ares:
> The cleanup test would take 24 hours to run (or require meddling with time). I don't think it's worth it.

Would it help if I tell you, where in the code to change it? The risk is, we may realize in few months that the DB contains GBs of data and there's no clean up. Even manual test would do I think.

> Why would the report generation be cancelled?

e.g. I scheduled something I know will take a long time to midnight, later that day I realize I made a typo in the input so I cancel and resubmit with new data. Probably not something that happens every day and would require the knowledge, there's a foreman task that I need to find and cancel.

tests/foreman/ui_airgun/test_reporttemplates.py
ares: Ah, got it, some value that us printed to the report will contain comma. Ok makes sense, thank you

lhellebr:
@ares I've just pushed CLI and API testcases

san7ket:
This should not go into master, just yet. We still need to branch 6.5. Date to branch is 26th April. After that we can merge this into master

tests/foreman/api/test_reporttemplates.py
```diff
@@ -293,3 +293,135 @@ def test_negative_delete_locked_report(self):
...
+        :expectedresults: Report cleanup task is running
+
+        :CaseImportance: Medium
+        """
+
+    @tier2
+    @stubbed()
+    def test_negative_nonauthor_of_report_cant_download_it(self):
+        """The resulting report should only be downloadable by
+           the user that generated it. Check.
```
ares: or admin, admins can download all reports

tests/foreman/cli/test_reporttemplates.py
```diff
@@ -349,3 +349,135 @@ def test_negative_delete_locked_report(self):
...
+           the user that generated it. Check.
+
+        :id: a4bc77db-146e-4871-a42e-e93887464986
+
+        :setup: Installed Satellite, user that can list running tasks
+
+        :steps:
+
+            1. hammer report-template schedule
+            2. hammer report-template report-data
```
ares: different user credentials need to be used here ofc, also make sure the user is not admin

lhellebr:
Can we merge this? @san7ket , @omkarkhatavkar ?

PR #6878, opened 2019-04-25
Implementing some stubbed reporting tests

lhellebr:
Depends on https://github.com/SatelliteQE/nailgun/pull/629

### DIFF
--- a/tests/foreman/ui/test_reporttemplates.py
+++ b/tests/foreman/ui/test_reporttemplates.py
@@ -6,7 +6,7 @@
 
 :CaseLevel: Acceptance
 
-:CaseComponent: UI
+:CaseComponent: Reporting
 
 :TestType: Functional
 
@@ -15,264 +15,29 @@
 :Upstream: No
 """
 
+import csv
+import os
+
 from nailgun import entities
 
 from robottelo.datafactory import gen_string
 from robottelo.decorators import (
     tier2,
-    tier3,
     stubbed,
-    fixture
+    fixture,
+    upgrade
 )
+from robottelo.ui.utils import create_fake_host
+
+
+@fixture(scope='module')
+def module_org():
+    return entities.Organization().create()
 
 
 @fixture(scope='module')
 def module_loc():
     return entities.Location().create()
-
-
-@tier2
-@stubbed()
-def test_positive_create_report(session):
-    """ Create a report template
-
-    :id: 906ac1f8-c02c-4197-9c98-01e8b9f2f075
-
-    :setup: User with reporting access rights
-
-    :steps:
-
-        1. Monitor -> Report Templates -> Create Template
-        2. Insert any content
-        3. Submit
-
-    :expectedresults: The report template should be created
-
-    :CaseImportance: High
-    """
-
-
-@tier2
-@stubbed()
-def test_positive_read_report(session):
-    """ Create a report template
-
-    :id: 906ac1f8-c02c-4197-9c97-01e8b9f2f075
-
-    :setup: User with reporting access rights, some existing report template
-
-    :steps:
-
-        1. Monitor -> Report Templates
-        2. Select template
-
-    :expectedresults: The report template's content is shown
-
-    :CaseImportance: High
-    """
-
-
-@tier2
-@stubbed()
-def test_positive_update_report(session):
-    """ Edit a report template
-
-    :id: 906ac1f8-c02c-4297-9c98-01e8b9f2f075
-
-    :setup: User with reporting access rights, some existing report template that is not locked
-
-    :steps:
-
-        1. Monitor -> Report Templates
-        2. Select template
-        3. Change template content
-        4. Submit
-
-    :expectedresults: The report template should be updated
-
-    :CaseImportance: High
-    """
-
-
-@tier2
-@stubbed()
-def test_positive_delete_report(session):
-    """ Edit a report template
-
-    :id: 906ac1f8-c02c-4297-9c98-01f8b9f2f075
-
-    :setup: User with reporting access rights, some existing report template that is not locked
-
-    :steps:
-
-        1. Monitor -> Report Templates
-        2. In the reports row, in Actions column, click Delete
-
-    :expectedresults: The report template should be deleted
-
-    :CaseImportance: High
-    """
-
-
-@tier2
-@stubbed()
-def test_positive_lock_report(session):
-    """ Edit a report template
-
-    :id: 903ac1f8-c02c-4297-9c98-01e8f9f2f075
-
-    :setup: User with reporting access rights, some existing report template that is not locked
-
-    :steps:
-
-        1. Monitor -> Report Templates
-        2. In the reports row, in Actions column, click Lock
-
-    :expectedresults: The report template should be locked
-
-    :CaseImportance: Medium
-    """
-
-
-@tier2
-@stubbed()
-def test_positive_unlock_report(session):
-    """ Edit a report template
-
-    :id: 976ac1f8-c02c-4297-9c98-01e8b9f2f075
-
-    :setup: User with reporting access rights, some existing report template that is locked
-
-    :steps:
-
-        1. Monitor -> Report Templates
-        2. In the reports row, in Actions column, click Unlock
-
-    :expectedresults: The report template should be unlocked
-
-    :CaseImportance: Medium
-    """
-
-
-@tier2
-def test_positive_clone(session):
-    """Assure ability to clone a report template
-
-    :id: 912f1619-abb0-4e0f-88ce-88b5726fdbe0
-
-    :Steps:
-        1.  Go to Report template UI
-        2.  Choose a template and attempt to clone it
-
-    :expectedresults: The template is cloned
-
-    :CaseLevel: Integration
-    """
-    name = gen_string('alpha')
-    clone_name = gen_string('alpha')
-    content = gen_string('alpha')
-    with session:
-        session.reporttemplate.create({
-            'template.name': name,
-            'template.template_editor.editor': content,
-        })
-        session.reporttemplate.clone(
-            name,
-            {
-                'template.name': clone_name,
-            }
-        )
-        assert session.reporttemplate.search(
-            clone_name)[0]['Name'] == clone_name
-        assert session.reporttemplate.read(
-               clone_name)['template']['template_editor']['editor'] == content
-
-
-@tier2
-@stubbed()
-def test_positive_export_report(session):
-    """ Edit a report template
-
-    :id: 916ac1f8-c02c-4297-9c98-01e8b9f2f075
-
-    :setup: User with reporting access rights, some existing report template
-
-    :steps:
-
-        1. Monitor -> Report Templates
-        2. In the reports row, in Actions column, click Export
-
-    :expectedresults: A file should be downloaded with the report's script
-
-    :CaseImportance: Medium
-    """
-
-
-@tier2
-@stubbed()
-def test_positive_generate_report_hoststatus(session):
-    """ Edit a report template
-
-    :id: 916fc1f8-c02c-4297-9c98-01e8b9f2f075
-
-    :setup: User with reporting access rights, existing Hosts Status report template,
-            at least two hosts
-
-    :steps:
-
-        1. Monitor -> Report Templates
-        2. In the reports row, in Actions column, click Generate
-        3. Submit
-
-    :expectedresults: A report should be generated with a list of hosts and their statuses
-
-    :CaseImportance: High
-    """
-
-
-@tier2
-@stubbed()
-def test_positive_generate_report_hoststatus_filter(session):
-    """ Edit a report template
-
-    :id: 916ec1f8-c02c-4297-9c98-01e8b9f2f075
-
-    :setup: User with reporting access rights, existing Hosts Status report template,
-            at least two hosts
-
-    :steps:
-
-        1. Monitor -> Report Templates
-        2. In the reports row, in Actions column, click Generate
-        3. Fill in such a filter that it only matches 1 host
-        4. Submit
-
-    :expectedresults: A report should be generated with a filtered hosts and its status
-
-    :CaseImportance: High
-    """
-
-
-@tier2
-@stubbed()
-def test_positive_generate_report_hoststatus_sanitized(session):
-    """ Edit a report template
-
-    :id: 916ed1f8-c02c-4297-9c98-01e8b9f2f075
-
-    :setup: User with reporting access rights, existing Hosts Status report template,
-            OS with comma in its label, at least two hosts with this OS
-
-    :steps:
-
-        1. Monitor -> Report Templates
-        2. In the reports row, in Actions column, click Generate
-        3. Submit
-
-    :expectedresults: A report should be generated with a list of hosts and their statuses
-
-    :CaseImportance: High
-    """
 
 
 @tier2
@@ -338,33 +103,8 @@ def test_positive_preview_report(session):
     """
 
 
-@tier3
-@stubbed()
-def test_positive_report_e2e(session):
-    """ Report templates end-to-end scenario
-
-    :id: cd19b90d-836f-4efd-8cbc-d5e09a908e67
-
-    :setup: Admin user, more than one host registered
-
-    :steps:
-
-        1. Admin creates a user with appropriate permissions, that user logs in
-        2. Creates new report template with host_search input that would e.g. print just host names
-        3. Generate the report out of it by specifying the valid host search syntax
-        4. The report contains expected data
-        5. User deletes the report template
-
-    :expectedresults: All steps succeeded. Generated report contains correct data.
-                      User can see this activity logged in audit
-                      page (besides generating report as it doesn't change DB state)
-
-    :CaseImportance: Medium
-    """
-
-
 @tier2
-def test_positive_e2e_crud(session, module_org, module_loc):
+def test_positive_end_to_end(session, module_org, module_loc):
     """Perform end to end testing for report template component's CRUD operations
 
     :id: b44d4cc8-a78e-47cf-9993-0bb871ac2c96
@@ -378,6 +118,7 @@ def test_positive_e2e_crud(session, module_org, module_loc):
     name = gen_string('alpha')
     content = gen_string('alpha')
     new_name = gen_string('alpha')
+    clone_name = gen_string('alpha')
     input_name = gen_string('alpha')
     template_input = [{
         'name': input_name,
@@ -386,6 +127,7 @@ def test_positive_e2e_crud(session, module_org, module_loc):
         'input_content.description': gen_string('alpha')
     }]
     with session:
+        # CREATE report template
         session.reporttemplate.create({
             'template.name': name,
             'template.default': True,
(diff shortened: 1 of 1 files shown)


oshtaier:
It is necessary to unite all these test cases (create, read, update) into one end_to_end scenario as it is required by our approach for WebUI tests

lhellebr:
There is an e2e test already.

lhellebr:
These are test cases agreed upon between devel and QE in the reporting feature discussion.

oshtaier:
Sure, but these test cases go against our test strategy that we defined here: https://github.com/SatelliteQE/robottelo/issues/6408. We must implement only one test case for crud operations in terms of new WebUI test suite, so it doesn't matter who agreed these test cases. It is direct code duplication and test functionality on component level which we are not doing in test automation anymore.

We can escalate that question to Matt or whoever you like to look on that

lhellebr:
OK.
Is it ok to make some tests (e.g. negative tests) as separate tests, or should **everything** be included in one single e2e test?

lhellebr:
Also depends on https://github.com/SatelliteQE/airgun/pull/322

lhellebr:
Also depends on https://github.com/SatelliteQE/airgun/pull/323

lhellebr:
Also depends on https://github.com/SatelliteQE/airgun/pull/324

lhellebr:
I think we can merge this and all the PRs this depends on, please review.

tests/foreman/ui_airgun/test_reporttemplates.py
```diff
@@ -407,12 +191,104 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+    :expectedresults: All subtests pass
+
+    :CaseLevel: Integration
+
+    :CaseImportance: High
+    """
+    # generate Host Status report
+    os_name = 'comma,' + gen_string('alpha')
+    os = entities.OperatingSystem(name=os_name).create().read()
+    host_template = entities.Host(organization=module_org,
```
oshtaier: can you use loops and lists?

tests/foreman/ui_airgun/test_reporttemplates.py
```diff
@@ -407,12 +191,104 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+    host_template.create_missing()
+    host_template2 = entities.Host(organization=module_org,
+                                   location=module_loc, operatingsystem=os)
+    host_template2.create_missing()
+    host_template3 = entities.Host(organization=module_org,
+                                   location=module_loc, operatingsystem=os)
+    host_template3.create_missing()
+    with session:
+        # create multiple hosts to test filtering
+        create_fake_host(
```
oshtaier: same here with loops

tests/foreman/ui_airgun/test_reporttemplates.py
```diff
@@ -407,12 +191,104 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+        )
+        host_name = create_fake_host(
+            session,
+            host_template,
+        )
+        create_fake_host(
+            session,
+            host_template3,
+        )
+        import pdb
```
oshtaier: please, remove

tests/foreman/ui_airgun/test_reporttemplates.py
```diff
@@ -407,12 +191,104 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+            columns = next(reader)
+            assert columns == ['Name', 'Ip', 'Operating System',
+                               'Subscriptions', 'Applicable Errata', 'Owner',
+                               'Kernel', 'Latest kernel available']
+            csvfile.seek(0)
+            dreader = csv.DictReader(csvfile)
+            res = next(dreader)
+            assert res['Name'] == host_name
+            # also tests comma in field contents
+            assert res['Operating System'] == os_name + " " + os.major
```
oshtaier: please use `.format` for string concatenation

tests/foreman/ui_airgun/test_reporttemplates.py
```diff
@@ -407,12 +191,104 @@ def test_positive_e2e_crud(session, module_org, module_loc):
         assert pt['type']['snippet'] is False
         assert pt['locations']['resources']['assigned'][0] == module_loc.name
         assert pt['organizations']['resources']['assigned'][0] == module_org.name
+        # U
```
oshtaier: ?

tests/foreman/ui_airgun/test_reporttemplates.py
```diff
@@ -396,7 +172,15 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
-        assert session.reporttemplate.search(name)[0]['Name'] == name
+        # also check other properties by API once BZ1702676 gets fixed
+        rt = entities.ReportTemplate().search(
+                query={'search': 'name={0}'.format(name)})[0].read()
+        assert rt.name == name
+        assert rt.organization[0].read().name == module_org.name
+        assert rt.location[0].read().name == module_loc.name
+        assert rt.template == content
+        assert rt.default is True
+        # R
```
oshtaier: ?

tests/foreman/ui_airgun/test_reporttemplates.py
```diff
@@ -386,6 +161,7 @@ def test_positive_e2e_crud(session, module_org, module_loc):
         'input_content.description': gen_string('alpha')
     }]
     with session:
+        # C
```
oshtaier: you kidding, right?

tests/foreman/ui_airgun/test_reporttemplates.py
```diff
@@ -407,12 +191,104 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+            host_template,
+        )
+        create_fake_host(
+            session,
+            host_template3,
+        )
+        import pdb
+        pdb.set_trace()
+        file_path = session.reporttemplate.generate("Registered hosts",
+                                                    values={'inputs': {'Hosts filter': host_name}})
```
mirzal: As list of hosts to include in report is filtered, this implements stubbed test `test_positive_generate_report_hosts_filter`, right? If it does, please remove this stub as well.

tests/foreman/ui_airgun/test_reporttemplates.py
```diff
@@ -396,7 +172,15 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+        # also check other properties by API once BZ1702676 gets fixed
+        rt = entities.ReportTemplate().search(
+                query={'search': 'name={0}'.format(name)})[0].read()
+        assert rt.name == name
+        assert rt.organization[0].read().name == module_org.name
+        assert rt.location[0].read().name == module_loc.name
+        assert rt.template == content
+        assert rt.default is True
+        # R
         pt = session.reporttemplate.read(name)
```
mirzal: Do you have a reason to keep `rt` around and introduce new variable?
I can see how `rt` stands for report template, but name `pt` is not very descriptive.

Same goes for `updated_pt` and `crt` below.

tests/foreman/ui_airgun/test_reporttemplates.py
```diff
@@ -407,12 +191,104 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+        assert crt.organization[0].read().name == module_org.name
+        assert crt.location[0].read().name == module_loc.name
+        assert crt.template == content
+        assert crt.default is True
+        # export
+        file_path = session.reporttemplate.export(new_name)
+        assert os.path.isfile(file_path)
+        with open(file_path) as dfile:
+            dfile_str = dfile.read()
+            assert content in dfile_str
```
mirzal: Maybe just `assert content in dfile.read()`? Or rename `dfile_str` to `dfile_content`.

tests/foreman/ui_airgun/test_reporttemplates.py
```diff
@@ -407,12 +191,104 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+        import pdb
+        pdb.set_trace()
+        file_path = session.reporttemplate.generate("Registered hosts",
+                                                    values={'inputs': {'Hosts filter': host_name}})
+        with open(file_path) as csvfile:
+            reader = csv.reader(csvfile)
+            columns = next(reader)
+            assert columns == ['Name', 'Ip', 'Operating System',
+                               'Subscriptions', 'Applicable Errata', 'Owner',
+                               'Kernel', 'Latest kernel available']
```
mirzal: I'm not 100% sure, but I think you can start with `csv.DictReader()`, assert number of items in list and then, for first item, assert that this dict has each of these column names. That would save `csvfile.seek()` call and duplicated invocation of `csv.reader()`.

tests/foreman/ui_airgun/test_reporttemplates.py
lhellebr: No. "C" for first part of CRUD. I can expand it to the full form if you think that really matters.

oshtaier: Yeah, please expand with 'create xxx (whatever you create)'. Man, please don't think that I want to offend your feelings, but we are working here on serious commercial product and that doesn't sound acceptable

lhellebr:
I think issues were addressed (thanks for pointing them out!) in this PR and other PRs this depends on. Please, continue with the review process.

lhellebr:
```
$ pytest test_reporttemplates.py
========================================= test session starts ==========================================
platform linux -- Python 3.6.8, pytest-3.6.1, py-1.7.0, pluggy-0.6.0
rootdir: /home/lhellebr/git/robottelo, inifile:
plugins: services-1.3.0, mock-1.10.0
collecting 5 items                                                                                     2019-05-23 15:39:58 - conftest - DEBUG - Collected 5 test cases

collected 5 items                                                                                      

test_reporttemplates.py sss..                                                                    [100%]

================================ 2 passed, 3 skipped in 527.09 seconds =================================
```

tests/foreman/ui_airgun/test_reporttemplates.py
```diff
@@ -396,23 +149,111 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+            )
+        host_name = host_names[1]  # pick some that is not first and is not last
+        file_path = session.reporttemplate.generate("Registered hosts",
+                                                    values={'inputs': {'Hosts filter': host_name}})
+        with open(file_path) as csvfile:
+            dreader = csv.DictReader(csvfile)
+            res = next(dreader)
+            assert list(res.keys()) == ['Name', 'Ip', 'Operating System',
+                                        'Subscriptions', 'Applicable Errata', 'Owner',
+                                        'Kernel', 'Latest kernel available']
```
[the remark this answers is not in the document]
lhellebr: In DictReader, the entries are of type OrderedDict so I think it's OK.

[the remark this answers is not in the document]
mirzal: @lhellebr that settles it, then.

lhellebr:
@oshtaier Can we get this merged?

lhellebr:
All the PRs this depends on have been merged. This can be merged, too.

lhellebr:
Can we finally get this merged?
@oshtaier @rplevka @san7ket @ntkathole

ntkathole:
@lhellebr 
Can you please update
1. casecomponent as per @mirzal comment to Reporting.
2. Feel free to add upgrade decorator on one of the test covering e2e functionality.
3. Small nitpick, import modules in order.

Once you updated it, will merge.

lhellebr:
@ntkathole 
1) Overwhelming majority of UI tests have CaseComponent set to "UI" regardless on what entities they test. For the sake of consistency, I think this should be left unchanged.
2) Done
3) Done

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -15,264 +15,40 @@
...
+@fixture(scope='module')
+def module_loc():
+    return entities.Location().create()
 
-    :expectedresults: A report should be generated with a list of hosts and their statuses
 
-    :CaseImportance: High
-    """
+@fixture(scope='module')
+def module_reporttemplate(module_org, module_loc):
```
oshtaier: It seems you do not need that method anymore

ntkathole:
@lhellebr Yes, this is recent change we are doing where we are updating metadata of tests. But I would not block this PR. We can have separate PR for updating metadata by component owner.

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -396,23 +149,112 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
-        assert pt['template']['default'] is True
-        assert pt['template']['template_editor']['editor'] == content
-        assert pt['inputs'][0]['name'] == input_name
-        assert pt['inputs'][0]['required'] is True
-        assert pt['inputs'][0]['input_type'] == 'User input'
-        assert pt['type']['snippet'] is False
-        assert pt['locations']['resources']['assigned'][0] == module_loc.name
-        assert pt['organizations']['resources']['assigned'][0] == module_org.name
+        # also check other properties by API once BZ1702676 gets fixed
+        rt = entities.ReportTemplate().search(
```
oshtaier: BZ is already ON_QA, anyway I am not sure why did you do search using API and then read through UI (which doing search through UI). That just doesn't make any sense

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -396,23 +149,112 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+        assert rt['inputs'][0]['input_type'] == 'User input'
+        assert rt['type']['snippet'] is False
+        assert rt['locations']['resources']['assigned'][0] == module_loc.name
+        assert rt['organizations']['resources']['assigned'][0] == module_org.name
+        # UPDATE report template
         session.reporttemplate.update(
             name,
             {'template.name': new_name, 'type.snippet': True}
         )
+        # check this by API once BZ1702676 gets fixed
```
oshtaier: seems that entire comment can be removed

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -396,23 +149,112 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+        # assert entities.ReportTemplate(name=new_name).search(
+        #    query={'search': 'name={0}'.format(new_name)})[0].read().snippet is True
         assert session.reporttemplate.search(new_name)[0]['Name'] == new_name
-        updated_pt = session.reporttemplate.read(new_name)
-        assert updated_pt['type']['snippet'] is True
+        rt = session.reporttemplate.read(new_name)
+        assert rt['type']['snippet'] is True
+        # LOCK
+        session.reporttemplate.lock(new_name)
+        # check this by API once BZ1702676 gets fixed
```
oshtaier: same that one

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -396,23 +149,112 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+        # assert rt.locked is True
+        assert session.reporttemplate.is_locked(new_name) is True
+        # UNLOCK
+        session.reporttemplate.unlock(new_name)
+        # check this by API once BZ1702676 gets fixed
```
oshtaier: and that

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -396,23 +149,112 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+        assert rt.default is True
+        # EXPORT
+        file_path = session.reporttemplate.export(new_name)
+        assert os.path.isfile(file_path)
+        with open(file_path) as dfile:
+            assert content in dfile.read()
+        # DELETE report template
         session.reporttemplate.delete(new_name)
-        assert not session.reporttemplate.search(new_name)
+        assert len(entities.ReportTemplate().search(
```
oshtaier: once more, if read is working then search is also working

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -396,23 +149,112 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+
+
+@upgrade
+@tier2
+def test_positive_generate_registered_hosts_report(session, module_org, module_loc):
+    """Use provided Registered Hosts report for testing
+
+    :id: b44d4cd8-a78e-47cf-9993-0bb871ac2c96
+
+    :expectedresults: All subtests pass
```
oshtaier: what subtests? what is goal of test case?

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -396,23 +149,112 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+
+    :CaseLevel: Integration
+
+    :CaseImportance: High
+    """
+    # generate Host Status report
+    os_name = 'comma,' + gen_string('alpha')
+    os = entities.OperatingSystem(name=os_name).create().read()
```
oshtaier: `create()`  is doing `read()` automatically

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -396,23 +138,112 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+    host_cnt = 3
+    host_templates = []
+    for i in range(host_cnt):
```
oshtaier: I feel
```
    host_cnt = 3
    host_templates = []
    for i in range(host_cnt):
        host_templates.append(entities.Host(organization=module_org,
                                            location=module_loc, operatingsystem=os)
                              )
        host_templates[i].create_missing()
```
can be improved using simple list comprehension

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -396,23 +138,112 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+        host_templates.append(entities.Host(organization=module_org,
+                                            location=module_loc, operatingsystem=os)
+                              )
+        host_templates[i].create_missing()
+    with session:
+        # create multiple hosts to test filtering
+        host_names = []
```
oshtaier: same here:
```
host_names = [create_fake_host(session, host_template) for host_template in host_templates]
```
or something like this

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -338,33 +103,8 @@ def test_positive_preview_report(session):
...
-                      User can see this activity logged in audit
-                      page (besides generating report as it doesn't change DB state)
-
-    :CaseImportance: Medium
-    """
-
-
 @tier2
-def test_positive_e2e_crud(session, module_org, module_loc):
+def test_positive_e2e(session, module_org, module_loc):
```
oshtaier: `test_positive_end_to_end`

tests/foreman/ui/test_reporttemplates.py
lhellebr: This tests that the report template is deleted

lhellebr:
@oshtaier comments addressed

tests/foreman/ui/test_reporttemplates.py
oshtaier: right, so verify it through UI. I can't imagine scenario where API validation is more preferable

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -396,23 +138,91 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
+        session.reporttemplate.unlock(new_name)
+        assert session.reporttemplate.is_locked(new_name) is False
+        # CLONE
+        session.reporttemplate.clone(
+            new_name,
+            {
+                'template.name': clone_name,
+            }
+        )
+        rt = entities.ReportTemplate().search(
```
oshtaier: same, why do you use API there?

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -396,23 +138,91 @@ def test_positive_e2e_crud(session, module_org, module_loc):
...
-        assert pt['inputs'][0]['input_type'] == 'User input'
-        assert pt['type']['snippet'] is False
-        assert pt['locations']['resources']['assigned'][0] == module_loc.name
-        assert pt['organizations']['resources']['assigned'][0] == module_org.name
+        # READ report template
+        rt = session.reporttemplate.read(name)
+        assert rt['template']['name'] == name
+        assert rt['template']['default'] is True
+        assert rt['template']['template_editor']['editor'] == content
+        assert rt['inputs'][0]['name'] == input_name
```
oshtaier: it is nitpick, but you can use dict values instead:
assert rt['inputs'][0]['name'] == template_input['name']

tests/foreman/ui/test_reporttemplates.py
lhellebr: Whenever there is discrepancy between API and WebUI. Testing through API can potentially find more bugs and is also more efficient.

oshtaier: Right, and that is a reason we have three different folders per each type. Just to test API in API folder, CLI in CLI folder and UI in UI folder. In that PR we have UI folder, so test through UI

lhellebr: In this part, I am testing WebUI by running tasks in WebUI and checking the results through API. That doesn't mean I am testing API, that means I am testing WebUI and I am not relying on the WebUI while doing so (which makes sense since I am **testing** it).

lhellebr: In this part, I am testing WebUI by running tasks in WebUI and checking the results through API. That doesn't mean I am testing API, that means I am testing WebUI and I am not relying on the WebUI while doing so (which makes sense since I am **testing** it).

jhutar: API is far faster. Only risk using API instead of UI to verify the entity was deleted is if UI would still be showing the entity after it was deleted. I do not recall any issue like that, so IMO speed benefit overweights this risk.

oshtaier: hmm. The whole idea of test automation is to verify result through same interface it was created.

Example:
You created through UI, deleted through UI, validated successfully through API (result showed entity is not present in application), but UI is still showing entity. So your test proved nothing.

We never, ever validate through different interfaces. Yes, we use API to speed up pre-configuration procedures, but attempt to use it for verifying results ruin the whole ideology of test automation. I can't even explain how many fundamental principles of quality assurance is violated by proposed idea.

Let's not waste time anymore on that pointless discussion

lhellebr: Imagine a simple example, where there are two databases - WebUI only uses one and API only uses the second, so whatever you do in WebUI is never visible through API. Which is obviously wrong. Yet, using WebUI to verify WebUI action results would never catch this.

tests/foreman/ui/test_reporttemplates.py
jhutar: I think the main disagreement here is @lhellebr says entity's **read** via UI is tested because it is done in `# READ report template` section and do not need to be tested again. On the other hand, @oshtaier says that **every entity** change done on in UI test needs to be done via UI - e.g. cloned entity might be unreadable via UI. Another Lukas's argument is he wants to verify change via another means that what caused the change to make sure (when thinking about it "ad absurdum"), UI might clone the entity only in data backend used by UI and not by API.
In ideal case, we would be checking with both UI and API, but that would be very expensive, so reasonable mix of these two ways seems to be a best solution to me.

tests/foreman/ui/test_reporttemplates.py
jhutar: > Let's not waste time anymore on that pointless discussion

OK, lets ack the PR then :-)

tests/foreman/ui/test_reporttemplates.py
oshtaier: @jhutar 
> In ideal case, we would be checking with both UI and API
We do not even both ways, but all three (including CLI), so we test *every entity* by *API, CLI, UI*, so I have no idea why we mix ways when we still validate same thing in API

tests/foreman/ui/test_reporttemplates.py
oshtaier: Not a chance =)

oshtaier: interesting example, but we do not test that frequently. For such scenario, we always create separate, independent test case and usually for existing issue as it will be too redundant to add so many test cases.

oshtaier: If you feel that it is must have validation for you, then validate whether both UI and API or only UI. I can't ack only API validation

lhellebr:
I nerfed the PR as requested. After 2 months, I don't really care enough already. Merge please.

lhellebr:
@oshtaier

lhellebr:
@oshtaier , this is taking **way** too long. Merge please.

PR #6888, opened 2019-05-02
Fix: sha256 repository create fails for on-demand repo

jyejare:
Fix: sha256 repository create fails for on-demand repo


san7ket:
I think we can merge, as just the download policy was changed , no  logic changes

omkarkhatavkar:
@san7ket ACK

PR #6901, opened 2019-05-08
Ostree repos1

### LINKED ISSUE #6770: Some tests syncing Fedora Atomic OSTree failing because repo no longer exists
We have a fair number of tests failing because they are trying to sync one of the 2 repos

```
FEDORA26_OSTREE_REPO = u'https://kojipkgs.fedoraproject.org/atomic/26/'
FEDORA27_OSTREE_REPO = u'https://kojipkgs.fedoraproject.org/atomic/27/'
```

Some of them seem to be syncing it just cause they need any ostree repo to sync, so we should try to find another repo to use for these tests.
@JacobCallahan Suggests we can use 
```
https://repos.fedorapeople.org/pulp/pulp/fixtures/ostree/small/
```
We still need a second one. 

Still looking into this, but if there are any test that need that repo specifically, we need to find out what the replacement is.

@latran can you add a comment with the test you see failing because of this. I will do the same.

### LINKED ISSUE #6772: OstreeContentViewTestCase isn't actually syncing the OSTree repo
Noticed this while looking into #6770 , looks like there was a typo where we are calling `cls.ostree_repo.sync` instead of `cls.ostree_repo.sync()` so the sync is never run. 

https://github.com/SatelliteQE/robottelo/blob/f617f6b4c40d5868105a579c0f240374a4e5b93a/tests/foreman/api/test_contentview.py#L1588

pgagne:
- Fixes #6770 Updated Fedora OSTree repos to point to an archived version of the repos since we need 2 repo URLs.
- Fixes #6772 added () to  cls.ostree_repo.sync

Added skip due to https://bugzilla.redhat.com/show_bug.cgi?id=1696988

```
(venv) [user@host robottelo]$ pytest -v tests/foreman/api/test_repository.py::OstreeRepositoryTestCase::test_positive_update_url
========================================================================================================================================================================== test session starts ===========================================================================================================================================================================
platform linux -- Python 3.6.8, pytest-4.0.2, py-1.8.0, pluggy-0.11.0 -- /redacted
cachedir: .pytest_cache
rootdir: //redacted/robottelo, inifile:
plugins: services-1.3.1, mock-1.10.0
collecting ... 2019-05-08 11:14:40 - conftest - DEBUG - Collected 1 test cases

collected 1 item

tests/foreman/api/test_repository.py::OstreeRepositoryTestCase::test_positive_update_url PASSED
```

### DIFF
--- a/tests/foreman/api/test_contentview.py
+++ b/tests/foreman/api/test_contentview.py
@@ -1595,6 +1595,7 @@ class ContentViewRolesTestCase(APITestCase):
         with self.assertRaises(HTTPError):
             entities.ContentView(id=content_view.id).read()
 
+
 @skip_if_bug_open('bugzilla', 1625783)
 class OstreeContentViewTestCase(APITestCase):
     """Tests for ostree contents in content views."""
--- a/tests/foreman/api/test_contentview.py
+++ b/tests/foreman/api/test_contentview.py
@@ -1612,7 +1612,7 @@ class OstreeContentViewTestCase(APITestCase):
             url=FEDORA27_OSTREE_REPO,
             unprotected=False
         ).create()
-        cls.ostree_repo.sync
+        cls.ostree_repo.sync()
         # Create new yum repository
         cls.yum_repo = entities.Repository(
             url=FAKE_1_YUM_REPO,
--- a/tests/foreman/api/test_contentview.py
+++ b/tests/foreman/api/test_contentview.py
@@ -1595,7 +1595,7 @@ class ContentViewRolesTestCase(APITestCase):
         with self.assertRaises(HTTPError):
             entities.ContentView(id=content_view.id).read()
 
-
+@skip_if_bug_open('bugzilla', 1625783)
 class OstreeContentViewTestCase(APITestCase):
     """Tests for ostree contents in content views."""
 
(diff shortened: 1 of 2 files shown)


tests/foreman/api/test_contentview.py
```diff
@@ -1596,6 +1596,7 @@ def test_negative_non_readonly_user_actions(self):
             entities.ContentView(id=content_view.id).read()
 
 
+@skip_if_bug_open('bugzilla', 1625783)
```
vijay8451: Having a query on it ..
As Bugzilla `1625783` is in verified state and afaik related tests would execute if any Bug get verified ?

pgagne: https://bugzilla.redhat.com/show_bug.cgi?id=1625783

Is in the NEW state.

vijay8451: thanks , I was referring the `1696988` .. 
Looks good now :)

vijay8451:
I just noticed that as per test result it ran for API tests , However CLI and UI also depended on same both repos ..
Could you please give a try to those as well ?

pgagne:
@vijay8451 I actually tried running those, but couldn't get them to run due to other known issues with  OSTree syncing (unrelated to the new repos). I actually have a separate task I am working on to address those.

vijay8451:
@pgagne thanks

vijay8451:
@pgagne it looks require to cherry-pick for 6.5 too as master is 6.6 now :)

PR #6908, opened 2019-05-09
Environment Variables for upgrade scenarios are now robottelo properties - Phase 1

jyejare:
Changes:

1. New Upgrade section in robottelo.properties
2. UpgradeClass is updated with upgrade proeprties in base.py
3. [ Phase 1] Only capsule and errata upgrade tests are updated to use settings from robottelo.properties instead of env vars are updated 

The rollout of this change will happen as part of phase 2 and on all the tests passed in phase 1.

PR #6947, opened 2019-05-22
Environment Variables for upgrade scenarios are now robottelo properties Phase -1

jyejare:
cherrypick - https://github.com/SatelliteQE/robottelo/pull/6908

### DIFF
--- a/robottelo.properties.sample
+++ b/robottelo.properties.sample
@@ -333,10 +333,24 @@ ssh_key=
 # "/var/lib/libvirt/images/".
 # libvirt_image_dir=/var/lib/libvirt/images
 
-# Section for PostUpgrade Verification Tests
-# [upgrade]
-# Option for providing a preUpgrade yaml data file location on server
-# upgrade_data=
+# Section for Upgrade Scenarios Verification Tests
+[upgrade]
+# RHEVM capsule hostname
+# rhev_cap_host=
+# Custom capsule hostname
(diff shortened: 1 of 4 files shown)


robottelo.properties.sample
```diff
@@ -333,10 +333,24 @@ ssh_key=
...
+# RHEVM capsule hostname
+# rhev_cap_host=
+# Custom capsule hostname
+# capsule_hostname=
+# RHEVM capsule activation key
+# rhev_capsule_ak=
+# Custom capsule activation key
+# capsule_ak=
+# The satellite end version
+# to_version=
```
vijay8451:
```suggestion
# to_version=6.5
```

jyejare: No, We dont want to hardcode as this will set from CI.

Because while upgrading from 6.5 to 6.6, this hardcoded value of 6.5 for TO_VERSION will have complications.

We have it in master though as 6.6 and there is no harm having it there.

jyejare: This was wantedky been removed.

vijay8451: Just a query .. Once we cut the branch for 6.6 then this will remain 6.6 and have to update/remove ? Is there anyway if could keep without version in master too ?

vijay8451: As  discussed you shall raise another PR to remove from master branch

PR #6963, opened 2019-05-27
Remove org/loc IDs dependency

### LINKED ISSUE #6937: Remove the tests dependency from DEFAULT_LOC_ID and DEFAULT_ORG_ID
Remove the tests dependency from DEFAULT_LOC_ID and DEFAULT_ORG_ID

vijay8451:
1. Fixed #6937
2. Recently hit bug[https://bugzilla.redhat.com/show_bug.cgi?id=1713269] due to which automation failed because of hardcoded org/loc IDs.

### DIFF
--- a/robottelo/cleanup.py
+++ b/robottelo/cleanup.py
@@ -5,7 +5,7 @@ from collections import deque, defaultdict
 from nailgun import entities, signals
 from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.proxy import Proxy
-from robottelo.constants import DEFAULT_ORG_ID
+from robottelo.constants import DEFAULT_ORG
 from robottelo.decorators import bz_bug_is_open
 
 
@@ -85,7 +85,9 @@ class EntitiesCleaner(object):
 
     def clean(self):
         """This method is called in TearDownClass only when cleanup=true"""
-        default_org = entities.Organization(id=DEFAULT_ORG_ID)
+        default_org_id = entities.Organization().search(
+            query={'search': 'name="{}"'.format(DEFAULT_ORG)})[0].id
+        default_org = entities.Organization(id=default_org_id)
         # reassign created hosts to default org
         self.update_entities(
             self.cleanup_queue.get(entities.Host.__name__, []),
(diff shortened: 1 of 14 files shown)


tests/foreman/rhai/conftest.py
```diff
@@ -18,7 +18,9 @@ def module_org():
 
     :rtype: :class:`nailgun.entities.Organization`
     """
-    return nailgun.entities.Organization(id=DEFAULT_ORG_ID).read()
+    default_org_id = nailgun.entities.Organization().search(
+        query={'search': 'name="{}"'.format(DEFAULT_ORG)})[0].i
```
omkarkhatavkar: you missed 'd' here 

```suggestion
        query={'search': 'name="{}"'.format(DEFAULT_ORG)})[0].id
```

robottelo/cleanup.py
```diff
@@ -85,7 +85,9 @@ def register_entity_for_cleanup(self, sender, entity, **kwargs):
 
     def clean(self):
         """This method is called in TearDownClass only when cleanup=true"""
-        default_org = entities.Organization(id=DEFAULT_ORG_ID)
+        default_org_id = entities.Organization().search(
+            query={'search': 'name="{}"'.format(DEFAULT_ORG)})[0].id
```
omkarkhatavkar: I thought if we used this line across multiple times, we can create a common function. e.g. 

default_org or default_org_id 

```python 
def get_default_org:
        return entities.Organization().search(query={'search': 'name="{}"'.format(DEFAULT_ORG)})[0].id
```
what do you think ?

tests/foreman/rhai/conftest.py
vijay8451: Ahh ... It was a TYPO :) ... pushed in new commit .

robottelo/cleanup.py
vijay8451: your point is valid .. I checked on that but somehow did not find any direct way where it could use directly call by all place where this dependency requires to remove hence wannted to go with same.

vijay8451:
As it's same call which is using/replacing on multiple file so below test result as sample:
```
tests/foreman/ui_airgun/test_computeresource.py::test_positive_update_organization[True] PASSED                                                                  [ 100%]
tests/foreman/ui_airgun/test_jobinvocation.py::test_positive_run_custom_job_template_by_ip ERROR                                                                 [100%]    <<== This test was fail with different reason not by this.
```

PR #6965, opened 2019-05-28
updated matadata for Settings component

omkarkhatavkar:
### PR Objective
Update the metadata for the component  Settings and validate against the testimony

### Testimony Result 
```
(env) [root@okhatavk robottelo]# check_testimony 
Total number of tests: 8
Total number of invalid docstrings: 0 (00.00%)
Test cases with no docstrings: 0 (0.00%)
Test cases missing minimal docstrings: 0 (0.00%)
Test cases with unexpected tags: 0 (0.00%)
Test cases with unexpected token values in docstrings: 0 (0.00%)
Test cases with unparseable docstrings: 0 (0.00%)
Total number of tests: 19
Total number of invalid docstrings: 0 (00.00%)
Test cases with no docstrings: 0 (0.00%)
Test cases missing minimal docstrings: 0 (0.00%)
Test cases with unexpected tags: 0 (0.00%)
Test cases with unexpected token values in docstrings: 0 (0.00%)
Test cases with unparseable docstrings: 0 (0.00%)  
```
### Tier Validation Result 
```
(env) [root@okhatavk robottelo]# files=`git diff --name-only | grep -E .py `; for file in $files; do testimony -c /home/okhatavk/Satellite/robottelo/testimony.yaml tiers $file; done
(env) [root@okhatavk robottelo]# 
```

### DIFF
--- a/tests/foreman/api/test_setting.py
+++ b/tests/foreman/api/test_setting.py
@@ -1,7 +1,7 @@
 # -*- encoding: utf-8 -*-
 """Test class for Setting Parameter values
 
-:Requirement: Test Cases for component Settings
+:Requirement: Settings
 
 :CaseAutomation: Automated
 
--- a/tests/foreman/api/test_setting.py
+++ b/tests/foreman/api/test_setting.py
@@ -1,13 +1,13 @@
 # -*- encoding: utf-8 -*-
 """Test class for Setting Parameter values
 
-:Requirement: Setting
+:Requirement: Test Cases for component Settings
 
 :CaseAutomation: Automated
 
 :CaseLevel: Acceptance
 
-:CaseComponent: API
+:CaseComponent: Settings
 
 :TestType: Functional
 
@@ -18,7 +18,13 @@
 from nailgun import entities
 from robottelo.cleanup import setting_cleanup
 from robottelo.datafactory import generate_strings_list, valid_data_list
-from robottelo.decorators import run_in_one_thread, stubbed, tier1, upgrade
+from robottelo.decorators import (
+    stubbed,
+    run_in_one_thread,
+    tier1,
+    tier2,
+    upgrade
+)
 from robottelo.test import APITestCase
 
 
@@ -47,7 +53,7 @@ class SettingTestCase(APITestCase):
                 self.assertEqual(login.value, login_text)
 
     @run_in_one_thread
-    @tier1
+    @tier2
     def test_positive_update_login_page_footer_text_without_value(self):
         """Updates parameter "login_text" without any string (empty value)
 
@@ -65,7 +71,7 @@ class SettingTestCase(APITestCase):
         self.assertEqual(login.value, "")
 
     @run_in_one_thread
-    @tier1
+    @tier2
     def test_positive_update_login_page_footer_text_with_long_string(self):
         """Attempt to update parameter "Login_page_footer_text"
             with long length string
@@ -86,7 +92,7 @@ class SettingTestCase(APITestCase):
                 self.assertEqual(login.value, login_text)
 
     @stubbed()
-    @tier1
+    @tier2
     def test_negative_update_hostname_with_empty_fact(self):
         """Update the Hostname_facts settings without any string(empty values)
 
@@ -98,7 +104,7 @@ class SettingTestCase(APITestCase):
         :CaseAutomation: notautomated
         """
 
-    @tier1
+    @tier2
     def test_positive_update_hostname_prefix_without_value(self):
         """Update the Hostname_prefix settings without any string(empty values)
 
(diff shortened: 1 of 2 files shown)


tests/foreman/api/test_setting.py
```diff
@@ -1,13 +1,13 @@
 # -*- encoding: utf-8 -*-
 """Test class for Setting Parameter values
 
-:Requirement: Setting
+:Requirement: Test Cases for component Settings
```
vijay8451:
```suggestion
:Requirement: Setting
```
Please keep as above ..

tests/foreman/api/test_setting.py
```diff
@@ -1,13 +1,13 @@
...
 
 :CaseAutomation: Automated
 
 :CaseLevel: Acceptance
 
-:CaseComponent: API
+:CaseComponent: Settings
```
vijay8451:
```suggestion
:CaseComponent: API
```
AFAIK this is still under discussion and we could keep as it is ..

tests/foreman/cli/test_setting.py
```diff
@@ -1,13 +1,13 @@
 # -*- encoding: utf-8 -*-
 """Test class for Setting Parameter values
 
-:Requirement: Setting
+:Requirement: Test Cases for component Settings
```
vijay8451: same as above

tests/foreman/cli/test_setting.py
```diff
@@ -1,13 +1,13 @@
...
 
 :CaseAutomation: Automated
 
 :CaseLevel: Acceptance
 
-:CaseComponent: CLI
+:CaseComponent: Settings
```
vijay8451: same as above

tests/foreman/api/test_setting.py
mirzal: Aligning CaseComponent with bugzilla components has one unintended consequence that was discovered recently and is being discussed, but I highly doubt it will stop the whole effort - we are much more likely to introduce another field to capture data we are missing, or just ignore the issue. Our management was very clear that they want to use CaseComponent for finding component owner, and it means that "API" value is usually not what we want.

vijay8451: @mirzal  thanks to clear on then it's good to keep the new change .

tests/foreman/api/test_setting.py
```diff
@@ -47,7 +53,7 @@ def test_positive_update_login_page_footer_text(self):
                 self.assertEqual(login.value, login_text)
 
     @run_in_one_thread
-    @tier1
+    @tier2
```
mirzal: Since this is functional test on acceptance level, I think it should be on Tier3 (according to test metadata doc, Tier2 is for functional tests on Component or Integration level).

tests/foreman/cli/test_setting.py
```diff
@@ -146,7 +151,7 @@ def test_positive_update_login_page_footer_text_without_value(self):
         login_text = Settings.list({'search': 'name=login_text'})[0]
         self.assertEqual('', login_text['value'])
 
-    @tier1
+    @tier2
```
mirzal: If test has importance Low, it has to be on Tier3 or Tier4.

tests/foreman/api/test_setting.py
```diff
@@ -1,13 +1,13 @@
 # -*- encoding: utf-8 -*-
 """Test class for Setting Parameter values
 
-:Requirement: Setting
+:Requirement: Settings
```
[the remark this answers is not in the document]
omkarkhatavkar: It should be `Settings` which is more appropriate because is it is not a singular, we can have multiple settings. This is even I can see Satellite UI as well.

tests/foreman/api/test_setting.py
omkarkhatavkar: I even thought about it making at tier3, did not have a more clear thought.

omkarkhatavkar: Look like I will support my comment, as while reviewing the code I can see that only single entity is been used. So keeping :CaseLevel: Acceptance does not make sense

omkarkhatavkar:
> I disagree with Tiers assignments, but other than that, looks OK. Thanks for working on this.
> 
> Please change tiers (I left two comments, but they apply to all tests), or explain why you think they should be where you put them.

@mirzal I have updated the PR with some more changes. I have even added tier validation to just get an idea of missing obvious tier mistakes.

tests/foreman/cli/test_setting.py
omkarkhatavkar: There is an update in CaseLevel, this might leads to tier2.

tests/foreman/api/test_setting.py
mirzal: Yes, functional test case on Component level definitely does belong to Tier2. It's good now.

tests/foreman/cli/test_setting.py
mirzal: Right, functional test on Component level belongs to Tier2, no matter the Importance.
Sometimes I get confused by these tier rules myself ;) .

PR #6975, opened 2019-05-30
Cleanup CaseAutomation,CaseComponent,CaseImportance metadata and tiers

pgagne:


pgagne:
Its not clear to my why the travis-ci run failed, is it just broken?

mirzal:
@pgagne I've seen failure in another test in the same module in another PR. Rebuild went all good. Seems we have intermittent failure in Robottelo test suite. Worth investigating, but unrelated to this PR.

tests/foreman/api/test_activationkey.py
```diff
@@ -6,7 +6,7 @@
 
 :CaseLevel: Acceptance
```
mirzal: I think you might want to change it to `:CaseLevel: Component`. This should fix most of testimony complaints about tier assignment, as well as simply make more sense - things like `test_positive_create_with_description` are testing one component independently of anything else, so they do belong to Component level.

tests/foreman/api/test_errata.py
```diff
@@ -6,7 +6,7 @@
 
 :CaseLevel: Acceptance
```
mirzal: Since all tests in this file overwrite it to set `:CaseLevel: System`, you might as well change this to `System` here and remove assignments on individual test level.
But you don't have to, effect is the same.

tests/foreman/api/test_lifecycleenvironment.py
```diff
@@ -10,7 +10,7 @@
 
 :CaseLevel: Acceptance
 
-:CaseComponent: API
+:CaseComponent: ErrataManagement
```
mirzal: Shouldn't that be `LifecycleEnvironments`?

tests/foreman/api/test_lifecycleenvironment.py
```diff
@@ -118,7 +118,7 @@ def test_positive_update_name(self):
 
         :expectedresults: Lifecycle environment is created and updated properly
 
-        :CaseImportance: Critical
+        :CaseImportance: High
```
mirzal: You have `High` set on module level, so you don't have to set it here.

tests/foreman/api/test_lifecycleenvironment.py
```diff
@@ -10,7 +10,7 @@
 
 :CaseLevel: Acceptance
```
mirzal: The same comment as I left for `CaseLevel` in `tests/foreman/api/test_activationkey.py` - it probably makes more sense to change it to `Component`.

tests/foreman/cli/test_activationkey.py
```diff
@@ -8,7 +8,7 @@
 
 :CaseLevel: Acceptance
```
mirzal: The same comment as I left for `CaseLevel` in `tests/foreman/api/test_activationkey.py` - it probably makes more sense to change it to `Component`.

tests/foreman/cli/test_activationkey.py
```diff
@@ -306,7 +306,7 @@ def test_negative_create_with_usage_limit_with_not_integers(self):
         :expectedresults: Activation key is not created. Appropriate error
             shown.
 
-        :CaseImportance: Critical
+        :CaseImportance: Low
```
mirzal: Tier1 is reserved for Functional tests on Component level with importance Critical, so this change should go with change to tier decorator. The same applies to many other tests in this file.

tests/foreman/cli/test_lifecycleenvironment.py
```diff
@@ -7,7 +7,7 @@
 
 :CaseLevel: Acceptance
```
mirzal: The same comment as I left for `CaseLevel` in `tests/foreman/api/test_activationkey.py` - it probably makes more sense to change it to `Component`.

tests/foreman/cli/test_lifecycleenvironment.py
```diff
@@ -122,8 +120,6 @@ def test_positive_create_with_description(self):
         :expectedresults: Lifecycle environment is created with Library as
             prior
 
-
-        :CaseImportance: Critical
```
mirzal: Changing importance from Critical to High changes tier assignment.

pgagne:
@mirzal I fixed the tier/caseimportance inconsistencies you pointed out. I also tried to adjust some of the caselevels. For this I was just prioritizing the stuff defined in the document. If you think CaseLevels are also something that needs review perhaps we should update the doc with links to the ISTQB definitions (I wasn't able to find them, I haven't done the ISTQB stuff yet, but am scheduled to).

tests/foreman/cli/test_activationkey.py
pgagne: moved to tier3

tests/foreman/cli/test_lifecycleenvironment.py
pgagne: moved to tier2

tests/foreman/installer/test_infoblox.py
```diff
@@ -2,11 +2,9 @@
 
 :Requirement: Infoblox
 
-:CaseAutomation: Automated
-
 :CaseLevel: System
 
-:CaseComponent: Installer
+:CaseComponent: Infobloxintegration
```
ntkathole: @pgagne Is this bugzilla component? I am not sure we are strictly bound to bz components or not. @mirzal am I right?

pgagne: It is 

https://bugzilla.redhat.com/describecomponents.cgi?product=Red%20Hat%20Satellite%206

ntkathole: @pgagne Thanks. I was looking in mojo doc. We might need to update there.

Issue #225, opened 2013-12-24
CLI : Implement the 'os' hammer cli 


tkolhar:
hammer os --help
Usage:
    hammer os [OPTIONS] SUBCOMMAND [ARG] ...

Parameters:
    SUBCOMMAND                    subcommand
    [ARG] ...                     subcommand arguments

Subcommands:
    add_configtemplate            Associate a resource
    remove_architecture           Disassociate a resource
    info                          Show an OS.
    update                        Update an OS.
    remove_ptable                 Disassociate a resource
    add_architecture              Associate a resource
    delete                        Delete an OS.
    add_ptable                    Associate a resource
    delete_parameter              Delete parameter for an operating system.
    set_parameter                 Create or update parameter for an operating system.
    remove_configtemplate         Disassociate a resource
    create                        Create an OS.
    list                          List all operating systems.

Options:
    -h, --help                    print help

PR #6976, opened 2019-05-31
Clean-Up: run_only_on decorator

jyejare:
Clean Up takes place in:

- All Tier Tests
- All Unit tests for robottelo
- Properties.sample File
- Functions that were using settings.project from robottelo properties


jyejare:
The Corresponding MR in Gitlab is raised @ #225.

jyejare:
Once this PR is in, I will raise cherrypicks for all other branches

robottelo/decorators/__init__.py
```diff
@@ -199,6 +199,8 @@ class ProjectModeError(Exception):
     """Indicates an error occurred while skipping based on Project Mode."""
 
 
+# This defination is not being used currently as there are no multiple projects to run the tests on
```
ntkathole: Typo "defination"

robottelo/decorators/__init__.py
```diff
@@ -199,6 +199,8 @@ class ProjectModeError(Exception):
...
+# Also, The setting project has been removed from properties untill wnd if we need in feature
```
ntkathole: Typo "untill wnd" and also sentence not well framed

robottelo/decorators/__init__.py
```diff
@@ -199,6 +199,8 @@ class ProjectModeError(Exception):
...
 def run_only_on(project):
```
ntkathole: What is point in keeping this definition if we are removing test class definition as well https://github.com/SatelliteQE/robottelo/pull/6976/files#diff-86bbbc36dd5587c78264c1b8f556a6b7L930

robottelo/decorators/__init__.py
mirzal: Also, "need in feature" or "need in future"?

robottelo/decorators/__init__.py
jyejare: As I said in comments above that this can be used later if a similar sort of scenario comes where we need to choose between two things.

About the test cases, if you saw those tests, they were tightly coupled with 'project' property in robottelo.propoerties which I have removed in this PR. Hence whenever you run 'make test', the tests will fail with no property found even though we have a definition in place.

So let's keep the definition hanging as it's not being used anywhere right now but can be used later. We can write tests later as per the function use.

jyejare:
@ntkathole @mirzal, Fixed silly typos, and sentence. Please revisit.

tests/foreman/api/test_repository_set.py
```diff
@@ -23,7 +23,7 @@
 from robottelo import manifests
 from robottelo.api.utils import upload_manifest
 from robottelo.constants import PRDS, REPOSET
-from robottelo.decorators import run_in_one_thread, run_only_on, tier1, upgrade
+from robottelo.decorators import run_in_one_thread,  tier1, upgrade
```
mirzal: There is now double space between `run_in_one_thread` and `tier1`.

jyejare: I wonder flake8 didnt catch it :)

jyejare:
@mirzal , I fixed all files u mentioned above except ```tests/foreman/cli/test_hostgroup.py```, I didnt find any trace. Can you help here ?

mirzal:
@jyejare Oh, I applied your patch on top of older master. You are right, `tests/foreman/cli/test_hostgroup.py` is fine, too.

jyejare:
@mirzal , Waiting for your merge comment. :)

PR #6978, opened 2019-05-31
fixes #4738 - removing the installer log test entirely

### LINKED ISSUE #4738: installer - SELinuxTestCase.test_positive_check_installer_logfile - should run at the very beginning
I'v got problem with this Test class. First of all, it does not match robottelo design, as it depends on the satellite state - in order to run correctly, it requires to be executed just after the installation (this might be somehow hacked by using some marker logic, but one can still run the test suite on a satellite which is not 'clean' anymore).
- As satellite constantly commulates state, it also logs messages to its logfiles.
`test_positive_check_installer_logfile` checks `/var/log/candlepin/error.log` - the time this file is checked, satellite has undergone several tests already and the file contains logs *not only* from the installer. This leads to bogus failures.

- My second concern is - why is this called `SELinuxTestCase`?
- the last concern - https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/installer/test_installer.py#L309-L310 - it seems the installer log always contains the following message:
```
[DEBUG 2017-05-23 14:52:03 main] calling :function_validate_re with [["ERROR", "^(UNKNOWN|FATAL|ERROR|WARN|INFO|DEBUG)$"]]
```
I guess this message will make the statement always positive

rplevka:
fixes https://github.com/SatelliteQE/robottelo/issues/4738
This test doesn't comply with the current testing strategy as it is designed to be written just after the installer finishes.
It checks for errors in the logfile, but since the order of the test is not guaranteed, it interferes with the rest of the tests, that get executed sooner and potentially feed logs with error/warn log messages (many of which are valid ones).

### DIFF
--- a/tests/foreman/installer/test_installer.py
+++ b/tests/foreman/installer/test_installer.py
@@ -14,8 +14,6 @@
 
 :Upstream: No
 """
-from datetime import datetime
-
 import re
 from six.moves import zip
 
@@ -24,7 +22,6 @@ from robottelo.config import settings
 from robottelo.constants import RHEL_6_MAJOR_VERSION, RHEL_7_MAJOR_VERSION
 from robottelo.decorators import tier1, upgrade
 from robottelo.helpers import get_host_info
-from robottelo.log import LogFile
 from robottelo.test import TestCase
 
 PREVIOUS_INSTALLER_OPTIONS = set([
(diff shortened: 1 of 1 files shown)


rplevka:
this one makes a candidate for backporting to other branches.

PR #6992, opened 2019-06-04
some refactoring for classparameters CLI and API tests

pondrejk:
Class parameter CLI and API tests were reviewed and slimmed down. I keep the changes in separate commits to make tracking somewhat easier. Test resuls for the CLI part (one thread):

```
pytest tests/foreman/cli/test_classparameters.py
================================================ test session starts                                  =================================================
platform linux -- Python 3.7.3, pytest-4.0.2, py-1.7.0, pluggy-0.8.1
collecting ... 2019-05-31 14:29:52 - conftest - DEBUG - BZ deselect is disabled in settings

collected 11 items

tests/foreman/cli/test_classparameters.                                                               py ...........                                                          [100%]

=========================================== 11 passed in 565.38 seconds                               =============================================
```

For comparison, before the change there were 47 tests, running for 1201.74 seconds in one thread. API tests went from 49  to 16, unfortunately test results are blocked by https://bugzilla.redhat.com/show_bug.cgi?id=1711929

rundown of changes:
- removed all instances of @run_only_on(sat)
- test_positive_update_parameter_type, test_negative_update_parameter_type removed from cli tests in  favor of their api counterparts
- list tests removed from api in favor of cli conterparts
- cli list tests were merged and chaned to reuse a host from setupclass,                           test_positive_list_by_puppetclass_id removed as it is tested already in                               test_positive_list_with_non_admin_user
- test_positive_list list queries provided in an array and looped over, makes it easier to add/remove items to be listed
- override tests in api removed in favor of cli counterparts, merged override cli tests
- a number of stubbed validation tests has been removed (kind of specific to classparam element.),    being there a long time, contained  excessive variation of inputs and big setup requirements (host provisioning)
- test_positive_hide removed as test_positive_unhide already contains the procedure
- create matcher tests removed from api in favor of cli conterparts, number of negative tests reduced
- validate tests kept mostly in API, with couple of tests in CLI merged
- test metadata updated


tests/foreman/api/test_classparameters.py
```diff
@@ -1523,35 +707,6 @@ def test_negative_enable_avoid_duplicates_checkbox(self):
...
-            match='is_virtual=true',
-            value=value,
-        ).create()
-        self.assertEqual(len(sc_param.read().override_values), 1)
-        override.delete()
-        self.assertEqual(len(sc_param.read().override_values), 0)
-
     @skip_if_bug_open('bugzilla', 1374253)
     @tier1
     def test_positive_impact_parameter_delete_attribute(self):
```
omkarkhatavkar: Looks like you are using more than 2-3 entities could be an ideal case for tier2?

tests/foreman/cli/test_classparameters.py
```diff
@@ -349,60 +137,22 @@ def test_positive_list_by_hostgroup_id(self):
...
-        # Check that only unique results are returned
-        self.assertEqual(len(sc_params), len({scp['id'] for scp in sc_params}))
-
-    @tier1
-    def test_positive_list_by_puppetclass_id(self):
-        """List all the parameters for specific puppet class by id.
 
-        :id: a7a8af1a-514b-4910-9e19-75306f634041
+        # check listing parameters for selected queries
+        for query in list_queries:
```
mirzal: I wonder if you shouldn't use [subtests](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests) here... or shuffle list before using it...

Imagine querying by environment is broken and it takes two months to provide fix. During these two months, we will know nothing about querying by host or puppet class, which might or might not be broken. If you use subtest, we should get separate failure for each query; if you shuffle list, we at least will have a chance to try other queries before encountering broken one and failing whole test.

The problem with subtests approach is that it won't work in pytest, as pytest doesn't support unittest subtests natively. There is `pytest-subtests` plugin, but it requires pytest 4.4 - newer than what we use. Of course updating pytest is out of scope for your PR.

To sum up: conceptually, I would prefer this part to not be blocked by any single failure. What do you think?

pondrejk: Good point, added the subTest. I think we can deal with the pytest problem when it arises, cheers

pondrejk:
re-test after changes:

```
pytest tests/foreman/cli/test_classparameters.py
============================================================== test session starts ==============================================================
platform linux -- Python 3.7.3, pytest-4.0.2, py-1.7.0, pluggy-0.8.1
collecting ... 2019-06-13 10:22:38 - conftest - DEBUG - BZ deselect is disabled in settings

collected 11 items                                                                                                                              

tests/foreman/cli/test_classparameters.py ...........                                                                                     [100%]

========================================================== 11 passed in 383.31 seconds ==========================================================
```

PR #6993, opened 2019-06-04
Make sure errata applicability task finishes before we check applicable errata for the host

jhutar:
This happens from time to time in automation so created a shared utility. I plan to add that function where this pops up.

### DIFF
--- a/robottelo/api/utils.py
+++ b/robottelo/api/utils.py
@@ -794,7 +794,6 @@ def wait_for_errata_applicability_task(host_id, search_rate=1, max_tries=10, pol
             if host_id in task.input['host_ids']:
                 task.poll(poll_rate=poll_rate, timeout=poll_timeout)
                 return task
-        import code; code.interact(local=locals())
         time.sleep(search_rate)
     else:
         raise AssertionError(
--- a/robottelo/api/utils.py
+++ b/robottelo/api/utils.py
@@ -791,11 +791,12 @@ def wait_for_errata_applicability_task(host_id, from_when, search_rate=1, max_tr
     assert isinstance(from_when, int), 'Param from_when have to be int'
     now = int(time.time())
     assert from_when <= now, 'Param from_when have to be timestamp in the past'
-    max_age = now - from_when + 1
-    search_query = '( label = Actions::Katello::Host::GenerateApplicability OR label = ' \
-        'Actions::Katello::Host::UploadPackageProfile ) AND started_at > "%s seconds ago"' \
-        % max_age
     for _ in range(max_tries):
+        now = int(time.time())
+        max_age = now - from_when + 1
+        search_query = '( label = Actions::Katello::Host::GenerateApplicability OR label = ' \
+            'Actions::Katello::Host::UploadPackageProfile ) AND started_at > "%s seconds ago"' \
+            % max_age
         tasks = entities.ForemanTask().search(query={'search': search_query})
         tasks_finished = 0
         for task in tasks:
--- a/robottelo/api/utils.py
+++ b/robottelo/api/utils.py
@@ -771,11 +771,12 @@ def wait_for_syncplan_tasks(repo_backend_id=None, timeout=10, repo_name=None):
         time.sleep(2)
 
 
-def wait_for_errata_applicability_task(host_id, search_rate=1, max_tries=10, poll_rate=None,
-                                       poll_timeout=15):
+def wait_for_errata_applicability_task(host_id, from_when, search_rate=1, max_tries=10,
+                                       poll_rate=None, poll_timeout=15):
     """Search the generate applicability task for given host and make sure it finishes
 
     :param int host_id: Content host ID of the host where we are regenerating applicability.
+    :param int from_when: Timestamp (in UTC) to limit number of returned tasks to investigate.
     :param int search_rate: Delay between searches.
     :param int max_tries: How many times search should be executed.
     :param int poll_rate: Delay between the end of one task check-up and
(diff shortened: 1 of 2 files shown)


jhutar:
```
$ pytest tests/foreman/cli/test_host.py::KatelloAgentTestCase::test_positive_apply_security_erratum --no-print-logs --capture=no --exitfirst 2>&1 | tee /tmp/pytest.log
[...]
========================== 1 passed in 232.61 seconds ==========================
```

vijay8451:
@jhutar Just having a query .. Can we not use the pre-exists utils `wait_for_tasks ` https://github.com/SatelliteQE/robottelo/blob/master/robottelo/api/utils.py#L677

jhutar:
Hello @vijay8451 . I do not want to use generic `wait_for_tasks`, because I want to check `if host_id in task.input['host_ids']:`. But I had forgotten debug there, so removed it - I do not understand how could I miss it.

robottelo/api/utils.py
```diff
@@ -771,6 +771,35 @@ def wait_for_syncplan_tasks(repo_backend_id=None, timeout=10, repo_name=None):
         time.sleep(2)
 
 
+def wait_for_errata_applicability_task(host_id, search_rate=1, max_tries=10, poll_rate=None,
+                                       poll_timeout=15):
+    """Search the generate applicability task for given host and make sure it finishes
```
vijay8451:
```suggestion
    """Search the generate applicability task for given host and return relevant task
```
Above looks more relevant as helper does not wait to finish task, right ?

robottelo/api/utils.py
```diff
@@ -771,6 +771,35 @@ def wait_for_syncplan_tasks(repo_backend_id=None, timeout=10, repo_name=None):
...
+
+    :param host_id: Content host ID of the host where we are regenerating applicability.
```
vijay8451:
```suggestion
    :param int host_id: Content host ID of the host where we are regenerating applicability.
```

robottelo/api/utils.py
```diff
@@ -771,6 +771,35 @@ def wait_for_syncplan_tasks(repo_backend_id=None, timeout=10, repo_name=None):
...
+    :param search_rate: Delay between searches.
```
vijay8451:
```suggestion
    :param int search_rate: Delay between searches.
```

robottelo/api/utils.py
```diff
@@ -771,6 +771,35 @@ def wait_for_syncplan_tasks(repo_backend_id=None, timeout=10, repo_name=None):
...
+    :param max_tries: How many times search should be executed.
```
vijay8451:
```suggestion
    :param int max_tries: How many times search should be executed.
```

robottelo/api/utils.py
```diff
@@ -771,6 +771,35 @@ def wait_for_syncplan_tasks(repo_backend_id=None, timeout=10, repo_name=None):
...
+    :param poll_rate: Delay between the end of one task check-up and
```
vijay8451:
```suggestion
    :param int poll_rate: Delay between the end of one task check-up and
```

robottelo/api/utils.py
```diff
@@ -771,6 +771,35 @@ def wait_for_syncplan_tasks(repo_backend_id=None, timeout=10, repo_name=None):
...
+            the start of the next check-up. Parameter for
+            ``nailgun.entities.ForemanTask.poll()`` method.
+    :param poll_timeout: Maximum number of seconds to wait until timing out.
```
vijay8451:
```suggestion
    :param int poll_timeout: Maximum number of seconds to wait until timing out.
```

vijay8451:
> Hello @vijay8451 . I do not want to use generic `wait_for_tasks`, because I want to check `if host_id in task.input['host_ids']:`. But I had forgotten debug there, so removed it - I do not understand how could I miss it.
@jhutar  thanks got it .

robottelo/api/utils.py
jhutar: It pools the task, so it waits for it to finish here:
```
task.poll(poll_rate=poll_rate, timeout=poll_timeout)
```

robottelo/api/utils.py
jhutar: Thank you! Fixed in:
```
[jhutar-4 490ed3a3] Make sure this is int
```

robottelo/api/utils.py
jhutar: Thank you! Fixed in:
```
[jhutar-4 57c285f0] Make sure this is int
```

robottelo/api/utils.py
jhutar: Thank you! Fixed in:
```
[jhutar-4 57c285f0] Make sure this is int
```

robottelo/api/utils.py
jhutar: Thank you! Fixed in:
```
[jhutar-4 57c285f0] Make sure this is int
```

robottelo/api/utils.py
jhutar: Thank you! Fixed in:
```
[jhutar-4 57c285f0] Make sure this is int
```

jhutar:
I'm now testing it more and I was able to reproduce the original issue even with this fix, so I'll be looking into it. Mnt please

robottelo/api/utils.py
```diff
@@ -771,6 +771,35 @@ def wait_for_syncplan_tasks(repo_backend_id=None, timeout=10, repo_name=None):
         time.sleep(2)
 
 
+def wait_for_errata_applicability_task(host_id, search_rate=1, max_tries=10, poll_rate=None,
```
rplevka: do you realize that this will not be very useful in scenarios with 2 consecutive errata applicability triggers?
as the second check would match the first task and immediately return it.
Perhaps accepting a datetime parameter e.g. `started_after` or something that would allow user to specify the target task as much as possible could help.
What do you think?

jhutar: Yep, rewrote it partly for this reason. Please review again.

jhutar:
@vijay8451 and @rplevka, please review again.
Seems stable now:
$ for i in $(seq 5); do pytest tests/foreman/cli/test_host.py::KatelloAgentTestCase::test_positive_apply_security_erratum --no-print-logs --capture=no --exitfirst 2>&1 >/tmp/test-$i.log; tail -n 1 /tmp/test-$i.log; done
========================== 1 passed in 244.39 seconds ==========================
========================== 1 passed in 248.22 seconds ==========================
========================== 1 passed in 251.22 seconds ==========================
========================== 1 passed in 260.54 seconds ==========================
========================== 1 failed in 260.13 seconds ==========================

jhutar:
Hmm, will do one more change. Need to count `max_age` in the loop as it incorporates waiting time. That might solve that last failure.

jhutar:
Now or never:
```
$ for i in $(seq 5); do pytest tests/foreman/cli/test_host.py::KatelloAgentTestCase::test_positive_apply_security_erratum --no-print-logs --capture=no --exitfirst 2>&1 >/tmp/test-$i.log; tail -n 1 /tmp/test-$i.log; done
========================== 1 passed in 245.16 seconds ==========================
========================== 1 passed in 244.33 seconds ==========================
========================== 1 passed in 246.54 seconds ==========================
========================== 1 passed in 249.47 seconds ==========================
========================== 1 passed in 257.52 seconds ==========================
```
!

jhutar:
Thank you!

PR #6999, opened 2019-06-06
tested updated pyup-update-wait-for-1.0.11-to-1.0.13 with rebase master

omkarkhatavkar:
```
tests/foreman/cli/test_repository.py::RepositoryTestCase::test_positive_synchronize_docker_repo_with_tags_whitelist PASSED [100%]

(env) [root@okhatavk robottelo]# pip list | grep wait
DEPRECATION: Python 3.4 support has been deprecated. pip 19.1 will be the last one supporting it. Please upgrade your Python as Python 3.4 won't be maintained after March 2019 (cf PEP 429).
wait-for                              1.0.13        
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,7 +18,7 @@ six==1.11.0
 testimony==1.6.0
 unittest2==1.1.0
 PyNaCl==1.2.1
-wait-for==1.0.11
+wait-for==1.0.13
 wrapanapi==3.2.0
 urllib3==1.24.2
 # python-bugzilla 2.0.0 changed a lot of function signatures


mirzal:
I take that we want this one instead of #6944? Can you close the other one already?

Could you also add result of at least one test from ui_airgun? Airgun uses `wait_for` extensively, and also with lambda function. So it's probably at biggest risk of regression here.

omkarkhatavkar:
> here

@mirzal If we consider Airgun, it is already using `Successfully installed airgun wait-for-1.0.13`  version. It won't impact regression as this change is only related to the only robottelo. Any underneath call to airgun code will use airgun's wait_for which is already updated. what do you say?

omkarkhatavkar:
@mirzal Still I have run this test which is failing because of host not able to run the command and failied with a timeout error.

```
 wait_for.TimedOutError: Could not do lambda defined as `lambda: session.jobinvocation.read('Run ls', hostname, 'overview.hosts_table')[
E                                   'overview']['hosts_table'][0]['Status'] == 'running',` at /home/okhatavk/Satellite/robottelo/tests/foreman/ui_airgun/test_remoteexecution.py:264 in time

test_remoteexecution.py:267: TimedOutError

```

mirzal:
> Any underneath call to airgun code will use airgun's wait_for which is already updated. what do you say?

[Airgun uses unversioned dependency on wait_for](https://github.com/SatelliteQE/airgun/blob/master/setup.py#L21). When we consider Robottelo, requirements.txt will first install wait_for 1.0.11 and then airgun; while installing airgun, airgun's dependency on wait_for will already be satisfied. So, airgun will use the same version of wait_for that Robottelo requires.

At least in situation that I consider standard, that is creating venv for Robottelo and `pip install -r robottelo/requirements.txt`. If you install airgun manually first, then it will pull in latest `wait_for`. But then, `pip install -r robottelo/requirements.txt` should complain you have two versions of the same package.

> Still I have run this test which is failing because of host not able to run the command and failied with a timeout error.

Can you check if this is all tests in ui_airgun, or only this particular test? I mean, run some other tests in ui_airgun (doesn't have to be entire directory) and see if they are ok or not. These tests don't have to use `wait_for`, it's enough that airgun is using it internally.

ntkathole:
@mirzal airgun is already using latest https://github.com/SatelliteQE/airgun/blob/master/setup.py#L21

omkarkhatavkar:
@ntkathole @mirzal @san7ket 

I tried running some tests that are getting passed which was using  https://github.com/SatelliteQE/airgun/blob/c1aae41133daa2f35df5f1f1ee75303980d3c789/airgun/views/product.py#L207

```
tests/foreman/ui_airgun/test_repository.py::test_positive_discover_module_stream_repo_via_existing_product PASSED [ 33%]
tests/foreman/ui_airgun/test_repository.py::test_positive_discover_repo_via_existing_product PASSED         [ 66%]
tests/foreman/ui_airgun/test_repository.py::test_positive_discover_repo_via_new_product PASSED              [100%]
```

san7ket:
Not well-versed with airgun , would request @mirzal to have a look

Issue #5349, opened 2017-10-06
vm.py - implement qemu-guest-agent support


rplevka:
`qemu-guest-agent` is a daemon installed on a virtual guest, which reports some storage and network interface information back to its hypervysor over serial channel.
- I aim to use this to source the IP address information. The main advantage of using a direct serial channel is that there is no network involved, making this method universally work even in the VLAN network environments.

- [x] **snap-guest** - add `--channel` argument to the resulting `virt-install` command in order to define the serial channel for the created domain
- [x] **base-image** - re-create the base images by installing and enabling the `qemu-guest-agent` (available in the base rhel-server repo)
- [ ] **vm.py** - add an option to prefer `virsh domifaddr` source of information over `icmp` method (possibly making it default to preserve test compatibility)

lpramuk:
the first two I've done already
- snap-guest is ready
- base-images are ready too

lpramuk:
@rplevka when implementing changes in vm.py you might find these two commands usefull (run at hypervisor)
```
virsh qemu-agent-command $VM '{"execute":"guest-network-get-interfaces"}'| jq -r '.return[]|select(.name!="lo")."ip-addresses"[]|select(."ip-address-type"=="ipv4")."ip-address"'
```
```
virsh -q domifaddr $VM --source agent
```
I find the first one more usefull as it returns json and you can parse it in python easily without using ```jq```

Issue #7013, opened 2019-06-07
VirtualMachine.create() has hardcoded top-level domain name


mirzal:
VirtualMachine.create() success depends on pinging new vm based on hardcoded domain name (
https://github.com/SatelliteQE/robottelo/blob/master/robottelo/vm.py#L236 ). This passes on our testing environment, but will fail everywhere else - unless you configure your local environment to mirror testing environment.

This class should be re-worked to be more robust. Any libvirt domain setting should be allowed. VirtualMachine class already has `hostname` and `domain` properties that might come in handy.

jhutar:
Hello. To get that `ping xxx.local` working, this is what Lukas advised me few days back:
```
rlPhaseStartSetup "Avahi discovery"
rlRun "yum -y install http://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm"
rlRun "yum -y install nss-mdns" # pulls in avahi
rlRun "rpm -e epel-release"
rlRun "service firewalld restart"
rlRun "firewall-cmd --add-service mdns && firewall-cmd --add-service mdns --permanent"
rlRun "service avahi-daemon restart"
rlAssertGrep mdns4_minimal /etc/nsswitch.conf
rlPhaseEnd
```
(ignore these `rlRun` wrappers, code copy&pasted from Lukas's Beaker task)

mirzal:
There is PR for using different approach to obtaining host IP address, one that does not depend on any particular network setup - https://github.com/SatelliteQE/robottelo/pull/7018

PR #7018, opened 2019-06-11
add support for qemu-guest-agent in vm.py

rplevka:
fixes https://github.com/SatelliteQE/robottelo/issues/5349, https://github.com/SatelliteQE/robottelo/issues/7013

this now implicitly uses qemu-guest-agent to query ip address info. Also, to maintain backward compatibility, the script falls back to the standard ping (Avahi) method in case the guest-agent is not configured or the script fails to source the info in time.

```
Starting install...
Domain creation completed.

2019-06-11 18:10:29 - robottelo.ssh - DEBUG - Destroyed Paramiko client 0x7f630584ef28
2019-06-11 18:10:30 - robottelo.ssh - DEBUG - Instantiated Paramiko client 0x7f630584ef98
2019-06-11 18:10:30 - robottelo.ssh - INFO - Connected to [***]
2019-06-11 18:10:30 - robottelo.ssh - INFO - >>> virsh qemu-agent-command adfxkkdmawd6joql.foo.bar.com '{"execute":"guest-network-get-interfaces"}'
2019-06-11 18:10:33 - robottelo.ssh - INFO - <<< stdout

2019-06-11 18:10:33 - robottelo.ssh - INFO - <<< stderr
error: Guest agent is not responding: QEMU guest agent is not connected
...

2019-06-11 18:10:45 - robottelo.ssh - INFO - >>> virsh qemu-agent-command adfxkkdmawd6joql.lab.eng.rdu2.redhat.com '{"execute":"guest-network-get-interfaces"}'
2019-06-11 18:10:47 - robottelo.ssh - INFO - <<< stdout
{"return":[{"name":"lo","ip-addresses":[{"ip-address-type":"ipv4","ip-address":"127.0.0.1","prefix":8}],"hardware-address":"00:00:00:00:00:00"},{"name":"ens3","ip-addresses":[{"ip-address-type":"ipv6","ip-address":"2620:52:0:800:5054:ff:fe31:d7d3","prefix":64},{"ip-address-type":"ipv6","ip-address":"fe80::5054:ff:fe31:d7d3","prefix":64}],"hardware-address":"52:54:00:31:d7:d3"}]}
...
2019-06-11 18:10:50 - robottelo.ssh - INFO - <<< stdout
{"return":[{"name":"lo","ip-addresses":[{"ip-address-type":"ipv4","ip-address":"127.0.0.1","prefix":8}],"hardware-address":"00:00:00:00:00:00"},{"name":"ens3","ip-addresses":[{"ip-address-type":"ipv4","ip-address":"10.8.30.227","prefix":19}],"hardware-address":"52:54:00:31:d7:d3"}]}
...
2019-06-11 18:10:52 - robottelo.ssh - INFO - >>> for i in {1..60}; do nc -vn 10.8.30.227 22 <<< "" && exit 0; sleep 1; done; exit 1
2019-06-11 18:10:54 - robottelo.ssh - INFO - <<< stdout
SSH-2.0-OpenSSH_7.4
Protocol mismatch.

2019-06-11 18:10:54 - robottelo.ssh - INFO - <<< stderr
Ncat: Version 7.60 ( https://nmap.org/ncat )
Ncat: Connected to 10.8.30.227:22.
Ncat: 1 bytes sent, 40 bytes received in 0.03 seconds.
```

### DIFF
--- a/robottelo/vm.py
+++ b/robottelo/vm.py
@@ -9,6 +9,7 @@ make sure that the server have in place: the base images for rhel66 and rhel71,
 snap-guest and its dependencies and the ``image_dir`` path created.
 
 """
+import json
 import logging
 import os
 import six
@@ -28,6 +29,7 @@ from robottelo.constants import (
 from robottelo.helpers import install_katello_ca, remove_katello_ca
 from robottelo.host_info import get_host_os_version
 from six.moves.urllib.parse import urlunsplit
+from time import sleep
 # This conditional is here to centralize use of urljoin
 if six.PY3:  # pragma: no cover
     from urllib.parse import urljoin  # noqa
@@ -77,6 +79,7 @@ class VirtualMachine(object):
             DISTRO_SLES12: settings.distro.image_sles12,
         }
         self.cpu = cpu
+        self.mac = None
         self.ram = ram
         self.nw_type = None
         if distro is None:
@@ -182,6 +185,7 @@ class VirtualMachine(object):
             '-m {vm_ram}',
             '-c {vm_cpu}',
             '-n {nw_type}={nw_name} -f',
+            '--qemu-ga',
         ]
 
         if self.image_dir is not None:
@@ -214,17 +218,16 @@ class VirtualMachine(object):
         )
 
         result = ssh.command(
-                command,
-                self.provisioning_server,
-                connection_timeout=30
+            command,
+            self.provisioning_server,
+            connection_timeout=30
         )
-
         if result.return_code != 0:
             raise VirtualMachineError(
                 u'Failed to run snap-guest: {0}'.format(result.stderr))
         else:
             self._created = True
-
+            self.mac = [n.split('MAC:')[1].strip() for n in result.stdout if 'MAC:' in n][0]
         # outside of VLANs ping from hypervisor, in VLANs ping from SAT
         if self.bridge == 'br0':
             ping_from_hostname = self.provisioning_server
@@ -232,19 +235,49 @@ class VirtualMachine(object):
             ping_from_hostname = settings.server.hostname
 
         # Give some time to machine boot
-        result = ssh.command(
-            u'for i in {{1..60}}; do ping -c1 {0}.local && exit 0; sleep 1;'
-            u' done; exit 1'.format(self._target_image),
-            ping_from_hostname,
-            connection_timeout=30
-        )
-        if result.return_code != 0:
-            logger.error('Failed to obtain VM IP, reverting changes')
-            self.destroy()
-            raise VirtualMachineError(
-                'Failed to fetch virtual machine IP address information')
-        output = ''.join(result.stdout)
-        self.ip_addr = output.split('(')[1].split(')')[0]
+        for i in range(60):
+            qemu_ga_check = ssh.command(
+                u'virsh qemu-agent-command {0} '
+                '\'{{"execute":"guest-network-get-interfaces"}}\''.format(self.hostname),
+                ping_from_hostname,
+                connection_timeout=30
+            )
+            if qemu_ga_check.return_code != 0:
+                if 'guest agent is not connected' in qemu_ga_check.stderr:
+                    # this means the agent wasn't started yet (vm still booting)
+                    sleep(1)
+                else:
+                    # this means there's another error with agent, e.g. it is not configured
+                    break
+            else:
+                ifaces = json.loads(qemu_ga_check.stdout[0])
+                mgmt_if = next((i for i in ifaces['return']
+                               if i['hardware-address'].lower() == self.mac.lower()), {})
+                try:
+                    # get only the ipv4 addresses
+                    self.ip_addr = next(i['ip-address'] for i in mgmt_if['ip-addresses']
+                                        if i['ip-address-type'] == 'ipv4')
+                    break
+                except(KeyError, StopIteration):
+                    sleep(1)
+                    continue
+
+        if not self.ip_addr:
+            # fallback to avahi in case of any issue with qemu-guest-agent
+            logger.warning('Failed to parse the mgmt IPv4 using qemu-guest-agent, trying Avahi')
+            result = ssh.command(
+                u'for i in {{1..60}}; do ping -c1 {0}.local && exit 0; sleep 1;'
+                u' done; exit 1'.format(self._target_image),
+                ping_from_hostname,
+                connection_timeout=30
+            )
+            if result.return_code != 0:
+                logger.error('Failed to obtain VM IP, reverting changes')
+                self.destroy()
+                raise VirtualMachineError(
+                    'Failed to fetch virtual machine IP address information')
+            output = ''.join(result.stdout)
+            self.ip_addr = output.split('(')[1].split(')')[0]
         ssh_check = ssh.command(
             u'for i in {{1..60}}; do nc -vn {0} 22 <<< "" && exit 0; sleep 1;'
             u' done; exit 1'.format(self.ip_addr),
--- a/tests/robottelo/test_vm.py
+++ b/tests/robottelo/test_vm.py
@@ -23,7 +23,7 @@ class VirtualMachineTestCase(unittest2.TestCase):
         super(VirtualMachineTestCase, self).tearDown()
         self.settings_patcher.stop()
 
-    def configure_provisoning_server(self):
+    def configure_provisioning_server(self):
         """Helper for configuring the provisioning server on robottelo config.
 
         """
@@ -31,14 +31,18 @@ class VirtualMachineTestCase(unittest2.TestCase):
 
     @patch('time.sleep')
     @patch('robottelo.ssh.command', side_effect=[
-        ssh.SSHCommandResult(),
+        ssh.SSHCommandResult(
+            return_code=0,
+            stdout=['CPUs:     1', 'Memory:   512 MB', 'MAC:      52:54:00:f7:bb:a8']
+        ),
+        ssh.SSHCommandResult(return_code=1, stderr=''),
         ssh.SSHCommandResult(stdout=['(192.168.0.1)']),
         ssh.SSHCommandResult()
     ])
     def test_dont_create_if_already_created(
             self, ssh_command, sleep):
         """Check if the creation steps are run more than once"""
-        self.configure_provisoning_server()
+        self.configure_provisioning_server()
         vm = VirtualMachine()
 
         with patch.multiple(
@@ -49,7 +53,7 @@ class VirtualMachineTestCase(unittest2.TestCase):
             vm.create()
             vm.create()
         self.assertEqual(vm.ip_addr, '192.168.0.1')
-        self.assertEqual(ssh_command.call_count, 3)
+        self.assertEqual(ssh_command.call_count, 4)
 
     def test_invalid_distro(self):
         """Check if an exception is raised if an invalid distro is passed"""
@@ -64,7 +68,7 @@ class VirtualMachineTestCase(unittest2.TestCase):
     @patch('robottelo.ssh.command')
     def test_run(self, ssh_command):
         """Check if run calls ssh.command"""
-        self.configure_provisoning_server()
+        self.configure_provisioning_server()
         vm = VirtualMachine()
 
         def create_mock():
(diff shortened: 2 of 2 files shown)


robottelo/vm.py
```diff
@@ -232,19 +235,50 @@ def create(self):
...
+                        if len(mgmt_if) > 0:
+                            if mgmt_if[0].get('ip-addresses'):
+                                # get only the ipv4 addresses
+                                ip4_addr = [i['ip-address'] for i in mgmt_if[0]['ip-addresses']
+                                            if i['ip-address-type'] == 'ipv4']
+                                if len(ip4_addr) > 0:
+                                    self.ip_addr = ip4_addr[0]
+                                    break
+        else:
+            # fallback to avahi in case of any issue with qemu-guest-agent
```
mirzal: As mentioned in previous comment, we enter this path based only on response of initial `virsh qemu-agent-command` run. If anything happens in subsequent call, we do not execute this path.

I think this `else` was supposed to be on different indentation level than it actually is.

robottelo/vm.py
```diff
@@ -232,19 +235,50 @@ def create(self):
...
             connection_timeout=30
         )
-        if result.return_code != 0:
-            logger.error('Failed to obtain VM IP, reverting changes')
-            self.destroy()
-            raise VirtualMachineError(
-                'Failed to fetch virtual machine IP address information')
-        output = ''.join(result.stdout)
-        self.ip_addr = output.split('(')[1].split(')')[0]
+        if qemu_ga_check.return_code != 0:
```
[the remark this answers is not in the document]
rplevka: ah, good catch. right, that is a leftover from the old logic, funny fact: this wouldn't return zero return code in like 99.99% cases as nothing boots that quick.
will fix this

rplevka:
@mirzal I addressed your comments + added some robottelo tests to assert the qemu-ga output parsing behaviour.

robottelo/vm.py
```diff
@@ -232,19 +235,50 @@ def create(self):
...
+                        mgmt_if = [i for i in ifaces['return']
+                                   if 'eth0' in i['name'] or 'ens3' in i['name']]
+                        if len(mgmt_if) > 0:
+                            if mgmt_if[0].get('ip-addresses'):
+                                # get only the ipv4 addresses
+                                ip4_addr = [i['ip-address'] for i in mgmt_if[0]['ip-addresses']
+                                            if i['ip-address-type'] == 'ipv4']
+                                if len(ip4_addr) > 0:
+                                    self.ip_addr = ip4_addr[0]
+                                    break
```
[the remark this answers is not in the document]
rplevka: I reworked the logic slightly, which decresed the number of levels slightly. Check if that works for you

robottelo/vm.py
```diff
@@ -232,19 +235,46 @@ def create(self):
...
+                ifaces = json.loads(qemu_ga_check.stdout[0])
+                mgmt_if = [i for i in ifaces['return']
+                           if 'eth0' in i['name'] or 'ens3' in i['name']]
+                if len(mgmt_if) > 0 and mgmt_if[0].get('ip-addresses'):
+                    # get only the ipv4 addresses
+                    ip4_addr = [i['ip-address'] for i in mgmt_if[0]['ip-addresses']
+                                if i['ip-address-type'] == 'ipv4']
+                    if len(ip4_addr) > 0:
+                        self.ip_addr = ip4_addr[0]
+                        break
```
mirzal: You can use iterator here (lines 258-262) instead:
```python
self.ip_addr = next((i['ip-address'] for i in mgmt_if[0]['ip-addresses'] if i['ip-address-type'] == 'ipv4'), None)
if self.ip_addr:
    break
```
(maybe move generator comprehension to separate variable to increase readability, but you get the idea)

tests/robottelo/test_vm.py
```diff
@@ -126,3 +127,48 @@ def test_destroy(self, ssh_command):
...
+        ssh.SSHCommandResult(stdout=[
+            '{"return":[{"name":"lo","ip-addresses":[{"ip-address-type":"ipv4",'
+            '"ip-address":"127.0.0.1","prefix":8}],"hardware-address":"00:00:00:00:00:00"}'
+            ',{"name":"ens3","ip-addresses":[{"ip-address-type":"ipv4",'
+            '"ip-address":"10.8.30.135","prefix":19}],"hardware-address":"52:54:00:f7:bb:a8"}]}'
+        ]),
+        ssh.SSHCommandResult()
+    ])
+    def test_qemu_ga_waits_for_ip(self, ssh_command, sleep):
+        """Verify that the qemu-ga waits for the IPv4 to be obtained"""
```
mirzal: Is there something that we could do to make this test more explicit? Because right now it looks like function that does nothing, because all the logic is moved into decorators and fixtures.

rplevka: Not sure what you'd like to see here. 
The purpose of the test is to assert that `vm.create()` properly acquires the ip address from the ssh command output, while the mocking simulates the possible qemu-ga outputs.

robottelo/vm.py
rplevka: ...but why?

robottelo/vm.py
```diff
@@ -232,19 +235,46 @@ def create(self):
...
+                    # this means the agent wasn't started yet (vm still booting)
+                    sleep(1)
+                else:
+                    # this means there's another error with agent, e.g. it is not configured
+                    break
+            else:
+                ifaces = json.loads(qemu_ga_check.stdout[0])
+                mgmt_if = [i for i in ifaces['return']
+                           if 'eth0' in i['name'] or 'ens3' in i['name']]
+                if len(mgmt_if) > 0 and mgmt_if[0].get('ip-addresses'):
```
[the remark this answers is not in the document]
rplevka: This is tricky.
If there were both, but the first didn't have the ip address, we still couldn't tell which one is the right one, as it would still be possible, that the 1st interface does not have the IP address YET.
- we should develop some mechanism to correctly tell which one of the interfaces is the management one.
Anyway, your suggestion makes sense. I'll try to implement it

robottelo/vm.py
```diff
@@ -232,19 +235,46 @@ def create(self):
...
+                if 'guest agent is not connected' in qemu_ga_check.stderr:
+                    # this means the agent wasn't started yet (vm still booting)
+                    sleep(1)
+                else:
+                    # this means there's another error with agent, e.g. it is not configured
+                    break
+            else:
+                ifaces = json.loads(qemu_ga_check.stdout[0])
+                mgmt_if = [i for i in ifaces['return']
+                           if 'eth0' in i['name'] or 'ens3' in i['name']]
```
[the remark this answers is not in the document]
rplevka: :D of course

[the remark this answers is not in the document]
lpramuk: +1

tests/robottelo/test_vm.py
lpramuk: we miss test body here?
```
        self.configure_provisioning_server()
        vm = VirtualMachine()
        vm.create()
        self.assertEqual(vm.ip_addr, '10.8.30.135')
```

robottelo/vm.py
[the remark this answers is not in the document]
lpramuk: ```I'll try to implement it``` <- in this PR or in future

I'm for future since this PR just introduced qemu-ga IP discovery and later on we can enhance its logic further

robottelo/vm.py
lpramuk: +1

mirzal: @rplevka for me, the biggest reason is that this way code clearly expresses your intent (which is "filter this list and assign first value to this variable"). Being more Pythonic and shorter are additional benefits.

robottelo/vm.py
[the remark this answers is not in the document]
rplevka: @mirzal @lpramuk - in fact I'll go further and pick the interface with the MAC address returned by snap-guest call.
- In future we want to update snap-guest to allow adding multiple interfaces and returning the MGMT_MAC.

robottelo/vm.py
```diff
@@ -214,37 +218,67 @@ def create(self):
...
+                    # get only the ipv4 addresses
+                    ip4_addr = [i['ip-address'] for i in mgmt_if[0]['ip-addresses']
+                                if i['ip-address-type'] == 'ipv4']
+                    self.ip_addr = next(
+                        (i['ip-address'] for i in mgmt_if[0]['ip-addresses']
+                            if i['ip-address-type'] == 'ipv4'),
+                        None
+                    )
+                    if self.ip_addr:
+                        break
```
mirzal: You create `ip4_addr` variable which you never use.

This whole `else` can be rewritten. Instead of:
```python
            else:
                ifaces = json.loads(qemu_ga_check.stdout[0])
                mgmt_if = [i for i in ifaces['return']
                           if i['hardware-address'].lower() == self.mac.lower()]
                if len(mgmt_if) > 0 and mgmt_if[0].get('ip-addresses'):
                    # get only the ipv4 addresses
                    ip4_addr = [i['ip-address'] for i in mgmt_if[0]['ip-addresses']
                                if i['ip-address-type'] == 'ipv4']
                    self.ip_addr = next(
                        (i['ip-address'] for i in mgmt_if[0]['ip-addresses']
                            if i['ip-address-type'] == 'ipv4'),
                        None
                    )
                    if self.ip_addr:
                        break
```
do:
```python
            else:
                ifaces = json.loads(qemu_ga_check.stdout[0])
                mgmt_if = next((i for i in ifaces['return']
                           if i['hardware-address'].lower() == self.mac.lower()), {})
                try:
                    self.ip_addr = next(i['ip-address'] for i in mgmt_if['ip-addresses']
                            if i['ip-address-type'] == 'ipv4')
                    break
                except (KeyError, StopIteration):
                    continue
```

`KeyError` will be raised if `mgmt_if` doesn't have `ip-addresses` key (because qemu-guest didn't return us interface with appropriate MAC address or returned interface is not assigned any IP address yet). `StopIteration` will be raised if generator comprehension returns empty list (because IPv4 address is not configured). I believe this covers all your cases?

rplevka: nice!

robottelo/vm.py
```diff
@@ -214,37 +218,67 @@ def create(self):
...
+            connection_timeout=30
         )
-
         if result.return_code != 0:
             raise VirtualMachineError(
                 u'Failed to run snap-guest: {0}'.format(result.stderr))
         else:
             self._created = True
-
+            self.mac = [n.split('MAC:')[1].strip() for n in result.stdout if 'MAC:' in n][0]
```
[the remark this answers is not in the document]
rplevka: I would normally use the regexp one, but...yeah, looks quite an overkill here.

rplevka:
@lpramuk mind revisiting this one?

rplevka:
@jhutar the guest images got taken care of by @lpramuk some time ago (https://github.com/SatelliteQE/robottelo/issues/5349) so this should work flawlessly already.
Yes, we could remove the avahi now, however I did not get rid of the code that uses the old method and I keep it as a fallback scenario.

robottelo/vm.py
```diff
@@ -214,37 +218,64 @@ def create(self):
...
+                mgmt_if = next((i for i in ifaces['return']
+                               if i['hardware-address'].lower() == self.mac.lower()), {})
+                try:
+                    # get only the ipv4 addresses
+                    self.ip_addr = next(i['ip-address'] for i in mgmt_if['ip-addresses']
+                                        if i['ip-address-type'] == 'ipv4')
+                    break
+                except(KeyError, StopIteration):
+                    continue
+        if not self.ip_addr:
```
lpramuk: can you pls add one empty line above to indicate end of the for loop

lpramuk: I miss some sleep when ```qemu_ga_check``` command suceeds but there is still no IPv4 adress due to network start-up.

robottelo/vm.py
```diff
@@ -214,37 +218,64 @@ def create(self):
...
+                ifaces = json.loads(qemu_ga_check.stdout[0])
+                mgmt_if = next((i for i in ifaces['return']
+                               if i['hardware-address'].lower() == self.mac.lower()), {})
+                try:
+                    # get only the ipv4 addresses
+                    self.ip_addr = next(i['ip-address'] for i in mgmt_if['ip-addresses']
+                                        if i['ip-address-type'] == 'ipv4')
+                    break
+                except(KeyError, StopIteration):
+                    continue
```
lpramuk: should be:
```
except(KeyError, StopIteration):
    sleep(1)    
    continue
```

lpramuk: hehe ```continue``` at very last line of the loop :-)

robottelo/vm.py
mirzal: It's handled by your previous (but chronologically later) comment. If there is no IPv4 address yet, generator will be empty and `next` will throw `StopIteration` on us.

lpramuk: yes immediately

rplevka:
@lpramuk comment addressed. could you re-re-review? :D

PR #7021, opened 2019-06-12
Environment Variables for upgrade scenarios are now robottelo properties - Phase 2

jyejare:
Changes in Files wrt $subject:

```
- tests/upgrades/test_client.py
- tests/upgrades/test_remoteexecution.py
- tests/upgrades/test_repository.py
- tests/upgrades/test_subscription.py
- tests/upgrades/test_syncplan.py
- tests/upgrades/test_yum_plugins.py
```

Also, Some nitpicks and small improvements wrt manifest upload.

### DIFF
--- a/tests/upgrades/test_client.py
+++ b/tests/upgrades/test_client.py
@@ -14,12 +14,11 @@
 
 :Upstream: No
 """
-import os
 import time
 
 from fabric.api import execute
 from nailgun import entities
-from robottelo.test import APITestCase
+from robottelo.test import APITestCase, settings
 from robottelo.api.utils import (
     attach_custom_product_subscription,
     call_entity_method_with_timeout
@@ -48,10 +47,10 @@ def create_activation_key_for_client_registration(
     :return nailgun.entity.ActivationKey: Activation key
     """
     client_os = client_os.upper()
-    from_ver = os.environ.get('FROM_VERSION')
+    from_ver = settings.upgrade.from_version
     rhel_prod_name = 'scenarios_rhel{}_prod'.format(client_os[-1])
     rhel_repo_name = '{}_repo'.format(rhel_prod_name)
-    rhel_url = os.environ.get('{}_CUSTOM_REPO'.format(client_os))
+    rhel_url = settings.rhel7_os
     if rhel_url is None:
         raise ValueError('The RHEL Repo URL environment variable for OS {} '
                          'is not provided!'.format(client_os))
@@ -95,8 +94,7 @@ def create_activation_key_for_client_registration(
         )[0]
     elif sat_state.lower() == 'post':
         product_name = 'scenarios_tools_product'
-        tools_repo_url = os.environ.get(
-            'TOOLS_{}'.format(client_os.upper()))
+        tools_repo_url = settings.sattools_repo[client_os]
         if tools_repo_url is None:
             raise ValueError('The Tools Repo URL environment variable for '
                              'OS {} is not provided!'.format(client_os))
@@ -248,7 +246,7 @@ class Scenario_upgrade_old_client_and_package_installation(APITestCase):
     """
     @classmethod
     def setUpClass(cls):
-        cls.docker_vm = os.environ.get('DOCKER_VM')
+        cls.docker_vm = settings.upgrade.docker_vm
         cls.default_org_id = entities.Organization().search(
             query={'search': 'name="{}"'.format(DEFAULT_ORG)})[0].id
         cls.org = entities.Organization(id=cls.default_org_id).read()
@@ -372,7 +370,7 @@ class Scenario_upgrade_new_client_and_package_installation(APITestCase):
     """
     @classmethod
     def setUpClass(cls):
-        cls.docker_vm = os.environ.get('DOCKER_VM')
+        cls.docker_vm = settings.upgrade.docker_vm
         cls.org_name = 'new_client_package_install'
         cls.ak_name = 'scenario_new_client_package_install'
         cls.le_name = cls.ak_name+'_env'
--- a/tests/upgrades/test_remoteexecution.py
+++ b/tests/upgrades/test_remoteexecution.py
@@ -14,13 +14,11 @@
 
 :Upstream: No
 """
-import os
-
 from nailgun import entities
 from robottelo.constants import DEFAULT_LOC, DEFAULT_ORG, DISTRO_RHEL7
 from robottelo.helpers import add_remote_execution_ssh_key
 from robottelo.vm import VirtualMachine
-from robottelo.test import APITestCase
+from robottelo.test import APITestCase, settings
 from upgrade_tests import post_upgrade, pre_upgrade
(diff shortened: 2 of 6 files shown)


tests/upgrades/test_remoteexecution.py
```diff
@@ -45,18 +43,18 @@ class Scenario_remoteexecution_external_capsule(APITestCase):
...
-        cls.vm_domain_name = os.environ.get('VM_DOMAIN')
+        cls.bridge = settings.vlan_networking.bridge
+        cls.subnet = settings.vlan_networking.subnet
+        cls.gateway = settings.vlan_networking.gateway
+        cls.netmask = settings.vlan_networking.netmask
+        cls.vm_domain_name = settings.upgrade.vm_domain
         cls.vm_domain = entities.Domain().search(query={'search': 'name="{}"'
                                                  .format(cls.vm_domain_name)})
-        cls.proxy_name = os.environ.get('RHEV_CAP_HOST')
+        cls.proxy_name = settings.upgrade.rhev_cap_host or settings.upgrade.capsule_hostname
```
omkarkhatavkar: `capsule_hostname` exist in robottelo.properties file but need to be sed while running the pipeline, this currently not exists in  https://github.com/SatelliteQE/robottelo-ci/blob/c89f0f2026e94e02672086b90dbebb1b983157a4/scripts/satellite6-upgrade-run-scenarios.sh#L19

tests/upgrades/test_yum_plugins.py
```diff
@@ -89,7 +89,6 @@ def _check_yum_plugins_count(self, client_container_id):
     def _check_package_installed(self, client_container_id, package):
         """Verify if package is installed on docker content host."""
         kwargs = {'host': self.docker_vm}
-        docker_execute_command
```
omkarkhatavkar: why `docker_execute_command` was removed from the argument?

tests/upgrades/test_yum_plugins.py
```diff
@@ -112,10 +111,8 @@ def _install_or_update_package(self, client_container_id, package, update=False)
 
     def _create_custom_rhel_tools_repos(self, product):
         """Install packge on docker content host."""
-        rhel_repo_url = os.environ.get('{}_CUSTOM_REPO'.format(self.client_os.upper()))
-
-        tools_repo_url = os.environ.get(
-            'TOOLS_{}'.format(self.client_os.upper()))
+        rhel_repo_url = settings.rhel7_os
+        tools_repo_url = settings.sattools_repo[DISTRO_RHEL7]
```
omkarkhatavkar: here also we hardcoding to rhel7, earlier it was dynamically generated anything (rhel6,7,8). we need to have some dynamic nature or adding if else would make sense.

tests/upgrades/test_client.py
```diff
@@ -48,10 +47,10 @@ def create_activation_key_for_client_registration(
     :return nailgun.entity.ActivationKey: Activation key
     """
     client_os = client_os.upper()
-    from_ver = os.environ.get('FROM_VERSION')
+    from_ver = settings.upgrade.from_version
     rhel_prod_name = 'scenarios_rhel{}_prod'.format(client_os[-1])
     rhel_repo_name = '{}_repo'.format(rhel_prod_name)
-    rhel_url = os.environ.get('{}_CUSTOM_REPO'.format(client_os))
+    rhel_url = settings.rhel7_os
```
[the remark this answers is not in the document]
jyejare: Again, I will give you the same answer given in my last phase-1 PR.

Actually, we setting an url for thel7 but all the validations are happening for rhel7 only and those are hardcoded.
When we will switch to RHEL8 as default all those tests will fail.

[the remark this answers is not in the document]
jyejare: I know we should do it for rhel8 as well. But then we should run the validations and setup twice, one for rhel7 and another for rhel8.

But that change is not part of this PR. We can handle that separately if we really want to test on rhel7 and rhel8.

tests/upgrades/test_remoteexecution.py
jyejare: This is a completely optional field and you can only use that field while running the tests locally on your local capsule.

In short, its either RHEV_CAP_HOST or CAPSULE_HOSTNAME

tests/upgrades/test_yum_plugins.py
jyejare: This was hanging without being used, I don't see that func call is doing anything. 

Also, I am not sure why r u calling it as an argument. I believe below this line, this func is being correctly called an argument.

tests/upgrades/test_yum_plugins.py
jyejare: Same as above.

tests/upgrades/test_remoteexecution.py
omkarkhatavkar: what if RHEV_CAP_HOST does not exist? you want to use CAPSULE_HOSTNAME if that is never a case then we can make it like this 

`cls.proxy_name = settings.upgrade.rhev_cap_host`

tests/upgrades/test_yum_plugins.py
omkarkhatavkar: my bad I did not notice that.

tests/upgrades/test_remoteexecution.py
jyejare: RHEV_CAP_HOST is there always when we run tests on Jenkins as we run upgrades on RHEV based images.

tests/upgrades/test_client.py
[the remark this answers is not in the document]
omkarkhatavkar: We can create an issue for RHEL8 and handle in separate.

tests/upgrades/test_remoteexecution.py
omkarkhatavkar: this means we don't require `CAPSULE_HOSTNAME` as an additional/optional argument, this where I'm pointing. If that is not true do we set  `CAPSULE_HOSTNAME` in the pipeline anywhere ?.

omkarkhatavkar: you can only use that field while running the tests locally on your local capsule. - sorry I missed this message. In this case, looks ok then.

jyejare:
Test Results:
------------------------

# test_repository.py
```
# py.test -s tests/upgrades/test_repository.py::Scenario_custom_repo_check -m 'pre_upgrade'
platform linux -- Python 3.6.5, pytest-4.0.2, py-1.8.0, pluggy-0.12.0
rootdir: /home/jenkins/workspace/automation-preupgrade-6.6-scenario-tests-rhel7, inifile:
plugins: xdist-1.27.0, forked-1.0.2, mock-1.10.4, services-1.3.1, cov-2.7.1
collected 2 items / 1 deselected
========================= 1 passed, 1 deselected, 6 warnings in 218.49 seconds ==========================
```

vijay8451:
LGTM .. ack

PR #7023, opened 2019-06-13
Remove most of unstable errors for host entity

oshtaier:
Sounds weird, but that is most elegant solution for issue when 'Submit' button is not clicked after 'Network Interface' window is closed. Of course, `ensure_page_safe` or `wait_for(lambda: self.submit.is_displayed is False...` for mentioned window doesn't help. Static sleep can help, but that is not solution for sure. Probably some java script events are not finished properly or whatever. Anyway, as issue specific to only one place, it doesn't make sense spend more time on investigation.
```
py.test /home/ashtayer/Desktop/robottelo/tests/foreman/ui/test_host.py -k 'test_positive_read_from_edit_page or test_positive_read_from_details_page'
==================================================================================== test session starts =====================================================================================
platform linux -- Python 3.6.5, pytest-4.0.2, py-1.5.3, pluggy-0.11.0
rootdir: /home/ashtayer/Desktop/robottelo, inifile:
plugins: services-1.3.1, mock-1.10.0
collecting ... 2019-06-13 11:30:17 - conftest - DEBUG - BZ deselect is disabled in settings

collected 31 items / 29 deselected                                                                                                                                                           

tests/foreman/ui/test_host.py ..                                                                                                                                                       [100%]

========================================================================= 2 passed, 29 deselected in 270.41 seconds ==========================================================================
```

### DIFF
--- a/robottelo/ui/utils.py
+++ b/robottelo/ui/utils.py
@@ -30,6 +30,7 @@ def create_fake_host(session, host, interface_id=gen_string('alpha'),
         'interfaces.interface.interface_additional_data.virtual_nic': False,
         'parameters.global_params': global_parameters,
         'parameters.host_params': host_parameters,
+        'additional_information.comment': 'Host with fake data'
     }
     values.update(extra_values)
     session.host.create(values)


vijay8451:
@oshtaier I am fine  by adding with `Additional Information comment` however would like to know the exact error which comes or maybe a screenshot explain more (I am more worry about Bug if it is a product one)

oshtaier:
Application works fine, but in around 55-65% of cases, after 'Save' button for network interface is clicked, 'Submit' click for 'Create Host' view is not actually clicked. Typical test automation issue. Maybe we have java script is not finished and it is not in `ensure_page_safe` procedure. Investigation for such issue can take weeks and it is definitely not worthwhile

vijay8451:
> @vijay8451
> Application works fine, but in around 55-65% of cases, after 'Save' button for network interface is clicked, 'Submit' click for 'Create Host' view is not actually clicked. Typical test automation issue. Maybe we have java script is not finished and it is not in `ensure_page_safe` procedure. Investigation for such issue can take weeks and it is definitely not worthwhile

okay , so adding the additional `Additional Information comment` make this work perfect ? (I am still thinking a product bug not automation because automation works fine 50%)

oshtaier:
@vijay8451 yes, because it can select tab just fine. I am not sure how you reproduce that behavior manually as issue happened in less than 1 second. And if the issue is a bug, imagine priority of it.

That mean `sleep(1)` can help, but I hate that solution

vijay8451:
> @vijay8451 yes, because it can select tab just fine. I am not sure how you reproduce that behavior manually as issue happened in less than 1 second. And if the issue is a bug, imagine priority of it.
> 
> That mean `sleep(1)` can help, but I hate that solution
I have seen same issue in upstream foreman too but that occur mutiple times and give a time(4-5 sec) time window too .
But as you mentioned it comes for one second then that might be not ...

oshtaier:
@vijay8451 if you see any generic pattern, please create an issue with details. There is always a chance that we need expand `ensure_page_safe` functionality or anything else

PR #7034, opened 2019-06-17
Make sure only plain RHEL subscription is used in this test

jhutar:
This test is failing, because auto attach choose subscription which was also providing Capsule.

```
$ pytest tests/foreman/cli/test_host.py::HostSubscriptionTestCase::test_negative_without_attach_with_lce --no-print-logs --capture=no --exitfirst 2>&1 | tee /tmp/pytest.log
[...]
==================== 1 passed, 7 warnings in 256.68 seconds ====================
```

### DIFF
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -70,6 +70,7 @@ from robottelo.constants import (
     CUSTOM_PUPPET_REPO,
     DEFAULT_CV,
     DISTRO_RHEL7,
+    DEFAULT_SUBSCRIPTION_NAME,
     ENVIRONMENT,
     FAKE_0_CUSTOM_PACKAGE,
     FAKE_0_CUSTOM_PACKAGE_GROUP,
@@ -3062,7 +3063,7 @@ class HostSubscriptionTestCase(CLITestCase):
         self.client.install_katello_ca()
 
     def _register_client(self, activation_key=None, lce=False,
-                         enable_repo=False, auto_attach=False):
+                         enable_repo=False, auto_attach=False, attach_to_default=False):
         """Register the client as a content host consumer
 
         :param activation_key: activation key if registration with activation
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_host.py
```diff
@@ -3073,8 +3074,13 @@ def _register_client(self, activation_key=None, lce=False,
         :param auto_attach: boolean to indicate whether to register with
             auto-attach option, in case of registration with activation key a
             command is launched
+        :param attach_to_default: boolean to indicate whether to attach to
+            plain RHEL subsctiption
         :return: the registration result
         """
+        assert not auto_attach or not attach_to_default, \
+            'Only one of auto_attach or attach_to_default can be set'
```
vijay8451: 1. 
```suggestion
            'Only one of auto_attach or attach_to_default must be set'
```
2. Due to above validation `self._register_client` will not work for other tests where it used i.e. https://github.com/SatelliteQE/robottelo/blob/cc4a6efe766fbe68eef99646c300287d46b7dcc6/tests/foreman/cli/test_host.py#L3208
please check
3. Why do we require it ?

jhutar: If none of these is set, assert should pass:
```
>>> auto_attach=True; attach_to_default=True; not auto_attach or not attach_to_default
False
>>> auto_attach=False; attach_to_default=True; not auto_attach or not attach_to_default
True
>>> auto_attach=True; attach_to_default=False; not auto_attach or not attach_to_default
True
>>> auto_attach=False; attach_to_default=False; not auto_attach or not attach_to_default
True
```

jhutar: Why is this needed: Test auto attaches and then attempts to enable `rhel-7-server-satellite-capsule-6.2-rpms` and asserts that repo is not enabled (because normal RHEL subscription does not provide capsule). Problem was auto attach was giving me subscription that had that entitlement, so repo enable passed. This way I explicitly request RHEL subscription, making test result more stable.

jhutar:
<jhutar> Vijay_Singh, Hello! Please does https://github.com/SatelliteQE/robottelo/pull/7034 look OK now?
<Vijay_Singh> jhutar, lmc
<Vijay_Singh> jhutar, ack
<jhutar> Vijay_Singh++ Thank you

PR #7035, opened 2019-06-17
Switch to using functools.lru_cache directly.

pgagne:
When trying to import robotello in another project I ran into an
import error cause by lru_cache in helpers .

Looks like we historically had a conditional around this to get the  cachetools version in python 2, however that was removed. Did a search and it seems the only place its used is `robottelo/host_info.py` so I changed it to import it directly from functools.

### DIFF
--- a/robottelo/host_info.py
+++ b/robottelo/host_info.py
@@ -1,15 +1,14 @@
 """Module that gather several informations about host"""
+import functools
 import logging
-
 import re
 from robottelo.cli.base import CLIReturnCodeError
-from robottelo.helpers import lru_cache
 
 from robottelo import ssh
 LOGGER = logging.getLogger(__name__)
 
 
-@lru_cache(maxsize=1)
+@functools.lru_cache(maxsize=1)
 def get_host_os_version():
     """Fetches host's OS version through SSH
     :return: str with version
@@ -37,7 +36,7 @@ _SAT_6_1_VERSION_COMMAND = (
 )
 
 
-@lru_cache(maxsize=1)
+@functools.lru_cache(maxsize=1)
 def get_host_sat_version():
     """Fetches host's Satellite version through SSH
     :return: Satellite version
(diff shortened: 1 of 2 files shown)


robottelo/host_info.py
```diff
@@ -1,15 +1,14 @@
 """Module that gather several informations about host"""
+import functools
```
mirzal: Why not `from functools import lru_cache`?

I mean - theoretical performance impact is probably negligible in context of whole project, but I wonder why you decided to import entire module instead of just part that you need.

pgagne: IDK I just thought it looked cleaner and more explicit that way. IMHO we do a little to much from x import y or *.

pgagne:
> Can you provide a little longer explanation why this change is needed? Was the other project using Python3? What was the error that you got?

The project was python3. TBH I didn't have much time to trace down why exactly the import was failing, other then to confirm that removing it fixed my issue. @JacobCallahan and I were looking at it and both decided it was probably better to take `lru_cache` out of helpers since there is no reason to have a conditional around it anymore.

robottelo/host_info.py
JacobCallahan: Explicit is better than Implicit. Basically, it allows the reader to better understand what is the scope of lru_cache without always having to reference the import.
Also, python loads the entire functools module even if importing lru_cache, so there is no impact to performance. The only change would be the namespace of functools.lru_cache vs lru_cache

JacobCallahan:
@mirzal The issue seems to occur only when robottelo is installed as a module on a system, not when running in our normal method.
It looks like, when installed, the import of lru_cache from robottelo.helpers is failing, as a sub-package likely isn't installed correctly.

We then looked through the code and found that it was literally a direct import from functools and then was only used in this one file.

After that, we think it is better to remove the inclusion in robottelo.helpers as it only makes things more confusing (someone trying to find out what lru_cache is in host_info.py would have to go to helpers.py to find out that it is actually a method in functools).

JacobCallahan:
Not to be too noisy, but I just looked at the history of the helpers file. We only had lru_cache in here from the days that we needed to support python 2. At the time, lru_cache had to be imported from different modules, depending on the version. In that case, it made sense to have the required reference in helpers. Now we're strictly python 3, so no need for the import check.

PR #7041, opened 2019-06-18
[6.6 Feature] GCE CR End To End Scenario

jyejare:
Test Result:

```
$ py.test -s tests/foreman/ui/test_computeresource_gce.py
2019-06-18 20:41:54 - conftest - DEBUG - Registering custom pytest_namespace

platform linux -- Python 3.6.8, pytest-4.0.2, py-1.7.0, pluggy-0.8.1
rootdir: /home/jitendrayejare/Desktop/RedHat/RoboTelloNew/robottelo, inifile:
plugins: xdist-1.26.1, services-1.3.1, mock-1.10.4, forked-1.0.2, cov-2.6.1
collecting ... 2019-06-18 20:41:54 - conftest - DEBUG - BZ deselect is disabled in settings

collected 1 item 

================================ 1 passed, 2 warnings in 524.81 seconds =================================
```

### DIFF
--- a/robottelo.properties.sample
+++ b/robottelo.properties.sample
@@ -388,6 +388,18 @@ port_range=9091, 14999
 # default value is private
 # managed_ip=Private
 
+#[gce]
+# Google Provider as Compute Resource
+# Project name in Google provider
+#project_id=sat6qe
+# Service Account email id which has compute admin permission
+#client_email=
+# client json Certificate path which is local path on satellite
+#cert_path=/usr/share/foreman/sat6qetester_key.json
+# Zones
+#zone=
+# client certificate URL
+#cert_url=
 
 # [rhev]
 # RHEV to be added as a compute resource.
--- a/robottelo/config/base.py
+++ b/robottelo/config/base.py
@@ -20,6 +20,7 @@ import airgun.settings
 from nailgun import entities, entity_mixins
 from nailgun.config import ServerConfig
 from robottelo.config import casts
+from robottelo.constants import VALID_GCE_ZONES
 
 LOGGER = logging.getLogger(__name__)
 SETTINGS_FILE_NAME = 'robottelo.properties'
@@ -511,6 +512,43 @@ class FakeManifestSettings(FeatureSettings):
             validation_errors.append(
                 'URL with key "default" is required if multiple URLs are provided'
             )
+        return validation_errors
+
+
+class GCESettings(FeatureSettings):
+    """Google Compute Engine settings definitions."""
+
+    def __init__(self, *args, **kwargs):
+        super(GCESettings, self).__init__(*args, **kwargs)
+        self.project_id = None
+        self.client_email = None
+        self.cert_path = None
+        self.zone = None
+        self.cert_url = None
+
+    def read(self, reader):
+        """Read GCE settings."""
+        self.project_id = reader.get('gce', 'project_id')
+        self.client_email = reader.get('gce', 'client_email')
+        self.cert_path = reader.get('gce', 'cert_path')
+        self.zone = reader.get('gce', 'zone')
+        self.cert_url = reader.get('gce', 'cert_url')
+
+    def validate(self):
+        """Validate GCE settings."""
+        valid_cert_path = '/usr/share/foreman/'
+        validation_errors = []
+        if not all(self.__dict__.values()):
+            validation_errors.append(
+                'All [gce] {} options must be provided'.format(self.__dict__.keys()))
+        if not str(self.cert_path).startswith(valid_cert_path):
+            validation_errors.append(
+                '[gce] cert_path - cert should be available '
+                'from satellites {}'.format(valid_cert_path))
+        if self.zone not in VALID_GCE_ZONES:
+            validation_errors.append(
+                'Invalid [gce] zone - {0}, The zone should be one of {1}'.format(
+                    self.zone, VALID_GCE_ZONES))
         return validation_errors
 
 
@@ -1122,6 +1160,7 @@ class Settings(object):
         self.ec2 = EC2Settings()
         self.fake_capsules = FakeCapsuleSettings()
         self.fake_manifest = FakeManifestSettings()
+        self.gce = GCESettings()
         self.ldap = LDAPSettings()
         self.ipa = LDAPIPASettings()
         self.oscap = OscapSettings()
--- a/tests/foreman/ui/test_computeresource_gce.py
+++ b/tests/foreman/ui/test_computeresource_gce.py
@@ -1,0 +1,159 @@
+"""Test for Compute Resource UI
+
+:Requirement: ComputeResources GCE
+
+:CaseAutomation: Automated
+
+:CaseLevel: Acceptance
+
+:CaseComponent: ComputeResourcesGCE
+
+:TestType: Functional
+
+:CaseImportance: High
+
+:Upstream: No
+"""
+from fauxfactory import gen_string
+from nailgun import entities
+from pytest import skip
+from robottelo import ssh
+from robottelo.config import settings
+from robottelo.constants import (
+    COMPUTE_PROFILE_SMALL,
+    FOREMAN_PROVIDERS,
+    GCE_EXTERNAL_IP_DEFAULT,
+    GCE_MACHINE_TYPE_DEFAULT,
+    GCE_NETWORK_DEFAULT
+)
+from robottelo.decorators import bz_bug_is_open, fixture, setting_is_set, tier2, upgrade
+
+
+if not setting_is_set('gce'):
+    skip('skipping tests due to missing gce settings', allow_module_level=True)
+
+
+class GCECertNotFoundError(Exception):
+    """An exception to raise when GCE Cert json is not available for creating GCE CR"""
+
+
+@fixture(scope='module')
+def module_org():
+    return entities.Organization().create()
+
+
+@fixture(scope='module')
+def module_loc():
+    return entities.Location().create()
+
+
+@fixture(scope='module')
+def module_gce_settings():
+    return dict(
+        project_id=settings.gce.project_id,
+        client_email=settings.gce.client_email,
+        cert_path=settings.gce.cert_path,
+        zone=settings.gce.zone
+    )
+
+
+@fixture(scope='module')
+def download_cert():
+    ssh.command('curl {0} -o {1}'.format(settings.gce.cert_url, settings.gce.cert_path))
+    if not ssh.command('[ -f {} ]'.format(settings.gce.cert_path)).return_code == 0:
+        raise GCECertNotFoundError("The GCE certificate {} is not found in satellite.".format(
+            settings.gce.cert_path))
+
+
+@tier2
+@upgrade
+def test_positive_default_end_to_end_with_custom_profile(
+        session, module_org, module_loc, module_gce_settings, download_cert):
+    """Create GCE compute resource with default properties and apply it's basic functionality.
+
+    :id: 59ffd83e-a984-4c22-b91b-cad055b4fbd7
+
+    :Steps:
+
+        1. Create an GCE compute resource with default properties.
+        2. Update the compute resource name and add new taxonomies.
+        3. Associate compute profile with custom properties to GCE compute resource
+        4. Delete the compute resource.
+
+    :expectedresults: The GCE compute resource is created, updated, compute profile associated and
+        deleted.
+
+    :CaseLevel: Integration
+
+    :CaseImportance: Critical
+    """
+    cr_name = gen_string('alpha')
+    new_cr_name = gen_string('alpha')
+    cr_description = gen_string('alpha')
+    new_org = entities.Organization().create()
+    new_loc = entities.Location().create()
+
+    with session:
+        # Compute Resource Create and Assertions
+        session.computeresource.create({
+            'name': cr_name,
+            'description': cr_description,
+            'provider': FOREMAN_PROVIDERS['google'],
+            'provider_content.google_project_id': module_gce_settings['project_id'],
+            'provider_content.client_email': module_gce_settings['client_email'],
+            'provider_content.certificate_path': module_gce_settings['cert_path'],
+            'organizations.resources.assigned': [module_org.name],
+            'locations.resources.assigned': [module_loc.name],
+        })
+        cr_values = session.computeresource.read(cr_name)
+        assert cr_values['name'] == cr_name
+        assert cr_values['provider_content']['zone']['value']
+        assert (cr_values['organizations']['resources']['assigned']
+                == [module_org.name])
+        assert (cr_values['locations']['resources']['assigned']
+                == [module_loc.name])
+        assert cr_values['provider_content']['google_project_id'] == module_gce_settings[
+            'project_id']
+        assert cr_values['provider_content']['client_email'] == module_gce_settings[
+            'client_email']
+        # Compute Resource Edit/Updates and Assertions
+        session.computeresource.edit(cr_name, {
+            'name': new_cr_name,
+            'organizations.resources.assigned': [new_org.name],
+            'locations.resources.assigned': [new_loc.name],
+        })
+        assert not session.computeresource.search(cr_name)
+        cr_values = session.computeresource.read(new_cr_name)
+        assert cr_values['name'] == new_cr_name
+        assert (set(cr_values['organizations']['resources']['assigned'])
+                == {module_org.name, new_org.name})
+        assert (set(cr_values['locations']['resources']['assigned'])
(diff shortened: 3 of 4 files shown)


jyejare:
Depends on https://github.com/SatelliteQE/airgun/pull/340

jyejare:
Fixing Travis guys ............. :1234:

robottelo.properties.sample
```diff
@@ -388,6 +388,16 @@ port_range=9091, 14999
 # default value is private
 # managed_ip=Private
 
+#[gce]
+# Google Provider as Compute Resource
+# Project name in Google provider
+#project_id=sat6qe
```
vijay8451: any specific reason to keep `sat6qe` value for `project_id` ?

robottelo.properties.sample
```diff
@@ -388,6 +388,16 @@ port_range=9091, 14999
...
+# Service Account email id which has compute admin permission
+#client_email=
+# client json Certificate path which is local path on satellite
+#cert_path=
+# Zones
+#zone=
```
vijay8451: to add support for above , requires to raise a PR at `robottelo-ci`  as well.

robottelo/config/base.py
```diff
@@ -514,6 +514,43 @@ def validate(self):
...
+            'us-east4-a', 'us-east4-b', 'us-east4-c']
+        validation_errors = []
+        if not all(self.__dict__.values()):
+            validation_errors.append(
+                'All [gce] {} options must be provided'.format(self.__dict__.keys()))
+        if not str(self.cert_path).startswith(valid_cert_path):
+            validation_errors.append(
+                '[gce] cert_path - cert should be available '
+                'from satellites {}'.format(valid_cert_path))
+        if str(self.zone).startswith('us') and self.zone not in valid_zones:
```
vijay8451:
```suggestion
        if self.zone not in valid_zones:
```
I think above is fine to validate the zone .. as `True` will only return on complete match  else it would remain `False`

tests/foreman/ui/test_computeresource_gce.py
```diff
@@ -0,0 +1,145 @@
+"""Test for Compute Resource UI
+
+:Requirement: Computeresource RHV
```
vijay8451:
```suggestion
:Requirement: Computeresource GCE
```

tests/foreman/ui/test_computeresource_gce.py
```diff
@@ -0,0 +1,145 @@
...
+            'organizations.resources.assigned': [module_org.name],
+            'locations.resources.assigned': [module_loc.name],
+        })
+        cr_values = session.computeresource.read(cr_name)
+        assert cr_values['name'] == cr_name
+        assert cr_values['provider_content']['zone']['value']
+        assert (cr_values['organizations']['resources']['assigned']
+                == [module_org.name])
+        assert (cr_values['locations']['resources']['assigned']
+                == [module_loc.name])
```
vijay8451: Can you please add the assert for `google_project_id` `client_email` as well .

robottelo/config/base.py
```diff
@@ -514,6 +514,43 @@ def validate(self):
...
+            validation_errors.append(
+                '[gce] zone - The us zone should be one of {}'.format(valid_zones))
```
omkarkhatavkar: no need to characterize to be 'us' zone. 
```suggestion
                '[gce] zone - The zone should be one of {}'.format(valid_zones))
```

tests/foreman/ui/test_computeresource_gce.py
```diff
@@ -0,0 +1,145 @@
...
+
+@fixture(scope='module')
+def module_org():
+    return entities.Organization().create()
+
+
+@fixture(scope='module')
+def module_loc():
+    # default_loc_id = entities.Location().search(
+    #    query={'search': 'name="{}"'.format(DEFAULT_LOC)})[0].id
```
omkarkhatavkar: no need for these comments

robottelo.properties.sample
jyejare: No specific reason, its a project created in GCE and its the name of a project there and not being created from Satellite.

robottelo/config/base.py
jyejare: There are many zones, so I don't want to get all of them stored under valid_zones. But as we tend to use more of us zones, I kept validation for us zones only.

tests/foreman/ui/test_computeresource_gce.py
jyejare: Forgot to remove it!

robottelo/config/base.py
jyejare: Because we are mostly tend to use us zones(and all the GCE tests will use us zones as our jenkins is there) and we have if condition for us only.

robottelo/config/base.py
vijay8451: @jyejare Thanks , But this `self.zone not in valid_zones` validation will auto takecare the `us` one because you only have these in your `valid_zones`

robottelo.properties.sample
vijay8451: @jyejare could you please remove that from here as that is more looking a harcode way for same.

jyejare: Its not mandatory to keep everything private, this value serves as a default value to project_id param.

robottelo/config/base.py
jyejare: But what if somebody changes it and give something other than 'us'. We won't check for any zones other than 'us' ones.
If I wont give this condition 'if str(self.zone).startswith('us')' the test will fail there itself saying the zone is not in valid zones, though it will be a valid zone.

robottelo.properties.sample
jyejare: Raised PR in gitlab PR number 231.

robottelo/config/base.py
vijay8451: on fail it would gives the clear hint around valid zones  `[gce] zone - The zone should be one of {}'.format(valid_zones))` .... Am i missing something ?

jyejare:
@omkarkhatavkar @vijay8451 , All your comments addressed. Also, I added the download GCE cert support which I forgot in 1st commit. This support makes GCE cert available in satellite to create CR successfully.

robottelo/config/base.py
jyejare: @vijay8451 , As I said in my first comment that there are many zones and as many as 20-25, but I have this validation just for 'us' based zones.

I don't want to keep the 20-25 list of zones in valid-zones as we won't be using or rarely use other zones than 'us' ones.

tests/foreman/ui/test_computeresource_gce.py
```diff
@@ -81,6 +91,7 @@ def test_positive_default_end_to_end_with_custom_profile(
     cr_description = gen_string('alpha')
     new_org = entities.Organization().create()
     new_loc = entities.Location().create()
+    download_cert
```
vijay8451: Are you trying to create above as var ?

jyejare: No ... its the fixture I am calling : https://github.com/SatelliteQE/robottelo/pull/7041/files#diff-90cf96329c7bb8b7528a8678e8224b06R61

vijay8451: as discussed , it already calling here:  https://github.com/SatelliteQE/robottelo/blob/317ea21e1cbe94a00c32edcb236d993a75988e03/tests/foreman/ui/test_computeresource_gce.py#L70
please remove from here .

vijay8451: addressed in latest push

jyejare:
Travis is pass.

Result:
```
$ py.test -s tests/foreman/ui/test_computeresource_gce.py
2019-06-25 18:21:23 - conftest - DEBUG - Registering custom pytest_namespace

platform linux -- Python 3.6.8, pytest-4.0.2, py-1.7.0, pluggy-0.8.1
rootdir: /home/jitendrayejare/Desktop/RedHat/RoboTelloNew/robottelo, inifile:
plugins: xdist-1.26.1, services-1.3.1, mock-1.10.4, forked-1.0.2, cov-2.6.1
collecting ... 2019-06-25 18:21:23 - conftest - DEBUG - BZ deselect is disabled in settings

collected 1 item 

================================ 1 passed, 2 warnings in 640.11 seconds =================================

```

robottelo/config/base.py
```diff
@@ -514,6 +514,45 @@ def validate(self):
...
+            'us-east4-a', 'us-east4-b', 'us-east4-c']
+        validation_errors = []
+        if not all(self.__dict__.values()):
+            validation_errors.append(
+                'All [gce] {} options must be provided'.format(self.__dict__.keys()))
+        if not str(self.cert_path).startswith(valid_cert_path):
+            validation_errors.append(
+                '[gce] cert_path - cert should be available '
+                'from satellites {}'.format(valid_cert_path))
+        if str(self.zone).startswith('us') and self.zone not in valid_zones:
```
mirzal: So, if I set `zone` to `sfdafasf`, or don't set it at all, I won't get any validation error?

As @vijay8451 , I think we should just check if `self.zone` is in `valid_zones`. `valid_zones` means "zones valid from Robottelo point of view". They can be subset of all zones valid from Google point of view.

tests/foreman/ui/test_computeresource_gce.py
```diff
@@ -0,0 +1,158 @@
...
+)
+from robottelo.decorators import bz_bug_is_open, fixture, setting_is_set, tier2
+
+
+if not setting_is_set('gce'):
+    skip('skipping tests due to missing gce settings', allow_module_level=True)
+
+
+class GCECertNotFoundError(Exception):
+    """An exception to raise when GCE Cert json is not available for creating GCE CR"""
```
mirzal: Is there a reason to create and raise custom exception here?

I see similar pattern is employed in `tests/foreman/cli/test_satellitesync.py`, but usually we raise one of existing exceptions, or simply `Exception`.

jyejare:
@JacobCallahan @rplevka @ntkathole Can you pitch in to review/ACK as tier2 reviewer ?

robottelo/config/base.py
jyejare: > So, if I set `zone` to `sfdafasf`, or don't set it at all, I won't get any validation error?

If you dont set the validation above will catch it and ask you to dont keep it empty and the zone has to be provided.

> As @vijay8451 , I think we should just check if `self.zone` is in `valid_zones`. `valid_zones` means "zones valid from Robottelo point of view". They can be subset of all zones valid from Google point of view.

Here, Let's take an example, there is a zone called 'asia-south1-a' actually present in GCE. And I have not listed that one in my valid_zones. So, according to the condition ``` if self.zone in valid_zones``` this becomes an invalid zone, which is not invalid zone at all.

We definitely don't validate zones/cloud properties like ```sfdafasf``` anywhere in our automation unless we have all them known or listed explicitly.

I have three options to choose guys:
1. Keep validation only for 'us' and as it is.
2. List all the valid zones from google.
3. Remove the zones validation at all.

tests/foreman/ui/test_computeresource_gce.py
jyejare: No special reason, but we get to know the issue by looking at the exception name itself.

This kind of exceptions are being employed from ages in satellite6 QE and I am very used to it :)

robottelo/config/base.py
mirzal: > > So, if I set `zone` to `sfdafasf`, or don't set it at all, I won't get any validation error?
> 
> If you dont set the validation above will catch it and ask you to dont keep it empty and the zone has to be provided.

Ah, it's caught by line 543. Alright.

> Here, Let's take an example, there is a zone called 'asia-south1-a' actually present in GCE. And I have not listed that one in my valid_zones. So, according to the condition ` if self.zone in valid_zones` this becomes an invalid zone, which is not invalid zone at all.

On the other hand, with current code zone called 'asia-south10-z' will make it through, while this is something that looks like it could be a real zone, but is not.

For me, letting invalid value through is more serious problem than rejecting value that is technically valid, but never used by the project.

Anyway, looking at https://cloud.google.com/compute/docs/regions-zones/ , I see they have 20 regions, each one with three zones (and one with four zones). Do we know how often regions and zones change? Is this something that we would have to maintain often, or can we just update it every year, or twice a year?

If maintenance burden is low, I would vote for listing all valid zones. `robottelo/constants.py` seems like good place to hold that data.

Personally, I would go as far as generating constant data structure dynamically to ease maintenance and save typing. Something like that:
```python
__zones = {'us-west1': ['a', 'b', 'c'], 'europe-west1': ['b','c','d']}  # and others
VALID_GCE_ZONES = [f'{loc}-{zone}' for loc, zones in zones.items() for zone in __zones]
```

jyejare:
@mirzal , There are as much as 61 zones in GCE. Just checked.

mirzal:
@jyejare `robottelo/constants.py` already has monstrosities like `BOOKMARK_ENTITIES` (over 100 lines of code) or `PERMISSIONS` (over 400 lines, mostly duplicated in `PERMISSIONS_UI` and `PERMISSIONS_WITH_BZ`), so list of 61 items doesn't sound that bad for me :) . Especially as we can generate it in the loop.

jyejare:
Changes are made @mirzal :+1: 

Here is an error for wrong zone:
```
robottelo.config.base.ImproperlyConfigured: Failed to validate the configuration, check the message(s):
Invalid [gce] zone - us-east1-k, The zone should be one of ['us-east1-b', 'us-east1-c', 'us-east1-d', ........ 
```

With correct zone is proceeds :
```
py.test tests/foreman/ui/test_computeresource_gce.py
platform linux -- Python 3.6.8, pytest-4.0.2, py-1.7.0, pluggy-0.8.1
rootdir: /home/jitendrayejare/Desktop/RedHat/RoboTelloNew/robottelo, inifile:
plugins: xdist-1.26.1, services-1.3.1, mock-1.10.4, forked-1.0.2, cov-2.6.1
collecting ... 2019-06-27 17:54:03 - conftest - DEBUG - BZ deselect is disabled in settings

collected 1 item                                                                                        

tests/foreman/ui/test_computeresource_gce.py
```

mirzal:
Thanks, @jyejare !

But we still need Tier2 reviewer ACK. Adding @JacobCallahan @rplevka @ntkathole

tests/foreman/ui/test_computeresource_gce.py
```diff
@@ -0,0 +1,158 @@
+"""Test for Compute Resource UI
+
+:Requirement: Computeresource GCE
+
+:CaseAutomation: Automated
+
+:CaseLevel: Acceptance
+
+:CaseComponent: UI
```
ntkathole: Please add casecomponent as per new metadata change.

tests/foreman/ui/test_computeresource_gce.py
```diff
@@ -0,0 +1,158 @@
...
+
+@fixture(scope='module')
+def download_cert():
+    ssh.command('curl {0} -o {1}'.format(settings.gce.cert_url, settings.gce.cert_path))
+    if not ssh.command('[ -f {} ]'.format(settings.gce.cert_path)).return_code == 0:
+        raise GCECertNotFoundError("The GCE certificate {} is not found in satellite.".format(
+            settings.gce.cert_path))
+
+
+@tier2
```
ntkathole: Would you like to add upgrade decorator for this?

tests/foreman/ui/test_computeresource_gce.py
```diff
@@ -0,0 +1,158 @@
...
+        zone=settings.gce.zone
+    )
+
+
+@fixture(scope='module')
+def download_cert():
+    ssh.command('curl {0} -o {1}'.format(settings.gce.cert_url, settings.gce.cert_path))
+    if not ssh.command('[ -f {} ]'.format(settings.gce.cert_path)).return_code == 0:
+        raise GCECertNotFoundError("The GCE certificate {} is not found in satellite.".format(
+            settings.gce.cert_path))
```
JacobCallahan: could you rewrite this to store the error output and include it in the exception?

jyejare: @JacobCallahan There wont be any stdout/stderr on checking if file exists or not with ```[ -f cert_path ]```. And -f does only checks for file existence and hence using this custom exception.

You want me to completely remove -f style file check ?

jyejare:
@ntkathole , Thanks for valuable updates, comments are addressed.

jyejare:
Ping @JacobCallahan ??

JacobCallahan:
ACK from me. Sorry for the delay.

PR #7047, opened 2019-06-20
Hammer-csv automation tests removed

vijay8451:
- Removed automation as feature `hammer csv` is removed under https://bugzilla.redhat.com/show_bug.cgi?id=1670092


tests/foreman/cli/test_subscription.py
```diff
@@ -244,479 +228,3 @@ def test_negative_manifest_refresh(self):
             Subscription.refresh_manifest({
                 'organization-id': self.org['id'],
             })
-
-    @tier4
-    def test_positive_restore_ak_and_content_hosts_subscriptions(self):
```
jyejare: I wonder why these test has been removed as part of hammer-csv automation tests removal ?

Cant we just remove CSV part from it and keep the entire test as it is.

tests/foreman/cli/test_subscription.py
```diff
@@ -244,479 +228,3 @@ def test_negative_manifest_refresh(self):
...
-                    'id': activation_key['id'],
-                    'host-id': host['id'],
-                }, output_format='json')
-                self.assertIn(
-                    DEFAULT_SUBSCRIPTION_NAME,
-                    [sub['name'] for sub in host_subscriptions]
-                )
-
-    @tier4
-    def test_positive_restore_content_hosts_with_modified_subscription(self):
```
jyejare: same as above and all the below removed tests ?

tests/foreman/cli/test_subscription.py
vijay8451: @jyejare the complete test steps are depended on `hammer-csv` hence not any worth this to keep it.

tests/foreman/cli/test_subscription.py
vijay8451: same as above.

tests/foreman/cli/test_subscription.py
jyejare: So isn't there anything that can take place of  CSV (```I mean just replace CSV with something similar```)

vijay8451: @jyejare good question ... As I have received the Bug to remove it however that does not have any answer to replace this one.
Also when I checked the features around sat 6.6 did not find anything coming as replacement.
So my ans is no (I am not sure if anything +/-).

Issue #6687, opened 2019-01-28
[UI] Fix test_positive_delete_with_ad


ntkathole:
https://github.com/SatelliteQE/robottelo/blob/eaa097ae1102e7bf0c545eb1ae5883b83081720c/tests/foreman/ui_airgun/test_ldap_authentication.py#L119

Test is trying to read table, but table will not present as created ldap auth is deleted. Reproducible when satellite is having no ldap auth configured before running this test.

```
> assert not session.ldapauthentication.read_table_row(name)

tests/foreman/ui_airgun/test_ldap_authentication.py:119: 

> raise NoSuchElementException('Could not find an element {}'.format(repr(locator)))
E selenium.common.exceptions.NoSuchElementException: Message: Could not find an element Locator(by='xpath', locator='.//table')
```

PR #7049, opened 2019-06-20
Fixes ldap authentication UI tests

ntkathole:
Depends on https://github.com/SatelliteQE/airgun/pull/343
Fixes https://github.com/SatelliteQE/robottelo/issues/6687
We have only one server and ldap auth configured in any other org in satellite was causing the test failure. The solution implemented will delete all ldap auth before test starts.

### DIFF
--- a/tests/foreman/ui/test_ldap_authentication.py
+++ b/tests/foreman/ui/test_ldap_authentication.py
@@ -17,7 +17,6 @@
 from fauxfactory import gen_url
 from navmazing import NavigationTriesExceeded
 from pytest import raises, skip
-from widgetastic.exceptions import NoSuchElementException
 
 from airgun.session import Session
 from nailgun import entities
@@ -31,7 +30,6 @@ from robottelo.constants import (
 )
 from robottelo.datafactory import gen_string
 from robottelo.decorators import (
-    bz_bug_is_open,
     fixture,
     run_in_one_thread,
     setting_is_set,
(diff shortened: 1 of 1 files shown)


ntkathole:
```
pytest tests/foreman/ui/test_ldap_authentication.py
============================================================== test session starts ===============================================================
platform linux -- Python 3.6.3, pytest-4.0.2, py-1.8.0, pluggy-0.12.0
rootdir: /home/nkathole/satQE/robottelo, inifile:
collecting ... 2019-06-20 20:04:31 - conftest - DEBUG - BZ deselect is disabled in settings

collected 13 items                                                                                                                               

tests/foreman/ui/test_ldap_authentication.py .............                                                                                 [100%]

==================================================== 13 passed in 2302.03 seconds ===================================================
```

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -126,8 +124,19 @@ def ldap_usergroup_name():
         user_groups[0].delete()
 
 
+@fixture()
```
oshtaier: as far as I see in documentation, default scope for fixtures is a function. That means, if you run 4 or 8 tests in parallel, you can delete all ldaps from other tests in run-time. And that can consequently fail all of them

ntkathole: All ldap tests always run in one thread as we won't have multiple ldap servers to test. https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/ui/test_ldap_authentication.py#L44

vijay8451: @ntkathole thanks , however it would be much great to set it `@fixture(scope='module')` as that will help to fix the scope of fixture too.

ntkathole: Fixed!

PR #7050, opened 2019-06-20
Metadata update for content view and content management

latran:
Add more metadata update from my component ownership for content view and content management.

### DIFF
--- a/tests/foreman/api/test_contentmanagement.py
+++ b/tests/foreman/api/test_contentmanagement.py
@@ -4,9 +4,9 @@
 
 :CaseAutomation: Automated
 
-:CaseLevel: Acceptance
+:CaseLevel: Component
 
-:CaseComponent: API
+:CaseComponent: ContentManagement
 
 :TestType: Functional
 
@@ -53,6 +53,7 @@ from robottelo.decorators import (
     skip_if_bug_open,
     skip_if_not_set,
     tier2,
+    tier3,
     tier4,
 )
 from robottelo.helpers import (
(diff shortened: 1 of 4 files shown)


tests/foreman/api/test_contentmanagement.py
```diff
@@ -88,6 +89,8 @@ def test_positive_sync_repos_with_large_errata(self):
 
         :CaseLevel: Integration
 
+        :CaseImportance: High
```
ntkathole: We have it at https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/api/test_contentmanagement.py#L13. I guess no need to add it for every test then.

tests/foreman/api/test_contentviewfilter.py
```diff
@@ -103,6 +107,8 @@ def test_positive_create_erratum_with_name(self):
             correct name and type
 
         :CaseLevel: Integration
+
+        :CaseImportance: High
```
ntkathole: Same here, we have it above https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/api/test_contentviewfilter.py#L17

tests/foreman/api/test_contentviewversion.py
```diff
@@ -160,6 +166,8 @@ def test_positive_promote_out_of_sequence_environment(self):
         :expectedresults: The promotion succeeds.
 
         :CaseLevel: Integration
+
+        :CaseImportance: High
```
ntkathole: Same for this file as well https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/api/test_contentviewversion.py#L13

latran:
Removed CaseImportance: High redundancy based on suggestions.

PR #7053, opened 2019-06-21
Add upgrade test for foreman_docker removal

mirzal:
My first upgrade test. Can we somehow run it on our CI environment first to ensure assumptions about system configuration are correct?

What I did:
1. Install Satellite 6.5 GA
2. Run pre-upgrade script:
```
$ pytest --pdb -v -m pre_upgrade tests/upgrades/test_containers_removal.py::Scenario_containers_support_removal::test_pre_scenario_containers_support_removal
platform linux -- Python 3.6.3, pytest-4.0.2, py-1.8.0, pluggy-0.12.0 -- /home/mzalewsk/.virtualenvs/robottelo/bin/python3
cachedir: .pytest_cache
plugins: mock-1.10.0, services-1.3.1
collected 1 item                                                                                                                                                                              

tests/upgrades/test_containers_removal.py::Scenario_containers_support_removal::test_pre_scenario_containers_support_removal PASSED                                                     [100%]

================================================================================== 1 passed in 82.07 seconds ==================================================================================
```
3. Upgrade Satellite to 6.6 snap
4. Run post-upgrade script (I had to remove argument to decorator first, because pytest complained about `'function' object has no attribute '__pytest_node_id'`):
```
$ pytest --pdb -v -m post_upgrade tests/upgrades/test_containers_removal.py::Scenario_containers_support_removal::test_post_scenario_containers_support_removal
platform linux -- Python 3.6.3, pytest-4.0.2, py-1.8.0, pluggy-0.12.0 -- /home/mzalewsk/.virtualenvs/robottelo/bin/python3
cachedir: .pytest_cache
plugins: mock-1.10.0, services-1.3.1
collected 1 item                                                                                                                                                                              

tests/upgrades/test_containers_removal.py::Scenario_containers_support_removal::test_post_scenario_containers_support_removal PASSED                                                    [100%]

================================================================================== 1 passed in 7.85 seconds ===================================================================================
```

### DIFF
--- a/tests/upgrades/test_containers_removal.py
+++ b/tests/upgrades/test_containers_removal.py
@@ -156,7 +156,7 @@ class Scenario_containers_support_removal(APITestCase):
         """Post-upgrade scenario test to verify containers created and run
         before upgrade are still running after upgrade.
 
-        :id: preupgrade-f6de07ae-14c7-4452-9cb1-cafe2aa648ae
+        :id: postupgrade-f6de07ae-14c7-4452-9cb1-cafe2aa648ae
 
         :steps:
 
--- a/tests/upgrades/test_containers_removal.py
+++ b/tests/upgrades/test_containers_removal.py
@@ -1,0 +1,205 @@
+"""Tests verifying that Containers support is properly removed
+
+:Requirement: Upgraded Satellite
+
+:CaseAutomation: Automated
+
+:CaseLevel: Acceptance
+
+:CaseComponent: ContainerManagement
+
+:TestType: Functional
+
+:CaseImportance: Medium
+
+:Upstream: No
+"""
+from fauxfactory import gen_string
+
+from nailgun import entities
+from robottelo import ssh
+from robottelo.test import APITestCase, settings
+from robottelo.vm import VirtualMachine
+from upgrade_tests import post_upgrade, pre_upgrade
+from upgrade_tests.helpers.scenarios import (
+    create_dict,
+    get_entity_data,
+)
+
+
+class Scenario_containers_support_removal(APITestCase):
+    """"Verify containers run before upgrade are still running
+    after upgrade, despite foreman_docker removal.
+
+    Test Steps:
+
+            1. Before Satellite upgrade.
+            2. Create docker host
+            3. Create and run container from dockerhub
+            4. Create and run container from external registry
+            5. Upgrade Satellite
+            6. Verify that upgrade procedure removed foreman_docker support
+            7. Verify container from dockerhub is still running
+            8. Verify container from external registry is still running
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        super(Scenario_containers_support_removal, cls).setUpClass()
+
+    def _vm_cleanup(self, hostname=None):
+        """ Cleanup the VM from provisioning server
+
+        :param str hostname: The content host hostname
+        """
+        if hostname:
+            vm = VirtualMachine(
+                hostname=hostname,
+                target_image=hostname,
+                provisioning_server=settings.clients.provisioning_server,
+                )
+            vm._created = True
+            vm.destroy()
+
+    @pre_upgrade
+    def test_pre_scenario_containers_support_removal(self):
+        """Pre-upgrade scenario test to verify containers created and run
+        before upgrade are still running after upgrade.
+
+        :id: preupgrade-f6de07ae-14c7-4452-9cb1-cafe2aa648ae
+
+        :steps:
+
+            1. Create docker host
+            2. Create and run container from dockerhub
+            3. Create and run container from external registry
+
+        :expectedresults:
+
+            1. Docker host is created
+            2. Container from dockerhub is created and running
+            3. Container from external registry is created and running
+
+        """
+        repo_name = 'rhel'
+        compute_resource_name = gen_string('alpha')
+        registry_url = settings.docker.external_registry_1
+
+        org = entities.Organization().create()
+
+        docker_host = VirtualMachine(
+            source_image=settings.docker.docker_image,
+            tag=u'docker'
+        )
+        docker_host.create()
+        docker_host.install_katello_ca()
+
+        compute_resource = entities.DockerComputeResource(
+            name=compute_resource_name,
+            organization=[org],
+            url='http://{0}:2375'.format(docker_host.ip_addr),
+        ).create()
+
+        # Only one registry with given URL can exist on Satellite,
+        # so search for it first and create it only if necessary
+        try:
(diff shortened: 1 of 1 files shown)


tests/upgrades/test_containers_removal.py
```diff
@@ -0,0 +1,205 @@
...
+            7. Verify container from dockerhub is still running
+            8. Verify container from external registry is still running
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        super(Scenario_containers_support_removal, cls).setUpClass()
+
+    def _vm_cleanup(self, hostname=None):
+        """ Cleanup the VM from provisioning server
```
omkarkhatavkar: I have one question to everyone, as per my knowledge we do not run post upgrade if the pre-upgrade test for the respective class is failed. In this case, the VM resource will be the lingering state. Should we also add validation such that kill the VM resource if those are created by failed pre-upgrade test? 

I'm not sure about our CI job take care of killing all lingering VM resources.

mirzal: I think you are right. Another test that uses `VirtualMachine` directly, `test_remoteexecution.py`, basically wraps entire pre_upgrade test in `try except` block and destroys vm on any exception: https://github.com/SatelliteQE/robottelo/blob/master/tests/upgrades/test_remoteexecution.py#L91

Let me do similar thing here.

mirzal: Done in 75ce0caebdbece643c26e1d3109a362954d4dd71 (commit diff is misleading, as I had to indent most of the code; but it's really 4 new lines)

omkarkhatavkar: NO Problem.

tests/upgrades/test_containers_removal.py
```diff
@@ -92,60 +92,64 @@ def test_pre_scenario_containers_support_removal(self):
...
+                'dockerhub_container': dockerhub_container.name,
+                'external_container': external_container.name,
+            }}
+            create_dict(scenario_dict)
+        except Exception as exp:
+            self._vm_cleanup(hostname=docker_host.hostname)
+            raise Exception(exp)
 
     @post_upgrade(depend_on=test_pre_scenario_containers_support_removal)
     def test_post_scenario_containers_support_removal(self):
```
omkarkhatavkar: @mirzal Do you think same try catch also should be applied to the postupgrade method as well? I can see the earlier example you gave from Repo Don't have.

mirzal: Exactly, test_remoteeecution doesn't have it.

I believe the idea is that if post_upgrade failed, it's probably due to error in environment or test, so we fix it quickly and re-run this part. This way we don't have to run pre-upgrade and costly upgrade process for every issue we encounter. So we want to keep VM around until we explicitly remove it.

tests/upgrades/test_containers_removal.py
```diff
@@ -0,0 +1,209 @@
...
+        except Exception as exp:
+            self._vm_cleanup(hostname=docker_host.hostname)
+            raise Exception(exp)
+
+    @post_upgrade(depend_on=test_pre_scenario_containers_support_removal)
+    def test_post_scenario_containers_support_removal(self):
+        """Post-upgrade scenario test to verify containers created and run
+        before upgrade are still running after upgrade.
+
+        :id: preupgrade-f6de07ae-14c7-4452-9cb1-cafe2aa648ae
```
omkarkhatavkar: should be post-upgrade

tests/upgrades/test_containers_removal.py
omkarkhatavkar: Hmm Yeah, that makes sense.

tests/upgrades/test_containers_removal.py
mirzal: Good catch! Changed in b30046f687c0a0de2e701ec29210706339f411e5

jyejare:
@mirzal , You are bang on with this first upgrade scenario :+1: Keep up the great work !

tests/upgrades/test_containers_removal.py
```diff
@@ -0,0 +1,209 @@
...
+                target_image=hostname,
+                provisioning_server=settings.clients.provisioning_server,
+                )
+            vm._created = True
+            vm.destroy()
+
+    @pre_upgrade
+    def test_pre_scenario_containers_support_removal(self):
+        """Pre-upgrade scenario test to verify containers created and run
+        before upgrade are still running after upgrade.
```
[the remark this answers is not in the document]
mirzal: Done in b0e6c4b50b0f53bf7506a1227122ecd4a942a3f7 .

tests/upgrades/test_containers_removal.py
```diff
@@ -0,0 +1,209 @@
...
+                            for line in docker_log))
+
+        removed_tables = ('docker_images', 'docker_tags',
+                          'docker_container_wizard_states_images',
+                          'containers')
+
+        for table in removed_tables:
+            self.assertTrue(any("table_exists?(:{})".format(table)
+                                in line for line in docker_log))
+
```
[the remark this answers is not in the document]
mirzal: This one is tricky. Removal of support for docker operations in Robottelo (CLI), Nailgun and Airgun is on my list of things to do. Once these commits are merged, our tooling will be unable to call these endpoints and verify that they are indeed removed on Satellite side.

My idea was to create separate test cases that would ensure endpoints are removed. This will be easy for CLI and doable for API; not yet sure how I can approach web UI. We can decorate these test cases with `@upgrade` to ensure they are run after upgrade, too.

Another approach would be to leave support for docker operations in our tooling for now, and remove it sometime after 6.6 GA. This way I can call API endpoint and expect exception.

What do you think?

tests/upgrades/test_containers_removal.py
mirzal: Fair point. Added in b0e6c4b50b0f53bf7506a1227122ecd4a942a3f7

mirzal:
> 1. Cherrypick the Scenario to 6.5.z

Why?

Looking at [satellite6-upgrade-run-scenarios.sh](https://github.com/SatelliteQE/robottelo-ci/blob/master/scripts/satellite6-upgrade-run-scenarios.sh), it ensures that we have nailgun from correct branch, but I don't see it changing Robottelo itself. Am I missing something?

> 2. Remove this entire Scenario for next version that is 6.7  - else - use dont_run_to_upgrade decorator from satellite6_upgrade repo.

Yes, I will remove it after 6.6 GA.

ntkathole:
@jyejare I don't think we need cherry-pick as we are cloning master for 6.6 upgrade.

PR #7056, opened 2019-06-24
Move remote from webdrivers, it should be a provider like sauce,docker

eko999:
Remote should not be a webdriver, it should be a provider and can be used for different remote webderivers. Webdriver and command_executor are required for Remote provider, everybody can use the provider to access any remote webdriver, the only thing is configure a valid remote webderiver url as below:

[robottelo]
browser=remote
webdriver=firefox 
command_executor=http://10.73.xxxx.xxx:4444/wd/hub


eko999:
This change depends on the airgun PR: https://github.com/SatelliteQE/airgun/pull/344

PR #7060, opened 2019-06-25
Fix Bootstrap casecomponent

swadeley:
Part of metadata review. As its a stubbed test I just fixed case component and added the full name of the script in case someone greps for that.

Thank you


swadeley:
@mirzal sorry about that. those were some artefacts from using meld. My fault for using "git ." without a proper check. Fixed and forced pushed.

Thank you

swadeley:
Thank you all.

PR #7063, opened 2019-06-25
Fix CaseComponent Email

swadeley:
Part of metadata review. As its a stubbed test I just fixed case component.

Thank you


swadeley:
Thank you all.

PR #7072, opened 2019-06-26
removing docker compute resource ui tests

omkarkhatavkar:
### PR Description 
As we are no longer supporting Docker Compute Resource, removing the UI Tests.   

### Note
Please correct if I'm wrong.


mirzal:
Is this removal of entire file?

When I discussed this topic with @san7ket , he gave me idea of having some kind of negative test that would check that these options are indeed removed. I would like to keep that option open, at least for a little longer. Removing file will make it slightly harder to implement it.

Can we leave the file, probably only with metadata and short comment why it is empty? I will make sure to add some test case, or remove file before 6.6 GA.

omkarkhatavkar:
> dd some test case, or remove file before 6.6 GA.

Ok Mirek I will close this PR in that case ?

mirzal:
@omkarkhatavkar that's an option, too. These tests failures are on me, we can ignore them for now.

Issue #6956, opened 2019-05-24
Add upgrade test for bulk Action update all packages(via katello agent)


ntkathole:
Steps:
1. Select 1 or N hosts on the content host page
2. Click on "Select Action" button.
3. Click on "Manage Packages" button.
4. Click on "Update All Packages"

For more details :
https://bugzilla.redhat.com/show_bug.cgi?id=1712069

swadeley:
Hello @ntkathole 

I could not find code to access the menu with the "Update All Packages" button from Content Hosts view. I found there was existing code[1] to access that menu from Host Collections view.

I am asking jturel in bugzilla [2] if testing that Update All Packages function via Host Collections would test the code path that caused this issue. 

[1] manage_packages https://github.com/SatelliteQE/airgun/blob/4ae23257a7f5ad3a722dd0ac521de6d292d15961/airgun/entities/hostcollection.py#L68
[2] https://bugzilla.redhat.com/show_bug.cgi?id=1712069#c9

swadeley:
Still need to cherry-pick to 6.5 branch

swadeley:
Hello, the test is now merged into the 6.5.z branch.

PR #7074, opened 2019-06-26
test_positive_end_to_end_bulk_update (#6956)

swadeley:
* Test the Update All Packages button on Update Packages dialogue
 * Use the Host Collection view to get to it.

### DIFF
--- a/tests/foreman/ui_airgun/test_contenthost.py
+++ b/tests/foreman/ui_airgun/test_contenthost.py
@@ -14,6 +14,7 @@
 
 :Upstream: No
 """
+from datetime import datetime
 import pytest
 
 from airgun.session import Session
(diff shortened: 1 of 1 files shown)


swadeley:
This is already in master branch.

 This change depends on https://github.com/SatelliteQE/airgun/pull/345 (now merged)

swadeley:
@mirzal sory about that, it was due to my misunderstanding. When I was told 6.5.z branch still uses /ui_airgun directory I applied the change to the file in that directory and then removed the file in /ui when I should have just done a checkout of the file to revert my changes in that /ui directory.

I have now done that and forced pushed.

Thank you

mirzal:
@swadeley no need to apologize. I'm pretty sure we could remove that file and nothing would happen, as it isn't used in 6.5.z branch anyway. It's just a little unexpected thing to do, that's all.

Anyway, we're all set now. I mean, once we have Tier2 reviewer ACK.

swadeley:
Thank you all

PR #7078, opened 2019-06-27
Metadata cleanup - Users and Roles

rdrazny:
Metadata updates for the Users&Roles component

### DIFF
--- a/tests/foreman/api/test_role.py
+++ b/tests/foreman/api/test_role.py
@@ -682,7 +682,7 @@ class CannedRoleTestCases(APITestCase):
                 None in cloned role
             2. Override flag is set to True in cloned role filter
 
-                :CaseLevel: Integration
+        :CaseLevel: Integration
         """
         role = entities.Role(
             name=gen_string('alpha'),
--- a/tests/foreman/api/test_role.py
+++ b/tests/foreman/api/test_role.py
@@ -10,7 +10,7 @@ http://theforeman.org/api/apidoc/v2/roles.html
 
 :CaseLevel: Acceptance
 
-:CaseComponent: API
+:CaseComponent: UsersRoles
 
 :TestType: Functional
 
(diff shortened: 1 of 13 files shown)


tests/foreman/cli/test_role.py
```diff
@@ -458,6 +458,8 @@ def test_negative_update_role_taxonomies(self):
 
         :expectedresults: The taxonomies of overridden filter should not be
             updated with role taxonomies
+
+        :CaseAutomation: notautomated
```
mirzal: Since all cases in `CannedRoleTestCases` class are stubbed, you can as well set `:CaseAutomation: notautomated` on class level.

Of course setting it for each test individually is working, too.

tests/foreman/api/test_role.py
```diff
@@ -682,9 +682,7 @@ def test_positive_emptiness_of_filter_taxonomies_on_role_clone(self):
                 None in cloned role
             2. Override flag is set to True in cloned role filter
 
-        :CaseAutomation: notautomated
-
-        :CaseLevel: Integration
+                :CaseLevel: Integration
```
mirzal: It doesn't matter for testimony, but I wonder how and why this get indented.

tests/foreman/cli/test_user.py
```diff
@@ -1150,6 +1150,8 @@ def test_positive_create_ssh_key(self):
         :expectedresults: SSH key should be added to new user
 
         :CaseImportance: Critical
+
+        :CaseAutomation: notautomated
```
mirzal: Same here. As far as I can see, all tests in `SshKeyInUserTestCase` are stubbed, so you can add it on class level.

tests/foreman/cli/test_role.py
rdrazny: You are right, of course - I have completey forgot that beside module-level and test-level it is possible to set class-level metadata as well. Fixed.

tests/foreman/api/test_role.py
rdrazny: That's easy - because I didn't notice the string has moved when I was deleting staled "notautomated" metadata :-). Fixed.

mirzal:
@jyejare @ntkathole @rplevka can you ACK and merge this, if you are OK with these changes?

PR #7082, opened 2019-06-28
Update Capsule tests metadata

lpramuk:


lpramuk:
@mirzal I'm done. Merge or another ACK?

mirzal:
@lpramuk mine ACK is in place, according to our processes we need Tier2 reviewer ACK to merge. I requested review from specific people.

tests/foreman/api/test_smartproxy.py
```diff
@@ -6,11 +6,11 @@
 
 :CaseLevel: Acceptance
```
ntkathole: @lpramuk It would be great if we change Caselevel as component here. I don't see any instance of acceptance caselevel in this file.

tests/foreman/cli/test_capsule.py
```diff
@@ -7,11 +7,11 @@
 
 :CaseLevel: Acceptance
```
ntkathole: Same for this file as well

tests/foreman/api/test_smartproxy.py
lpramuk: or yet better remove ```:CaseLevel:``` from module level and have it only defined for every single test

lpramuk:
Resolved @ntkathole  comment

tests/foreman/api/test_smartproxy.py
lpramuk: Ok. Updated ```:CaseLevel:```

PR #7094, opened 2019-07-01
Fix containers removal upgrade tests

mirzal:
The problem is that due to changes in registry.access.redhat.com, Satellite is no longer able to pull in content from that registry. See [bz#1703397](https://bugzilla.redhat.com/show_bug.cgi?id=1703397), especially comment # 5. There are couple of possible solutions, each with their own problems:

## 1. Use registry.redhat.io

registry.redhat.io is new Red Hat registry that will eventually deprecate registry.access.redhat.com. The thing is, it requires authentication to pull content in. Portal credentials can be used, and while we do have QE account there, it is not public and out tests never use it. Robottelo **could** use it, if robottelo-ci is modified to get secret credentials from Jenkins and modify certain placeholders in robottelo.properties before running test. Which sounds like a lot of work for test that will have to be removed after 6.6 GA.

## 2. Apply hotfix manually

Fix for the issue is quite small - [foreman_docker PR#218](https://github.com/theforeman/foreman_docker/pull/218/files). It only requires running `foreman-maintain service restart` after applying patch. Robottelo has all the infrastructure needed to pull this off, but:
1. It would modify running Satellite, possibly impacting other tests or operations going on at the same time
2. It's The Wrong Way™ to do it. It works, but it's not something we would be proud of.

## 3. Check Bugzilla state and conditionally skip portions of test

Robottelo can skip certain portions of tests if relevant bugzilla is open. This is easy to implement and certain to "fix" the issue, but at the expense of coverage - we basically lose half of it. On the other hand, I covered that during test development and it's not like we expect this particular scenario to suddenly start failing. 

Of course the best option would be to release 6.5.z with proper fix before 6.6 GA, but this is outside of our control.

-----------------------------

This PR goes with solution # 3, but at the same time serves as platform for discussion. Possibly other approach would be better used, or there is some other solution that I missed.

```
$ pytest --pdb -v -m pre_upgrade tests/upgrades/test_containers_removal.py::Scenario_containers_support_removal::test_pre_scenario_containers_support_removal
platform linux -- Python 3.6.3, pytest-4.0.2, py-1.8.0, pluggy-0.12.0 -- /home/mzalewsk/.virtualenvs/robottelo/bin/python3
cachedir: .pytest_cache
plugins: mock-1.10.0, services-1.3.1
collected 1 item                                                                                                                                                                              

tests/upgrades/test_containers_removal.py::Scenario_containers_support_removal::test_pre_scenario_containers_support_removal PASSED                                                     [100%]

============================================================================ 1 passed, 2 warnings in 65.24 seconds ============================================================================
```
Warnings are skipped from output, they are related to SafeConfigParser being deprecated.

### DIFF
--- a/tests/upgrades/test_containers_removal.py
+++ b/tests/upgrades/test_containers_removal.py
@@ -18,6 +18,7 @@ from fauxfactory import gen_string
 
 from nailgun import entities
 from robottelo import ssh
+from robottelo.decorators import bz_bug_is_open
 from robottelo.test import APITestCase, settings
 from robottelo.vm import VirtualMachine
 from upgrade_tests import post_upgrade, pre_upgrade
@@ -81,6 +82,7 @@ class Scenario_containers_support_removal(APITestCase):
             3. Container from external registry is created and running
 
         """
+        rh_registry_available = not bz_bug_is_open(1703397)
         repo_name = 'rhel'
         compute_resource_name = gen_string('alpha')
         registry_url = settings.docker.external_registry_1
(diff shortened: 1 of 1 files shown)


san7ket:
@jyejare @ntkathole @vijay8451 ping

tests/upgrades/test_containers_removal.py
```diff
@@ -81,6 +82,7 @@ def test_pre_scenario_containers_support_removal(self):
             3. Container from external registry is created and running
 
         """
+        rh_registry_available = not bz_bug_is_open(1703397)
```
jyejare: :+1:  Good way of doing things :)

tests/upgrades/test_containers_removal.py
```diff
@@ -121,30 +123,35 @@ def test_pre_scenario_containers_support_removal(self):
...
+                                    for line in running_containers.stdout))
+
+            ext_container_name = external_container.name if rh_registry_available else ''
 
             scenario_dict = {self.__class__.__name__: {
+                'rh_registry_available': rh_registry_available,
                 'docker_host': docker_host.hostname,
                 'dockerhub_container': dockerhub_container.name,
-                'external_container': external_container.name,
+                'external_container': ext_container_name,
```
jyejare: Once the bug is verified, this will be '', Are you ok for that ?

mirzal: This bug should be fixed in z-stream, so it will probably only take a week or two for it to move from VERIFIED to CLOSED. If we have some info it will take longer than that, I am very much OK with another PR that would revert this patch.

But my guess/worry is that we will release 6.6 before this bug is CLOSED. And this test is going to be removed after 6.6 GA anyway.

mirzal:
@jyejare Does my answer satisfy you, or do you want to this it further?
@san7ket my last commit dismissed your review, can you re-ACK?

PR #7099, opened 2019-07-02
Testimony fixes: tiers

lhellebr:


tests/foreman/api/test_computeprofile.py
```diff
@@ -34,6 +34,8 @@ def test_positive_create_with_name(self):
         :expectedresults: Compute Profile is created
 
         :CaseImportance: Critical
+
+        :CaseLevel: Component
```
vijay8451: Can you please define the `CaseLevel` at module level so that you no need to set this for every test where it works ?

tests/foreman/api/test_computeprofile.py
```diff
@@ -49,6 +51,8 @@ def test_negative_create(self):
         :expectedresults: Compute Profile is not created
 
         :CaseImportance: Critical
+
+        :CaseLevel: Component
```
vijay8451: same as above

tests/foreman/api/test_computeresource_libvirt.py
```diff
@@ -97,7 +101,9 @@ def test_positive_create_libvirt_with_display_type(self):
         :expectedresults: Compute resources are created with expected
             display_type value
 
-        :CaseImportance: Critical
+        :CaseImportance: High
+
+        :CaseLevel: Component
```
vijay8451: same as above for `CaseLevel` and `CaseImportance`

tests/foreman/api/test_computeresource_libvirt.py
```diff
@@ -119,6 +125,8 @@ def test_positive_create_with_provider(self):
         :expectedresults: Compute resources are created with expected providers
 
         :CaseImportance: Critical
+
+        :CaseLevel: Component
```
vijay8451: same as above and below tests where applicable for `CaseLevel` and `CaseImportance`

tests/foreman/api/test_computeresource_libvirt.py
```diff
@@ -76,6 +78,8 @@ def test_positive_create_with_description(self):
             descriptions
 
         :CaseImportance: Critical
+
+        :CaseLevel: Component
```
vijay8451: same as above

tests/foreman/api/test_computeresource_libvirt.py
```diff
@@ -55,6 +55,8 @@ def test_positive_create_with_name(self):
         :expectedresults: Compute resources are created with expected names
 
         :CaseImportance: Critical
+
+        :CaseLevel: Component
```
vijay8451: same as above

tests/foreman/api/test_computeprofile.py
```diff
@@ -100,6 +108,8 @@ def test_positive_delete(self):
         :expectedresults: Compute Profile is deleted successfully.
 
         :CaseImportance: Critical
+
+        :CaseLevel: Component
```
vijay8451: same as above

tests/foreman/api/test_computeprofile.py
```diff
@@ -81,6 +87,8 @@ def test_negative_update_name(self):
         :expectedresults: Compute Profile is not updated.
 
         :CaseImportance: Critical
+
```
vijay8451: same as above

tests/foreman/api/test_computeprofile.py
```diff
@@ -64,6 +68,8 @@ def test_positive_update_name(self):
         :expectedresults: Compute Profile is updated.
 
         :CaseImportance: Critical
+
+        :CaseLevel: Component
```
vijay8451: same as above

tests/foreman/api/test_computeprofile.py
mirzal: @vijay8451 It's perfectly fine to define the same metadata for each case individually, too. Hope you don't consider this comment blocking, because I think it is not. It's mostly personal preference thing.

vijay8451: @mirzal you are right ... it is a suggestion only not blocker ...
However when I checked the test file I see that at module level it is still requires to change too: https://github.com/SatelliteQE/robottelo/blob/bf73f96589b80a30375f5dd41b680eca5c084bc5/tests/foreman/api/test_computeprofile.py#L10

Hence above may help to achieve both part of it.

lhellebr:
CaseComponent fixed

mirzal:
Adding DO NOT MERGE due to unclear relationship with #7148 . Please see https://github.com/SatelliteQE/robottelo/pull/7148#issuecomment-510597567

lhellebr:
Those were just two different sets of changes, but for the same reason - I merged both into this PR. Please review this one and I'll close another one.

Issue #7006, opened 2019-06-06
[Discussion] Virt-who UI config test and approach to automate


sghai:
All,

I would like to start a discussion on automating virt-who-config UI tests. Earlier we were thinking to automate as part of virt-who repo but then thought if we should automate as part of robottelo. I see value having the automation as part of robottelo but then I wanted to understand if there are concerns keeping then in robottelo. Few considerations, in case we are going to keep tests in robottelo:

a) We will run tests with virt-who jenkins as all the hypervisors are part of China lab
b) Plan to define a config  file only for virt-who hypevisor configuration.

Now question is how we should read the virt-who config file ? Do we have some common read functions in robottelo ?

Please expect some questions here are there and lets be open to come up with conclusion.

sghai:
@san7ket @jyejare @ntkathole: ^^

ntkathole:
> a) We will run tests with virt-who jenkins as all the hypervisors are part of China lab

This would be good. They just need to setup jenkins job that will run tests from robottelo in their Jenkins.

> b) Plan to define a config file only for virt-who hypevisor configuration.
> 

The main change, we will require to have same base config file as robottelo https://github.com/SatelliteQE/robottelo/blob/master/robottelo/config/base.py#L25 . 
We can create new directory at https://github.com/SatelliteQE/robottelo/tree/master/robottelo with name `virtwho_config `(similar to config directory used by robottelo) and add functions to read properties file.

For virt_who tests, we can have separate directory same as we have for upgrades under robottelo/tree/master/tests/foreman/

This way we can keep tests as well as properties file of virtwho team separated and it won't affect our automation. 

Adding just high level idea, need to lookup in depth for overall plan. Lets see what others think.

jyejare:
I see only following values keeping these tests in Robottelo is:
1. If we are planning to run in upstream(if they are not running already)
2. If we want all the tests in a single place regardless of component. (Just like we moved all upgrade scenario tests in robottelo)

And,

> a) We will run tests with virt-who jenkins as all the hypervisors are part of China lab
Yes, but if we really want to run those tests on our environement(may be in future), we may need PRs in robottelo-ci wrt virt-who.
> b) Plan to define a config file only for virt-who hypevisor configuration.
I think keeping two config files isnt good for any test-framework So either we can add one-more config section for virt-who. Which I believe would be a challenge as Virt-who might have long list of properties.
[...]

PR #7102, opened 2019-07-03
Support to run virtwho-configure UI cases

eko999:
This PR is used to support virt-who configure UI cases and the first test case is included.
1. add a new section `[virtwho]` in robottelo.properties.sample
2. add a new class `VirtWhoSettings` in robottelo/config/base.py to read the config
3. add a new file `robottelo/virtwho.py`, all the virtwho functions will be defined there.
4. test case: tests/foreman/ui/test_virtwho_configure.py 

Note: Now there is another functions file existing: `robottelo/virt_who_configure.py`, it still be used by `tests/foreman/api/test_virt_who_config.py`, we will re-write the api/cli cases in the future, but now, continue to keep them.


eko999:
Test result as:
```
# pytest tests/foreman/ui/test_virtwho_configure.py 
2019-07-03 14:12:30 - conftest - DEBUG - Registering custom pytest_configure

platform linux -- Python 3.5.3, pytest-4.0.2, py-1.8.0, pluggy-0.12.0
rootdir: /root/eko/robottelo, inifile:
plugins: mock-1.10.4, services-1.3.1, xdist-1.29.0, cov-2.7.1, forked-1.0.2
collecting ... 2019-07-03 14:12:30 - conftest - DEBUG - BZ deselect is disabled in settings

collected 1 item                                                                                                                                                            

tests/foreman/ui/test_virtwho_configure.py 
.                                                                                                                          [100%]

============================================================================= warnings summary ==============================================================================
tests/foreman/ui/test_virtwho_configure.py::test_positive_end_to_end
  /usr/lib/python3.5/site-packages/widgetastic/widget/base.py:827: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() instead
    c_args, c_varargs, c_keywords, c_defaults = inspect.getargspec(condition)

=================================================================== 1 passed, 1 warnings in 92.58 seconds ===================================================================
```

sghai:
Good to see the coverage around Virt-who UI tests as part of robottelo.
@jyejare @mirzal @ntkathole @omkarkhatavkar Please review..

tests/foreman/ui/test_virtwho_configure.py
```diff
@@ -0,0 +1,44 @@
...
+
+:Upstream: No
+"""
+from fauxfactory import gen_string
+
+from robottelo.decorators import tier2
+from robottelo.virtwho import get_form_data
+
+
+@tier2
```
ntkathole: @sghai are we planning to run this tests in our jenkins? I am asking because if we add this test here, it will always fail for us.

tests/foreman/ui/test_virtwho_configure.py
```diff
@@ -0,0 +1,44 @@
...
+
+    :expectedresults: All expected CRUD actions finished successfully
+
+    :CaseLevel: Acceptance
+
+    :CaseImportance: High
+    """
+    name = gen_string('alpha')
+    with session:
+        form_data = get_form_data(name)
```
omkarkhatavkar: you can move this form_data = get_form_data(name) line above session.

eko999: Yes, it should be, thank you for suggestion.

tests/foreman/ui/test_virtwho_configure.py
eko999: A existing hypervisor environment is required to run the virtwho cases, virtwho CI system will provision the environment and run the cases, but any suggestion to store the case in robottelo?

ntkathole: @eko999 then as I said https://github.com/SatelliteQE/robottelo/issues/7006#issuecomment-499467701 We need to have separate directory under robottelo/tree/master/tests/foreman/ for virt_who tests

robottelo/virtwho.py
```diff
@@ -0,0 +1,44 @@
+"""Utilities for Virtwho Configure Plugin Testing
+
+All the virtwho-configure UI/CLI/API interfaces should be defined there.
+
+"""
+from robottelo.config import settings
+
+
+def get_hypervisor_type():
```
mirzal: Shouldn't this function take constant name as argument and return specific string? Something like:
```python
def get_hypervisor_type(hypervisor):
    mapping = {
        'esx': 'VMware vSphere / vCenter (esx)',
        # skip for brevity
    }
    return mapping[hypervisor]
```

And then you will use it like that:
```python
hypervisor_type = get_hypervisor_type(hypervisor)
```

Also, I'm not sure about the name here. It doesn't really return hypervisor type, which is already decided in robottelo.properties. We obtain user-visible name based on type.

tests/foreman/ui/test_virtwho_configure.py
eko999: okay, looking forward to the final decision from @sghai :+1:

robottelo.properties.sample
```diff
@@ -518,3 +518,26 @@ port_range=9091, 14999
 # redis_password=
 # How much time we retry if a function call fail, by default call_retries=2
 # call_retries=2
+
+# Section for virtwho configure function
+# [virtwho]
+# The hypervisor type, one of the values: esx,xen,hyperv,rhevm,libvirt,kubevirt
+# hypervisor_type=
```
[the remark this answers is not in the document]
eko999: Yes, it's true, I will add validation for hypervisor_type

robottelo.properties.sample
```diff
@@ -518,3 +518,26 @@ port_range=9091, 14999
...
+# The hostname or ip or other values for hypervisor server, depends on the type
+# hypervisor_server=
+# Admin username to access hypervisor server
+# hypervisor_username=
+# Admin password to access hypervisor server
+# hypervisor_password=
+# The config file to connect hypervisor, only for kubevirt
+# hypervisor_config_file=
+# The guest ip or hostname in the hypervisor
+# guest=
```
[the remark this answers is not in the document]
eko999: Different hypervisor has different guest ip and it's created by dhcp when provisioning the env.

robottelo/config/base.py
```diff
@@ -1075,6 +1075,54 @@ def validate(self):
...
+
+    def validate(self):
+        """Validate virtwho settings."""
+        validation_errors = []
+        values = [v for k, v in vars(self).items() if k != 'hypervisor_config_file']
+        if not all(values):
+            validation_errors.append(
+                'All [virtwho] hypervisor_type, hypervisor_server, hypervisor_username, '
+                'hypervisor_password, guest, guest_username, guest_password, '
+                'sku_vdc_physical, sku_vdc_virtual options must be provided.'
```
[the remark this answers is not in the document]
eko999: `guest, guest_username, guest_password, sku_vdc_physical, sku_vdc_virtual` must be required, we have to run the following test steps:
1. register the guest to satellite, check the association is created or not between hypervisor and guest.
2. attach the sku_vdc_physical to hypervisor, and then check the sku_vdc_virtual is generated or not in the guest.

robottelo/virtwho.py
```diff
@@ -0,0 +1,44 @@
...
+        'xen': 'XenServer (xen)',
+        'rhevm': 'Red Hat Virtualization Hypervisor (rhevm)',
+        'hyperv': 'Microsoft Hyper-V (hyperv)',
+        'libvirt': 'libvirt',
+        'kubevirt': 'Container-native virtualization',
+    }
+
+
+def get_form_data(
+        name, debug=True, interval='Every hour', hypervisor_id='hostname'):
```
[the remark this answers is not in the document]
eko999: My original idea was define all the functions in the virtwho.py, because we have some functions need to be used by ui/api/cli, such as:
 virtwho_logfile_analyzer()
 deploy_by_hammer()
 deploy_by_script()
 register_guest()

An example to understand, the UI have to involve some CLI function:
1. create a esx config by UI
2. get the deploy command line from UI
3. run the command line in virt-who host by CLI
4. check the virtwho log message by CLI
5. check the hypervisor/guest sent to satellite or not by UI
6. attach a physical sku to hypervisor by UI
7. check the virtual sku is generated or not and then attach it to the guest by UI

robottelo/virtwho.py
eko999: The background is:
The real virt-who config file is defined in /etc/virt-who.d/xxx.conf such as:
[virtwho-config-1]
type=esx
server=xxx.xxx.xxx.xxx
username=xxx
password=xxx
owner=xxx

Virtwho Configure UI change the "esx" type to "VMware vSphere / vCenter (esx)" for user-visible, that means when we select "VMware vSphere / vCenter (esx)" in UI, we have to verify the type is "esx" or not in /etc/virt-who.d/xxx.conf.

tests/foreman/ui/test_virtwho_configure.py
jyejare: @eko999 , I think it is more how/which we run the tests on Jenkins CI from robottelo.

We run tests from specific folders under tests/foreman/ - API/CLI/UI and hence probably @ntkathole is asking u to keep the tests in a separate directory.

robottelo/virtwho.py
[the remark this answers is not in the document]
eko999: I have to pass name parameter to `get_form_data` to create the UI form, and also need some conditional statements for different hypervisors' form data.  I have no good idea how to pass name parameter to `get_form_data` if I change it to a fixture.

robottelo/virtwho.py
```diff
@@ -0,0 +1,44 @@
...
+
+All the virtwho-configure UI/CLI/API interfaces should be defined there.
+
+"""
+from robottelo.config import settings
+
+
+def get_hypervisor_type():
+    return{
+        'esx': 'VMware vSphere / vCenter (esx)',
```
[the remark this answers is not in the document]
eko999: The background is:
When creating a config in the UI and deploy it by CLI, the real virt-who config file will be created in /etc/virt-who.d/xxx.conf such as:
[virtwho-config-1]
type=esx
server=xxx.xxx.xxx.xxx
username=xxx
password=xxx
owner=xxx

Virtwho Configure UI change the "esx" type to "VMware vSphere / vCenter (esx)" for user-visible, that means when we select "VMware vSphere / vCenter (esx)" in UI, we have to verify the type is "esx" or not in /etc/virt-who.d/xxx.conf.

tests/foreman/ui/test_virtwho_configure.py
eko999: Sure, I will move the test file to tests/foreman/virtwho/, thank you all.

robottelo/virtwho.py
eko999: Another choice is we change the type in robottelo.properties as:
# The hypervisor type, should be one of the values: 
#  'VMware vSphere / vCenter (esx)',
#  'XenServer (xen)',
#  'Red Hat Virtualization Hypervisor (rhevm)',
#  'Microsoft Hyper-V (hyperv)',
#  'libvirt',
#  'Container-native virtualization'
# hypervisor_type=VMware vSphere / vCenter (esx)

robottelo/virtwho.py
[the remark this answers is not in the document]
mirzal: >  I have no good idea how to pass name parameter to `get_form_data` if I change it to a fixture.

There are at least three ways. I would consider them more or less the same, so just pick whichever you find easiest to implement and reason about.

1. [Use indirect parametrization](https://stackoverflow.com/a/33879151/3552063)
2. [Create fixture that will return a function](https://stackoverflow.com/a/28570677/3552063) and call that function with parameters in test
3. Let fixture return part of the dict that is constant and update this dict in test. This will be fine here, as you are just taking parameters and putting them in dict, they are not involved in any decision-making.

robottelo/virtwho.py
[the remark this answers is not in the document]
mirzal: I think I wasn't clear enough, so let me try again.

In robottelo.properties and robottelo tests you would use short names (`esx`, `xen` etc.). In your test, you would do:
```python
session.virtwho_configure.create({'hypervisor': 'esx', ...})
```
Then, in Airgun, you translate short name to UI name before calling `.fill()`:
```python
class VirtwhoConfigureEntity(BaseEntity):

    def create(self, values):
        """Create new virtwho configure entity"""
        view = self.navigate_to(self, 'New')
        mapping = {'esx': 'VMware vSphere / vCenter (esx)', 'xen': 'XenServer (xen)', # skip others} 
        vals = values.copy()
        vals.update({'hypervisor_type': mapping[vals['hypervisor']]})
        vals.pop('hypervisor')
        # vals now has correct hypervisor_type and doesn't have hypervisor:
        # {'name': 'new name', 'hypervisor_type': 'VMware vSphere / vCenter (esx)'}
        view.fill(vals)
        view.submit.click()
        view.flash.assert_no_error()
        view.flash.dismiss()
```

Of course you should have similar code for `edit`, so `mapping` should be class variable shared between these two functions. Or create private method in `VirtwhoConfigureEntity` that will be responsible for translating between short name and UI string.

Does that make sense to you? Do you like this idea? If not, we can leave it as is - as I said, some people disagree with that approach and this is only a suggestion.

robottelo/virtwho.py
mirzal: Oh no, putting UI strings in robottelo.properties doesn't sound like a good design for me. Short names, like `esx`, `xen` etc. are OK.

And if I understand correctly, CLI tests might need to use short names, so you have to do translation between short name and UI string anyway. I think it's better to map from short name to UI string than the other way around.

robottelo/virtwho.py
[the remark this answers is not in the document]
mirzal: I forgot to add this link.
We use similar idea in Airgun already here: https://github.com/SatelliteQE/airgun/blob/master/airgun/entities/contenthost.py#L120
and then this is how it's called in Robottelo: https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/ui/test_contenthost.py#L831

robottelo/virtwho.py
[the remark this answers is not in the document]
eko999: Make sense to me, drop the utils.py now, and update the form_data in test

robottelo/virtwho.py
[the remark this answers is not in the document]
eko999: The key idea is: don't need to update the hypervisor_type in robottelo if the type element is changed in UI, only need to update it in airgun. A new implement as below in entities/virtwho_configure.py:
```
    def reset_values(self, values):
        mapping = {
            'esx': 'VMware vSphere / vCenter (esx)',
            'xen': 'XenServer (xen)',
            'hyperv': 'Microsoft Hyper-V (hyperv)',
            'rhevm': 'Red Hat Virtualization Hypervisor (rhevm)',
            'libvirt': 'libvirt',
            'kubevirt': 'Container-native virtualization',
        }
        if 'hypervisor_type' in values:
            hypervisor_value = values['hypervisor_type']
            values['hypervisor_type'] = mapping[hypervisor_value]
        return values

    def create(self, values):
        """Create new virtwho configure entity"""
        view = self.navigate_to(self, 'New')
        values = self.reset_values(values)
        view.fill(values)
        view.submit.click()
        view.flash.assert_no_error()
        view.flash.dismiss()
```

robottelo/virtwho.py
eko999: Changed it back and comment as above.

tests/foreman/virtwho/conftest.py
```diff
@@ -0,0 +1,105 @@
+import logging
```
mirzal: I see this file is copied from `tests/foreman/ui/conftest.py`, with some parts removed. I wonder if there is a way for us to share these fixtures between two directories.
If nobody has any good idea, we can deal with that duplication later.

robottelo/virtwho.py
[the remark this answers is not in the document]
mirzal: Thanks, Eko. This is definitely moving in right direction.
But this specific implementation has one problem. I explained in https://github.com/SatelliteQE/airgun/pull/355#discussion_r300464632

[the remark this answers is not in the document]
eko999: Yes, got it,  let's move to https://github.com/SatelliteQE/airgun/pull/355 and fix it.

tests/foreman/virtwho/conftest.py
eko999: We can use `sys.path.append('../')` to import the conftest.py from ui, but I think it's not a good idea.
about the @stubbed, yes, I have add some doc string cases in virthwo, these cases should be automated in the future, it will not finished in this PR.

eko999:
> Overall looks good. Thanks @eko999 for putting great efforts. Few suggestions (It's up to you if you want to do it in same PR or another one, I don't want to block this PR on them)
> 
>     * [ ]   You can add this virtwho directory for `uuid-check` https://github.com/SatelliteQE/robottelo/blob/master/Makefile#L69
> 
>     * [ ]   For documentation of modules, you can add `robottelo.virtwho.rst` https://github.com/SatelliteQE/robottelo/tree/master/docs/api
> 
> 
> `Note for merging : Please use squash and merge while merging.`

Hey Nikhil, thanks for your suggestion, I will do this in another PR, because this PR is very huge now.

ntkathole:
@mirzal Feel free to merge if you are good with it.

PR #7103, opened 2019-07-03
Fixing Satellite subscription name we have in the manifest now

jhutar:
```
pytest tests/foreman/cli/test_host.py::HostSubscriptionTestCase::test_positive_remove --no-print-logs --capture=no --exitfirst 2>&1 | tee /tmp/pytest.log
...
=================== 1 passed, 5 warnings in 1641.36 seconds ====================
```

### DIFF
--- a/robottelo/constants.py
+++ b/robottelo/constants.py
@@ -452,7 +452,7 @@ LANGUAGES = {
     u'zh_TW': u'zh_TW'
 }
 
-SATELLITE_SUBSCRIPTION_NAME = 'Red Hat Satellite Employee Subscription'
+SATELLITE_SUBSCRIPTION_NAME = 'Red Hat Satellite Infrastructure Subscription'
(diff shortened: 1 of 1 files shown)


ntkathole:
ok to test : tests/test_advanced.py -k "test_positive_beta_repositories, test_positive_repositories_setup"

jhutar:
This also needs to be backported to 6.4 and 6.5 branch. Shall I just create another PRs for that?

vijay8451:
> This also needs to be backported to 6.4 and 6.5 branch. Shall I just create another PRs for that?

Cherry-pick would works fine for 6.4 and 6.5.

mirzal:
> This also needs to be backported to 6.4 and 6.5 branch. Shall I just create another PRs for that?

Yes, after this is merged, cherry-pick commit in these branches and send new PRs. Best to add reference to this PR as well, so we know this is only cherry-pick and not a new thing.

PR #7106, opened 2019-07-03
[6.4.z] move old ui tests to ui_deprecated and ui_airgun to ui

rplevka:
I basically did:

```bash
mv ui ui_deprecated && mv ui_airgun ui
```
the confusing diff is probably because the `ui` dir still exists but contains different files, where some files have the same name so it takes it as changes made to those files


Makefile
```diff
@@ -5,9 +5,9 @@ FOREMAN_CLI_TESTS_PATH=$(join $(FOREMAN_TESTS_PATH), cli)
 FOREMAN_RHAI_TESTS_PATH=$(join $(FOREMAN_TESTS_PATH), rhai)
 FOREMAN_RHCI_TESTS_PATH=$(join $(FOREMAN_TESTS_PATH), rhci)
 FOREMAN_ENDTOEND_TESTS_PATH=$(join $(FOREMAN_TESTS_PATH), endtoend)
-FOREMAN_TIERS_TESTS_PATH=$(join $(FOREMAN_TESTS_PATH), {api,cli,ui_airgun})
+FOREMAN_TIERS_TESTS_PATH=$(join $(FOREMAN_TESTS_PATH), {api,cli,ui,ui})
```
pgagne: You have two `ui` items here

pgagne:
@rplevka you might want to redo this using 'git mv old_dir new_dir' instead

rplevka:
the travis failure seem completely unrelated. Also the failure doesn't make sense as complains about docstring indentation on an import line.

Makefile
rplevka: thanks, fixed.

JacobCallahan:
@rplevka there are some massive diffs in this PR. was your most recent force push where you switched from ```mv``` to ```git mv```?

rplevka:
@JacobCallahan those massive diffs originate from the fact that there already was UI dir before contraining some files with the same name as those in UI_AIRGUN.
the latest forced-push was me trying to redo the whole operation using `git mv` which unfortunately did not handle the situation any better

JacobCallahan:
@rplevka thanks. just having to check the procedure, since it isn't feasible to check each line of code in this change.

rplevka:
@JacobCallahan not sure which automation job you're referring to. Can you be more specific?

JacobCallahan:
@rplevka check out ```satellite6-standalone-automation```. I believe @ntkathole can help you get it running if you're having problems with it.

ntkathole:
@rplevka to fix Travis remove the module for each entity, we will just need it like https://github.com/SatelliteQE/robottelo/blob/master/docs/api/robottelo.ui.rst

mirekdlugosz:
Since there hasn't been any activity here in over 6 months, and 6.4 is going to reach end of maintenance support in ~2 months, do we want to keep that open? Or just close it right here and now?

JacobCallahan:
@mirekdlugosz it looks like this change has already taken place based on the current robottelo tests directory structure. 
@rplevka am I correct on my previous statement? If so can this be closed, or is there still something that needs to be resolved?

PR #7107, opened 2019-07-03
[6.5.z] move old ui tests to ui_deprecated and ui_airgun to ui

rplevka:
I basically did:

```bash
mv ui ui_deprecated && mv ui_airgun ui
```
the confusing diff is probably because the `ui` dir still exists but contains different files, where some files have the same name so it takes it as changes made to those files


pgagne:
@rplevka you might want to redo this using 'git mv old_dir new_dir' instead

rplevka:
@JacobCallahan 
the test results might be found at <jnkns instance>/job/rplevka-t2-airgun-zalenium/4/

JacobCallahan:
@rplevka sorry for the delay, I'm looking through these results and I'm seeing a number of failures in general, but most of the UI tests are passing. So that proves that your commit is likely not the cause of the failures. ACK from me!

PR #7110, opened 2019-07-03
update metadata for provisioning template tests

rplevka:


tests/foreman/api/test_template_combination.py
```diff
@@ -5,14 +5,12 @@
 
 :CaseAutomation: Automated
 
-:CaseLevel: Acceptance
+:CaseLevel: Component
 
-:CaseComponent: API
+:CaseComponent: ProvisioningTemplate
```
pondrejk: Should be ProvisioningTemplates?

PR #7111, opened 2019-07-04
reverted the manifest_clone

vijay8451:
- Fixed https://github.com/SatelliteQE/satellite6-upgrade/issues/332
- Test result:
```
PASSED2019-07-04 16:12:53 - robottelo - INFO - Started tearDownClass: tests.upgrades.test_subscription/Scenario_manifest_refresh
```


vijay8451:
> ACK
> just a warning, that this approach can be only used once. If there was another test doing manifest refresh, it would need to be able to upload the same manifest (not a clone). Thus we should ensure that the current test does a proepr cleanup

Agree .. will address that in separate PR.

PR #7114, opened 2019-07-05
fixed all content_credentials tests

omkarkhatavkar:
### PR objective
Currently when a table has not any qualified rows and it contains empty row message with some useful text was not return correctly, hence added custom table to return that.

### What was the problem?
The earlier table was read and the empty message was returned a row with column widget which is absolutely wrong and dependent assertions were failed.

### Solution
Created the custom table widget to solve this in Airgun (https://github.com/SatelliteQE/airgun/pull/358) .
Possible Upstream Commit 
Katello/katello@48acd2e#diff-df08c0b5c1111c905e5e022b202c6415R6105

### Test Result 

wait.....

### DIFF
--- a/tests/foreman/ui/test_contentcredential.py
+++ b/tests/foreman/ui/test_contentcredential.py
@@ -28,6 +28,8 @@ from robottelo.datafactory import gen_string
 from robottelo.decorators import fixture, tier2, upgrade
 from robottelo.helpers import get_data_file, read_data_file
 
+table_empty_message = "You currently don't have any Products associated with this Content Credential."
+
 
 @fixture(scope='module')
 def module_org():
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_contentcredential.py
```diff
@@ -28,6 +28,8 @@
 from robottelo.decorators import fixture, tier2, upgrade
 from robottelo.helpers import get_data_file, read_data_file
 
+table_empty_message = "You currently don't have any Products associated with this Content Credential."
```
jyejare: length is > 99. Fix it.

omkarkhatavkar: done!

PR #7116, opened 2019-07-05
Fix testimony validate errors outside of {api,cli,ui}

mirzal:
Mostly update metadata, but also RST parsing errors in virtwho

### DIFF
--- a/tests/foreman/longrun/test_puppet_upgrade.py
+++ b/tests/foreman/longrun/test_puppet_upgrade.py
@@ -2,11 +2,11 @@
 
 :Requirement: Puppet
 
-:CaseAutomation: Automated
+:CaseAutomation: NotAutomated
 
-:CaseLevel: Acceptance
+:CaseLevel: System
 
-:CaseComponent: CLI
+:CaseComponent: ConfigurationManagement
 
 :TestType: Functional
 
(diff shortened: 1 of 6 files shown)


tests/foreman/longrun/test_puppet_upgrade.py
```diff
@@ -2,11 +2,11 @@
...
 :Requirement: Puppet
 
-:CaseAutomation: Automated
+:CaseAutomation: NotAutomated
 
-:CaseLevel: Acceptance
+:CaseLevel: System
 
-:CaseComponent: CLI
+:CaseComponent: ConfigurationManagement
```
vijay8451: I think the more appropriate CaseComponent is `Puppet` here as per Bugzilla component ? what do you think ?

mirzal: I'd say both are equally good. `Puppet` has description "Bugs around Puppet Integration", `ConfigurationManagement` has description "Bugs related to Puppet functionality".

I thought ConfigurationManagement will be slightly better match, but since all these tests are `@stubbed`, I don't think it really matters. If you insist, I can change it to `Puppet`.

vijay8451: @mirzal yeah .. keep it `ConfigurationManagement` , also I just noticed that change to `Puppet` also does not change the component owner as well.

PR #7123, opened 2019-07-09
Port can be changed for SSH Paramiko connection

eko999:
As we known, the default ssh port is 22, but how about ssh to a container? ssh to a kubevirt's vm? their ports may be not 22.
I created this PR because:
 1). we use the container as our jenkins slave, and have to ssh it with container's port, the port is not 22.
 2). we have to login the kubevirt virtual machine by ssh and then execute some commands, the port is not 22.

The usage as:
```
from robottelo import ssh
results = ssh.command(
    cmd='uname -a', 
    hostname='10.xxx.xxx.xxx',
    username='root',
    password='xxx',
    port=53225,
) 
```

### DIFF
--- a/robottelo/ssh.py
+++ b/robottelo/ssh.py
@@ -86,7 +86,7 @@ def _call_paramiko_sshclient():  # pragma: no cover
 
 
 def get_client(hostname=None, username=None, password=None,
-               key_filename=None, timeout=None):
+               key_filename=None, timeout=None, port=22):
     """Returns a SSH client connected to given hostname"""
     if hostname is None:
         hostname = settings.server.hostname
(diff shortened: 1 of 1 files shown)


vijay8451:
@eko999 approach looks really good .. however would you please share some tests result to see if this works fine with existing tests after change ?

eko999:
A test result as below:
```
# pytest  tests/foreman/cli/test_activationkey.py::ActivationKeyTestCase::test_positive_create_with_name
2019-07-09 16:14:57 - conftest - DEBUG - Registering custom pytest_configure

platform linux -- Python 3.5.3, pytest-4.0.2, py-1.8.0, pluggy-0.12.0
rootdir: /root/eko/robottelo, inifile:
plugins: mock-1.10.4, services-1.3.1, xdist-1.29.0, cov-2.7.1, forked-1.0.2
collecting ... 2019-07-09 16:14:58 - conftest - DEBUG - BZ deselect is disabled in settings

collected 1 item                                                                                                                                                            

tests/foreman/cli/test_activationkey.py .                                                                                                                             [100%]

========================================================================= 1 passed in 99.42 seconds =========================================================================
```

PR #7142, opened 2019-07-09
vars cli fix name selection

pondrejk:
```
pytest tests/foreman/cli/test_variables.py -k CRUD
2019-07-09 13:38:43 - conftest - DEBUG - Registering custom pytest_configure

================================================================ test session starts ================================================================
platform linux -- Python 3.7.3, pytest-4.0.2, py-1.7.0, pluggy-0.12.0
plugins: cov-2.7.1, mock-1.10.4, forked-0.2, services-1.3.1, xdist-1.29.0
collecting ... 2019-07-09 13:38:44 - conftest - DEBUG - BZ deselect is disabled in settings

collected 19 items / 18 deselected                                                                                                                  

tests/foreman/cli/test_variables.py .       
 1 passed, 18 deselected, 4 warnings in 165.67 seconds 
```

### DIFF
--- a/tests/foreman/cli/test_variables.py
+++ b/tests/foreman/cli/test_variables.py
@@ -181,7 +181,7 @@ class SmartVariablesTestCase(CLITestCase):
         self.assertEqual(smart_variable['variable'], name)
 
         # Update name and puppet class
-        new_name = valid_data_list()[1]
+        new_name = valid_data_list()[0]
         new_puppet = Puppet.info(
             {u'name': choice(self.puppet_subclasses)['name']})
         SmartVariable.update({


tests/foreman/cli/test_variables.py
```diff
@@ -181,7 +181,7 @@ def test_positive_CRUD(self):
         self.assertEqual(smart_variable['variable'], name)
 
         # Update name and puppet class
-        new_name = valid_data_list()[1]
+        new_name = valid_data_list()[0]
```
mirzal: I'm not sure this is the right approach. `[0]` contains only alphabetic characters, `[1]` contains alphanumeric characters. From what I see, smart variable can contain numbers:
```
# hammer smart-variable update --id 31 --new-variable '213124'
Smart variable [213124] updated.
# hammer smart-variable list
---|----------|---------------|--------|-----------------------
ID | VARIABLE | DEFAULT VALUE | TYPE   | PUPPET CLASS          
---|----------|---------------|--------|-----------------------
31 | 213124   |               | string | access_insights_client
---|----------|---------------|--------|-----------------------
```

Perhaps the problem is that smart variable name has upper length limit lower than 255, and sometimes `valid_data_list()` gives you string that is too long? Then your test should use `gen_string` directly, or provide your own version of `valid_org_data_list` (see https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/api/test_organization.py#L35 ).

pondrejk: @mirzal Not really, the problem is elsewhere. I forgot about the filtered_datapoint thing here https://github.com/SatelliteQE/robottelo/blob/master/robottelo/datafactory.py#L19

valid_org_data_list() is decorated with that, so when run in our automation, the returned list has only one member (randomly selected), hence I got `IndexError: list index out of range`.  

Maybe there's a time for a discussion if we want to have something like that in robottelo, it was added some time ago out of necessity, it kind of randomizes the results which can be seen both as good and bad

mirzal: @pondrejk ah, I see. I thought `filtered_datapoint` only makes list out of dict for backward compatibility. I didn't notice that it also completely changes the meaning based on `settings.run_one_datapoint` value. And you are right, we have this set to True in our automation.

I did not run `settings.configure()` in my test, that's why I was getting completely different result.

Thanks for explanation.

PR #7144, opened 2019-07-09
Fixes bz flagged for 6.6.0 deselected tests

ntkathole:
Bugzillas flagged for 6.6.0 was considered in list of missing version flags. This will fix it.


san7ket:
This blocks running of tests that were blocked due to 66 bz

PR #7151, opened 2019-07-10
globalparam metadata update

pondrejk:


tests/foreman/cli/test_globalparam.py
```diff
@@ -7,7 +7,7 @@
 
 :CaseLevel: Acceptance
 
-:CaseComponent: CLI
+:CaseComponent: Settings
```
mirzal: Shouldn't that be "Parameters", which is for "Bugs related to Parameters such as Global Parameter, Host Parameter, Domain Parameter etc"?

"Settings" is for "Bugs related to the settings feature of Satellite 6".

PR #7161, opened 2019-07-11
Clean-up of docker test files. Because sat 6.6 remove foreman docker …

lvrtelov:
Clean-up of docker test files. Because sat 6.6 remove foreman docker integration - https://redhatmbu.aha.io/epics/SAT-E-159


lvrtelov:
@mirzal Ahh..thanks for reviewing it. I wasn't sure about these test because I just get them after Ales and I was trying to figure it out..I'll take another look at it

tests/foreman/api/test_docker.py
```diff
@@ -1157,30 +1198,80 @@ class DockerUnixSocketContainerTestCase(APITestCase):
 
     @classmethod
     @skip_if_not_set('docker')
+    @skip_if_bug_open('bugzilla', 1478966)
```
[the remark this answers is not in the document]
mirzal: Not sure why github showed me this line is being added...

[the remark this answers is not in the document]
lvrtelov: When I think about it it's not needed there..because it's negative test :)

tests/foreman/api/test_docker.py
```diff
@@ -1149,144 +1146,43 @@ def test_positive_remove_docker_repo_ccv(self):
...
-                    settings.docker.get_unix_socket_url()):
-            with self.subTest(url):
-                resource = entities.DockerComputeResource(url=url).create()
-                self.assertEqual(resource.url, url)
-                self.assertEqual(resource.provider, DOCKER_PROVIDER)
-                resource.delete()
+                description = gen_string('alphanumeric')
                 with self.assertRaises(HTTPError):
-                    resource.read()
+                    entities.Registry(
```
JacobCallahan: can you add a check to this test that would also work when/if Registry is removed from nailgun?

tests/foreman/api/test_docker.py
```diff
@@ -1318,178 +1218,16 @@ def setUp(self):
...
-            pulled from the external registry
-        """
-        repo_name = 'rhel'
-        registry = entities.Registry(
-            url=settings.docker.external_registry_1).create()
-        try:
-            container = entities.DockerRegistryContainer(
-                compute_resource=self.compute_resource,
+        with self.assertRaises(HTTPError):
+            self.compute_resource = entities.DockerComputeResource(
```
JacobCallahan: same as my comment above.

tests/foreman/api/test_docker.py
```diff
@@ -1502,173 +1240,37 @@ class DockerUnixSocketContainerTestCase(APITestCase):
...
-        :CaseImportance: Critical
+        :expectedresults: Compute Resource is not created.
         """
         for name in valid_data_list():
-            with self.subTest(name):
-                description = gen_string('alphanumeric')
-                registry = entities.Registry(
-                    description=description,
+            with self.assertRaises(HTTPError):
+                entities.DockerComputeResource(
```
JacobCallahan: same as my comment above.

tests/foreman/api/test_docker.py
```diff
@@ -1502,173 +1240,37 @@ class DockerUnixSocketContainerTestCase(APITestCase):
...
-        :CaseImportance: Critical
+        :expectedresults: Compute Resource is not created.
         """
-        registry = entities.Registry(url=self.url).create()
-        registry.delete()
-        with self.assertRaises(HTTPError):
-            registry.read()
+        for name in valid_data_list():
+            with self.assertRaises(HTTPError):
+                entities.DockerComputeResource(
```
JacobCallahan: same as my comment above.

JacobCallahan:
Oh, also can you squash your commits?

tests/foreman/api/test_docker.py
mirzal: When support for Registries (and other entities) is removed from Nailgun, these tests should be removed. Personally, I would leave them as they are and let them fail after Nailgun changes.

Lucie, if you want to go Jake's way, you need:
```python
with self.assertRaises((HTTPError, AttributeError)):
```

lvrtelov: Personally, I agree with @mirzal. Because - first why would we need this test if we drop this entity(in my POV it will become test of Nailgun instead of Satellite) and - second I think it would be only wasting of resources to have this type of tests in robotello if we will agree to drop these entities from Nailgun.

JacobCallahan: To answer, first, we need to ask what the purpose of the test is. The purpose is to make sure that the feature doesn't exist on Satellite.
How do we do that? We use the currently existing nailgun entity to attempt to hit the previous end point. This should fail, so we pass the test.

If the entity is removed in nailgun, then the test will error. Is this correct? No. The feature still doesn't exist in Satellite, so the test should pass. 

The only benefit to not having this deprecation handled in-test is to serve as a reminder to the component owner to remove the test? So again, is the purpose of the test to check that a feature doesn't exist in Satellite, or is it a time-delayed reminder for a QE?

mirzal: Once we remove support for entity in nailgun, we can't say anything about Satellite anymore. Feature could be re-introduced in Satellite, but since there is no nailgun support, we have no way of checking that.

Once nailgun support is removed, this test becomes broken (outdated, if you will). It will serve no purpose in context of testing Satellite. Right thing to do will be removing test. I prefer it to fail loudly, so it's easy for us to find. Otherwise, it will just waste shared resources doing nothing of value.

jhutar: Hello @JacobCallahan . Actually "If the entity is removed in nailgun, then the test will error. Is this correct? No. The feature still doesn't exist in Satellite, so the test should pass." is IMO missleading because if I remove "Host" entity from Nailgun, it does not mean it was removed from Satellite.
Anyway, probability of reintroducing Docker entity in Satellite by accident is high now right after the change, but will be very small in few months when we will remove it from Nailgun.
We can not afford running tests testing something is not there indefinitely, esp. given that "automation under 12 hours" OKR.

lhellebr: I agree with Miroslaw and Lucie. Entity support in Nailgun doesn't tell us anything about the existence or nonexistence of the endpoint.  A test that uses functionality A to test functionality B is missing, that fails on missing A, is a broken test.

However, Jacob originally suggested the following: "_can you add a check to this test that would also work when/if Registry is removed from nailgun?_"
That makes sense because exactly what I said in previous paragraph - a test won't test anything and will be broken. Testing this the way that isn't dependent of Nailgun would be a fix. We want to be able to test endpoint nonexistence without Nailgun. If we wanted to test that, we'd need to use API directly (curl...).

I believe both is possible: using Nailgun for now and later (when it fails) fixing this by using Nailgun independent approach, or doing it now.

JacobCallahan: @mirzal If failing loudly (via test error) is the intention, that is one thing. I'm personally not a fan.
@jhutar this test wouldn't pass once the entity is removed from nailgun as the assert if for an HTTPError.
If everyone is good with the test erroring temporarily (until it is removed), then I won't block this PR.

jhutar: Thank you @JacobCallahan ! What is the process now? Should Lucie just get one more ack and ignore your nack when merging?

JacobCallahan: @mirzal @jhutar @lvrtelov  I need an answer as to whether you both are fine with this test failing via error (once nailgun removed the entity). If so, then we can merge as-is.

jhutar: @JacobCallahan Yes, I'm OK with this test failing via error (once nailgun removed the entity). Thank you!

mirzal: @JacobCallahan yes, I believe they should fail loudly, so we are notified and can ensure they are removed once they are no longer valid.

JacobCallahan: Fair enough.

Issue #7173, opened 2019-07-15
Option --environment-id is deprecated. Use --puppet-environment-id instead


ntkathole:
Many tests are failing in robottelo due to $subject 
```
tests.foreman.cli.test_host*
tests.foreman.cli.test_classparameters*
tests.foreman.cli.test_hostgroup*
tests.foreman.cli.test_variables*
```

Ref : https://bugzilla.redhat.com/show_bug.cgi?id=1713222

PR #7174, opened 2019-07-15
Fixing failed tests because #7173

omkarkhatavkar:
### PR Objective 
To fix all failing tests because of #7173 

### What is the change 
Now the environment option has changed from environment to puppet-environment hence all tests are failing    

### Test Result 
Some tests are failing not because of this commit.
 
```
Testing started at 4:34 PM ...
/home/okhatavk/Satellite/robottelo/env/bin/python /home/okhatavk/Downloads/pycharm-community-2018.1.4/helpers/pycharm/_jb_pytest_runner.py --target test_host.py::HostCreateTestCase
Launching py.test with arguments test_host.py::HostCreateTestCase in /home/okhatavk/Satellite/robottelo/tests/foreman/cli

2019-07-15 16:34:35 - conftest - DEBUG - Registering custom pytest_configure

2019-07-15 16:34:35 - conftest - DEBUG - Fetching BZs to deselect...

2019-07-15 16:34:57 - conftest - DEBUG - Deselected tests reason: BZ resolution ['1479291', '1414821', '1204686', '1226425', '1475443', '1310422', '1156555', '1278917', '1199150', '1217635', '1147100', '1311113', '1214312', '1230902', '1347658', '1487317']

2019-07-15 16:34:57 - conftest - DEBUG - Deselected tests reason: missing version flag ['1479291', '1682940', '1194476', '1204686', '1436209', '1226425', '1475443', '1610309', '1625783', '1310422', '1156555', '1489322', '1278917', '1581628', '1199150', '1217635', '1147100', '1311113', '1214312', '1378442', '1230902', '1347658', '1487317', '1701118', '1701132']

2019-07-15 16:34:57 - conftest - DEBUG - Deselected tests reason: sat-backlog ['1479291', '1682940', '1194476', '1204686', '1436209', '1226425', '1475443', '1610309', '1625783', '1310422', '1156555', '1489322', '1278917', '1581628', '1199150', '1217635', '1147100', '1311113', '1214312', '1378442', '1230902', '1347658', '1487317', '1701118', '1701132']

platform linux -- Python 3.4.8, pytest-4.0.2, py-1.7.0, pluggy-0.9.0
plugins: xdist-1.26.1, services-1.3.1, mock-1.10.0, forked-1.0.2, cov-2.6.12019-07-15 16:34:57 - conftest - DEBUG - Collected 34 test cases

collected 34 items

test_host.py 


======== 6 failed, 27 passed, 1 skipped, 6 warnings in 3138.27 seconds =========
```

### DIFF
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -149,7 +149,7 @@ class HostCreateTestCase(CLITestCase):
             'search': u'content_view="{0}"'.format(cls.puppet_cv['name'])})[0]
         cls.puppet_class = Puppet.info({
             'name': puppet_modules[0]['name'],
-            'environment': cls.puppet_env['name'],
+            'puppet-environment': cls.puppet_env['name'],
         })
         # adding org to a puppet env
         Org.set_parameter({
@@ -576,7 +576,7 @@ class HostCreateTestCase(CLITestCase):
         """
         host = make_fake_host({
             'puppet-class-ids': self.puppet_class['id'],
-            'environment-id': self.puppet_env['id'],
+            'puppet-environment-id': self.puppet_env['id'],
             'organization-id': self.new_org['id'],
         })
         host_classes = Host.puppetclasses({'host-id': host['id']})
@@ -597,7 +597,7 @@ class HostCreateTestCase(CLITestCase):
         """
         host = make_fake_host({
             'puppet-classes': self.puppet_class['name'],
-            'environment': self.puppet_env['name'],
+            'puppet-environment': self.puppet_env['name'],
             'organization-id': self.new_org['id'],
         })
         host_classes = Host.puppetclasses({'host': host['name']})
@@ -813,12 +813,12 @@ class HostCreateTestCase(CLITestCase):
         # Create hostgroup with associated puppet class
         host = make_fake_host({
             'puppet-classes': self.puppet_class['name'],
-            'environment': self.puppet_env['name'],
+            'puppet-environment': self.puppet_env['name'],
             'organization-id': self.new_org['id'],
         })
         # Override one of the sc-params from puppet class
         sc_params_list = SmartClassParameter.list({
-            'environment': self.puppet_env['name'],
+            'puppet-environment': self.puppet_env['name'],
             'search': u'puppetclass="{0}"'.format(self.puppet_class['name'])
         })
         scp_id = choice(sc_params_list)['id']
@@ -840,12 +840,12 @@ class HostCreateTestCase(CLITestCase):
         # Create hostgroup with associated puppet class
         host = make_fake_host({
             'puppet-classes': self.puppet_class['name'],
-            'environment': self.puppet_env['name'],
+            'pupppet-environment': self.puppet_env['name'],
             'organization-id': self.new_org['id'],
         })
         # Override one of the sc-params from puppet class
         sc_params_list = SmartClassParameter.list({
-            'environment': self.puppet_env['name'],
+            'puppet-environment': self.puppet_env['name'],
             'search': u'puppetclass="{0}"'.format(self.puppet_class['name'])
         })
         scp_id = choice(sc_params_list)['id']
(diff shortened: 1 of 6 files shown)


tests/foreman/cli/test_host.py
```diff
@@ -926,7 +926,6 @@ def test_positive_list(self):
             self.assertTrue(client.subscribed)
             hosts = Host.list({
                 'organization-id': self.new_org['id'],
-                'environment-id': self.new_lce['id'],
```
ntkathole: why we completely remove environment-id from here?

omkarkhatavkar: As per below output, we don't have environment-id then I thought to use `'puppet-environment-id '` but if you see the test case we are passing lifecycle-environment id to that. isn't that test case is kind wrong? and I'm not sure how is running and passing?     

```
# hammer host list --help | grep environ
 --puppet-environment PUPPET_ENVIRONMENT_NAME  Puppet environment name
 --puppet-environment-id PUPPET_ENVIRONMENT_ID  

```

jyejare:
@omkarkhatavkar Why the Test Execution has 6 failures and are these product failures? If yes we should skip them with bug else fix them from automation.

pondrejk:
hi @omkarkhatavkar  code changes look ok but I see test results just from test_host.py but the changes were also made for classparameters, hostgroups, and variables, so we should check if those work too

pondrejk:
test_puppetclass is also affected

omkarkhatavkar:
> hi @omkarkhatavkar code changes look ok but I see test results just from test_host.py but the changes were also made for classparameters, hostgroups, and variables, so we should check if those work too

As mentioned in #7173 I just consider all other Test Classes. let me check of test_puppetclass as well.

omkarkhatavkar:
> @omkarkhatavkar Why the Test Execution has 6 failures and are these product failures? If yes we should skip them with bug else fix them from automation.

@jyejare 
This fix was only related to #7173 and 6 failure apart for that won't satisfied with this PR. 6 failures are different with each other will take time for making them as an available fix. I do understand that we should not have a bug or fix available to those tests.  We can handle separate PR.

mirzal:
At the risk of stating the obvious: `environment` and `environment-id` are deprecated, but still supported. They do work in simplest case (`hammer hostgroup create --name test --environment-id 1`).

If switching from `--environment` to `--puppet-environment` actually causes the test to pass, it means that something (maybe specific options combination) mandates usage of new option, which means that promise of backward compatibility is not delivered. **This is product bug**.

Do we have bugzilla filled for that?

omkarkhatavkar:
> At the risk of stating the obvious: `environment` and `environment-id` are deprecated, but still supported. They do work in simplest case (`hammer hostgroup create --name test --environment-id 1`).
> 
> If switching from `--environment` to `--puppet-environment` actually causes the test to pass, it means that something (maybe specific options combination) mandates usage of new option, which means that promise of backward compatibility is not delivered. **This is product bug**.
> 
> Do we have bugzilla filled for that?

@mirzal Can you please add same comments here https://bugzilla.redhat.com/show_bug.cgi?id=1713222, We can wait for this PR to get merged.

omkarkhatavkar:
> test_puppetclass is also affected

@pondrejk thanks for identifying that, I have added https://github.com/SatelliteQE/robottelo/pull/7174/commits/ec143b8e37fefebe477ad66d0016403d15e91a95 as well.

mirzal:
@omkarkhatavkar I have created https://bugzilla.redhat.com/show_bug.cgi?id=1730026 and added it in comment in issue you mentioned. I also sent email to satqe with more details.

What I meant is that we need to go through all tests one by one and run them against latest snap with this patch applied and without this patch applied - possibly using Robottelo from before #7176 was merged. If one test fails when using `--environment`, but passes when using `--puppet-environment`, then it's a bug of new options not being backwards compatible. It needs to be reported, so PM can understand full scope of the issue.

omkarkhatavkar:
> o PM can understand

@mirzal It's a valid point, I do agree about it. Deprecation means discouragement of the use of some terminology, feature, design, or practice. I did not think about it earlier when you raised this. Instead of deprecation if it was mentioned unsupported would have made more sense.

As you already raised this issue in this PR considering as Product Bug, You are the correct person to raise this issue, I dont wanted to encroachment here. As QE I do understand that It needs to be reported because of the backwards compatibility issue.

omkarkhatavkar:
No need for this PR, closing

PR #7194, opened 2019-07-17
ansible recurring rex -- fixed waiting for execution

pondrejk:
The previous way of polling for rex result in test_positive_run_reccuring_job for ansible caused automation failures -- it checked only the first invocation of recurring job which led to asserting too soon. Fixed to match regular rex ttest

### DIFF
--- a/tests/foreman/cli/test_remoteexecution.py
+++ b/tests/foreman/cli/test_remoteexecution.py
@@ -36,7 +36,6 @@ from robottelo.constants import (
 )
 from robottelo.decorators import (
     bz_bug_is_open,
-    skip_if_bug_open,
     skip_if_not_set,
     stubbed,
     tier3,
@@ -154,11 +153,12 @@ class TestRemoteExecution():
             raise AssertionError(result)
 
     @tier3
-    @skip_if_bug_open('bugzilla', 1451675)
     def test_positive_run_job_effective_user_by_ip(self, fixture_vmsetup, fixture_org):
         """Run default job template as effective user on a host by ip
 
         :id: 0cd75cab-f699-47e6-94d3-4477d2a94bb7
+
+        :bz: 1451675
 
         :expectedresults: Verify the job was successfully run under the
             effective user identity on host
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_remoteexecution.py
```diff
@@ -583,16 +583,7 @@ def test_positive_run_reccuring_job(self, fixture_vmsetup, fixture_org):
...
-        pending_state = u'1'
-        for _ in range(10):
-            if pending_state != u'0':
-                invocation_info = JobInvocation.info({
-                    'id': invocation_command[u'id']})
-                pending_state = invocation_info[u'pending']
-                sleep(30)
-            else:
-                break
+        sleep(150)
```
omkarkhatavkar: Putting hard sleep is discouraged in robottelo, do you have some better solution with the use of wait_for?

pondrejk: So why is the wait_for better? It's hard for me to say as the tool has no documentation. I'd say here the use sleep is justified, I create a recurring job with a fixed set of iterations, so the time of execution is quite predictable, I set the time accordingly with some buffer, what is the risk in that?

mirzal: Explicit sleeps are widely considered one of code smells in automation. If something happens before said time, we waste time being idle. If something happens slightly after that time, we fail without good reason. Which usually leads to increasing sleep time, which only results in more time wasted being idle.

That's why we should react to events (if we can subscribe to them), or poll for status.

Now, I think I understand your problem with current code. It polls couple of times when we know there is no way task will be completed yet. It will be completed soon after second-to-last poll, but since polling interval is big, we waste quite some time being idle. Reducing polling interval is not really an option on the table.

Personally, I would explicitly sleep for 2 minutes (because we know task can't complete before that). Then I would poll for completion with high frequency (every 5 seconds, or less). This polling needs some upper boundary, like 30 seconds or something.

This is robust enough to allow for variable execution time (due to network slowness etc), but also tries to minimize test running time.

(I'm completely ignoring `wait_for` here, as it's not really part of the conversation. Yes, we introduced it as dependency of Robottelo, yes, we use it in some other parts of code, yes, it basically does the main loop for you. If you don't want to use it and prefer to code loop yourself, I'm ok with that.)

pondrejk: The main problem with previous code was polling at the wrong place (JobInvocation instead of RecurringLogic), which caused the failure. I'd like to get it merged before the next snap so that we actually test this path. Anyway, if it comes out wrong it will get back on my plate.

As for the approach to timing I filed this  #7204 to review it in a more systematic way as the sleep approach not exclusive to this test. 

@omkarkhatavkar @mirzal does it sound acceptable to you?

omkarkhatavkar: I'm good in that case.

pondrejk: talked to Mirek via other channel, he won't be available for ack today due to travel, he agreed to continue the discussion in #7204

PR #7201, opened 2019-07-18
Added skip for bug#1716429

vijay8451:
- https://bugzilla.redhat.com/show_bug.cgi?id=1716429

### DIFF
--- a/tests/foreman/cli/test_repository.py
+++ b/tests/foreman/cli/test_repository.py
@@ -2064,6 +2064,7 @@ class OstreeRepositoryTestCase(CLITestCase):
                 self.assertEqual(new_repo['name'], name)
                 self.assertEqual(new_repo['content-type'], u'ostree')
 
+    @skip_if_bug_open('bugzilla', 1716429)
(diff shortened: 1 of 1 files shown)


mirzal:
I wonder if we shouldn't just rewrite / remove the test. Apparently, nobody noticed this changed behavior in over a year (6.5 and 6.4 cycle) and we can't name any negative consequence of this change, except "test is failing". Maybe this test never checked something that is part of Satellite contract and it doesn't really provide value.

vijay8451:
> I wonder if we shouldn't just rewrite / remove the test. Apparently, nobody noticed this changed behavior in over a year (6.5 and 6.4 cycle) and we can't name any negative consequence of this change, except "test is failing". Maybe this test never checked something that is part of Satellite contract and it doesn't really provide value.

We should keep this test as it is until bug is closed with fix/nofix and  take call accordingly after that.
Also not sure how same was not notice before(assume skipped due to another but but not sure).

vijay8451:
ping @mirzal

PR #7222, opened 2019-07-23
Deleted duplicated tests in test_hostgroup, some tests merged into one

xdragu01:
Before refactoring:
```
tests/foreman/cli/test_hostgroup.py .................................                                                                                                                  [100%]

========================================================================== 33 passed, 0 warnings in 1207.39 seconds ==========================================================================
```
After refactoring:
```
tests/foreman/cli/test_hostgroup.py .......                                                                                                                                            [100%]

=========================================================================== 7 passed, 0 warnings in 267.61 seconds ===========================================================================
```
Tests for smartvariables were redundat. See https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/ui/test_smartvariable.py#L756

### DIFF
--- a/tests/foreman/cli/test_hostgroup.py
+++ b/tests/foreman/cli/test_hostgroup.py
@@ -15,8 +15,6 @@
 
 :Upstream: No
 """
-from random import choice
-
 from fauxfactory import gen_integer
 from robottelo.cleanup import capsule_cleanup
 from robottelo.cli.base import CLIReturnCodeError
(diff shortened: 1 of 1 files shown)


pondrejk:
We were discussing this with Peter while he was working on this, and I realize these removals might be controversial, on the other hand, there're some arguments for it.

- for test_positive_create_with_synced_content, the purpose of the related bz was to find if an option is exposed in the api, the resulting test ended up synchronizing the rhel kickstart repo taking up half of the execution time of the test class. So I lean towards removal as there are some upstream katello tests checking the kickstart-repo option at the model level

- for smart variables and scparams tests, the decision was due to some setting and listing of these parameters at the host group level is being done in the test_smartclassparameter.py and test_variables.py ui tests. I realize UI is not the same as cli, though I wonder what is the risk of not having those tests in cli. Another option is to do this testing within test_positive_create_with_multiple_entities_and_delete, though that one is quite lengthy already

PR #7233, opened 2019-07-30
fix accidental change and add skip for bugzilla

omkarkhatavkar:
reverting to older commit code as conten_view_filter represent handle empty table issue internally. No need of conversion this was added accidentally

### DIFF
--- a/tests/foreman/ui/test_contentview.py
+++ b/tests/foreman/ui/test_contentview.py
@@ -1642,7 +1642,6 @@ def test_positive_remove_cv_version_from_default_env(session, module_org):
         assert ENVIRONMENT not in cvv['Environments']
 
 
-@skip_if_bug_open('bugzilla', 1729153)
 @tier2
 def test_positive_remove_promoted_cv_version_from_default_env(session, module_org):
     """Remove promoted content view version from Library environment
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_contentview.py
```diff
@@ -1642,6 +1642,7 @@ def test_positive_remove_cv_version_from_default_env(session, module_org):
         assert ENVIRONMENT not in cvv['Environments']
 
 
+@skip_if_bug_open('bugzilla', 1729153)
```
vijay8451: I think above Bug is not applicable for this test , can you please recheck ? also I have seen that this test did not fail in last snap too.

tests/foreman/ui/test_contentview.py
```diff
@@ -2078,7 +2079,7 @@ def test_positive_remove_filter(session, module_org):
         assert session.contentviewfilter.search(
             cv.name, filter_name)[0]['Name'] == filter_name
         session.contentviewfilter.delete(cv.name, filter_name)
-        assert session.contentviewfilter.search(cv.name, filter_name)[0]['Name'] != filter_name
+        assert not session.contentviewfilter.search(cv.name, filter_name)
```
vijay8451: do we requires to change it as well ? I doubt .

vijay8451: Ignore above ... it looks fine.

tests/foreman/ui/test_contentview.py
omkarkhatavkar: Hmm let correct that

PR #7242, opened 2019-08-01
fix of the repository name for rhst7_64 and rhst7_65

tstrych:
Hello I think, there was a bug in the names of the `REPOS rhst7_64` and `rhst7_65`. So I fix them.
Please contact me and I will send you log where I debug the change. 

I understand why we need this in our automation, but this is not good way. After new release we will add another rhst ? 
And imagine situation, when you need to work with REPO which is not yet on CDN. Is there a way how we can easily do that in robottelo ?


tstrych:
@pondrejk problem I discuss with you.

PR #7244, opened 2019-08-01
RHEL7 needs 2GB RAM and install media is needed

jhutar:
I do not know why it worked from time to time before. Anyway, now we have installation media and also VM is created with 2 GB RAM which is needed by RHEL7 (it won't boot with 1 GB).


jhutar:
```
time pytest tests/foreman/ui/test_host.py::test_positive_delete_libvirt --no-print-logs --capture=no --exitfirst 2>&1 | tee /tmp/pytest.log
[...]
==================== 1 passed, 3 warnings in 372.57 seconds ====================

real	6m13.235s
user	0m16.057s
sys	0m5.350s
```

PR #7249, opened 2019-08-02
implement docstring tokens recording as xml properties

rplevka:
This automatically parses tokens from the module, class and test case docstrings using `:([A-Za-z_-]+): ([0-9a-zA-Z-_\n]+)` regexp on the pytest runtime.
Then it records them as `<property>` elements in the generated junit xml file.

This can be used by all the tools working with the tags (polarion upload, report portal upload, etc.)

```xml
   <testsuite errors="0" failures="0" name="pytest" skipped="0" tests="1" time="91.310">
    <properties>
      <property name="start_time" value="1564739387819"/>
    </properties>
    <testcase classname="tests.foreman.api.test_bookmarks.BookmarkTestCase" file="tests/foreman/api/test_bookmarks.py" line="156" name="test_negative_create_empty_query" time="91.197">
      <properties>
        <property name="start_time" value="1564739395254"/>
        <property name="Requirement" value="Bookmarks"/>
        <property name="CaseAutomation" value="Automated"/>
        <property name="CaseLevel" value="Acceptance"/>
        <property name="CaseComponent" value="Usability"/>
        <property name="TestType" value="Functional"/>
        <property name="CaseImportance" value="Critical"/>
        <property name="Upstream" value="No"/>
        <property name="id" value="674d569f-6f86-43ba-b9cc-f43e05e8ab1c"/>
        <property name="expectedresults" value="Error"/>
      </properties>
      <system-out>...</system-out>
    </testcase>
  </testsuite>
```

### DIFF
--- a/tests/foreman/conftest.py
+++ b/tests/foreman/conftest.py
@@ -4,6 +4,7 @@ import datetime
 import logging
 
 import pytest
+import re
 try:
     from pytest_reportportal import RPLogger, RPLogHandler
 except ImportError:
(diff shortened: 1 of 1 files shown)


rplevka:
@mirzal -
I'll try to address the `parent` issue to add some more checking
I'll also address the varying case of the tokens to put them all to lowercase
It doesn't handle multiline metadata - We're not interested in these and as a matter of fact, i'll probably put these on a blacklist, or I'll rather come up with a whitelist of tokens i'm after.

- I tried to use the code from testimony but it didn't find any way to do that elegantly. The `DocstringParser` seem to use methods from `ast` lib, which gave me errors on passing the classes and functions to it.

JacobCallahan:
@rplevka +1 to a whitelist

ntkathole:
@rplevka can you please add more details on what we are trying to achieve here? We already have Betelgeuse to parse docstrings and create xml. Asking because the xml file in description is quite conflicting with the xml generated by Betelgeuse. 
Example :  In context to polarion, property is custom field and testcase has some inbuild fields, thus, 
`<property name="id" value="674d569f-6f86-43ba-b9cc-f43e05e8ab1c"/>` -> this corresponds to custom field, while it should be testcase field like
`<testcase id="674d569f-6f86-43ba-b9cc-f43e05e8ab1c">`

PR #7251, opened 2019-08-05
Add 6 new cases for virt-who configure UI testing

eko999:
Some new functions are added to utils.py:
```
    get_configure_id()
    get_configure_file()
    get_configure_option()
    get_configure_command()
```
6 new cases are added to test_ui_virtwho.py:
```
    test_positive_debug_option():
    test_positive_interval_option():
    test_positive_hypervisor_id_option():
    test_positive_filtering_option():
    test_positive_proxy_option():
    test_positive_virtwho_roles():
```


eko999:
Test Result as:
``# pytest test_ui_virtwho.py 
2019-08-05 11:34:00 - conftest - DEBUG - Registering custom pytest_configure

platform linux -- Python 3.7.3, pytest-4.6.3, py-1.7.0, pluggy-0.12.0
rootdir: /root/eko/robottelo, inifile: tox.ini
plugins: mock-1.10.4, xdist-1.29.0, cov-2.7.1, forked-1.0.2, services-1.3.1, flake8-1.0.1
collecting ... 2019-08-05 11:34:21 - conftest - DEBUG - Collected 8 test cases

2019-08-05 11:34:21 - conftest - DEBUG - Deselected test tests.foreman.virtwho.test_ui_virtwho.test_positive_filtering_option

collected 8 items / 1 deselected / 7 selected                                                                                                                               

test_ui_virtwho.py .......                                                                                                                                            [100%]

================================================================= 7 passed, 1 deselected in 2416.98 seconds =================================================================
`

```

File: tests/foreman/virtwho/utils.py [hunk 1]
```diff
@@ -180,43 +236,47 @@ def _get_hypervisor_mapping(logs):
...
+    # Delete the hypervisor entry and always make sure it's new.
+    hosts = Host.list({'search': hypervisor_name})
+    for index, host in enumerate(hosts):
+        Host.delete({'id': host['id']})
+    runcmd("systemctl restart virt-who")
+    return hypervisor_name, guest_name
 
 
-def deploy_configure_by_command(command, debug=True):
+def deploy_configure_by_command(command, debug=False):
```
vijay8451: Any specific reason to move parameter from `debug=True` to `debug=False` ?

tests/foreman/virtwho/test_ui_virtwho.py
```diff
@@ -134,3 +128,220 @@ def test_positive_deploy_configure_by_script(session, form_data):
...
+
+    :CaseImportance: Medium
+    """
+    name = gen_string('alpha')
+    form_data['name'] = name
+    with session:
+        session.virtwho_configure.create(form_data)
+        config_id = get_configure_id(name)
+        config_command = get_configure_command(config_id)
+        deploy_configure_by_command(config_command)
```
vijay8451: Just a thought , can we add assertion for UI `Enable debugging output?` as well to make sure both(UI and file- /etc/sysconfig/virt-who ) end same reflect .

tests/foreman/virtwho/test_ui_virtwho.py
```diff
@@ -134,3 +128,220 @@ def test_positive_deploy_configure_by_script(session, form_data):
...
+        assert get_configure_option('VIRTWHO_DEBUG', VIRTWHO_SYSCONFIG) == '1'
+        session.virtwho_configure.edit(name, {'debug': False})
+        deploy_configure_by_command(config_command)
+        assert get_configure_option('VIRTWHO_DEBUG', VIRTWHO_SYSCONFIG) == '0'
```
vijay8451: same as above.

tests/foreman/virtwho/test_ui_virtwho.py
```diff
@@ -134,3 +128,220 @@ def test_positive_deploy_configure_by_script(session, form_data):
...
+            'Every 4 hours': '14400',
+            'Every 8 hours': '28800',
+            'Every 12 hours': '43200',
+            'Every 24 hours': '86400',
+            'Every 2 days':  '172800',
+            'Every 3 days':  '259200',
+        }
+        for option, value in sorted(intervals.items(), key=lambda item: int(item[1])):
+            session.virtwho_configure.edit(name, {'interval': option})
+            deploy_configure_by_command(config_command)
```
vijay8451: same as above .. add assertion for UI end too.

tests/foreman/virtwho/test_ui_virtwho.py
```diff
@@ -134,3 +128,220 @@ def test_positive_deploy_configure_by_script(session, form_data):
...
+
+    :CaseImportance: Medium
+    """
+    name = gen_string('alpha')
+    form_data['name'] = name
+    with session:
+        session.virtwho_configure.create(form_data)
+        config_id = get_configure_id(name)
+        config_command = get_configure_command(config_id)
+        intervals = {
```
vijay8451: There is no harm to check for each interval type .. however we are just repeating the same thing in functionality wise .
Can we just keep only one/two from UI (as that also help to check around the same)... and cover everything in way from API/CLI ... your thought on it ?

tests/foreman/virtwho/test_ui_virtwho.py
```diff
@@ -134,3 +128,220 @@ def test_positive_deploy_configure_by_script(session, form_data):
...
+    :CaseImportance: Medium
+    """
+    name = gen_string('alpha')
+    form_data['name'] = name
+    with session:
+        session.virtwho_configure.create(form_data)
+        config_id = get_configure_id(name)
+        config_command = get_configure_command(config_id)
+        config_file = get_configure_file(config_id)
+        values = ['uuid', 'hwuuid', 'hostname']
```
vijay8451: Again there is no harm to check for each `Hypervisor ID` .. however we are just repeating the same thing in functionality wise .
Can we just keep only one/two from UI (as that also help to check around the same)... and cover everything in way from API/CLI ... your thought on it ?

tests/foreman/virtwho/test_ui_virtwho.py
```diff
@@ -134,3 +128,220 @@ def test_positive_deploy_configure_by_script(session, form_data):
...
+            'filtering': 'Blacklist',
+            'filtering_content.exclude_hosts': regex,
+        }
+        if hypervisor_type == 'esx':
+            whitelist['filtering_content.filter_host_parents'] = regex
+            blacklist['filtering_content.exclude_host_parents'] = regex
+        # Update Whitelist and check the result
+        session.virtwho_configure.edit(name, whitelist)
+        deploy_configure_by_command(config_command)
+        assert regex == get_configure_option('filter_hosts', config_file)
```
vijay8451: why are you not validating the `assert` based on UI search ? and same for all below `assert` where used in same way ..

tests/foreman/virtwho/utils.py
```diff
@@ -142,6 +145,59 @@ def get_virtwho_status():
...
+    :raises: VirtWhoError: If failed to get the configure info by hammer.
+    """
+    config = VirtWhoConfig.info({'name': name})
+    if 'id' in config['general-information']:
+        return config['general-information']['id']
+    else:
+        raise VirtWhoError("No configure id found for {}".format(name))
+
+
+def get_configure_command(config_id, org=1):
```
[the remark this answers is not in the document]
eko999: 'Default_Organization' is used for virt-who testing because the subscription manifest is belongs to the default org when we deployed satellite. Maybe we should get the org_id by the org_name(Default_Organization)

tests/foreman/virtwho/utils.py
eko999: Most of the time, `debug=True` only used by two cases:
test_positive_deploy_configure_by_id()
test_positive_deploy_configure_by_script()
Because these two cases need to check the rhsm.log message with the DEBUG enable.

Other 6 UI cases don't need to check the DEBUG message in rhsm.log, so I think `debug=False` is a common setting, otherwise, we will have many  statement as `deploy_configure_by_command(command, debug=False)`

tests/foreman/virtwho/test_ui_virtwho.py
eko999: It's not enough if we only validate the UI `debug` checked or uncheck, the significance of virtwhon configure is help to create the configure file in the host. After deploy the configure, it will change the /etc/sysconfig/virt-who, and create a configure file in /etc/virt-who.d/. 

So if without deployment and the validation , we are not sure the configure file can be working or not  which is created by the UI.

tests/foreman/virtwho/test_ui_virtwho.py
eko999: as above

tests/foreman/virtwho/test_ui_virtwho.py
eko999: as above

tests/foreman/virtwho/test_ui_virtwho.py
eko999: It will become more quickly if only check one/two  options' values, but we still not sure if we select the interval to 'Every 2 days', and the actually interval value is set to a unexpected value in /etc/sysconfig/virt-who.

As the above, need to make sure: if the UI changes, the related options also should be changed in /etc/sysconfig/virt-who or /etc/virt-who.d/

tests/foreman/virtwho/test_ui_virtwho.py
eko999: as the above, the final goal of virt-who configure is help to change the /etc/sysconfig/virt-who or the configure file in /etc/virt-who.d.

tests/foreman/virtwho/utils.py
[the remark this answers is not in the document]
vijay8451: Yes , you can get the ID by using org_name .

tests/foreman/virtwho/utils.py
vijay8451: okay, make sense.

tests/foreman/virtwho/test_ui_virtwho.py
vijay8451: Just clearing on my comment ...
Along with configuration validation(/etc/sysconfig/virt-who,) add validation at UI end as well ... so that we could see if UI and filesystem level has same thing or different ?

tests/foreman/virtwho/test_ui_virtwho.py
vijay8451: same as above ..

tests/foreman/virtwho/test_ui_virtwho.py
vijay8451: same as above ..

tests/foreman/virtwho/test_ui_virtwho.py
eko999: as the above.

tests/foreman/virtwho/test_ui_virtwho.py
vijay8451: okay.

tests/foreman/virtwho/test_ui_virtwho.py
vijay8451: same as above ..

tests/foreman/virtwho/utils.py
```diff
@@ -180,43 +236,47 @@ def _get_hypervisor_mapping(logs):
...
-        hosts = Host.list({'search': hypervisor_name})
-        for index, host in enumerate(hosts):
-            Host.delete({'id': host['id']})
-        runcmd("systemctl restart virt-who")
-        return hypervisor_name, guest_name
+    hypervisor_name, guest_name = _get_hypervisor_mapping(logs)
+    # Delete the hypervisor entry and always make sure it's new.
+    hosts = Host.list({'search': hypervisor_name})
+    for index, host in enumerate(hosts):
+        Host.delete({'id': host['id']})
```
JacobCallahan: why are you enumerating here?

tests/foreman/virtwho/utils.py
```diff
@@ -142,6 +145,59 @@ def get_virtwho_status():
...
+
+def get_configure_option(option, filename):
+    """Return the option's value for the specific file.
+    :param str option: the option name in the configure file
+    :param str filename: the configure file, it could be:
+        /etc/sysconfig/virt-who
+        /etc/virt-who.d/virt-who-config-{}.conf
+    :raises: VirtWhoError: If no this option name in the file.
+    """
+    cmd = "grep -v '^#' {} |grep ^{}".format(filename, option)
```
JacobCallahan: nitpick, but can you add a space between | and grep?

tests/foreman/virtwho/test_ui_virtwho.py
eko999: I have a little afraid it will spend more time to run the case if validating all the options' value, but according to the latest test result with the new commit, it's acceptable:
` 7 passed, 1 skipped in 2829.53 seconds`

tests/foreman/virtwho/utils.py
eko999: Yeah, it should be dropped because we don't need to use the `index` in the current method.

tests/foreman/virtwho/utils.py
eko999: Nice

tests/foreman/virtwho/test_ui_virtwho.py
eko999: @vijay8451  please review again with the new commit.

eko999:
Depends on the PR in airgun: https://github.com/SatelliteQE/airgun/pull/386

PR #7257, opened 2019-08-06
Fixes 3 Upgrade Scenarios of Capsule, SyncPlan and Client

jyejare:
Fixes Below Upgrade Scenarios:

1. tests.upgrades.test_capsule.Scenario_capsule_sync.test_post_user_scenario_capsule_sync
```
Was Failing with : Timeout error, increased to 3600, It actually runs for 46 minutes
```
2. tests.upgrades.test_client.Scenario_upgrade_new_client_and_package_installation.test_post_scenario_postclient_package_installation
```
Was Failing with: KeyError : Capital RHEL7 was not found, lowered it.
```
3. tests.upgrades.test_syncplan.ScenarioSyncPlan.test_post_sync_plan_migration
```
Was Failing with: The last sync format is completely changed in 6.6 from 6.5. So removed the assertion as its not the intend of the test. I will handle if its still required.

Change:
E       - 2858-05-25 21:24:39 UTC
E       ?     ^  ^            ^^^
E       + 2858/05/25 21:24:39 +0000
E       ?     ^  ^            ^^^^^
```

### DIFF
--- a/tests/upgrades/test_syncplan.py
+++ b/tests/upgrades/test_syncplan.py
@@ -96,6 +96,9 @@ class ScenarioSyncPlan(APITestCase):
         self.assertEqual(product.sync_plan.id, sync_plan.id)
         self.assertEqual(sync_plan.name, entity_data.get("sync_plan_name"))
         self.assertEqual(sync_plan.interval, entity_data.get("interval"))
+        if '-' in entity_data.get('sync_date'):
+            entity_data['sync_date'] = ''.join(
+                entity_data.get('sync_date').replace('-', '/').replace('UTC', '+0000'))
         self.assertEqual(sync_plan.sync_date, entity_data.get("sync_date"))
         # checking sync plan update on upgraded satellite
         sync_plan.interval = SYNC_INTERVAL['custom']
--- a/tests/upgrades/test_capsule.py
+++ b/tests/upgrades/test_capsule.py
@@ -17,7 +17,7 @@
 from fabric.api import execute, run
 from nailgun import entities
 from robottelo.test import APITestCase, settings
-from robottelo.api.utils import promote
+from robottelo.api.utils import promote, call_entity_method_with_timeout
 from upgrade.helpers.tasks import wait_untill_capsule_sync
 from upgrade_tests import post_upgrade, pre_upgrade
 from upgrade_tests.helpers.scenarios import (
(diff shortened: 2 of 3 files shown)


tests/upgrades/test_syncplan.py
```diff
@@ -96,7 +96,6 @@ def test_post_sync_plan_migration(self):
         self.assertEqual(product.sync_plan.id, sync_plan.id)
         self.assertEqual(sync_plan.name, entity_data.get("sync_plan_name"))
         self.assertEqual(sync_plan.interval, entity_data.get("interval"))
-        self.assertEqual(sync_plan.sync_date, entity_data.get("sync_date"))
```
vijay8451: I understand this not the end purpose of this test however we should keep this inorder to keep a check after upgrade and it shall add a plus for us too.

jyejare: If some other function or test ran before this test and capsule sync happened then this is likely to fail and unnecessarily causing this fail test. So thought to remove it. I am still at the side of removing it.

vijay8451: I did not get this one "capsule sync happened then this is likely to fail" ? How this is going to impact the sync plan 'sync date' ?

JacobCallahan: @vijay8451 I believe that field references when the sync plan will trigger next. So if the sync plan hits between when entity_data is collected and when sync_plan is queried, then we could get a false negative result.
@jyejare feel free to correct me if I'm wrong.

vijay8451: @JacobCallahan by default sync plan `Start Date` and `Next Sync` are part of future date so should not get impacted by any upgrade interval.
@jyejare please correct me if i missed any other part in it.

jyejare: @JacobCallahan , You are absolutely correct.

@vijay8451 , That field represents the last sync time and by 'capsule sync' I mean if somehow another test or just capsule CV publish will change that time and as @JacobCallahan  said this test will fail with false negative result.

jyejare: Just removed my last comment as I was confused between sync plan date and capsule sync date.

jyejare: @vijay8451 , Updated the code with sync plan sync date check. Please review.

vijay8451: Thanks @jyejare :+1:

PR #7261, opened 2019-08-08
Added helper to update content in template

vijay8451:
- Added helper to update content in template -> Special going to use around GCE/Azure/EC2 
- Use as below:
```
update = update_provisioning_template(name="Kickstart default finish",
                             old="yum update -t -y",
                             new="#yum update -t -y",
                             org_id=1)
if update:
     ... continue further
```

### DIFF
--- a/robottelo/api/utils.py
+++ b/robottelo/api/utils.py
@@ -909,6 +909,9 @@ def update_provisioning_template(name=None, old=None, new=None, org_id=None):
             'search': 'name="{}"'.format(name),
             'organization_id': org_id
         })[0].read()
-    temp.template = temp.template.replace(old, new)
-    update = temp.update(['template'])
-    return new in update.template
+    if old in temp:
+        temp.template = temp.template.replace(old, new, 1)
+        update = temp.update(['template'])
+        return new in update.template
+    else:
+        raise '{} does not exists in template {}'.format(old, name)
--- a/robottelo/api/utils.py
+++ b/robottelo/api/utils.py
@@ -891,3 +891,24 @@ def attach_custom_product_subscription(prod_name=None, host_name=None):
         query={'search': 'name={0}'.format(prod_name)})[0]
     entities.HostSubscription(host=host.id).add_subscriptions(
         data={'subscriptions': [{'id': product_subscription.id, 'quantity': 1}]})
+
+
+def update_provisioning_template(name=None, old=None, new=None, org_id=None):
+    """ Update provisioning template content
+
+    :param str name: template provisioning name
+    :param str old: current content
+    :param str new: replace content
+    :param int org_id: organization ID
+
+    :return boolean: True/False
+    """
+    temp = entities.ProvisioningTemplate().search(
+        query={
+            'per_page': 1000,
+            'search': 'name="{}"'.format(name),
+            'organization_id': org_id
+        })[0].read()
+    temp.template = temp.template.replace(old, new)
+    update = temp.update(['template'])
+    return new in update.template
--- a/robottelo/api/utils.py
+++ b/robottelo/api/utils.py
@@ -893,13 +893,12 @@ def attach_custom_product_subscription(prod_name=None, host_name=None):
         data={'subscriptions': [{'id': product_subscription.id, 'quantity': 1}]})
 
 
-def update_provisioning_template(name=None, old=None, new=None, org_id=None):
+def update_provisioning_template(name=None, old=None, new=None):
     """ Update provisioning template content
 
     :param str name: template provisioning name
     :param str old: current content
     :param str new: replace content
-    :param int org_id: organization ID
 
     :return boolean: True/False
     """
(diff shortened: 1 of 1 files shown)


robottelo/api/utils.py
```diff
@@ -891,3 +891,24 @@ def attach_custom_product_subscription(prod_name=None, host_name=None):
...
+    """
+    temp = entities.ProvisioningTemplate().search(
+        query={
+            'per_page': 1000,
+            'search': 'name="{}"'.format(name),
+            'organization_id': org_id
+        })[0].read()
+    temp.template = temp.template.replace(old, new)
+    update = temp.update(['template'])
+    return new in update.template
```
omkarkhatavkar: If we pass values  update_provisioning_template('foo', 'too') having template as template = 'testing too', In this case the method will return True. 

Do you think we need to add one more layer of validation to see if old value truly exist in the template? As we are returning the update based on new value exist in the template.

vijay8451: @omkarkhatavkar That is not added to avoid the condition where you have similer as below and condition will be shown always a false `True` hence to avoid that only good to validate the new:

```
i.e.  in below new replaced with #new :

                                                                                                                                     
In [4]: test = 'this is test on <#new>'                                                                                                                                      

In [5]: 'new' in test                                                                                                                                                        
Out[5]: True

In [6]: '#new' in test                                                                                                                                                       
Out[6]: True
```

omkarkhatavkar: @vijay8451 My point was before updating try to check if the old value really exists in the template. I was not saying you to check the new value in the template. If we don't have the old value nothing will replace and we will only update the same thing. 

Whatever you mentioned above has the old value, which means the replacement will work and get updated to the new value. My thought process was to see if an old value exists in the template to ensure that replacement was correct and update calls make sense with checking new value. 

I'm ok if you think that we don't want that validation or assuming this condition will never occur.

vijay8451: Done :)

robottelo/api/utils.py
```diff
@@ -891,3 +891,27 @@ def attach_custom_product_subscription(prod_name=None, host_name=None):
...
+            'per_page': 1000,
+            'search': 'name="{}"'.format(name),
+            'organization_id': org_id
+        })[0].read()
+    if old in temp:
+        temp.template = temp.template.replace(old, new)
+        update = temp.update(['template'])
+        return new in update.template
+    else:
+        return False
```
jyejare: Better if we can raise an error here saying the old isnt available. Else the user will confuse why its failed wheather due to New in update or old in temp.

robottelo/api/utils.py
```diff
@@ -891,3 +891,27 @@ def attach_custom_product_subscription(prod_name=None, host_name=None):
...
+    :return boolean: True/False
+    """
+    temp = entities.ProvisioningTemplate().search(
+        query={
+            'per_page': 1000,
+            'search': 'name="{}"'.format(name),
+            'organization_id': org_id
+        })[0].read()
+    if old in temp:
+        temp.template = temp.template.replace(old, new)
```
jyejare: Provide a parameter to choose either 1 occurrence or all.

The current behavior replaces more than one matching lines with new.

robottelo/api/utils.py
vijay8451: Added :)

robottelo/api/utils.py
vijay8451: yeah  I forgot about that ... added now.

vijay8451:
> Ack pending comments.

Added , please review.

robottelo/api/utils.py
```diff
@@ -891,3 +891,27 @@ def attach_custom_product_subscription(prod_name=None, host_name=None):
...
+    :param str new: replace content
+    :param int org_id: organization ID
+
+    :return boolean: True/False
+    """
+    temp = entities.ProvisioningTemplate().search(
+        query={
+            'per_page': 1000,
+            'search': 'name="{}"'.format(name),
+            'organization_id': org_id
```
jyejare: I dont think that org id is required. OR we shouldn't make it mandatory.

vijay8451: It was added to restrict the search with specific org only because in provisioning scenario we create a new org and templates are aligned to that only .

If you think it could make some false call .. I am happy to remove it .

jyejare: Provisioning Templates are global and they are not bound to any perticular org.

Hence, change in provisinoing template is univesal and not bound to any perticular taxonomy. If you change in template from org, the other org will show same change.

Hence, it doesnt make sense for org parameter here.

To reinitialize things, resetting the 'old' text in teardown is mandatory.

vijay8451: True Provisioning Templates are global but they only become part of an organization if the get assigned to that org else they never get visible to it .
i.e.
snap123

in above the `Kickstart default` template never be available to org `test123` .
hence a org requires for a provisioning template to become visible into that.

Now coming to the point: by default newly created org does have the all templates assigned with it.
I understand removing it  does not impact anything in functionality thus removing it now ..

vijay8451: Removed in latest snap .. I hope good to merge this PR now :)

PR #7264, opened 2019-08-09
GCE Host Provision E2E

jyejare:
New:

- Fixture for GCE Domain
- Fixture for GCE Hostgroup
- Fixture for wrappenapi GCE Client
- Fixture Compute Resource with Image (from UI as API / CLI support is not yet available for GCE)
- Fixture for GCE Download cert in Satellite for creating GCE CR and in Test running machine for wrappenAPI GCE client end point validation
- GCE Image UUID in constants for creating an image in GCE CR

Updates:

- Updates to the name of most of the existing fixtures created for libvirt, now they are generalized
- GCE CR test is modified to select the zone provided in the properties file, earlier it was choosing the default one
- Update to ServerFileDownloader class for binary permission

Fixes:
- Fixes to https://github.com/SatelliteQE/robottelo/pull/7261

### DIFF
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -34,6 +34,7 @@ from robottelo.api.utils import (
     create_role_permissions,
     promote,
     publish_puppet_module,
+    update_provisioning_template
 )
 from robottelo.cli.factory import (
     make_content_view,
@@ -73,7 +74,9 @@ from robottelo.decorators import (
     skip_if_not_set,
     upgrade,
 )
+from robottelo.helpers import download_server_file
 from robottelo.ui.utils import create_fake_host
+from wrapanapi import GoogleCloudSystem
 
 
 def _get_set_from_list_of_dict(value):
@@ -85,6 +88,21 @@ def _get_set_from_list_of_dict(value):
         tuple(sorted(list(global_param.items()), key=lambda t: t[0]))
         for global_param in value
     }
+
+
+def skip_yum_update_during_provisioning(template=None, reverse=False):
+    """Hides the yum update command with echo text
+
+    :param str template: The template name where the yum update will be hidden
+    :param bool reverse: Reverses the echo text to yum update
+    :return: Boolean True on success else exception
+    """
+    old = 'yum -t -y update'
+    new = 'echo "Yum update skipped for faster automation testing"'
+    if not reverse:
+        return update_provisioning_template(name=template, old=old, new=new)
+    else:
+        return update_provisioning_template(name=template, old=new, new=old)
 
 
 @pytest.fixture
@@ -179,20 +197,21 @@ def default_architecture():
 
 
 @pytest.fixture(scope='module')
-def module_libvirt_environment(module_org, module_loc):
+def module_environment(module_org, module_loc):
     # Create new puppet environment
     return entities.Environment(
         organization=[module_org], location=[module_loc]).create()
 
 
 @pytest.fixture(scope='module')
-def module_libvirt_os(
+def module_os(
         default_architecture, default_partition_table, module_org, module_loc):
     # Get the OS ID
-    os = entities.OperatingSystem().search(query={
-        u'search': u'name="RedHat" AND (major="{0}" OR major="{1}")'.format(
-            RHEL_6_MAJOR_VERSION, RHEL_7_MAJOR_VERSION)}
-    )[0].read()
+    os = entities.OperatingSystem().search(
+        query={u'search': u'name="RedHat" AND major="7"'}
+    ) or entities.OperatingSystem().search(
+        query={u'search': u'name="RedHat" AND major="6"'})
+    os = os[0].read()
     # Get the templates and update with OS, Org, Location
     templates = []
     for template_name in [
@@ -223,11 +242,11 @@ def module_libvirt_os(
 
 
 @pytest.fixture(scope='module')
-def os_path(module_libvirt_os):
+def os_path(module_os):
     # Check what OS was found to use correct media
-    if module_libvirt_os.major == str(RHEL_6_MAJOR_VERSION):
+    if module_os.major == str(RHEL_6_MAJOR_VERSION):
         os_distr_url = settings.rhel6_os
-    elif module_libvirt_os.major == str(RHEL_7_MAJOR_VERSION):
+    elif module_os.major == str(RHEL_7_MAJOR_VERSION):
         os_distr_url = settings.rhel7_os
     else:
         raise ValueError('Proposed RHEL version is not supported')
@@ -338,25 +357,25 @@ def module_libvirt_subnet(
 
 
 @pytest.fixture(scope='module')
-def module_libvirt_lce(module_org):
+def module_lce(module_org):
     return entities.LifecycleEnvironment(organization=module_org).create()
 
 
 @pytest.fixture(scope='module')
-def module_libvirt_product(module_org):
+def module_product(module_org):
     return entities.Product(organization=module_org).create()
 
 
 @pytest.fixture(scope='module')
-def module_libvirt_repository(os_path, module_libvirt_product):
+def module_repository(os_path, module_product):
     repo = entities.Repository(
-        product=module_libvirt_product, url=os_path).create()
+        product=module_product, url=os_path).create()
     call_entity_method_with_timeout(entities.Repository(id=repo.id).sync, timeout=3600)
     return repo
 
 
 @pytest.fixture(scope='module')
-def module_libvirt_media(module_org, module_loc, os_path, module_libvirt_os):
+def module_libvirt_media(module_org, module_loc, os_path, module_os):
     media = entities.Media().search(
         query={u'search': u'path="{0}"'.format(os_path)}
     )
@@ -365,7 +384,7 @@ def module_libvirt_media(module_org, module_loc, os_path, module_libvirt_os):
         media = media[0].read()
         media.organization.append(module_org)
         media.location.append(module_loc)
-        media.operatingsystem.append(module_libvirt_os)
+        media.operatingsystem.append(module_os)
         media.os_family = 'Redhat'
         media = media.update(['organization', 'location', 'operatingsystem', 'os_family'])
     else:
@@ -373,7 +392,7 @@ def module_libvirt_media(module_org, module_loc, os_path, module_libvirt_os):
         media = entities.Media(
             organization=[module_org],
             location=[module_loc],
-            operatingsystem=[module_libvirt_os],
+            operatingsystem=[module_os],
             path_=os_path,
             os_family='Redhat',
         ).create()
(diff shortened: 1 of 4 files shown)


jyejare:
This PR is dependent on 
Airgun : https://github.com/SatelliteQE/airgun/pull/388

and 

Nailgun:  https://github.com/SatelliteQE/nailgun/pull/651

tests/foreman/ui/test_host.py
```diff
@@ -338,25 +341,25 @@ def module_libvirt_subnet(
...
     repo = entities.Repository(
-        product=module_libvirt_product, url=os_path).create()
+        product=module_product, url=os_path).create()
     call_entity_method_with_timeout(entities.Repository(id=repo.id).sync, timeout=3600)
     return repo
 
 
 @pytest.fixture(scope='module')
-def module_libvirt_media(module_org, module_loc, os_path, module_libvirt_os):
+def module_libvirt_media(module_org, module_loc, os_path, module_os):
```
omkarkhatavkar: will this name also be changed to module_media? instead module_libvirt_media

tests/foreman/ui/test_host.py
```diff
@@ -1701,3 +1704,190 @@ def test_positive_delete_libvirt(
...
+
+    :CaseLevel: System
+    """
+    name = gen_string('alpha').lower()
+    hostname = u'{0}.{1}'.format(name, class_gce_domain.name)
+    root_pwd = gen_string('alpha', 15)
+    with Session('gce_tests') as session:
+        session.organization.select(org_name=module_org.name)
+        session.location.select(loc_name=module_loc.name)
+        # Provision GCE Host
```
omkarkhatavkar: Does above code require? I suppose the dashboard will open module_org_name and location?

tests/foreman/ui/test_host.py
```diff
@@ -1701,3 +1704,190 @@ def test_positive_delete_libvirt(
...
+        try:
```
vijay8451: Just a quick suggestion to use helper :https://github.com/SatelliteQE/robottelo/pull/7261 in order to save the `yum update` time on client ... 

I'll review this PR bit later :)

tests/foreman/ui/test_host.py
jyejare: As of now this being used only for libvirt and need not to generalized it. Whenever other tests will be added and if they want to generalize it they can.

For GCE that function is not required.

tests/foreman/ui/test_host.py
jyejare: Nop, we have to explicitly navigate to that org and loc.

tests/foreman/ui/test_host.py
jyejare: YUP, will do ! By the way the comments are given on that PR for some improvements.

vijay8451: They are addressed , please check.

tests/foreman/ui/test_host.py
omkarkhatavkar: okay make sense

tests/foreman/ui/test_host.py
```diff
@@ -1701,3 +1704,194 @@ def test_positive_delete_libvirt(
...
+def class_gce_template(gce_client):
+    max_rhel7_template = max(
+        [img.name for img in gce_client.list_templates(
+            True) if str(img.name).startswith('rhel-7')])
+    return gce_client.get_template(max_rhel7_template, project='rhel-cloud').uuid
+
+
+@pytest.fixture
+def class_gce_domain(module_org, module_loc):
+    domain_name = '{zone}.c.{project_id}.internal'.format(
```
vijay8451:
```suggestion
    domain_name = '{}.c.{}.internal'.format(
```

tests/foreman/ui/test_host.py
```diff
@@ -1701,3 +1704,194 @@ def test_positive_delete_libvirt(
...
+                'provider_content.virtual_machine.network': 'default',
+                'operating_system.operating_system': module_os.title,
+                'operating_system.image': 'autogce_img',
+                'operating_system.root_password': root_pwd
+            })
+            wait_for(
+                lambda: session.host.get_details(hostname)[
+                            'properties']['properties_table']['Build'] != 'Pending installation',
+                timeout=600,
+                delay=30,
```
vijay8451: Can we use some API way to check the host Build status ? and once status(!Pending installation) get from API move further ...
as that would help to save the delay time (because on UI this check always takes time to go through entire navigation)

tests/foreman/ui/test_host.py
```diff
@@ -1701,3 +1704,194 @@ def test_positive_delete_libvirt(
...
+                fail_func=session.browser.refresh,
+                silent_failure=True,
+                handle_exception=True,
+            )
+            # 1. Host Creation Assertions
+            # 1.1 UI based Assertions
+            host_info = session.host.get_details(hostname)
+            assert session.host.search(hostname)[0]['Name'] == hostname
+            assert host_info['properties']['properties_table']['Build'] == 'Installed'
```
vijay8451: Can we add assertion here to see `Zone`  `IP`, 'machineType', `Interface` as well , as that would help to make sure there is not empty values around it ?

jyejare:
Latest Test Result with 600s timeout in render_page_safe and hide yum update from provisioning template:
```
$ py.test tests/foreman/ui/test_host.py -k test_positive_gce_provision_end_to_end
2019-08-14 13:22:03 - conftest - DEBUG - Registering custom pytest_configure

collecting 32 items                                                                                                                                                                                                                          2019-08-14 13:22:05 - conftest - DEBUG - BZ deselect is disabled in settings

collected 32 items / 31 deselected / 1 selected                                                                                                                                                                                              

tests/foreman/ui/test_host.py .                                                                                                                                                                                                        [100%]
=========================================================================================== 1 passed, 31 deselected in 1177.73 seconds ===========================================================================================
```

tests/foreman/ui/test_host.py
```diff
@@ -1701,3 +1704,194 @@ def test_positive_delete_libvirt(
...
+def gce_client(download_cert):
+    return GoogleCloudSystem(
+        project=settings.gce.project_id,
+        zone=settings.gce.zone,
+        file_path=download_cert,
+        file_type='json')
+
+
+@pytest.fixture
+def class_gce_template(gce_client):
```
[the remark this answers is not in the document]
jyejare: Actually, this aligns with existing fixtures in this file which starts with ```module_``` which suggests that it can be used at module level.

Here the class says that it just aligns with particular CR. I will remove class itself from all and will start with ```gce_```.

tests/foreman/ui/test_host.py
jyejare: 1. It is good to check 'Installed' status from UI as its UI test.
2. GCE doesnt supports API endpoint for now.

vijay8451: You are right `Installed` should check from UI and same you could check in assertion under [1] ..

My thought here is  under wait_for we are checking each time through UI only(it is too slower then API) ... so same can check from API (by search the host and check status ... no GCE API support needed for that) ..

[1 ]https://github.com/SatelliteQE/robottelo/pull/7264/files#diff-339c21393e2e46d258dccc49edb476dbR1885

tests/foreman/ui/test_host.py
jyejare: I am already testing IP, Zone and machineType already and that too directly using GCE APIs they are below this assert statement.

What do you want to check in interface, I could only see these fields available in GCE Host:
Screenshot from 2019-08-14 20-04-36

vijay8451: the idea is to check the values around these field are as expected on Satellite ..

As you are doing validation of GCE directly hence does not shows the validation at Satellite as well for same VM/host.

About interface please ignore as nothing much exists .

jyejare: > As you are doing validation of GCE directly(that is +1 as validating there too) hence does not show the validation at Satellite as well for same VM/host.

didnt get this, but Do you sill want me to test things from UI ?

jyejare:
@vijay8451, Post 'Installation' changes check from API,

Results:
```
============ 1 passed, 31 deselected in 1105.07 seconds =================
```

tests/foreman/ui/test_host.py
vijay8451: The thought it :

If you are checking `IP, Zone and machineType` using GCE API then can you please check the same at **Satellite end as well** ?
Why I am saying to check it because it may happen that sometime related values(IP, Zone and machineType) are not populating at Satellite end but from GCE IP you able to see it.

I believe `host_info` would able to provide these details and you only need to add the related **assert** .

tests/foreman/ui/test_host.py
```diff
@@ -1701,3 +1720,196 @@ def test_positive_delete_libvirt(
...
+                'operating_system.operating_system': module_os.title,
+                'operating_system.image': 'autogce_img',
+                'operating_system.root_password': root_pwd
+            })
+            wait_for(
+                lambda: entities.Host().search(
+                    query={'search': 'name={}'.format(
+                        hostname)})[0].build_status_label != 'Pending installation',
+                timeout=600,
+                delay=30,
```
vijay8451:
```suggestion
                delay=10,
```
Just a suggestion to change the `delay` as that would help to save atleast sometime in test.

tests/foreman/ui/test_host.py
jyejare: @vijay8451 Only IP value is accessible in host_info. For getting Zone and machineType I have to switch to the VM tab in the host page for which support/PR needs to be airgun.

Currently I am trying so, but unfortunately, the VM tab doesn't have an ID or something unique to add here: https://github.com/SatelliteQE/airgun/blob/master/airgun/views/host.py#L491

I am looking into it.

tests/foreman/ui/test_host.py
jyejare: Dont you think its so frequent? May be 20 seconds ?

vijay8451: 15-20 seconds gap should be fine. please add.

tests/foreman/ui/test_host.py
vijay8451: @jyejare I have another thought , let's validate IP here only on UI...
and keep  `Zone and machineType` validation through API(whenever GCE API support added to Satellite) ... what do you think ?

jyejare: @vijay8451 , Ok then! I will straightaway update the PR, that would be quick.

jyejare:
@vijay8451 , All your comments addressed / discussed.

vijay8451:
Thanks @jyejare to addressed all comments ...

Ack .. pending https://github.com/SatelliteQE/nailgun/pull/651

PR #7271, opened 2019-08-13
updating images in distro in robotello properties sample

tstrych:


robottelo.properties.sample
```diff
@@ -187,8 +187,8 @@ ssh_key=
 
 # For tests that uses the images for content-host testcases.
 # [distro]
-# image_el6=rhel68
-# image_el7=rhel73
+# image_el6=rhel610
```
vijay8451: Not requires to update this one as RHEL6.10 is latests one ..

robottelo.properties.sample
```diff
@@ -187,8 +187,8 @@ ssh_key=
...
+# image_el7=rhel76
```
vijay8451:
```suggestion
# image_el7=rhel77
```
RHEL7.7 is latest GA here ..

robottelo.properties.sample
tstrych: @vijay8451 Can you please explain to me, why we do not need to change this ? I am probably missing the idea which version should be in distro section.

vijay8451: ahh .. you can ignore ... I missed the version change rhel610 :)

PR #7287, opened 2019-08-15
Simple CLI Test For logging

swadeley:
Hello

 one test (Candlepin) is still unfinished, the test for manifest load, but I wanted more time to get feedback.

Edit: Candlepin test now works with own org.

Edit: Done:  Candlepin test working just with manifest upload event.

### DIFF
--- a/tests/foreman/cli/test_logging.py
+++ b/tests/foreman/cli/test_logging.py
@@ -1,0 +1,266 @@
+"""CLI tests for logging.
+
+:Requirement: Logging
+
+:CaseAutomation: Automated
+
+:CaseLevel: Acceptance
+
+:CaseComponent: logging
+
+:TestType: Functional
+
+:CaseImportance: Medium
+
+:Upstream: No
+"""
+import re
+from fauxfactory import gen_string
+from robottelo import (
+    manifests,
+    ssh,
+)
+from robottelo.cli.factory import (
+    make_org,
+    make_product_wait,
+    make_repository,
+)
+from robottelo.cli.subscription import Subscription
+from robottelo.ssh import upload_file
+from robottelo.test import CLITestCase
+from robottelo.decorators import tier4
+
+
+def line_count(file, connection=None):
+    """Get number of lines in a file."""
+    connection = connection or ssh.get_connection()
+    result = connection.run('wc -l < {0}'.format(file),
+                            output_format='plain')
+    count = result.stdout.strip('\n')
+    return count
+
+
+def cut_lines(start_line, end_line, source_file, out_file, connection=None):
+    """Given start and end line numbers, cut lines from source file
+    and put them in out file."""
+    connection = connection or ssh.get_connection()
+    result = connection.run('sed -n "{0},{1} p" {2} < {2} > {3}'.format(
+                start_line,
+                end_line,
+                source_file,
+                out_file))
+    return result
+
+
+class SimpleLoggingTestCase(CLITestCase):
+    """Test class for default logging to files."""
+
+    org = None
+    product = None
+
+    @classmethod
+    def setUpClass(cls):
+        """Tests for logging to files"""
+        super(SimpleLoggingTestCase, cls).setUpClass()
+
+        # need own org for the manifest refresh test
+
+        cls.org = make_org(cached=True)
+        cls.product = make_product_wait(
+            {u'organization-id': cls.org['id']},
+            )
+
+    def _make_repository(self, options=None):
+        """Makes a new repository and asserts its success"""
+        if options is None:
+            options = {}
+
+        if not options.get('product-id'):
+            options[u'product-id'] = self.product['id']
+
+        return make_repository(options)
+
+    @tier4
+    def test_positive_logging_from_foreman_core(self):
+        """Check that GET command to Hosts API is logged and has request ID.
+
+        :id: 0785260d-cb81-4351-a7cb-d7841335e2de
+
+        :expectedresults: line of log with GET has request ID
+
+        :CaseImportance: Medium
+        """
+
+        GET_line_found = False
+        source_log = '/var/log/foreman/production.log'
+        test_logfile = '/var/tmp/logfile_from_foreman_core'
+        with ssh.get_connection() as connection:
+            # get the number of lines in the source log before the test
+            line_count_start = line_count(source_log, connection)
+            # hammer command for this test
+            result = connection.run('hammer host list')
+            self.assertEqual(result.return_code, 0, "BASH command error?")
+            # get the number of lines in the source log after the test
+            line_count_end = line_count(source_log, connection)
+            # get the log lines of interest, put them in test_logfile
+            cut_lines(line_count_start, line_count_end, source_log, test_logfile, connection)
+        # use same location on remote and local for log file extract
+        ssh.download_file(test_logfile)
+        # search the log file extract for the line with GET to host API
+        with open(test_logfile, "r") as logfile:
+            for line in logfile:
+                if re.search(r'Started GET \"\/api/hosts\?page=1', line):
+                    self.logger.info('Found the line with GET to hosts API')
+                    GET_line_found = True
+                    # Confirm the request ID was logged in the line with GET
+                    match = re.search(r'\[I\|app\|\w{8}\]', line)
+                    assert match, "Request ID not found"
+                    self.logger.info("Request ID found for logging from foreman core")
+                    break
+        assert GET_line_found, "The GET command to list hosts was not found in logs."
+
+    @tier4
+    def test_positive_logging_from_foreman_proxy(self):
+        """Check PUT to Smart Proxy API to refresh the features is logged and has request ID.
+
+        :id: 0ecd8406-6cf1-4520-b8b6-8a164a1e60c2
+
+        :expectedresults: line of log with PUT has request ID
+
+        :CaseImportance: Medium
+        """
+
+        PUT_line_found = False
+        request_id = None
+        source_log_1 = '/var/log/foreman/production.log'
+        test_logfile_1 = '/var/tmp/logfile_1_from_proxy'
+        source_log_2 = '/var/log/foreman-proxy/proxy.log'
+        test_logfile_2 = '/var/tmp/logfile_2_from_proxy'
+        with ssh.get_connection() as connection:
+            # get the number of lines in the source logs before the test
+            line_count_start_1 = line_count(source_log_1, connection)
+            line_count_start_2 = line_count(source_log_2, connection)
+            # hammer command for this test
+            result = connection.run('hammer proxy refresh-features --id 1')
+            self.assertEqual(result.return_code, 0, "BASH command error?")
+            # get the number of lines in the source logs after the test
+            line_count_end_1 = line_count(source_log_1, connection)
+            line_count_end_2 = line_count(source_log_2, connection)
+            # get the log lines of interest, put them in test_logfile_1
+            cut_lines(line_count_start_1, line_count_end_1, source_log_1, test_logfile_1,
+                      connection)
+            # get the log lines of interest, put them in test_logfile_2
+            cut_lines(line_count_start_2, line_count_end_2, source_log_2, test_logfile_2,
+                      connection)
+        # use same location on remote and local for log file extract
+        ssh.download_file(test_logfile_1)
+        # use same location on remote and local for log file extract
+        ssh.download_file(test_logfile_2)
+        # search the log file extract for the line with PUT to host API
+        with open(test_logfile_1, "r") as logfile:
+            for line in logfile:
+                if re.search(r'Started PUT \"\/api\/smart_proxies\/1\/refresh', line):
+                    self.logger.info('Found the line with PUT to foreman proxy API')
+                    PUT_line_found = True
+                    # Confirm the request ID was logged in the line with PUT
+                    match = re.search(r'\[I\|app\|\w{8}\]', line)
+                    assert match, "Request ID not found"
+                    self.logger.info("Request ID found for logging from foreman proxy")
+                    p = re.compile(r"\w{8}")
+                    result = p.search(line)
+                    request_id = result.group(0)
+                    break
+        assert PUT_line_found, "The PUT command to refresh proxies was not found in logs."
+        # search the local copy of proxy.log file for the same request ID
+        with open(test_logfile_2, "r") as logfile:
+            for line in logfile:
+                # Confirm request ID was logged in proxy.log
+                match = line.find(request_id)
+                assert match, "Request ID not found in proxy.log"
+                self.logger.info("Request ID also found in proxy.log")
+                break
+
+    @tier4
+    def test_positive_logging_from_candlepin(self):
+        """Check logging after manifest upload.
+
+        :id: 8c06e501-52d7-4baf-903e-7de9caffb066
+
+        :expectedresults: line of logs with POST has request ID
+
+        :CaseImportance: Medium
+        """
+
+        POST_line_found = False
+        source_log = '/var/log/candlepin/candlepin.log'
+        test_logfile = '/var/tmp/logfile_from_candlepin'
+        # regex for a version 4 UUID (8-4-4-12 format)
+        regex = r"\b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b"
+        with ssh.get_connection() as connection:
+            # get the number of lines in the source log before the test
+            line_count_start = line_count(source_log, connection)
+            # command for this test
+            with manifests.clone() as manifest:
+                upload_file(manifest.content, manifest.filename)
+                Subscription.upload({
+                    'file': manifest.filename,
+                    'organization-id': self.org['id'],
+                    })
+            # get the number of lines in the source log after the test
+            line_count_end = line_count(source_log, connection)
+            # get the log lines of interest, put them in test_logfile
+            cut_lines(line_count_start, line_count_end, source_log, test_logfile, connection)
+        # use same location on remote and local for log file extract
+        ssh.download_file(test_logfile)
+        # search the log file extract for the line with POST to candlepin API
+        with open(test_logfile, "r") as logfile:
+            for line in logfile:
+                if re.search(r'verb=POST, uri=/candlepin/owners/{0}', line
+                             .format(self.org['name'])):
+                    self.logger.info('Found the line with POST to candlepin API')
+                    POST_line_found = True
+                    # Confirm the request ID was logged in the line with POST
+                    match = re.search(regex, line)
+                    assert match, "Request ID not found"
+                    self.logger.info("Request ID found for logging from candlepin")
+                    break
+        assert POST_line_found, "The POST command to candlepin was not found in logs."
+
+    @tier4
+    def test_positive_logging_from_dynflow(self):
+        """Check POST to repositories API is logged while enabling a repo \
+            and it has the request ID.
+
+        :id: 2d1a5f64-0b1c-4f95-ad20-881134717c4c
+
+        :expectedresults: line of log with POST has request ID
+
+        :CaseImportance: Medium
+        """
+
+        POST_line_found = False
+        source_log = '/var/log/foreman/production.log'
+        test_logfile = '/var/tmp/logfile_dynflow'
+        with ssh.get_connection() as connection:
+            # get the number of lines in the source log before the test
+            line_count_start = line_count(source_log, connection)
+            # command for this test
+            new_repo = self._make_repository({u'name': gen_string('alpha')})
+            self.logger.info('Created Repo {0} for dynflow log test'.format(new_repo[u'name']))
+            # get the number of lines in the source log after the test
+            line_count_end = line_count(source_log, connection)
+            # get the log lines of interest, put them in test_logfile
+            cut_lines(line_count_start, line_count_end, source_log, test_logfile, connection)
+        # use same location on remote and local for log file extract
+        ssh.download_file(test_logfile)
+        # search the log file extract for the line with POST to to repositories API
+        with open(test_logfile, "r") as logfile:
+            for line in logfile:
+                if re.search(r'Started POST \"/katello\/api\/repositories', line):
+                    self.logger.info('Found the line with POST to repositories API.')
+                    POST_line_found = True
+                    # Confirm the request ID was logged in the line with POST
+                    match = re.search(r'\[I\|app\|\w{8}\]', line)
+                    assert match, "Request ID not found"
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,189 @@
...
+    product = None
+
+    def setUp(self):
+        """Tests for logging to files"""
+        super(SimpleLoggingTestCase, self).setUp()
+
+        # need own org for the manifest refresh test
+
+        if SimpleLoggingTestCase.org is None:
+            SimpleLoggingTestCase.org = make_org(cached=True)
```
mirzal: Perhaps you want [`setUpClass`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass), which is called once per class (and not before each test in class)? This way you will be able to drop these conditionals.

Also, why `SimpleLoggingTestCase.org` and not `self.org`? You will have single instance of this class anyway, so it's not like you need to share this data between instances.

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,189 @@
...
+            SimpleLoggingTestCase.product = make_product_wait(
+                {u'organization-id': SimpleLoggingTestCase.org['id']},
+            )
+
+    def _make_repository(self, options=None):
+        """Makes a new repository and asserts its success"""
+        if options is None:
+            options = {}
+
+        if options.get('product-id') is None:
```
mirzal:
```suggestion
        if not options.get('product-id'):
```

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,189 @@
...
+        self.logger.info("Testing test_positive_logging_from_foreman_core")
+        with ssh.get_connection() as connection:
+            result = connection.run('hammer host list')
+            self.assertEqual(result.return_code, 0)
+            # extract last ten lines from log
+            result = connection.run('tail -20 /var/log/foreman/production.log > \
+                        /var/tmp/logfile_from_foreman_core')
+            self.assertEqual(result.return_code, 0)
+        # use same location on remote and local for log file extract
+        logfile_location = '/var/tmp/logfile_from_foreman_core'
```
mirzal: Move that variable above and use it in `connection.run()`:
```python
            result = connection.run(('tail -20 /var/log/foreman/production.log > '
                        '{}'.format(logfile_location)))
```

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,189 @@
...
+                if re.search(r'Started GET \"\/api/hosts\?page=1', line):
+                    self.logger.info('Found the line with GET to hosts API')
+                    GET_line_found = 1
+                    # Confirm the request ID was logged in the line with GET
+                    match = re.search(r'\[I\|app\|\w{8}\]', line)
+                    if match:
+                        self.logger.info("Request ID found for logging from foreman core")
+                    else:
+                        assert match == 1, "Request ID not found"
+        assert GET_line_found == 1, "The GET command to list hosts was not found in logs."
```
mirzal: Perhaps it's all edge cases, but some ideas when that assertion will produce incorrect result:

* another test generates a lot of data between us calling "GET" and getting log for further processing
* Satellite throws exception on receiving GET and saves it in log file
* Satellite starts logging much more data for each request and 20 lines are not enough
* Satellite never receives our GET, but some other test asked for list of hosts recently and their request is still in log

These can be mitigated by sending unique request (like asking for host with specific name, one that we created a while earlier) and reading logs from last X minutes, no matter how many lines there are.

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,189 @@
...
+            for line in logfile:
+                if re.search(r'Started GET \"\/api/hosts\?page=1', line):
+                    self.logger.info('Found the line with GET to hosts API')
+                    GET_line_found = 1
+                    # Confirm the request ID was logged in the line with GET
+                    match = re.search(r'\[I\|app\|\w{8}\]', line)
+                    if match:
+                        self.logger.info("Request ID found for logging from foreman core")
+                    else:
+                        assert match == 1, "Request ID not found"
```
mirzal: I find this code hard to understand. Since `re.search()` will return None if match was not found, in else branch `match` will be None, meaning this assertion can never succeed. I guess the goal is to fail test with specific message? Then it's better to use [`self.fail()`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.fail).

If you want to fail test if `re.search()` doesn't find anything, then perhaps it's better to refactor it a little:

```python
match = re.search(r'\[I\|app\|\w{8}\]', line)
assert match, "Request ID not found"
# `match` is guaranteed to be proper Match object
self.logger.info("Request ID found for logging from foreman core")
```

Also, you probably want to call `return` or `break` after finding the match, as there is no point in analyzing subsequent lines.

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,189 @@
...
+        assert PUT_line_found == 1, "The PUT command to refresh proxies was not found in logs."
+
+    def test_positive_logging_from_candlepin(self):
+        """Check logging after load or manifest refresh."""
+        GET_line_found = 0
+        POST_line_found = 0
+        with ssh.get_connection() as connection:
+            # get the number of lines in the log before the test
+            line_count_1 = connection.run('wc -l < \
+                                    /var/log/foreman/production.log'.format(),
```
mirzal: This path is used three times in this test, which means you should think about putting it in variable.

tests/foreman/cli/test_logging.py
swadeley: Yes, thank you.

tests/foreman/cli/test_logging.py
swadeley: OK, fixed, thank you.

tests/foreman/cli/test_logging.py
swadeley: I can use it higher up too.

tests/foreman/cli/test_logging.py
swadeley: 1] For the more verbose candlepin logs @JacobCallahan suggested the use of sed to cut the lines from the log that were written during the test. I have now implemented that method for the candlepin log test and I could use that method in the other tests if required.

2] The important thing we are looking for is the ID in the first line of an event. This was fixed recently and is essential for following logs using logging software. The ID was present from 2nd line only before, which made it very difficult to follow logs in busy system.

3] I can use the same method as for candlepin test if required.

4] I think we do not care about that. See point 2.

Thank you

mirzal: > 2] The important thing we are looking for is the ID in the first line of an event. This was fixed recently and is essential for following logs using logging software. The ID was present from 2nd line only before, which made it very difficult to follow logs in busy system.

Yes, but you are processing only last 20 lines of `/var/log/foreman/production.log`. And I mean there is a very slight chance these 20 lines will not contain first line of an event. You need something like this sed cutting to ensure you obtain all lines you need, regardless of how many that is.

I don't really require you to implement these changes, this is more food for thought.

tests/foreman/cli/test_logging.py
swadeley: Thank you, I have now fixed two sections like that as you suggested,

tests/foreman/cli/test_logging.py
swadeley: ACK, now that I have that sed code working I can use everywhere rather than tail. Will come back to this.

tests/foreman/cli/test_logging.py
swadeley: Hello

How do I make use of the cls org in the candlepin test? Something like:

    -            result = connection.run('hammer subscription refresh-manifest --organization-id 1')
    +            result = connection.run('hammer subscription refresh-manifest --organization-id {0}'
    +                                    .format(cls.org.id))

or  cls.org['id']

mirzal: It's in the same class, right? Reference to organization should be available under `self.org`. `make_org` returns dict, so you want `self.org['id']`.

If it did not work, let me know what error did you encounter.

tests/foreman/cli/test_logging.py
swadeley: I have updated all tests to use variables for source and test log files

tests/foreman/cli/test_logging.py
swadeley: I have updated all tests to use sed method same as the candlepin test

tests/foreman/cli/test_logging.py
swadeley: This works, thanks:

     +result = connection.run('hammer subscription refresh-manifest --organization-id {0}'
     +                                    .format(self.org['id']))


However this raises another problem. The clone manifest function changes the consumer UUID so when you try to refresh it you get an error. Options that I can see: Just use default org (that might make the test disruptive if not destructive); download and use another manifest just for this test.

Thank you

swadeley: I will try to get logs from the upload event, that would solve the problem.

swadeley:
Hello

I have changed the candlepin test to use manifest upload event (because refreshing the cloned manifest gives errors)

Test coming soon.

swadeley:
robottelo]$ git status
On branch issue#7286
Your branch is up-to-date with 'origin/issue#7286'.

robottelo]$ pytest  --show-capture=no tests/foreman/cli/test_logging.py
2019-09-20 14:17:35 - conftest - DEBUG - Registering custom pytest_configure

======================= test session starts =======================================
platform linux -- Python 3.6.6, pytest-4.6.3, py-1.8.0, pluggy-0.12.0
rootdir: /sat6/robottelo
plugins: services-1.3.1, forked-1.0.2, xdist-1.29.0, cov-2.7.1, mock-1.10.4
collecting ... 2019-09-20 14:17:36 - conftest - DEBUG - BZ deselect is disabled in settings

collected 4 items                                                                                                                                            

tests/foreman/cli/test_logging.py ....                                                                        [100%]

====================== 4 passed, 2 warnings in 164.70 seconds ======================

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,278 @@
...
+from robottelo.ssh import upload_file
+from robottelo import manifests
+from robottelo.test import CLITestCase
+from robottelo.cli.factory import (
+    make_org,
+    make_product_wait,
+    make_repository,
+)
+from robottelo.cli.subscription import Subscription
+log = logging.getLogger(__name__)
```
mirzal: Am I blind, or you don't actually use this?
`self.logger` is from https://github.com/SatelliteQE/robottelo/blob/master/robottelo/test.py#L178

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,278 @@
...
+        :expectedresults: line of log with GET has request ID
+
+        :CaseImportance: Medium
+        """
+
+        GET_line_found = 0
+        source_log = '/var/log/foreman/production.log'
+        test_logfile = '/var/tmp/logfile_from_foreman_core'
+        # log the start of the test in case of problems with slow systems and time zones
+        self.logger.info("Testing test_positive_logging_from_foreman_core")
```
mirzal: Doesn't pytest do that for us?
It's definitely not a common pattern in our tests.

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,278 @@
...
+        with ssh.get_connection() as connection:
+            # get the number of lines in the log before the test
+            line_count_1 = connection.run('wc -l < \
+                                    {0}'.format(source_log),
```
mirzal:
```python
            line_count_1 = connection.run('wc -l < {0}'.format(source_log),
                                          output_format='plain')
```
Line break here is not needed at all, and backslashes are generally discouraged by PEP8.

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,278 @@
...
+        """Check that GET command to Hosts API is logged and has request ID.
+
+        :id: 0785260d-cb81-4351-a7cb-d7841335e2de
+
+        :expectedresults: line of log with GET has request ID
+
+        :CaseImportance: Medium
+        """
+
+        GET_line_found = 0
```
mirzal: Can you make these boolean?

```python
GET_line_found = False
....
# in for loop:
GET_line_found = True
....
# after for loop:
assert GET_line_found, "failure message"  # or "assert GET_line_found == True", if you want to be very explicit
```

tests/foreman/cli/test_logging.py
swadeley: OK, removed. Thank you.

tests/foreman/cli/test_logging.py
swadeley: OK, I have removed logging lines from the start of the tests.

tests/foreman/cli/test_logging.py
swadeley: Removed those plus some other unnecessary lines breaks after sed commands

tests/foreman/cli/test_logging.py
swadeley: flake8 does not like this form "assert GET_line_found == True"
tests/foreman/cli/test_logging.py:104:31: E712 comparison to True should be 'if cond is True:' or 'if cond:'

But this format works: "assert GET_line_found,"

All changed.
Thank you

swadeley:
robottelo]$ git status
On branch issue#7286
Your branch is up-to-date with 'origin/issue#7286'.

robottelo]$ pytest --show-capture=no tests/foreman/cli/test_logging.py 
2019-09-23 08:05:57 - conftest - DEBUG - Registering custom pytest_configure

====================== test session starts ========================================
platform linux -- Python 3.6.6, pytest-4.6.3, py-1.8.0, pluggy-0.12.0
rootdir: /sat6/robottelo
plugins: services-1.3.1, forked-1.0.2, xdist-1.29.0, cov-2.7.1, mock-1.10.4
collecting ... 2019-09-23 08:05:57 - conftest - DEBUG - BZ deselect is disabled in settings

collected 4 items                                                                                                                                            

tests/foreman/cli/test_logging.py ....                                                                                                                 [100%]

====================== 4 passed, 2 warnings in 162.73 seconds =====================

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,251 @@
...
+:CaseImportance: Medium
+
+:Upstream: No
+"""
+import re
+from fauxfactory import gen_string
+from robottelo import ssh
+from robottelo.ssh import upload_file
+from robottelo import manifests
+from robottelo.test import CLITestCase
```
omkarkhatavkar: Import should be in order.

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,251 @@
...
+
+        :id: 0785260d-cb81-4351-a7cb-d7841335e2de
+
+        :expectedresults: line of log with GET has request ID
+
+        :CaseImportance: Medium
+        """
+
+        GET_line_found = False
+        source_log = '/var/log/foreman/production.log'
```
omkarkhatavkar: source_log is used multiple times, you can either have in a class or module level or keep in constant.py.

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,251 @@
...
+        test_logfile = '/var/tmp/logfile_from_foreman_core'
+        with ssh.get_connection() as connection:
+            # get the number of lines in the log before the test
+            line_count_1 = connection.run('wc -l < {0}'.format(source_log),
+                                          output_format='plain')
+            self.assertEqual(line_count_1.return_code, 0, "BASH command error?")
+            # hammer command for this test
```
omkarkhatavkar: above two lines are common all the test functions, good to be a candidate for reusable function in a class

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,251 @@
...
+
+        :expectedresults: line of logs with POST has request ID
+
+        :CaseImportance: Medium
+        """
+
+        POST_line_found = False
+        source_log = '/var/log/candlepin/candlepin.log'
+        test_logfile = '/var/tmp/logfile_from_candlepin'
+        regex = r"\b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b"
```
omkarkhatavkar: can you please mention a small comment what above regex is doing, it's getting hard for a reviewer to know what that regex is used for.

tests/foreman/cli/test_logging.py
swadeley: Hello @omkarkhatavkar 
isort suggests this:

```
import re

from robottelo import manifests, ssh
from robottelo.cli.factory import make_org, make_product_wait, make_repository
from robottelo.cli.subscription import Subscription
from robottelo.ssh import upload_file
```

or do you prefer:

```
import re
from robottelo import (
    manifests,
    ssh,
)
from robottelo.cli.factory import (
    make_org,
    make_product_wait,
    make_repository,
)
from robottelo.cli.subscription import Subscription
from robottelo.ssh import upload_file

```

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,251 @@
...
+                                          output_format='plain')
+            self.assertEqual(line_count_2.return_code, 0, "BASH command error?")
+            # get the log lines of interest, put them in test_logfile
+            result = connection.run('sed -n "{0},{1} p" {2} < {2} > {3}'.format(
+                line_count_1.stdout.strip('\n'),
+                line_count_2.stdout.strip('\n'),
+                source_log,
+                test_logfile))
+            self.assertEqual(result.return_code, 0, "BASH command error?")
+        # use same location on remote and local for log file extract
```
omkarkhatavkar: above lines are used every test case, looks like good candidate for common reusable function

tests/foreman/cli/test_logging.py
swadeley: But in the test for candlepin its not the same. Also, future tests will have different source.

tests/foreman/cli/test_logging.py
omkarkhatavkar: both look good, any of them is fine

swadeley: OK, using 2nd format as it seems commonly used.
Thank you

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,254 @@
...
+
+        :id: 0ecd8406-6cf1-4520-b8b6-8a164a1e60c2
+
+        :expectedresults: line of log with PUT has request ID
+
+        :CaseImportance: Medium
+        """
+
+        PUT_line_found = False
+        source_log = '/var/log/foreman/production.log'
```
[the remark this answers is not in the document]
swadeley: Hello @vijay8451 , I asked lzap, he said it would be good to verify the UUID I find in the test as it is now is also present in '/var/log/foreman-proxy/proxy.log' because there are two different processes running. I will add an extra step for that. Thanks for spotting that.

swadeley:
robottelo]$ git status
On branch issue#7286
Your branch is up to date with 'origin/issue#7286'.

nothing to commit, working tree clean
robottelo]$ pytest  --show-capture=no tests/foreman/cli/test_logging.py
2019-10-03 15:55:19 - conftest - DEBUG - Registering custom pytest_configure

======================== test session starts =============================
platform linux -- Python 3.7.4, pytest-4.6.3, py-1.8.0, pluggy-0.13.0
rootdir: /sat6/robottelo
plugins: cov-2.7.1, services-1.3.1, mock-1.10.4, forked-1.0.2, xdist-1.29.0
collecting ... 2019-10-03 15:55:19 - conftest - DEBUG - BZ deselect is disabled in settings

collected 4 items                                                                                                

tests/foreman/cli/test_logging.py ....                                                     [100%]

=================== 4 passed, 2 warnings in 174.20 seconds ======================

tests/foreman/cli/test_logging.py
swadeley: @omkarkhatavkar please see new function: "def line_count(self, file, connection=None)"
Line 33

tests/foreman/cli/test_logging.py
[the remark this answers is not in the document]
swadeley: Hello @vijay8451, test updated to check for presence of same Request ID in proxy.log

Thank you

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,272 @@
...
+        """Makes a new repository and asserts its success"""
+        if options is None:
+            options = {}
+
+        if not options.get('product-id'):
+            options[u'product-id'] = self.product['id']
+
+        return make_repository(options)
+
+    def test_positive_logging_from_foreman_core(self):
```
vijay8451: Could you please update the tier for all these tests ...

tests/foreman/cli/test_logging.py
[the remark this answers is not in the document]
vijay8451: @swadeley looks good to me :+1:

tests/foreman/cli/test_logging.py
```diff
@@ -0,0 +1,272 @@
...
+from robottelo.test import CLITestCase
+
+
+def line_count(self, file, connection=None):
+    """Get number of lines in a file."""
+    connection = connection or ssh.get_connection()
+    result = connection.run('wc -l < {0}'.format(file),
+                            output_format='plain')
+    self.assertEqual(result.return_code, 0, "BASH command error?")
+    return result
```
mirekdlugosz: Standalone functions shouldn't take class instances as first argument. Either remove `self` and assertion, or move that function into `SimpleLoggingTestCase` and call it like: `number_of_lines = self.line_count(file, connection)`.

tests/foreman/cli/test_logging.py
swadeley: OK, done. Thank you.

swadeley:
robottelo]$ git status
On branch issue#7286
Your branch is up to date with 'origin/issue#7286'.

nothing to commit, working tree clean
robottelo]$ pytest tests/foreman/cli/test_logging.py 
2019-10-10 15:12:11 - conftest - DEBUG - Registering custom pytest_configure

======================= test session starts =====================================
platform linux -- Python 3.7.4, pytest-4.6.3, py-1.8.0, pluggy-0.13.0
rootdir: /sat6/robottelo
plugins: cov-2.8.1, services-1.3.1, xdist-1.30.0, mock-1.10.4, forked-1.0.2
collecting ... 2019-10-10 15:12:12 - conftest - DEBUG - BZ deselect is disabled in settings

collected 4 items                                                                                                

tests/foreman/cli/test_logging.py ....                                                [100%]

================== 4 passed, 2 warnings in 79.68 seconds ==========================

PR #7296, opened 2019-08-19
failing because of issue bugzilla-1743271

omkarkhatavkar:
failing because of issue https://bugzilla.redhat.com/show_bug.cgi?id=1743271


pondrejk:
@omkarkhatavkar hey, looks good but I'd add the bz number also to the docstring so that we don't forget about it later.

PR #7275, opened 2019-08-13
refactor smart class params ui tests

pondrejk:
some cleanup for smartclass parameters ui tests. This testclass is much similar to smartvariables tests handled in

results:
```
pytest tests/foreman/ui/test_smartclassparameter.py 
2019-08-13 13:21:39 - conftest - DEBUG - Registering custom pytest_configure

================= test session starts ===============================
platform linux -- Python 3.7.4, pytest-4.6.3, py-1.7.0, pluggy-0.12.0
shared_function enabled - OFF - scope:  - storage: file
rootdir: /home/pondrejk/Documents/robottelo
plugins: cov-2.7.1, mock-1.10.4, forked-0.2, services-1.3.1, xdist-1.29.0
collecting ... 2019-08-13 13:21:39 - conftest - DEBUG - BZ deselect is disabled in settings

collected 6 items                                                                                                                                            
tests/foreman/ui/test_smartclassparameter.py .s....    

5 passed, 1 skipped in 1249.89 seconds
```

One skip due to https://bugzilla.redhat.com/show_bug.cgi?id=1734022
Brief summary of changes:

- hidden parameters merged,
- negative tests reduced,
- test_positive_view_yaml_output merged to avoid duplicates test,
- update_with_long_priority_list removed as the related bz is tested in smartvariables class
- parameter merge tests reduced - merge merge overrides are tested in smartvariables class, merge defaults tested within avoid duplicates test

PR #7303, opened 2019-08-20
fixing assertion for test_positive_create_matcher_attribute_priority

omkarkhatavkar:
E           assert "['ap1dniwgdl...65:c6:f8:6f']" == ['ap1dniwgdl',...:65:c6:f8:6f']
E             -"['ap1dniwgdl', 'b6:19:65:c6:f8:6f']"
E             +['ap1dniwgdl', 'b6:19:65:c6:f8:6f']

output received from the browser was in string format, fixing the input in the same format

### DIFF
--- a/tests/foreman/ui/test_smartclassparameter.py
+++ b/tests/foreman/ui/test_smartclassparameter.py
@@ -292,14 +292,14 @@ def test_positive_create_matcher_attribute_priority(
         )
         output = yaml.load(session.host.read_yaml_output(module_host.name))
         output_scp = output['classes'][PM_NAME][sc_param.parameter]
-        assert output_scp == [domain.name, module_host.mac]
+        assert output_scp == str([domain.name, module_host.mac])
 
         # That update operation will change nothing, but submit the same form
         session.sc_parameter.update(
             sc_param.parameter, {'parameter.override': True})
         output = yaml.load(session.host.read_yaml_output(module_host.name))
         output_scp = output['classes'][PM_NAME][sc_param.parameter]
-        assert output_scp == [domain.name, module_host.mac]
+        assert output_scp == str([domain.name, module_host.mac])
 
 
 @tier2


vijay8451:
@omkarkhatavkar related bug: https://bugzilla.redhat.com/show_bug.cgi?id=1734022#attach_1594267

omkarkhatavkar:
> @omkarkhatavkar related bug: https://bugzilla.redhat.com/show_bug.cgi?id=1734022#attach_1594267 is verified in latest snap and marked under reportportal too ..
> 
> I am not sure why this test is failed in recent snap ? is this a new one or regression ?

@vijay8451 This test is related to smart_class_parameter and bug is about smart_class_vartiable.  I guess these are two separate issues.  @pondrejk can you please quickly comment on this?

pondrejk:
@vijay8451  @pondrejk that bug affected both sc_params and variables but manifested differently and is no longer around so it does not relate to this failure

vijay8451:
@pondrejk @omkarkhatavkar I see this test was running fine till last snap and failed in recent snap only ? do you think a regression or new bug here?

omkarkhatavkar:
> d failed in recent snap only ? do you think a regression or new bug here?

This has happened after https://github.com/SatelliteQE/robottelo/pull/7275/files#diff-017434d10ada2d5285efac96b4ed30b8R295 change. If you see the failure the both are showing the same data only formatting is not correct. when selenium fetched data is in string format whereas expected data in list format. As per my knowledge, it should not be regression or bug.

vijay8451:
> > d failed in recent snap only ? do you think a regression or new bug here?
> 
> This has happened after https://github.com/SatelliteQE/robottelo/pull/7275/files#diff-017434d10ada2d5285efac96b4ed30b8R295 change. If you see the failure the both are showing the same data only formatting is not correct. when selenium fetched data is in string format whereas expected data in list format. As per my knowledge, it should not be regression or bug.

you are right on it and i saw that as well...(offcourse not a big deal to change it) but one question coming in my mind is that , how this test was working fine till last snap ? why the data format is changed ?

vijay8451:
I got it here https://github.com/SatelliteQE/robottelo/pull/7275/files

Issue #7285, opened 2019-08-15
reconsidering ui dashboard testclass


pondrejk:
Currently test_dashboard.py works like this: create organization/location context, create some entities or tasks, then check the reaction on dashboard.

This seems rather wasteful given that individual entities are tested in their own testclasses often in a separate org/loc context. So it should be possible to reuse existing setups and just add some assertions to check dashboard functionality.

PR #7311, opened 2019-08-22
ui dashboard tests reorganized

pondrejk:
fixes https://github.com/SatelliteQE/robottelo/issues/7285
Asserts from ui dashboard tests attached to their respective entities to reuse components.

summary:
```
test_positive_latest_warning_error_tasks merged to test_positive_task_status
test_positive_content_view_history moved to pytest tests/foreman/ui/test_contentview.py::test_positive_promote_with_custom_content
test_positive_rendering_after_env_removed moved to pytest tests/foreman/ui/test_contentview.py::test_positive_promote_with_custom_content
test_positive_host_collections moved to tests/foreman/ui/test_hostcollection.py::test_positive_end_to_end
test_positive_current_subscription_totals moved to tests/foreman/ui/test_subscription.py::test_positive_end_to_end
test_positive_content_host_subscription_status move tests/foreman/ui/test_contenthost.py::test_positive_search_by_subscription_status
test_positive_discovered_host moved to tests/foreman/ui/test_discoveredhost.py::test_positive_update_default_taxonomies
test_positive_sync_overview_widget moved to: pytest tests/foreman/ui/test_repository.py::test_positive_sync_custom_repo_yum
test_positive_new_host moved to tests/foreman/ui/test_host.py::test_positive_create
```

tests:
```
pytest tests/foreman/ui/test_host.py::test_positive_create
==================================================================== test session starts ====================================================================

tests/foreman/ui/test_host.py .                                                                                                                       [100%] 

========================================================== 1 passed in 126.64 seconds ===========================================================
```

tests/foreman/ui/test_hostcollection.py::test_positive_end_to_end -- no succesfull result due to https://github.com/SatelliteQE/airgun/issues/392 -- but checked visually the test passes through the dashboard check step

```
pytest tests/foreman/ui/test_discoveredhost.py::test_positive_update_default_taxonomies
===================================================== test session starts ======================================================

collected 1 item

tests/foreman/ui/test_discoveredhost.py .                                                                                [100%]

============================================ 1 passed in 116.20 seconds ============================================
```

```
pytest tests/foreman/ui/test_subscription.py::test_positive_end_to_end

=================================================================== test session starts ===================================================================
collected 1 item

tests/foreman/ui/test_subscription.py .                                                                                                             [100%]

=============================================================== 1 passed in 173.63 seconds ================================================================
```

```
pytest tests/foreman/ui/test_dashboard.py::test_positive_task_status
======================================================== test session starts ========================================================
collected 1 item 
tests/foreman/ui/test_dashboard.py .                                                                                          [100%]

===================================================== 1 passed in 80.62 seconds =====================================================

```
```
pytest tests/foreman/ui/test_contentview.py::test_positive_promote_with_custom_content

======================================================== test session starts ========================================================
collected 1 item

tests/foreman/ui/test_contentview.py .                                                                                        [100%]

==================================================== 1 passed in 181.75 seconds =====================================================
```

### DIFF
--- a/tests/foreman/ui/test_contentview.py
+++ b/tests/foreman/ui/test_contentview.py
@@ -1358,7 +1358,8 @@ def test_positive_publish_version_changes_in_target_env(session, module_org):
 
 @tier2
 def test_positive_promote_with_custom_content(session, module_org):
-    """Attempt to promote a content view containing custom content
+    """Attempt to promote a content view containing custom content,
+        check dashboard
 
     :id: 7c2fd8f0-c83f-4725-8953-9590112fae50
 
@@ -1367,6 +1368,8 @@ def test_positive_promote_with_custom_content(session, module_org):
     :expectedresults: Content view can be promoted
 
     :CaseLevel: Integration
+
+    :BZ: 1361793
 
     :CaseImportance: Critical
     """
@@ -1382,6 +1385,26 @@ def test_positive_promote_with_custom_content(session, module_org):
         assert result['Version'] == VERSION
         result = session.contentview.promote(cv_name, VERSION, lce.name)
         assert 'Promoted to {}'.format(lce.name) in result['Status']
+        # dashboard
+        values = session.dashboard.read('ContentViews')
+        assert cv_name in values[
+            'content_views'][0]['Content View']
+        assert values['content_views'][0][
+            'Task'] == 'Promoted to {0}'.format(lce.name)
+        assert 'Success' in values['content_views'][0]['Status']
+        assert cv_name in values[
+            'content_views'][1]['Content View']
+        assert values['content_views'][1]['Task'] == 'Published new version'
+        assert 'Success' in values['content_views'][1]['Status']
+        values = session.dashboard.search('lifecycle_environment={}'.format(
+            lce.name))
+        assert cv_name in values['ContentViews'][
+            'content_views'][0]['Content View']
+        entities.LifecycleEnvironment(id=lce.id).delete()
+        values = session.dashboard.search('lifecycle_environment={}'.format(
+            lce.name))
+        assert cv_name in values['ContentViews'][
+            'content_views'][0]['Content View']
 
 
 @run_in_one_thread
--- a/tests/foreman/ui/test_dashboard.py
+++ b/tests/foreman/ui/test_dashboard.py
@@ -16,18 +16,14 @@
 """
 from fauxfactory import (
     gen_integer,
-    gen_ipaddr,
-    gen_mac,
 )
 
 from airgun.session import Session
 from nailgun import entities
 from nailgun.entity_mixins import TaskFailedError
 from pytest import raises
-from requests.exceptions import HTTPError
 
-from robottelo import manifests
-from robottelo.api.utils import create_role_permissions, create_sync_custom_repo, promote
+from robottelo.api.utils import create_role_permissions
 from robottelo.config import settings
 from robottelo.constants import (
     DISTRO_RHEL7,
@@ -39,7 +35,6 @@ from robottelo.datafactory import gen_string
 from robottelo.decorators import (
     bz_bug_is_open,
     run_in_one_thread,
-    skip_if_bug_open,
     skip_if_not_set,
     tier2,
     tier3,
@@ -155,35 +150,6 @@ def test_positive_host_configuration_chart(session):
         assert dashboard_values['chart']['No report'] == '100%'
 
 
-@tier2
-def test_positive_new_host(session):
-    """Check if the New Hosts widget is working in the Dashboard UI
-
-    :id: 123fadc6-e3e0-4a49-8bca-2bf672460359
-
-    :Steps:
-
-        1. Create new host in the application
-        2. Navigate to Monitor -> Dashboard
-        3. Review the New Hosts widget
-
-    :expectedresults: New Hosts widget contains information about just created host
-
-    :CaseLevel: Integration
-    """
-    org = entities.Organization().create()
-    loc = entities.Location().create()
-    host = entities.Host(organization=org, location=loc).create()
-    with session:
-        session.organization.select(org_name=org.name)
-        session.location.select(loc_name=loc.name)
-        dashboard_values = session.dashboard.read('NewHosts')['hosts']
-        assert len(dashboard_values) == 1
-        assert dashboard_values[0]['Host'] == host.name
-        assert host.operatingsystem.read().name in dashboard_values[0]['Operating System']
-        assert dashboard_values[0]['Installed'] == '-'
-
-
 @run_in_one_thread
 @tier2
 def test_positive_task_status(session):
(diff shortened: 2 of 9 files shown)


tests/foreman/ui/test_contentview.py
```diff
@@ -1382,6 +1385,27 @@ def test_positive_promote_with_custom_content(session, module_org):
         assert result['Version'] == VERSION
         result = session.contentview.promote(cv_name, VERSION, lce.name)
         assert 'Promoted to {}'.format(lce.name) in result['Status']
+        # dashboard
+        session.organization.select(org_name=module_org.name)
```
omkarkhatavkar: why there is a need to select org again?

tests/foreman/ui/test_dashboard.py
```diff
@@ -226,287 +193,17 @@ def test_positive_task_status(session):
...
+            'LatestFailedTasks': {'name': 'Synchronize'}
         })
-        values = session.task.read('Create')
-        assert values['task']['result'] == 'error'
+        values = session.task.read('Synchronize')
+        assert values['task']['result'] == 'warning'
         assert (
             values['task']['errors'] ==
-            'Validation failed: Name has already been taken'
+            'PLP0000: Importer indicated a failed response'
```
omkarkhatavkar: Sorry, I'm missing something here, where are steps create this error?

tests/foreman/ui/test_dashboard.py
```diff
@@ -226,287 +193,17 @@ def test_positive_task_status(session):
...
-    host = entities.Host(organization=org, location=loc).create()
-    host_collection = entities.HostCollection(
-        host=[host], organization=org).create()
-    with session:
-        session.organization.select(org_name=org.name)
-        session.location.select(loc_name=loc.name)
-        values = session.dashboard.read('HostCollections')
-        assert values['collections'][0]['Name'] == host_collection.name
-        assert values['collections'][0]['Content Hosts'] == '1'
-
```
omkarkhatavkar: What was the reason for removing this test?

tests/foreman/ui/test_dashboard.py
```diff
@@ -226,287 +193,17 @@ def test_positive_task_status(session):
...
-        subscription_values = session.dashboard.read('SubscriptionStatus')['subscriptions']
-        assert subscription_values[0][
-            'Subscription Status'] == 'Active Subscriptions'
-        assert int(subscription_values[0]['Count']) >= 1
-        assert subscription_values[1][
-            'Subscription Status'] == 'Subscriptions Expiring in 120 Days'
-        assert int(subscription_values[1]['Count']) == 0
-        assert subscription_values[2][
-            'Subscription Status'] == 'Recently Expired Subscriptions'
-        assert int(subscription_values[2]['Count']) == 0
```
omkarkhatavkar: What was the reason for removing this test?

tests/foreman/ui/test_dashboard.py
```diff
@@ -226,287 +193,17 @@ def test_positive_task_status(session):
...
-        4. Click each link:
-
-            a. Invalid Subscriptions
-            b. Valid Subscriptions
-
-    :expectedresults: The widget is updated with all details for Valid,
-        and Invalid Subscriptions
-
-    :CaseLevel: System
-    """
```
omkarkhatavkar: What was the reason for removing this test?

tests/foreman/ui/test_dashboard.py
```diff
@@ -226,287 +193,17 @@ def test_positive_task_status(session):
...
-    :Steps:
-
-        1. Create a Discovered Host.
-        2. Navigate Monitor -> Dashboard
-        3. Review the Discovered Host Status.
-
-    :expectedresults: The widget is updated with all details.
-
-    :CaseLevel: Integration
-    """
```
omkarkhatavkar: What was the reason for removing this test?

tests/foreman/ui/test_dashboard.py
```diff
@@ -624,7 +292,7 @@ def test_positive_virtwho_configs_widget(session):
 
     :expectedresults: The widget is updated with all details.
 
-    :CaseLevel: Integration
+    CaseLevel: Integration
     """
```
omkarkhatavkar: Everywhere we use `:CaseLevel:`

tests/foreman/ui/test_dashboard.py
omkarkhatavkar: Ignore above comment

omkarkhatavkar:
@pondrejk I look at the whole PR, Almost all dashboard tests are moved the specific component. I think it is not good practice. Consider an example if Dashboard is not working then all test contains Dashboard code in other entities will fail. This will give noise and wrong interpretation of other entities.

I'm not in favour of this, merging for the same component/entity makes sense but merging across the component/entity is not a good idea. Let's see what others say.

vijay8451:
I agree with @omkarkhatavkar on it that we should not move such dashboard tests under the specific component ...as that may give us the `False` True results around that component.

pondrejk:
Hi guys, I understand your concerns, though I don't think the risk is very big, esp. compared to performance gains we could get from this PR. If dashboard fails the number of affected wouldn't be big, certainly it wouldn't occlude the overall results for the given component (for example 1 of 21 content host tests would be affected.). Plus from investigation it would be clear where the failures are coming from, and we have bz_bug_is_open to deal with these situation, so I'm not too afraid of misinterpretation.

What bothered me about these dashboard tests is that they created quite some number of entities (which is anyways done elsewhere) to do some basic read tests to check for failures that would be by no means blocking for the release. That was my logic for this PR.

jyejare:
Thanks, Peter, for opening this PR.

I would like to highlight some of the points wrt this PR:

1. Dashboard entity is equally important as other entities and it's completely driven by other entities. Any changes to other entities have a reflection on Dashboard entity. So keeping Dashboard assertions along with their entities is a fair idea.
2. Customer heavily uses Dashboard for getting a quick idea about their production environment and hence Dashboard assertions are 'NOT LESS' important. And hence failure in Dashboard for the particular entity should be equally treated as a failure in the entity.
3. As Dashboard depends on other entities, we have to create an entity to check if the Dashboard has that reflection. **By separating Dashboard tests, we were doubling the work of CRUDing** the entities which are being/have been created with regular Robottelo CRUD tests. And this PR would really save the time eliminating this **doubling work** by just adding the assertions of Dashboard at the end of existing tests.
4. Also, we are not compromising with the functionality here by adding Dashboard assertions as if the test fails due to this we can surely conclude the reasons by looking at the logs. (+1 to what peter has said about it above)

So, I will give 'GO' for this change,

sghai:
I liked the comments from @jyejare and reasoning from @pondrejk  because of two reasons:
a) performance gain 
b) entities created elsewhere to reflect the data on dashboard.

tests/foreman/ui/test_contentview.py
pondrejk: you're right it's not necessary

pondrejk:
I updated the PR based on feedback, is there any chance to resolve the discussion before the nearest snap arrives?

PR #7330, opened 2019-09-03
Fixed the failed cases for virtwho-configure testing

eko999:
We have run the test cases for all the hypervisors, and found some cases failed for with specific hypervisor.
1. check the guest_uuid for hyperv, only need the last segment and upper it
2. need to clean all the hosts before we start to run these cases for the next hypervisor, otherwise, there are many redundant subscription items existing to attach.

### DIFF
--- a/tests/foreman/virtwho/test_cli_virtwho.py
+++ b/tests/foreman/virtwho/test_cli_virtwho.py
@@ -89,17 +89,23 @@ class VirtWhoConfigTestCase(CLITestCase):
         self.assertEquals(
             VirtWhoConfig.info({'id': vhd['id']})['general-information']['status'], 'OK')
         hosts = [
-            (hypervisor_name, 'product_id={}'.format(self.vdc_physical)),
-            (guest_name, 'type=STACK_DERIVED')]
+            (hypervisor_name, 'product_id={} and type=NORMAL'.format(self.vdc_physical)),
+            (guest_name, 'product_id={} and type=STACK_DERIVED'.format(self.vdc_physical))]
         for hostname, sku in hosts:
             host = Host.list({'search': hostname})[0]
-            vdc = Subscription.list({
+            subscriptions = Subscription.list({
                 'organization': DEFAULT_ORG,
                 'search': sku,
-            })[0]
+            })
+            vdc_id = subscriptions[0]['id']
+            if sku == 'type=STACK_DERIVED':
+                for item in subscriptions:
+                    if hypervisor_name in item['type']:
+                        vdc_id = item['id']
+                        break
             result = Host.subscription_attach({
                 'host-id': host['id'],
-                'subscription-id': vdc['id']
+                'subscription-id': vdc_id
             })
             self.assertTrue('attached to the host successfully' in '\n'.join(result))
         VirtWhoConfig.delete({'name': name})
@@ -127,17 +133,23 @@ class VirtWhoConfigTestCase(CLITestCase):
         self.assertEquals(
             VirtWhoConfig.info({'id': vhd['id']})['general-information']['status'], 'OK')
         hosts = [
-            (hypervisor_name, 'product_id={}'.format(self.vdc_physical)),
-            (guest_name, 'type=STACK_DERIVED')]
+            (hypervisor_name, 'product_id={} and type=NORMAL'.format(self.vdc_physical)),
+            (guest_name, 'product_id={} and type=STACK_DERIVED'.format(self.vdc_physical))]
         for hostname, sku in hosts:
             host = Host.list({'search': hostname})[0]
-            vdc = Subscription.list({
+            subscriptions = Subscription.list({
                 'organization': DEFAULT_ORG,
                 'search': sku,
-            })[0]
+            })
+            vdc_id = subscriptions[0]['id']
+            if sku == 'type=STACK_DERIVED':
+                for item in subscriptions:
+                    if hypervisor_name in item['type']:
+                        vdc_id = item['id']
+                        break
             result = Host.subscription_attach({
                 'host-id': host['id'],
-                'subscription-id': vdc['id']
+                'subscription-id': vdc_id
             })
             self.assertTrue('attached to the host successfully' in '\n'.join(result))
         VirtWhoConfig.delete({'name': name})
--- a/tests/foreman/virtwho/test_cli_virtwho.py
+++ b/tests/foreman/virtwho/test_cli_virtwho.py
@@ -98,7 +98,7 @@ class VirtWhoConfigTestCase(CLITestCase):
                 'search': sku,
             })
             vdc_id = subscriptions[0]['id']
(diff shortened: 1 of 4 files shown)


tests/foreman/virtwho/utils.py
```diff
@@ -131,6 +133,8 @@ def virtwho_cleanup():
     runcmd("rm -f /var/run/virt-who.pid")
     runcmd("rm -f /var/log/rhsm/rhsm.log")
     runcmd("rm -rf /etc/virt-who.d/*")
+    for host in Host.list():
+        Host.delete({'id': host['id']})
```
mirzal: Just to confirm - these tests are running on separate Satellite, right? There is no risk of this part interfering with our (Sat QE) automation, correct?

Overall, removing all registered hosts doesn't sound like a good idea for me, but if you have full control over that environment and believe this is the best approach, I'm OK with that.

eko999: @mirzal Yes, there is no influence for the SatQE automation, we deploy the satellite by virtwho jenkins and only used to test virtwho-configure cases.

eko999: Although our satellite is a separated server, I still agree with you it's not a good idea to remove all the hosts, imaging that someone wants to use his satellite server to debug these cases, his original hosts will be deleted, if that, it's too bad.

Why I need to delete the host which sent by virt-who:
When I use hammer to list a specific subscription, it will return two or more items, the redundant
items is caused by virt-who have consumed the physical subscription and generate the virtual bonus subscription for guests, such as:
```
# hammer -v -u admin -p admin  --output=csv subscription list --organization="Default Organization" --search="product_id=RH00001"
ID,UUID,Name,Type,Contract,Account,Support,End Date,Quantity,Consumed
33,8ac213ac6cdbced2016cfac546da0d45,"Red Hat Enterprise Linux for Virtual Datacenters, Premium",Guests of virt-who-bootp-73-131-242.rhts.eng.pek2.redhat.com-1,...
32,8ac213ac6cdbced2016cfab1031c0d39,"Red Hat Enterprise Linux for Virtual Datacenters, Premium",Guests of virt-who-win-virt-who-ci-1,...
5,8ac213ac6cdbced2016cdbeb606d0814,"Red Hat Enterprise Linux for Virtual Datacenters, Premium",Physical,...
```
So I found another solution to fix this issue as below:
```
            host = Host.list({'search': hostname})[0]
            items = Subscription.list({
                'organization': DEFAULT_ORG,
                'search': sku,
            })
            vdc_id = items[0]['id']
            if sku == 'type=STACK_DERIVED':
                for item in items:
                    if hypervisor_name in item['type']:
                        vdc_id = item['id']
                        break
            result = Host.subscription_attach({
                'host-id': host['id'],
                'subscription-id': vdc_id
            })
```
When `type=STACK_DERIVED`, it will return two or more subscriptions, I need the item which including `hypervisor_name`.

eko999:
Test result as:
```
# pytest /root/eko/robottelo/tests/foreman/virtwho/test_api_virtwho.py 
2019-09-04 09:38:57 - conftest - DEBUG - Registering custom pytest_configure

platform linux -- Python 3.7.3, pytest-4.6.3, py-1.7.0, pluggy-0.12.0
rootdir: /root/eko/robottelo, inifile: tox.ini
plugins: html-1.22.0, mock-1.10.4, xdist-1.29.0, metadata-1.8.0, cov-2.7.1, forked-1.0.2, services-1.3.1, flake8-1.0.1
collecting ... 2019-09-04 09:38:57 - conftest - DEBUG - BZ deselect is disabled in settings

collected 7 items                                                                                                                                                           

test_api_virtwho.py .......                                                                                                                                           [100%]

======================================================================== 7 passed in 1559.57 seconds ========================================================================
```

tests/foreman/virtwho/test_cli_virtwho.py
```diff
@@ -89,17 +89,23 @@ def test_positive_deploy_configure_by_id(self):
...
-            vdc = Subscription.list({
+            subscriptions = Subscription.list({
                 'organization': DEFAULT_ORG,
                 'search': sku,
-            })[0]
+            })
+            vdc_id = subscriptions[0]['id']
+            if 'type=STACK_DERIVED' in sku:
+                for item in subscriptions:
+                    if hypervisor_name in item['type']:
```
mirzal: Can you confirm, that this condition is correct? That `hypervisor_name` will be in `item['type']`? 

I noticed that in another file, condition is `if hypervisor_name.lower() in item['hypervisor']['name']`.

tests/foreman/virtwho/test_cli_virtwho.py
```diff
@@ -127,17 +133,23 @@ def test_positive_deploy_configure_by_script(self):
...
+                        vdc_id = item['id']
+                        break
```
mirzal: This condition and loop is present four times, and almost identical in all cases. Would it be possible to use helper function instead?

tests/foreman/virtwho/test_cli_virtwho.py
eko999: Yes,  the subscription list result is difference between cli and api.
**to cli, can check the hypervisor_name by type:**
```
    {
        "account": "xxx",
        "consumed": "0",
        "contract": "xxx",
        "end-date": "2020/05/24 03:59:59",
        "id": "39",
        "name": "Red Hat Enterprise Linux for Virtual Datacenters, Premium",
        "quantity": "Unlimited",
        "support": "Premium",
        "type": "Guests of virt-who-win-virt-who-ci-1",
        "uuid": "8ac213ba6cf11645016cfc6db3710d63"
    }
```
**to api**, I have to check hypervisor_name by ['hypervisor']['name'] :
```
    {
        "id": 24,
        "cp_id": "xxx",
        "subscription_id": 7,
        "name": "Red Hat Enterprise Linux for Virtual Datacenters, Premium",
        "start_date": "2019-05-24 04:00:00 UTC",
        "end_date": "2020-05-24 03:59:59 UTC",
        "available": -1,
        "quantity": -1,
        "consumed": 0,
        "account_number": xxx,
        "contract_number": xxx,
        "support_level": "Premium",
        "product_id": "RH00049",
        "sockets": "2",
        "cores": null,
        "ram": null,
        "instance_multiplier": 1,
        "stacking_id": null,
        "multi_entitlement": null,
        "type": "STACK_DERIVED",
        "product_name": "Red Hat Enterprise Linux for Virtual Datacenters, Premium",
        "unmapped_guest": false,
        "virt_only": true,
        "virt_who": false,
        "upstream": true,
        "upstream_pool_id": null,
        "hypervisor": {
            "id": 67,
            "name": "virt-who-win-virt-who-ci-1"
        }
    }
```

tests/foreman/virtwho/test_cli_virtwho.py
eko999: As above, api and cli are difference condition, the helper function will be limited.

PR #7334, opened 2019-09-04
Failure in end to end test for content host fix

latran:
```ui/test_contenthost.py::test_positive_end_to_end``` has been broken for a while.  This is a fix for that failed test case.

### DIFF
--- a/tests/foreman/ui/test_contenthost.py
+++ b/tests/foreman/ui/test_contenthost.py
@@ -163,8 +163,10 @@ def test_positive_end_to_end(session, repos_collection, vm):
         chost = session.contenthost.read(vm.hostname,
                                          widget_names=['details',
                                                        'provisioning_details',
-                                                       'subscriptions',
-                                                       'repository_sets'])
+                                                       'subscriptions'])
+        session.contenthost.update(vm.hostname, {'repository_sets.limit_to_lce': True})
+        ch_reposet = session.contenthost.read(vm.hostname, widget_names=['repository_sets'])
+        chost = {**chost, **ch_reposet}
         # Ensure all content host fields/tabs have appropriate values
         assert chost['details']['name'] == vm.hostname
         assert (
@@ -208,7 +210,7 @@ def test_positive_end_to_end(session, repos_collection, vm):
         packages = session.contenthost.search_package(vm.hostname, FAKE_0_CUSTOM_PACKAGE_NAME)
         assert packages[0]['Installed Package'] == FAKE_0_CUSTOM_PACKAGE
         # Install errata
-        result = session.contenthost.install_errata(vm.hostname, FAKE_2_ERRATA_ID)
+        result = session.errata.install(FAKE_2_ERRATA_ID, vm.hostname)
         assert result['result'] == 'success'
         # Ensure errata installed
         packages = session.contenthost.search_package(vm.hostname, FAKE_2_CUSTOM_PACKAGE_NAME)

### CODE REVIEW

**Comment by latran** [comment]
@mirzal can you also check this out?  It's in regard to limit to lce checkbox.  So I took your suggestion and did option 2 from: https://github.com/SatelliteQE/airgun/pull/396.  However, something is up with the automation and it's failing 
```
assert ( repos_collection.custom_product['name'] in {repo['Repository Name'] for repo in chost['subscriptions']['resources']['assigned']})
```
I'm not understanding why it's not pulling all the necessary information from the subscriptions page now for that assertion.

**Comment by mirzal** [comment]
@latran the problem you are observing is completely unrelated to `limit_to_lce` checkbox we discussed earlier.

The root cause is in [`SatTable._read_all`](https://github.com/SatelliteQE/airgun/blob/master/airgun/widgets.py#L1637). It tries to go through all pages, read the table and combine results into one big list, that is returned. Under the hood, it calls [`widgetastic.widget.table.row`](https://github.com/RedHatQE/widgetastic.core/blob/master/src/widgetastic/widget/table.py#L893), which in our case simply does `[row.read() for row in rows]`. However, on second page our `Table` object will still think that table has as many rows as it had on first page. As the result, it will try to read row with number greater than number of rows that there are in the table.

This causes exception. But somehow this exception is swallowed and instead of failing test right away, whole `subscriptions.resources.read()` call is thrown away and test continues as if nothing happened. It then fails on assertion, when it tries to access `chost['subscriptions']['resources']` - which is not defined.

--------------------

Workaround is to get content of "assigned" tab directly:

```python
chost = session.contenthost.read(vm.hostname,
           widget_names=['subscriptions.resources.list_remove_tab'])
```

----------------------------

So, the final code, with both workarounds, can look like this:
```python
chost = session.contenthost.read(vm.hostname,
                widget_names=['details',       
                                'provisioning_details',
                                'subscriptions.resources.list_remove_tab'])
        session.contenthost.update(vm, {'repository_sets.limit_to_lce': True})
        ch_reposet = session.contenthost.read(vm, widget_names=['repository_sets'])
        chost = {**chost, **ch_reposet}
        assert (
            repos_collection.custom_product['name'] in
            {repo['Repository Name'] for repo in chost['subscriptions']['resources']['list_remove_tab']}
            
        )
        actual_repos = {repo['Repository Name'] for repo in chost['repository_sets']['table']}
        expected_repos = {
            repos_collection.repos_data[repo_index].get(
                'repository-set',
                repos_collection.repos_info[repo_index]['name']
            )
            for repo_index in range(len(repos_collection.repos_info))
        }
        assert actual_repos == expected_repos
```
(Note that first assertion is slightly changed)

This should work, although I have not fully tested it (`repos_collection` setup takes way too much time).

mirzal:
@latran Honestly, I have no idea. I would rule out Airgun (even ran very old version, the same issue), widgetastic (our dependency is versioned and unchanged since March or so) and switch from Saucelabs to Zalenium (Zalenium is generally faster, but couple of `sleep()`s in strategic places didn't "fix" the problem).

I did not try running against older version of Satellite, so I'm little tempted to bet on change in the product. But I wouldn't run screaming I found a bug.

Let me know if proposed workaround allowed you to move forward with this.

latran:
@mirzal Your proposed workaround has helped me in moving forward but now there's an issue with installing the security errata.  I wonder if this is a bug and I'm investigating it at the moment.

JacobCallahan:
@latran it looks like there is some overlap with your version updates in this PR and mine in [1]. Since you are already updating versions, do you want to add in what I have changed?

As mentioned in that PR, the entire automation suite will have to be run after these changes.

[1] - https://github.com/SatelliteQE/robottelo/pull/7453

latran:
@JacobCallahan done!

latran:
@mirekdlugosz Yeah, tell me about it.  All this time it was mainly setup issues and not automation issues except for those 2 changes in the file :)

PR #7335, opened 2019-09-05
CV optimize along with Entity Fixtures

jyejare:
Improvements:

1. 14 minutes and 48 seconds has been saved in this CV test module.
2. Module-wide, Class-wide, function wide fixtures for automation ease.
3. Pytest asserts instead of legacy unit-test asserts.
4. Repeating of publish/promote in some tests reduced to 3(max) instead of 5(max) which was unnecessary earlier.
5. New entity_fixtures lib in robottelo/api to be used everywhere in Tests of API/CLI/UI.
6. Lines reduced in Test module: 379

Removals:
1. Repeated asserts/actions tests (most of them asserting/repeating once replaced with multiple assertion/repetition)

### DIFF
--- a/tests/foreman/api/test_contentview.py
+++ b/tests/foreman/api/test_contentview.py
@@ -44,6 +44,7 @@ from robottelo.constants import (
 )
 from robottelo.datafactory import invalid_names_list, valid_data_list
 from robottelo.decorators import (
+    fixture,
     run_in_one_thread,
     skip_if_not_set,
     stubbed,
@@ -61,6 +62,11 @@ from robottelo.test import APITestCase
 # How many times should that be done? A higher number means a more interesting
 # but longer test.
 REPEAT = 3
+
+
+@fixture
+def module_org():
+    return entities.Organization().create()
 
 
 class ContentViewTestCase(APITestCase):
--- a/tests/foreman/api/test_contentview.py
+++ b/tests/foreman/api/test_contentview.py
@@ -20,7 +20,6 @@ import random
 
 from fauxfactory import gen_integer, gen_string, gen_utf8
 from nailgun import entities
-from nailgun.entity_mixins import TaskFailedError
 from requests.exceptions import HTTPError
 from robottelo import ssh
 from robottelo.api.utils import apply_package_filter, enable_rhrepo_and_fetchid, promote
@@ -55,7 +54,6 @@ from robottelo.decorators import (
 )
 from robottelo.decorators.host import skip_if_os
 from robottelo.helpers import get_data_file, get_nailgun_config
-from robottelo.test import APITestCase
 
 
 # Some tests repeatedly publish content views or promote content view versions.
@@ -64,17 +62,51 @@ from robottelo.test import APITestCase
 REPEAT = 3
 
 
+@fixture(scope='class')
+def class_cv(module_org):
+    return entities.ContentView(organization=module_org).create()
+
+
+@fixture(scope='class')
+def class_published_cv(class_cv):
+    class_cv.publish()
+    return class_cv.read()
+
+
+@fixture(scope='class')
+def class_promoted_cv(
+        class_published_cv, module_lce):
+    promote(class_published_cv.version[0], module_lce.id)
+    return class_published_cv.read()
+
+
+@fixture(scope='class')
+def class_cloned_cv(class_cv):
+    copied_cv_id = entities.ContentView(id=class_cv.id).copy(
+        data={u'name': gen_string('alpha', gen_integer(3, 30))})['id']
+    return entities.ContentView(id=copied_cv_id).read()
+
+
+@fixture(scope='class')
+def class_published_cloned_cv(class_cloned_cv):
+    class_cloned_cv.publish()
+    return entities.ContentView(id=class_cloned_cv.id).read()
+
+
 @fixture
-def module_org():
-    return entities.Organization().create()
+def content_view(module_org):
+    return entities.ContentView(organization=module_org).create()
 
 
-class ContentViewTestCase(APITestCase):
-    """Tests for content views."""
+@fixture
+def role():
+    return entities.Role().create()
 
+
+class TestContentView:
     @upgrade
     @tier3
-    def test_positive_subscribe_host(self):
+    def test_positive_subscribe_host(self, class_cv, class_promoted_cv, module_lce, module_org):
         """Subscribe a host to a content view
 
         :id: b5a08369-bf92-48ab-b9aa-10f5b9774b79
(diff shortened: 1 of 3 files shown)


tests/foreman/api/test_contentview.py
```diff
@@ -60,8 +62,50 @@
 REPEAT = 3
 
 
+@fixture(scope='class', autouse=True)
+def module_org(request):
+    request.cls.module_org = entities.Organization().create()
```
mirzal: It's idiomatic for pytest fixtures to return, or yield, value, not assign it to something.

Look at e.g. https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/ui/test_activationkey.py#L68 :
```python
@fixture(scope='module')
def module_org():
    return entities.Organization().create()

...

def test_positive_search_scoped(session, module_org):
    repo_id = create_sync_custom_repo(module_org.id)
```

tests/foreman/api/test_contentview.py
```diff
@@ -361,6 +355,7 @@ def test_positive_restart_publish_via_dynflow(self):
         """
 
 
+@pytest.mark.usefixtures('module_org')
```
mirzal: Actually, is there a reason for each class to have their own organization? Perhaps you could speed it up further by using one organization everywhere?

jyejare: Probably to avoid the clash of entities/actions that affect the asserts of tests. Its been followed since long and when we developed robottelo.

tests/foreman/api/test_contentview.py
jyejare: The things work differently when we have to class module-level fixtures in a class and especially when the class in inheriting unittest.TestCase.

I don't want to completely replace the class level separation of tests.

You can find more about the fixtures when used with unit test in pytest documentation.

mirzal: I see, you are right. They cover it in document on [UnitTest support](https://docs.pytest.org/en/latest/unittest.html), with explanation provided in Note at the bottom (easy to miss!).

tests/foreman/api/test_contentview.py
mirzal: "We have always done it this way" is not a very good reason, but I understand you don't want to touch too many things at one time.

tests/foreman/api/test_contentview.py
```diff
@@ -60,8 +62,50 @@
...
+        organization=request.cls.module_org).create()
+
+
+@fixture(scope='class', autouse=True)
+def module_product(request, module_org):
+    request.cls.module_product = entities.Product(
+        organization=request.cls.module_org).create()
+
+
+@pytest.mark.usefixtures('module_org', 'module_lce', 'module_product')
```
devendra104: Can we keep all the fixtures in conftest file to avoid the bulkiness of test case file?

omkarkhatavkar: We can even keep in Single file (not in conftest to avoid getting it bulky) it can be used across other files. e.g. https://gist.github.com/peterhurford/09f7dcda0ab04b95c026c60fa49c2a68 mentioned here

jyejare: @devendra104, That's a really good idea but as I said there are higher chances of collision and failures if we do CRUD to those entity objects. These objects are mutable and hence having them as test modules is OK and can be handled. But for the entire test bed it's quite uncontrollable.

Still, there could be few entities like org and loc can be used for the entire testbed as fixture (except org and loc itself) which are less likely to operate on.

jyejare: @devendra104, @omkarkhatavkar  That's a really good idea but as I said there are higher chances of collision and failures if we do CRUD to those entity objects. These objects are mutable and hence having them as test modules is OK and can be handled. But for the entire test bed it's quite uncontrollable.

Still, there could be few entities like org and loc can be used for the entire testbed as fixture (except org and loc itself) which are less likely to operate on.

jyejare:
Post Rework Difference In Test Execution Time:

Before(in master):
```
=========== 1 failed, 51 passed, 14 skipped, 2 deselected in 2460.97 seconds ==========
```
After(in my branch):
```
========== 1 failed, 81 passed, 14 skipped, 2 deselected in 1591.34 seconds =========
```

tests/foreman/api/test_contentview.py
jyejare: @omkarkhatavkar @devendra104, All the fixtures are now moved to a special library under robottelo/api/entity_fixtures.py and those all has been imported to conftest. The approach works great.

jyejare:
@vijay8451 , @omkarkhatavkar and @devendra104 @SatelliteQE/quality-engineers ,

Also, the total number of lines in this (ContentView) test module has been reduced to 1890 from 2269 and that is 379 lines are saved.

tests/foreman/api/test_contentview.py
omkarkhatavkar: @jyejare Awesome Work!

jyejare:
Dont merge until I resolve conflicts from #7625

jyejare:
The PR is all set to merge now as I have resolved conflicts with #7625

tests/foreman/api/test_contentview.py
```diff
@@ -407,11 +382,12 @@ def test_positive_restart_publish_via_dynflow(self):
...
 
 
-class ContentViewCreateTestCase(APITestCase):
+class TestContentViewCreate:
     """Create tests for content views."""
 
+    @pytest.mark.parametrize('composite', (True, False))
     @tier1
-    def test_positive_create_composite(self):
+    def test_positive_create_composite(self, composite):
```
lpramuk: passes

tests/foreman/api/test_contentview.py
```diff
@@ -421,17 +397,11 @@ def test_positive_create_composite(self):
...
-                    entities.ContentView(
-                        composite=composite
-                    ).create().composite,
-                )
+        assert entities.ContentView(composite=composite).create().composite == composite
 
+    @pytest.mark.parametrize('name', valid_data_list())
     @tier1
-    def test_positive_create_with_name(self):
+    def test_positive_create_with_name(self, name):
```
lpramuk: fails in multithread

tests/foreman/api/test_contentview.py
```diff
@@ -440,15 +410,11 @@ def test_positive_create_with_name(self):
...
-                self.assertEqual(
-                    entities.ContentView(name=name).create().name,
-                    name
-                )
+        assert entities.ContentView(name=name).create().name == name
 
+    @pytest.mark.parametrize('desc', valid_data_list())
     @tier1
-    def test_positive_create_with_description(self):
+    def test_positive_create_with_description(self, desc):
```
lpramuk: fails in multithread

tests/foreman/api/test_contentview.py
```diff
@@ -457,17 +423,10 @@ def test_positive_create_with_description(self):
...
-                    desc,
-                    entities.ContentView(
-                        description=desc
-                    ).create().description,
-                )
+        assert entities.ContentView(description=desc).create().description == desc
 
     @tier1
-    def test_positive_clone(self):
+    def test_positive_clone(self, content_view, module_org):
```
lpramuk: passes in multithread

PR #7338, opened 2019-09-06
automation bug, query doesn't support seconds

tstrych:
more information in https://bugzilla.redhat.com/show_bug.cgi?id=1740790
Maybe I should change it to minutes instead of putting bug there, anyway hot fix.


vijay8451:
@tstrych It's fine to skip this test using `skip_if_bug_open` and priorities the Bugzilla with dev to fix ...because setting up mins may mislead the others ..

Issue #7336, opened 2019-09-05
Remove Robozilla dependency on skip_if_bug_open decorator 


rochacbruno:
For now to be applied only on master branch.

PR 1: https://github.com/SatelliteQE/robottelo/pull/7346

- [x] Explicitly mark wontfixed test cases with `@pytest.mark.deselect` without calling Bugzilla and remove all the calls for bugzilla in the test collection phase. we want to keep the tests in codebase for historical reasons.
- [x] Identify tests that can run (passing tests) and remove the blocker.
- [x] Replace `@skip_if_bug_open` with explicit `@pytest.mark.skip(reason="BZ:123456")` 
- [x] Remove direct dependency on Robozilla for decorators and remove bugzilla calls.

PR 2: Done! https://gitlab.{internal-url}/satelliteqe/nagger_bz
- [x] Crate a nagger script that parses that BZ numbers and notifies component owners if a test is skipped by an already closed BZ and ask for update.

PR3: IN progress: https://gitlab.{internal_url}/satelliteqe/nagger_bz/blob/master/DOCS.md

- [ ] Update the documentation on how to use the decorators and strategies for BZ based workarounds.

rochacbruno:
@pgagne @flatfender @jyejare @sghai @mirzal 

This is the plan and I am `in progress` on the item 2, so I'll send the first PR soon.

If you have any comment.

mirzal:
> * Crate a nagger script that parses that BZ numbers and notifies component owners if a test is skipped by an already closed BZ and ask for update.

I understand it would take BZ number from new decorator (`@pytest.mark.skip(reason="BZ:123456")`), is that correct?

> The script will send emails based on
> 
> * last person that made changes on that line?
> * component owner mapping?
> * CODEOWNERS file
> * BZ qa-contact?
> * A single email to the list

I'm undecided between "component owner mapping" and "BZ qa-contact".
On one hand, component owners should own their component, which sometimes mean they need to take care of bugzillas created by other QE. On the other hand, BZ qa-contact is the best person to make any decisions regarding specific bz and any automation around it. Most of the time component owner and BZ qa-contact will be the same person, and if they differ, then probably there is a good reason for that.

But then, there might be rare cases when BZ qa-contact is unresponsive (left the team or something), while component owners are always up-to-date.

Perhaps we can do both?

[...]

PR #7346, opened 2019-09-12
Cleanup robozilla decorators and conditionals PR 1/3 ref #7336

rochacbruno:
This PR is ref #7336

- Remove call to Bugzilla to deselect WONTFIX, DEFERRED blockd tests
- Explicitly added `@pytest.mark.deselect` to 13 WONTFIX test cases
- Remove all calls to bugzilla via decorator or function
- Explicitly added `@pytest.mark.skip` on blocked tests (in PR 3/3 a nagger script will take care of it)
- Added `BEGIN:END` comment blocks, that will be parsed by nagger script in place of in function conditionals.

This PR is WIP and is 1/3 from #7336

## explanations

- Where an `if bz_bug_is_open` or a `@skip_if_bug..` was deleted means that **BZ was not open** (a.k.a Fixed, verified)
- Where `@pytest.mark.deselect` was used, means BZ is Wontfix, Deferred, Cantfix and there is no solution planned, test kept only for historical reasons.
- Where a comment `BEGIN: END` was added means that the `if` conditional was always evaulationg to `True` and then the nagger script will take care in PR 3/3

## Bonus

Added a `py.test tests/foreman/ --bzcollect /tmp/data.json` this will generate a file describing all the usage of the BZ markers. This file is the input for the nagger script to notifications.

<details>
<summary>Example</summary>

```json
{
    "skip": [
        {
            "reason": "BZ:1374253",
            "filepath": "tests/foreman/api/test_classparameters.py",
            "lineno": 710,
            "testcase": "SmartClassParametersTestCase.test_positive_impact_parameter_delete_attribute",
            "component": "Puppet" 
        },
        {
            "reason": "BZ:1463811",
            "filepath": "tests/foreman/api/test_contentmanagement.py",
            "lineno": 76,
            "testcase": "ContentManagementTestCase.test_positive_sync_repos_with_large_errata",
            "component": "Content"
        }
    ],
    "deselect": [
        {
            "reason": "BZ:1147100",
            "filepath": "tests/foreman/api/test_contentview.py",
            "lineno": 1246,
            "testcase": "ContentViewUpdateTestCase.test_negative_update_label",
            "component": "ContentView"
        },
        {
            "reason": "BZ:1310422",
            "filepath": "tests/foreman/api/test_product.py",
            "lineno": 249,
            "testcase": "ProductTestCase.test_positive_update_organization",
            "component": "Product"
        }
    ],
    "workaround": [
        {
            "reason": "BZ:1439691",
            "filepath": "tests/foreman/api/test_contentmanagement.py",
            "lineno": 396,
            "testcase": "CapsuleContentManagementTestCase.test_positive_capsule_sync"
        },
        {
            "reason": "BZ:1709683",
            "filepath": "tests/foreman/api/test_operatingsystem.py",
            "lineno": 99,
            "testcase": "OperatingSystemTestCase.test_positive_create_with_os_family"
        }
    ]
}

```

</details>


tests/foreman/ui/test_bookmark.py
```diff
@@ -49,6 +49,8 @@ def ui_entities(module_org, module_loc):
         skip = entity.get('skip_for_ui')
         if isinstance(skip, tuple):
             bug_type, bug_id = skip
+            # TODO: What to do with those dynamic cases?
+            # NOTE: we want to completely remove `*_bug_is_open` calls.
```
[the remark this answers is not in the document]
mirzal: As I understand the code, it iterates through [`BOOKMARK_ENTITIES`](https://github.com/SatelliteQE/robottelo/blob/master/robottelo/constants.py#L1672) and looks at value of `skip_for_ui` key. If that value is True, it skips entity unconditionally. If that value is tuple, it does `*_bug_is_open` call.

There are only two entities using tuple, `ContentCredential` depending on bz#1638781 (CLOSED ERRATA) and `DiscoveryRule` depending on bz#1387569 (CLOSED WONTFIX).

In context of master, both these can be switched to booleans and this entire `if isinstance(skip, tuple):` part can be removed.

tests/foreman/api/test_syncplan.py
```diff
@@ -776,11 +777,9 @@ def test_positive_synchronize_custom_product_future_sync_date(self):
         self.validate_repo_content(
             repo, ['erratum', 'package', 'package_group'], after_sync=False)
         # Create and Associate sync plan with product
-        if bz_bug_is_open(1695733):
-            self.logger.info('Need to set seconds to zero because BZ#1695733')
-            sync_date = datetime.utcnow().replace(second=0) + timedelta(seconds=delay)
-        else:
-            sync_date = datetime.utcnow() + timedelta(seconds=delay),
+        self.logger.info('Need to set seconds to zero because BZ:1695733')
+        # sync_date = datetime.utcnow() + timedelta(seconds=delay)
```
mirzal: Why is old code left commented? Shouldn't we use these `BEGIN: END` markers here?

tests/foreman/cli/test_computeresource_rhev.py
```diff
@@ -267,7 +272,7 @@ def test_negative_add_image_rhev_with_invalid_name(self):
         """
 
     @tier2
-    @skip_if_bug_open('bugzilla', 1278917)
+    @pytest.mark.deselect(reason="BZ:1278917")
     @stubbed()
```
mirzal: It's stubbed, meaning it would always be deselected anyway (never be selected in first place? don't know what pytest does internally)

tests/foreman/cli/test_repository.py
```diff
@@ -1227,9 +1226,12 @@ def test_positive_synchronize_rpm_repo_ignore_content(self):
...
         self.assertEqual(repo['content-counts']['errata'], '2',
                          'content not synced correctly')
-        if not bz_bug_is_open(1664549):
-            self.assertEqual(repo['content-counts']['source-rpms'], '3',
-                             'content not synced correctly')
+
+        # BEGIN workaround for BZ: 1664549
+        # self.assertEqual(repo['content-counts']['source-rpms'], '3',
+        #                 'content not synced correctly')
+        # END workaround for BZ: 1664549
```
mirzal: Workaround that is commented out?

tests/foreman/api/test_contentmanagement.py
```diff
@@ -493,28 +491,27 @@ def test_positive_capsule_sync(self):
...
-            sync_status = capsule.content_get_sync()
-            self.assertTrue(
-                len(sync_status['active_sync_tasks']) >= 1
-                or sync_status['last_sync_time'] != last_sync_time
-            )
-            for task in sync_status['active_sync_tasks']:
-                entities.ForemanTask(id=task['id']).poll()
-            sync_status = capsule.content_get_sync()
-            last_sync_time = sync_status['last_sync_time']
+        # If BZ1439691 is CLOSED WONTFIX, need to sync repo once more,
```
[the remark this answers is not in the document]
rochacbruno: @mirzal another BZ can be opened as a Duplicate of this one, which will make it dependent of this one and if the dup is resolved this one whould also be considered resolved. Our nagger script will need to lookup for duplicates and other relations.

tests/foreman/cli/test_computeresource_rhev.py
rochacbruno: We run tests in CI using `-m not stubbed` this will `deselect` all the stubbed test cases before pytest can even catch the `.deselect` marker.

But I think we can keep the marker here, so our script can catch the status of the BZ (the script will collect everything including stubs) and notify changes that will be useful to update the stubbed for when it is automated.

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -283,12 +276,17 @@ def test_negative_create_with_invalid_hostname(self):
         :CaseImportance: Medium
 
         :CaseLevel: Component
+
+        :BZ: 1378427
         """
         for name in self.invalid_hostnames_list():
             with self.subTest(name):
                 bug_id = name.pop('bugzilla', None)
```
rochacbruno: The nagger script will also look for BZs on collections like this `invalid_hostnames_list():` and notify if changes, so it addresses dynamic cases like this one

tests/foreman/cli/test_repository.py
rochacbruno: When BZ is resolved then the assertion can be re-included. I added a better note about it

tests/foreman/ui/test_bookmark.py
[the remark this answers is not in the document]
rochacbruno: ok, so the script will need to look to that `BOOKMARK_ENTITIES` also.

tests/foreman/api/test_contentmanagement.py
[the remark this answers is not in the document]
mirzal: I wonder if there is any instance of bug being closed WONTFIX, then getting a duplicate, and then that duplicate getting properly fixed. I would expect CLOSED WONTFIX duplicates to be closed almost automatically.

Anyway, I was mostly referring to grammar of this comment. "If BZ1439691 is CLOSED WONTFIX" is misleading for me, because that BZ is CLOSED WONTFIX already. If we care about possible duplicates with other states, I propose "If BZ1439691 and all duplicates are CLOSED WONTFIX".

rochacbruno:
This is ready for review and merge

Tier 1 tests runs:

<SATQE_JENKINS>/view/Satellite%206/job/satellite6-standalone-automation/1630/testReport/
and
<SATQE_JENKINS>/view/Satellite%206/job/satellite6-standalone-automation/1631/testReport/

Thw 214 failures are not related to any BZ blocker.

rochacbruno:
@lpramuk that was stated that `we will completely remove robozilla` that was decided on the email thread, so I came up with 2 ideas.

1. Using comment block:

```py
# BEGIN BZ: 12345
workaround here
# END BZ: 123456
```

2. Using a context manager

```py
with BZ("123456"):
    workaround here
```

I implemented the option 1 because I was thinking the comment was easier to do, but if others agree we can easily go with the context manager.

Or if you have a suggestion on how to resolve this that will be appreciated! Considering we don't call bugzilla API anymore, we don't have a way to to an `if` conditional based on its status.

rochacbruno:
another thing we can do is:

Do not use workarounds anymore!

Then we can mark blocked test as `xfail(reason="BZ:123456)`  the  reports will know it is expected failure, will not get our reports down. And when test suddenly passes we will get the `xpassed` result to act on it by removing the xfail mark.

tests/foreman/conftest.py
```diff
@@ -119,63 +131,90 @@ def log_test_execution(robottelo_logger, request):
...
+    bzcollect_data = {'skip': [], 'deselect': [], 'workaround': []}
 
-def _extract_setup_class_ids(item):
-    setup_class_method = getattr(item.parent.obj, 'setUpClass', None)
-    return getattr(setup_class_method, 'bugzilla_ids', [])
+    for item in items:
+        filepath, lineno, testcase = item.location
+        # Take the module level :CaseComponent:
+        module_component = None
+        with open(os.path.join(os.path.sep, filepath), 'r') as modulefile:
```
mirzal: Was that here before? because now when I try to run it, I get exception:
```
$ pytest tests/foreman/api/ --bzcollect
...
INTERNALERROR>   File "/home/mzalewsk/.virtualenvs/robottelo/lib64/python3.6/site-packages/pluggy/callers.py", line 187, in _multicall
INTERNALERROR>     res = hook_impl.function(*args)
INTERNALERROR>   File "/home/mzalewsk/sources/robottelo/tests/foreman/conftest.py", line 201, in pytest_collection_modifyitems
INTERNALERROR>     generate_bz_collection(items, config, filename=bzcollect)
INTERNALERROR>   File "/home/mzalewsk/sources/robottelo/tests/foreman/conftest.py", line 143, in generate_bz_collection
INTERNALERROR>     with open(os.path.join(os.path.sep, filepath), 'r') as modulefile:
INTERNALERROR> FileNotFoundError: [Errno 2] No such file or directory: '/tests/foreman/api/test_activationkey.py'
```

rochacbruno: @mirzal I just pushed a change to use the `module.__file__` as looks like the `item.location` is not absolute.

.gitignore
```diff
@@ -50,3 +50,6 @@
 
 .pytest_cache/
 tests/foreman/pytest.ini
+
+# For --bzcollect generated file
+/bzcollect.json
```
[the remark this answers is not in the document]
rochacbruno: Changed to `**/bzcollect.json` so it filters this file in every subdir

tests/foreman/conftest.py
```diff
@@ -1,18 +1,29 @@
...
+
+
+BEGIN_BZ = re.compile(
+    # To match `# BEGIN BZ:12345`
+    r"\s*#\s*BEGIN\s*(?P<source>\D{2})\s*:\s*(?P<number>\d*)"
+)
+COMPONENT = re.compile(
+    # To match :CaseComponent: FooBar
+    r"\s*:CaseComponent:\s*(?P<component>\S*)"
+)
```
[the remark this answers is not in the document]
rochacbruno: done, the only problem is that `re.IGNORECASE` makes the search a bit slow, now it takes ~30 seconds.. but I think it is not a problem.

flatfender:
> I went through file changes last time, so this time I focused only on `pytest --bzcollect`.
> 
> LGTM, but not giving ACK to allow for discussion. But honestly, "CleanUp: skip_if_bug_open / bz_bug_is_open decorator" thread was started on satqe-list by @jyejare back in July, and Bruno shared his plan month ago. At some point we need to draw the line and assume everyone who was interested in the topic already spoke out.

I'll 2nd this, There was ample opportunity to participate and shape the direction here.  Let's not hold the train up at the 2nd to last stop b/c someone didn't get on 3 stops ago.

tests/foreman/conftest.py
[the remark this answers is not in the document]
mirzal: I'll take it as a price we have to pay for being technically correct.

docs/features/decorators.rst
```diff
@@ -34,36 +34,6 @@ Please note that ''stubbed'' is a decorator generator, and cannot be used as a
...
-        """
-
-Decorator communicates with bug tracker to know the status of respective bug
-id. If it is still open, test will be skipped. Once bz is fixed and moved to
-ON_QA or Verified state, test will automatically be run with next builds.
-
-This is useful to allow QE team automating test cases even before a bug is
-fixed or a feature is implemented, while test will not be considered a build
-fail until the issue has it state changed.
-
```
pondrejk: Well, I'd say we should document the suggested way of doing things to ease the adoption. Explaining the rationale in docs for future use is definitely time well spent

rochacbruno: @pondrejk to anble easier review I added this task to https://github.com/SatelliteQE/robottelo/issues/7336 and will send the new docs in a separate PR.

pondrejk:
Hi all, having read trough the discussion here and in the email thread, I didn't find an explanation on how this initiative plays along with the closed loop process. I'm concerned it kind of contradicts it. I had put more details here https://github.com/SatelliteQE/robottelo/issues/7336#issuecomment-535018824

robottelo/vm_capsule.py
```diff
@@ -203,13 +203,13 @@ def _capsule_cleanup(self):
...
-                        self.hostname, exp))
-                else:
-                    logger.error('Failed to cleanup the host: {0}\n{1}'.format(
-                        self.hostname, exp))
-                    raise
+
+                # BEGIN BZ:1622064
+                logger.warn('Failed to cleanup the host: {0}\n{1}'.format(
+                    self.hostname, exp))
+                # when BZ is closed the exception should `raise` here.
```
rplevka: so how exactly I'll find out when the BZ gets closed and I'm supposed to update something here?

robottelo/vm_capsule.py
```diff
@@ -295,16 +291,17 @@ def _setup_capsule(self):
...
+            # BEGIN BZ:1446930
+            result = self.run('systemctl start pulp_celerybeat.service')
+            if result.return_code != 0:
                 raise CapsuleVirtualMachineError(
-                    'pulp_celerybeat service not running')
+                    'Failed to start pulp_celerybeat service\n{}'.format(
+                        result.stderr)
+                )
+            # When BZ is fixed the above can be replaced with the below.
+            # raise CapsuleVirtualMachineError(
```
rplevka: so we already have the code here, but we just going to comment it out so it requires manual intervention once the bug gets closed? like...seriously? :D

tests/foreman/api/test_classparameters.py
```diff
@@ -707,7 +708,7 @@ def test_negative_enable_avoid_duplicates_checkbox(self):
         )
         self.assertEqual(sc_param.read().avoid_duplicates, False)
 
-    @skip_if_bug_open('bugzilla', 1374253)
+    @pytest.mark.skip(reason="BZ:1374253")
```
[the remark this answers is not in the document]
rplevka: what if i'm running the tests on my local machine with no `nagger` script output or anything? The benefit of of the `bug_open`-related methods was, that it already contained the logic to figure out whether the test/workaround should be ran or skipped.

rochacbruno:
@rplevka 

The `bz_bug_is_open` is provided by Robozilla as you can see in:

https://github.com/SatelliteQE/robottelo/blob/master/robottelo/decorators/__init__.py#L10
https://github.com/SatelliteQE/robottelo/blob/master/robottelo/decorators/__init__.py#L349

We can implement a new/better `bz_bug_is_open` function, but for that we will need to re-create all the logic behind it which is very complicated right now
https://github.com/SatelliteQE/robozilla/blob/bf6f599c4279f77eeccb9c263ba329ec993b9913/robozilla/decorators/__init__.py#L170-L454

By the email thread conversation it was clear that waht we want is actually more manual intervention and less bugzilla calls.

I am ok to re-implement the bugzilla logic (as part of it will be in nagger script anyway).

robottelo/vm_capsule.py
rochacbruno: that is the only I found given the constraint "no more bugzilla calls at test runtime"

robottelo/vm_capsule.py
rochacbruno: You will receive an email saying "BZ number 1622064 is now ON_QA, action required on robottelo/vm_capsule.py".

Then you will need to figure out what to do based on the BZ , code context + any good comment that was left on the block. To ensure that we will need to be attentive on reviewing the addition of those blocks.

rochacbruno: and of course you will need to verify the BZ state before making changes on the code

rochacbruno:
@rplevka 

> So we literally add requirement for a manual intervention and a dependency on some external script that would remind us to take action instead of an already existing and implemented logic that pretty much kept doing this for us.

The existing logic is behind robozilla, which we are removing, as I said before I am happy to implement a better version of it, it will take some more time and will require maintainance as our BZ rules are not so simple and because of that the logic begind robozilla got so complex.

It was coordinated with management and others on the email thread that we are moving to a more "component owner manual action" based approach.

> So in some cases, once the bug gets fixed, the test is actually going to fail as false negative, negatively influencing the PQI of the product?

Yes that is something we can further discuss, however I think the goal here is to take action ASAP, so if the nagger script runs twice a day, we have 12 hours window of a notification received to verify the BZ state and take the action, if it doesn't work then we can think in a better approach.

So what would be the ideal workflow?

I think:

1. There is a case I expect to fail because of an open BZ
2. I go to the test case and I use the `@pytest.mark.xfail(BZ:123456)` (expected failure)
3. If test fails, it will not decrease the PQI because it was an xfail
4. If test passes it will be reported as an `xpassed` (exceed pass) and will count positively on PQI and also raise the action to remove the `xfail` form it.

rplevka:
@rochacbruno - why don't we simply keep everything as is (skip_if_bz_open-wise) and just extend the decorators/conditional functions to log messages (or make pytest to generate reports) about already closed bugzillas that our code made calls for, so once in a while the component owner (or really anybody) can look at this report and do some regular cleanup PR without causing any disruption whatsoever?

rochacbruno:
@rplevka satqe-list July 1st "CleanUp: skip_if_bug_open / bz_bug_is_open decorator" thread.

flatfender:
> Hi all, having read trough the discussion here and in the email thread, I didn't find an explanation on how this initiative plays along with the closed loop process. I'm concerned it kind of contradicts it. I had put more details here [#7336 (comment)](https://github.com/SatelliteQE/robottelo/issues/7336#issuecomment-535018824)

This has no effect on closed loop what so ever.

rochacbruno:
Rebased and Cleaned more tests that got closed after this PR was opened.

rochacbruno:
New PR is coming :)

Issue #7347, opened 2019-09-13
Code refactoring of Pre/Post upgrade test cases. 


devendra104:
I am planning to create a common file in which we will keep all the reusable function/methods at a single place related to pre/post-upgrade. It will help us to reduce the bulkiness of the test case file and prevent the duplication.

PR #7090, opened 2019-07-01
fixed the goferd checking process

omkarkhatavkar:
### PR Objective 
Fix upgrade test, fix the verification of goferd process check

### Reason
```
[root@0scenariokatelloclientdifGbJjMaQrhel7 ~]# ps -aux
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root         1  0.0  0.0  11688  1428 ?        Ss   Jun30   0:00 /bin/sh /tmp/st
root       214  0.0  0.0   4404   360 ?        S    Jun30   0:00 tail -f /var/lo
root       215  0.0  0.0  11820  1904 ?        Ss   05:36   0:00 /bin/bash
root       255  0.2  0.3 852376 30204 ?        Tl   05:42   0:00 python2 /usr/bi
root       295  0.0  0.0  51744  1732 ?        R+   05:45   0:00 ps -aux

[root@0scenariokatelloclientdifGbJjMaQrhel7 ~]# ps -ef 
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0  0 Jun30 ?        00:00:00 /bin/sh /tmp/startup.sh
root       214     1  0 Jun30 ?        00:00:00 tail -f /var/log/rhsm/rhsm.log
root       215     0  0 05:36 ?        00:00:00 /bin/bash
root       255   215  0 05:42 ?        00:00:00 python2 /usr/bin/goferd -f
root       296   215  0 05:46 ?        00:00:00 ps -ef
```
 ### Solution
Some how ps aux not showing goferd process, so using ps -ef instead ps aux

PR #7349, opened 2019-09-16
code refactor of upgrade test cases

devendra104:
#7347

### DIFF
--- a/robottelo/cleanup.py
+++ b/robottelo/cleanup.py
@@ -6,7 +6,6 @@ from nailgun import entities
 from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.proxy import Proxy
 from robottelo.decorators import bz_bug_is_open
-from robottelo.vm import VirtualMachine
 
 
 LOGGER = logging.getLogger(__name__)
@@ -58,21 +57,3 @@ def vm_cleanup(vm):
     :param robottelo.vm.VirtualMachine vm: virtual machine to destroy
     """
     vm.destroy()
-
-
-def cleanup_of_provisioned_server(hostname=None, provisioning_server=None, distro=None):
-    """ Cleanup the VM from provisioning server
-
-    :param str hostname: The content host hostname
-    :param str provisioning_server: provision server name
-    :param str distro: distro type
-    """
-    if hostname:
-        vm = VirtualMachine(
-            hostname=hostname,
-            target_image=hostname,
-            provisioning_server=provisioning_server,
-            distro=distro,
-            )
-        vm._created = True
-        vm.destroy()
--- a/robottelo/cleanup.py
+++ b/robottelo/cleanup.py
@@ -6,7 +6,7 @@ from nailgun import entities
 from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.proxy import Proxy
 from robottelo.decorators import bz_bug_is_open
-
+from robottelo.vm import VirtualMachine
 
 LOGGER = logging.getLogger(__name__)
 
@@ -57,3 +57,22 @@ def vm_cleanup(vm):
     :param robottelo.vm.VirtualMachine vm: virtual machine to destroy
     """
     vm.destroy()
+
+
+def cleanup_of_provisioned_server(hostname=None, provisioning_server=None,
+                                  distro=None):
+    """ Cleanup the VM from provisioning server
+
+    :param: str hostname: The content host hostname
+    :param: str provisioning_server: provision server name
+    :param: str distro: distro type
+    """
+    if hostname:
+        vm = VirtualMachine(
+            hostname=hostname,
+            target_image=hostname,
+            provisioning_server=provisioning_server,
+            distro=distro,
+        )
+        vm._created = True
+        vm.destroy()
--- a/robottelo/upgrade_utility.py
+++ b/robottelo/upgrade_utility.py
@@ -9,10 +9,8 @@ from robottelo.api.utils import call_entity_method_with_timeout
 
 
 class CommonUpgradeUtility(object):
-    """
-    This class will be used by upgrade test cases to re-utilize all the defined module
-    in their test cases execution.
-    """
+    """ CommonUpgradeUtility class for upgrade test cases """
+
     def __init__(self, container_id=None, package=None):
         self.client_container_id = container_id
         self.package = package
--- a/robottelo/upgrade_utility.py
+++ b/robottelo/upgrade_utility.py
@@ -1,10 +1,11 @@
 """Common Upgrade test utilities """
 
-from wait_for import wait_for
 from fabric.api import execute, run
-from robottelo.vm import VirtualMachine, settings
+from nailgun import entities
+from robottelo.test import settings
+from robottelo.api.utils import call_entity_method_with_timeout
 from upgrade.helpers.docker import docker_execute_command
-from robottelo.api.utils import call_entity_method_with_timeout, entities
+from wait_for import wait_for
 
 
 class CommonUpgradeUtility(object):
@@ -102,9 +103,10 @@ class CommonUpgradeUtility(object):
             query={'search': '{0}'.format(client_container_name)})
         return host
 
-    def host_location_update(self, client_container_name=None, loc=None):
-        """ Check the content host status (as package profile update task does take time to
-        upload) and update location.
+    def host_location_update(self, client_container_name=None,
+                             logger_obj=None, loc=None):
+        """ Check the content host status (as package profile update task does
+        take time to upload) and update location.
 
         :param: str client_container_name: The content host hostname
         :param: str loc: Location
@@ -115,7 +117,7 @@ class CommonUpgradeUtility(object):
                                              )) > 0,
                 timeout=100,
                 delay=2,
-                logger=self.logger
+                logger=logger_obj
             )
         host_loc = self.host_status(client_container_name=client_container_name)[0]
         host_loc.location = loc
@@ -136,22 +138,3 @@ class CommonUpgradeUtility(object):
         call_entity_method_with_timeout(content_view.publish, timeout=3400)
         content_view = content_view.read()
         return content_view
-
-    @staticmethod
-    def cleanup_of_provisioned_server(hostname=None, provisioning_server=None,
-                                      distro=None):
-        """ Cleanup the VM from provisioning server
-
-        :param: str hostname: The content host hostname
-        :param: str provisioning_server: provision server name
-        :param: str distro: distro type
-        """
-        if hostname:
-            vm = VirtualMachine(
-                hostname=hostname,
-                target_image=hostname,
-                provisioning_server=provisioning_server,
-                distro=distro,
-            )
-            vm._created = True
-            vm.destroy()
--- a/robottelo/upgrade_utility.py
+++ b/robottelo/upgrade_utility.py
@@ -2,17 +2,16 @@
 
 from wait_for import wait_for
 from fabric.api import execute, run
-from robottelo.vm import VirtualMachine
+from robottelo.vm import VirtualMachine, settings
 from upgrade.helpers.docker import docker_execute_command
-from robottelo.api.utils import call_entity_method_with_timeout, entities, settings
+from robottelo.api.utils import call_entity_method_with_timeout, entities
 
 
 class CommonUpgradeUtility(object):
     """ CommonUpgradeUtility class for upgrade test cases """
 
-    def __init__(self, container_id=None, package=None):
+    def __init__(self, container_id=None):
         self.client_container_id = container_id
-        self.package = package
 
     def run_goferd(self):
         """
@@ -33,11 +32,11 @@ class CommonUpgradeUtility(object):
             'goferd -f',
             **kwargs
         )
-        status = execute(docker_execute_command, self.client_container_id, 'ps -aux',
+        status = execute(docker_execute_command, self.client_container_id, 'ps -ef',
                          host=settings.upgrade.docker_vm)[settings.upgrade.docker_vm]
         return status
 
-    def check_package_installed(self):
+    def check_package_installed(self, package=None):
         """
         Verify if package is installed on docker content host."""
 
@@ -45,15 +44,16 @@ class CommonUpgradeUtility(object):
         installed_package = execute(
             docker_execute_command,
             self.client_container_id,
-            'yum list installed {}'.format(self.package),
+            'yum list installed {}'.format(package),
             **kwargs
         )[settings.upgrade.docker_vm]
         return installed_package
 
-    def install_or_update_package(self, update=False):
+    def install_or_update_package(self, update=False, package=None):
         """
         Install/update the package on docker content host.
         :param bool update:
+        :param str package:
         :return:
         """
         kwargs = {'host': settings.upgrade.docker_vm}
@@ -62,13 +62,13 @@ class CommonUpgradeUtility(object):
                 'subscription-manager repos --enable=*;yum clean all',
                 **kwargs)[settings.upgrade.docker_vm]
         if update:
-            command = 'yum update -y {}'.format(self.package)
+            command = 'yum update -y {}'.format(package)
         else:
-            command = 'yum install -y {}'.format(self.package)
+            command = 'yum install -y {}'.format(package)
 
         execute(docker_execute_command, self.client_container_id, command, **kwargs)[
             settings.upgrade.docker_vm]
-        self.assertIn(self.check_package_installed(), self.package)
+        self.assertIn(self.check_package_installed(), package)
 
     def create_repo(self, rpm_name, repo_path, post_upgrade=False, other_rpm=None):
         """ Creates a custom yum repository, that will be synced to satellite
--- a/robottelo/upgrade_utility.py
+++ b/robottelo/upgrade_utility.py
@@ -5,6 +5,7 @@ from wait_for import wait_for
 from fabric.api import execute, run
 from robottelo.test import settings
 from upgrade.helpers.docker import docker_execute_command
+from robottelo.api.utils import call_entity_method_with_timeout
 
 
 class CommonUpgradeUtility(object):
@@ -17,7 +18,10 @@ class CommonUpgradeUtility(object):
         self.package = package
 
     def run_goferd(self):
-        """Start the goferd process."""
+        """
+        Start the goferd process.
+        :return: status of execution
+        """
 
         kwargs = {'async': True, 'host': settings.upgrade.docker_vm}
         execute(
@@ -37,7 +41,8 @@ class CommonUpgradeUtility(object):
         return status
 
     def check_package_installed(self):
-        """Verify if package is installed on docker content host."""
+        """
+        Verify if package is installed on docker content host."""
 
         kwargs = {'host': settings.upgrade.docker_vm}
         installed_package = execute(
@@ -52,6 +57,7 @@ class CommonUpgradeUtility(object):
         """
         Install/update the package on docker content host.
         :param bool update:
+        :return:
         """
         kwargs = {'host': settings.upgrade.docker_vm}
         execute(docker_execute_command,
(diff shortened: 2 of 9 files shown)


robottelo/upgrade_utility.py
```diff
@@ -0,0 +1,150 @@
+"""Common Upgrade test utilities """
+
+from nailgun import entities
+from wait_for import wait_for
+from fabric.api import execute, run
+from robottelo.test import settings
+from upgrade.helpers.docker import docker_execute_command
+from robottelo.api.utils import call_entity_method_with_timeout
+
```
omkarkhatavkar: sort imports

robottelo/upgrade_utility.py
```diff
@@ -0,0 +1,150 @@
+"""Common Upgrade test utilities """
+
+from nailgun import entities
+from wait_for import wait_for
+from fabric.api import execute, run
+from robottelo.test import settings
```
omkarkhatavkar: is that correct? or` from robottelo.config import settings`

robottelo/upgrade_utility.py
```diff
@@ -0,0 +1,150 @@
...
+            'pkill -f gofer',
+            **kwargs
+        )
+        execute(
+            docker_execute_command,
+            self.client_container_id,
+            'goferd -f',
+            **kwargs
+        )
+        status = execute(docker_execute_command, self.client_container_id, 'ps -aux',
```
omkarkhatavkar: Can we keep `ps -ef` I have seen this https://github.com/SatelliteQE/robottelo/pull/7090 earlier

robottelo/upgrade_utility.py
```diff
@@ -0,0 +1,169 @@
...
+                         host=settings.upgrade.docker_vm)[settings.upgrade.docker_vm]
+        return status
+
+    def check_package_installed(self):
+        """
+        Verify if package is installed on docker content host."""
```
vijay8451: Add `Return` as well ..

robottelo/upgrade_utility.py
```diff
@@ -0,0 +1,169 @@
...
+            'yum list installed {}'.format(self.package),
+            **kwargs
+        )[settings.upgrade.docker_vm]
+        return installed_package
+
+    def install_or_update_package(self, update=False):
+        """
+        Install/update the package on docker content host.
+        :param bool update:
+        :return:
```
vijay8451:
```suggestion
    
```

robottelo/upgrade_utility.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def create_repo(self, rpm_name, repo_path, post_upgrade=False, other_rpm=None):
+        """ Creates a custom yum repository, that will be synced to satellite
+        and later to capsule from satellite
+        :param str rpm_name : rpm name, required to create a repository.
+        :param str repo_path: Name of the repository path
+        :param bool post_upgrade: For Pre-upgrade, post_upgrade value will be False
+        :param str other_rpm: If we want to clean a specific rpm and update with
+        latest then we pass other rpm.
+        :return:
```
vijay8451:
```suggestion
```

robottelo/upgrade_utility.py
```diff
@@ -0,0 +1,169 @@
...
+            query={'search': '{0}'.format(client_container_name)})
+        return host
+
+    def host_location_update(self, client_container_name=None, loc=None):
+        """ Check the content host status (as package profile update task does take time to
+        upload) and update location.
+
+        :param: str client_container_name: The content host hostname
+        :param: str loc: Location
+        :return:
```
vijay8451:
```suggestion
```

robottelo/upgrade_utility.py
```diff
@@ -0,0 +1,169 @@
...
+                logger=self.logger
+            )
+        host_loc = self.host_status(client_container_name=client_container_name)[0]
+        host_loc.location = loc
+        host_loc.update(['location'])
+
+    def publish_content_view(self, org=None, repolist=None):
+        """
+        publish content view and return content view
+        :param str org: Name of the organisation
```
vijay8451:
```suggestion
        :param: str org: Name of the organisation
```

robottelo/upgrade_utility.py
```diff
@@ -0,0 +1,169 @@
...
+        :param str repolist: Name of the repolist
```
vijay8451: same as above

robottelo/upgrade_utility.py
```diff
@@ -0,0 +1,169 @@
...
+        :return Return content view
```
vijay8451: same as above.

robottelo/upgrade_utility.py
```diff
@@ -0,0 +1,169 @@
...
+        content_view = content_view.read()
+        return content_view
+
+    def cleanup_of_provisioned_server(self, hostname=None, provisioning_server=None,
+                                      distro=None):
+        """ Cleanup the VM from provisioning server
+
+        :param str hostname: The content host hostname
+        :param str provisioning_server: provision server name
+        :param str distro: distro type
```
vijay8451: same for all above `:param`

robottelo/upgrade_utility.py
```diff
@@ -0,0 +1,169 @@
...
+            vm.destroy()
+
+    def assertIn(self, member, container):
+        """
+        This method is used to check the member existence in
+        container otherwise raise an exception.
+
+        :param member
+        :param container
+        :return:
```
vijay8451: same as above

tests/upgrades/test_capsule.py
```diff
@@ -85,7 +76,7 @@ def test_pre_user_scenario_capsule_sync(self):
         ak_env = ak.environment.read()
         product = entities.Product(
             name=self.prod_name, organization=self.org_id).create()
-        self.create_repo()
+        CommonUpgradeUtility().create_repo(rpm1, self.repo_name)
```
omkarkhatavkar: is it we need to pass self.repo_name, I thought self.repo_path

tests/upgrades/test_capsule.py
```diff
@@ -200,7 +181,7 @@ def test_post_user_scenario_capsule_sync_2(self):
         ak_env = ak.environment.read()
         product = entities.Product(
             name=self.prod_name, organization=self.org_id).create()
-        self.create_repo()
+        CommonUpgradeUtility().create_repo(rpm2, self.repo_name)
```
omkarkhatavkar: same here we only passing repo_name, why not passing repo_path

tests/upgrades/test_capsule.py
```diff
@@ -18,6 +18,7 @@
 from nailgun import entities
 from robottelo.test import APITestCase, settings
 from robottelo.api.utils import promote, call_entity_method_with_timeout
+from robottelo.upgrade_utility import CommonUpgradeUtility
```
omkarkhatavkar: sort imports

robottelo/cleanup.py
```diff
@@ -58,21 +57,3 @@ def vm_cleanup(vm):
     :param robottelo.vm.VirtualMachine vm: virtual machine to destroy
     """
     vm.destroy()
-
-
-def cleanup_of_provisioned_server(hostname=None, provisioning_server=None, distro=None):
```
vijay8451: Why are we removing this one from here ?

tests/upgrades/test_errata.py
```diff
@@ -249,11 +173,16 @@ def test_pre_scenario_generate_errata_for_client(self):
             delay=2,
             logger=self.logger
         )
-        self._install_or_update_package(client_container_id, 'katello-agent')
-        self._run_goferd(client_container_id)
+
+        CommonUpgradeUtility(container_id=client_container_id, package="katello-agent"). \
+            install_or_update_package(update=True)
+        self.assertIn("goferd", CommonUpgradeUtility(container_id=client_container_id).
```
vijay8451: I think it would be great to keep assertion in utils itself .. else we have to everywhere specify this assertion everywhere when calling ..

robottelo/upgrade_utility.py
```diff
@@ -0,0 +1,150 @@
...
+from upgrade.helpers.docker import docker_execute_command
+from robottelo.api.utils import call_entity_method_with_timeout
+
+
+class CommonUpgradeUtility(object):
+    """ CommonUpgradeUtility class for upgrade test cases """
+
+    def __init__(self, container_id=None, package=None):
+        self.client_container_id = container_id
+        self.package = package
```
omkarkhatavkar: I think you can remove the package from here and add it in the method, package is not constant over the class whereas container_id will be constant most of the time over a test case.

tests/upgrades/test_errata.py
```diff
@@ -316,10 +245,10 @@ def test_post_scenario_errata_count_installtion(self):
         content_view = content_view.update(['repository'])
         content_view.publish()
 
-        self._install_or_update_package(client_container_id,
-                                        "katello-agent",
-                                        update=True)
-        self._run_goferd(client_container_id)
+        CommonUpgradeUtility(container_id=client_container_id, package="katello-agent"). \
+            install_or_update_package(update=True)
+        self.assertIn("goferd", CommonUpgradeUtility(container_id=client_container_id).
```
vijay8451: same as above

tests/upgrades/test_errata.py
```diff
@@ -451,8 +382,10 @@ def test_pre_scenario_generate_errata_with_previous_version_katello_agent_client
                 client_container_id,
                 'yum update -y',
                 host=self.docker_vm)[self.docker_vm]
-        self._install_or_update_package(client_container_id, 'katello-agent')
-        self._run_goferd(client_container_id)
+        CommonUpgradeUtility(container_id=client_container_id, package="katello-agent"). \
+            install_or_update_package(update=True)
+        self.assertIn("goferd", CommonUpgradeUtility(container_id=client_container_id).
```
vijay8451: same as above ..

tests/upgrades/test_yum_plugins.py
```diff
@@ -223,8 +157,10 @@ def test_pre_scenario_yum_plugins_count(self):
                          host=self.docker_vm)[self.docker_vm]
         self.assertIn(self.org.label, status)
 
-        self._install_or_update_package(client_container_id, 'katello-agent')
-        self._run_goferd(client_container_id)
+        CommonUpgradeUtility(container_id=client_container_id, package="katello-agent").\
+            install_or_update_package(update=True)
+        self.assertIn("goferd", CommonUpgradeUtility(container_id=client_container_id).
```
vijay8451: same as above ..

tests/upgrades/test_errata.py
```diff
@@ -316,10 +245,10 @@ def test_post_scenario_errata_count_installtion(self):
...
         content_view.publish()
 
-        self._install_or_update_package(client_container_id,
-                                        "katello-agent",
-                                        update=True)
-        self._run_goferd(client_container_id)
+        CommonUpgradeUtility(container_id=client_container_id, package="katello-agent"). \
+            install_or_update_package(update=True)
+        self.assertIn("goferd", CommonUpgradeUtility(container_id=client_container_id).
+                      run_goferd())
```
omkarkhatavkar:
```suggestion
          common_upgrade = CommonUpgradeUtility(container_id=client_container_id)
          common_upgrade.install_or_update_package(update=True, package='katello-agent')
          common_upgrade.run_goferd()
```

tests/upgrades/test_yum_plugins.py
```diff
@@ -267,6 +203,8 @@ def test_post_scenario_yum_plugins_count(self):
 
         attach_custom_product_subscription(prod_name=product.name,
                                            host_name=client_container_name)
-        self._install_or_update_package(client_container_id, "katello-agent", update=True)
-        self._run_goferd(client_container_id)
+        CommonUpgradeUtility(container_id=client_container_id, package="katello-agent").\
+            install_or_update_package(update=True)
+        self.assertIn("goferd", CommonUpgradeUtility(container_id=client_container_id).
```
vijay8451: same as above ..

robottelo/cleanup.py
devendra104: I want to keep everything related to upgrade into upgrade_utility file.

tests/upgrades/test_errata.py
```diff
@@ -249,11 +173,16 @@ def test_pre_scenario_generate_errata_for_client(self):
...
 
         for package in FAKE_9_YUM_OUTDATED_PACKAGES:
-            self._install_or_update_package(client_container_id, package)
+            CommonUpgradeUtility(container_id=client_container_id,
+                                 package=package).install_or_update_package(update=True)
+
```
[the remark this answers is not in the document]
devendra104: Fixed.

robottelo/upgrade_utility.py
devendra104: Fixed.

tests/upgrades/test_errata.py
devendra104: Fixed

tests/upgrades/test_errata.py
devendra104: Fixed

tests/upgrades/test_yum_plugins.py
devendra104: Fixed

tests/upgrades/test_yum_plugins.py
devendra104: fixed

tests/upgrades/test_errata.py
devendra104: Fixed

robottelo/upgrade_utility.py
```diff
@@ -0,0 +1,165 @@
+"""Common Upgrade test utilities """
+
+from wait_for import wait_for
+from fabric.api import execute, run
+from robottelo.vm import VirtualMachine, settings
+from upgrade.helpers.docker import docker_execute_command
+from robottelo.api.utils import call_entity_method_with_timeout, entities
+
+
+class CommonUpgradeUtility(object):
```
mirzal: I challenge creating this class.

1. It contains many completely unrelated methods.
2. Most of the methods could be decorated with `@staticmethod`. They are self-contained, they don't take any data from class instance (in fact, they don't care if class was instantiated at all), they don't modify instance or class.
3. The only argument it takes during instantiation is optional, and used only by three methods.

I believe this file should just contain a loose collection of functions. Then tests can import functions they need and call them directly.

robottelo/upgrade_utility.py
```diff
@@ -0,0 +1,165 @@
...
+    """ CommonUpgradeUtility class for upgrade test cases """
+
+    def __init__(self, client_container_id=None):
+        self.client_container_id = client_container_id
```
mirzal: `client_container_id` might have made sense in first function that used code that is now refactored into `run_goferd`, but I believe it would be much better to call it `client_hostname`. After all, this is what this string represents - FQDN of host that we work with (execute commands, install software etc.).

robottelo/upgrade_utility.py
```diff
@@ -0,0 +1,165 @@
...
+            vm = VirtualMachine(
+                hostname=hostname,
+                target_image=hostname,
+                provisioning_server=provisioning_server,
+                distro=distro,
+            )
+            vm._created = True
+            vm.destroy()
+
+    def assertIn(self, member, container):
```
mirzal: unittest already provides that function. The fact that you need to copy it to your class is sign of deeper problems with this approach, for me.

Either force tests to run these assertions themselves, or just use `assert` in this file.

devendra104: Agree, I think no need of assertIn, We can evaluate it by simple assert statement.

robottelo/upgrade_utility.py
devendra104: Right, fixed.

robottelo/upgrade_utility.py
devendra104: 1. Yes It contains many unrelated methods but all of them are part of the upgrade, that's the reason I kept them at a single place.
2. Acknowledged.

robottelo/cleanup.py
vijay8451: @devendra104 `cleanup_of_provisioned_server` is more a common utils and could be used by tiers tests as well .. hence would not suggests to move it from here ..

devendra104: Agree, But the current scope of this module was only upgrade tests, that was the only reason for keeping this method/function in upgrade_utility. 
Anyway, I rollback this change.

robottelo/upgrade_utility.py
mirzal: > 1. Yes It contains many unrelated methods but all of them are part of the upgrade, that's the reason I kept them at a single place.

And let them be in single place - `robottelo.upgrade_utility` module. But there is no point in having them in class if you are not getting any benefit out of them being in the class.

devendra104: > > 1. Yes It contains many unrelated methods but all of them are part of the upgrade, that's the reason I kept them at a single place.
> 
> And let them be in single place - `robottelo.upgrade_utility` module. But there is no point in having them in class if you are not getting any benefit out of them being in the class.

You are right Mirek,  I know there is no short term benefit of this class but in the future, if re-utilization requires then it will be useful.
what do you suggest?
As we know most of the methods are completely unrelated to each other, Can I separate them in a different file(function base/class(static) base) and make it more meaningful?

mirzal: > You are right Mirek, I know there is no short term benefit of this class but in the future, if re-utilization requires then it will be useful.

Do you have anything specific on your mind? Some scenario where it would be helpful to group these functions into a class?

Also: if that scenario requires new functions that can well work together in a class, we can just put a class into this module. There is no rule that every module should have a class, or exactly one class, or that functions can't be put into module directly etc.

> what do you suggest?

I suggest that this class is removed and all methods are stored as plain functions. Tests should import functions they need.
We can add classes later if they are beneficial.

> As we know most of the methods are completely unrelated to each other, Can I separate them in a different file(function base/class(static) base) and make it more meaningful?

Thinking about this, some of the methods that you have here probably implement patterns used in other (non-upgrade) tests, and it would make sense to move functions to `helpers` or somewhere else, and port other tests to use them. But that's probably more than you wanted to tackle in this PR, so let's just keep everything in `upgrade_utility` module for now.

devendra104: > > You are right Mirek, I know there is no short term benefit of this class but in the future, if re-utilization requires then it will be useful.
> 
> Do you have anything specific on your mind? Some scenario where it would be helpful to group these functions into a class?
> 
> Also: if that scenario requires new functions that can well work together in a class, we can just put a class into this module. There is no rule that every module should have a class, or exactly one class, or that functions can't be put into module directly etc.

Not yet Mirek, Anyway If it will come then amend it with proper discussion.

Thanks Mirek for valuable comment.

mirzal: Thanks for implementing these changes!

devendra104:
Test Results:

PASSED2019-09-21 05:48:32 - robottelo - INFO - Started tearDownClass: tests.upgrades.test_errata/Scenario_errata_count
PASSED2019-09-21 04:59:15 - robottelo - INFO - Started tearDownClass: tests.upgrades.test_capsule/Scenario_capsule_sync
PASSED2019-09-21 05:20:40 - robottelo - INFO - Started tearDownClass: tests.upgrades.test_containers_removal/Scenario_containers_support_removal
PASSED2019-09-21 06:18:22 - robottelo - INFO - Started tearDownClass: tests.upgrades.test_repository/Scenario_repository_upstream_authorization_check
FAILED tests/upgrades/test_repository.py::Scenario_custom_repo_check::test_pre_scenario_custom_repo_check (Due to different reason)
PASSED2019-09-21 06:26:50 - robottelo - INFO - Started tearDownClass: tests.upgrades.test_subscription/Scenario_manifest_refresh
PASSED2019-09-21 06:30:29 - robottelo - INFO - Started tearDownClass: tests.upgrades.test_subscription/Scenario_contenthost_subscription_autoattach_check
PASSED2019-09-21 06:40:50 - robottelo - INFO - Started tearDownClass: tests.upgrades.test_yum_plugins/Scenario_yum_plugins_count

vijay8451:
@devendra104 GOOD WORK ... thanks to share the tests results .

PR #7350, opened 2019-09-16
Added SRPMs APIs tests

vijay8451:
- Added SRPMs APIs  tests.

### DIFF
--- a/robottelo/constants.py
+++ b/robottelo/constants.py
@@ -149,7 +149,7 @@ ZOO_CUSTOM_GPG_KEY = "zoo_custom_gpgkey.txt"
 VALID_GPG_KEY_BETA_FILE = "valid_gpg_key_beta.txt"
 
 RPM_TO_UPLOAD = "which-2.19-6.el6.x86_64.rpm"
-SRPM_TO_UPLOAD = "which-2.19-6.el6.x86_64.src.rpm"
+SRPM_TO_UPLOAD = "which-2.19-6.el6.src.rpm"
 
 ENVIRONMENT = "Library"
 
--- a/tests/foreman/api/test_repository.py
+++ b/tests/foreman/api/test_repository.py
@@ -797,21 +797,50 @@ class RepositoryTestCase(APITestCase):
 
     @skip_if_bug_open('bugzilla', 1378442)
     @tier1
-    def test_positive_upload_contents_srpm(self):
-        """Create a repository and upload SRPM contents.
+    @upgrade
+    def test_positive_upload_delete_srpm(self):
+        """Create a repository and upload, delete SRPM contents.
 
         :id: e091a725-048f-44ca-90cc-c016c450ced9
 
-        :expectedresults: The repository's contents include one SRPM.
+        :expectedresults: The repository's contents include one SRPM and delete after that.
 
         :CaseImportance: Critical
         """
-        # Create a repository and upload source RPM content.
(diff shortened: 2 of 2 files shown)


vijay8451:
Nailgun PR => https://github.com/SatelliteQE/nailgun/pull/662

robottelo/constants.py
```diff
@@ -149,7 +149,7 @@
 VALID_GPG_KEY_BETA_FILE = "valid_gpg_key_beta.txt"
 
 RPM_TO_UPLOAD = "which-2.19-6.el6.x86_64.rpm"
-SRPM_TO_UPLOAD = "which-2.19-6.el6.x86_64.src.rpm"
+SRPM_TO_UPLOAD = "which-2.19-6.el6.src.rpm"
```
mirzal: This constant is also used by `tests/foreman/cli/test_repository.py::test_positive_upload_content_srpm`. Is this change safe for that test?

tests/foreman/api/test_repository.py
```diff
@@ -1738,81 +1767,49 @@ class SRPMRepositoryTestCase(APITestCase):
...
-                self.product.label,
-                repo.label,
-            )
-        )
-        self.assertEqual(result.return_code, 0)
-        self.assertGreaterEqual(len(result.stdout), 1)
+        self.assertEqual(cv.repository[0].read().content_counts['srpm'], 3)
+        self.assertEqual(len(entities.Srpms().search(query={'organization_id': self.org.id})), 3)
+
+        cv = cv.read()
```
mirzal: Shouldn't that be called after `cv.publish()`, before assertions?

Also, usually we re-assign new data back with update:
```python
cv = cv.update(['repository'])
```
So maybe drop this line completely. I think publishing doesn't change anything in CV object.

robottelo/constants.py
vijay8451: Yeah it's safe as name was incorrect at all  :)

tests/foreman/api/test_repository.py
vijay8451: 1) It made intentionally to call it if below assertion pass else not require to call further and test would be fail:
```
self.assertEqual(cv.repository[0].read().content_counts['srpm'], 3)
self.assertEqual(len(entities.Srpms().search(query={'organization_id': self.org.id})), 3)
``` 
And same way promote is also done (to save test time and made assertion more aligned with action only instead of wait at the end)

2)  the end goal to read is (cv.version) and that only getting with `cv.read()` ... I tried below way but it only comes with same:
 
```
(Pdb) cv = cv.update(['repository'])
(Pdb) cv.publish()
(Pdb) cv.version
[]
(Pdb) cv = cv.read()
(Pdb) cv.version
[nailgun.entities.ContentViewVersion(id=21)]
(Pdb) 
```

mirzal: Ad 2: ah I see. Agreed.
Ad 1: For me, it felt unusual that you update the object, do not assign it back to variable and run assertions on old version, not newest one you received from server. But more I think about it, more I realize improvements I could propose are superficial.

Let's leave it as it is. Thanks for explanations.

jyejare:
ok to test : tests/foreman/api/test_puppetmodule.py tests/foreman/api/test_repository_set.py tests/foreman/api/test_errata.py

vijay8451:
Dependencies PR is merged now .. Good to merge this one.

Issue #7273, opened 2019-08-13
some ui testing for tasks


pondrejk:
Some checking of monitor > tasks page is missing in our test suite. I'd say appropriate testing can be accomplished without adding a new test class, just by adding asserts checking the tasks page to existing tests that already trigger tasks, e.g. remote execution or syncplan tests. Some supporting airgun infrastructure should be created accordingly.

PR #7363, opened 2019-09-24
add test for tasks dashboard

xdragu01:
waiting on SatelliteQE/airgun/pull/399
resolving: #7273 
test result:
```
tests/foreman/ui/test_dashboard.py .                                                                                                                                                                                                   [100%]

================================================================================================== 1 passed, 4 deselected in 144.64 seconds ==================================================================================================

```

### DIFF
--- a/tests/foreman/ui/test_dashboard.py
+++ b/tests/foreman/ui/test_dashboard.py
@@ -153,7 +153,8 @@ def test_positive_host_configuration_chart(session):
 @run_in_one_thread
 @tier2
 def test_positive_task_status(session):
-    """Check if the Task Status is working in the Dashboard UI
+    """Check if the Task Status is working in the Dashboard UI and
+        filter from Tasks dashboard is working correctly
 
     :id: fb667d6a-7255-4341-9f79-2f03d19e8e0f
 
@@ -161,10 +162,15 @@ def test_positive_task_status(session):
 
         1. Navigate to Monitor -> Dashboard
         2. Review the Latest Warning/Error Tasks widget
-        3. Review the Task Status widget
-        4. Click few links from the widget
+        3. Review the Running Chart widget
+        4. Review the Task Status widget
+        5. Review the Stopped Chart widget
+        6. Click few links from the widget
 
-    :expectedresults: Each link shows the right info
+    :expectedresults: Each link shows the right info and filter can be set
+        from Tasks dashboard
+
+    :BZ: 1718889
 
     :CaseLevel: Integration
     """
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_dashboard.py
```diff
@@ -180,13 +182,22 @@ def test_positive_task_status(session):
...
         })
-        tasks = session.task.read_all()
-        assert tasks['searchbox'] == 'state=running&result=pending'
+        searchbox = session.task.read_widget('searchbox')
+        assert searchbox == 'state=running&result=pending'
+        session.task.set_chart_filter('RunningChart')
+        row_count = session.task.table_row_count()
+        chart = session.task.read_widget('RunningChart')
+        assert row_count == int(chart['total']['Total'])
         session.dashboard.action({
```
omkarkhatavkar: Do you think we need to add one more assertion to see that we are navigated to latest result of task status? There could be a possibility that we could have multiple entries of 'TaskStatus': {'state': 'running', 'result': 'pending'}(Inregression). We should make sure that the we are navigated the to the task that we created as part of the test

tests/foreman/ui/test_dashboard.py
```diff
@@ -180,13 +182,22 @@ def test_positive_task_status(session):
...
             'TaskStatus': {'state': 'stopped', 'result': 'warning'}
         })
         tasks = session.task.read_all()
         assert tasks['searchbox'] == 'state=stopped&result=warning'
+        session.task.set_chart_filter(
+            'StoppedChart', {'row': 1, 'focus': 'Total'})
+        row_count = session.task.table_row_count()
+        chart = session.task.read_widget('StoppedChart')
+        assert row_count == int(chart['table'][1]['Total'])
```
omkarkhatavkar: Do you think we should also validate the count present on dashboard?

tests/foreman/ui/test_dashboard.py
xdragu01: If I understand correctly, there is still only one row per 'TaskStatus': {'state': 'running', 'result': 'pending'} in 'Task Status' table in dashboard. So clicking in table gets you to the list of tasks. There is no change of multiple entries.

tests/foreman/ui/test_dashboard.py
xdragu01: This could be a bit problem when you run tests in more threads. By the time test read whole dashboard and navigate to 'Tasks' this count could not be same.

tests/foreman/ui/test_dashboard.py
omkarkhatavkar: Yes if we are running tests in threads there could be a similar task is running and pending status?

xdragu01: Yes, there could be a similar task, but this part of test validate just filtering and if number shown in Tasks dashboard is same as number of tasks in table. Maybe I am missing something, but you don't care if there are more similar tasks.

xdragu01:
Adding new test results due to some changes in airgun
```
tests/foreman/ui/test_dashboard.py .                                                                                                                                                                                                   [100%]

================================================================================================== 1 passed, 4 deselected in 144.48 seconds ==================================================================================================
```

PR #7369, opened 2019-09-25
added test around 'Recommended Repositories'

vijay8451:
- Added test around 'Recommended Repositories'
- Test result
```
tests/foreman/ui/test_repository.py::test_positive_recommended_repos PASSED                                                                                      [100%]

============================================================== 1 passed, 20 deselected in 149.54 seconds ===============================================================
```

### DIFF
--- a/tests/foreman/ui/test_repository.py
+++ b/tests/foreman/ui/test_repository.py
@@ -41,6 +41,7 @@ from robottelo.constants import (
     INVALID_URL,
     REPO_DISCOVERY_URL,
     REPO_TYPE,
+    REPOSET,
     VALID_GPG_KEY_FILE,
     VALID_GPG_KEY_BETA_FILE,
 )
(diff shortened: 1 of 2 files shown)


tests/foreman/ui/test_repository.py
```diff
@@ -864,3 +864,24 @@ def test_positive_delete_random_docker_repo(session, module_org):
...
+
+    :expectedresults: Shows repositories as per On/Off 'Recommended Repositories'.
+
+    :CaseLevel: Integration
+    """
+    manifests.upload_manifest_locked(module_org.id)
+    with session:
+        session.organization.select(module_org.name)
+        rrepos_on = session.redhatrepository.read(recommended_repo='on')
+        assert rrepos_on
```
mirzal: Can we make better assertions than that? "recommended returns some data, and non-recommended returns some data, and second data has more items than first" isn't particularly strong.

In `robottelo/constants.py` we have `REPOS` and `REPOSET` variables. They look similar to what we can see in red hat repositories. Perhaps we can use them and verify that they appear in both recommended and non-recommended? Or one specific of them is in list? Or random one is in lists?

vijay8451: >> Can we make better assertions than that? 

It was not added because that does change from version to version .. (some repos are available under recommended repos for version 6.5 and some for 6.6).

It's  added now to keep a data check around it .

>> recommended returns some data, and non-recommended returns some data, and second data has more items than first" isn't particularly strong.

As checked the same over version 6.3/4/5/6 it found same (second data has more items than first) behaviour.

mirzal: Thank you, Vijay

PR #7370, opened 2019-09-26
Fix broken test test_positive_access_with_non_admin_user_without_mani…

JacobCallahan:
…fest

The test was failing because it encountered a NoSuchElement exception, since
the page renders differently with no manifest.
I put the line in question under a pytest.raises context manager.
Expected results are that this would always raise an exception.

Test is passing in our environment.
platform linux -- Python 3.6.6, pytest-4.6.3, py-1.8.0, pluggy-0.13.0 -- /home/[*******]/shiningpanda/jobs/ad0ef6b5/virtualenvs/d41d8cd9/bin/python3.6
cachedir: .pytest_cache
rootdir: /home/[*******]/workspace/satellite6-standalone-automation
plugins: xdist-1.25.0, services-1.3.1, forked-1.0.2, mock-1.10.4, env-0.6.2
collecting ... 2019-09-25 20:51:55 - conftest - DEBUG - Collected 6 test cases

tests/foreman/ui/test_subscription.py::test_positive_access_with_non_admin_user_without_manifest PASSED [100%]

- generated xml file: /home/[*******]/workspace/satellite6-standalone-automation/foreman-results.xml -
=================== 1 passed, 5 deselected in 88.02 seconds ====================

### DIFF
--- a/tests/foreman/ui/test_subscription.py
+++ b/tests/foreman/ui/test_subscription.py
@@ -19,7 +19,7 @@ from tempfile import mkstemp
 from airgun.session import Session
 from fauxfactory import gen_string
 from nailgun import entities
-from pytest import skip
+from pytest import raises, skip
 
 from robottelo import manifests
 from robottelo.api.utils import create_role_permissions
@@ -45,6 +45,7 @@ from robottelo.decorators import (
 )
 from robottelo.products import RepositoryCollection, RHELAnsibleEngineRepository
 from robottelo.vm import VirtualMachine
+from widgetastic.browser import NoSuchElementException
 import time
 pytestmark = [run_in_one_thread]
 
@@ -147,7 +148,9 @@ def test_positive_access_with_non_admin_user_without_manifest(test_name):
     ).create()
     with Session(test_name, user=user.login, password=user_password) as session:
         assert not session.subscription.search('')
-        assert not session.subscription.has_manifest
+        with raises(NoSuchElementException) as no_elem_err:
+            has_manifest = session.subscription.has_manifest or True
+            assert not has_manifest or no_elem_err.type is NoSuchElementException
 
 
 @tier2


JacobCallahan:
@mirzal if you wouldn't mind implementing that on the Airgun side, then I'm all about it!

mirzal:
@JacobCallahan you were right, this button is not displayed at all for users with limited permissions. I absently tried this as admin before.

Patch on Airgun side is submitted in https://github.com/SatelliteQE/airgun/pull/402

I'm not sure if that `session.subscription.has_manifest` should be called in this test at all. Theoretically, it doesn't make much sense - we just created this organization and have not uploaded manifest, we **know** it doesn't have manifest. We don't need to check that.

Practically, `has_manifest` uses "Add subscription" button state as proxy for manifest presence. Since user with "viewer" role doesn't see that button (probably rightfully so), that proxy now fails. It seems that there is no reliable way for user with limited permissions to verify if manifest was uploaded or not.

Anyway, with proposed Airgun patch, this test passes without further modifications. I'm leaving it up to you to decide if you want to remove reference to `has_manifest` or not.

vijay8451:
@JacobCallahan  
- As this test was passing till snap 16.0 and started failing after that so is this a bug or a new change ?

JacobCallahan:
@vijay8451 The current behavior is more correct, in my opinion. I don't recall exactly what that page looked like with limited permissions before, though.

@mirekdlugosz The limited permissions can tell a manifest isn't uploaded because they will still be presented with the "Import a Manifest", which then pops up the import modal. However, since they don't have permission to upload, they just see this.
Screenshot from 2019-10-01 09-19-07

I would like to keep the test as-is in the automation, in case there is a regression in rbac. Closing this PR out, since it is now handled on the Airgun side. Thanks, @mirekdlugosz!

PR #7381, opened 2019-09-30
separate organization and default manifest for rename tests

pondrejk:
The manifest refresh after satellite rename, that has been a subject to consistent timeouts and has low importance in the context of the whole scenario

### DIFF
--- a/tests/foreman/sys/test_rename.py
+++ b/tests/foreman/sys/test_rename.py
@@ -18,6 +18,8 @@
 """
 from fauxfactory import gen_string
 from nailgun import entities
+from robottelo import manifests
+from robottelo.api.utils import upload_manifest
 from robottelo.config import settings
 from robottelo.constants import DEFAULT_ORG
 from robottelo.decorators import (
@@ -47,7 +49,9 @@ class RenameHostTestCase(TestCase):
         cls.password = settings.server.admin_password
         cls.default_org_id = entities.Organization().search(
             query={'search': 'name="{}"'.format(DEFAULT_ORG)})[0].id
-        cls.org = entities.Organization(id=cls.default_org_id)
+        cls.org = entities.Organization().create()
+        with manifests.original_manifest() as manifest:
+            upload_manifest(cls.org.id, manifest.content)
         cls.product = entities.Product(organization=cls.org).create()
 
     @run_in_one_thread


mirekdlugosz:
Do we know the cause of "consistent timeouts"? Shouldn't they be reported and fixed? Or was there bugzilla already and devs/PM just don't care?

Can you explain why you say manifest refresh has "low importance"? Refreshing manifest seems pretty important to me, with or without host rename.

ntkathole:
@pondrejk can you please check comment by Mirek?

jyejare:
+1 to what @mirekdlugosz and @ntkathole have said about removing refresh from assertions. The manifest refresh should be closely tied up with the satellite hostname. and hence it should be a part of this test. I would like to see this PR closed if we consider the loss.

pondrejk:
Hi guys, sorry for longer silence. Refreshing manifest is tested in subscriptions and organizations tests. Low importance *in the context of this scenario* means that medium and repository used here are not dependent on the availability of subscriptions from the manifest, hence it appeared to me it is not worth to let the test wait (after successful rename) on something it doesn't need. But I guess opinions here may vary.

pondrejk:
@ntkathole, @jyejare, @JacobCallahan hi guys, can we go for this change?

jyejare:
> Hi guys, sorry for longer silence. Refreshing manifest is tested in subscriptions and organizations tests. Low importance _in the context of this scenario_ means that medium and repository used here are not dependent on the availability of subscriptions from the manifest, hence it appeared to me it is not worth to let the test wait (after successful rename) on something it doesn't need. But I guess opinions here may vary.

Hello Peter, considering the repository used here isn't dependent on the subscriptions, I would ACK this PR. But we should also consider adding the 'new test' which checks for 'the refresh post rename' and as there is no such test exists in org and subscriptions test files.

JacobCallahan:
I'm on the same page as @jyejare. I'd prefer to have the new test in place or merge ready before merging this.

rplevka:
I checked the the failed test logs and found out that the refresh doesn't really time out -
the tas kcompletes with warning:
```
'errors': ['Upstream identity certificate not available']
```

This looks like you're using a cloned manifest, whereas you should use the original one (portal has no knowledge of the fake IDs of our cloned manifests).
- you seems to use the `Default Organization` for the tests while the setupClass doesn't import the manifest - it looks like there is other test from completely different class inside the same tier, which interacts with the Default Org and imports the cloned manifest to it.
I think if you create your custom org in the setupClass and import an original manifest to it and then use that org across your rename tests, you should be fine and the refresh should pass.

pondrejk:
well, be it then 

```
pytest tests/foreman/sys/test_rename.py -k test_positive_rename_satellite
                                                

tests/foreman/sys/test_rename.py .                                                                             [100%]

====================================== 1 passed, 4 deselected in 993.80 seconds ======================================
```

PR #7391, opened 2019-10-03
added test case for bugzilla 1745575

omkarkhatavkar:
### Test Result 
```
Testing started at 4:20 PM ...
/home/okhatavk/Satellite/robottelo/env/bin/python /home/okhatavk/Downloads/pycharm-community-2018.1.4/helpers/pydev/pydevd.py --multiproc --qt-support=auto --client 127.0.0.1 --port 42534 --file /home/okhatavk/Downloads/pycharm-community-2018.1.4/helpers/pycharm/_jb_pytest_runner.py --target test_product.py::ProductTestCase.test_product_list_with_default_settings
pydev debugger: process 21959 is connecting

Connected to pydev debugger (build 181.5087.37)
Launching py.test with arguments test_product.py::ProductTestCase::test_product_list_with_default_settings in /home/okhatavk/Satellite/robottelo/tests/foreman/cli

2019-10-03 10:50:43 - conftest - DEBUG - Registering custom pytest_configure

platform linux -- Python 3.4.8, pytest-4.0.2, py-1.7.0, pluggy-0.9.0
plugins: xdist-1.26.1, services-1.3.1, mock-1.10.0, forked-1.0.2, cov-2.6.12019-10-03 10:50:44 - conftest - DEBUG - BZ deselect is disabled in settings

collected 1 item
==================== 1 passed, 2 warnings in 392.82 seconds ==================== 
```

### DIFF
--- a/tests/foreman/cli/test_product.py
+++ b/tests/foreman/cli/test_product.py
@@ -17,7 +17,9 @@
 """
 
 from fauxfactory import gen_alphanumeric, gen_string
+from robottelo import ssh
 from robottelo.cli.base import CLIReturnCodeError
+from robottelo.cli.defaults import Defaults
 from robottelo.cli.factory import (
     CLIFactoryError,
     make_gpg_key,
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_product.py
```diff
@@ -190,3 +192,62 @@ def test_negative_create_with_label(self):
...
+            u'param-name': 'organization_id',
+            u'param-value': default_org['id'],
+        })
+        result = ssh.command('hammer defaults list')
+        self.assertTrue(default_org['id'] in "".join(result.stdout))
+        try:
+            # Verify --organization-id is not required to pass if defaults are set
+            result = ssh.command('hammer product list')
+            self.assertTrue(default_product_name in "".join(result.stdout))
+            result = ssh.command('hammer repository list')
```
vijay8451: Do we require to verify this as part of BZ ?

tests/foreman/cli/test_product.py
```diff
@@ -190,3 +192,62 @@ def test_negative_create_with_label(self):
...
+            self.assertTrue(default_product_name in "".join(result.stdout))
+
+            # verify that defaults setting should not affect other entities
+            product_list = Product.list({u'organization-id': non_default_org['id']})
+            self.assertEquals(non_default_product_name, product_list[0]['name'])
+            repository_list = Repository.list({u'organization-id': non_default_org['id']})
```
vijay8451: same as above..

tests/foreman/cli/test_product.py
```diff
@@ -190,3 +192,62 @@ def test_negative_create_with_label(self):
...
+            self.assertEquals(non_default_product_name, repository_list[0]['product'])
+
+        finally:
+            Defaults.delete({u'param-name': 'organization_id'})
+            result = ssh.command('hammer defaults list')
```
vijay8451: same as above ..

tests/foreman/cli/test_product.py
omkarkhatavkar: This bug has been fixed in 6.6 satellite beta version, I have added this scenario to add coverage in test automation.

tests/foreman/cli/test_product.py
omkarkhatavkar: I think we don't have a test case where we can leverage the default settings, don't pass organization id, as it is taken care by default setting.  It is a good candidate.

tests/foreman/cli/test_product.py
vijay8451: thanks @omkarkhatavkar ..let's keep it here.

tests/foreman/cli/test_product.py
```diff
@@ -190,3 +192,62 @@ def test_negative_create_with_label(self):
...
+
+        :BZ: 1745575
+
+        :expectedresults: product/reporsitory list should work as expected.
+
+        """
+        default_product_name = gen_string('alpha')
+        non_default_product_name = gen_string('alpha')
+        default_org = make_org()
+        non_default_org = make_org()
```
pondrejk: I wonder if we really need to create any organizations in this test. One org is created in the setup and there's also Default org so that is two. I'd say reusing those would speed up the test execution

tests/foreman/cli/test_product.py
```diff
@@ -190,3 +192,62 @@ def test_negative_create_with_label(self):
...
+        for product in (default_product, non_default_product):
+            make_repository({
+                'product-id': product['id'],
+                'url': FAKE_0_YUM_REPO,
+            })
+            Product.synchronize({
+                'id': product['id'],
+            })
+            packages = Package.list({'product-id': product['id']})
+            self.assertEqual(len(packages), FAKE_0_YUM_REPO_PACKAGES_COUNT)
```
pondrejk: I'm not sure why we are synchronizing and checking packages. The bz and the following asserts are just about listing the prods and repos. So just creating the entities should do, potential for speedup. Syncing is well tested elsewhere

omkarkhatavkar: I think no customer will only just create the repository, they will sync those repos ultimately. I'm just trying to think like the customer while designing this scenario.

tests/foreman/cli/test_product.py
omkarkhatavkar: The reason I created separate or is that there are multiple other products are already created in that org. I wanted to avoid that, even if we add the default or to module org won't make the problem. let me test that gets back to you

omkarkhatavkar:
@SatelliteQE/quality-engineers Can someone please take a look at this PR?

PR #7396, opened 2019-10-08
Add test related to kickstart encode error

kersommoura:
Add test to verify encode gzip error for kickstart repositories.

### DIFF
--- a/tests/foreman/api/test_contentmanagement.py
+++ b/tests/foreman/api/test_contentmanagement.py
@@ -28,6 +28,7 @@ from robottelo.api.utils import (
 )
 from robottelo.config import settings
 from robottelo.constants import (
+    CUSTOM_KICKSTART_REPO,
     CUSTOM_PUPPET_REPO,
     DISTRO_RHEL7,
     ENVIRONMENT,
@@ -122,6 +123,51 @@ class ContentManagementTestCase(APITestCase):
         ).create()
         with self.assertNotRaises(TaskFailedError):
             repo.sync()
+
+    @tier4
+    def test_positive_sync_kickstart_repo(self):
+        """No encoding gzip errors on kickstart repositories
+        sync.
+
+        :id: dbdabc0e-583c-4186-981a-a02844f90412
+
+        :expectedresults: No encoding gzip errors present in /var/log/messages.
+
+        :CaseLevel: Integration
+
+        :steps:
+
+            1. Sync a kickstart repository.
+            2. After the repo is synced, change the download policy to
+                immediate.
+            3. Sync the repository again.
+            4. Assert that no errors related to encoding gzip are present in
+                /var/log/messages.
+            5. Assert that sync was executed properly.
+
+        :CaseComponent: Pulp
+
+        :BZ: 1687801
+        """
+        org = entities.Organization().create()
+        product = entities.Product(organization=org).create()
+        repo = entities.Repository(
+            product=product,
+            url=CUSTOM_KICKSTART_REPO
+        ).create()
+        repo.sync()
+        repo.download_policy = 'immediate'
+        repo = repo.update(['download_policy'])
+        call_entity_method_with_timeout(repo.sync, timeout=600)
+        result = ssh.command(
+            'grep pulp /var/log/messages | grep failed | grep encoding | grep gzip'
+        )
+        self.assertEqual(result.return_code, 1)
+        self.assertEqual(len(result.stdout), 0)
+        repo = repo.read()
+        self.assertGreater(repo.content_counts['package'], 0)
+        self.assertGreater(repo.content_counts['package_group'], 0)
+        self.assertGreater(repo.content_counts['rpm'], 0)
 
 
 @run_in_one_thread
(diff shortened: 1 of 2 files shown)


tests/foreman/api/test_contentmanagement.py
```diff
@@ -125,6 +126,48 @@ def test_positive_sync_repos_with_lots_files(self):
...
+        :BZ: 1687801
+        """
+        org = entities.Organization().create()
+        product = entities.Product(organization=org).create()
+        repo = entities.Repository(
+            product=product,
+            url=CUSTOM_KICKSTART_REPO
+        ).create()
+        repo.sync()
+        repo = repo.read()
```
mirekdlugosz: Do you need that `read`? You call `update` and re-assign variable just two lines below.

tests/foreman/api/test_contentmanagement.py
```diff
@@ -125,6 +126,48 @@ def test_positive_sync_repos_with_lots_files(self):
...
+        repo.download_policy = 'immediate'
+        repo = repo.update(['download_policy'])
+        repo.sync()
+        results = [
+            ssh.command('cat /var/log/messages | grep {}'.format(key))
```
mirekdlugosz:
```suggestion
            ssh.command('grep {} /var/log/messages'.format(key))
```

tests/foreman/api/test_contentmanagement.py
```diff
@@ -125,6 +126,48 @@ def test_positive_sync_repos_with_lots_files(self):
...
+            for key in ('encoding', 'gzip', 'failed')
+        ]
+        self.assertTrue(all([result.return_code == 1 for result in results]))
+        self.assertTrue(all([len(result.stdout) == 0 for result in results]))
```
mirekdlugosz:
```python
        self.assertTrue(all(result.return_code == 1 for result in results))
        self.assertTrue(all(len(result.stdout) == 0 for result in results))
```

`(x for x in items)` is generator expression - it gives one value at the time. It's good for functions like `all` or `any`, as they can exit earlier, on first value that determines entire result (i.e. on first False for `all`). Of course it doesn't make any difference for 3-element lists.

But cool thing about generator expressions is that you can omit parentheses if they are the only argument to the function. So you can write `all(x for x in items)` instead of `all([x for x in items])`.

tests/foreman/api/test_contentmanagement.py
```diff
@@ -125,6 +126,48 @@ def test_positive_sync_repos_with_lots_files(self):
...
+            url=CUSTOM_KICKSTART_REPO
+        ).create()
+        repo.sync()
+        repo = repo.read()
+        repo.download_policy = 'immediate'
+        repo = repo.update(['download_policy'])
+        repo.sync()
+        results = [
+            ssh.command('cat /var/log/messages | grep {}'.format(key))
+            for key in ('encoding', 'gzip', 'failed')
```
mirekdlugosz: Shouldn't we use some more specific terms? I can imagine some completely unrelated component might fail, polluting messages with `failed`, or some other process may add message about `encoding`. Maybe grep for `pulp.*{key}`?

tests/foreman/api/test_contentmanagement.py
kersommoura: That is a great hint. Thanks

tests/foreman/api/test_contentmanagement.py
kersommoura: The error is related to the encoding of headers. I will work on improve those assertions.

jyejare:
@kersommoura , We are planning to merge 6.6 branch,  can you actively work on comments to merge this before we could branch 6.6. Thanks :)

kersommoura:
@jyejare, thanks, I will work on this one today.

kersommoura:
@rplevka, thanks for your review. The first used kickstart repo was created focused to help test the encoding error. No packages present there. The I did some research, I could not find test to sync kickstart repos. Based on your feedback, I decided to change the repository used.

kersommoura:
```shell
PASSED2019-10-30 17:21:25 - robottelo - DEBUG - Finished Test: ContentManagementTestCase/test_positive_sync_kickstart_repo
2019-10-30 17:21:25 - robottelo - INFO - Started tearDownClass: tests.foreman.api.test_contentmanagement/ContentManagementTestCase


========================================================================================================= 1 passed in 409.43 seconds =======================================================================================================
```

tests/foreman/api/test_contentmanagement.py
```diff
@@ -123,6 +124,51 @@ def test_positive_sync_repos_with_lots_files(self):
...
+        result = ssh.command(
+            'grep pulp /var/log/messages | grep failed | grep encoding | grep gzip'
+        )
+        self.assertEqual(result.return_code, 1)
+        self.assertEqual(len(result.stdout), 0)
+
+        repo = repo.read()
+        self.assertGreater(repo.content_counts['package'], 0)
+        self.assertGreater(repo.content_counts['package_group'], 0)
+        self.assertGreater(repo.content_counts['rpm'], 0)
```
ntkathole: flake8 not well 
`./tests/foreman/api/test_contentmanagement.py:173:1: E302 expected 2 blank lines, found 1 `

kersommoura: `make can-i-push` does not detect this. Updated.

PR #7401, opened 2019-10-09
Fix repo sync on multi env capsule

latran:
Added repo sync to docker repo


mirekdlugosz:
@latran If you mean this codecov report, it's safe to ignore. It complains about change in file that you have not modified in PR. Overall, value provided by codecov is questionable, but that's discussion for another time.

latran:
Tested and seems to be stable as is.  Thought I had some sycing issues, but turns out fine.

PR #7402, opened 2019-10-10
refactored UI subnet test cases

rplevka:
~~**DEPENDS ON**: https://github.com/SatelliteQE/nailgun/commit/e0c2fb2d69d7ca703f18213563e75c65ce793c9e being cherry-picked to `nailgun-stable` branch~~

merged 2 tests into one (they shared the same execution path),
rewritten the assertion part to use API.

original tests took: ~2mins 40secs
after refactoring:  ~52secs

```
$ pytest test_subnet.py -k end -v
2019-10-10 14:12:58 - conftest - DEBUG - Registering custom pytest_configure

======================================================================== test session starts =========================================================================
platform linux -- Python 3.6.8, pytest-4.6.3, py-1.8.0, pluggy-0.12.0 -- /home/rplevka/.virtualenvs/robottelo/bin/python3
cachedir: .pytest_cache
rootdir: /home/rplevka/work/rplevka/robottelo/tests/foreman, inifile: pytest.ini
collecting ... 2019-10-10 14:12:58 - conftest - DEBUG - BZ deselect is disabled in settings

collected 1 item                                                                                                                                                     

test_subnet.py::test_positive_end_to_end PASSED                                                                                                                [100%]

=== warnings summary ===
ui/test_subnet.py::test_positive_end_to_end
ui/test_subnet.py::test_positive_end_to_end
  /home/rplevka/.virtualenvs/robottelo/lib/python3.6/site-packages/widgetastic/browser.py:721: DeprecationWarning: use driver.switch_to.alert instead
    return self.selenium.switch_to_alert()

=== 1 passed, 2 warnings in 52.67 seconds ===
```

### DIFF
--- a/tests/foreman/ui/test_subnet.py
+++ b/tests/foreman/ui/test_subnet.py
@@ -30,34 +30,16 @@ def module_loc():
     return entities.Location().create()
 
 
-@tier2
-def test_positive_create_with_domain(session, module_org, module_loc):
-    """Create new subnet with domain in same organization
-
-    :id: adbc7189-b451-49df-aa10-2ae732832dfe
-
-    :expectedresults: Subnet is created with domain associated
-
-    :CaseLevel: Integration
-    """
-    name = gen_string('alpha')
-    domain = entities.Domain(
-        organization=[module_org], location=[module_loc]).create()
-    with session:
-        session.subnet.create({
-            'subnet.name': name,
-            'subnet.network_address': gen_ipaddr(ip3=True),
-            'subnet.network_prefix': 12,
-            'domains.resources.assigned': [domain.name]
-        })
-        subnet_values = session.subnet.read(name)
-        assert subnet_values[
-            'domains']['resources']['assigned'][0] == domain.name
+@fixture(scope='module')
+def module_dom(module_org, module_loc):
+    d = entities.Domain(organization=[module_org.id], location=[module_loc.id]).create()
+    yield d.read()
+    d.delete()
 
 
 @tier2
 @upgrade
-def test_positive_end_to_end(session):
+def test_positive_end_to_end(session, module_dom):
     """Perform end to end testing for subnet component in ipv6 network
 
     :id: f77031c9-2ca8-44db-8afa-d0212aeda540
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_subnet.py
```diff
@@ -81,19 +63,30 @@ def test_positive_end_to_end(session):
...
-        subnet_values = session.subnet.read(name)
-        assert subnet_values['subnet']['name'] == name
-        assert subnet_values['subnet']['description'] == description
-        assert subnet_values['subnet']['protocol'] == 'IPv6'
-        assert subnet_values['subnet']['network_address'] == network_address
-        assert subnet_values['subnet']['network_prefix'] == '24'
-        assert subnet_values['subnet']['ipam'] == 'EUI-64'
-        assert subnet_values['subnet']['mtu'] == '1600'
+        sn = entities.Subnet().search(query={'search': 'name={0}'.format(name)})
+        assert len(sn) > 0, 'Subnet {0} expected to exist, but it\'s not listed'
```
mirekdlugosz: If you have single-quote character inside string, please surround string with double-quote characters. Allows you to drop backslash.
Also, seems you are missing `.format()` method call on assertion failure message.

tests/foreman/ui/test_subnet.py
```diff
@@ -81,19 +63,30 @@ def test_positive_end_to_end(session):
...
+        sn = sn[0].read()
+        assert sn.description == description, 'Subnet description expected to be {0} \
+                but it\'s {1} instead'.format(description, sn.description)
```
mirekdlugosz:
```python
        assert sn.description == description, ("Subnet description expected to be {0} "
                "but it's {1} instead").format(description, sn.description)
```
The same applies to all assertions below.

vijay8451:
I missed that on earlier PR .. thanks @mirekdlugosz..
Yes the API assertion  is altogether a new way in UI tests .. I am also not sure about background on this one as well and would like to know about same.

tests/foreman/ui/test_subnet.py
rplevka: I'm following an ancient codestyle rule, where it was settled on using single quotes everywhere. I don't mind escaping a character here and there - is that so big issue?

rplevka: Will fix the missing `format()`

rplevka:
@mirekdlugosz no, this wasn't discussed anywhere.

rplevka:
@mirekdlugosz - what i can do here (and i probably should) is to bring back the C __R__ UD (Read) part of the test, so we actually test reading of the subnet info using UI.
I'm intending to do all the assertions using API.

rplevka:
The updated test takes
```
==================== 1 passed, 2 warnings in 89.64 seconds =====================
```
- I also updated the `read` method call to only load the widgets with data necessary for the assertions.

tests/foreman/ui/test_subnet.py
mirekdlugosz: That rule is justified as attempt to reduce visual noise. Backslash inside a string is a definition of visual noise, if you ask me.

But no, this is not big, blocking issue. More in would-be-nice-to-have and personal preference territory.

jyejare:
@rplevka , @JacobCallahan and @mirekdlugosz Guys can we merge this now, I dont see any dependency PR to merged now.

mirekdlugosz:
Well, it has "DO NOT MERGE" label. @rplevka is it waiting for something, or can we merge it already? I think you can remove label and merge youself, you have 3 ACKs already.

PR #7404, opened 2019-10-11
New Bugzilla skipping logic - Replacing Robozilla Issue #7336 

### LINKED ISSUE #7336: Remove Robozilla dependency on skip_if_bug_open decorator

rochacbruno:
Fix #7336 
Replaces PR: #7346 

#### Replaces Robozilla decorators with Pytest Markers and adds new logic for skipping and deselect.

This PR is separated in 5 commits. 
(when merging it can be squased in to one commit, or just rebased keeping 5 different commits)

1. Cleanup of redundant decorators from test cases https://github.com/SatelliteQE/robottelo/pull/7404/commits/23197ae9d3df38786643b33b19b15c3595c694eb
2. Fix of Robottelo Unit Tests for the decorators changes https://github.com/SatelliteQE/robottelo/pull/7404/commits/25e1fb0e8a9d12491ddf673010709cac21cb8074
3. Cleanup of redundant bz functions used within robottelo modules https://github.com/SatelliteQE/robottelo/pull/7404/commits/1e8b9df0725d8dbcda718b00b7f3467e7a20d120
4. Simple .gitignore and Makefile changes https://github.com/SatelliteQE/robottelo/pull/7404/commits/bc35716de207455d14fbf5d54fa56beee34ab582
5. Replaces Robozilla decorators with Pytest markers and adds new logic for `is_open` function https://github.com/SatelliteQE/robottelo/pull/7404/commits/6d4f0fc3de550b851bd2823294ec2acc31acf585

#### The commit n.5 is the most important:
**Replaces Robozilla decorators with Pytest markers and adds new logic for `is_open` function https://github.com/SatelliteQE/robottelo/pull/7404/commits/6d4f0fc3de550b851bd2823294ec2acc31acf585**
Its logic is based on flowcharts described here -> https://hackmd.io/01mIqiG6TOKTrPpTk4jejg?view

At the `files changed` tab reviewers can select which commit to review.

### Cleanups

I checked every BZ using the old `skip_if_bug_open` and also `bz_bug_is_open` function and cleaned up the code when the BZ status was evaluating the function always to True. (in other workds we were doing useless calls to Bz API)

### BZ REST API

> **IMPORTANT!!!**

The new implementation is using `httpx` module to call Bugzilla REST API, it introduces changes on `robottelo.properties` file, that will need to be changed to include `api_key` under `[bugzilla]` section.

So CI configs and also your local configs will need to be updated.

### The new skipping implementation

- Uses httpx to communicate with Robozilla REST API
- Authenticates using `api_key` (that you can generate in your BZ profile page)
- Makes a single call to get all the BZs data at the test collection phase.
- Uses `tenacity.retry` to perform retries if HTTP connection fails.
- Can bypass the api call if `--bz-cache` is passed to pytest then it will use pre-fetched data (useful for local testing).
- Using `pytest tests/foreman --collect-only` will create a new `bz_cache.json` on the root of folder so this file can be used as `--bz-cache` for future runs, and also can be used by the nagger script.
- Has simpler logic based on https://hackmd.io/01mIqiG6TOKTrPpTk4jejg?view

#### To skip a test case based on BZ status use:

```python
import pytest

@pytest.mark.skip_if_open('BZ:123456')
def test_that_will_skip_if_bz_is_not_yet_fixed_for_current_system_version(...)
    ...
```

#### To deselect a test case based on BZ status use:

```python
import pytest

@pytest.mark.skip_if_open('BZ:123456')
def test_that_will_be_deselected_from_testing_collection_if_bz_is_wontfix(...)
    # WONTFIX,CANTFIX,DEFERRED makes this test to be ignored by pytest runner
    ...
```

#### To use a `workaround` inside a test case

```python
from robottelo.helpers import is_open

def test_something(...):
    ...
    if is_open('BZ:213456'):
        # this block executes only if BZ is not yet fixed for current system version
```

### The nagger script

There will still be a nagger script, that will notify weekly, component owners to do cleanups of redundant marks, the nagger script will use the `bz_cache.json` generated on every robottelo run.

### TO-DOs

- [x] Send a PR to gitlab config adding new `[bugzilla]`  configs
- [ ] Send PR to robottelo-ci saving the `bz_cache.json` artifact on every build.

### How to test this PR?

- checkout to this branch 
  if you have `hub` so `hub pr checkout 7404`
- install the new dependencies `httpx` and `tenacity`
  `pip install -r requirements.txt`
- Change your `robottelo.properties` adding
   ```ini
    [bugzilla]
    url=https://bugzilla.redhat.com
    api_key=(generate your key on bugzilla profile page)
   ```
- Run `make test-robottelo`
- Run `py.test -s tests/foreman/ --collect-only` 
  No test will run, it will just collect data and generate the `bz_cache.json` take a look at this file :)
- Run `py.test -s tests/foreman/ --collect-only --bz-cache` 
  No test will run, it will just collect data use existing `bz_cache.json`
- Pick some test marked with `@pytest.mark.skip_if_open()` and run it separatelly to see if it works.


robottelo/cli/factory.py
```diff
@@ -884,8 +883,6 @@ def make_product_wait(options=None, wait_for=5):
     try:
         product = make_product(options)
     except CLIFactoryError as err:
-        if not bz_bug_is_open(1332650):
```
rochacbruno: A second look here,this would alwwys raise?

robottelo/datafactory.py
```diff
@@ -294,9 +290,8 @@ def valid_docker_repository_names():
              gen_string('latin1', random.randint(1, 255)),
              gen_string('numeric', random.randint(1, 255)),
              gen_string('utf8', random.randint(1, 85)),
+             gen_string('html', random.randint(1, 85))
              ]
-    if not bz_bug_is_open(1483622):
```
rochacbruno: + 1 to review

robottelo.properties.sample
```diff
@@ -25,6 +25,9 @@ ssh_key=
 # Admin password when accessing API and UI
 # admin_password=changeme
 
+# Version is taken from here OR SAT_VERSION envvar or SSH call
+# version=6.6
```
mirekdlugosz: Do we really need new setting?
I understand why you don't want to depend on version obtained from Satellite itself, but isn't environmental variable enough?

robottelo/helpers.py
```diff
@@ -6,7 +6,12 @@
 import random
 import re
 import requests
+import inspect
+import json
+import pytest
```
mirekdlugosz: Nitpick: I think that usually we sort modules alphabetically within a group.

robottelo/helpers.py
```diff
@@ -750,3 +749,190 @@ def host_provisioning_check(ip_addr):
...
+    Arguments:
+        issue {str} -- A string containing handler + number e.g: BZ:123465
+        data {dict} -- Issue data indexed by <handler>:<number> or None
+    """
+    # Handlers can be extended to support different issue trackers.
+    handlers = {'BZ': bugzilla.is_open_bz}
+    supported_handlers = tuple(handlers.keys())
+
+    if str(issue).startswith(supported_handlers):
+        handler_code, _, _ = str(issue).partition(":")
```
mirekdlugosz: Technically, this branch will execute for issue=`BZ1234`. `handler_code` will then be set to `issue` variable value, eventually throwing Key Error in last `return`.

Possibly you want slightly different definition of supported_handlers:
```python
supported_handlers = tuple(f"{handler}:" for handler in handlers.keys())
```

robottelo/issue_handlers/bugzilla.py
```diff
@@ -0,0 +1,259 @@
...
+        "creation_time",
+        "flags",
+        "keywords",
+        "dupe_of",
+        "target_milestone",
+        "cf_clone_of",
+        "clone_ids",
+        "depends_on",
+    ]
+    # Following fields are dinamically calculated/loaded
```
mirekdlugosz: typo: dinamically -> dynamically

requirements.txt
```diff
@@ -23,14 +23,9 @@ widgetastic.core==0.33
...
-# python-bugzilla 2.0.0 changed a lot of function signatures
-# also it removed some functions use e.g: getbugsimple
-# to update it code changes is needed.
-python-bugzilla==1.2.2  # pyup: ignore
-
 PyYAML==5.1.2
-robozilla==0.2.6
-
+tenacity==5.1.1
+httpx==0.7.5
```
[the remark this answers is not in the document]
rochacbruno: You are right, it could be `requests` I can change it, I just copied those 2 functions from the nagger script without changing. gonna do that.

robottelo/issue_handlers/bugzilla.py
```diff
@@ -0,0 +1,259 @@
...
+        bz['status'] in CLOSED_STATUSES
+        and bz['resolution'] in WONTFIX_RESOLUTIONS
+    ):
+        return True
+
+    # BZ is CLOSED with a resolution in (ERRATA, CURRENT_RELEASE, ...)
+    # server.version is higher or equal than BZ version
+    # Consider fixed,  BZ is not open
+    if float(settings.server.version) >= extract_min_version(bz):
+        return False
```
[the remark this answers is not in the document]
rochacbruno: this is other example taken from Robozilla and logic just moved to here, I agree with you, we can improve that.

Just for curiosity @JacobCallahan do you know if historically we had a 2 digit `y` stream?

Anyway we can improve that.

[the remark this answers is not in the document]
rochacbruno: @mirekdlugosz actually it is an easy change :)

```py
from packaging.version import Version
Version('6.12') > Version('6.20')
```

robottelo/vm_capsule.py
```diff
@@ -203,7 +206,7 @@ def _capsule_cleanup(self):
                 # exception has 'return_code=70(Error: host not found)'
                 if exp.return_code == 70:
                     super(CapsuleVirtualMachine, self).destroy()
-                if bz_bug_is_open('1622064'):
+                if is_open('BZ:1622064'):
```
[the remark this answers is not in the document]
mirekdlugosz: There is one place in constants where we do use f-strings. I could swear there are more than that.

Either way, I would say f-strings are safe to be used in robottelo. By the time we finally dropped Python2 support, Python 3.6 was already widely available.

robottelo/issue_handlers/bugzilla.py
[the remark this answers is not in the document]
mirekdlugosz: `packaging.version.Version` will work, too. Looks like we already pull it in through one of our dependencies.

tests/foreman/conftest.py
```diff
@@ -185,3 +172,32 @@ def pytest_collection_modifyitems(items, config):
...
+def pytest_configure(config):
+    """Register custom markers to avoid warnings."""
+    markers = [
+        "stubbed: Tests that are not automates yet.",
+        "deselect(reason=None): Mark test to be removed from collection.",
+        "skip_if_open(issue): Skip test based on issue status.",
+        "tier1: CRUD tests",
+        "tier2: Association tests",
+        "tier3: Systems integration tests",
+        "tier4: Long running tests",
```
mirekdlugosz: I know this is not a time and place for this conversation, but I just have to point out that not everyone in the team agrees with these definitions. They aren't aligned with QE org guidelines, either. Personally, I would prefer to leave this whole matter out of here and give tautological explanations, like `tier1: Tier 1 tests`.

tests/robottelo/test_helpers.py
```diff
@@ -133,3 +141,317 @@ def test_multiple_dicts_converted_to_storage(self):
...
+                "target_milestone": "Unspecified",
+                "flags": [{"status": "+", "name": "sat-7.7.z"}],
+                "clones": [
+                    {
+                        "id": 999999,
+                        "status": "POST",
+                        "resolution": "",
+                        "target_milestone": "Unspecified",
+                        "flags": [{"status": "+", "name": "sat-6.6.z"}],
+                        "clones": []
```
mirekdlugosz: Does bugzilla ensure `clones` to be symmetrical relationship?
In other words, is it guaranteed that if BZ 1 clones BZ 2, and BZ 2 clones BZ 3, then BZ 3 "clones" field will contain both BZ 1 and 2?

If not, then perhaps "clones" field handling should be done recursively or in the loop, to ensure all clones are found.

tests/robottelo/test_helpers.py
```diff
@@ -133,3 +141,317 @@ def test_multiple_dicts_converted_to_storage(self):
...
+                    }
+                ]
+            }
+        }
+        self.assertTrue(is_open("BZ:123456", data))
+
+    def test_bz_is_not_open_using_dupe_data_clones(self):
```
mirekdlugosz: Speaking of which, maybe we could use test case where clone has duplicate?

tests/robottelo/test_helpers.py
```diff
@@ -133,3 +141,317 @@ def test_multiple_dicts_converted_to_storage(self):
...
+                        "resolution": "",
+                        "target_milestone": "Unspecified",
+                        "flags": []
+                    }
+                    self.assertFalse(_should_deselect("BZ:123456", data))
+
+    def test_invalid_handler_for_is_open_raises_error(self):
+        """Assert is_open w/ invalid handlers raise AttributeError"""
+
+        for issue in ("XX:123456", "KK:89456", "123456", 999999):
```
mirekdlugosz: Please add "BZ1234" to this list.

tests/foreman/ui/test_bookmark.py
```diff
@@ -47,12 +47,7 @@ def ui_entities(module_org, module_loc):
         # Skip the entities, which can't be tested ATM (not implemented in
         # airgun or have open BZs)
         skip = entity.get('skip_for_ui')
-        if isinstance(skip, tuple):
-            bug_type, bug_id = skip
-            if (bug_type == 'bugzilla' and bz_bug_is_open(bug_id)
-                    or bug_type == 'redmine' and rm_bug_is_open(bug_id)):
-                skip = True
-        if skip is True:
+        if skip:
```
mirekdlugosz: There are two references to bugzilla in `BOOKMARK_ENTITIES`: bz#1638781 (CLOSED ERRATA) and bz#1387569 (CLOSED WONTFIX). According to current logic, one will return False for `is_open`, while other would return True.

I think we are incorrectly skipping `ContentCredential` here.

robottelo/issue_handlers/bugzilla.py
```diff
@@ -0,0 +1,259 @@
...
+# match any version as in `sat-6.2.x` or `sat-6.2.0` or `6.2.9`
+# The .version group being a `d.d` string that can be casted to float()
+VERSION_RE = re.compile(r'(?:sat-)*?(?P<version>\d\.\d)\.\w*')
+
+
+def is_open_bz(issue, data=None):
+    """Check if specific BZ is open consulting a cached `data` dict or
+    calling Bugzilla REST API.
+
+    https://hackmd.io/01mIqiG6TOKTrPpTk4jejg?view#Is-Bug-Open
```
mirekdlugosz: Going forward, can we have this information in `docs` instead of external resource?

tests/robottelo/test_helpers.py
rochacbruno: BUgzilla returns only a list `clone_ids` holding the direct clones, if the clone has a clone than you are right and we need to follow it.

robottelo/issue_handlers/template.py
```diff
@@ -0,0 +1,44 @@
+"""An issue handler must expose the following functions:
```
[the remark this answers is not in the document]
rochacbruno: I gonna change it to a README.md as it is a static file just as a dev doc for if/when we need to handle another issue tracker e.g: Jira

robottelo/issue_handlers/bugzilla.py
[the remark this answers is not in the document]
rochacbruno:
```
❯ pipdeptree -p packaging
packaging==19.1
  - attrs [required: Any, installed: 19.1.0]
  - pyparsing [required: >=2.0.2, installed: 2.4.2]
  - six [required: Any, installed: 1.12.0]

```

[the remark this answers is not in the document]
rochacbruno: And also nailgun

```
nailgun==0.32.0
  - blinker-herald [required: Any, installed: 0.2.0]
  - fauxfactory [required: Any, installed: 3.0.6]
  - inflection [required: Any, installed: 0.3.1]
  - packaging [required: Any, installed: 19.1]
    - attrs [required: Any, installed: 19.1.0]
    - pyparsing [required: >=2.0.2, installed: 2.4.2]
    - six [required: Any, installed: 1.12.0]
  - pyxdg [required: Any, installed: 0.26]
  - requests [required: >=2.7, installed: 2.22.0]
    - certifi [required: >=2017.4.17, installed: 2019.6.16]
    - chardet [required: >=3.0.2,<3.1.0, installed: 3.0.4]
    - idna [required: >=2.5,<2.9, installed: 2.8]
    - urllib3 [required: >=1.21.1,<1.26,!=1.25.1,!=1.25.0, installed: 1.25.3]
```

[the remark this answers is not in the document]
mirekdlugosz: I was looking for something like `pipdeptree` :) . Good to know.
I was a little worried this package might be removed from availablae packages without us noticing, but that's not a huge problem. And if that's direct dependency of nailgun, we have full control over it anyway.

tests/robottelo/test_helpers.py
rochacbruno: We will not follow clones anymore, @mirekdlugosz this responsibility is being deferred to the nagger script.

robottelo/issue_handlers/bugzilla.py
```diff
@@ -0,0 +1,258 @@
...
+        bz_numbers {list of str} -- ['123456', ...]
+        cached_data
+
+    Returns:
+        [list of dicts] -- [{'id':..., 'status':..., 'resolution': ...}]
+    """
+    if cached_data:
+        LOGGER.debug("Using cached data for {}".format(set(bz_numbers)))
+        if not all(
+            ['BZ:{}'.format(number) in cached_data for number in bz_numbers]
```
JacobCallahan: f-strings in this section too, please :)

robottelo/issue_handlers/bugzilla.py
```diff
@@ -0,0 +1,258 @@
...
+    # Following fields are dynamically calculated/loaded
+    for field in ('is_open', 'clones', 'version'):
+        assert field not in bz_fields
+
+    # Ensure API key is set
+    if not settings.bugzilla.api_key:
+        raise ImproperlyConfigured("Config file is missing bugzilla api_key")
+
+    response = requests.get(
+        "{}/rest/bug".format(settings.bugzilla.url),
```
JacobCallahan: another f-string possibility :)

robottelo/helpers.py
```diff
@@ -750,3 +751,218 @@ def host_provisioning_check(ip_addr):
...
+    Arguments:
+        issue {str} -- A string containing handler + number e.g: BZ:123465
+        data {dict} -- Issue data indexed by <handler>:<number> or None
+    """
+    # Handlers can be extended to support different issue trackers.
+    handlers = {'BZ': bugzilla.is_open_bz}
+    supported_handlers = tuple(f"{handler}:" for handler in handlers.keys())
+
+    if str(issue).startswith(supported_handlers):
+        handler_code, _, _ = str(issue).partition(":")
```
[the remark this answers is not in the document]
rochacbruno: Consider if one day we want to use other issue tracker like `GL:gitlab.server:8000/repo/issue/1234`

```python
> 'GL:gitlab.server:8000/repo/issue/1234'.split(':')                                                                                                                                         
['GL', 'gitlab.server', '8000/repo/issue/1234']
```

So with partition I ensure `[-1]` will always have what I want

```python
> 'GL:gitlab.server:8000/repo/issue/1234'.partition(':')                                                                                                                                     
('GL', ':', 'gitlab.server:8000/repo/issue/1234')

```

robottelo/helpers.py
```diff
@@ -750,3 +751,218 @@ def host_provisioning_check(ip_addr):
...
+
+
+def _should_deselect(issue, data=None):
+    """Check if test should be deselected based on marked issue."""
+    # Handlers can be extended to support different issue trackers.
+    handlers = {'BZ': bugzilla.should_deselect_bz}
+    supported_handlers = tuple(f"{handler}:" for handler in handlers.keys())
+    if str(issue).startswith(supported_handlers):
+        handler_code, _, _ = str(issue).partition(":")
+        return handlers[handler_code.strip()](issue.strip(), data)
```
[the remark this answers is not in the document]
rochacbruno: I thought about that, but at the end chose to go with 2 separate functions, because one is public, exposed to be imported and other is to be used only by the the `generate_issue_collection`, also the first is throwing error if can't handle the issue, the other is just returning None.

Well maybe it falls on [YAGNI](https://www.martinfowler.com/bliki/Yagni.html) but for example, if we decide to use github issues to mark tests, maybe github is not deselecting tests.

robottelo/helpers.py
```diff
@@ -750,3 +751,218 @@ def host_provisioning_check(ip_addr):
...
+        }
+    """
+
+    valid_markers = ["skip_if_open", "skip", "deselect"]
+    collected_data = defaultdict(lambda: {"data": {}, "used_in": []})
+
+    bz_cache = config.getvalue('bz_cache')  # use existing json cache?
+    cached_data = None
+    if bz_cache:
+        with open(bz_cache) as bz_cache_file:
```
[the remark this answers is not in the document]
rochacbruno: This block is executed only if someone passes `--bz-cache` explicitly to `py.test` so I think it is ok to have just the `FileNotFoundError` that `open` is going to throw.

robottelo/helpers.py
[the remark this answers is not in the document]
JacobCallahan: Interesting. I could never come up with a reason to use partition... until now!

robottelo/helpers.py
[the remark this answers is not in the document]
JacobCallahan: fair enough.

robottelo/helpers.py
```diff
@@ -750,3 +751,218 @@ def host_provisioning_check(ip_addr):
...
+                )
+
+                # Store issue key to lookup in the deselection process
+                deselect_data[item.location] = issue_key
+
+        # Then take the workarounds using `is_open` helper.
+        if 'is_open(' in source:
+            def add_workaround(matches, condition):
+                for match in matches:
+                    issue = "{0}:{1}".format(*match)
```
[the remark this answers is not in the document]
rochacbruno: Here we have the options for:

```python
IS_OPEN.findall(source)
# and then
"{0}:{1}".format(*match)
```
or

```python
found = IS_OPEN.search(source)
if found:
    src = found.group('src')
    num = found.group('num')
# and then

f"{src}:{num}"
```

robottelo/helpers.py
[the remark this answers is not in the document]
JacobCallahan: I'm probably too over-cautious, so I lean toward the "users are going to break it" side. If it becomes an issue, we can add a check later.

robottelo/helpers.py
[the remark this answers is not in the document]
JacobCallahan: I was thinking exactly like 2. *match is elegant, but explicit indexes, in this case, are more apparent. Plus... f-strings!!!

JacobCallahan:
ACK, thanks @rochacbruno

robottelo/helpers.py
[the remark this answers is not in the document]
mirekdlugosz: @rochacbruno FYI, second argument to `split` says how many splits at most you want to perform:
```python
>>> 'GL:gitlab.server:8000/repo/issue/1234'.split(':', 1)
['GL', 'gitlab.server:8000/repo/issue/1234']
```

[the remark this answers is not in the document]
rochacbruno: @mirekdlugosz great! I was not aware about that, that would be saved me lots of `partition` uses... 

For now I dont see a reason for changing, looks the performance is the same, maybe partition is an alias to `split(sep, 1)`

JacobCallahan:
@SatelliteQE/quality-engineers can we get at least one T1 reviewer for this?

omkarkhatavkar:
@rochacbruno Do you think we should the whole regression with this branch or change at once before applying this change in master, Just to ensure we are not breaking other things. I know it is big ask but this will give more confidence.

robottelo/cli/subscription.py
```diff
@@ -35,8 +34,6 @@ class Subscription(Base):
     def upload(cls, options=None, timeout=None):
         """Upload a subscription manifest."""
         cls.command_sub = 'upload'
-        if bz_bug_is_open(1669186) and (timeout is None or timeout < 1500):
-            timeout = 1500
```
omkarkhatavkar: @rochacbruno earlier this line was 
`timeouttimeout = 1500 if bz_bug_is_open(1339696) else 300 = 1500 if bz_bug_is_open(1339696) else 300` Do you think we  need  to keep it 300 ?

robottelo/cli/subscription.py
```diff
@@ -47,8 +44,6 @@ def upload(cls, options=None, timeout=None):
     def delete_manifest(cls, options=None, timeout=None):
         """Deletes a subscription manifest."""
         cls.command_sub = 'delete-manifest'
-        if bz_bug_is_open(1669186) and (timeout is None or timeout < 1500):
-            timeout = 1500
```
omkarkhatavkar: same here

robottelo/cli/subscription.py
```diff
@@ -59,8 +54,6 @@ def delete_manifest(cls, options=None, timeout=None):
     def refresh_manifest(cls, options=None, timeout=None):
         """Refreshes a subscription manifest."""
         cls.command_sub = 'refresh-manifest'
-        if bz_bug_is_open(1669186) and (timeout is None or timeout < 1500):
-            timeout = 1500
```
omkarkhatavkar: same here

tests/foreman/api/test_multiple_paths.py
```diff
@@ -214,9 +214,8 @@ def test_positive_post_status_code(self):
                 # partials cannot be compared for class identity and the class
                 # hierarchy needs fixing (SatelliteQE/nailgun#42), so we just
                 # comment it out above.
-                if (entity_cls in BZ_1118015_ENTITIES and
-                        bz_bug_is_open(1118015)):
-                    self.skipTest('Bugzilla bug 1118015 is open.')
+                if entity_cls in BZ_1118015_ENTITIES and is_open('BZ:1118015'):
+                    continue  # pytest can't skip inside a subTest.
```
omkarkhatavkar: As I understood we are removing the skipping part. because of "# pytest can't skip inside a subTest." ? can we please some logger statement as well.

tests/foreman/api/test_multiple_paths.py
```diff
@@ -292,12 +290,12 @@ def test_positive_put_status_code(self):
             content-type
 
         :CaseImportance: Critical
+
+        :BZ: 1378009
```
omkarkhatavkar: missed to add 1122257 Bugzilla number

tests/foreman/api/test_multiple_paths.py
```diff
@@ -226,7 +225,6 @@ def test_positive_post_status_code(self):
                 )
 
     @tier1
-    @skip_if_bug_open('bugzilla', 1122257)
```
omkarkhatavkar: missed to add :BZ: 1122257 statement

tests/foreman/api/test_syncplan.py
```diff
@@ -835,11 +838,11 @@ def test_positive_synchronize_custom_products_future_sync_date(self):
             with self.assertRaises(AssertionError):
                 self.validate_task_status(repo.id)
         # Create and Associate sync plan with products
-        if bz_bug_is_open(1695733):
-            self.logger.info('Need to set seconds to zero because BZ#1695733')
-            sync_date = datetime.utcnow().replace(second=0) + timedelta(seconds=delay)
-        else:
-            sync_date = datetime.utcnow() + timedelta(seconds=delay),
```
omkarkhatavkar: https://bugzilla.redhat.com/show_bug.cgi?id=1695733 still an open bug, why did else condition was removed here? if it valid then The same case also should be updated in test test_positive_synchronize_custom_product_future_sync_date

tests/foreman/cli/test_discoveredhost.py
```diff
@@ -104,16 +102,6 @@ def setUpClass(cls):
...
-        if bz_bug_is_open(1578290):
-            ssh.command('mkdir /var/lib/tftpboot/boot/fdi-image')
-            ssh.command('ln -s /var/lib/tftpboot/boot/'
-                        'foreman-discovery-image-3.4.4-1.iso-vmlinuz'
-                        ' /var/lib/tftpboot/boot/fdi-image/vmlinuz0')
-            ssh.command('ln -s /var/lib/tftpboot/boot/'
-                        'foreman-discovery-image-3.4.4-1.iso-img'
-                        ' /var/lib/tftpboot/boot/fdi-image/initrd0.img')
-            ssh.command('chown -R foreman-proxy /var/lib/tftpboot/boot/')
-
```
omkarkhatavkar: One Open Question here @SatelliteQE/quality-engineers,
we are removing all good amount tweaking code in case of issues, similar above case. If tomorrow we get same issue I'm not sure we open a new issue or reopen previously closed issues.

But importatnt thing is that we remove these all lines from code that we again have to reinvent unless someone tries to see history for the quick fix.

jyejare: I think the best would be to keep this code in the commented format instead of removing it. Their is no harm or increase in time either by keeping it commented.

We can keep something like this:
```
'''
Patch for BZ: 1578290
ssh.command('mkdir /var/lib/tftpboot/boot/fdi-image')
ssh.command('ln -s /var/lib/tftpboot/boot/'
         'foreman-discovery-image-3.4.4-1.iso-vmlinuz'
         ' /var/lib/tftpboot/boot/fdi-image/vmlinuz0')
ssh.command('ln -s /var/lib/tftpboot/boot/'
         'foreman-discovery-image-3.4.4-1.iso-img'
         ' /var/lib/tftpboot/boot/fdi-image/initrd0.img')
ssh.command('chown -R foreman-proxy /var/lib/tftpboot/boot/')
'''
```

tests/foreman/api/test_syncplan.py
swadeley: yes, line 784 & 785 needs to be copied to 845

tests/foreman/longrun/test_oscap.py
```diff
@@ -651,9 +649,6 @@ def test_positive_oscap_run_with_tailoring_file_with_ansible(self):
                 'organization': self.config_env['org_name'],
                 'ansible-role-ids': role_id
             })
-            # needed to work around BZ#1650103
-            if bz_bug_is_open(1650103):
-                ssh.command('''sed -i '/ProxyCommand/s/^/#/g' /etc/ssh/ssh_config''')
```
omkarkhatavkar: Good to keep this workaround in comment format.

tests/foreman/cli/test_discoveredhost.py
mirekdlugosz: I wouldn't worry about losing workaround if similar issue ever re-surfaces. I would trust component owner to know about this trick. It's probably something that could be added as a note in "how to test component X" page on Mojo, one of these that management asked us to create earlier this year.

And please, don't leave it as a comment. Dead code place is in version control history.

JacobCallahan: +1 to no comment
However, loss of knowledge is a legitimate concern, especially as component ownership will change. Maybe we need to store these kinds of workarounds somewhere. If not in a private comment on the bug itself, then maybe on mojo or a new repo called "deprecated workarounds" or something.

JacobCallahan:
Thanks, @omkarkhatavkar! 
@rochacbruno I definitely agree with Omkar's desire to have your branch ran through our standalone automation job. I think that will give us all a little peace of mind :)

robottelo/cli/subscription.py
rochacbruno: Hi @omkarkhatavkar this line was

```py
if bz_bug_is_open(1669186) and (timeout is None or timeout < 1500):
    timeout = 1500
```

So the first condition was already evaluating to false `if bz_bug_is_open(1669186)` then `timeout = 1500` this was never being set since the BZ was closed 5 months ago.

tests/foreman/cli/test_discoveredhost.py
rochacbruno: Git has a tool for that https://git-scm.com/docs/git-notes
+1 to keep that as a private BZ comment until we figureout better use of git notes.

robottelo/cli/subscription.py
omkarkhatavkar: okay, make sense.

JacobCallahan:
@rochacbruno once you've got positive results from the standalone automation, we can merge this. Thanks for all the updates!

rochacbruno:
@JacobCallahan I have the build `/view/Satellite 6/job/satellite6-standalone-automation/1718/` all the BZ part is ok as you can see in the .log artifacts.

But got some failures on `test setup` because my sat instance was not correctly set.

JacobCallahan:
@rochacbruno I can give you a satellite to run the standalone automation against. Hit me up on irc

JacobCallahan:
@SatelliteQE/quality-engineers I've looked through all the failures from the standalone automation run and nothing was caused by this pull request.
This PR has all acks and comments have been resolved. Merging!

requirements.txt
```diff
@@ -24,14 +24,8 @@ widgetastic.core==0.33
...
 urllib3==1.25.3
-# python-bugzilla 2.0.0 changed a lot of function signatures
-# also it removed some functions use e.g: getbugsimple
-# to update it code changes is needed.
-python-bugzilla==1.2.2  # pyup: ignore
-
 PyYAML==5.1.2
-robozilla==0.2.6
-
+tenacity==5.1.1
```
ntkathole: @rochacbruno any blocker for not using latest tenacity 5.1.4  ?

rochacbruno: @ntkathole no blockers I am using 5.1.5 locally and works.

on #7443 Travis is not happy with it.

PR #7407, opened 2019-10-15
Added test around BZ#1667129

vijay8451:
- BZ 1667129
- Test results:
```
tests/foreman/api/test_product.py::ProductTestCase::test_positive_filter_product_list PASSED                                                                          [100%]

================================================================= 1 passed, 17 deselected in 42.32 seconds ==================================================================
```

### DIFF
--- a/tests/foreman/api/test_product.py
+++ b/tests/foreman/api/test_product.py
@@ -21,6 +21,8 @@ http://<sat6>/apidoc/v2/products.html
 from fauxfactory import gen_string
 from nailgun import entities
 from requests.exceptions import HTTPError
+from robottelo import manifests
+from robottelo.api.utils import upload_manifest
 from robottelo.constants import (
     FAKE_1_PUPPET_REPO,
     FAKE_1_YUM_REPO,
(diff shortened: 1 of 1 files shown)


tests/foreman/api/test_product.py
```diff
@@ -367,3 +369,30 @@ def test_positive_sync_several_repos(self):
...
+        :CaseLevel: Integration
+
+        :BZ: 1667129
+        """
+        product = entities.Product(organization=self.org.id).create()
+        # Manifest upload to create RH Product
+        with manifests.clone() as manifest:
+            upload_manifest(self.org.id, manifest.content)
+
+        custom_product = entities.Product(organization=self.org.id).search(query={'custom': True})
```
omkarkhatavkar: can you change the custom_product to custom_products for more clarity

tests/foreman/api/test_product.py
```diff
@@ -367,3 +369,30 @@ def test_positive_sync_several_repos(self):
...
+        rh_product = entities.Product(organization=self.org.id).search(query={'redhat_only': True})
```
omkarkhatavkar: same here

tests/foreman/api/test_product.py
vijay8451: it's there now .

tests/foreman/api/test_product.py
vijay8451: same as above.

tests/foreman/api/test_product.py
```diff
@@ -367,3 +369,31 @@ def test_positive_sync_several_repos(self):
...
+        # Manifest upload to create RH Product
+        with manifests.clone() as manifest:
+            upload_manifest(self.org.id, manifest.content)
+
+        custom_products = entities.Product(organization=self.org.id).search(query={'custom': True})
+        rh_products = entities.Product(organization=self.org.id).search(
+            query={'redhat_only': True})
+
+        self.assertEqual(len(custom_products), 1)
+        self.assertIn(product.name, [prod.name for prod in custom_products])
```
omkarkhatavkar: why need to create list comprehension as there will only 1 custom_product ? as you already have added check above statement.
`
self.assertEqual(product.name, custom_products[0].name)`

tests/foreman/api/test_product.py
```diff
@@ -367,3 +369,31 @@ def test_positive_sync_several_repos(self):
...
+        self.assertNotIn('Red Hat Beta', [prod.name for prod in custom_products])
```
omkarkhatavkar: same here

vijay8451: It's added because if incase any regression comes then it could contain the every Product(including RH and Custom) and assertion will not True.

tests/foreman/api/test_product.py
vijay8451: yeah we not needed list comprehension here .......nice catch ...thanks

vijay8451: it's update now . revisit please.

PR #7424, opened 2019-10-22
add coverage for 1760701

pondrejk:
Adding an assert to cover https://bugzilla.redhat.com/show_bug.cgi?id=1760701
update: edited to add new style skipping, therefore this is blocked by #7404

### DIFF
--- a/tests/foreman/api/test_role.py
+++ b/tests/foreman/api/test_role.py
@@ -30,6 +30,7 @@ from robottelo.decorators import (
     tier3,
     upgrade
 )
+from robottelo.helpers import is_open
 from robottelo.test import APITestCase
 from robozilla.decorators import skip_if_bug_open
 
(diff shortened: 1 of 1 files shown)


tests/foreman/api/test_role.py
```diff
@@ -1498,6 +1501,10 @@ def test_positive_create_user_by_org_admin(self):
         ).create()
         self.assertEqual(user_login, user.login)
         self.assertEqual(org_admin.id, user.role[0].id)
+        if is_open('BZ:1760701'):
```
JacobCallahan: should this be ```if not is_open(```?

rplevka:
Is that a valid bug? Should org_admin be implicitly able to create locations? (not a blocker)

JacobCallahan:
@rplevka it's a new bug, so we'll just have to see what dev decides. Feel free to raise your concerns on the bug as well! Very well could be not a bug :)

PR #7425, opened 2019-10-22
Closed loop: BZ1750924

lhellebr:


lhellebr:
No test run results provided because right now, the test is skipped, otherwise it would fail (true positive).

tests/foreman/cli/test_reporttemplates.py
```diff
@@ -654,3 +655,41 @@ def test_negative_nonauthor_of_report_cant_download_it(self):
...
+            'name': 'Host statuses'
+        })
+        result = ReportTemplate.generate({
+            'name': rt_host_statuses['name'],
+            'organization': DEFAULT_ORG
+        })
+
+        self.assertIn(
+            host['name'],
+            [item.split(',')[0] for item in result]
```
JacobCallahan: if each item is a string, and the hostname isn't going to be in a location that could cause a false positive, then you shouldn't need the split here. Do you have an example of what item and/or result should look like?

mirekdlugosz: `split()` is poor man's `csv.reader`, but personally I would rather have `split()` than weak assertion of "happens to be anywhere in generated file".

JacobCallahan: I love my split(), but sometimes it isn't needed for simple cases, which I think this could potentially be. Depending on the composition of result, a simple assertIn would suffice and be more efficient. If the data is complex, or could potentially contain the hostname in other locations, then the current method is preferred.

lhellebr: The report contains columns "Name,Global,Addons,Build,Compliance,Configuration,Errata,Execution,Role,Service Level,Subscription,System Purpose,Traces,Usage".

On the scale from full CSV reader to grep, I think checking if "there is hostname before the first comma" is reasonable because from line 681, I know the name should not contain comma and on the other hand, I want to test that it's the name that matches and not something else (for example if order of columns didn't match the header, it would be a bug).

JacobCallahan: That's a few columns! Splitting is perfect for this scenario, ACK.

lhellebr:
Can we get this merged? @mirekdlugosz ?

tests/foreman/cli/test_reporttemplates.py
```diff
@@ -654,3 +655,41 @@ def test_negative_nonauthor_of_report_cant_download_it(self):
                 'id': report_template['name'],
                 'job-id': schedule[0].split("Job ID: ", 1)[1],
             })
+
+    @skip_if_bug_open('bugzilla', 1750924)
```
pondrejk: now that https://github.com/SatelliteQE/robottelo/pull/7404 is merged we'll probably need  @pytest.mark.skip_if_open('BZ:1750924') here

pondrejk:
@lhellebr ah, this got merged while I was commenting, so probably a separate PR then :)

PR #7426, opened 2019-10-22
Closed Loop: added test coverage for bugzilla 1649011

omkarkhatavkar:
### PR Description 
This test was added to cover the behaviour failed in https://bugzilla.redhat.com/show_bug.cgi?id=1649011

### Why Tier4?
As this test dependent on defaults setting and in an unsuccessful call for default delete can cause adverse effects in test automation, good candidate for tier4 test. I wanted to ensure it should be tested at last.

### DIFF
--- a/tests/foreman/cli/test_operatingsystem.py
+++ b/tests/foreman/cli/test_operatingsystem.py
@@ -16,7 +16,9 @@
 :Upstream: No
 """
 from fauxfactory import gen_alphanumeric, gen_string
+from robottelo import ssh
 from robottelo.cli.base import CLIReturnCodeError
+from robottelo.cli.defaults import Defaults
 from robottelo.cli.operatingsys import OperatingSys
 from robottelo.cli.factory import (
     CLIFactoryError,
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_operatingsystem.py
```diff
@@ -373,3 +382,33 @@ def test_positive_update_parameters_attributes(self):
...
+        make_os()
+        os_list_before_default = OperatingSys.list()
+        self.assertTrue(len(os_list_before_default) > 0)
+        try:
+            Defaults.add({
+                u'param-name': 'organization',
+                u'param-value': DEFAULT_ORG,
+            })
+            result = ssh.command('hammer defaults list')
+            self.assertTrue(DEFAULT_ORG not in "".join(result.stdout))
```
ntkathole: Can you paste results? At this step, DEFAULT_ORG should be in defaults. am I missing something?

tests/foreman/cli/test_operatingsystem.py
```diff
@@ -373,3 +382,33 @@ def test_positive_update_parameters_attributes(self):
...
+            os_list_after_default = OperatingSys.list()
+            self.assertTrue(len(os_list_after_default) > 0)
```
ntkathole: Shall we check return code as well?

JacobCallahan:
@omkarkhatavkar if you're concerned that this test could significantly interfere with other tests, it can always be tagged with @destructive

tests/foreman/cli/test_operatingsystem.py
```diff
@@ -373,3 +383,36 @@ def test_positive_update_parameters_attributes(self):
         os = OperatingSys.info({'id': os_id}, output_format='json')
         self.assertEqual(param_name, os['parameters'][0]['name'])
         self.assertEqual(param_value, os['parameters'][0]['value'])
+
+    @tier4
```
ntkathole: you can keep either tier4 or destructive. Not both.

omkarkhatavkar: okay

PR #7431, opened 2019-10-24
Added SRPM CLIs tests

vijay8451:
- Added SRPMs CLIs tests.

### DIFF
--- a/tests/foreman/cli/test_repository.py
+++ b/tests/foreman/cli/test_repository.py
@@ -43,6 +43,7 @@ from robottelo.cli.filter import Filter
 from robottelo.cli.repository import Repository
 from robottelo.cli.role import Role
 from robottelo.cli.settings import Settings
+from robottelo.cli.srpm import Srpm
 from robottelo.cli.user import User
 from robottelo.constants import (
     FEDORA27_OSTREE_REPO,
@@ -1849,14 +1850,13 @@ class RepositoryTestCase(CLITestCase):
             {'organization-id': org['id']})
         self.assertEqual(len(repos), 0)
 
-    @skip_if_bug_open('bugzilla', 1378442)
-    @tier1
-    def test_positive_upload_content_srpm(self):
-        """Create repository and upload a SRPM content
+    @tier2
+    def test_positive_upload_remove_srpm_content(self):
+        """Create repository, upload and remove an SRPM content
 
         :id: 706dc3e2-dacb-4fdd-8eef-5715ce498888
 
-        :expectedresults: File successfully uploaded
+        :expectedresults: SRPM successfully uploaded and removed
 
         :CaseImportance: Critical
         """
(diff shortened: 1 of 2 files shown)


tests/foreman/cli/test_repository.py
```diff
@@ -1865,16 +1865,77 @@ def test_positive_upload_content_srpm(self):
...
+            'name': new_repo['name'],
+            'organization': new_repo['organization'],
+            'path': "/tmp/{0}".format(SRPM_TO_UPLOAD),
+            'product-id': new_repo['product']['id'],
+            'content-type': 'srpm',
+        })
+        srpm_list = Srpm.list()
+        assert srpm_list[0]['filename'] == SRPM_TO_UPLOAD
+        assert len(srpm_list) == 1
+        assert Srpm.info({'id': int(srpm_list[0]['id'])})[0]['filename'] == SRPM_TO_UPLOAD
```
mirekdlugosz: Is it required in `Srpm.info()` for id to be int? Other entities doesn't seem to have such requirement. 
And amount of parentheses in this line is close to the max, it's getting hard to read at this point.

tests/foreman/cli/test_repository.py
```diff
@@ -1865,16 +1865,77 @@ def test_positive_upload_content_srpm(self):
...
+        assert int('{}'.format(Repository.info(
```
mirekdlugosz: Another `format` cast as `int()`.

tests/foreman/cli/test_repository.py
```diff
@@ -1865,16 +1865,77 @@ def test_positive_upload_content_srpm(self):
...
+            {'id': new_repo['id']})['content-counts']['source-rpms'])) == len(
+            Srpm.list({'repository-id': new_repo['id']}))
+        assert len(Srpm.list({'organization': new_repo['organization'], 'product-id': new_repo[
+            'product']['id']})) == 1
+        assert len(Srpm.list({'organization': new_repo['organization']})) == 1
```
mirekdlugosz: This will look for SRPMs in default organization, so assumption that SRPM created in this test will be the only one seems a little too strong / optimistic for me. It's enough for another test to fail with cleanup for this to fail.

tests/foreman/cli/test_repository.py
```diff
@@ -1865,16 +1865,77 @@ def test_positive_upload_content_srpm(self):
...
+        assert len(Srpm.list(
+            {'organization': new_repo['organization'], 'lifecycle-environment': 'Library'})) == 1
+        assert len(Srpm.list({'content-view-version-id': 1})) == 1
```
mirekdlugosz: Are content view version ids unique within content view? Or are they unique globally?
In latter case, I wouldn't count that this test will be the first to run in session. Most likely first created content view version will not have any SRPM at all.

tests/foreman/cli/test_repository.py
vijay8451: New org is creating as part of `Setup` hence using new_repo org details to filter SRPMs org wise .

tests/foreman/cli/test_repository.py
vijay8451: `'content-view-version-id': 1`  is fixed for `Default Organization` C.V.  so here this assert is checking SRPMs detail in `Default Organization View` version 1.0 wise..

```
ID:                           1
Name:                         Default Organization View
Label:                        Default_Organization_View
Composite:                    false
Description:                  
Content Host Count:           0
Force Puppet:                 false
Solve Dependencies:           false
Organization:                 Default Organization
Yum Repositories:             

Container Image Repositories: 

OSTree Repositories:          

Puppet Modules:               

Lifecycle Environments:       
 1) ID:   1
    Name: Library
Versions:                     
 1) ID:        1                 <===========
    Version:   1.0
    Published: 2019/10/12 07:09:04
```

vijay8451: I have added another more user readable assertion now .

vijay8451:
@mirekdlugosz thanks to review .. I am done with code re-factor part(which was pending earlier) and removed [WIP] from Title ..
In recent push all your comments are addressed , so please re-visit .

tests/foreman/cli/test_repository.py
mirekdlugosz: `new_repo` is created using [`self._make_repository`](https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/cli/test_repository.py#L117-L125), which does not define organization at all. As far as I know, command like that will assume default organization unless explicitly stated.

You would need something like
```python
new_repo = self._make_repository({
    'name': gen_string('alpha', 15),
    'organization-id': self.org['id']
})
```

tests/foreman/cli/test_repository.py
```diff
@@ -1849,14 +1850,13 @@ def test_negative_restricted_user_cv_add_repository(self):
             {'organization-id': org['id']})
         self.assertEqual(len(repos), 0)
 
-    @skip_if_bug_open('bugzilla', 1378442)
-    @tier1
-    def test_positive_upload_content_srpm(self):
-        """Create repository and upload a SRPM content
+    @tier2
+    def test_positive_upload_remove_srpm_content(self):
+        """Create repository , upload and remove a SRPM content
```
JacobCallahan: Create repository, upload and remove an SRPM content

tests/foreman/cli/test_repository.py
```diff
@@ -1865,16 +1865,78 @@ def test_positive_upload_content_srpm(self):
...
+        Repository.remove_content({
+            'id': new_repo['id'],
+            'ids': [Srpm.list({'repository-id': new_repo['id']})[0]['id']],
+            'content-type': 'srpm'
+        })
+        assert int(Repository.info({'id': new_repo['id']})['content-counts']['source-rpms']) == 0
+
+    @tier2
+    def test_positive_srpm_list_end_to_end(self):
+        """Create repository,  upload, list and remove a SRPM content
```
JacobCallahan: Create repository,  upload, list and remove an SRPM content

tests/foreman/cli/test_repository.py
vijay8451: new_org is taking care under `setup`  ..
https://github.com/SatelliteQE/robottelo/blob/f3162b1ddf4eb271b47c85ad3d1574bd8f5654c1/tests/foreman/cli/test_repository.py#L106-L116

tests/foreman/cli/test_repository.py
vijay8451: Disagree due to [1][2]
[1] https://wiki.centos.org/HowTos/RebuildSRPM
[2] https://www.theclassroom.com/use-vs-sentence-2221937.html

tests/foreman/cli/test_repository.py
vijay8451: same as above.

tests/foreman/cli/test_repository.py
mirekdlugosz: @vijay8451 You are right, this is tied to new organization through product, created at setUp phase. I see it now. Sorry for being stubborn.

vijay8451:
@JacobCallahan thanks but this PR was require to merge in `master branch` when branching is done for `6.6.z` ... now I have to raise one more PR for same to `revert` it as change are only applicable for 6.7.

PR #7449, opened 2019-10-31
Fix #7445 - allow to run tests without having bugzilla api_key

### LINKED ISSUE #7445: Not able to skip bugzilla logic if bugzilla section removed from properties file
With Introduction of new bugzilla logic #7404 , we loose the functionality of running tests without checking anything related to bugzilla. Add skipping of bugzilla logic if robottelo properties file has bugzilla section not set.

rochacbruno:
Fix #7445 

If `[bugzilla]` section in settings is empty, logger will output this.

```bash
2019-10-30 21:00:28 - robottelo.issue_handlers.bugzilla - WARNING - Config file is missing bugzilla api_key so all tests with skip_if_open mark is skipped. Provide api_key or a bz_cache.json.
```

### DIFF
--- a/robottelo/issue_handlers/bugzilla.py
+++ b/robottelo/issue_handlers/bugzilla.py
@@ -9,7 +9,6 @@ from packaging.version import Version
 from tenacity import retry, stop_after_attempt, wait_fixed
 
 from robottelo.config import settings
-from robottelo.config.base import ImproperlyConfigured
 from robottelo.constants import (
     CLOSED_STATUSES,
     OPEN_STATUSES,
@@ -31,6 +30,7 @@ def is_open_bz(issue, data=None):
         issue {str} -- The BZ reference e.g: BZ:123456
         data {dict} -- Issue data indexed by <handler>:<number> or None
     """
+
     bz = try_from_cache(issue, data)
     if bz.get("is_open") is not None:  # bug has been already processed
         return bz["is_open"]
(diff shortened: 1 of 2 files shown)


rochacbruno:
@ntkathole

robottelo/issue_handlers/bugzilla.py
```diff
@@ -232,6 +233,28 @@ def get_data_bz(bz_numbers, cached_data=None):  # pragma: no cover
...
+            "Provide api_key or a bz_cache.json."
+        )
+        # Provide default data for collected BZs
+        return [
+            {
+                "id": number,
+                "is_open": True,  # All marked is skipped
+                "is_deselected": False,  # nothing is deselected
+                "status": "",
+                "resolution": "",
```
vijay8451: I have a query here , in return why `status, resolution, cf_clone_of, clone_ids` are required without values ? or am i missing something here ?

rochacbruno: You are right @mirekdlugosz to skip or not so skip will be decided by 
`is_open": True` and `"is_deselected": False`. But there this will also be looked up id BZ has duplicates ot clones, so we add that there just to defult it as a NO duplicate and having no clones.

JacobCallahan: in the future, doing dict lookups with .get() avoids this :)

rochacbruno: @JacobCallahan I am not a big fan of using `.get` for some cases,  that data is schemaless yet, so we need it to not silence the problem if for example Bugzilla API changes in future.

I would prefer to make some kind of Type Serialization, define a `Bug` Schema using Marshmallow and then serialize API response to that object.

rochacbruno: @JacobCallahan not Marshmallow, I mean `Serde` (inspired by Rust) https://pypi.org/project/serde/

PR #7451, opened 2019-11-01
Proposal: Use ROBOTTELO_ prefixed envvars to override .properties

rochacbruno:
With the follwing simple changes one can do:

```bash
$ export ROBOTTELO_SERVER_HOSTNAME=server.foo.bar.com
$ export ROBOTTELO_BUGZILLA_API_KEY=123456sdfksdjkfsd
```

Then `robottelo.config.base.INIReader` will first look for
variables in form: **ROBOTTELO_{SECTION}_{OPTION}** all uppercase
before it tries to read the value from the proper .properties file.

Following unix principle, envvars have precedence over settings files.

### DIFF
--- a/robottelo/config/base.py
+++ b/robottelo/config/base.py
@@ -68,6 +68,9 @@ class INIReader(object):
     def get(self, section, option, default=None, cast=None):
         """Read an option from a section of a INI file.
 
+        First try to lookup for the value as an environment variable having the
+        following format: ROBOTTELO_{SECTION}_{OPTION}.
+
         The default value will return if the look up option is not available.
         The value will be cast using a callable if specified otherwise a string
         will be returned.
@@ -78,10 +81,17 @@ class INIReader(object):
             defined.
         :param cast: If provided the value will be cast using the cast
             provided.
-
         """
+        # First try to read from environment variable.
+        # [bugzilla]
+        # api_key=123456
+        # can be expressed as:
+        # $ export ROBOTTELO_BUGZILLA_API_KEY=123456
+        value = os.environ.get(f'ROBOTTELO_{section.upper()}_{option.upper()}')
+
         try:
-            value = self.config_parser.get(section, option)
+            # If envvar does not exist then try from .properties file.
+            value = value or self.config_parser.get(section, option)
             if cast is not None:
                 if cast is bool:
                     value = self.cast_boolean(value)
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -135,23 +135,33 @@ To configure Robottelo, create a file named ``robottelo.properties``. You can
 use the ``robottelo.properties.sample`` file as a starting point. Then, edit the
 configuration file so that at least the following attributes are set::
 
-    server.hostname=[FULLY QUALIFIED DOMAIN NAME OR IP ADDRESS]
-    server.ssh.key_private=[PATH TO YOUR SSH KEY]
-    server.ssh.username=root
-    project=sat
-    locale=en_US
-    remote=0
-    smoke=0
+    [server]
+    hostname=[FULLY QUALIFIED DOMAIN NAME OR IP ADDRESS]
+    ssh_key=[PATH TO YOUR SSH KEY]
 
-    [foreman]
-    admin.username=admin
-    admin.password=changeme
+    [bugzilla]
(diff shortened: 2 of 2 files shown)


robottelo/config/base.py
```diff
@@ -78,10 +81,17 @@ def get(self, section, option, default=None, cast=None):
...
         :param cast: If provided the value will be cast using the cast
             provided.
-
         """
+        # First try to read from environment variable.
+        # [bugzilla]
+        # api_key=123456
+        # can be expressed as:
+        # $ export ROBOTTELO_BUGZILLA_API_KEY=123456
+        value = os.environ.get(f'ROBOTTELO_{section.upper()}_{option.upper()}')
```
JacobCallahan: we could just pull everything prefixed with ROBOTTELO_. If something we're not expecting is already there, then it can still be added to the runtime config and never used.

rochacbruno: @JacobCallahan then I recommend the use of http://dynaconf.readthedocs.io

mirekdlugosz: @JacobCallahan since both section and option can have any number of underscores, I don't think there is generic way to correctly extract these out of arbitrary string. At least, you need to compare section name against list of known names. I'm afraid this will be much more code than literally two lines added here.

And then, I don't see a difference between "never added from environment" (this PR) and "added from environment, but never retrieved (since nothing references it)" (your proposal, as I understand it).

If there is no visible difference between this and your propsal, and this PR achieves it in smaller and simpler code, what would be benefit of your proposal?

JacobCallahan: @mirekdlugosz I do something similar with plinko [1] where I just pull everything that has a pre-defined prefix. In plinko's case, it is PLINKO_. If we wanted it to be compatible with other presets, we could simply strip the "ROBOTTELO_", make the rest lowercase, and apply it to the running config. No need to get overly complicated.

rochacbruno:
# Initial Configuration

To configure Robottelo, create a file named ``robottelo.properties``. You can
use the ``robottelo.properties.sample`` file as a starting point. Then, edit the
configuration file so that at least the following attributes are set::

    [server]
    hostname=[FULLY QUALIFIED DOMAIN NAME OR IP ADDRESS]
    ssh_key=[PATH TO YOUR SSH KEY]

    [bugzilla]
    api_key=sdfsdg654g8df4gdf6g4df8g468dfg

Note that you only need to configure the SSH key if you want to run CLI tests.
There are other settings to configure what web browser to use for UI tests and
even configuration to run the automation using `SauceLabs`_. For more
information about what web browsers you can use, check Selenium's `WebDriver`_
documentation.

## Using environment variables

Each of the sections in the ``robottelo.properties`` file can be mapped to an 
environment variable prefixed with ``ROBOTTELO_`` so for example if you want
to override the ``server.hostname`` without changing the properties file you can do::

    $ export ROBOTTELO_SERVER_HOSTNAME=other.hostname.com

The envars follows the format ``ROBOTTELO_{SECTION}_{VALUE}`` all uppercase, more examples::

    $ export ROBOTTELO_SERVER_SSH_KEY=path/to/your/key
    $ export ROBOTTELO_BUGZILLA_API_KEY=sdfsdg654g8df4gdf6g4df8g468dfg

robottelo/config/base.py
JacobCallahan: @rochacbruno i'm just thinking about the possibility of a setting not conforming to the ROBOTTELO_X_Y naming convention. Not a blocker at all, just some ideas to make it more flexible.

rochacbruno: @JacobCallahan I am ok to make it more flexible, but I am concerned this will require too much complexity right now with no immediate advantage (use case). to do that we need to avoid conflicts with existing sections on properties file.

I suggest we do that in a different PR, and by adding a `meta` section to the config like `settings.env` and then we can access arbitrary variables only via that endpoint.

```bash
export ROBOTTELO_FOO_BAR=1
export ROBOTTELO_BUGZILLA_API_KEY=nskfnsakjnfajskf
```

then

```py
settings.env.foo_bar == 1
settings.env.bugzilla_api_key == 'nskfnsakjnfajskf'
settings.bugzilla.api_key == 'nskfnsakjnfajskf'
```
`settings.env` will be a catchall for all existing `ROBOTTELO_` envvars.

PR #7453, opened 2019-11-01
Update many repos to point to 6.6 instead of 6.2

JacobCallahan:
It's been a while since 6.2 has not been the latest GA'd version.
It is time to update to the latest (6.6).
In the future, this could be a rolling reference.


JacobCallahan:
Test results for this are waiting on all BZ fixes to be merged. After that, A full automation run will have to be triggered via standalone.

JacobCallahan:
@mirekdlugosz, since these are based on names/labels, we don't have to worry about checking repo path validity. But, as you said, if something isn't named correctly then we'll have quite a few additional failures! :D
I'll work on getting this tested this week.

PR #7458, opened 2019-11-06
[DO NOT MERGE!!]Change code formatting to black[DO NOT MERGE]

JacobCallahan:
This PR is a demo for the proposal to switch robottelo's code formatting
away from flak8 standards and over to python black.


JacobCallahan:
@SatelliteQE/quality-engineers Please check out the example changes in this PR. This will NOT be the final code to be merged when the change happens, as that should also include some additional pieces.

jyejare:
@JacobCallahan Formatting looks way better than flake8 and as it supports autoformatting its a cherry on cake. :)

Btw Travis is upset with you :(

JacobCallahan:
@jyejare yeah travis will be much less upset when the flake checks are removed!

rochacbruno:
@JacobCallahan we can still keep flake8 to warn about things like unused imports, unused variables etc..

Black can live with flake8 as long we ignore some specific checks example: https://github.com/rochacbruno/dynaconf/blob/master/.pre-commit-config.yaml#L35

kersommoura:
@rochacbruno  I have not used yet. But https://pypi.org/project/flake8-black/

JacobCallahan:
@SatelliteQE/robottelo-automation I've updated this PR with the new pre-commit hook.
Note that engineers will still need to run black before committing, as it follows the non-changing standard.
See also: https://github.com/pre-commit/pre-commit/issues/747#issuecomment-386782080

rochacbruno:
@mirekdlugosz you can use `pre-commit run -a -v` manually before commiting, I usually do this to avoid to do git add twice.

pyproject.toml
```diff
@@ -0,0 +1,17 @@
+[tool.black]
+line-length = 99
```
rochacbruno: OMG! I am on the 79 team.. but I think I can try to accommodate 99 now that I have a widescreen monitor.

JacobCallahan:
Ok, so I ran a bunch of automation runs and couldn't find any failures linked to this PR. There is another issue I've identified, so we may want to wait on resolving that. The solution to the other problem might be easier to find before this is merged.

JacobCallahan:
@rochacbruno @mshriver @mirekdlugosz So the issue in our standalone doesn't seem to be affecting our tiered automation runs in 6.7. If you all don't have any reservations, we should be safe to merge.

PR #7460, opened 2019-11-07
Remove f-string to make it compatible with py2

ntkathole:
To avoid polarion test case job failure

### DIFF
--- a/tests/foreman/api/test_reporttemplates.py
+++ b/tests/foreman/api/test_reporttemplates.py
@@ -141,7 +141,7 @@ class ReportTemplateTestCase(APITestCase):
         input_name = gen_string('alpha').lower()
         input_value = gen_string('alpha').lower()
         template_name = gen_string('alpha').lower()
-        template = f'<%= "value=\\"" %><%= input(\'{input_name}\') %><%= "\\"" %>'
+        template = '<%= "value=\\"" %><%= input(\'{0}\') %><%= "\\"" %>'.format(input_name)
         entities.Host(name=host_name).create()
         rt = entities.ReportTemplate(name=template_name, template=template).create()
         entities.TemplateInput(name=input_name,


mirekdlugosz:
@ntkathole why is Python 2 compatibility something we care about again?

There are other places in Robottelo where we use f-strings, and at least one place where we use pathlib (introduced in Py 3.4). This one PR doesn't make Robottelo py2-compatible, so how is this file different from everything else we have?

mirekdlugosz:
Re-reading initial message, I guess it has to do with betelgeuse. Didn't betelgeuse re-introduce all the features after switching to Python3? What exactly is it missing?

ntkathole:
@mirekdlugosz Yes, you got it correct, it is required for Betelgeuse. 

 "Betelgeuse stopped creating requirements because pylarion is not supported anymore. This command will be updated and will generate a XML file to be used with the requirements importer in a future release. Betelgeuse stopped creating test plans because pylarion is not supported anymore. This command will be removed in a future release."

JacobCallahan:
This is a big red flag for me. Bringing anything back to python 2 compatibility is the wrong direction. Python 2 is end of life in less than 2 months. If Betelgeuse has issues with moving to python 3, then we should be fixing it at that end.

@elyezer, iirc you're still maintaining Betelguese correct? Is there anything you need help with to make this not an issue?

elyezer:
IMO those are not reasons to come back to Python 2. Working with test plans and what not is team related and vary so each team should have some extra scripts to deal with their way to deal with versions and releases. Pylarion has been recently updated to Python 3 and could be used there.

About the requirements that is not hard to add and it was not because I was prioritizing working on https://github.com/SatelliteQE/betelgeuse/pull/123. It will eventually come as I have time to work on it or someone implement it.

PR #7436, opened 2019-10-25
Test container on robo

latran:
Container to replace vm for efficiency in contenthost creation.

PR #7469, opened 2019-11-11
bootstrap positive register and reregister

swadeley:
Closed loop: BZ#1739367

These are two container based tests and depend on:
https://github.com/SatelliteQE/robottelo/pull/7436

### DIFF
--- a/tests/foreman/cli/test_bootstrap_script.py
+++ b/tests/foreman/cli/test_bootstrap_script.py
@@ -1,5 +1,5 @@
 # -*- encoding: utf-8 -*-
-"""Test for bootstrap script (bootstrap.py)
+"""Test for bootstrap script (bootstrap.py).
 
 :Requirement: Bootstrap Script
 
@@ -15,8 +15,22 @@
 
 :Upstream: No
 """
-from robottelo.decorators import stubbed, tier1, upgrade
+from nailgun import entities
+from robottelo.decorators import stubbed, tier1, tier3, upgrade
+from robottelo import manifests
+from robottelo.api.utils import (
+    promote, enable_rhrepo_and_fetchid, upload_manifest
+    )
 from robottelo.test import CLITestCase
+from robottelo.config import settings
+from robottelo.containers import Container
+from robottelo.constants import (
+    DEFAULT_ARCHITECTURE,
+    RHEL_7_MAJOR_VERSION,
+    REPOS,
+    PRDS,
+    REPOSET,
+)
 
 
 class BootstrapScriptTestCase(CLITestCase):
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_bootstrap_script.py
```diff
@@ -15,62 +15,205 @@
...
+        subs = entities.Subscription().search(
+            query={'search': 'name~{0}'.format(prod.name)}
+        )
+
+        cls.ak.add_subscriptions(data={'subscriptions': [{'id': subs[0].id}]})
+
+        # Search for SmartProxy
+        proxy = entities.SmartProxy().search(
+            query={
+                u'search': u'name={0}'.format(
```
mirekdlugosz: You can drop this `u` prefix in new code. It was there only to force Python2 to treat string as unicode. Python3 does that by default. And we no longer pretend we support Python2.

tests/foreman/cli/test_bootstrap_script.py
```diff
@@ -15,62 +15,205 @@
...
 
-        :CaseAutomation: notautomated
+        :CaseAutomation: automated
 
-        :CaseImportance: Critical
+        :CaseImportance: High
         """
+        my_host = Container(agent=True)
+        host_name = my_host.execute("hostname")
+        my_fqdn = host_name.strip() + '.' + self.domain.name + '.' 'com'
```
mirekdlugosz: Wouldn't `format` be cleaner?
```python
my_fqdn = "{}.{}.com".format(host_name.strip(), self.domain.name)
```
f-string would be even more readable, but I think it trips betlegeuse and we shouldn't use it in test code for now.

tests/foreman/cli/test_bootstrap_script.py
```diff
@@ -15,62 +15,205 @@
...
+                                     self.org.name, self.loc.name,
+                                     self.hostgroup.name,
+                                     self.ak.name,
+                                     my_fqdn
+                                         )
+                                 )
+        assert "The system has been registered" in result, 'Not registered'
+        Container.delete(my_host)
 
     @tier1
```
mirekdlugosz: Tier1? This looks comparable to this other test here, which is Tier4.

tests/foreman/cli/test_bootstrap_script.py
```diff
@@ -15,62 +15,205 @@
...
+        :CaseAutomation: automated
+
+        :CaseImportance: Medium
         """
+        my_host = Container(agent=True)
+        host_name = my_host.execute("hostname")
+        my_fqdn = host_name.strip() + '.' + self.domain.name + '.' 'com'
+        # Register host the first time
+        my_host.execute("curl -O http://{}/pub/bootstrap.py".format(settings.server.hostname))
+        result = my_host.execute("python bootstrap.py -l admin -p changeme -s {} -o '{}' "
```
mirekdlugosz: Do you actually need this `bootstrap.py` call?

Seems in bugzilla steps are:
1. run bootstrap.py
2. remove `katello-ca-consumer` and any package that could be used to report status back to host
3. install `katello-ca-consumer`
4. register host using `subscription-manager`
5. run bootstrap.py again

Could you start from step 3 above? Would it reproduce the issue as well?

Your steps are:
1. run bootstrap.py
2. register host using `subscription-manager`
3. run bootstrap.py again

Shouldn't you de-register and re-install `katello-ca-consumer` between steps 2 and 3, then?

swadeley: Hello @mirekdlugosz 

At first I did try register the first time with `subscription-manager` but I needed the `--add-domain` option the first time for some reason so I decided to just use bootstrap.py twice.

In `test_positive_reregister` I see:
1] run bootstrap.py first time on line 183
2] run `subscription-manager attach --auto` on line 193
3] run bootstrap.py second time on line 198

I will ask jturel about the need to de-register and re-install `katello-ca-consumer` between steps 2 and 3.

Thank you

tests/foreman/cli/test_bootstrap_script.py
swadeley: OK, thank you

tests/foreman/cli/test_bootstrap_script.py
swadeley: Hello @mirekdlugosz 
As I am using the .format method, could you please elaborate?
Thank you

tests/foreman/cli/test_bootstrap_script.py
swadeley: That comes from the original spec, I mean how it was when stubbed out. Tier1 is supposed to be for "Critical component level functional tests". Lets make them both Tier3 as they are quite quick and Tier4 is for test with longer set-up times.

Thank you

tests/foreman/cli/test_bootstrap_script.py
mirekdlugosz: I mean this particular line.
You have:
```python
my_fqdn = host_name.strip() + '.' + self.domain.name + '.' 'com'
```
My proposal is:
```python
my_fqdn = "{}.{}.com".format(host_name.strip(), self.domain.name)
```

tests/foreman/cli/test_bootstrap_script.py
jturel: These test cases seem comprehensive since using bootstrap.py to register was broken completely in 6.5.2

I don't think there's any reason to manage the katello-ca-consumer RPM in any way - bootstrap.py does that on its own

tests/foreman/cli/test_bootstrap_script.py
swadeley: oh yes, now I see it. Will do, thank you

mirekdlugosz:
@swadeley it's been couple months. I think we should decide if we try to merge it (finish any remaining work, if any, and remove DO NOT MERGE label), or just close it.

swadeley:
> @swadeley it's been couple months. I think we should decide if we try to merge it (finish any remaining work, if any, and remove DO NOT MERGE label), or just close it.

Hello @mirekdlugosz , this depends on container context manager and related changes in #7436, and it looks like that will never be accepted. So I will close this PR and rewrite these tests some other way.

Thank you

PR #7471, opened 2019-11-12
added scenario for bugzilla_1655130

omkarkhatavkar:
### PR Description
An added scenario for Bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=1655130
### Dependancy 
Needs Airgun PR  https://github.com/SatelliteQE/airgun/pull/420

### DIFF
--- a/tests/foreman/ui/test_errata.py
+++ b/tests/foreman/ui/test_errata.py
@@ -540,7 +540,7 @@ def test_positive_content_host_errata_details(session, erratatype_vm, module_org
         2. Create a User with 'Viewer' Permission.
         3. Go to Content Hosts -> Select content host -> Errata Tab -> Select Errata_Id.
 
-    :expectedresults: The errata details apge should be disaplyed for User.
+    :expectedresults: The errata details page should be displayed for User.
 
     :BZ: 1655130
 
--- a/tests/foreman/ui/test_errata.py
+++ b/tests/foreman/ui/test_errata.py
@@ -14,12 +14,14 @@
 
 :Upstream: No
 """
+import pytest
 from airgun.session import Session
 from fauxfactory import gen_string
 from nailgun import entities
 
 from robottelo.api.utils import promote
 from robottelo.constants import (
+    DEFAULT_LOC,
     DISTRO_RHEL6,
     DISTRO_RHEL7,
     FAKE_1_CUSTOM_PACKAGE,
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_errata.py
```diff
@@ -525,6 +527,49 @@ def test_positive_content_host_search_type(session, erratatype_vm):
...
+
+    :id: 236de56f-8035-4072-b0fa-03abbf3fc692
+
+    :Steps:
+
+        1. Content Host with applicable security errata.
+        2. Create a User with 'Viewer' Permission.
+        3. Go to Content Hosts -> Select content host -> Errata Tab -> Select Errata_Id.
+
+    :expectedresults: The errata details apge should be disaplyed for User.
```
vijay8451:
```suggestion
    :expectedresults: The errata details page should be displayed for User.
```

tests/foreman/ui/test_errata.py
```diff
@@ -525,6 +527,49 @@ def test_positive_content_host_search_type(session, erratatype_vm):
...
+        query={'search': u'name="Viewer"'})
+    default_loc_id = entities.Location().search(
+         query={'search': 'name="{}"'.format(DEFAULT_LOC)})[0].id
+    default_loc = entities.Location(id=default_loc_id).read()
+
+    entities.User(role=viewer_role, login=login, password=password,
+                  default_location=default_loc, organization=[module_org]).create()
+
+    pkgs = " ".join(FAKE_9_YUM_OUTDATED_PACKAGES)
+    assert _install_client_package(erratatype_vm, pkgs, errata_applicability=True)
```
vijay8451: I see the `pkgs` in only using at one place ..Can we use as below it :
```
assert _install_client_package(erratatype_vm, 
                               " ".join(FAKE_9_YUM_OUTDATED_PACKAGES), 
                               errata_applicability=True)
```

tests/foreman/ui/test_errata.py
```diff
@@ -525,6 +527,49 @@ def test_positive_content_host_search_type(session, erratatype_vm):
...
+    :BZ: 1655130
+
+    :CaseLevel: Integration
+    """
+    login = gen_string('alpha')
+    password = gen_string('alpha')
+    viewer_role = entities.Role().search(
+        query={'search': u'name="Viewer"'})
+    default_loc_id = entities.Location().search(
+         query={'search': 'name="{}"'.format(DEFAULT_LOC)})[0].id
```
vijay8451: Any specific reason to use `DEFAULT_LOC`? Can we not create a new one here?

omkarkhatavkar: avoid increase time and this is a UI test we are seeing issues where we are selecting the location causing the satellite to take time load. I have seen even other tests most of them have used the default location. 

What are the befits are you seeing adding the different location?

tests/foreman/ui/test_errata.py
omkarkhatavkar: In your above example, we are using three lines of code where an existing example uses the 2 lines of code. At looking at cleanness I think the current solution looks better.

tests/foreman/ui/test_errata.py
vijay8451: Benefits are as similar to create a new org ... but it was a suggestion only ... 
as you mentioned it's already using (which I think not a right way) at mutiple place we can ignore it here and can handle under separate PR ... thanks

tests/foreman/ui/test_errata.py
vijay8451: It was a suggestion only :) .. thanks

PR #7472, opened 2019-11-12
BZ1475443 and BZ1278917 cleanup

lhellebr:


tests/foreman/cli/test_computeresource_rhev.py
```diff
@@ -271,110 +270,6 @@ def test_negative_add_image_rhev_with_invalid_name(self):
         :expectedresults: The image should not be added to the CR
         """
 
-    @tier2
-    @pytest.mark.skip_if_open("BZ:1278917")
-    @stubbed()
```
ntkathole: we have hammer compute-profile list command available. Can't this be automated now?

lhellebr: That command lists **all** compute profiles. I think this testcase was about listing compute profiles from the CR view, like in WebUI going to Infra -> CRs -> <CR> -> Compute Profiles (which IMO doesn't make much sense in CLI). The command you mentioned **should** be tested  by another test in another file: test_computeprofile. I will create that file and stubs later.

mirekdlugosz: @lhellebr "later" as in later in this PR, or "later" as in future PR?
Was about to merge it, but I wasn't sure if I shouldn't wait a little.

lhellebr: In future PR.

PR #7474, opened 2019-11-12
basic host view with neccesary permisssions with non-admin-user

tstrych:
QE-test-coverage
[BZ-link](https://bugzilla.redhat.com/show_bug.cgi?id=1642076#h6)

I figured that I have to use read_all, because read action in Host entity in airgun is defined by `edit 'btn-group'` but when only view_host permissions are used there is no edit button. 

Test result:
```
============= 1 passed, 28 deselected, 2 warnings in 66.39 seconds =============
```

### DIFF
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -880,6 +880,49 @@ def test_negative_delete_primary_interface(session, module_host_template):
         assert 'Interface Delete button is disabled' in str(context.value)
 
 
+@tier2
+def test_positive_view_hosts_with_non_admin_user(test_name, module_org, module_loc):
+    """View hosts and content hosts as a non-admin user with only view_hosts, edit_hosts
+    and view_organization permissions
+
+    :BZ: 1642076
+
+    :id: 19a07026-0550-11ea-bfdc-98fa9b6ecd5a
+
+    :expectedresults: user with only view_hosts, edit_hosts and view_organization permissions
+        is able to read content hosts and hosts
+
+    :CaseLevel: System
+    """
+    user_password = gen_string('alpha')
+    role = entities.Role(organization=[module_org]).create()
+    create_role_permissions(
+        role,
+        {
+            'Organization': ['view_organizations'],
+            'Host': ['view_hosts'],
+        }
+    )
+    user = entities.User(
+        role=[role],
+        admin=False,
+        password=user_password,
+        organization=[module_org],
+        location=[module_loc],
+        default_organization=module_org,
+        default_location=module_loc,
+    ).create()
+    created_host = entities.Host(
+        location=module_loc,
+        organization=module_org,
+    ).create()
+    with Session(test_name, user=user.login, password=user_password) as session:
+        host = session.host.get_details(created_host.name, widget_names='breadcrumb')
+        assert host['breadcrumb'] == created_host.name
+        content_host = session.contenthost.read(created_host.name, widget_names='breadcrumb')
+        assert content_host['breadcrumb'] == created_host.name
+
+
 @tier3
 def test_positive_remove_parameter_non_admin_user(test_name, module_org, module_loc):
     """Remove a host parameter as a non-admin user with enough permissions


tstrych:
@vijay8451 @omkarkhatavkar sorry for pinging you, but no one take a look at this PR for 10 days.

tests/foreman/ui/test_host.py
```diff
@@ -880,6 +880,49 @@ def test_negative_delete_primary_interface(session, module_host_template):
...
+        default_organization=module_org,
+        default_location=module_loc,
+    ).create()
+    host = entities.Host(
+        location=module_loc,
+        organization=module_org,
+    ).create()
+    with Session(test_name, user=user.login, password=user_password) as session:
+        hosts = session.host.read_all()
+        assert any([True for h in hosts['table'] if h['Name'] == host.name])
```
mirekdlugosz: This feels unnecessary complicated. You filter the list on the fly and return True for every element that exists in new list.
Instead, you can just get the list of boolean values that tell if condition is true.
With generator comprehension, you can stop execution on first match.
```python
assert any(h['Name'] == host.name for h in hosts['table'])
```

tests/foreman/ui/test_host.py
```diff
@@ -880,6 +880,49 @@ def test_negative_delete_primary_interface(session, module_host_template):
...
+        content_hosts = session.contenthost.read_all()
+        assert any([True for ch in content_hosts['table'] if ch['Name'] == host.name])
```
mirekdlugosz: Same as above

tests/foreman/ui/test_host.py
```diff
@@ -880,6 +880,49 @@ def test_negative_delete_primary_interface(session, module_host_template):
...
+        default_location=module_loc,
+    ).create()
+    host = entities.Host(
+        location=module_loc,
+        organization=module_org,
+    ).create()
+    with Session(test_name, user=user.login, password=user_password) as session:
+        hosts = session.host.read_all()
+        assert any([True for h in hosts['table'] if h['Name'] == host.name])
+        content_hosts = session.contenthost.read_all()
```
omkarkhatavkar: please avoid the read_all(), try to use the widget parameter with read_all. read_all takes too much time to read all widgets.

tests/foreman/ui/test_host.py
```diff
@@ -880,6 +880,49 @@ def test_negative_delete_primary_interface(session, module_host_template):
...
+        location=[module_loc],
+        default_organization=module_org,
+        default_location=module_loc,
+    ).create()
+    host = entities.Host(
+        location=module_loc,
+        organization=module_org,
+    ).create()
+    with Session(test_name, user=user.login, password=user_password) as session:
+        hosts = session.host.read_all()
```
omkarkhatavkar: please avoid the read_all()

tests/foreman/ui/test_host.py
omkarkhatavkar: Instead writing only `h` please use `host`, helps in reading the code. Just a suggestion

omkarkhatavkar:
> @vijay8451 @omkarkhatavkar sorry for pinging you, but no one takes a look at this PR for 10 days.

@tstrych me and @vijay8451 were busy in ISTQB training whole week and could not able to look at this PR. If you have any urgency please feel free to tag SatelliteQE/quality-engineers.

tstrych:
I measure the time for two approaches in this case where we expect just one host:
```
start = time.time()
host = session.host.get_details(created_host.name)
assert host['breadcrumb'] == created_host.name
print(time.time() - start)
start = time.time()
hosts = session.host.read_all()
assert any(host['Name'] == created_host.name for host in hosts['table'])
print(time.time() - start)
```
with results ` 35 seconds for get_details and 15 seconds for read_all`

Test results:
`============= 1 passed, 28 deselected, 2 warnings in 63.90 seconds =============`

tests/foreman/ui/test_host.py
tstrych: as I explain at descrption, in this setup you can't use `read` because ui expect the button with ▼
( i don't want to change edit button )

tests/foreman/ui/test_host.py
tstrych: Thanks @mirekdlugosz.

mirekdlugosz:
@tstrych Yes, all tests in `tests/foreman/ui/test_host.py` will share organization and location. There are 28 tests total, assuming each will leave one host behind, we can expect 30 hosts to be present when your test runs. `read_all()` might still be faster than other approach, but difference won't be so spectacular.

But I'm more concerned that with 30 hosts, we will take a code path in Airgun that was giving us troubles in the past. I'm afraid `read_all()` approach may simply fail.

Would it be possible for you to register ~30 fake hosts in single organization and run `read_all()` approach on that? So we know whether it works and how much time it might take?

Otherwise, I would say to just use `session.host.get_details()` and `session.contenthost.read()` for individual host you are interested in. 35 seconds is not that bad.

tstrych:
@mirekdlugosz  as you said I tried it with 30 fake host and the result was 70 seconds for `session.host.read_all()` 

I use the 2. approach - read in contenthost and get_details in host

test-result
`============ 1 passed, 28 deselected, 2 warnings in 127.82 seconds =============`

tstrych:
DONE, thanks for your participation mirek. 
============= 1 passed, 28 deselected, 2 warnings in 87.30 seconds =============

tstrych:
Requested changes after discussion with @rplevka: 

- navigate directly  use https://github.com/SatelliteQE/airgun/pull/417#issue-336758935

- use fixture for setup

rplevka:
After a discussion with @tstrych we came to conclusion that using additional fixtures wouldn't really bring in any benefit, it would actually obfuscate the code, so I'm merging this as is

PR #7475, opened 2019-11-12
Remove containers support removal test

mirekdlugosz:
This was added in 6.6 to check that containers support is correctly removed during 6.5 -> 6.6 upgrade. Since there is no support for containers in 6.6, this test doesn't make sense during 6.6 -> 6.7 upgrade.


ntkathole:
This PR needs branching so adding do not merge for now.

mirekdlugosz:
> Will it be worth to have any test to verify container related stuff is not part of any upcoming build ?

I don't think so. Features don't magically appear all by itself. Even if foreman_docker plugin gets pulled in by mistake, latest version removed all code that hooked into Foreman/Satellite.

We had deprecation period for this in 6.6, now we can finally remove it.

tests/upgrades/test_containers_removal.py
```diff
@@ -1,204 +0,0 @@
...
-        :expectedresults:
-
-            1. Docker host is created
-            2. Container from dockerhub is created and running
-            3. Container from external registry is created and running
-
-        """
-        repo_name = 'rhel'
-        compute_resource_name = gen_string('alpha')
-        registry_url = settings.docker.external_registry_1
```
omkarkhatavkar: Do we have to remove these setting from .properties files?

mirekdlugosz: `external_registry_1` is still required in robottelo.properties, as other tests use it.
Other than that, as far as I am aware, there is no downside of having key in robottelo.properties that robottelo no longer uses or recognizes.

omkarkhatavkar: I thought maybe you could miss it, that's fine.

PR #7478, opened 2019-11-13
Remove docker container, registry and compute resource

mirekdlugosz:
These were removed from Satellite during 6.6 cycle, but we agreed to remove them from frameworks in release after 6.6.

- removed CLI hammer docker container and hammer docker registry
- removed leftover API and CLI tests
- removed leftover references in constants
- removed some config keys:
  - docker.external_url - used only to create docker compute resource
  - docker.unix_socket - used only in docker compute resources
  - docker.external_registry_2 not used since 7f74af32cd15208c5150b01a937b790bdd4e611d, used only in docker compute resource

See also https://github.com/SatelliteQE/airgun/pull/421 and https://github.com/SatelliteQE/nailgun/pull/673 . Technically, nailgun should be merged only after this one is merged, so we avoid importing entities that are not there anymore.


robottelo.properties.sample
```diff
@@ -190,29 +190,13 @@ ssh_key=
...
-# in other words, the daemon is initialized with `--host tcp://0.0.0.0:<port>`.
-# This will be used by tests that test external docker daemon.
-# external_url=http://localhost:2375
-# These will be used by tests that exercise external docker registries. Second
-# one will be used by tests that validate updating from one docker registry url
-# to another
-# External docker registries urls in the format http[s]://<server>[:<port>].
+# This will be used by tests that exercise external docker registries.
+# External docker registry url in the format http[s]://<server>[:<port>].
 # external_registry_1=
```
ntkathole: external_registry_1= is still useful?

mirekdlugosz: Yes, it is used by product tests (when you add repo with containers).

I thought about renaming it to just `external_registry`, but then I would have to update jenkins config and ask everyone to update their local config. I decided effort is not worth it for so superficial change.

mirekdlugosz:
@SatelliteQE/quality-engineers It's been almost two weeks now, can I have another review and merge?

pgagne:
Whats up with the codecov check not passing, do we know what the issue is?

mirekdlugosz:
@pgagne This measures coverage of "lines adjusted in the pull request" ([source](https://docs.codecov.io/docs/commit-status#section-patch-status)). When I removed part of the line, git records that as addition and removal, and codecov complains this particular added line is not covered by any test.

PR #7481, opened 2019-11-14
Discovery BZ:1638403 automated along with first pxe discovery API test

jyejare:
Automated:

- test_positive_provision_pxe_host
```
=============================== 1 passed, 19 deselected in 807.79 seconds ==============================
```
- test_positive_provision_pxe_host_non_admin (BZ 1638403)
```
=============================== 1 passed, 20 deselected in 443.69 seconds ===============================
```

### DIFF
--- a/robottelo/api/utils.py
+++ b/robottelo/api/utils.py
@@ -944,3 +944,53 @@ def apply_package_filter(content_view, repo, package, inclusion=True):
     content_view = content_view.read()
     content_view_version_info = content_view.version[0].read()
     return content_view_version_info
+
+
+def create_org_admin_role(orgs, locs, name=None):
+    """Helper function to create org admin role for particular
+    organizations and locations by cloning 'Organization admin' role.
+
+    :param list orgs: The list of organizations for which the org admin is
+        being created
+    :param list locs: The list of locations for which the org admin is
+        being created
+    :param str name: The name of cloned Org Admin role, autogenerates if None provided
+    :return dict: The object of ```nailgun.Role``` of Org Admin role.
+    """
+    name = gen_string('alpha') if not name else name
+    default_org_admin = entities.Role().search(
+        query={'search': u'name="Organization admin"'})
+    org_admin = entities.Role(id=default_org_admin[0].id).clone(
+        data={
+            'role': {
+                'name': name,
+                'organization_ids': orgs or [],
+                'location_ids': locs or []
+            }
+        }
+    )
+    return entities.Role(id=org_admin['id']).read()
+
+
+def create_org_admin_user(orgs, locs):
+    """Helper function to create an Org Admin user by assigning org admin role and assign
+    taxonomies to Role and User
+
+    The taxonomies for role and user will be assigned based on parameters of this function
+
+    :return User: Returns the ```nailgun.entities.User``` object with passwd attr
+    """
+    # Create Org Admin Role
+    org_admin = create_org_admin_role(orgs=orgs, locs=locs)
+    # Create Org Admin User
+    user_login = gen_string('alpha')
+    user_passwd = gen_string('alphanumeric')
+    user = entities.User(
+        login=user_login,
+        password=user_passwd,
+        organization=orgs,
+        location=locs,
+        role=[org_admin.id]
+    ).create()
+    user.passwd = user_passwd
+    return user
--- a/tests/foreman/api/test_discoveredhost.py
+++ b/tests/foreman/api/test_discoveredhost.py
@@ -13,14 +13,23 @@
 
 :Upstream: No
 """
+import time
+
 from fauxfactory import gen_string, gen_ipaddr, gen_mac
 from nailgun import entities
+from nailgun.config import ServerConfig
+from robottelo import ssh
+from robottelo.api.utils import create_org_admin_user
+from robottelo.cli.factory import configure_env_for_provision
 from robottelo.datafactory import valid_data_list
 from robottelo.decorators import (
+    skip_if_not_set,
     stubbed,
     tier2,
     tier3,
 )
+from robottelo.helpers import get_nailgun_config
+from robottelo.libvirt_discovery import LibvirtGuest
 from robottelo.test import APITestCase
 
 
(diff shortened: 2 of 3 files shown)


tests/foreman/api/test_discoveredhost.py
```diff
@@ -55,9 +61,94 @@ def _create_discovered_host(name=None, ipaddress=None, macaddress=None):
...
+
+        Introduced a delay of 300secs by polling every 10 secs to get expected
+        host
+        """
+        for _ in range(30):
+            discovered_host = entities.DiscoveredHost().search(
+                query={'search': 'name={}'.format(hostname)})
+            if discovered_host:
+                break
+            time.sleep(10)
```
mirekdlugosz: How about using `wait_for` here?
We pull it in as dependency and have couple of tests using it already.

tests/foreman/api/test_discoveredhost.py
```diff
@@ -55,9 +61,94 @@ def _create_discovered_host(name=None, ipaddress=None, macaddress=None):
...
+        """
+        super(DiscoveryTestCase, cls).setUpClass()
+
+        # Build PXE default template to get default PXE file
+        entities.ConfigTemplate().build_pxe_default()
+        # let's just modify the timeouts to speed things up
+        ssh.command("sed -ie 's/TIMEOUT [[:digit:]]\\+/TIMEOUT 1/g' "
+                    "/var/lib/tftpboot/pxelinux.cfg/default")
+        ssh.command("sed -ie '/APPEND initrd/s/$/ fdi.countdown=1/' "
+                    "/var/lib/tftpboot/pxelinux.cfg/default")
```
mirekdlugosz: Is such modification of host files safe? Isn't there a risk that something else will run at the same time or later, but will require different values?

rplevka: @mirekdlugosz fair point. However, I'd consider `fdi.countdown=1` a default for our automation. It'd be probably better to set this timeout on a global level as a part of the provisioning job.
Anyway, the risk is not that big (we have no aspirations on testing different pxelinux configs for now)

tests/foreman/api/test_discoveredhost.py
jyejare: I see the existing implementation is much easier and with less efforts(LOC) than using wait_for like this :
```
        try:
            wait_for(
                lambda: entities.DiscoveredHost(user_config or default_config).search(
                    query={'search': 'name={}'.format(hostname)}),
                timeout=300,
                delay=10,
                logger=self.logger
            )
        except TimeoutError:
            raise HostNotDiscoveredException(
                "The host {} is not discovered within 300 seconds".format(hostname))
        return entities.DiscoveredHost(user_config or default_config).search(
                    query={'search': 'name={}'.format(hostname)})[0]
```

jyejare: To save 10 lines, I am ending up writing 13 lines :) with no extra benefits.

tests/foreman/api/test_discoveredhost.py
jyejare: Yes, Its less likely that other tests are using this setting than Discovery and even if its provisioning then its 1 as @rplevka said.

@rplevka We won't do that at the global level as we will end up two different component tests dependent on single settings and hence change in either one component in one discovery/provisioning could be the cause for another to fail.

So let's handle those settings independently and as per the convinience of components.

jyejare:
@rplevka @mirekdlugosz Thanks for reviewing my PR guys but my PR was in progress while you reviewed it. There are additional changes and new test case automated.

Removing the DO_NOT_MERGE label now to unblock review and merging. :)

tests/foreman/api/test_discoveredhost.py
mirekdlugosz: `wait_for` returns call result, so this would be more like:
```python
try:
    res = wait_for(lambda: entities.DiscoveredHost.search(), *args)
except:
    raise HostNotDiscoveredException(...)
return res['out'][0]
```
And then, wait_for already raises exception, so unless you have to rewrite it, you could get away with:
```python
res = wait_for(lambda: entities.DiscoveredHost.search(), *args)
return res['out'][0]
```

Of course it's your choice to make, that was only a suggestion.

jyejare: Just observed, It's not ```TimeoutError``` its ```TimedoutError``` for which the 'wait_for' doesnt provides an exception to be handled. And I dont want to miss any other bug by simply doing except :(
Complexed filling.

Keeping as it is for now @mirekdlugosz , let me know if you wanna go hard on it. Thanks for the comments though.

mirekdlugosz: No, use whichever way you prefer. This PR is ACK from me already, with or without `wait_for`.

jyejare:
@rplevka Need your approvel.

tests/foreman/api/test_discoveredhost.py
```diff
@@ -163,6 +255,75 @@ def test_positive_provision_pxe_host(self):
...
+            assert provisioned_host.operatingsystem.read(
+                ).title == self.configured_env['os']['title']
+            assert not entities.DiscoveredHost().search(
+                query={'search': 'name={}'.format(discovered_host.name)})
+
+    @tier3
+    def test_positive_provision_pxe_host_non_admin(self):
+        """Provision a pxe-based discovered hosts
+
+        :id: e805b9c5-e8f6-4129-a0e6-ab54e5671ddb
```
ntkathole: duplicate id

tests/foreman/api/test_discoveredhost.py
```diff
@@ -163,6 +255,75 @@ def test_positive_provision_pxe_host(self):
...
+
+        :Setup: Provisioning should be configured and a host should be
+            discovered
+
+        :Steps: PUT /api/v2/discovered_hosts/:id
+
+        :expectedresults: Host should be provisioned successfully
+
+        :CaseAutomation: notautomated
```
ntkathole: automated

tests/foreman/api/test_discoveredhost.py
```diff
@@ -163,6 +255,75 @@ def test_positive_provision_pxe_host(self):
 
         :CaseImportance: Critical
         """
+        if not self.configured_env:
```
ntkathole: :CaseAutomation: automated

robottelo/api/utils.py
```diff
@@ -944,3 +945,66 @@ def apply_package_filter(content_view, repo, package, inclusion=True):
...
+        password=user_passwd,
+        organization=orgs,
+        location=locs,
+        role=[org_admin.id]
+    ).create()
+    user.passwd = user_passwd
+    return user
+
+
+def create_user_config(user):
```
ntkathole: Isn't from robottelo.helpers import get_nailgun_config helper useful ?

jyejare: get_nailgun_config returns the config for the default / admin user. But here it will return a config for the specified user.

There is a parameter for ```user``` in this helper.

So for now, I am modifying the existing get_nailgun_config to suit my needs without affecting the existing functionality of returning admin config.

jyejare:
@ntkathole Fixed silly mistakes done by me and PR is ready for your another look. Thanks !

jyejare:
@ntkathole , Here it is :

```
$ py.test -s tests/foreman/api/test_discoveredhost.py -k test_positive_provision_pxe_host_non_admin
collecting ... 2019-11-25 10:29:00 - conftest - DEBUG - Collected 21 test cases

.
=============================== 1 passed, 20 deselected in 419.36 seconds ==============================
```

jyejare:
@ntkathole , Here is external Test from this PR not affected by get_nailgun_config change:

```
$ py.test tests/foreman/api/test_contentview.py -k test_positive_readonly_user_actions
collecting ... 2019-11-25 10:47:26 - conftest - DEBUG - Collected 68 test cases

collected 68 items / 67 deselected / 1 selected                                                         

tests/foreman/api/test_contentview.py .                                                           [100%]

=============================== 1 passed, 67 deselected in 113.11 seconds ===============================
```

But TBH with you, in all the tests this function is just used to fetch the url and verify=False, later they are updating the auth. All of them can be replaced with my new function. :))

PR #7497, opened 2019-11-20
Add endtoend tests to run in automation

ntkathole:


mirekdlugosz:
Could you elaborate a little on this change and why it is needed?

I think we used to have separate Jenkins job for endtoend tests. I take that we don't anymore?
Wouldn't it be better to move these tests to `tests/api` or `tests/cli` and remove endtoend dir?

ntkathole:
> Could you elaborate a little on this change and why it is needed?
> 
> I think we used to have separate Jenkins job for endtoend tests. I take that we don't anymore?
> Wouldn't it be better to move these tests to `tests/api` or `tests/cli` and remove endtoend dir?

Earlier tests were run on only upgrade automation,not sure why we didn't add them in past to tier automation. Tests catches my attention while working on pipeline, these are only 11 tests and it is not worth to have separate job, endpoint and test run in polarion as well. So I thought to include them in tier automation and remove separate job.

ntkathole:
@rplevka This PR is focused on making them part of automation. If anyone want to refactor them or change the directory or make them smoke tests, we can do it in separate PR.  
The tests were renamed from smoke to endtoend in https://github.com/SatelliteQE/robottelo/pull/3382

rplevka:
the original purpose of having these in a separate job was to block the release pipeline. As i said - smoke tests.
Now we're trying to blend them in together with tierX tests. They use different style than the tierX tests. I think we really need a proper discussion on what we want to achieve, as the whole automation is starting to be a bunch of various tests executed at the same time.

I'm not going to block this PR, if you get another ACK, feel free to merge it, I just wanted to raise my concern.

ntkathole:
@SatelliteQE/quality-engineers can someone look into this? Discussion is on satqe list.

PR #7503, opened 2019-11-25
Added check around last sat version capsule/tools repos

vijay8451:
- Added check to check last version Capsule/Tools repos exist or not .. i.e. bug https://bugzilla.redhat.com/show_bug.cgi?id=1776108

### DIFF
--- a/tests/foreman/ui/test_repository.py
+++ b/tests/foreman/ui/test_repository.py
@@ -889,10 +889,11 @@ def test_positive_recommended_repos(session, module_org):
         session.organization.select(module_org.name)
         rrepos_on = session.redhatrepository.read(recommended_repo='on')
         assert REPOSET['rhel7'] in [repo['name'] for repo in rrepos_on]
-        last_sat_version = '{}.{}'.format(get_sat_version().release[0],
-                                          get_sat_version().release[1] - 1)
-        assert 0 == len([repo['name'] for repo in rrepos_on if last_sat_version in repo[
-            'name']]), 'Last Satellite version Capsule/Tools repos are exist'
+        sat_version = get_sat_version().public
+        cap_tool_repos = [repo['name'] for repo in rrepos_on if 'Tools' in repo['name'] or
+                                                                'Capsule' in repo['name']]
+        cap_tools_repos = [repo for repo in cap_tool_repos if repo.split()[4] != sat_version]
+        assert not cap_tools_repos, 'Tools/Capsule repos do not match with Satellite version'
         rrepos_off = session.redhatrepository.read(recommended_repo='off')
         assert REPOSET['fdrh8'] in [repo['name'] for repo in rrepos_off]
         assert len(rrepos_off) > len(rrepos_on)
--- a/tests/foreman/ui/test_repository.py
+++ b/tests/foreman/ui/test_repository.py
@@ -55,6 +55,7 @@ from robottelo.decorators import (
     upgrade
 )
 from robottelo.helpers import read_data_file
+from robottelo.host_info import get_sat_version
 from robottelo.products import SatelliteToolsRepository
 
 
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_repository.py
```diff
@@ -867,21 +868,31 @@ def test_positive_delete_random_docker_repo(session, module_org):
...
+
+    :BZ: 1776108
     """
     manifests.upload_manifest_locked(module_org.id)
     with session:
         session.organization.select(module_org.name)
         rrepos_on = session.redhatrepository.read(recommended_repo='on')
         assert REPOSET['rhel7'] in [repo['name'] for repo in rrepos_on]
+        last_sat_version = '{}.{}'.format(get_sat_version().release[0],
+                                          get_sat_version().release[1] - 1)
```
mirekdlugosz: This won't work for first version with increased major number. I.e. for 7.0, it will produce 7.-1 instead of 6.X.

1. I think `Version` class has comparison overloads, so you can parse version from UI and verify that version in UI is lesser than version we have.
2. Do we even care if version is lower? Probably we want to ensure that version in UI is the same as version of Satellite. Doesn't matter in UI has lower or higher, if it's different, it's a bug.

vijay8451: Thanks @mirekdlugosz for your thoughts/input ...

> This won't work for first version with increased major number. I.e. for 7.0, it will produce 7.-1 instead of 6.X.

It should be fine now with latest commit

>     1. I think `Version` class has comparison overloads, so you can parse version from UI and verify that version in UI is lesser than version we have.

I found `Version/get_sat_version' slightly fast hence used that ..

>     2. Do we even care if version is lower? Probably we want to ensure that version in UI is the same as version of Satellite. Doesn't matter in UI has lower or higher, if it's different, it's a bug.

Yes we are taking care of the lower version of Satellite here because current Satellite `Capsule/Tools repos` only comes as part of manifest when content pushed to CDN after Beta/GA .. so test will keep fail until that.
thus adding reverse check does help to report the issue/bug at initial stage (as mentions in bug) and test will keep pass once that fix.

tests/foreman/ui/test_repository.py
```diff
@@ -867,21 +868,30 @@ def test_positive_delete_random_docker_repo(session, module_org):
 
 
 @tier2
+# @pytest.mark.skip_if_open("BZ:1776108")
```
omkarkhatavkar: why did this comment?

vijay8451: ahh... it's uncommented now :)

tests/foreman/ui/test_repository.py
mirekdlugosz: @vijay8451 well, current approach will only work for 6.9 -> 7.0 transition. If we get from 6.8 to 7.0, or from 6.10 to 7.0, it will still fail.

I would still advice using Version, performance impact notwithstanding (it's UI test, I'm pretty sure couple of `Version()` calls are not the biggest waste), or change assertion from "there is no repo for previous version of Sat" to "there are only repos from current version of Sat". But if you are committed to current approach, I won't block you.

vijay8451: > @vijay8451 well, current approach will only work for 6.9 -> 7.0 transition. If we get from 6.8 to 7.0, or from 6.10 to 7.0, it will still fail.

Yeah, But we never no 6.x will go till what .. so have to keep test in either `-1 version` or `hardcoded` ... 
Please let me know if you see any other better approach here ?

> I would still advice using Version, performance impact notwithstanding (it's UI test, I'm pretty sure couple of `Version()` calls are not the biggest waste), or change assertion from "there is no repo for previous version of Sat" to "there are only repos from current version of Sat". But if you are committed to current approach, I won't block you.

> in this check the idea is to keep a reverse check using last version of sat.

vijay8451: @mirekdlugosz I have slightly change the last version dependency (-1) in new push to avoid the such scenarios i.e.  ` If we get from 6.8 to 7.0, or from 6.10 to 7.0, it will still fail`and it would work in any approach now . 
please revisit .

tests/foreman/ui/test_repository.py
```diff
@@ -867,21 +868,33 @@ def test_positive_delete_random_docker_repo(session, module_org):
...
     manifests.upload_manifest_locked(module_org.id)
     with session:
         session.organization.select(module_org.name)
         rrepos_on = session.redhatrepository.read(recommended_repo='on')
         assert REPOSET['rhel7'] in [repo['name'] for repo in rrepos_on]
+        sat_version = get_sat_version().public
+        cap_tool_repos = [repo['name'] for repo in rrepos_on if 'Tools' in repo['name'] or
+                                                                'Capsule' in repo['name']]
+        assert 0 == len([repo for repo in cap_tool_repos if repo.split()[4] < sat_version or repo.
+                        split()[4] > sat_version]), 'Tools/Capsule repos do not' \
```
mirekdlugosz: `num < sat_version or num > sat_version` can be simplified to `num != sat_version`, can't it?

Given how long this line becomes, I would extract this list comprehension to separate variable and assert that:
```python
cap_tools_repos = [repo for repo in cap_tool_repos if repo.split()[4] != sat_version]
assert not cap_tools_repos, "Tools/Capsule repos do not match with Satellite version"
```

vijay8451: Push a new commit , please revisit.

PR #7473, opened 2019-11-12
Remove output_format='plain'

swadeley:
There is no "hammer --output=plain" in help and other docs, and it fails in 6.7

PR #7510, opened 2019-11-27
Add back 'plain' output_format option

swadeley:
* 'plain' was removed because hammer stopped accepting it.
   However, some non-hammer tests use it (in the code, to signal ssh.py, not on the command line)

 Remove output_format='plain' #7473

### DIFF
--- a/robottelo/ssh.py
+++ b/robottelo/ssh.py
@@ -424,8 +424,8 @@ def execute_command(cmd, connection, output_format=None, timeout=None,
         # Convert to unicode string and remove all color codes characters
         stderr = regex.sub('', decode_to_utf8(stderr))
         logger.info('<<< stderr\n%s', stderr)
-    # we don't want a list as output of 'base' just pure text
-    if stdout and output_format not in ('json', 'base'):
+    # Skip converting to list if 'plain', or the hammer options 'json' or 'base' are passed
+    if stdout and output_format not in ('json', 'base', 'plain'):
         # Mostly only for hammer commands
         # for output we don't really want to see all of Rails traffic
         # information, so strip it out.


swadeley:
robottelo]$ pytest  tests/foreman/cli/test_logging.py -k proxy
========================= test session starts ================================
platform linux -- Python 3.7.5, pytest-4.6.3, py-1.8.0, pluggy-0.13.1
rootdir: /sat6/robottelo
plugins: cov-2.8.1, repeat-0.8.0, services-1.3.1, xdist-1.30.0, mock-1.10.4, forked-1.0.2
collecting ... 2019-11-27 16:31:05 - conftest - DEBUG - Collected 4 test cases
============ 4 passed in 73.82 seconds =========

mirekdlugosz:
> However, some non-hammer tests use it.

Could you explain what are these tests, and what's stopping us from updating them to use `base` instead of `plain`?

swadeley:
I could just use 'base' in my test, but that has no meaning as I am using a bash command not a hammer command. Other's tests would still fail.
For example, in tests/foreman/cli/test_hammer.py:
in this code:  
       `raw_output = ssh.command('hammer full-help', output_format='plain').stdout`
 'plain' does not go to the command line, it only signals ssh.py not to make stdout into a list.

    `[root@dell-r330-12 ~]# hammer full-help --base
     Error: Unrecognised option '--base'.

     See: 'hammer full-help --help'.
    [root@dell-r330-12 ~]# hammer full-help --plain
    Error: Unrecognised option '--plain'.

    See: 'hammer full-help --help'.`

swadeley:
In https://github.com/SatelliteQE/robottelo/blob/181f66dad2f0d8342037e8d575163236a30744a9/tests/foreman/sys/test_rename.py#L182

in this code:
    `result = connection.run(
                'satellite-change-hostname -y \
                        {0} -u {1} -p {2}'.format(
                    hostname, self.username, self.password),
                output_format='plain'
            )
`

it would not be obvious what 'base' meant

swadeley:
I should improve this:
`     # we don't want a list as output of 'base' just pure text
`

swadeley:
I used it here too:
https://github.com/SatelliteQE/robottelo/blob/181f66dad2f0d8342037e8d575163236a30744a9/tests/foreman/sys/test_katello_certs_check.py#L116

swadeley:
> Honestly, I find this whole thing confusing. As if initially this `output_format` was for hammer only, but at one point it started living its own life. Perhaps it would help to draw a clear line between them.
> 
It is confusing. The code was written to convert stdout to a list format unless you or the hammer command uses one of those keywords.
> Can you only verify that comments around this line, and in docstring for function, are correct? And update them if that's necessary?
I improved the text a bit, I think that is all we have time for now as its year end.

Thank you

Issue #7514, opened 2019-11-29
hammer parser have problems with host info, discussion how to change hammer parser


tstrych:
TLDR: I would like to create some discussion how to change hammer parser ? And after this change all cli tests probably needs to be runned and fixed if the parsed structure will change. 

Imagine this output from host info:
```
input = [
'Subscription Information: ',
'    Release Version:               ',
'    Autoheal:                      true',
'    Registered by Activation Keys: ',
'     1) rafkIs9iuG',
'    System Purpose:                ',
'        Service Level:  ',
]
```

The function [parse_info](https://github.com/SatelliteQE/robottelo/blob/ec7f27d5af391f03b8a962b73d2b86dce720d22c/robottelo/cli/hammer.py#L188) will be called on this input. This input value is just part of output from `Host.info` command.

This will end with:
```
contents[sub_prop][key] = value
TypeError: list indices must be integers or slices, not str
```
This issue cause a lot of problems in our tests. Since `System purpose` fields were added to host info.

What happens inside [parse_info](https://github.com/SatelliteQE/robottelo/blob/ec7f27d5af391f03b8a962b73d2b86dce720d22c/robottelo/cli/hammer.py#L188) function:

Some examples:
Input
```
[
'Subscription Information: ',
'    Release Version:               ',
'    Autoheal:                      true',
'    Registered by Activation Keys: ',
]
```

Output:
`{'subscription-information': {'release-version': '', 'autoheal': 'true', 'registered-by-activation-keys': ''}}`
Looks fine

Then append this line at the end of Input
```
'     1) rafkIs9iuG',
```
Output: `{'subscription-information': ['rafkIs9iuG']}`
Here you can see all registered-by-activation-keys was rewritten by list. 
[Here in code](https://github.com/SatelliteQE/robottelo/blob/ec7f27d5af391f03b8a962b73d2b86dce720d22c/robottelo/cli/hammer.py#L234)

So I assume the problem was there before `System purpose` fields was added, just addition of these fields makes them visible for our tests.
Probably we don't have a test which verifies information about `subscription-information`. Therefore we didn't find this problem sooner.

And finally when we add another line with [indentation level](https://github.com/SatelliteQE/robottelo/blob/ec7f27d5af391f03b8a962b73d2b86dce720d22c/robottelo/cli/hammer.py#L164) = 1.
In our example append this line to Input. `'    System Purpose:                ',`
And output is gonna be `Typeerror` which is upper in this issue. 

[...]

PR #7522, opened 2019-12-04
hammer parser, wrong insertion to dictionary when value is a list

tstrych:
Connected to issue: #7514 

I think this needs a bit of testing, as it can change every `hammer entity info output`. 
But I didn't find any input, output hammer_parser test.


rplevka:
I think this is a good candidate for `tests/robottelo`

PR #7523, opened 2019-12-04
Invalid AK test case

ColeHiggins2:
Based on BZ:1156555 with a won't fix status, deleting test case.


ColeHiggins2:
@mirekdlugosz I've pushed the changes but Travis still has not kicked off.

PR #7528, opened 2019-12-09
added support for ldap OTP based login tests cases

omkarkhatavkar:
### PR Description 
added support for LDAP OTP based login tests cases

### Test Result 
```
(env) [root@okhatavk robottelo]# pytest tests/foreman/ui/test_ldap_authentication.py -k 'otp'
platform linux -- Python 3.7.2, pytest-4.6.3, py-1.8.0, pluggy-0.13.0
plugins: cov-2.8.1, xdist-1.30.0, mock-1.10.4, services-1.3.1, forked-1.1.1
collecting ... 2019-12-09 06:51:36 - conftest - DEBUG - Collected 15 test cases
2019-12-09 12:21:37 - robottelo.helpers - DEBUG - Generated file bz_cache.json with BZ collect data
collected 15 items / 13 deselected / 2 selected                                                                                                                             

tests/foreman/ui/test_ldap_authentication.py ..                                                                                                                       [100%]

================================================================= 2 passed, 13 deselected in 118.12 seconds =================================================================
 
```

### DIFF
--- a/tests/foreman/ui/test_ldap_authentication.py
+++ b/tests/foreman/ui/test_ldap_authentication.py
@@ -169,6 +169,15 @@ def generate_otp(secret):
     return time_otp.now()
 
 
+def ldap_tear_down():
+    ldap = entities.AuthSourceLDAP().search()
+    for ldap_auth in range(len(ldap)):
+        users = entities.User(auth_source=ldap[ldap_auth]).search()
+        for user in range(len(users)):
+            users[user].delete()
+        ldap[ldap_auth].delete()
+
+
 @tier2
 def test_positive_end_to_end_ad(session, ldap_data, ldap_auth_name):
     """Perform end to end testing for LDAP authentication component with AD
@@ -796,6 +805,7 @@ def test_positive_login_ad_user_basic_roles(
         assert ldapsession.architecture.search(name)[0]['Name'] == name
 
 
+@upgrade
 @tier2
 def test_positive_login_user_password_otp(test_name, ipa_data, auth_source_ipa):
     """Login with password with time based OTP
@@ -809,18 +819,21 @@ def test_positive_login_user_password_otp(test_name, ipa_data, auth_source_ipa):
     :expectedresults: Log in to foreman UI successfully
 
     """
-    password_with_otp = "{0}{1}".format(
-        ipa_data['ldap_ipa_user_passwd'],
-        generate_otp(ipa_data['time_based_secret']))
-    with Session(
-            test_name,
-            ipa_data['ipa_otp_username'],
-            password_with_otp
-    ) as ldapsession:
-        with raises(NavigationTriesExceeded):
-            ldapsession.user.search('')
-        expected_user = "{} {}".format(ipa_data['ipa_otp_username'], ipa_data['ipa_otp_username'])
-        assert ldapsession.task.read_all()['current_user'] == expected_user
+    try:
+        password_with_otp = "{0}{1}".format(
+            ipa_data['ldap_ipa_user_passwd'],
+            generate_otp(ipa_data['time_based_secret']))
+        with Session(
+                test_name,
+                ipa_data['ipa_otp_username'],
+                password_with_otp
+        ) as ldapsession:
+            with raises(NavigationTriesExceeded):
+                ldapsession.user.search('')
+            expected_user = "{} {}".format(ipa_data['ipa_otp_username'], ipa_data['ipa_otp_username'])
+            assert ldapsession.task.read_all()['current_user'] == expected_user
+    finally:
+        ldap_tear_down()
 
 
 @tier2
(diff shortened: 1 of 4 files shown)


tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -763,3 +794,55 @@ def test_positive_login_ad_user_basic_roles(
...
+    with Session(
+            test_name,
+            ipa_data['ipa_otp_username'],
+            password_with_otp
+    ) as ldapsession:
+        with raises(NavigationTriesExceeded):
+            ldapsession.user.search('')
+        expected_user = "{} {}".format(ipa_data['ipa_otp_username'], ipa_data['ipa_otp_username'])
+        assert ldapsession.task.read_all()['current_user'] == expected_user
+
```
ntkathole: login with admin user and verifying otp user is created or not will be right assert. Please add it as well. Same below.

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -763,3 +794,55 @@ def test_positive_login_ad_user_basic_roles(
             ldapsession.usergroup.search('')
         ldapsession.architecture.create({'name': name})
         assert ldapsession.architecture.search(name)[0]['Name'] == name
+
+
+@tier2
+def test_positive_login_user_password_otp(test_name, ipa_data, auth_source_ipa):
```
ntkathole: add teardown to delete the user and auth source

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -763,3 +794,55 @@ def test_positive_login_ad_user_basic_roles(
             ldapsession.usergroup.search('')
         ldapsession.architecture.create({'name': name})
         assert ldapsession.architecture.search(name)[0]['Name'] == name
+
+
+@tier2
```
ntkathole: add upgrade decorator as well

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -763,3 +803,68 @@ def test_positive_login_ad_user_basic_roles(
...
+        password_with_otp = "{0}{1}".format(
+            ipa_data['ldap_ipa_user_passwd'],
+            generate_otp(ipa_data['time_based_secret']))
+        with Session(
+                test_name,
+                ipa_data['ipa_otp_username'],
+                password_with_otp
+        ) as ldapsession:
+            with raises(NavigationTriesExceeded):
+                ldapsession.user.search('')
```
mirekdlugosz: I don't understand this. Why do we expect to see exception in positive path? Why do we search for empty string among users after logging in?

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -763,3 +803,68 @@ def test_positive_login_ad_user_basic_roles(
...
+            ipa_data['ldap_ipa_user_passwd'],
+            gen_string(str_type='numeric', length=6))
+        with Session(
+                test_name,
+                ipa_data['ipa_otp_username'],
+                password_with_otp
+        ) as ldapsession:
+            with raises(NavigationTriesExceeded) as error:
+                ldapsession.user.search('')
+            assert error.typename == "NavigationTriesExceeded"
```
mirekdlugosz: This assertion is already done by [`pytest.raises`](https://docs.pytest.org/en/latest/reference.html#pytest-raises). If block within context raises different exception than specified, or does not raise exception at all, pytest will fail the test.

omkarkhatavkar: That's okay right, it should only lead same error otherwise there has to be a problem with the user able to login or suppress.

tests/foreman/ui/test_ldap_authentication.py
omkarkhatavkar: There are multiple tests are using the same code. https://github.com/SatelliteQE/robottelo/blob/45fe1205f71a2956d6544cef8508360227f552b2/tests/foreman/ui/test_ldap_authentication.py#L777, https://github.com/SatelliteQE/robottelo/blob/45fe1205f71a2956d6544cef8508360227f552b2/tests/foreman/ui/test_ldap_authentication.py#L753 let me check what could be the exact reason.

omkarkhatavkar: I guess the unless we fire any command e.g. ldapsession.user.search('') browser will not et open? that is only one reason I could guess

omkarkhatavkar: let me check by removing that code as we have next step where we are reading the task details.

mirekdlugosz: Browser is opened on first request to entity, so in your case if you remove context manager and proceed directly to `ldapsession.task.read_all()`, it will still open browser.
Hacks like that would only be needed if you want to interact directly with `session.browser` before you interact with any entity (see "Browser needs to launch to grab datetime() for test case" on satqe-list posted around two weeks ago).

From what I see, in tests that you referenced this code tries to verify that logged in user doesn't have access to specific parts of Satellite. So that would be an answer to my question. Next thing is - do you need such assertion in your test? Seems you are more interested in login process itself than permissions after user logs in.

omkarkhatavkar: my thought to verify that user got logged-in and has correct permissions.

mirekdlugosz: OK

mirekdlugosz:
ACK

PR #7535, opened 2019-12-11
Updated assert based on BZ#1778055

vijay8451:
- Updated assert based on BZ:1778055
- Add `upgrade` decorator.
- Test results :
```
PASSED2019-12-11 10:15:56 - robottelo - DEBUG - Finished Test: RepositoryTestCase/test_positive_srpm_list_end_to_end
```

### DIFF
--- a/tests/foreman/cli/test_repository.py
+++ b/tests/foreman/cli/test_repository.py
@@ -1889,6 +1889,7 @@ class RepositoryTestCase(CLITestCase):
         })
         assert int(Repository.info({'id': new_repo['id']})['content-counts']['source-rpms']) == 0
 
+    @upgrade
     @tier2
     def test_positive_srpm_list_end_to_end(self):
         """Create repository,  upload, list and remove an SRPM content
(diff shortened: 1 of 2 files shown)


tests/foreman/cli/test_repository.py
```diff
@@ -1937,7 +1937,7 @@ def test_positive_srpm_list_end_to_end(self):
         # Remove uploaded SRPM
         Repository.remove_content({
             'id': new_repo['id'],
-            'ids': [Srpm.list()[0]['id']],
+            'ids': [Srpm.list({'repository-id': new_repo['id']})[0]['id']],
```
omkarkhatavkar: what is the difference between id and id's? is that both required while removing the content?

If I'm trying to understand here if I removed the repository it does not deletes the all associated Srpms, I could be wrong?

vijay8451: Ids => We could pass multiple SRPM packages IDs under it .
id=> id filtering the first package ID only in that repo.

omkarkhatavkar: means we can have the privilege to delete specific srpms from the uploaded repo?

vijay8451: yes, this is how it works for every content i.r. rpm/file/ ..etc

omkarkhatavkar: not for module streams :)

vijay8451: I mean content which can upload same delete as well :) ... I am not sure if module streams can upload manual :)

PR #7543, opened 2019-12-13
Get sesh

rplevka:
This PR extends the `get_web_session` helper so it accepts custom credentials - this is required for UI tests as they all create their own user.

Also, it makes the `session` fixture to use it, so all the tests will start skipping the UI login.


rplevka:
```
<accent="Arnie">
    I'll be BACK!
</accent>
```

PR #7546, opened 2019-12-16
Add test for Ansible Tower Inventory syncing

vsedmik:
Report template named 'Ansible Inventory' (default name is specified in settings) must be present in Satellite 6.7 and later in order to provide enhanced functionality for Ansible Tower inventory synchronization with Satellite.
This test checks that the proper template is present on the Satellite side and that it is renderable by a user with minimal required permissions.

### DIFF
--- a/tests/foreman/cli/test_reporttemplates.py
+++ b/tests/foreman/cli/test_reporttemplates.py
@@ -35,6 +35,7 @@ from robottelo.cli.location import Location
 from robottelo.cli.org import Org
 from robottelo.cli.report_template import ReportTemplate
 from robottelo.cli.user import User
+from robottelo.cli.settings import Settings
 from robottelo.constants import (
     DEFAULT_LOC,
     DEFAULT_ORG,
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_reporttemplates.py
```diff
@@ -693,3 +694,70 @@ def test_positive_generate_with_name_and_org(self):
...
+            1. A user with minimal required permissions: 'Ansible Tower Inventory Reader' role
+            2. A fake host to be checked in report output
+
+        :steps:
+            1. Check settings for default Ansible Inventory template name and ensure
+               the template is present
+            2. Try to render the template using the user with ATIR role
+            3. Check the fake host is present in the output
+
+        :expected result: Report template is present, renederable and provides output
```
mirekdlugosz: It's `:expectedresults:` - without space and in plural. Right now it causes travis failure.

tests/foreman/cli/test_reporttemplates.py
```diff
@@ -693,3 +694,70 @@ def test_positive_generate_with_name_and_org(self):
...
+        user = make_user({
+                'login': login,
+                'password': password,
+                'organization-ids': org['id'],
+                'location-ids': loc['id']
+        })
+
+        User.add_role({
+            'login': user['login'],
+            'role': 'Ansible Tower Inventory Reader',
```
mirekdlugosz: Usually we create new role, define new filter for that role with necessary permissions and use that. Probably for us to not be tied to specific string, which may change in the future.

I'm not saying you have to change this, I'm only pointing out difference from established pattern.

tests/foreman/cli/test_reporttemplates.py
vsedmik: Thank you for pointing out, I have overlooked this.

tests/foreman/cli/test_reporttemplates.py
vsedmik: I agree, I don't consider using constant strings in code as good practice too. However, in this case the referenced role is shipped as predefined role within Satellite to be used for rendering the template, so I would like to test this explicit role too. Does moving the string to the 'robottelo.constants' make sense?

mirekdlugosz: Nah, I would just leave it as is.
New role and filter makes it unnecessary to commit when existing role name change. If change of role name forces us to submit PR, then I don't see much difference between changing test itself and changing constants.
Plus constants turned out to be kitchen sink and adding one more random value used by single test won't really help anyone.

PR #7552, opened 2019-12-18
Added pre-upgrade and post-upgrade test scenarios for performance & Tuning

devendra104:


tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,153 @@
...
+
+:Upstream: No
+"""
+
+from upgrade_tests import post_upgrade, pre_upgrade
+from robottelo import ssh
+from unittest2.case import TestCase
+
+
+MEDIUM_TUNING_DATA = "apache::mod::passenger::passenger_max_pool_size: 30" \
```
mirekdlugosz: Please use parentheses instead of backslash:
```python
MEDIUM_TUNING_DATA = ("apache::mod::passenger::passenger_max_pool_size: 30"
    "\napache::mod::passenger::passenger_max_request_queue_"
    ...
)
```

In this case, when you create data to save in file, instead of polluting strings with newline characters, it might be better idea to have a list of strings that will be joined before writing:
```python
MEDIUM_TUNING_DATA = ["apache::mod::passenger::passenger_max_pool_size: 30",
    "apache::mod::passenger::passenger_max_request_queue_"
    ...
]
with open('custom-hiera.yaml', 'w') as fd:
    fd.write("\n".join(DEFAULT_CUSTOM_HIERA_DATA))
```
`"".join()` inserts string between list elements, so you probably want to first merge both lists together.

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,153 @@
...
+            1. Run satellite-installer --tuning medium --disable-system-checks.
+
+        :expectedresults: Medium tuning parameter should be applied.
+
+         """
+        tuning_state_after_upgrade = ssh.command('less /etc/foreman-installer/'
+                                                 'scenarios.d/satellite.yaml|'
+                                                 'grep "tuning: medium";if [ $? -eq 0 ]; '
+                                                 'then echo "PASSED"; else echo "FAILED";'
+                                                 ' fi').stdout
```
mirekdlugosz: I find this command strange. Piping less to grep, and printing out string depending on exit code? Just look for string in file and assert on exit code.

```
cmd = 'grep "tuning: medium" /etc/foreman-installer/scenarios.d/satellite.yaml'
tuning_state_after_upgrade = ssh.command(cmd).return_code
assert tuning_state_after_upgrade == 0  # zero indicates success, non-zero indicates error
```

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,153 @@
...
+    """The test class contains pre-upgrade and post-upgrade scenarios to test
+    Performance Tuning utility
+
+    Test Steps:
+
+    1. Before Satellite upgrade, we will apply the medium tune size.
+    2. Once the size get apply we will wait for post upgrade test case to check
+    whether the size is same or not.
+    3. In post upgrade we will check whether satellite installer would able to change
+    the size or not
```
mirekdlugosz: Typo? "would" -> "was"

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,153 @@
...
+        :expectedresults: Medium tuning parameter should be applied.
+
+         """
+        with open('custom-hiera.yaml', 'w') as fd:
+            fd.write("{}\n".format(DEFAULT_CUSTOM_HIERA_DATA))
+            fd.write("{}".format(MEDIUM_TUNING_DATA))
+        ssh.upload_file('custom-hiera.yaml', '/etc/foreman-installer')
+        command_output = [data.strip(" ") for data in
+                          ssh.command('satellite-installer --disable-system-checks')
+                              .stdout]
```
mirekdlugosz: I think it would be more readable if you had `ssh` call outside of comprehension:
```python
command_output = ssh.command('satellite-installer --disable-system-checks').stdout
command_output = [data.strip(" ") for data in command_output]
```

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+
+    @pre_upgrade
+    def test_pre_performance_tuning_apply(self):
+        """In preupgrade scenario we will apply the medium tuning size.
+
+        :id: preupgrade-83404326-20b7-11ea-a370-48f17f1fc2e1
+
+        :steps:
+            1. Run satellite-installer --tuning medium --disable-system-checks.
```
mirekdlugosz: This is not the command that you run in this step. You write custom properties and run installer without any additional parameters.

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,153 @@
...
+                                                 ' fi').stdout
+        assert "PASSED" in tuning_state_after_upgrade
+        with open('custom-hiera.yaml', 'w') as fd:
+            fd.write("{}".format(DEFAULT_CUSTOM_HIERA_DATA))
+
+        ssh.upload_file('custom-hiera.yaml', '/etc/foreman-installer')
+        command_output = [data.strip(" ") for data in
+                          ssh.command('satellite-installer --tuning default '
+                                      '--disable-system-checks').stdout]
+        assert "Success!" in command_output
```
mirekdlugosz: I assume the goal here is to revert Satellite to default state after test succeeds.

I think it would be better idea to download file as-is in pre-upgrade and upload it in post-upgrade. Right now you hardcode current defaults in test. If defaults change, you will need to update this test (or, what is more likely, we will not notice, not update and post-upgrade will restore slightly different state than it was before pre-upgrade).

Also, you possibly want this in `try ... finally` block, to ensure it runs even if test itself failed for whatever reason.

tests/upgrades/test_performance_tuning.py
devendra104: Fixed.

tests/upgrades/test_performance_tuning.py
devendra104: Yes, you are correct, my bad I pushed the testing code :)

tests/upgrades/test_performance_tuning.py
devendra104: fixed

tests/upgrades/test_performance_tuning.py
devendra104: fixed.

tests/upgrades/test_performance_tuning.py
devendra104: You are right, fixed.

tests/upgrades/test_performance_tuning.py
devendra104: Fixed.

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,195 @@
...
+
+        :steps:
+            1. Run satellite-installer --disable-system-checks.
+
+        :expectedresults: Medium tuning parameter should be applied.
+
+         """
+        cmd = 'grep "mongodb::server::storage_engine: \'wiredTiger\'" ' \
+              '/etc/foreman-installer/custom-hiera.yaml'
+        mongodb_type = ssh.command('{}'.format(cmd)).return_code
```
mirekdlugosz: `cmd` is a string already, there is no reason to use it inside `format` to cast it to string.

tests/upgrades/test_performance_tuning.py
mirekdlugosz: Constants are lists now, but there are still strings with backslashes elsewhere. Please do not use backslash before newline, use parentheses instead.

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,195 @@
...
+        :steps:
+            1. Collect the current tuning state from satellite.yaml file.
+            2. Compare it with the expected value.
+            3. If Expected value get match then this scenario would be passed.
+            1. Run satellite-installer --tuning medium --disable-system-checks.
+
+        :expectedresults: Medium tuning parameter should be applied.
+
+         """
+        cmd = 'grep "tuning: medium /etc/foreman-installer/scenarios.d/satellite.yaml'
```
mirekdlugosz: Seems you are missing ending double quote here:
```suggestion
        cmd = 'grep "tuning: medium" /etc/foreman-installer/scenarios.d/satellite.yaml'
```

devendra104: Fixed

tests/upgrades/test_performance_tuning.py
devendra104: Fixed.

devendra104:
> As far as I know, we use single Satellite instance to run all pre-upgrade checks, then we run upgrade, and finally we run all post-upgrade checks. Is manipulating these values safe? Isn't there a risk of interfering with other tests?

The risk happens(in terms of performance) only when we do not rollback the original state, but here 
we are rollback the actual tuning size(default) that was before the upgrade and also validate the content of custom-hiera.yaml file after post-upgrade.

tests/upgrades/test_performance_tuning.py
devendra104: > I assume the goal here is to revert Satellite to default state after test succeeds.
> 
> I think it would be better idea to download file as-is in pre-upgrade and upload it in post-upgrade. Right now you hardcode current defaults in test. If defaults change, you will need to update this test (or, what is more likely, we will not notice, not update and post-upgrade will restore slightly different state than it was before pre-upgrade).
> 
> Also, you possibly want this in `try ... finally` block, to ensure it runs even if test itself failed for whatever reason.

Thanks, Mirek, The test cases would be moved from one version to another, In this case, we will need to track the changes of the custom-hiera.yaml file, that's the reason I have written two test cases to verify the custom-hiera.yaml file in pre-upgrade and post-upgrade. If something gets change after an upgrade then we should aware of it.

Yes, you are right `try ... finally` block is required to remediate if any failure happens with default tune parameters after post-upgrade.

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,219 @@
...
+                                         connection_timeout=1000).stdout
+            command_output = [data.strip(" ") for data in command_output]
+            assert "Success!" in command_output
+        finally:
+            ssh.upload_file('custom-hiera.yaml', '/etc/foreman-installer')
+            command_output = ssh.command('satellite-installer '
+                                         '--disable-system-checks',
+                                         connection_timeout=1000).stdout
+            command_output = [data.strip(" ") for data in command_output]
+            assert "Success!" in command_output
```
mirekdlugosz: `finally` runs every time, doesn't matter if there was exception in `try` or not.
If everything works as intended, we upload `custom-hiera.yaml` twice and run `satellite-installer` two times (but with different parameters).
If uploading file throws exception, we upload file again.
I'm not sure this is how we should approach situation like that.

devendra104:
> > The risk happens(in terms of performance) only when we do not rollback the original state, but here we are rollback the actual tuning size(default) that was before the upgrade and also validate the content of custom-hiera.yaml file after post-upgrade.
> 
> My point is, you don't control order of test execution. There might be any number of tests running between your pre-upgrade and post-upgrade.
> Unless I'm missing something and you can confidently say that your pre-upgrade will be the last test running before Satellite upgrade, and your post-upgrade will be the first running after Satellite upgrade. Then it's ok, there will be no impact at all on other tests.

> Or you can ensure that changes that you do here are not significant enough to affect other tests status (passed/failed). I don't mind too much about performance, we can re-think it once it turns out that performance impact is big.

As you know, we can not control the order of execution, but I can ensure you that there are no such data(most of the data related to improving the performance) that impact the test case's execution, But If any failure happens while setting the tuning parameters then it will impact the other test case's execution.

tests/upgrades/test_performance_tuning.py
devendra104: > `finally` runs every time, doesn't matter if there was exception in `try` or not.
> If everything works as intended, we upload `custom-hiera.yaml` twice and run `satellite-installer` two times (but with different parameters).
> If uploading file throws exception, we upload file again.
> I'm not sure this is how we should approach situation like that.

Yes, It will be executed whether the previous command gets successful or not,
The good thing is, It will not only apply the default data in custom-hiera.yaml file but also fixes any system-related failure.

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,221 @@
...
+            fd.write("\n".join(DEFAULT_CUSTOM_HIERA_DATA))
+            if mongodb_type:
+                fd.write("\n".join(MMPV1_MONGODB))
+            else:
+                fd.write("\n".join(WIREDTIGER_MONGODB))
+            if tune_size == "medium":
+                fd.write("\n".join(MEDIUM_TUNING_DATA))
+
+    @pre_upgrade
+    def test_pre_custom_hiera_file_validation(self):
```
jyejare: This should be in another class as another scenario. We shouldnt create two scenarios under one class.

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,221 @@
...
+        """ In preupgrade scenario we will validate the default custom hiera file
```
jyejare: Missing id, steps and extected_results .

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,221 @@
...
+         """
+        cmd = 'grep "mongodb::server::storage_engine: \'wiredTiger\'" ' \
+              '/etc/foreman-installer/custom-hiera.yaml'
+        mongodb_type = ssh.command(cmd).return_code
+        self._custom_hiera_file_creation(mongodb_type, "medium")
+
+        ssh.upload_file('custom-hiera.yaml', '/etc/foreman-installer')
+        command_output = ssh.command('satellite-installer --disable-system-checks',
+                                     connection_timeout=1000).stdout
+        command_output = [data.strip(" ") for data in command_output]
```
jyejare: You need not to convert the outpu string into list, as you can directly use 'in' operator to check in output string. So remove this line and say:
assert "Success!" in command_output   (here command_output is string)

But the question is what if there are more than one Success / Failure ?

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,221 @@
...
+        try:
+            self._custom_hiera_file_creation(mongodb_type)
+            cmd = 'grep "tuning: medium" /etc/foreman-installer/scenarios.d/satellite.yaml'
+            tuning_state_after_upgrade = ssh.command(cmd).return_code
+            assert tuning_state_after_upgrade == 0
+            ssh.upload_file('custom-hiera.yaml', '/etc/foreman-installer')
+            command_output = ssh.command('satellite-installer --tuning default '
+                                         '--disable-system-checks',
+                                         connection_timeout=1000).stdout
+            command_output = [data.strip(" ") for data in command_output]
```
jyejare: same here no need to list comprehenssion.

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,221 @@
...
+:Upstream: No
+"""
+
+from robottelo.decorators import destructive
+from upgrade_tests import post_upgrade, pre_upgrade
+from robottelo import ssh
+from unittest2.case import TestCase
+
+
+DEFAULT_CUSTOM_HIERA_DATA = [
```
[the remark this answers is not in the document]
devendra104: Yes, we can achieve it by using multi-line but it will make the code bulky and dirty.

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,221 @@
...
+    "# apache::mod::prefork::maxclients: 256",
+    "# apache::mod::prefork::maxrequestsperchild: 4000",
+    "# Here are some examples of how you tune the PostgreSQL options if needed:",
+    "#",
+    "# postgresql::server::config_entries:",
+    "#   max_connections: 600",
+    "#   shared_buffers: 1024MB",
+]
+
+MEDIUM_TUNING_DATA = [
```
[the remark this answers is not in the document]
devendra104: Yes, we can achieve it by using multi-line but it will make the code bulky and dirty.

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,221 @@
...
+]
+
+WIREDTIGER_MONGODB = [
+    "# Do not remove",
+    "mongodb::server::storage_engine: 'wiredTiger'"
+]
+
+
+@destructive
+class ScenarioPerfornaceTuning(TestCase):
```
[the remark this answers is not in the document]
devendra104: Fixed

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,221 @@
...
+    """The test class contains pre-upgrade and post-upgrade scenarios to test
+    Performance Tuning utility
+
+    Test Steps:
+
+    1. Before Satellite upgrade, we will apply the medium tune size.
```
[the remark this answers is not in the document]
devendra104: Yes.

tests/upgrades/test_performance_tuning.py
devendra104: fixed.

tests/upgrades/test_performance_tuning.py
devendra104: fixed.

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,256 @@
...
+        mongodb_type = ssh.command(cmd).return_code
+        self._custom_hiera_file_creation(mongodb_type, "medium")
+
+        ssh.upload_file('custom-hiera.yaml', '/etc/foreman-installer')
+        command_output = ssh.command('satellite-installer --disable-system-checks',
+                                     connection_timeout=1000).stdout
+        command_output = [data.strip(" ") for data in command_output]
+        assert "Success!" in command_output
+
+    @post_upgrade
```
ntkathole: @post_upgrade(depend_on=test_pre_performance_tuning_apply)

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,256 @@
...
+              '/etc/foreman-installer/custom-hiera.yaml'
+        mongodb_type = ssh.command(cmd).return_code
+        if mongodb_type:
+            expected_custom_hiera = set(DEFAULT_CUSTOM_HIERA_DATA + MMPV1_MONGODB)
+        else:
+            expected_custom_hiera = set(DEFAULT_CUSTOM_HIERA_DATA + WIREDTIGER_MONGODB)
+
+        assert expected_custom_hiera == actual_custom_hiera
+
+    @post_upgrade
```
ntkathole: @post_upgrade(depend_on=test_pre_custom_hiera_file_validation)

tests/upgrades/test_performance_tuning.py
devendra104: Fixed

tests/upgrades/test_performance_tuning.py
devendra104: Fixed

tests/upgrades/test_performance_tuning.py
devendra104: Fixed.
And added an exception in pre-upgrade to keep the systems original state.

tests/upgrades/test_performance_tuning.py
devendra104: Fixed.

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,221 @@
+"""Test for Performance Tuning related Upgrade Scenario's
+
+:Requirement: Upgraded Satellite
+
+:CaseAutomation: Automated
+
+:CaseLevel: Acceptance
+
+:CaseComponent: CLI
```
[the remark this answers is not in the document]
jyejare: Still Pending ..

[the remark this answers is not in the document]
devendra104: Fixed

devendra104:
> > > The risk happens(in terms of performance) only when we do not rollback the original state, but here we are rollback the actual tuning size(default) that was before the upgrade and also validate the content of custom-hiera.yaml file after post-upgrade.
> > 
> > 
> > My point is, you don't control order of test execution. There might be any number of tests running between your pre-upgrade and post-upgrade.
> > Unless I'm missing something and you can confidently say that your pre-upgrade will be the last test running before Satellite upgrade, and your post-upgrade will be the first running after Satellite upgrade. Then it's ok, there will be no impact at all on other tests.
> 
> > Or you can ensure that changes that you do here are not significant enough to affect other tests status (passed/failed). I don't mind too much about performance, we can re-think it once it turns out that performance impact is big.
> 
> As you know, we can not control the order of execution, but I can ensure you that there are no such data(most of the data related to improving the performance) that impact the test case's execution,

>But If any failure happens while setting the tuning parameters then it will impact the other test case's execution.
We have handled this problem too If some issue happens in pre-upgrade then the system will be rollbacked to their actual states.

And If the pre-upgrade test cases get failed then the post-upgrade test wouldn't be executed( @post_upgrade(depend_on=test_pre_performance_tuning_apply)).

And these scenarios will be executed when destructive scenarios checks enable in the pipeline.

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,264 @@
...
+                                         connection_timeout=1000).stdout
+            assert '  Success!' in command_output
+        except Exception:
+            test_case_status = "FAILED"
+            self._custom_hiera_file_creation(mongodb_type, "default")
+            ssh.upload_file('custom-hiera.yaml', '/etc/foreman-installer')
+            command_output = ssh.command('satellite-installer -s --disable-system-checks',
+                                         connection_timeout=1000).stdout
+            assert '  Success!' in command_output
+            assert 'PASSED' == test_case_status
```
mirekdlugosz: This last assertion has no way of succeeding, as `test_case_status` has unconditionally assigned `"FAILED"`.

If your intention is to run some clean up code on failure, but still fail test, just re-raise caught exception:
```python
try:
    ssh.upload_file('custom-hiera.yaml', '/etc/foreman-installer')
    ...
except Exception:
    self._custom_hiera_file_creation(mongodb_type, "default")
    ... #  other cleanup code
    raise #  raise without arguments in exception handler re-raises this exception
```

tests/upgrades/test_performance_tuning.py
```diff
@@ -0,0 +1,264 @@
...
+    1. Before Satellite upgrade, we will apply the medium tune size.
+    2. Once the size get apply we will wait for post upgrade test case to check
+    whether the size is same or not.
+    3. In post upgrade we will check whether satellite installer was able to change
+    the size or not
+
+    :expectedresults: The performance parameter should not be changed after upgrade.
+    """
+    @staticmethod
+    def _custom_hiera_file_creation(mongodb_type, tune_size="default"):
```
mirekdlugosz: Try to use verbs instead of nouns in function names, as functions do something: `_create_custom_hiera_file`.

tests/upgrades/test_performance_tuning.py
devendra104: The purpose is only to restore the state if some kind of issue happens in the try part.
And the post-upgrade test depends on the pre-upgrade test if it gets complete successfully then and only the post-upgrade test gets to execute.

PR #7554, opened 2019-12-19
add tier2 test for cockpit webconsole integration

rplevka:
adds coverage for the cockpit integration feature.

Please, do not merge before https://github.com/SatelliteQE/automation-tools/pull/819 
and https://github.com/SatelliteQE/airgun/pull/436 get merged, otherwise the test will fail due environment not being configured correctly.

```
================================================================== test session starts ==================================================================
platform linux -- Python 3.7.5, pytest-4.6.3, py-1.8.0, pluggy-0.12.0
sensitiveurl: .*
rootdir: /home/rplevka/work/rplevka/robottelo/tests/foreman, inifile: pytest.ini
plugins: variables-1.7.1, services-1.3.1, selenium-1.17.0, metadata-1.8.0, xdist-1.29.0, repeat-0.8.0, mock-1.10.4, html-1.22.0, forked-1.0.2, base-url-1.4.1
collecting 1 item                                                                                                                                       2019-12-19 14:45:49 - conftest - DEBUG - Collected 1 test cases
2019-12-19 15:45:49 - robottelo.ssh - DEBUG - Instantiated Paramiko client 0x7f92e29c0290
2019-12-19 15:45:49 - robottelo.ssh - INFO - Connected to [sat-6-7-qa-rhel7.localhost.example.com]
2019-12-19 15:45:49 - robottelo.ssh - INFO - >>> rpm -q satellite
2019-12-19 15:45:50 - robottelo.ssh - INFO - <<< stdout
satellite-6.7.0-4.beta.el7sat.noarch

2019-12-19 15:45:50 - robottelo.ssh - DEBUG - Destroyed Paramiko client 0x7f92e29c0290
2019-12-19 15:45:50 - robottelo.host_info - DEBUG - Host Satellite version: 6.7
2019-12-19 15:45:50 - robottelo.helpers - DEBUG - Generated file bz_cache.json with BZ collect data
collected 1 item                                                                                                                                        

test_host.py .                                                                                                                                    [100%]
========================================================= 1 passed, 6 warnings in 58.90 seconds =========================================================
```

### DIFF
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -1970,3 +1970,31 @@ def test_positive_gce_cloudinit_provision_end_to_end(
             gce_client.disconnect()
             skip_yum_update_during_provisioning(
                 template='Kickstart default user data', reverse=True)
+
+
+@tier2
+def test_positive_cockpit(session):
+    """Test whether webconsole button and cockpit integration works
+
+    :id: 5a9be063-cdc4-43ce-91b9-7608fbebf8bb
+
+    :expectedresults: Cockpit page is loaded and displays sat host info
+
+    :CaseLevel: System
+
+    """
+    with session:
+        session.organization.select(org_name='Default Organization')
+        session.location.select(loc_name='Any Location')
+        session.host.open_webconsole(entity_name=settings.server.hostname)
+        wd = session.browser.selenium
+        assert len(wd.find_elements_by_xpath('//iframe')) > 0, (
+            'Unable to find iframe on the page - are we on the Cockpit page?')
+        # the remote host content is loaded in an iframe, let's switch to it
+        wd.switch_to.frame(0)
+        hostname_element = wd.find_elements_by_id('system_information_hostname_button')
+        assert len(hostname_element) > 0, (
+            'element "system_information_hostname_button" not found in the iframe')
+        assert hostname_element[0].text == settings.server.hostname, (
+            'cockpit page shows hostname {0} instead of {1}'.format(
+                   hostname_element[0].text, settings.server.hostname))


tests/foreman/ui/test_host.py
```diff
@@ -1970,3 +1970,32 @@ def test_positive_gce_cloudinit_provision_end_to_end(
...
+        session.location.select(loc_name='Any Location')
+        session.host.open_webconsole(entity_name=settings.server.hostname)
+        wd = session.browser.selenium
+        assert(len(wd.find_elements_by_xpath('//iframe')) > 0,
+               'Unable to find iframe on the page - are we on the Cockpit page?'
+        )
+        # the remote host content is loaded in an iframe, let's switch to it
+        wd.switch_to.frame(0)
+        hostname_element = wd.find_elements_by_id('system_information_hostname_button')
+        assert(len(hostname_element) > 0,
```
mirekdlugosz: assert is a statement, not a function. So instead of passing two arguments to `assert`, you create-two element tuple and pass it to `assert` to evaluate its truthiness - and non-empty tuple will always be truthy.

If you want assertion message in new line, you need:
```python
assert len(hostname_element) > 0, (
        'element "system_information_hostname_button" not found in the iframe'
)
```

rplevka: oops, you're right...this would always return true

rplevka:
>     1. There are multiple flake8 errors that cause travis to fail. Please take a look at them.
> 
>     2. As a general rule, we strongly discourage interacting with selenium directly in robottelo test code, and prefer to use abstractions in Airgun instead. If there are no abstractions built in already, we should add them first.
>        Speaking of which, I can't find any reference to `open_webconsole` method in Airgun. So this PR seems to already assume some unmerged change.

1. will fix those
2. _As a general rule, we strongly discourage interacting with selenium directly in robottelo test code_  - citation needed. 
Anyway, this is navigating outside of satellite UI and I'm really not willing to make abstraction for 3rd party project (Cockpit) UI into airgun.
3. https://github.com/SatelliteQE/airgun/pull/436

rplevka:
```
$ py.test test_host.py::test_positive_cockpit
===================================================== test session starts =====================================================
platform linux -- Python 3.7.5, pytest-4.6.3, py-1.8.0, pluggy-0.12.0
sensitiveurl: .*
rootdir: /home/rplevka/work/rplevka/robottelo/tests/foreman, inifile: pytest.ini
plugins: variables-1.7.1, services-1.3.1, selenium-1.17.0, metadata-1.8.0, xdist-1.29.0, repeat-0.8.0, mock-1.10.4, html-1.22.0, forked-1.0.2, base-url-1.4.1
collecting 1 item                                                                                                             2019-12-19 16:23:38 - conftest - DEBUG - Collected 1 test cases
2019-12-19 17:23:38 - robottelo.ssh - DEBUG - Instantiated Paramiko client 0x7f323bfd5e10
2019-12-19 17:23:38 - robottelo.ssh - INFO - Connected to [sat-6-7-qa-rhel7.localhost.example.com]
2019-12-19 17:23:38 - robottelo.ssh - INFO - >>> rpm -q satellite
2019-12-19 17:23:39 - robottelo.ssh - INFO - <<< stdout
satellite-6.7.0-4.beta.el7sat.noarch

2019-12-19 17:23:39 - robottelo.ssh - DEBUG - Destroyed Paramiko client 0x7f323bfd5e10
2019-12-19 17:23:39 - robottelo.host_info - DEBUG - Host Satellite version: 6.7
2019-12-19 17:23:39 - robottelo.helpers - DEBUG - Generated file bz_cache.json with BZ collect data
collected 1 item                                                                                                              

test_host.py .                                                                                                          [100%]

============================================ 1 passed, 3 warnings in 71.91 seconds ============================================
```

tests/foreman/ui/test_host.py
```diff
@@ -1970,3 +1970,31 @@ def test_positive_gce_cloudinit_provision_end_to_end(
...
+    :CaseLevel: System
+
+    """
+    with session:
+        session.organization.select(org_name='Default Organization')
+        session.location.select(loc_name='Any Location')
+        session.host.open_webconsole(entity_name=settings.server.hostname)
+        wd = session.browser.selenium
+        assert len(wd.find_elements_by_xpath('//iframe')) > 0, 'Unable to find iframe on the ' \
+                                                               'page - are we on the Cockpit page?'
```
mirekdlugosz: Please use parentheses instead of backslash:
```python
assert len(wd.find_elements_by_xpath('//iframe')) > 0, (
    'Unable to find iframe on the page - are we on the Cockpit page?'
)
```

tests/foreman/ui/test_host.py
```diff
@@ -1970,3 +1970,31 @@ def test_positive_gce_cloudinit_provision_end_to_end(
...
+        # the remote host content is loaded in an iframe, let's switch to it
+        wd.switch_to.frame(0)
+        hostname_element = wd.find_elements_by_id('system_information_hostname_button')
+        assert len(hostname_element) > 0, 'element "system_information_hostname_button" not ' \
```
omkarkhatavkar: Why don't we encapsulate these above codes in airgun and used as a function return whatever required instead call selenium web driver code?  Correct me if I'm wrong.

tests/foreman/ui/test_host.py
rplevka: why?

rplevka:
> Of course implementing entire Cockipt in Airgun would not be reasonable. 
I wasn't talking about entire Cockpit implementation either, of course.

>...
> How does that sound?

Honestly, pretty awful :) (no offence).
What I always hated about nailgun has already crawled into airgun as well and this is mixing the functional logic with the testing one.
I want to have my testing logic in the test, I don't want to pendulate between various projects just to keep my test in shape.

> Then, your test would be literally couple lines of code

Why do you think the goal is to have as little lines of code in the test as possible? What would be the ideal state for you?
```
def test_foo():
   test_stuff_in_airgun
```
?

rplevka:
Anyway, can we keep it this way until  there's a new test accessing cockpit? As currently I doubt there will be any more such tests and I really don't want t overengineer it ( we already have a lot oh single-purpose helper methods)

tests/foreman/ui/test_host.py
omkarkhatavkar: whatever @mirekdlugosz  suggested looks nicer!

mirekdlugosz: @rplevka 
1. PEP8 discourages backslashes at the end of line
2. With parentheses it's more readable, as you can have message in one line
3. `black`, which we might start using at one point, will reformat it from backslashes to parentheses anyway

rplevka: fair enough

mirekdlugosz:
> I want to have my testing logic in the test, I don't want to pendulate between various projects just to keep my test in shape.

Airgun is mostly build on two principles:
1. To hide away complexities of interacting with website from test
2. To provide set of reasonable goals that user would like to achieve, mostly by trying to follow CRUD

If `host.has_working_webconsole()` crosses the boundary by hiding away assertions and oracles in Airgun, then you can stop couple of steps earlier. Something like `host.read_webconsole()`, which would navigate to cockpit UI, read all points of interest and return it to test. Then test will be responsible for asserting whether what it get matches expectations.

And by the way, `session.host.open_webconsole()` is exactly single purpose helper that you don't like. The only difference is that it's simpler than what I proposed before, and that it doesn't fit rest of the framework too well. Opening the page is hardly the goal - it's only a necessary step on your way of achieving the real goal.

> What would be the ideal state for you?

To have code consistent with rest of our tools - one that follows already established practices and principles. 
In this case it also makes test shorter, but that's only a by-product caused by test itself being simple.

rplevka:
This hurts. 
I'm going to implement it in your way just because i'm too tired already. whatever, the world doesn't end here.

Hopefully we'll do better with the next major framework rewrite :'D

tstrych:
I run this test locally.  I already had enabled remote-execution-cockpit.
So I tested just this PR and [airgun PR](https://github.com/SatelliteQE/airgun/pull/436)

What i found test works correctly but there is small problem, if your test is fast enough you are unable to catch hostname, because it's not there yet, or it's not loaded yet.
```
0. hostname text:  
1. hostname text: hpe-dl380pgen8-01.hpe2.lab.eng.bos.redhat.com 
```
where 0. - print right now 
and 1. - print after 1 second of sleeping.

Wait for can be solution for this problem.

mirekdlugosz:
Closing, since #7679 is now merged.

PR #7568, opened 2019-12-30
Upgrade Scenario for GCE

jyejare:
- An upgrade Scenario which tests:

- [x] In 6.6, The Compute Resource(UI support) with an image can be created

- [x] In 6.7, On the same 6.6 Compute Resource, attempt to provision Host(API, CLI support) and manipulate existing GCE CR.

### DIFF
--- a/robottelo/helpers.py
+++ b/robottelo/helpers.py
@@ -24,8 +24,9 @@ from robottelo.config import settings
 from robottelo.constants import (
     PULP_PUBLISHED_YUM_REPOS_PATH,
     RHEL_6_MAJOR_VERSION,
-    RHEL_7_MAJOR_VERSION,
+    RHEL_7_MAJOR_VERSION
 )
+from robottelo.errors import GCECertNotFoundError
 from robottelo.issue_handlers import bugzilla
 
 LOGGER = logging.getLogger(__name__)
@@ -1077,3 +1078,10 @@ def _handle_version(key, value):  # pragma: no cover
             for item in value
         ]
     return value
+
+
+def download_gce_cert(gce_cert_url, gce_cert_path):
+    ssh.command('curl {0} -o {1}'.format(gce_cert_url, gce_cert_path))
+    if ssh.command('[ -f {} ]'.format(gce_cert_path)).return_code != 0:
+        raise GCECertNotFoundError(
+            "The GCE certificate {} is not found in satellite.".format(gce_cert_path))
--- a/tests/foreman/ui/test_computeresource_gce.py
+++ b/tests/foreman/ui/test_computeresource_gce.py
@@ -17,8 +17,6 @@
 from fauxfactory import gen_string
 from nailgun import entities
 from pytest import skip
-from robottelo import ssh
-from robottelo.config import settings
 from robottelo.constants import (
     COMPUTE_PROFILE_SMALL,
     FOREMAN_PROVIDERS,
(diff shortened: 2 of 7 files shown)


tests/upgrades/test_host.py
```diff
@@ -0,0 +1,152 @@
...
+    :expectedresults:
+
+        1. The host should be provisioned on GCE CR created in previous version
+        2. The GCE CR attributes should be manipulated
+    """
+    @classmethod
+    def setUpClass(cls):
+        cls.hostname = None
+        cls.arch = entities.Architecture().search(query={'search': 'name=x86_64'})[0]
+        cls.img_uuid = '7726764279310511390'
```
ntkathole: Is this random value or defined somewhere?

tests/upgrades/test_host.py
```diff
@@ -0,0 +1,152 @@
...
+            silent_failure=True,
+            handle_exception=True
+        )
+        assert gce_hst.name == '{}.{}'.format(self.hostname, self.domain_name).lower()
+        gce_hst = entities.Host(id=gce_hst.id).read()
+        assert gce_hst.build_status_label == 'Installed'
+        # CR Manipulation Tests
+        newgce_name = gen_string('alpha')
+        gce_cr.name = newgce_name
+        gce_cr.zone = 'europe-west4-a'
```
ntkathole: Use from robottelo.constants import VALID_GCE_ZONES

robottelo/helpers.py
```diff
@@ -1077,3 +1078,10 @@ def _handle_version(key, value):  # pragma: no cover
             for item in value
         ]
     return value
+
+
+def download_gce_cert(gce_cert_url, gce_cert_path):
+    ssh.command('curl {0} -o {1}'.format(gce_cert_url, gce_cert_path))
+    if not ssh.command('[ -f {} ]'.format(gce_cert_path)).return_code == 0:
```
ntkathole: can be != 0

tests/upgrades/test_host.py
jyejare: With this update, I am checking if the 'update' functionality works fine not 'GCE zones'. For GCE Zones, the robottelo test will take care of it. Hence hardcoded this value.

tests/upgrades/test_host.py
jyejare: This is a UUID of the image on Google Cloud. This is not going to change.

ntkathole: I would suggest to keep it in constant with docstring added about how and where to find this value for future reference.

tests/upgrades/test_host.py
ntkathole: If VALID_GCE_ZONES is used, then we don't need to maintain the test case if zone value is changed in future.

jyejare:
The Results:


Pre:
```
=============================================================================== 1 passed, 1 deselected, 2 warnings in 93.84 seconds ================================================================================
```
Post: 
```
=============================================================================== 1 passed, 1 deselected, 2 warnings in 578.72 seconds ===============================================================================
```

jyejare:
@ntkathole , Post -Comment Results:

Pre:
```
========================== 1 passed, 1 deselected, 2 warnings in 97.08 seconds ==========================
```

Post:
```
========================= 1 passed, 1 deselected, 2 warnings in 253.80 seconds ==========================
```

tests/upgrades/test_host.py
jyejare: Done

tests/foreman/ui/test_computeresource_gce.py
```diff
@@ -17,58 +17,24 @@
...
-from robottelo import ssh
-from robottelo.config import settings
 from robottelo.constants import (
     COMPUTE_PROFILE_SMALL,
     FOREMAN_PROVIDERS,
     GCE_EXTERNAL_IP_DEFAULT,
     GCE_MACHINE_TYPE_DEFAULT,
-    GCE_NETWORK_DEFAULT
+    GCE_NETWORK_DEFAULT,
+    GCE_SETTINGS
```
omkarkhatavkar: I don't see this added in https://github.com/SatelliteQE/robottelo/blob/8c62891b3e880f3461ae9fe8c3a4350956752747/robottelo/constants.py

tests/foreman/ui/test_computeresource_gce.py
```diff
@@ -127,9 +97,9 @@ def test_positive_default_end_to_end_with_custom_profile(
         cr_values = session.computeresource.read(new_cr_name)
         assert cr_values['name'] == new_cr_name
         assert (set(cr_values['organizations']['resources']['assigned'])
-                == {module_org.name, new_org.name})
+                == {org.name, new_org.name})
```
omkarkhatavkar: no need for extra brackets after assert

tests/foreman/ui/test_computeresource_gce.py
```diff
@@ -127,9 +97,9 @@ def test_positive_default_end_to_end_with_custom_profile(
...
         assert (set(cr_values['locations']['resources']['assigned'])
-                == {module_loc.name, new_loc.name})
+                == {loc.name, new_loc.name})
```
omkarkhatavkar: no need for extra brackets after assert

tests/upgrades/test_host.py
```diff
@@ -0,0 +1,154 @@
...
+from wait_for import wait_for
+
+
+GCE_SETTINGS = dict(
+    project_id=settings.gce.project_id,
+    client_email=settings.gce.client_email,
+    cert_path=settings.gce.cert_path,
+    zone=settings.gce.zone,
+    cert_url=settings.gce.cert_url,
+)
```
omkarkhatavkar: do you think we could add this in robottelo constants, in case other test classes need them?

tests/upgrades/test_host.py
```diff
@@ -0,0 +1,154 @@
...
+                entities.Host(id=hst[0].id).delete()
+
+    @pre_upgrade
+    def test_pre_create_gce_cr_and_host(self):
+        """
+        """
+        cr_name = gen_string('alpha')
+        org = entities.Organization().create()
+        loc = entities.Location().create()
+        with Session('gce_upgrade_tests') as session:
```
omkarkhatavkar: Any specific reason for creating GCE resource using AIrgun? Is that not possible with API?

tests/foreman/ui/test_computeresource_gce.py
jyejare: Great Catch, I wonder flake8 didnt catch that. Fixed

tests/foreman/ui/test_computeresource_gce.py
jyejare: Flake8 was complaining for 'needs statement'. I reformated things.

tests/upgrades/test_host.py
jyejare: It doesn't work in constants as the settings are not initiated at that level, I tried that and after facing errors I had to move it back to test file.

tests/upgrades/test_host.py
jyejare: This is a pre part of upgrade scenario, that means its going to be run in 6.6. And 6.6 only supports UI not API.

omkarkhatavkar: okay make sense, I was not aware that

PR #7570, opened 2020-01-02
Fix 6.7 Automation Fixes

jyejare:
Fixes tests failing in 6.7 for my components.

### DIFF
--- a/tests/foreman/endtoend/test_api_endtoend.py
+++ b/tests/foreman/endtoend/test_api_endtoend.py
@@ -16,6 +16,7 @@
 :Upstream: No
 """
 import random
+import pytest
 
 from fauxfactory import gen_string
 from nailgun import client, entities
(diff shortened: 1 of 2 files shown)


tests/foreman/endtoend/test_api_endtoend.py
```diff
@@ -1064,6 +1127,7 @@ def test_positive_ping(self):
             u'Not all services seem to be up and running!'
         )
 
+    @pytest.mark.skip_if_open("BZ:1787147")
```
mirekdlugosz: Could you only skip parts of test, instead of entire thing?
If I understand correctly, only parts related to manifest and depending on manifest are blocked, everything else should work. By blocking entire thing we lose information until this bug is fixed, which might result in us noticing another regression later than we could.

jyejare: The test is the end-to-end test and currently, this issue is breaking that link of end-to-end. All above the failing assert is already being exercised in CRUD/integration tests and I believe it is not required to test them here.

So skipping the entire test is better here. What do u say ?

mirekdlugosz: These exact scenarios, when action is run as another user with admin permissions, are generally not covered by specific CRUD tests. Most of the time we work as default admin, sometimes we create new user with very specific permissions.

That was only the suggestion, I'm fine with current approach as well.

tests/foreman/endtoend/test_api_endtoend.py
```diff
@@ -708,14 +755,8 @@
...
         u'/foreman_tasks/api/recurring_logics/:id',
-        u'/foreman_tasks/api/recurring_logics/:id/cancel',
-    ),
-    u'registries': (
-        u'/docker/api/v2/registries',
-        u'/docker/api/v2/registries',
-        u'/docker/api/v2/registries/:id',
-        u'/docker/api/v2/registries/:id',
-        u'/docker/api/v2/registries/:id',
+        u'/foreman_tasks/api/recurring_logics/:id',
```
ntkathole: seems duplicated

tests/foreman/endtoend/test_api_endtoend.py
```diff
@@ -729,6 +770,19 @@
         u'/api/reports/:id',
         u'/api/reports/:id',
     ),
+    u'report_templates': (
+        u'/api/report_templates',
+        u'/api/report_templates/:id',
+        u'/api/report_templates',
+        u'/api/report_templates/import',
+        u'/api/report_templates/:id',
+        u'/api/report_templates/:id',
```
ntkathole: duplicates

tests/foreman/endtoend/test_api_endtoend.py
```diff
@@ -882,6 +940,12 @@
     u'template_kinds': (
         u'/api/template_kinds',
     ),
+    u'trends': (
+        u'/api/trends',
+        u'/api/trends/:id',
+        u'/api/trends',
```
ntkathole: duplicate

jyejare:
@ntkathole , Those are actually duplicates in API/APIDoc.

PR #7575, opened 2020-01-06
test_positive_bulk_delete_host fix

pondrejk:
blocked by https://github.com/SatelliteQE/airgun/pull/437
Less hosts and faster assert
```
 pytest tests/foreman/ui/test_host.py -k test_positive_bulk_delete_host
...

tests/foreman/ui/test_host.py .                                                                                            [100%]
===================================== 1 passed, 28 deselected, 3 warnings in 103.86 seconds ======================================
```

### DIFF
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -1543,7 +1543,7 @@ def test_positive_set_multi_line_and_with_spaces_parameter_value(session, module
 @tier4
 @upgrade
 def test_positive_bulk_delete_host(session, module_loc):
-    """Delete a multiple hosts from the list
+    """Delete multiple hosts from the list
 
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_host.py
```diff
@@ -1568,13 +1568,13 @@ def test_positive_bulk_delete_host(session, module_loc):
             operatingsystem=host_template.operatingsystem,
             ptable=host_template.ptable,
         ).create().name
-        for _ in range(18)
+        for _ in range(3)
```
omkarkhatavkar: Why we are changing from 18 to 3? any specific issue

pondrejk: No specific reason, just wanted to reduce the number, 3 seems just enough but not a strong opinion

Issue #7574, opened 2020-01-06
virtual_machine.install_katello_agent()E robottelo.vm.VirtualMachineError: Failed to install katello-agent


vijay8451:
Mutiple tests are failing due to below:

```
virtual_machine.install_katello_agent()E robottelo.vm.VirtualMachineError: Failed to install katello-agent
```

PR #7579, opened 2020-01-08
Added waits for to avoid Flakiness of tests

vijay8451:
From last couple of automation ran I have noticed that multiple tests(10-15) are failing due  error `VirtualMachineError: Failed to install katello-agent` ..When I checked on automation Satellite I did noticed that everything looks pretty good however due to some delay in repolist is taking bit time at client end(might be task is taking a bit of time to complete but not sure) also when you run same test manually all looks good and works fine.

hence to avoid this flakiness further I have raised a small fix[1] here  

test result:
```
tests/foreman/ui/test_contenthost.py::test_positive_host_re_registion_with_host_rename 2020-01-08 15:54:07

PASSED2020-01-08 16:08:01 - robottelo.vm - INFO - Destroying the VM

======================================================================== 1 passed in 866.23 seconds =========================================================================
```

### DIFF
--- a/robottelo/vm.py
+++ b/robottelo/vm.py
@@ -13,6 +13,7 @@ import json
 import logging
 import os
 import six
+from wait_for import wait_for
 
 from fauxfactory import gen_string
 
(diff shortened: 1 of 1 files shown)


vijay8451:
> Wouldn't it be better to wrap `yum install katello-agent` in `wait_for`/loop?
> 
> yum returns code 0 on successful installation, 1 on failure (including case where package can't be found). So you can easily hook into that.
> 
> Trying wanted thing right away will generally be faster, as it will install without any delay in best case when package is already available on the time of function call. In your approach, we first spend some time figuring out if operation will succeed, and only then trying to install.

Good catch ... we could that ..thanks

vijay8451:
It's there now.

vijay8451:
Fixed: #7574

PR #7589, opened 2020-01-17
Removed bug and updated assertion

vijay8451:
- Removed fix bug.
- Updated assert to avoid false fail.
- test result:
```
$ py.test tests/foreman/ui/test_repository.py -k test_positive_recommended_repos

collected 21 items / 20 deselected / 1 selected                                                                                                                             

tests/foreman/ui/test_repository.py .                                                                                                                                 [100%]

================================================================= 1 passed, 20 deselected in 177.06 seconds =================================================================

```

### DIFF
--- a/robottelo/constants.py
+++ b/robottelo/constants.py
@@ -213,7 +213,6 @@ REPOSET = {
     'rhscl7': ('Red Hat Software Collections RPMs for Red Hat Enterprise'
                ' Linux 7 Server'),
     'rhae2': 'Red Hat Ansible Engine 2.7 RPMs for Red Hat Enterprise Linux 7 Server',
-    'fdrh8': 'Fast Datapath for RHEL 8 x86_64 (RPMs)',
 }
 
 REPOS = {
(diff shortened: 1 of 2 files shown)


vijay8451:
> Looks innocent enough, but can you also explain why is it necessary to change assertion from "Fast Datapath for RHEL 8" to "Red Hat Ansible Engine"? 

Yeah, because as checked with current manifest the `Red Hat Ansible Engine` repo shows on first page of 'Red Hat Repositories' instead of `Fast Datapath for RHEL 8` hence updated it

> Should we remove "fdrh8" from constants as well?

yeah, it's removed now.

mirekdlugosz:
> Yeah, because as checked with current manifest the `Red Hat Ansible Engine` repo shows on first page of 'Red Hat Repositories' instead of `Fast Datapath for RHEL 8` hence updated it

So the real problem is that `session.redhatrepository.read(recommended_repo='off')` is unable to read more than one page?

If that's the case, please file issue in Airgun.

vijay8451:
> So the real problem is that `session.redhatrepository.read(recommended_repo='off')` is unable to read more than one page?
> If that's the case, please file issue in Airgun.

We could add ... but I do not see much value add except to see the repos in a list ... 
I have opened the issue for same: https://github.com/SatelliteQE/airgun/issues/440

PR #7604, opened 2020-01-24
GCE API Tests - Part 1

jyejare:
$subject

### DIFF
--- a/robottelo/helpers.py
+++ b/robottelo/helpers.py
@@ -1080,8 +1080,10 @@ def _handle_version(key, value):  # pragma: no cover
     return value
 
 
-def download_gce_cert(gce_cert_url, gce_cert_path):
-    ssh.command('curl {0} -o {1}'.format(gce_cert_url, gce_cert_path))
-    if ssh.command('[ -f {} ]'.format(gce_cert_path)).return_code != 0:
+def download_gce_cert():
+    ssh.command('curl {0} -o {1}'.format(settings.gce.cert_url, settings.gce.cert_path))
+    if ssh.command('[ -f {} ]'.format(settings.gce.cert_path)).return_code != 0:
         raise GCECertNotFoundError(
-            "The GCE certificate {} is not found in satellite.".format(gce_cert_path))
+            "The GCE certificate in path {} is not found in satellite.".format(
+                settings.gce.cert_path))
+    return download_server_file('json', settings.gce.cert_url)
--- a/tests/foreman/api/test_computeresource_gce.py
+++ b/tests/foreman/api/test_computeresource_gce.py
@@ -1,0 +1,360 @@
+# -*- encoding: utf-8 -*-
+"""Tests for the GCE ``compute_resource`` paths.
+
+A full API reference for compute resources can be found here:
+http://www.katello.org/docs/api/apidoc/compute_resources.html
+
+:Requirement: Computeresource GCE
+
+:CaseAutomation: Automated
+
+:CaseLevel: Acceptance
+
+:CaseComponent: computeresources-gce
+
+:TestType: Functional
+
+:CaseImportance: High
+
+:Upstream: No
+"""
+import pytest
+import random
+
+from fauxfactory import gen_string
+from nailgun import entities
+from robottelo.config import settings
+from robottelo.constants import VALID_GCE_ZONES
+from robottelo.decorators import tier1, tier3
+
+
+GCE_SETTINGS = dict(
+    project_id=settings.gce.project_id,
+    client_email=settings.gce.client_email,
+    cert_path=settings.gce.cert_path,
+    zone=settings.gce.zone,
+    cert_url=settings.gce.cert_url)
+
+
+clouduser = gen_string('alpha')
+finishuser = gen_string('alpha')
+
+
+@pytest.fixture(scope='module')
+def module_gce_cloudimg(
+        module_architecture, module_gce_compute, module_os, gce_custom_cloudinit_uuid):
+    """Creates cloudinit image on GCE Compute Resource"""
+    cloud_image = entities.Image(
+        architecture=module_architecture, compute_resource=module_gce_compute,
+        name=gen_string('alpha'), operatingsystem=module_os,
+        username=clouduser, uuid=gce_custom_cloudinit_uuid,
+        user_data=True
+    ).create()
+    return cloud_image
+
+
+@pytest.fixture(scope='module')
+def module_gce_finishimg(module_architecture, module_gce_compute, module_os, gce_latest_rhel_uuid):
+    """Creates finish image on GCE Compute Resource"""
+    finish_image = entities.Image(
+        architecture=module_architecture, compute_resource=module_gce_compute,
+        name=gen_string('alpha'), operatingsystem=module_os,
+        username=finishuser, uuid=gce_latest_rhel_uuid
+    ).create()
+    return finish_image
+
+
+@pytest.fixture(scope='class')
+def gce_domain(module_org, module_location, module_smart_proxy):
+    """Sets Domain for GCE Host Provisioning"""
+    _, _, dom = settings.server.hostname.partition('.')
+    domain = entities.Domain().search(
+        query={
+            u'search': u'name="{0}"'.format(dom)
+        }
+    )
+    domain = domain[0].read()
+    domain.location.append(module_location)
+    domain.organization.append(module_org)
+    domain.dns = module_smart_proxy
+    domain.update(['dns', 'location', 'organization'])
+    return entities.Domain(id=domain.id).read()
+
+
+@pytest.fixture(scope='class')
+def gce_hostgroup(
+        module_architecture, module_gce_compute, gce_domain, module_location,
+        module_puppet_environment, module_smart_proxy, module_os, module_org,
+        module_partiontable, googleclient):
+    """Sets Hostgroup for GCE Host Provisioning"""
+    hgroup = entities.HostGroup(
+        architecture=module_architecture,
+        compute_resource=module_gce_compute,
+        domain=gce_domain,
+        location=[module_location],
+        environment=module_puppet_environment,
+        puppet_proxy=module_smart_proxy,
+        puppet_ca_proxy=module_smart_proxy,
+        root_pass=gen_string('alphanumeric'),
+        operatingsystem=module_os,
+        organization=[module_org],
+        ptable=module_partiontable).create()
+    return hgroup
+
+
+class TestGCEComputeResourceTestCases:
+    """Tests for ``api/v2/compute_resources``."""
+
+    @tier1
+    def test_positive_crud_gce_cr(self, module_org, module_location):
+        """Create, Read, Update and Delete GCE compute resources
+
+        :id: b324f8cd-d509-4d7a-b738-08aefafdc2b5
+
+        :expectedresults: GCE Compute resource should be created, read, updated and deleted
+
+        :CaseImportance: Critical
+
+        :CaseLevel: Component
+        """
+        cr_name = gen_string('alpha')
+        # Testing Create
+        compresource = entities.GCEComputeResource(
+            name=cr_name,
+            provider='GCE',
+            email=GCE_SETTINGS['client_email'],
+            key_path=GCE_SETTINGS['cert_path'],
+            project=GCE_SETTINGS['project_id'],
+            zone=GCE_SETTINGS['zone'],
+            organization=[module_org],
+            location=[module_location]
+        ).create()
+        assert compresource.name == cr_name
+        # Testing Update
+        new_name = gen_string('alpha')
+        description = gen_string('utf8')
+        new_zone = random.choice(VALID_GCE_ZONES)
+        compresource.name = new_name
+        compresource.description = description
+        compresource.zone = new_zone
+        compresource.update(['name', 'description', 'zone'])
+        # Testing Read
+        updated = entities.GCEComputeResource(id=compresource.id).read()
+        assert updated.name == new_name
+        assert updated.description == description
+        assert updated.zone == new_zone
+        # Testing Delete
+        updated.delete()
+        assert not entities.GCEComputeResource().search(
+            query={'search': 'name={}'.format(new_name)})
+
+    @pytest.mark.skip_if_open("BZ:1794744")
+    @tier3
+    def test_positive_check_available_images(self, module_gce_compute, googleclient):
+        """Verify all the images from GCE are available to select from
+
+        :id: 5cdfab18-a591-4442-8c19-a01e9b10ac36
+
+        :expectedresults: All the images from Google CR should be available to select in GCE CR
+
+        :CaseLevel: Integration
+        """
+        satgce_images = module_gce_compute.available_images()
+        gcloudclinet_images = googleclient.list_templates(True)
+        assert len(satgce_images) == len(gcloudclinet_images)
+
+    @tier3
+    def test_positive_check_available_networks(self, module_gce_compute, googleclient):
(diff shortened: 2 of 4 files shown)


tests/foreman/api/test_computeresource_gce.py
```diff
@@ -0,0 +1,250 @@
...
+            2. Add a finish template based image in GCE CR.
+
+        :expectedresults: Finish template image should be added in GCE CR along with username
+
+        :CaseLevel: Integration
+        """
+        assert GCEnvironment.finishimg(
+            gceclient).compute_resource.id == GCEnvironment.gce_compute().id
+        assert GCEnvironment.finishimg(gceclient).uuid == GCEnvironment.latest_rhel_uuid()
+        assert GCEnvironment.finishimg().username == GCEnvironment.finishimg_user
```
mirekdlugosz: This is one of weaknesses of that wrapper and API - it's not clear where test data comes from and what exactly is verified here. Since `GCEnvironment` controls both `finishimg` and `finishimg_user`, it looks more like unit test for wrapper that verifies if certain data is available through two different interfaces.

tests/foreman/api/test_computeresource_gce.py
```diff
@@ -0,0 +1,250 @@
...
+        :return:
+        """
+        super(GCEHostProvisioningTestCase, cls).tearDownClass()
+        cls.host.delete()
+        cls.actual_vm.wait_for_state('VmState.STOPPED', '2m')
+        assert cls.actual_vm.is_stopped
+
+    @tier1
+    def test_positive_host_provisioned(self):
+        assert self.host
```
mirekdlugosz: Can this assertion ever be false? `host` is created in class setup. If `Host.create()` throws exception, that test will not even run. And as far as I know, there is no way for `Host.create()` to silently return falsey value.

tests/foreman/api/test_computeresource_gce.py
```diff
@@ -0,0 +1,250 @@
...
+            }
+        }
+        cls.host = entities.Host(
+            architecture=GCEnvironment.architecture(), compute_attributes=cls.compute_attrs,
+            domain=GCEnvironment.domain(), hostgroup=GCEnvironment.hostgroup(gceclient),
+            image=GCEnvironment.finishimg(gceclient), organization=GCEnvironment.organization(),
+            operatingsystem=GCEnvironment.operatingsystem(), location=GCEnvironment.location(),
+            name=cls.hostname, provision_method='image', root_pass=gen_string('alphanumeric')
+        ).create()
+        cls.fullhostname = '{}.{}'.format(cls.hostname, GCEnvironment.domain().name).lower()
```
mirekdlugosz: That should be specified in relevant tests, not in class setup. Setup is reserved for time-consuming operations where results can be used by multiple tests.

tests/foreman/api/test_computeresource_gce.py
```diff
@@ -0,0 +1,250 @@
...
+        assert self.host.name == self.fullhostname
+        assert self.host.build_status_label == "Installed"
+
+    @tier3
+    def test_positive_host_ip(self):
+        self.host.ip = self.actual_vm.ip
+
+    @tier3
+    def test_positive_power_on_off(self):
+        self.host.power()
```
mirekdlugosz: Shouldn't you add `power_action="off"` argument? At least on Nailgun side, there doesn't seem to be any magic to decide which action should be called.

jyejare:
@mirekdlugosz , Thanks for all the comments, and extremely sorry for not putting Do_NOT_REVIEW or MERGE on PR. The PR is being worked on. I will have latest updates for you soon.

jyejare:
Test Results:
```
================ 1 failed, 6 passed, 2 skipped in 482.46 seconds =========
```

tests/foreman/api/test_computeresource_gce.py
```diff
@@ -0,0 +1,250 @@
...
+class GCEComputeResourceTestCase(APITestCase):
+    """Tests for ``api/v2/compute_resources``."""
+
+    @tier1
+    def test_positive_crud_gce_cr(self):
+        """Create GCE compute resources with different names
+
+        :id:
+
+        :expectedresults: GCE Compute resource should be created
```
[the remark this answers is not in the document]
jyejare: updated!

tests/foreman/api/test_computeresource_gce.py
```diff
@@ -0,0 +1,250 @@
...
+
+        :expectedresults: All the networks from Google CR should be available to select in GCE CR
+
+        :CaseLevel: Integration
+        """
+        assert len(
+            GCEnvironment.gce_compute().available_networks()
+        ) == len(
+            self.gclient.list_network()
+        )
```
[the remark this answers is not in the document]
jyejare: I prefer to not set a variable which won't be used multiple lines. still, I updated them for more clearity of assertions.

tests/foreman/api/test_computeresource_gce.py
```diff
@@ -0,0 +1,250 @@
...
+        :id:
+
+        :expectedresults: All the networks from Google CR should be available to select in GCE CR
+
+        :CaseLevel: Integration
+        """
+        assert len(
+            GCEnvironment.gce_compute().available_networks()
+        ) == len(
+            self.gclient.list_network()
```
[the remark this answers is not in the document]
jyejare: The half baked PR comment. Now it has a correct thing.

tests/foreman/api/test_computeresource_gce.py
```diff
@@ -0,0 +1,250 @@
...
+    def test_positive_check_available_flavors(self):
+        """Verify flavors from GCE are available to select
+
+        :id:
+
+        :expectedresults: All the flavors from Google CR should be available to select in GCE Host
+
+        :CaseLevel: Integration
+        """
+        assert len(GCEnvironment.gce_compute().available_flavors()) > 1
```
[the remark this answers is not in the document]
jyejare: There is Bug [1] Raised in BZ to get the exact count of flavors. Currently the count from Satellite, Google Cloud UI and WrappenAPI Google Cloud API, all of them show different count.

So think it as just a placeholder and the Test is already being skipped with that BZ now.

tests/foreman/api/test_computeresource_gce.py
jyejare: That can be fixed by passing a username parameter. But That way I will end up with multiple usernames and this being a helper function which sets only once, I dont want to do that.

tests/foreman/api/test_computeresource_gce.py
jyejare: Removed.

tests/foreman/api/test_computeresource_gce.py
jyejare: Probably not in current tests, but it will be used multiple times in future.

tests/foreman/api/test_computeresource_gce.py
jyejare: Half baked. Now the correct is there.

jyejare:
> All tests miss Id metadata, which cause Travis to fail. Of course these must be added.

Half baked PR. Now all has Id's.
> Overall this looks like heading in right direction, despite number of comments that I added. It adds coverage, which is obviously a good thing.

Great to hear that.
> I am not sold on `robottelo/api/provisioning.py` idea and implementation.
> 
>     * It duplicates some of the wrappers that we already have. I would rather re-use them, so if there is need for changes in future, we can change only one wrapper.

Are you talking about :

- robottelo/api/utils -> configure_provisioning: This sets up katello components as well which is not required for image-based provisioning. Hence It was required to separate Global Entities, Katello Entities and ComputeResource specific Entities, so that all of them can be used as per once convenience.
> 
>     * It is very heavy on singletons and deferred execution. Given these objects are supposed to be global in context of couple of test classes, it seems unnecessary complication. Just create all entities on class initialization.

Yeah, the scope part can be improved by ```object.methods``` instead of ```class.methods``` where we can initiate an object classwise. Should I raise a different PR for it or are you insisting on doing in this PR only ?
> 
>     * I believe these wrappers try to do too much, to the point that it's unclear what is the point of some of the tests. I left comment in on of them.

I fixed most of them. Probably it was there because the PR was half baked. Let me know if you have more.
>
>     * In general, that wrapper should be implemented in class setup or module-scoped fixture.

Need more context or what wrapper your specific on ?

jyejare:
@mirekdlugosz Hold the review as we are updating the provisioning library as fixtures.

jyejare:
@mirekdlugosz , @vijay8451 :

Results post fixturization:
```
$ py.test tests/foreman/api/test_computeresource_gce.py
collecting ... 2020-01-30 10:59:23 - conftest - DEBUG - Collected 9 test cases
2020-01-30 16:29:23 - robottelo.issue_handlers.bugzilla - DEBUG - Calling Bugzilla API for {'1794744'}
2020-01-30 16:29:34 - robottelo.ssh - DEBUG - Instantiated Paramiko client 0x7f23008da350
2020-01-30 16:29:34 - robottelo.ssh - INFO - Connected to [qe-sat67-rhel7-tier3.satqe.lab.eng.rdu2.redhat.com]
2020-01-30 16:29:34 - robottelo.ssh - INFO - >>> rpm -q satellite
2020-01-30 16:29:36 - robottelo.ssh - INFO - <<< stdout
satellite-6.7.0-5.beta.el7sat.noarch

2020-01-30 16:29:36 - robottelo.ssh - DEBUG - Destroyed Paramiko client 0x7f23008da350
2020-01-30 16:29:36 - robottelo.host_info - DEBUG - Host Satellite version: 6.7
2020-01-30 16:29:36 - robottelo.helpers - DEBUG - Generated file bz_cache.json with BZ collect data
collected 9 items                                                                                       

tests/foreman/api/test_computeresource_gce.py .s.s.....                                           [100%]

================================= 7 passed, 2 skipped in 444.26 seconds =================================
```

Also, PR is ready for review. and thats final I promise. :)

conftest.py
```diff
@@ -1,2 +1,201 @@
...
+        file_type='json')
+    yield gceclient
+    gceclient.disconnect()
+
+
+@pytest.fixture(scope='session')
+def gce_latest_rhel_uuid(googleclient):
+    max_rhel7_template = max(
+        [img.name for img in googleclient.list_templates(
+            True) if str(img.name).startswith('rhel-7')])
```
mirekdlugosz: List comprehension doesn't really serve readability of this. I would break it into two lines:
```python
template_names = [img.name for img in googleclient.list_templates(True)]
latest_rhel7_template = max(name for name in template_names if name.startswith('rhel-7'))
```

robottelo/helpers.py
```diff
@@ -1080,8 +1080,10 @@ def _handle_version(key, value):  # pragma: no cover
     return value
 
 
-def download_gce_cert(gce_cert_url, gce_cert_path):
-    ssh.command('curl {0} -o {1}'.format(gce_cert_url, gce_cert_path))
-    if ssh.command('[ -f {} ]'.format(gce_cert_path)).return_code != 0:
+def download_gce_cert():
+    ssh.command('curl {0} -o {1}'.format(settings.gce.cert_url, settings.gce.cert_path))
+    if not ssh.command('[ -f {} ]'.format(settings.gce.cert_path)).return_code == 0:
```
mirekdlugosz: In this line, previous way was more natural. It was "if command return is not equal zero" instead of "if not command return equals zero".

tests/foreman/api/test_computeresource_gce.py
```diff
@@ -0,0 +1,357 @@
...
+        gcloudclinet_images = googleclient.list_templates(True)
+        assert len(satgce_images) == len(gcloudclinet_images)
+
+    @tier3
+    def test_positive_check_available_networks(self, module_gce_compute, googleclient):
+        """Verify all the networks from GCE are available to select
+
+        :id: 259c9f5f-19dd-4e06-af5e-874612914f44
+
+        :expectedresults: All the networks from Google CR should be available to select in GCE CR
```
mirekdlugosz: Technically, this only tests if number of networks is the same, not if they are actually the same networks.

Does `module_gce_compute.available_networks()` provide list of networks as well? Perhaps you could put that into `set()` and compare against `set()` of `googleclient.list_network()`?

tests/foreman/api/test_computeresource_gce.py
```diff
@@ -0,0 +1,357 @@
...
+            query={'search': 'name={}'.format(new_name)})
+
+    @pytest.mark.skip_if_open("BZ:1794744")
+    @tier3
+    def test_positive_check_available_images(self, module_gce_compute, googleclient):
+        """Verify all the images from GCE are available to select from
+
+        :id: 5cdfab18-a591-4442-8c19-a01e9b10ac36
+
+        :expectedresults: All the images from Google CR should be available to select in GCE CR
```
mirekdlugosz: Same here - technically we only check if number of images is the same, not if these are the same images.

tests/foreman/api/test_computeresource_gce.py
```diff
@@ -0,0 +1,357 @@
...
+
+class TestGCEHostProvisioningTestCase:
+    """GCE Host Provisioning Tests
+
+    :Requirement: Host
+
+    :CaseComponent: Hosts
+    """
+    @pytest.fixture(scope='class')
+    def gce_domain(self, module_org, module_location, module_smart_proxy):
```
mirekdlugosz: Can you move these fixtures to top of file? Usually fixtures are defined as global functions, not as methods of class. having them here is unexpected.

conftest.py
```diff
@@ -1,2 +1,201 @@
 # coding: utf-8
```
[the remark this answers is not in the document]
jyejare: Yes, So I wanted to put that in tests/foreman, so that it will be used API,CLI and UI tests. Thanks for confirmation.

conftest.py
jyejare: Again, I dont waste variable if its not reusable, but updating for better understanding.

tests/foreman/api/test_computeresource_gce.py
jyejare: Actually what Google Compute returns is different in names is different than what Foreman GCE shows, Slight change in name. But I will be able to confirm/ fix this post the BZ bug for the test is fixed. This is just a placeholder for now.

tests/foreman/api/test_computeresource_gce.py
jyejare: Not mandatory to Keep in Module Scope but yes the newly added tests (by new contributor) for that person/tests it will be helpful.

Moving at module level.

jyejare:
Post Changes result:
```
===================== 7 passed, 2 skipped in 367.76 seconds ==================
```

tests/foreman/api/test_computeresource_gce.py
```diff
@@ -0,0 +1,360 @@
...
+            module_org, module_os, module_location, gce_latest_rhel_uuid, module_gce_finishimg):
+        """Provisions the host on GCE
+
+        Later in tests this host will be used to perform assertions
+        """
+        host = entities.Host(
+            architecture=module_architecture, compute_attributes=self.compute_attrs,
+            domain=gce_domain, hostgroup=gce_hostgroup, organization=module_org,
+            operatingsystem=module_os, location=module_location, name=self.hostname,
+            provision_method='image', image=module_gce_finishimg,
```
vijay8451: Do you plan to cover `user data/cloud-init ` host create as well ?

jyejare: The cloud-init has been covered by UI plus the CLI automation would be based on Cloud-init.

PR #7605, opened 2020-01-24
Remove rubygem-foreman-redhat_access, add InsightsInventoryPlugin and InsightsPlugin

mirekdlugosz:
Remove component that disappeared from bugzilla, add two components that appeared.
Based on discussion on today's standup, tests assigned to `rubygem-foreman-redhat_access` are moved to `InsightsPlugin`.

Notification: @sghai

### DIFF
--- a/testimony.yaml
+++ b/testimony.yaml
@@ -66,6 +66,8 @@ CaseComponent:
     - Hosts-Content
     - Infobloxintegration
     - Infrastructure
+    - InsightsInventoryPlugin
+    - InsightsPlugin
     - Installer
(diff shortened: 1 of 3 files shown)


mirekdlugosz:
We can wait with that until components are finalized sometime next week, OR we can merge it to align with what we have right now, and do any follow-ups once components are finalized.

After components finalization we will need to do more work anyway, so I vote for option 2.

JacobCallahan:
I'm agreeing with option 2, so I will merge. Thanks @mirekdlugosz

PR #7608, opened 2020-01-27
Added support for AzureRM properties

vijay8451:
- Added support for AzureRM properties

### DIFF
--- a/robottelo/constants.py
+++ b/robottelo/constants.py
@@ -39,6 +39,7 @@ FOREMAN_PROVIDERS = {
     'openstack': 'RHEL OpenStack Platform',
     'rackspace': 'Rackspace',
     'google': 'Google',
+    'azurerm': 'Azure Resource Manager',
 }
 
 EC2_REGION_CA_CENTRAL_1 = 'ca-central-1'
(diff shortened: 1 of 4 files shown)


robottelo/constants.py
```diff
@@ -83,6 +84,34 @@
...
+    'norwayeast': 'Norway East'}
+AZURERM_RHEL7_FT_IMG_URN = 'RedHat:RHEL:7-RAW:latest'
+AZURERM_RHEL7_UD_IMG_URN = 'RedHat:RHEL:7-RAW-CI:7.6.2019072418'
+AZURERM_RG_DEFAULT = 'SATQE'
+AZURERM_PLATFORM_DEFAULT = 'Linux'
+AZURERM_VM_SIZE_DEFAULT = 'Standard_B2ms'
+AZURERM_PREMIUM_OS_Disk = True
+AZURERM_FILE_URI = 'https://gist.githubusercontent.com/' \
+                   'vijay8451/72150405d99e0711056e50151251e2b1/raw/' \
+                   '734316fd56578e0b5777e686fb4119733ea457d9/check.sh'
```
mirekdlugosz: Try to avoid backslash unless they are absolutely necessary:
```python
AZURERM_FILE_URI = ('https://gist.githubusercontent.com/'
                    'vijay8451/72150405d99e0711056e50151251e2b1/raw/'
                    '734316fd56578e0b5777e686fb4119733ea457d9/check.sh')
```

Also - is there a chance to not use private gist here? Perhaps introduce new file into repo that will be uploaded?

robottelo/constants.py
```diff
@@ -83,6 +84,34 @@
 GCE_NETWORK_DEFAULT = 'default'
 GCE_EXTERNAL_IP_DEFAULT = True
 
+# AzureRM specific constants
+AZURERM_VALID_REGIONS = {
+    'eastasia': 'East Asia', 'southeastasia': 'Southeast Asia',
```
mirekdlugosz: For those of us missing the context - why is that a dict?

And would it make sense to generate dict dynamically? Looks like there is very simple transformation to make:
```python
AZURERM_VALID_REGIONS = ['East Asia', 'UK South', 'West US 2', ...]
AZURERM_VALID_REGIONS = {key.lower().replace(' ', '') : key for key in AZURERM_VALID_REGIONS}
```

robottelo/constants.py
vijay8451: Good observation :+1:  .. added it.

robottelo/constants.py
vijay8451: updated the same..

vijay8451: earlier I was thinking to keep the region as a part of dict because `wrapanapi and UI` needed in caps however CLI/API needed in lower case ...
But i think the lower case could handle at test case level hence keeping only list of caps regions now.
Thanks for your input. please revisit.

vijay8451:
@mirekdlugosz please revisit.

robottelo/constants.py
mirekdlugosz: Maybe `helpers` could have function to transform wrapanapi form to CLI form, and from CLI form to wrapanapi form. This can be introduced along with first test that actually needs it.

I think keeping simple list in constants is better choice.

Issue #7599, opened 2020-01-21
_by_ip will no longer be an option for rex under IPv6


pondrejk:
For IPv6 we'll need a way to run rex jobs on host by fqdn, as _by_ip setting (used predominantly now) won't be implemented

PR #7609, opened 2020-01-27
by-ip param set in fixture, removed some stubbs

pondrejk:
removed some less promising stubbs that hang around for long time, by_ip setting is now specified in fixture making it easier to change the setting for the whole class (#7599)

tests (standalone auto 1842):
```
py.test -v --junit-xml=foreman-results.xml -o junit_suite_name=standalone-automation -m 'not stubbed' tests/foreman/cli/test_remoteexecution.py

collected 13 items / 3 deselected / 10 selected

tests/foreman/cli/test_remoteexecution.py::TestRemoteExecution::test_positive_run_default_job_template_by_ip[rhel7] PASSED [ 10%]
tests/foreman/cli/test_remoteexecution.py::TestRemoteExecution::test_positive_run_job_effective_user_by_ip[rhel7] FAILED [ 20%]
tests/foreman/cli/test_remoteexecution.py::TestRemoteExecution::test_positive_run_custom_job_template_by_ip[rhel7] PASSED [ 30%]
tests/foreman/cli/test_remoteexecution.py::TestRemoteExecution::test_positive_run_default_job_template_multiple_hosts_by_ip[rhel7] PASSED [ 40%]
tests/foreman/cli/test_remoteexecution.py::TestRemoteExecution::test_positive_install_multiple_packages_with_a_job_by_ip[rhel7] PASSED [ 50%]
tests/foreman/cli/test_remoteexecution.py::TestRemoteExecution::test_positive_run_recurring_job_with_max_iterations_by_ip[rhel7] PASSED [ 60%]
tests/foreman/cli/test_remoteexecution.py::TestRemoteExecution::test_positive_run_scheduled_job_template_by_ip[rhel7] PASSED [ 70%]
tests/foreman/cli/test_remoteexecution.py::TestAnsibleREX::test_positive_run_effective_user_job[rhel7] PASSED [ 80%]
tests/foreman/cli/test_remoteexecution.py::TestAnsibleREX::test_positive_run_reccuring_job[rhel7] PASSED [ 90%]
tests/foreman/cli/test_remoteexecution.py::TestAnsibleREX::test_positive_run_packages_and_services_job[rhel7] PASSED [100%]
============= 1 failed, 9 passed, 3 deselected in 1589.96 seconds ==============
```

One failure unrelated

### DIFF
--- a/tests/foreman/cli/test_remoteexecution.py
+++ b/tests/foreman/cli/test_remoteexecution.py
@@ -37,13 +37,11 @@ from robottelo.constants import (
     FAKE_0_YUM_REPO
 )
 from robottelo.decorators import (
-    skip_if_not_set,
     stubbed,
     tier3,
     upgrade
 )
 from robottelo.helpers import add_remote_execution_ssh_key
-from robottelo.test import CLITestCase
 from robottelo.vm import VirtualMachine
 from time import sleep
 
@@ -52,7 +50,6 @@ import pytest
 TEMPLATE_FILE = u'template_file.txt'
 TEMPLATE_FILE_EMPTY = u'template_file_empty.txt'
 distros = [DISTRO_DEFAULT]
-
 
 @pytest.fixture(scope="module")
 def fixture_org():
@@ -66,7 +63,7 @@ def fixture_org():
 
 
 @pytest.fixture(params=distros, scope="module")
-def fixture_vmsetup(request, fixture_org):
+def fixture_vmsetup(request, fixture_org, subnet_id=None, by_ip=True):
     """Create Org, Lifecycle Environment, Content View, Activation key,
     VM, install katello-ca, register it, add remote execution key
     """
@@ -85,6 +82,18 @@ def fixture_vmsetup(request, fixture_org):
         )
         assert client.subscribed
         add_remote_execution_ssh_key(client.ip_addr)
+        if subnet_id is not None:
+            Host.update({
+                'name': client.hostname,
+                'subnet-id': subnet_id,
+            })
+        if by_ip:
+            # connect to host by ip
+            Host.set_parameter({
+                'host': client.hostname,
+                'name': 'remote_execution_connect_by_ip',
+                'value': 'True',
+            })
         yield client
     finally:
         client._hostname = None
@@ -93,33 +102,6 @@ def fixture_vmsetup(request, fixture_org):
 
 class TestRemoteExecution():
     """Implements job execution tests in CLI."""
-
-    @stubbed()
-    @tier3
-    def test_positive_run_job_multiple_hosts_time_span(self):
-        """Run job against multiple hosts with time span setting
-
-        :id: 82d69069-0592-4083-8992-8969235cc8c9
-
-        :expectedresults: Verify the jobs were successfully distributed
-            across the specified time sequence
-        """
-        # currently it is not possible to get subtasks from
-        # a task other than via UI
-
-    @stubbed()
-    @tier3
-    @upgrade
-    def test_positive_run_job_multiple_hosts_concurrency(self):
-        """Run job against multiple hosts with concurrency-level
-
-        :id: 15639753-fe50-4e33-848a-04fe464947a6
-
-        :expectedresults: Verify the number of running jobs does comply
-            with the concurrency-level setting
-        """
-        # currently it is not possible to get subtasks from
-        # a task other than via UI
 
     @tier3
     def test_positive_run_default_job_template_by_ip(self, fixture_vmsetup, fixture_org):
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_remoteexecution.py
```diff
@@ -65,7 +62,7 @@ def fixture_org():
 
 
 @pytest.fixture(params=distros, scope="module")
-def fixture_vmsetup(request, fixture_org):
+def fixture_vmsetup(request, fixture_org, subnet_id=None, by_ip=True):
```
mirekdlugosz: How are `subnet_id` and `by_ip` supposed to be set?

As far as I know, fixture has special properties that separate it from normal python function. Fixture is always pulled in by test as a dependency, and as such, is always referenced unparametrized. Fixture can be parametrized, but then all combinations of all parameters all called, creating multiple fixtures in process, and each test depending on that fixture is called multiple times (once for every unique combination).

To allow test to set `subnet_id` and `by_ip` (which I assume is what you want), you would need to modify this fixture to return function creating VM with parameters. Then tests would need to call that function explicitly:
```python
@pytest.fixture
def fixture_vmsetup(request):
    def inner(subnet_id=None, by_ip=True):
        # do what you have to do
        return created_vm
    return inner  # return reference to function - nothing is called yet at this point

def test_whatever(fixture_vmsetup):
    vm = fixture_vmsetup()  # call function returned by fixture
    # OR vm = fixture_vmsetup(by_ip=False)
```
One drawback of that approach is that every single test referencing this fixture would need to be modified, i.e. instead of `self.client = fixture_vmsetup` now you need `self.client = fixture_vmsetup()` (as fixture returns function, and you want to call that function and assign returned value.

Another problem is that in naive implementation (above), vm will be created independently in each function-returned-by-fixture invocation. So you need to introduce storage for created vms. I would go with "one VM for each unique combination of parameters" (so all tests calling `fixture_vmsetup()` will get the same VM). Of course that mandates re-implementation of teardown mechanism, as you are no longer using yield.

pondrejk: Hi @mirekdlugosz, these are good points but I wasn't really aiming for per-test parametrization, rather a way to easily turn off by_ip for the class when needed (as it will probably out or support under IPv6). I don't expect it will be worth to test by_ip along with by_hostname, I was thinking that the subnet_id and by_ip settings would be set globally according to some settings in robottelo.properties.

vijay8451: @pondrejk I agree with @mirekdlugosz  approach on it.
However another side of coin , afaik all rex tasks and tests in automation are running(get success) only by setting `remote_execution_connect_by_ip True` at host level hence as per my understanding we should to go with below straight forward approach and if someone really need to set this param to `False` at test level than this parameter could be override at test level it self ..
i.e.
```
@pytest.fixture(params=distros, scope="module")
def fixture_vmsetup(request, fixture_org):
    """Create Org, Lifecycle Environment, Content View, Activation key,
    VM, install katello-ca, register it, add remote execution key
    """
    # Create VM and register content host
    client = VirtualMachine(distro=request.param)
    try:
        client.create()
        if request.param in [DISTRO_SLES11, DISTRO_SLES12]:
            # SLES hostname in subscription-manager facts doesn't include domain
            client._hostname = client.hostname.split(".")[0]
        client.install_katello_ca()
        # Register content host
        client.register_contenthost(
            org=fixture_org.label,
            lce='Library'
        )
        assert client.subscribed
        add_remote_execution_ssh_key(client.ip_addr)
        Host.update({
            'name': client.hostname,
            'subnet-id': subnet_id,
            })
        Host.set_parameter({
            'host': client.hostname,
            'name': 'remote_execution_connect_by_ip',
            'value': 'True',
            })
        yield client
    finally:
        client._hostname = None
        client.destroy()
```

pondrejk: @mirekdlugosz @vijay8451 how about making the by_ip setting a fixture parameter before fixture_vmsetup like it is currently done with distros? This will allow 3 scenarios: run all tests with by_ip, without by_ip or with both.

vijay8451: @pondrejk could you please share an example how exactly you wants to achieve this  ... as that would help to understand the idea around it.

JacobCallahan: @pondrejk any luck with this?

pondrejk: @mirekdlugosz @vijay8451 @JacobCallahan looking back at this, I'd rework this to be put in after branching. I'd keep the PR open to preserve the discussion if that's OK.

mshriver:
@pondrejk I've marked this as draft as it appears to be forgotten/stale. Please rebase and update if its something you would like re-reviewed. From your last comments, it may be open to "preserve the discussion"?  The discussion will still be available if the PR is closed, and any important points from the discussion should be added to the code itself as in-line comments as to why something is written the way it is. 

Thanks!

mshriver:
With no response, I'm going to close this PR. Please update your branch and re-open if there are changes you would still like to propose.

PR #7623, opened 2020-02-03
test negative without attach rewritten and fixed

tstrych:
For a long time this test was failing, the whole logic of this test before does not make a sense to me, so with a help I change it. 

```
test_host.py::HostSubscriptionTestCase::test_negative_without_attach 
test_host.py::HostSubscriptionTestCase::test_negative_without_attach_with_lce 
================== 2 passed, 60 deselected in 343.34 seconds ===================
```

### DIFF
--- a/robottelo/vm.py
+++ b/robottelo/vm.py
@@ -376,10 +376,10 @@ class VirtualMachine(object):
         if force or settings.cdn or not downstream_repo:
             self.run(u'subscription-manager repos --enable {0}'.format(repo))
 
-    def list_repos(self):
+    def subscription_manager_list_repos(self):
         return self.run("subscription-manager repos --list")
 
-    def status(self):
+    def subscription_manager_status(self):
         return self.run("subscription-manager status")
 
     def create_custom_repos(self, **kwargs):
@@ -505,7 +505,7 @@ gpgcheck=0'''.format(
         cmd = u'subscription-manager register --org {0}'.format(org)
         if activation_key is not None:
             cmd += u' --activationkey {0}'.format(activation_key)
-        elif lce is not None:
+        elif lce:
             if username is None and password is None:
                 username = settings.server.admin_username
                 password = settings.server.admin_password
@@ -515,7 +515,7 @@ gpgcheck=0'''.format(
             )
             if auto_attach:
                 cmd += u' --auto-attach'
-        elif consumerid is not None:
+        elif consumerid:
             if username is None and password is None:
                 username = settings.server.admin_username
                 password = settings.server.admin_password
--- a/robottelo/vm.py
+++ b/robottelo/vm.py
@@ -376,6 +376,12 @@ class VirtualMachine(object):
         if force or settings.cdn or not downstream_repo:
             self.run(u'subscription-manager repos --enable {0}'.format(repo))
 
+    def list_repos(self):
+        return self.run("subscription-manager repos --list")
+
+    def status(self):
+        return self.run("subscription-manager status")
+
     def create_custom_repos(self, **kwargs):
         """Create custom repofiles.
         Each ``kwargs`` item will result in one repository file created. Where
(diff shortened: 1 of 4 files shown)


robottelo/vm.py
```diff
@@ -365,6 +365,12 @@ def enable_repo(self, repo, force=False):
         if force or settings.cdn or not downstream_repo:
             self.run(u'subscription-manager repos --enable {0}'.format(repo))
 
+    def list_repos(self):
+        return self.run("subscription-manager repos --list")
+
+    def status(self):
```
JacobCallahan: This method name could be more descriptive. Looking at the name, a user isn't sure of what kind of status they are getting. maybe ```sub_status``` or ```subscription_status```?

robottelo/vm.py
```diff
@@ -498,6 +506,18 @@ def register_contenthost(self, org, activation_key=None, lce=None,
             )
             if auto_attach:
                 cmd += u' --auto-attach'
+        elif consumerid is not None:
```
JacobCallahan: This conditional will also pass for ```"", [], {}``` and other "empty" values you may not want. 
To avoid this, the conditional could simply be changed to ```elif consumerid:```

robottelo/vm.py
```diff
@@ -498,6 +506,18 @@ def register_contenthost(self, org, activation_key=None, lce=None,
...
+                username = settings.server.admin_username
+                password = settings.server.admin_password
+
+            cmd += u' --consumerid {0} --username {1} --password {2}'.format(
+                consumerid,
+                username,
+                password,
+            )
+            if auto_attach:
+                cmd += u' --auto-attach'
```
JacobCallahan: you can strip the ```u``` from your string definitions, as we don't need to support python 2 anymore.

robottelo/vm.py
tstrych: Okay,  I will do that, 
just this is vm module, so user can expect subscription manager,
but in future there can be another status, 
but still when you look at function, what I always do, I can see the only line where is the `subscription-manager status`

robottelo/vm.py
tstrych: @JacobCallahan  also agree, but I did it this way because I want continue with same code style as in these [lines right above](https://github.com/SatelliteQE/robottelo/pull/7623/files#diff-2be7fd316a53f5055bef542dab4c3ad7R487-R505)

robottelo/vm.py
tstrych: from my point of view this should be done as one PR, where all `u'` definitions should be removed.

tstrych: or at least, I can clear the whole file in another PR, I don't think it's good idea to mix this stuff together

tstrych:
`tests/robottelo/test_vm.py` results:

together with test_host_info.py, why ? - explained in #7635 

```py.test tests/robottelo/test_host_info.py tests/robottelo/test_vm.py
== test session starts ==                                                                                                                                                                                                                                             

tests/robottelo/test_host_info.py ................                                                                                                                                                                                                               [ 64%]
tests/robottelo/test_vm.py .........                                                                                                                                                                                                                             [100%]

== 25 passed, 2 warnings in 0.29 seconds ==
```
deprecation warnings omitted

tests/robottelo/test_vm.py
```diff
@@ -6,9 +6,13 @@
 
 from robottelo import ssh
 from robottelo.config.base import DistroSettings
+from robottelo.constants import NO_REPOS_AVAILABLE
+from robottelo.constants import SM_OVERALL_STATUS
 from robottelo.vm import VirtualMachine
 from robottelo.vm import VirtualMachineError
 
+from unittest.mock import call, patch
```
mirekdlugosz: Looks like `reorder-python-imports` in `pre-commit` doesn't like that line - I guess it would prefer to have it split to two lines.

Best to install `pre-commit` locally and run it, then commit changes it did.

PR #7624, opened 2020-02-03
Wait for image tags to be non-zero unconditionally

mirekdlugosz:
bz1664631 has been closed WONTFIXed.


mirekdlugosz:
@rplevka Fair enough.
I added docstring with explanation and reference to bz, and force-pushed changes.

Issue #310, opened 2014-01-23
UI: Validate given URL is valid and exists


sghai:
URL validation is required for helper function `download_template` under `robottelo/common/helper.py

I tried to implement but I need to validate it with different request codes. Currently, I'm getting 502, while accessing the correct github.gist URL. Ideally, its should be 200.
Will try to fix it.

PR #7625, opened 2020-02-05
Added API tests around AzureRM - Part1

vijay8451:
- nailgun PR https://github.com/SatelliteQE/nailgun/pull/697
- test results :
```
api/test_computeresource_azurerm.py::TestAzureRMComputeResourceTestCase::test_positive_crud_azurerm_cr PASSED                                                  [ 16%]
api/test_computeresource_azurerm.py::TestAzureRMComputeResourceTestCase::test_positive_create_finish_template_image PASSED                                            [ 33%]
api/test_computeresource_azurerm.py::TestAzureRMComputeResourceTestCase::test_positive_create_cloud_init_image PASSED                                                 [ 50%]
api/test_computeresource_azurerm.py::TestAzureRMComputeResourceTestCase::test_positive_check_available_networks PASSED                                                [ 66%]
api/test_computeresource_azurerm.py::TestAzureRMHostProvisioningTestCase::test_positive_azurerm_host_provisioned PASSED                                               [ 83%]
api/test_computeresource_azurerm.py::TestAzureRMHostProvisioningTestCase::test_positive_azurerm_host_power_on_off PASSED                                              [100%]

======================================================================== 6 passed in 892.47 seconds =========================================================================

```

### DIFF
--- a/tests/foreman/api/test_computeresource_azurerm.py
+++ b/tests/foreman/api/test_computeresource_azurerm.py
@@ -79,7 +79,7 @@ class TestAzureRMComputeResourceTestCase:
         description = gen_string('utf8')
         compresource.name = new_cr_name
         compresource.description = description
-        compresource.update(['name', 'description'])
+        compresource = compresource.update(['name', 'description'])
         assert compresource.name == new_cr_name
         assert compresource.description == description
 
--- a/tests/foreman/api/test_computeresource_azurerm.py
+++ b/tests/foreman/api/test_computeresource_azurerm.py
@@ -1,0 +1,288 @@
+"""Test class for AzureRM Compute Resource
+
+:Requirement: ComputeResources AzureRM
+
+:CaseAutomation: Automated
+
+:CaseLevel: Component
+
+:CaseComponent: ComputeResources-Azure
+
+:TestType: Functional
+
+:CaseImportance: High
+
+:Upstream: No
+"""
+import pytest
+
+from fauxfactory import gen_string
+from nailgun import entities
+from robottelo.api.utils import skip_yum_update_during_provisioning
+from robottelo.config import settings
+from robottelo.constants import (
+    AZURERM_FILE_URI,
+    AZURERM_RHEL7_FT_IMG_URN,
+    AZURERM_RHEL7_UD_IMG_URN,
+    AZURERM_RG_DEFAULT,
+    AZURERM_PREMIUM_OS_Disk,
+    AZURERM_PLATFORM_DEFAULT,
+    AZURERM_VM_SIZE_DEFAULT,
+)
+from robottelo.decorators import (
+    tier1,
+    tier2,
+    tier3,
+    upgrade,
+)
+
+
+class TestAzureRMComputeResourceTestCase:
+    """Tests for ``api/v2/compute_resources``"""
+
+    @upgrade
+    @tier1
+    def test_positive_crud_azurerm_cr(self, module_org, module_location, azurerm_settings):
+        """Create, Read, Update and Delete AzureRM compute resources
+
+        :id: da081a1f-9614-4918-91cb-c900c40ac121
+
+        :expectedresults: Compute resource should be created, read, updated and deleted
+
+        :CaseImportance: Critical
+
+        :CaseLevel: Component
+        """
+
+        # Create CR
+        cr_name = gen_string('alpha')
+        compresource = entities.AzureRMComputeResource(
+            name=cr_name,
+            provider='AzureRm',
+            tenant=azurerm_settings['tenant'],
+            app_ident=azurerm_settings['app_ident'],
+            sub_id=azurerm_settings['sub_id'],
+            secret_key=azurerm_settings['secret'],
+            region=azurerm_settings['region'],
+            organization=[module_org],
+            location=[module_location],
+        ).create()
+        assert compresource.name == cr_name
+        assert compresource.provider == 'AzureRm'
+        assert compresource.tenant == azurerm_settings['tenant']
+        assert compresource.app_ident == azurerm_settings['app_ident']
+        assert compresource.sub_id == azurerm_settings['sub_id']
+        assert compresource.region == azurerm_settings['region']
+
+        # Update CR
+        new_cr_name = gen_string('alpha')
+        description = gen_string('utf8')
+        compresource.name = new_cr_name
+        compresource.description = description
+        compresource.update(['name', 'description'])
+        assert compresource.name == new_cr_name
+        assert compresource.description == description
+
+        # Delete CR
+        compresource.delete()
+        assert not entities.AzureRMComputeResource().search(
+            query={'search': 'name={}'.format(new_cr_name)})
+
+    @upgrade
+    @tier2
+    def test_positive_create_finish_template_image(self,
+                                                   module_architecture,
+                                                   module_azurerm_cr,
+                                                   module_azurerm_finishimg):
+        """ Finish template image along with username is being added in AzureRM CR
+
+        :id: 78facb19-4b27-454b-abc5-2c69c0a6c28a
+
+        :steps:
+            1. Create a AzureRM Compute Resource.
+            2. Add a finish template based image in it.
+
+        :expectedresults: Finish template image should be added in AzureRM CR along with username
+
+        :CaseImportance: Critical
+
+        :CaseLevel: Integration
+        """
+
+        assert module_azurerm_finishimg.architecture.id == module_architecture.id
+        assert module_azurerm_finishimg.compute_resource == module_azurerm_cr
+        assert module_azurerm_finishimg.username == settings.azurerm.username
+        assert module_azurerm_finishimg.uuid == AZURERM_RHEL7_FT_IMG_URN
+
+    @upgrade
+    @tier2
+    def test_positive_create_cloud_init_image(self,
+                                              module_azurerm_cloudimg,
+                                              module_azurerm_cr,
+                                              module_architecture):
+        """Cloud Init template image along with username is being added in AzureRM CR
+
+        :id: 05ea1b20-0dfe-4af3-b1b7-a82604aa1e79
+
+        :steps:
+            1. Create a AzureRM Compute Resource.
+            2. Add a Cloud init supported image in it.
+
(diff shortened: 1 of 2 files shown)


tests/foreman/api/test_computeresource_azurerm.py
```diff
@@ -0,0 +1,288 @@
...
+        assert compresource.app_ident == azurerm_settings['app_ident']
+        assert compresource.sub_id == azurerm_settings['sub_id']
+        assert compresource.region == azurerm_settings['region']
+
+        # Update CR
+        new_cr_name = gen_string('alpha')
+        description = gen_string('utf8')
+        compresource.name = new_cr_name
+        compresource.description = description
+        compresource.update(['name', 'description'])
```
mirekdlugosz: You should assign update result back to variable to ensure that these values are what you expect **after** updating. Otherwise, if update silently fails, your test will still pass.
```suggestion
        compresource = compresource.update(['name', 'description'])
```

tests/foreman/api/test_computeresource_azurerm.py
```diff
@@ -0,0 +1,288 @@
...
+            name=self.hostname,
+            provision_method='image',
+            image=module_azurerm_finishimg,
+            root_pass=gen_string('alphanumeric'),
+            environment=module_puppet_environment,
+            puppet_proxy=module_smart_proxy,
+            puppet_ca_proxy=module_smart_proxy,
+        ).create()
+        yield host
+        skip_yum_update_during_provisioning(template='Kickstart default finish', reverse=True)
```
mirekdlugosz: Why is host provisioned in fixture teardown? Shouldn't that be done before `yield host`? Or just skipped altogether?

tests/foreman/conftest.py
```diff
@@ -389,3 +396,77 @@ def module_gce_compute(module_org, module_location):
         zone=settings.gce.zone, organization=[module_org],
         location=[module_location]).create()
     return gce_cr
+
+
+@pytest.fixture(scope='module')
+def azurerm_settings():
```
mirekdlugosz: Do we really need that as global fixture? Can't you have it as fixture in `tests/foreman/api/test_computeresource_azurerm.py`, or even function there?

The same applies to other fixtures below.

tests/foreman/api/test_computeresource_azurerm.py
vijay8451: after the yield statement serves as the teardown code :)

tests/foreman/api/test_computeresource_azurerm.py
vijay8451: good catch .. will update it .

tests/foreman/conftest.py
vijay8451: Yeah I would like these to be part of `conftest` and wherever it would need further can utilize at any test level.

tests/foreman/api/test_computeresource_azurerm.py
vijay8451: it is pushed on new commit now.

tests/foreman/api/test_computeresource_azurerm.py
mirekdlugosz: I know, I asked for a reason why.
But I know now, we first break kickstart template, and then feel compelled to return it back to what Robottelo thinks it was. There are obvious issues with that approach, starting with race conditions, but I see you just follow pattern established elsewhere.

tests/foreman/conftest.py
mirekdlugosz: I would rather put that at the most specific level and move to more general levels if we actually have something that uses it, rather than the other way around. You know, YAGNI.

tests/foreman/api/test_computeresource_azurerm.py
vijay8451: ahh ..okay ..  so could you please let me know in which case we could see race condition?

tests/foreman/conftest.py
vijay8451: I am fine to put them at specific level as well however the approach is used as similar to:
https://github.com/SatelliteQE/robottelo/blob/80beb97d8762146893c5c1b45f9e0aedab942539/tests/foreman/conftest.py#L364-L398

tests/foreman/api/test_computeresource_azurerm.py
mirekdlugosz: Let's say we have two tests following that pattern running in parallel. Following sequence of events is possible:
1. Test A does `skip_yum_update_during_provisioning(template='Kickstart default finish')`
2. Test B does `skip_yum_update_during_provisioning(template='Kickstart default finish')` (effectively no-op, I presume)
3. Test A does it's thing
4. Test B does expensive operation
5. Test A finishes and runs teardown, doing `skip_yum_update_during_provisioning(template='Kickstart default finish', reverse=True)`
6. Test B does it's thing, still thinking kickstart modification is in effect; it's unaware that Test A reverted that in the meantime.

Obviously, each test should have it's own kickstart template with content it needs. Right now we have bunch of random tests modifying global state at will.

As I said, I see you just reused pattern established elsewhere, so I don't expect you to fix that as part of your PR. But if you wan't to, be my guest.

tests/foreman/conftest.py
mirekdlugosz: Yes, and in that PR I also suggested to put these fixtures closer to where they are used.

tests/foreman/api/test_computeresource_azurerm.py
vijay8451: Agree with you on it ...but first would like to observe the automation results around these tests and then would put a call to clone the existing template and run `skip_yum_update_during_provisioning` on that .

tests/foreman/conftest.py
vijay8451: Hummm ... I have seen in another PR https://github.com/SatelliteQE/robottelo/pull/7335 @jyejare where he is putting them separately now , So if that approach looks good then I'll put same there ..
but yeah for that I'll raise another PR.

tests/foreman/api/test_computeresource_azurerm.py
```diff
@@ -0,0 +1,288 @@
...
+        assert compresource.description == description
+
+        # Delete CR
+        compresource.delete()
+        assert not entities.AzureRMComputeResource().search(
+            query={'search': 'name={}'.format(new_cr_name)})
+
+    @upgrade
+    @tier2
+    def test_positive_create_finish_template_image(self,
```
jyejare: Move all params to the second line, they should be fitting in one line. Currently they all looks like hanging.

tests/foreman/api/test_computeresource_azurerm.py
```diff
@@ -0,0 +1,288 @@
...
+        :id: aee5d077-668e-4f4d-adee-6472f0e4ecaa
+
+        :expectedresults: All the networks from AzureRM CR should be available.
+
+        :CaseLevel: Integration
+        """
+
+        cr_nws = module_azurerm_cr.available_networks()
+        portal_nws = azurermclient.list_network()
+        assert len(portal_nws) == len(cr_nws['results'])
```
jyejare: Can we compare with actual network names instead of their length of them?

tests/foreman/api/test_computeresource_azurerm.py
```diff
@@ -0,0 +1,288 @@
...
+        request.cls.platform = AZURERM_PLATFORM_DEFAULT
+        request.cls.vm_size = AZURERM_VM_SIZE_DEFAULT
+        request.cls.hostname = gen_string('alpha')
+        request.cls.fullhostname = '{}.{}'.format(self.hostname, module_domain.name).lower()
+
+        request.cls.compute_attrs = {
+            "resource_group": self.rg_default,
+            "vm_size": self.vm_size,
+            "username": settings.azurerm.username,
+            "password": settings.azurerm.password,
```
jyejare: I think this could be randomly generated, need not to pass ot from properties.

tests/foreman/conftest.py
```diff
@@ -389,3 +396,77 @@ def module_gce_compute(module_org, module_location):
...
+        name=gen_string('alpha'),
+        operatingsystem=module_os,
+        username=settings.azurerm.username,
+        uuid=AZURERM_RHEL7_UD_IMG_URN,
+        user_data=True
+    ).create()
+    return finish_image
+
+
+@pytest.fixture(scope='module')
```
jyejare: session scope ?

tests/foreman/conftest.py
```diff
@@ -389,3 +396,77 @@ def module_gce_compute(module_org, module_location):
         zone=settings.gce.zone, organization=[module_org],
         location=[module_location]).create()
     return gce_cr
+
+
+@pytest.fixture(scope='module')
```
jyejare: session scope? We are not going to change the properties AFAIK in any tiers / test.

tests/foreman/api/test_computeresource_azurerm.py
vijay8451: done

tests/foreman/api/test_computeresource_azurerm.py
vijay8451: no , as both end provided same network in different ways ..hence only idea to see the count here.

tests/foreman/api/test_computeresource_azurerm.py
vijay8451: done.

tests/foreman/conftest.py
vijay8451: good catch .. we could.

tests/foreman/conftest.py
vijay8451: done.

vijay8451:
> ACK pending comments/questions.

Comments are addressed in recent push .. thanks

> Also, I didn't see constants for AzureRM declared in this PR, are they ? If not have you raised a PR for constants separately ?

Support for properties and constant are already added in robottelo in under PR https://github.com/SatelliteQE/robottelo/pull/7608

vijay8451:
MR #310 raised as well.

vijay8451:
@jyejare please revisit and possible merge :)

PR #7627, opened 2020-02-06
Implement test generating Subscriptions report, in JSON format

lhellebr:


tests/foreman/ui/test_reporttemplates.py
```diff
@@ -233,6 +234,31 @@ def test_positive_generate_registered_hosts_report(session, module_org, module_l
...
+
+    :CaseImportance: Medium
+    """
+    # generate Subscriptions report
+    with session:
+        file_path = session.reporttemplate.generate("Subscriptions",
+                                                    values={'output_format': 'JSON'})
+    with open(file_path) as json_file:
+        data = json.load(json_file)
+        assert len(data) > 0
```
mirekdlugosz: I suggest to de-indent that and further lines, so if assertion fails, we don't leave file opened (pytest will complain file was left opened, but other than that, I don't think it has any practical consequences; it's just good practice).

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -233,6 +234,31 @@ def test_positive_generate_registered_hosts_report(session, module_org, module_l
...
+        for subscription in data:
+            keys_expected = ['ID', 'Name', 'Available', 'Quantity', 'SKU', 'Contract number']
```
mirekdlugosz: Could you move that up, so we don't re-define variable on every loop iteration (I fully expect compiler to optimize it away, but still).

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -233,6 +234,31 @@ def test_positive_generate_registered_hosts_report(session, module_org, module_l
...
+    :expectedresults: The Subscriptions report is generated in JSON
+
+    :CaseLevel: Integration
+
+    :CaseImportance: Medium
+    """
+    # generate Subscriptions report
+    with session:
+        file_path = session.reporttemplate.generate("Subscriptions",
+                                                    values={'output_format': 'JSON'})
```
mirekdlugosz: Shouldn't you first upload manifest, or create product?
Looks like this report is using subscriptions from all organizations, but if there is no manifest in your Satellite at all, generated file will just contain `[]` (empty array).

lhellebr: I thought there is some manifest already in the environment when the tests are run. Isn't that the case?

lhellebr:
Depends on https://github.com/SatelliteQE/airgun/pull/445

tests/foreman/ui/test_reporttemplates.py
mirekdlugosz: I don't think we upload manifest as part of environment preparation.

If manifest will be already available when this test is run, it's most likely only because some other test uploaded it. Depending on test execution order has many drawbacks, which I probably don't have to state here.

If you are confident this test will always run after some other test uploading manifest, and you want to leave it as that, I won't block you.

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -233,6 +234,31 @@ def test_positive_generate_registered_hosts_report(session, module_org, module_l
...
+    with open(file_path) as json_file:
+        data = json.load(json_file)
+        assert len(data) > 0
+        for subscription in data:
+            keys_expected = ['ID', 'Name', 'Available', 'Quantity', 'SKU', 'Contract number']
+            assert sorted(list(subscription.keys())) == sorted(keys_expected)
+
```
pondrejk: I wonder if we should also check the number of subscriptions in the report against the number of known subs (obtaining it via API would be the fastest option). The test currently checks if the report is not empty and doesn't contain rubbish, but if the report drops a sub, we wouldn't know

pondrejk:
+1 to Mirek's comments, pending test resutls

lhellebr:
I finally found time to address your comments. Can we make this happen, please? @pondrejk @mirekdlugosz 

```
$ pytest tests/foreman/ui/test_reporttemplates.py::test_positive_generate_subscriptions_report_json
platform linux -- Python 3.6.6, pytest-4.6.3, py-1.7.0, pluggy-0.12.0
rootdir: /home/lhellebr/git/robottelo, inifile: pytest.ini
collecting ... 2020-03-13 14:39:21 - conftest - DEBUG - Collected 1 test cases
2020-03-13 15:39:23 - robottelo.ssh - DEBUG - Instantiated Paramiko client 0x7fce971062e8
2020-03-13 15:39:23 - robottelo.ssh - INFO - Connected to [dhcp-3-242.vms.sat.rdu2.redhat.com]
2020-03-13 15:39:23 - robottelo.ssh - INFO - >>> rpm -q satellite
2020-03-13 15:39:25 - robottelo.ssh - INFO - <<< stdout
satellite-6.7.0-6.el7sat.noarch

2020-03-13 15:39:25 - robottelo.ssh - DEBUG - Destroyed Paramiko client 0x7fce971062e8
2020-03-13 15:39:25 - robottelo.host_info - DEBUG - Host Satellite version: 6.7
2020-03-13 15:39:25 - robottelo.helpers - DEBUG - Generated file bz_cache.json with BZ collect data
collected 1 item                                                                                                       

tests/foreman/ui/test_reporttemplates.py .                                                                       [100%]

============================================== 1 passed in 252.30 seconds ==============================================
```

JacobCallahan:
black is also failing. simply 
```pre-commit run --all-files```
and handle the rest

lhellebr:
@mirekdlugosz @JacobCallahan Hopefully fixed; everything seems fine, the check passed, it just doesn't seem to be up to date here in GitHub.

PR #7638, opened 2020-02-11
automation failure, power hosts permissions in host filter was missing

tstrych:
Test result.
`================== 1 passed, 28 deselected in 129.66 seconds ===================`

### DIFF
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -865,6 +865,7 @@ def test_negative_delete_primary_interface(session, module_host_template):
         assert 'Interface Delete button is disabled' in str(context.value)
 
 
+@pytest.mark.skip_if_open("BZ:1801630")
 @tier2
 def test_positive_view_hosts_with_non_admin_user(test_name, module_org, module_loc):
     """View hosts and content hosts as a non-admin user with only view_hosts, edit_hosts


tstrych:
@mirekdlugosz  Hi, I know it past in the past, what is funny, I run it multiple times in my local management, but after merge when I look into history of automation snaps, I can see it never pass. 
Not sure if there was a change after I tried it last time and first automation run. Maybe `power hosts `permissions were added.

What I know for sure there is a bug on infinite spinner if there is no `power_hosts` permissions there. And I already created the [bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1801630).

So I can create there is another option and probably better one, to skip test if bug is open. Solved bug, should solve this test-failiure in future.

tstrych:
`=================== 1 skipped, 28 deselected in 3.57 seconds ===================`

mirekdlugosz:
@tstrych thanks for bugzilla.

Here's what is going on: Airgun runs [`ensure_page_safe`](https://github.com/SatelliteQE/airgun/blob/master/airgun/browser.py#L609) before interacting with page. This function runs some javascript on browser and expects to get dictionary with only truthy values in `values()` (you can see implementation in [widgetastic](https://github.com/RedHatQE/widgetastic.core/blob/master/src/widgetastic/browser.py#L45), and `ENSURE_PAGE_SAFE` in `airgun/browser.py`).

One of the things we check is that there is no visible element with class "spinner" (that was introduced in https://github.com/SatelliteQE/airgun/pull/243). This is check that fails in this test - spinners are visible, because Satellite is trying to load power state of hosts, but that can't be loaded due to lack of necessary permission. As a result, Airgun thinks that page is not fully loaded.

With all that in mind, I must take back my previous comment and admit that your initial implementation was probably better idea. We have broken that test in our automation, we should fix it in our automation. There's probably no Satellite bug to fix.

We can wait for devs opinion before proceeding with that, but with everything I know now, I would ACK your previous version.

tstrych:
Yes, but from development point of view, there is bug, because there is no reason why the spinner should spin forever, there should be an empty space in that field. Of course the decision can wait for devs, but it can take a long time.

PR #7641, opened 2020-02-13
Add pre-commit, apply to all files

mshriver:
Include python-reorder, flake, and local uuids-fix to start

Remove pylint config, makefile commands for flake/lint/pre-commit

Run pre-commit on all files in travis

gitignore venv/egg directories

Docs should be updated to reflect these changes, if I get agreement on making this switch I'll make those updates.


rochacbruno:
As we are already touching almost every file in this PR why not applying `black`? so it will be just one PR to review.

@mshriver @JacobCallahan

mshriver:
@rochacbruno absolutely we'll be using `black` via pre-commit, but I think that they are separate considerations for the team. One can go without the other, and vice versa.

It will take some rebasing, but I'd rather consider using pre-commit first, and then add more checks to pre-commit once we have the team using it.

My preference would be to consider them as separate PRs/changes, but I can certainly be convinced otherwise.

JacobCallahan:
@rochacbruno i'll revive the black PR once this is merged, as I can just add black to the pre-commit

vijay8451:
@mshriver  is same change going to adopt in 6.6.z/6.5.z/6.4.z branches  as well ?

PR #7644, opened 2020-02-18
modified assert in test org end-to-end

pondrejk:
Occasionally in automation, the removed ptables and templates are not found in the unassigned list.  While not to reproduce this outside of auto, I decided to compare the length of lists in hope to avoid the issue.

```
pytest tests/foreman/ui/test_organization.py -k end_to_end
tests/foreman/ui/test_organization.py .                                                                        [100%]

================================ 1 passed, 5 deselected, 4 warnings in 265.97 seconds ================================
```


JacobCallahan:
@pondrejk can you made the new variable names more indicative of what they represent?
maybe something like ```ptables_before_remove``` or something similar?

PR #7651, opened 2020-02-20
UI report

ColeHiggins2:
Continuation of https://github.com/SatelliteQE/robottelo/pull/7637

PR #7652, opened 2020-02-20
CLI scheduling Test and API fixture update

ColeHiggins2:
CLI test for scheduling reports using the Entitlements report template.

platform linux -- Python 3.7.5, pytest-4.6.3, py-1.8.0, pluggy-0.13.1 -- /home/colehiggins/projects/venv/bin/python3
cachedir: .pytest_cache
rootdir: /home/colehiggins/projects/robottelo
plugins: mock-1.10.4, xdist-1.31.0, forked-1.1.3, cov-2.8.1, services-1.3.1
collecting ... 2020-02-20 19:08:34 - conftest - DEBUG - Collected 1 test cases
2020-02-20 14:08:35 - robottelo.ssh - DEBUG - Instantiated Paramiko client 0x7f230b5d65d0
2020-02-20 14:08:35 - robottelo.ssh - INFO - Connected to [dhcp-2-221.vms.sat.rdu2.redhat.com]
2020-02-20 14:08:35 - robottelo.ssh - INFO - >>> rpm -q satellite
2020-02-20 14:08:37 - robottelo.ssh - INFO - <<< stdout
satellite-6.7.0-5.beta.el7sat.noarch

2020-02-20 14:08:37 - robottelo.ssh - DEBUG - Destroyed Paramiko client 0x7f230b5d65d0
2020-02-20 14:08:37 - robottelo.host_info - DEBUG - Host Satellite version: 6.7
2020-02-20 14:08:37 - robottelo.helpers - DEBUG - Generated file bz_cache.json with BZ collect data
collected 1 item

test_reporttemplates.py::ReportTemplateTestCase::test_positive_schedule_Entitlements_report 

========================== 1 passed in 214.77 seconds ==========================

Updated API with fixtures for future tests.

platform linux -- Python 3.7.5, pytest-4.6.3, py-1.8.0, pluggy-0.13.1 -- /home/colehiggins/projects/venv/bin/python3
cachedir: .pytest_cache
rootdir: /home/colehiggins/projects/robottelo
plugins: mock-1.10.4, xdist-1.31.0, forked-1.1.3, cov-2.8.1, services-1.3.1
collecting ... 2020-02-20 19:16:53 - conftest - DEBUG - Collected 1 test cases
2020-02-20 14:16:53 - robottelo.issue_handlers.bugzilla - DEBUG - Calling Bugzilla API for {'1680458'}
2020-02-20 14:16:54 - robottelo.ssh - DEBUG - Instantiated Paramiko client 0x7fa0a1447bd0
2020-02-20 14:16:54 - robottelo.ssh - INFO - Connected to [dhcp-2-221.vms.sat.rdu2.redhat.com]
2020-02-20 14:16:54 - robottelo.ssh - INFO - >>> rpm -q satellite
2020-02-20 14:16:56 - robottelo.ssh - INFO - <<< stdout
satellite-6.7.0-5.beta.el7sat.noarch

2020-02-20 14:16:56 - robottelo.ssh - DEBUG - Destroyed Paramiko client 0x7fa0a1447bd0
2020-02-20 14:16:56 - robottelo.host_info - DEBUG - Host Satellite version: 6.7
2020-02-20 14:16:56 - robottelo.helpers - DEBUG - Generated file bz_cache.json with BZ collect data
collected 1 item

test_reporttemplates.py::ReportTemplateTestCase::test_positive_generate_entitlements_report 

========================== 1 passed in 198.36 seconds ==========================

### DIFF
--- a/tests/foreman/api/test_reporttemplates.py
+++ b/tests/foreman/api/test_reporttemplates.py
@@ -15,6 +15,7 @@
 
 :Upstream: No
 """
+import pytest
 from fauxfactory import gen_string
 from nailgun import entities
 from requests import HTTPError
@@ -38,6 +39,52 @@ from robottelo.test import APITestCase
 from robottelo.vm import VirtualMachine
 
 
+@pytest.fixture(scope='class')
+def setup_content(request):
+    """"""
+    org = entities.Organization().create()
+    with manifests.clone() as manifest:
+        upload_manifest(org.id, manifest.content)
+    rh_repo_id = enable_rhrepo_and_fetchid(
+        basearch='x86_64',
+        org_id=org.id,
+        product=PRDS['rhel'],
+        repo=REPOS['rhst7']['name'],
+        reposet=REPOSET['rhst7'],
+        releasever=None,
+    )
+    rh_repo = entities.Repository(id=rh_repo_id).read()
+    rh_repo.sync()
+    custom_repo = entities.Repository(
+        product=entities.Product(organization=org).create(),
+    ).create()
+    custom_repo.sync()
+    lce = entities.LifecycleEnvironment(organization=org).create()
+    cv = entities.ContentView(
+        organization=org,
+        repository=[rh_repo_id, custom_repo.id],
+    ).create()
+    cv.publish()
+    cvv = cv.read().version[0].read()
+    promote(cvv, lce.id)
+    ak = entities.ActivationKey(
+        content_view=cv,
+        max_hosts=100,
+        organization=org,
+        environment=lce,
+        auto_attach=True
+    ).create()
+    subscription = entities.Subscription(organization=org).search(
+        query={'search': 'name="{}"'.format(DEFAULT_SUBSCRIPTION_NAME)})[0]
+    ak.add_subscriptions(data={
+        'quantity': 1,
+        'subscription_id': subscription.id,
+    })
+    request.cls.org_setup = org
+    request.cls.ak_setup = ak
+
+
+@pytest.mark.usefixtures("setup_content")
 class ReportTemplateTestCase(APITestCase):
     """Tests for ``katello/api/v2/report_templates``."""
 
@@ -156,7 +203,10 @@ class ReportTemplateTestCase(APITestCase):
         template = '<%= "value=\\"" %><%= input(\'{0}\') %><%= "\\"" %>'.format(input_name)
         entities.Host(name=host_name).create()
         rt = entities.ReportTemplate(name=template_name, template=template).create()
-        entities.TemplateInput(name=input_name, input_type="user", template=rt.id).create()
+        entities.TemplateInput(name=input_name,
+                               input_type="user",
+                               template=rt.id,
+                               ).create()
         ti = entities.TemplateInput(template=rt.id).search()[0].read()
         self.assertEquals(input_name, ti.name)
         res = rt.generate(data={"input_values": {input_name: input_value}})
@@ -200,11 +250,8 @@ class ReportTemplateTestCase(APITestCase):
         self.assertTrue(rt.locked)
         # 3. Clone template, check cloned data
         rt.clone(data={'name': template_clone_name})
-        cloned_rt = (
-            entities.ReportTemplate()
-            .search(query={'search': u'name="{}"'.format(template_clone_name)})[0]
-            .read()
-        )
+        cloned_rt = entities.ReportTemplate().search(
+                  query={'search': u'name="{}"'.format(template_clone_name)})[0].read()
         self.assertEquals(template_clone_name, cloned_rt.name)
         self.assertEquals(template1, cloned_rt.template)
         # 4. Try to delete template
@@ -212,14 +259,8 @@ class ReportTemplateTestCase(APITestCase):
             with self.assertRaises(HTTPError):
                 rt.delete()
             # In BZ1680458, exception is thrown but template is deleted anyway
-            self.assertNotEquals(
-                0,
-                len(
-                    entities.ReportTemplate().search(
-                        query={'search': u'name="{}"'.format(template_name)}
-                    )
-                ),
-            )
+            self.assertNotEquals(0, len(entities.ReportTemplate().search(
+                query={'search': u'name="{}"'.format(template_name)})))
         # 5. Try to edit template
         with self.assertRaises(HTTPError):
             entities.ReportTemplate(id=rt.id, template=template2).update(["template"])
@@ -235,14 +276,8 @@ class ReportTemplateTestCase(APITestCase):
         self.assertEquals(template2, rt.template)
         # 8. Delete template
         rt.delete()
-        self.assertEquals(
-            0,
-            len(
-                entities.ReportTemplate().search(
-                    query={'search': u'name="{}"'.format(template_name)}
-                )
-            ),
-        )
+        self.assertEquals(0, len(entities.ReportTemplate().search(
+            query={'search': u'name="{}"'.format(template_name)})))
 
     @tier2
     @stubbed()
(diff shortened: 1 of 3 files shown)


tests/foreman/api/test_reporttemplates.py
```diff
@@ -38,6 +39,52 @@
 from robottelo.vm import VirtualMachine
 
 
+@pytest.fixture(scope='class')
+def setup_content(request):
+    """"""
```
mirekdlugosz: Please write proper docstring, or just don't have any at all.

tests/foreman/api/test_reporttemplates.py
```diff
@@ -38,6 +39,52 @@
...
+        query={'search': 'name="{}"'.format(DEFAULT_SUBSCRIPTION_NAME)})[0]
+    ak.add_subscriptions(data={
+        'quantity': 1,
+        'subscription_id': subscription.id,
+    })
+    request.cls.org_setup = org
+    request.cls.ak_setup = ak
+
+
+@pytest.mark.usefixtures("setup_content")
```
mirekdlugosz: Right now if I run `test_positive_CRUDL`, it will first execute your fixture - even though this test doesn't use anything from fixture.

This matters because your test is tier3, so it will be invoked in separate `pytest` run by automation nad won't leverage fixture that run for tier1. Also, during failure analysis we might run any single test in isolation.

I think it's better to move that decorator to `test_positive_generate_entitlements_report` method. This way this test will be the only one that pulls in fixture - which is good, because it's only one that makes actual use of it.

If tests are gradually ported to make use of fixture, they will need to have decorator, too. At some point we will just remove individual methods decorators and slap that on top of class.

tests/foreman/cli/test_reporttemplates.py
```diff
@@ -857,3 +857,37 @@ def test_positive_generate_entitlements_report_multiple_formats(self):
...
+                          regarding Entitlements.
+
+        :CaseImportance: High
+        """
+        with VirtualMachine(distro=DISTRO_RHEL7) as vm:
+            vm.install_katello_ca()
+            vm.register_contenthost(self.setup_org['label'], self.setup_new_ak['name'])
+            assert vm.subscribed
+            scheduled_csv = ReportTemplate.schedule({
+                'id': 115,
```
mirekdlugosz: What is '115'? Will it always 115? Could we obtain that number dynamically? Would it make sense to add it to constants?

tests/foreman/cli/test_reporttemplates.py
```diff
@@ -857,3 +857,37 @@ def test_positive_generate_entitlements_report_multiple_formats(self):
...
+                'organization': self.setup_org['name'],
+                'report-format': 'csv'
+            })
+            data_csv = ReportTemplate.report_data({
+                'id': 115,
+                'job-id': scheduled_csv[0].split("Job ID: ", 1)[1]
+            })
+            assert vm.hostname in data_csv[1]
+            assert self.setup_subs_id[0]['name'] in data_csv[1]
```
mirekdlugosz: Usual comment about hardcoded numbers. Would it make sense to rewrite that to loop to make it more robust? We don't want test to fail just because order of items is slightly different.

tests/foreman/cli/test_reporttemplates.py
ColeHiggins2: 115 is the entitlements ID. each report template has a preset ID assigned to it. This should not change.

omkarkhatavkar: Can we add that into constants? Variable name help to understand.

tests/foreman/api/test_reporttemplates.py
```diff
@@ -38,6 +39,50 @@
 from robottelo.vm import VirtualMachine
 
 
+@pytest.fixture(scope='class')
+def setup_content(request):
```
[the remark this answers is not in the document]
ColeHiggins2: Since the two fixtures are set up slightly differently and have different return values, I think it seems redundant to add this in common code

tests/foreman/cli/test_reporttemplates.py
ColeHiggins2: Since each ReportTemplate has its own unique ID, we will still have to specify it anytime we call schedule or report_data. I can look into making it a constant but I don't think it is necessary for these tests. What do you think?

tests/foreman/cli/test_reporttemplates.py
```diff
@@ -857,3 +857,41 @@ def test_positive_generate_entitlements_report_multiple_formats(self):
...
+                'organization': self.setup_org['name'],
+                'report-format': 'csv'
+            })
+            data_csv = ReportTemplate.report_data({
+                'id': 115,
+                'job-id': scheduled_csv[0].split("Job ID: ", 1)[1]
+            })
+            for item in data_csv:
+                if vm.hostname == item:
+                    assert vm.hostname in data_csv
```
omkarkhatavkar: I have a query here if the` if vm.hostname == item` False over the loop we don't have the assertion in place ?
why not directly using `assert vm.hostname in data_csv ` maybe I'm missing something ?

tests/foreman/cli/test_reporttemplates.py
```diff
@@ -857,3 +857,41 @@ def test_positive_generate_entitlements_report_multiple_formats(self):
...
+            for item in data_csv:
+                if self.setup_subs_id[0]['name'] == item:
+                    assert self.setup_subs_id[0]['name'] in data_csv
```
omkarkhatavkar: same as https://github.com/SatelliteQE/robottelo/pull/7652/files#r383632971

tests/foreman/cli/test_reporttemplates.py
ColeHiggins2: report_data has a unique return value. It is a list with multiple strings, each containing multiple values. If I try to `assert vm.hostname in data_cav` it raises an assertion error. Hence me looping through instead.

mirekdlugosz: 1. If `assert vm.hostname in data_csv` failed outside of loop, there is no reason to believe it will succeed inside loop.
2. That assert will never be executed, because `item` never contains only `vm.hostname` - there are multiple other things in line.

What you might have meant (untested):
```python
for item in data_csv:
    if vm.hostname in item:
        assert vm.hostname in item
        break
else:  # this will be executed only if loop body did not encounter `break`
    # we need, because if vm.hostname is not in file anywhere, we don't run any assertion and we never fail
    pytest.fail("VM hostname {} not found in CSV".format(vm.hostname)
```

But it's much better written this way:
```python
assert any(vm.hostname in line for line in data_csv)
```

Alternatively, you can call `csv.reader()` on `data_csv` (`reader` supports list of strings as well as any file-like object) and check if `vm.hostname` is equal to specific field in any line.

tests/foreman/api/test_reporttemplates.py
```diff
@@ -38,6 +39,41 @@
...
```
rochacbruno: On https://github.com/SatelliteQE/robottelo/pull/7651/files it is called `content_setup` here it is `setup_content` any reason for different names?

rochacbruno: Also in future we might want to revisit this fixtures and put a single dynamic one in `tests/conftest.py` so we avoid repetition.


same_code

rochacbruno: Figures out that we already have a standard for reusable fixtures, examples here: https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/conftest.py#L8-L27

ColeHiggins2: I will make a note for future tests that require a full content setup.

PR #7661, opened 2020-02-25
Stubbed out iso library sync test case

latran:
Test case uses capsule iso 6.6 which we don't have.  We stopped creating capsule iso after 6.4 so I'm stubbing this test case out.

### DIFF
--- a/tests/foreman/api/test_contentmanagement.py
+++ b/tests/foreman/api/test_contentmanagement.py
@@ -50,6 +50,7 @@ from robottelo.constants import REPOSET
 from robottelo.constants import RPM_TO_UPLOAD
 from robottelo.decorators import run_in_one_thread
 from robottelo.decorators import skip_if_not_set
+from robottelo.decorators import stubbed
 from robottelo.decorators import tier2
 from robottelo.decorators import tier3
 from robottelo.decorators import tier4
@@ -562,6 +563,7 @@ class CapsuleContentManagementTestCase(APITestCase):
             get_repo_files(lce_repo_path, hostname=self.capsule_ip), get_repo_files(cvv_repo_path)
         )
 
+    @stubbed()
     @tier4
     def test_positive_iso_library_sync(self):
         """Ensure RH repo with ISOs after publishing to Library is synchronized


tests/foreman/api/test_contentmanagement.py
```diff
@@ -562,68 +562,68 @@ def test_positive_capsule_sync(self):
...
-        )
-        # Wait till capsule sync finishes
-        for task in sync_status['active_sync_tasks']:
-            entities.ForemanTask(id=task['id']).poll(timeout=600)
-        # Verify all the ISOs are present on capsule
-        capsule_isos = get_repo_files(repo_path, extension='iso', hostname=self.capsule_ip)
-        self.assertGreater(len(result), 0)
-        self.assertEqual(set(sat_isos), set(capsule_isos))
+    # @tier4
+    # def test_positive_iso_library_sync(self):
```
mirekdlugosz: Wouldn't it be better to just add `@stubbed` decorator?

As far as I know, `@stubbed` still end up in Polarion. I'm not versed enough in DDR/PQI to tell what exact impact that has, but I heard we generally don't want things removed from Polarion once they appear there.

JacobCallahan:
@stubbed is definitely preferred.

tests/foreman/api/test_contentmanagement.py
latran: using @stubbed as @JacobCallahan suggested.

omaciel:
@SatelliteQE/quality-engineers why not just completely delete it? It doesn't look like we're planning to automate this anymore as we don't generate ISOs anymore?

latran:
@omaciel, I think there was discussion on how it might affect polarion and it was better stubbed out than to delete.  Of course there's many stubbed out test cases in robottelo that doesn't have any have implementation that we need to address as well.
@JacobCallahan @mirekdlugosz, thoughts on this?

omaciel:
I would really delete it and mark the test case in Polarion as obsolete or whatever. We don't want to add to our technical debt.

latran:
There's no other tests that covers ISO content sync to a capsule so I'm leaving this test case stubbed out instead of deleting it.

PR #7666, opened 2020-02-28
Add test for assigning Http proxy to Repository and Products.

jameerpathan111:
Test Result:
```
$ pytest tests/foreman/ui/test_http_proxy.py -k test_positive_assign_http_proxy_to_products_repositories
2020-02-28 05:53:17 - conftest - DEBUG - Collected 2 test cases

collected 2 items / 1 deselected / 1 selected

tests/foreman/ui/test_http_proxy.py .                                    [100%]

============ 1 passed, 1 deselected, 17 warnings in 494.82 seconds =============
```

### DIFF
--- a/tests/foreman/ui/test_http_proxy.py
+++ b/tests/foreman/ui/test_http_proxy.py
@@ -17,9 +17,24 @@
 from fauxfactory import gen_integer
 from fauxfactory import gen_string
 from fauxfactory import gen_url
+from nailgun import entities
 
+from robottelo.constants import FAKE_0_PUPPET_REPO
+from robottelo.constants import FAKE_1_YUM_REPO
+from robottelo.constants import REPO_TYPE
+from robottelo.decorators import fixture
 from robottelo.decorators import tier2
 from robottelo.decorators import upgrade
+
+
+@fixture(scope='module')
+def module_org():
+    return entities.Organization().create()
+
+
+@fixture(scope='module')
+def module_loc():
+    return entities.Location().create()
 
 
 @tier2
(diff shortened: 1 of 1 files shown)


jameerpathan111:
Following PR's needs to be merged before merging this one.
Nailgun PR: [https://github.com/SatelliteQE/nailgun/pull/705](https://github.com/SatelliteQE/nailgun/pull/705)
Airgun PR: [https://github.com/SatelliteQE/airgun/pull/461](https://github.com/SatelliteQE/airgun/pull/461)

tests/foreman/ui/test_http_proxy.py
```diff
@@ -67,3 +82,117 @@ def test_positive_create_update_delete(session, module_org, module_loc):
...
+        organization=[module_org.id],
+        location=[module_loc.id],
+    ).create()
+    # Create products
+    product_a = entities.Product(organization=module_org.id,).create()
+    product_b = entities.Product(organization=module_org.id,).create()
+    # Create repositories from UI.
+    with session:
+        session.organization.select(org_name=module_org.name)
+        session.location.select(loc_name=module_loc.name)
```
mirekdlugosz: Do you need these two lines? What would happen if you don't use them?

I thought UI uses new user that is assigned to new organization and location by default, so there is no need for you to manually select it from menu after logging in.

tests/foreman/ui/test_http_proxy.py
```diff
@@ -67,3 +82,117 @@ def test_positive_create_update_delete(session, module_org, module_loc):
...
+                'repo_type': REPO_TYPE['yum'],
+                'repo_content.upstream_url': FAKE_1_YUM_REPO,
+                'repo_content.http_proxy_policy': 'Use specific HTTP Proxy',
+                'repo_content.proxy_policy.http_proxy': http_proxy_a.name,
+            },
+        )
+        repo_a2_values = session.repository.read(product_a.name, repo_a2_name)
+        assert repo_a2_values['repo_content'][
+            'http_proxy_policy'
+        ] == 'Use specific HTTP Proxy ({})'.format(http_proxy_a.name)
```
mirekdlugosz: This is not wrong, but would be nicer if written:
```python
        expected_policy = 'Use specific HTTP Proxy ({})'.format(http_proxy_a.name)
        assert repo_a2_values['repo_content']['http_proxy_policy'] == expected_policy
```

tests/foreman/ui/test_http_proxy.py
```diff
@@ -67,3 +82,117 @@ def test_positive_create_update_delete(session, module_org, module_loc):
...
+            product_b.name,
+            {
+                'name': repo_b2_name,
+                'repo_type': REPO_TYPE['puppet'],
+                'repo_content.upstream_url': FAKE_0_PUPPET_REPO,
+                'repo_content.http_proxy_policy': 'No HTTP Proxy',
+            },
+        )
+        # Add http_proxy to products
+        session.product.search('')
```
mirekdlugosz: You shouldn't need to do that - it's `session.product.manage_http_proxy` to figure out that elements to select are not displayed, and reset any filter there might be to make them visible.

I don't want to block your PR on that, so can you open Airgun issue?

tests/foreman/ui/test_http_proxy.py
```diff
@@ -67,3 +82,117 @@ def test_positive_create_update_delete(session, module_org, module_loc):
...
+            {
+                'http_proxy_policy': 'Use specific HTTP Proxy',
+                'proxy_policy.http_proxy': http_proxy_b.name,
+            },
+        )
+        # Verify that Http Proxy is updated for all repos of product_a and product_b.
+        repo_a1_values = session.repository.read(product_a.name, repo_a1_name)
+        assert repo_a1_values['repo_content'][
+            'http_proxy_policy'
+        ] == 'Use specific HTTP Proxy ({})'.format(http_proxy_b.name)
```
mirekdlugosz: There is a lot of string duplication here. Can you get away with following?
```python
        proxy_policy = 'Use specific HTTP Proxy ({})'
        assert repo_a1_values['repo_content']['http_proxy_policy'] == proxy_policy.format(http_proxy_b.name)
        repo_a2_values = session.repository.read(product_a.name, repo_a2_name)
        assert repo_a2_values['repo_content']['http_proxy_policy'] == proxy_policy.format(http_proxy_b.name)
        ...
```

tests/foreman/ui/test_http_proxy.py
jameerpathan111: Created issue https://github.com/SatelliteQE/airgun/issues/465

tests/foreman/ui/test_http_proxy.py
jameerpathan111: yeah, it doesn't look good. Updated PR.

tests/foreman/ui/test_http_proxy.py
jameerpathan111: yes, looks way better.

tests/foreman/ui/test_http_proxy.py
```diff
@@ -17,11 +17,26 @@
...
+from robottelo.constants import FAKE_0_PUPPET_REPO
+from robottelo.constants import FAKE_1_YUM_REPO
+from robottelo.constants import REPO_TYPE
+from robottelo.decorators import fixture
 from robottelo.decorators import tier2
 from robottelo.decorators import upgrade
 
 
+@fixture(scope='module')
+def module_org():
```
jyejare: This is already defined in robottelo/api/entity_fixtures.py and imported in conftest.py , use it from there instead of defining here.

tests/foreman/ui/test_http_proxy.py
```diff
@@ -17,11 +17,26 @@
...
+    return entities.Organization().create()
+
+
+@fixture(scope='module')
```
jyejare: same here

jameerpathan111:
> Also, I see that some of the base entities like repository are being created using UI. Port them to API instead to fasten the process.

@jyejare the test is for assigning the http proxy to repos and products from UI, so we shouldn't port them to API.

tests/foreman/ui/test_http_proxy.py
jameerpathan111: @jyejare If I use the approach you suggested then, I will have to add code for manually selecting the required org/loc when UI part of test starts executing otherwise the test will fail because it won't be able to find http proxy for that org/loc. But If I keep the current approach then I won't have to do any of that. Also most of the UI tests seems to be following the same approach.

jameerpathan111: This is what I have observed so far, I will look further into it and if I find any other way then I will update the code accordingly. :)

jyejare: Probabbly one of your entity(may be http proxy) automatically creating different orgs/locs and hence the entity is not discoverable. You can take a look at this later if its taking more time to troubleshoot.

Running the tests ASAP that code quality is more important for now. You please revisit to fix this later.

jyejare:
> > Also, I see that some of the base entities like repository are being created using UI. Port them to API instead to fasten the process.
> 
> @jyejare the test is for assigning the http proxy to repos and products from UI, so we shouldn't port them to API.

Just assigning part should be done via UI then and not the repo/product creation.

Take your time and improve it later. For now, merging this as its 6.7 feature automation PR.

jameerpathan111:
> Just assigning part should be done via UI then and not the repo/product creation.

@jyejare I am assigning Http Proxy during creation of repository itself and that's what this test intends to do, so I can't port them to API. We want to check whether Http Proxy field is working while creation of repo from UI. That's why I have only created products from API and not repositories. :)

tests/foreman/ui/test_http_proxy.py
jameerpathan111: @jyejare yes probably, I will see why it's not working and fix it in another PR.

jyejare:
> > Just assigning part should be done via UI then and not the repo/product creation.
> 
> @jyejare I am assigning Http Proxy during creation of repository itself and that's what this test intends to do, so I can't port them to API. We want to check whether Http Proxy field is working while creation of repo from UI. That's why I have only created products from API and not repositories. :)

Gotch you ! Thanks for the clarification !

PR #7676, opened 2020-03-04
AzureRm CLI automation - part1

vijay8451:
- Added tests around AzureRm CLI 
- Test results:
```
tests/foreman/cli/test_computeresource_azurerm.py::TestAzureRMComputeResourceTestCase::test_positive_crud_azurerm_cr PASSED                                                            [ 14%]
tests/foreman/cli/test_computeresource_azurerm.py::TestAzureRMComputeResourceTestCase::test_positive_crud_finish_template_image PASSED                                                 [ 28%]
tests/foreman/cli/test_computeresource_azurerm.py::TestAzureRMComputeResourceTestCase::test_positive_crud_cloud_init_image PASSED                                                      [ 42%]
tests/foreman/cli/test_computeresource_azurerm.py::TestAzureRMComputeResourceTestCase::test_positive_check_available_networks PASSED                                                   [ 57%]
tests/foreman/cli/test_computeresource_azurerm.py::TestAzureRMComputeResourceTestCase::test_positive_create_compute_profile_values PASSED                                              [ 71%]
tests/foreman/cli/test_computeresource_azurerm.py::TestAzureRm_FinishTemplate_Provisioning::test_positive_azurerm_host_provisioned PASSED                                              [ 85%]
tests/foreman/cli/test_computeresource_azurerm.py::TestAzureRm_UserData_Provisioning::test_positive_azurerm_host_provisioned PASSED                                                    [100%]

================================================================================ 7 passed in 1475.09 seconds =================================================================================
```

### DIFF
--- a/robottelo/api/utils.py
+++ b/robottelo/api/utils.py
@@ -948,12 +948,12 @@ def skip_yum_update_during_provisioning(template=None, reverse=False):
         return update_provisioning_template(name=template, old=new, new=old)
 
 
-def hammer_api_timeout_set(timeout=-1, reverse=False):
+def set_hammer_api_timeout(timeout=-1, reverse=False):
     """Set hammer API request timeout on Satellite
 
     :param int timeout: request timeout in seconds
     :param bool reverse: Reverses the request timeout
-    :return: Boolean True on success else exception
+    :return: ssh.command
     """
     default_timeout = ':request_timeout: {}'.format(120)
     new_timeout = ':request_timeout: {}'.format(timeout)
--- a/robottelo/api/utils.py
+++ b/robottelo/api/utils.py
@@ -946,3 +946,26 @@ def skip_yum_update_during_provisioning(template=None, reverse=False):
         return update_provisioning_template(name=template, old=old, new=new)
     else:
         return update_provisioning_template(name=template, old=new, new=old)
+
+
+def hammer_api_timeout_set(timeout=-1, reverse=False):
+    """Set hammer API request timeout on Satellite
+
+    :param int timeout: request timeout in seconds
+    :param bool reverse: Reverses the request timeout
+    :return: Boolean True on success else exception
+    """
+    default_timeout = ':request_timeout: {}'.format(120)
+    new_timeout = ':request_timeout: {}'.format(timeout)
+    if not reverse:
+        return ssh.command(
+            "sed -ie 's/{}/{}/' ~/.hammer/cli.modules.d/foreman.yml".format(
+                default_timeout, new_timeout
+            )
+        )
+    else:
+        return ssh.command(
+            "sed -ie 's/{}/{}/' ~/.hammer/cli.modules.d/foreman.yml".format(
+                new_timeout, default_timeout
+            )
+        )
--- a/robottelo/cli/computeresource.py
+++ b/robottelo/cli/computeresource.py
@@ -54,7 +54,7 @@ class ComputeResource(Base):
 
     @classmethod
     def image_list(cls, options):
-        """list an image"""
+        """Show the list of images"""
         cls.command_sub = 'image list'
         return cls.execute(cls._construct_command(options), output_format='csv')
 
--- a/robottelo/cli/computeresource.py
+++ b/robottelo/cli/computeresource.py
@@ -27,3 +27,45 @@ class ComputeResource(Base):
     """
 
     command_base = 'compute-resource'
+
+    @classmethod
+    def image_create(cls, options):
+        """Create an image"""
+        cls.command_sub = 'image create'
+        return cls.execute(cls._construct_command(options), output_format='csv')
+
+    @classmethod
+    def image_info(cls, options):
+        """Show an image"""
+        cls.command_sub = 'image info'
+        return cls.execute(cls._construct_command(options), output_format='csv')
+
+    @classmethod
+    def image_available(cls, options):
+        """Show images available for addition"""
+        cls.command_sub = 'image available'
+        return cls.execute(cls._construct_command(options), output_format='csv')
+
+    @classmethod
+    def image_delete(cls, options):
+        """delete an image"""
+        cls.command_sub = 'image delete'
+        return cls.execute(cls._construct_command(options), output_format='csv')
+
+    @classmethod
+    def image_list(cls, options):
+        """list an image"""
+        cls.command_sub = 'image list'
+        return cls.execute(cls._construct_command(options), output_format='csv')
+
+    @classmethod
+    def image_update(cls, options):
+        """update an image"""
+        cls.command_sub = 'image update'
+        return cls.execute(cls._construct_command(options), output_format='csv')
+
+    @classmethod
+    def networks(cls, options):
+        """List available networks for a compute resource"""
+        cls.command_sub = 'networks'
+        return cls.execute(cls._construct_command(options), output_format='csv')
--- a/tests/foreman/cli/test_computeresource_azurerm.py
+++ b/tests/foreman/cli/test_computeresource_azurerm.py
@@ -31,6 +31,7 @@ from robottelo.constants import AZURERM_RG_DEFAULT
 from robottelo.constants import AZURERM_RHEL7_FT_IMG_URN
 from robottelo.constants import AZURERM_RHEL7_UD_IMG_URN
 from robottelo.constants import AZURERM_VM_SIZE_DEFAULT
+from robottelo.decorators import parametrize
 from robottelo.decorators import run_in_one_thread
 from robottelo.decorators import tier1
 from robottelo.decorators import tier2
@@ -67,24 +68,7 @@ def azurerm_hostgroup(
 
 
 class TestAzureRMComputeResourceTestCase:
-    """
-    Usage::
-
-    hammer compute_resource [OPTIONS] SUBCOMMAND [ARG] ...
-
-Parameters::
-
-    SUBCOMMAND                    subcommand
-    [ARG] ...                     subcommand arguments
-
-Subcommands::
-
-    create                        Create a compute resource.
-    info                          Show an compute resource.
-    list                          List all compute resources.
-    update                        Update a compute resource.
-    delete                        Delete a compute resource.
-    image                         View and manage compute resource's images"""
+    """ AzureRm compute resource Tests"""
 
     @upgrade
     @tier1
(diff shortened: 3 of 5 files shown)


tests/foreman/cli/test_computeresource_azurerm.py
```diff
@@ -0,0 +1,629 @@
...
+        module_org,
+        module_os,
+        module_smart_proxy,
+        module_puppet_environment,
+    ):
+        """
+        Provisions the host on AzureRM using Finish template
+        Later in tests this host will be used to perform assertions
+        """
+        hammer_api_timeout_set()
```
vijay8451: Setting notimeout `-1` to avoid hammer error `Timed out reading data from server` as host create/delete takes time (depends on Azure Cloud response)

tests/foreman/cli/test_computeresource_azurerm.py
```diff
@@ -0,0 +1,629 @@
...
+                'interface': self.interfaces_attributes,
+                'location-id': module_location.id,
+                'organization-id': module_org.id,
+                'domain-id': module_domain.id,
+                'architecture-id': module_architecture.id,
+                'operatingsystem-id': module_os.id,
+                'root-password': gen_string('alpha'),
+                'image': module_azurerm_finishimg.name,
+            },
+            timeout=1800,
```
vijay8451: set timeout to avoid error `command did not respond in the predefined time` as host create/delete takes time (depends on Azure Cloud response)

tests/foreman/cli/test_computeresource_azurerm.py
```diff
@@ -0,0 +1,629 @@
...
+            {'name': new_img_name, 'compute-resource': module_azurerm_cr.name}
+        )[0]
+        assert img_info['username'] == new_username
+
+        # Delete Image
+        result = ComputeResource.image_delete(
+            {'name': new_img_name, 'compute-resource': module_azurerm_cr.name}
+        )[0]
+        assert result['message'] == 'Image deleted.'
+        assert result['name'] == new_img_name
```
pondrejk: Maybe I'm overlooking something but if the only difference between test_positive_crud_cloud_init_image and test_positive_crud_finish_template_image is in the constant passed to uuid (AZURERM_RHEL7_UD_IMG_URN vs AZURERM_RHEL7_FT_IMG_URN), then we could make it into one parametrized test for easier maintenance

tests/foreman/cli/test_computeresource_azurerm.py
```diff
@@ -0,0 +1,629 @@
...
+        assert class_host_ud['name'] == self.fullhostname
+        assert class_host_ud['status']['build-status'] == "Pending installation"
+        assert class_host_ud['network-interfaces'][0]['ipv4-address'] == azureclient_host.ip
+        assert class_host_ud['compute-resource'] == module_azurerm_cr.name
+        assert class_host_ud['operating-system']['image'] == module_azurerm_cloudimg.name
+        assert class_host_ud['host-group'] == azurerm_hostgroup.name
+
+        # Azure cloud
+        assert self.hostname.lower() == azureclient_host.name
+        assert self.vm_size == azureclient_host.type
```
pondrejk: Is it really needed to have two separate classes for provisioning? It's a bit hard to wrap my head around this but it seems there's just a difference in the image (which is not even asserted later).

tests/foreman/cli/test_computeresource_azurerm.py
```diff
@@ -0,0 +1,629 @@
...
+
+        :id: 9e08463c-c700-47fc-8a58-e03aa8bcd097
+
+        :expectedresults: All the networks from AzureRM CR should be available.
+
+        :CaseLevel: Integration
+        """
+
+        result = ComputeResource.networks({'id': module_azurerm_cr.id})
+        assert len(result) > 0
```
pondrejk: I wonder if this shouldn't be rather a prerequisite step in provisioning tests rather than a separate test case. If those networks are not available the provisioning will fail, or?

tests/foreman/cli/test_computeresource_azurerm.py
```diff
@@ -0,0 +1,629 @@
...
+
+Subcommands::
+
+    create                        Create a compute resource.
+    info                          Show an compute resource.
+    list                          List all compute resources.
+    update                        Update a compute resource.
+    delete                        Delete a compute resource.
+    image                         View and manage compute resource's images"""
+
```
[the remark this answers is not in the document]
vijay8451: okay

tests/foreman/cli/test_computeresource_azurerm.py
vijay8451: Good idea , let me give try in that way too.

tests/foreman/cli/test_computeresource_azurerm.py
vijay8451: no this should not be prerequisite , because this test is checking list of networks however if this fails then it does not impacting host provisioning as both are two different things ..

But for user experience for easy read it's good to see the list of n/ws and could use during provisioning ..

tests/foreman/cli/test_computeresource_azurerm.py
vijay8451: yes it's needed as this PR is part-1 to add support around UD and FT provisioning ... so there would be few more tests where they either requires put under UD or FT class ...

hence for easy of separation (more utilization) this requires ..

tests/foreman/cli/test_computeresource_azurerm.py
vijay8451: Done:
```
tests/foreman/cli/test_computeresource_azurerm.py::TestAzureRMComputeResourceTestCase::test_positive_image_crud[RedHat:RHEL:7-RAW:latest] PASSED                                       [ 50%]
tests/foreman/cli/test_computeresource_azurerm.py::TestAzureRMComputeResourceTestCase::test_positive_image_crud[RedHat:RHEL:7-RAW-CI:7.6.2019072418] PASSED                            [100%]

```

vijay8451:
@pondrejk revisit please .thanks

robottelo/api/utils.py
```diff
@@ -946,3 +946,26 @@ def skip_yum_update_during_provisioning(template=None, reverse=False):
         return update_provisioning_template(name=template, old=old, new=new)
     else:
         return update_provisioning_template(name=template, old=new, new=old)
+
+
+def hammer_api_timeout_set(timeout=-1, reverse=False):
```
omkarkhatavkar: any reason keeping the timeout as -1

robottelo/api/utils.py
```diff
@@ -946,3 +946,26 @@ def skip_yum_update_during_provisioning(template=None, reverse=False):
...
+    """Set hammer API request timeout on Satellite
+
+    :param int timeout: request timeout in seconds
+    :param bool reverse: Reverses the request timeout
+    :return: Boolean True on success else exception
```
omkarkhatavkar: ssh.command returns Boolean, I doubt here.

robottelo/api/utils.py
omkarkhatavkar: I would have kept the function name as 'set_hammer_api_timeout'.

robottelo/cli/computeresource.py
```diff
@@ -27,3 +27,45 @@ class ComputeResource(Base):
...
+
+    @classmethod
+    def image_delete(cls, options):
+        """delete an image"""
+        cls.command_sub = 'image delete'
+        return cls.execute(cls._construct_command(options), output_format='csv')
+
+    @classmethod
+    def image_list(cls, options):
+        """list an image"""
```
omkarkhatavkar: show the list of images

robottelo/api/utils.py
vijay8451: -1 is for no timeout[1]  and could help in case not clear on exact time-out lengths 

[1] https://access.redhat.com/solutions/1274723

robottelo/api/utils.py
vijay8451: that what i understood from is's behaviour and we are following same here too: 

https://github.com/SatelliteQE/robottelo/blob/445eea73d227f4f3f68de58eeab2cec43918463d/robottelo/api/utils.py#L936-L941

robottelo/api/utils.py
vijay8451: done

robottelo/cli/computeresource.py
vijay8451: done

vijay8451:
@omkarkhatavkar revisit please,  thanks

tests/foreman/cli/test_computeresource_azurerm.py
```diff
@@ -0,0 +1,530 @@
...
+        :expectedresults: All the networks from AzureRM CR should be available.
+
+        :CaseLevel: Integration
+        """
+
+        result = ComputeResource.networks({'id': module_azurerm_cr.id})
+        assert len(result) > 0
+
+    @tier2
+    def test_positive_create_compute_profile_values(self, azurermclient, module_azurerm_cr):
```
jyejare:
```suggestion
    def test_positive_create_compute_profile(self, azurermclient, module_azurerm_cr):
```

robottelo/api/utils.py
omkarkhatavkar: That can be incorrect, take a look at  https://github.com/omkarkhatavkar/robottelo/blob/a55948b7a4575a72506c845e81e2d3049dbf5056/robottelo/ssh.py#L449 file we are ssh.command does not return Boolean

vijay8451: good catch , could just add the 'ssh.command' in return

tests/foreman/cli/test_computeresource_azurerm.py
vijay8451: tests is more related to `values` create instead profile create hence used same naming ..

robottelo/api/utils.py
vijay8451: done

vijay8451:
@jyejare please merge if looks good to you :)

tests/foreman/cli/test_computeresource_azurerm.py
jyejare: We always test values of any component as a componenet is made of values only. :)

But not a blocker comment to hold this PR. ACKing and Merging.

PR #7677, opened 2020-03-04
Remove deprecated ui tests

JacobCallahan:
These are all commented out, so let's just remove them.


JacobCallahan:
@mirekdlugosz my thought as well. with them still not being automated, either that functionality is now covered in other tests, or they weren't valuable enough to port.

mirekdlugosz:
There's dreaded third possibility: nobody knew or had enough time to dig into them, to decide if there is value to begin with. That's what my initial concern revolved around.

But if that wasn't widely known or was pushed "for tomorrow" for so long, then perhaps that's the sign there is nothing valuable inside.

JacobCallahan:
@mirekdlugosz definitely a possibility. component owners can always go back into the 6.6.z branch if they want to review what used to be here.

PR #7679, opened 2020-03-05
add tier2 test for cockpit webconsole integration

tstrych:
Changing PR: #7554
Moving test more into airgun.
Airgun PR [466](https://github.com/SatelliteQE/airgun/pull/466)

test results:
```
test_host.py::test_positive_cockpit 
=================== 1 passed, 29 deselected in 95.11 seconds ===================
```

### DIFF
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -1908,3 +1908,25 @@ def test_positive_gce_cloudinit_provision_end_to_end(
             skip_yum_update_during_provisioning(
                 template='Kickstart default user data', reverse=True
             )
+
+
+@tier2
+def test_positive_cockpit(session):
+    """Test whether webconsole button and cockpit integration works
+
+    :id: 5a9be063-cdc4-43ce-91b9-7608fbebf8bb
+
+    :expectedresults: Cockpit page is loaded and displays sat host info
+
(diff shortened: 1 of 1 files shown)


jameerpathan111:
@pondrejk we should close PR #7554, now that we have this one.

tests/foreman/ui/test_host.py
```diff
@@ -1908,3 +1908,27 @@ def test_positive_gce_cloudinit_provision_end_to_end(
...
+    :id: 5a9be063-cdc4-43ce-91b9-7608fbebf8bb
+
+    :expectedresults: Cockpit page is loaded and displays sat host info
+
+    :CaseLevel: System
+
+    """
+    with session:
+        session.organization.select(org_name='Default Organization')
+        session.location.select(loc_name='Any Location')
```
JacobCallahan: does this only work with the default org and any location selected?
if not, was there a specific need to select them here?

mirekdlugosz: This needs host, and uses Satellite to save time. Satellite is assigned to default organization and default location and won't be displayed in table when any other combination is selected.

JacobCallahan: interesting. thanks for the explanation

PR #7693, opened 2020-03-17
slight change in rename error message

pondrejk:
```
pytest tests/foreman/sys/test_rename.py -k est_negative_rename_sat_to_invalid_hostname
============================================================= test session starts =============================================================
                                                                               

tests/foreman/sys/test_rename.py .                                                                                                      [100%]

=================================================== 1 passed, 4 deselected in 19.42 seconds ===================================================
```

### DIFF
--- a/tests/foreman/sys/test_rename.py
+++ b/tests/foreman/sys/test_rename.py
@@ -30,8 +30,9 @@ from robottelo.ssh import get_connection
 from robottelo.test import TestCase
 
 BCK_MSG = "**** Hostname change complete! ****"
-BAD_HN_MSG = "{0} is not a valid fully qualified domain name, please \
-use a valid FQDN and try again."
+BAD_HN_MSG = (
+    "{0} is not a valid fully qualified domain name. Please use a valid FQDN and try again."
+)
 NO_CREDS_MSG = "Username and/or Password options are missing!"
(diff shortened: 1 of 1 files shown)


tests/foreman/sys/test_rename.py
```diff
@@ -30,7 +30,7 @@
 from robottelo.test import TestCase
 
 BCK_MSG = "**** Hostname change complete! ****"
-BAD_HN_MSG = "{0} is not a valid fully qualified domain name, please \
+BAD_HN_MSG = "{0} is not a valid fully qualified domain name. Please \
 use a valid FQDN and try again."
```
JacobCallahan: non-blocking, but do you think this looks better?
```python
BAD_HN_MSG = "{0} is not a valid fully qualified domain name. " \
             "Please use a valid FQDN and try again."
```

pondrejk: I do think it looks better, but black than "fixes" it to this:
```
 "{0} is not a valid fully qualified domain name. " "Please use a valid FQDN and try again."
```
`¯\_(ツ)_/¯`
Anyway, went with the one liner without the extra quotes

JacobCallahan: hahaha nice. i would think that would put it over to 99 character limit. so perhaps a bug on black's end.

PR #7698, opened 2020-03-19
added test to verify the external logout in integration with RHSSO

omkarkhatavkar:
### PR Description 
Added Scenario for logout using RHSSO integration  

### Test Result 
```

2020-03-19 15:34:31 - robottelo.ssh - DEBUG - Destroyed Paramiko client 0x7f10af4f02e8
2020-03-19 15:34:31 - robottelo.host_info - DEBUG - Host Satellite version: 6.7
2020-03-19 15:34:31 - robottelo.helpers - DEBUG - Generated file bz_cache.json with BZ collect data
collected 1 item

========================== 1 passed in 108.46 seconds ==========================
```

### Airgun PR
https://github.com/SatelliteQE/airgun/pull/468

### DIFF
--- a/tests/foreman/ui/test_ldap_authentication.py
+++ b/tests/foreman/ui/test_ldap_authentication.py
@@ -1228,3 +1228,35 @@ def test_single_sign_on_using_rhsso(enable_external_auth_rhsso, session):
 
     finally:
         update_rhsso_settings_in_satellite(revert=True)
+
+
+@destructive
+@tier4
+def test_ext_logout_rhsso(session):
+    """Verify the ext logout page navigation with external authentication RH-SSO
+
+    :id: 87b5e08e-69c6-11ea-8126-e74d80ea4308
+
+    :setup: Enroll the RH-SSO Configuration for External Authentication
+
+    :steps:
+        1. Create Mappers on RHSSO Instance and Update the Settings in Satellite
+        2. Login into Satellite using RHSSO login page redirected by Satellite
+        3. Logout from Satellite and Verify the ext_logout page displayed
+
+    :expectedresults: After logout from Satellite navigate should be ext_loout page
+    """
+    try:
+        update_rhsso_settings_in_satellite()
+        with session(login=False):
+            login_details = {
+                'username': settings.rhsso.rhsso_user,
+                'password': settings.rhsso.password,
+            }
+            session.rhsso_login.login(login_details)
+            view = session.rhsso_login.logout()
+            assert view['login_again'] == "Click to log in again"
+            session.rhsso_login.login(login_details, external_login=True)
+            assert session.task.read_all()['current_user'] == settings.rhsso.rhsso_user
+    finally:
+        update_rhsso_settings_in_satellite(revert=True)
--- a/tests/foreman/ui/test_ldap_authentication.py
+++ b/tests/foreman/ui/test_ldap_authentication.py
@@ -45,7 +45,6 @@ from robottelo.decorators import run_in_one_thread
 from robottelo.decorators import setting_is_set
 from robottelo.decorators import skip_if_not_set
 from robottelo.decorators import tier2
-from robottelo.decorators import tier4
 from robottelo.decorators import upgrade
 from robottelo.helpers import file_downloader
 
@@ -1158,7 +1157,6 @@ def test_negative_login_user_with_invalid_password_otp(test_name, ipa_data, auth
 
 @destructive
 @ldap_wrapper
-@tier4
 def test_single_sign_on_ldap_ipa_server(enroll_idm_and_configure_external_auth):
     """Verify the single sign-on functionality with external authentication
 
@@ -1201,7 +1199,6 @@ def test_single_sign_on_ldap_ipa_server(enroll_idm_and_configure_external_auth):
 
 
 @destructive
-@tier4
 def test_single_sign_on_using_rhsso(enable_external_auth_rhsso, session):
     """Verify the single sign-on functionality with external authentication RH-SSO
 
@@ -1231,7 +1228,6 @@ def test_single_sign_on_using_rhsso(enable_external_auth_rhsso, session):
 
 
 @destructive
-@tier4
 def test_ext_logout_rhsso(session):
     """Verify the ext logout page navigation with external authentication RH-SSO
 
--- a/tests/foreman/ui/test_ldap_authentication.py
+++ b/tests/foreman/ui/test_ldap_authentication.py
@@ -294,6 +294,10 @@ def enroll_configure_rhsso_external_auth():
             settings.rhsso.password, settings.rhsso.host_url, settings.rhsso.realm,
         )
     )
+    run_command(cmd="satellite-installer --foreman-keycloak true "
+                    "--foreman-keycloak-app-name 'foreman-openidc' "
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1228,3 +1225,34 @@ def test_single_sign_on_using_rhsso(enable_external_auth_rhsso, session):
...
+    :setup: Enroll the RH-SSO Configuration for External Authentication
+
+    :steps:
+        1. Create Mappers on RHSSO Instance and Update the Settings in Satellite
+        2. Login into Satellite using RHSSO login page redirected by Satellite
+        3. Logout from Satellite and Verify the ext_logout page displayed
+
+    :expectedresults: After logout from Satellite navigate should be ext_loout page
+    """
+    try:
```
JacobCallahan: what is the most volatile part of this code? i'd like us to wrap as little as possible under a ```try```

omkarkhatavkar: @JacobCallahan can you please explain in detail here. Not sure what do you mean here.

omkarkhatavkar: if you mean why I'm doing try-catch here, This is because we are changing the settings, if we don't use that can cause a radical issue on the server. hence I'm reverting the settings. Please let me know if you are looking for.

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1228,3 +1225,34 @@ def test_single_sign_on_using_rhsso(enable_external_auth_rhsso, session):
 
     finally:
         update_rhsso_settings_in_satellite(revert=True)
+
+
+@destructive
+def test_ext_logout_rhsso(session):
+    """Verify the ext logout page navigation with external authentication RH-SSO
```
mirekdlugosz: I think we may use "external" in docstring, so everyone knows what it is about.

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1228,3 +1225,34 @@ def test_single_sign_on_using_rhsso(enable_external_auth_rhsso, session):
...
+    :id: 87b5e08e-69c6-11ea-8126-e74d80ea4308
+
+    :setup: Enroll the RH-SSO Configuration for External Authentication
+
+    :steps:
+        1. Create Mappers on RHSSO Instance and Update the Settings in Satellite
+        2. Login into Satellite using RHSSO login page redirected by Satellite
+        3. Logout from Satellite and Verify the ext_logout page displayed
+
+    :expectedresults: After logout from Satellite navigate should be ext_loout page
```
mirekdlugosz: Typo:
```suggestion
    :expectedresults: After logout from Satellite navigate should be ext_logout page
```

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1228,3 +1225,34 @@ def test_single_sign_on_using_rhsso(enable_external_auth_rhsso, session):
...
+        with session(login=False):
+            login_details = {
+                'username': settings.rhsso.rhsso_user,
+                'password': settings.rhsso.password,
+            }
+            session.rhsso_login.login(login_details)
+            view = session.rhsso_login.logout()
+            assert view['login_again'] == "Click to log in again"
+            session.rhsso_login.login(login_details, external_login=True)
+            assert session.task.read_all()['current_user'] == settings.rhsso.rhsso_user
```
mirekdlugosz: I think I've seen this pattern before, but it seems wasteful:

* You are not on Tasks page after logging in, are you? So you navigate to Tasks only to read data that you also have on current page.
* There's no reason to read everything just to run assertion on content of single widget. You should be able to use `session.task.read_all(widget_names="current_user")`.

omkarkhatavkar: okay make sense

tests/foreman/ui/test_ldap_authentication.py
JacobCallahan: @omkarkhatavkar i'm basically trying to find out what part of this is most likely to fail. then wrap only that line (or two) in the try

omkarkhatavkar: I can write the decorator wrapper like ldap_wrapper here is e.g. https://github.com/omkarkhatavkar/robottelo/blob/b164fe01e51d6b7b2015762a9eacbc21189ac651/tests/foreman/ui/test_ldap_authentication.py#L356 if you think it is a better idea?

JacobCallahan: so that decorator would wrap the entire test in a try/finally :)
basically, if the most volatile lines of this test are ```session.rhsso_login.login(login_details)``` those are the only ones that should be wrapped in the try/finally

omkarkhatavkar: @JacobCallahan `session.rhsso_login.login(login_details, external_login=True)` line also need that try catch, if we add only on `session.rhsso_login.login(login_details)`, mentioned line will fail.

mirekdlugosz: @JacobCallahan the idea here is that we want to run cleanup no matter what happens. This is to restore global configurations to known working state, clean up external objects etc. In pytest you could model that with fixtures and finalizers. I don't know if there is better way to do that in unittest, but we kind of settled on `try/finally`, with large chunks of code inside `try`. Just grep robottelo tests for `finally:` for evidence.

It seems to me that you two kind of talk past each other, and as far as I can tell, root cause is that you have two different use-cases in mind.

omkarkhatavkar: I'm open here to know the solution or hint about how I can implement. Is that something you guys can let me then it works for me.

omkarkhatavkar:
```
Testing started at 5:06 PM ...
/home/okhatavk/okhatavk/Satellite/robottelo/env/bin/python /home/okhatavk/Downloads/pycharm-community-2018.1.4/helpers/pycharm/_jb_pytest_runner.py --target test_ldap_authentication.py::test_external_logout_rhsso
Launching py.test with arguments test_ldap_authentication.py::test_external_logout_rhsso in /home/okhatavk/Satellite/robottelo/tests/foreman/ui                                           [100%]

========================== 1 passed in 141.56 seconds ==========================

Testing started at 4:49 PM ...
/home/okhatavk/okhatavk/Satellite/robottelo/env/bin/python /home/okhatavk/Downloads/pycharm-community-2018.1.4/helpers/pycharm/_jb_pytest_runner.py --target test_ldap_authentication.py::test_single_sign_on_using_rhsso
Launching py.test with arguments test_ldap_authentication.py::test_single_sign_on_using_rhsso in                     [100%]

========================== 1 passed in 696.66 seconds ==========================
```

tests/foreman/ui/test_ldap_authentication.py
omkarkhatavkar: @JacobCallahan Do you have more concerns, I need this need to be merge before next snap. Please let me know

JacobCallahan: @mirekdlugosz we'll have to keep this in mind during the great unittest purge. While not ideal, I won't block this time.

omkarkhatavkar:
Thanks @JacobCallahan

PR #7706, opened 2020-03-26
added rhsso scenario, fix flaky tests and refactored some tests

omkarkhatavkar:
### PR Details 

- an added scenario for new RHSSO user and verify the login details. 
- Fixed flaky test-related to https AD instance
- Refactored rhsso helper methods 

### Test Result
```
Launching py.test with arguments test_ldap_authentication.py::test_positive_create_with_ad_https in /home/okhatavk/Satellite/robottelo/tests/foreman/ui                     [100%]

========================== 1 passed in 275.29 seconds ==========================

========================== 1 passed in 128.38 seconds ==========================
```

### DIFF
--- a/robottelo/constants.py
+++ b/robottelo/constants.py
@@ -1778,3 +1778,14 @@ AUDIENCE_MAPPER = {
     "protocol": "openid-connect",
     "protocolMapper": "oidc-audience-mapper",
 }
+
+RHSSO_NEW_USER = {
+    "email": "test_user@mailinator.com",
+    "emailVerified": "true",
+    "enabled": "true",
+    "firstName": "first_name",
+    "lastName": "last_name",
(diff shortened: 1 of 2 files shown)


robottelo/constants.py
```diff
@@ -1778,3 +1778,14 @@
     "protocol": "openid-connect",
     "protocolMapper": "oidc-audience-mapper",
 }
+
+RHSSO_NEW_USER = {
+    "email": "test_user@mailinator.com",
```
mirekdlugosz: As this is going to be overwritten anyway, could you use `@example.com`?
We don't really care about this value, so I don't see a reason to use any specific service.

omkarkhatavkar: sure I can

PR #7707, opened 2020-03-29
Refactor tests to end to end

dvagala:
Merge small tests.

### DIFF
--- a/tests/foreman/api/test_discoveryrule.py
+++ b/tests/foreman/api/test_discoveryrule.py
@@ -37,7 +37,7 @@ class DiscoveryRuleTestCase(APITestCase):
         :id: 25366930-b7f4-4db8-a9c3-a470fe4f3583
 
         :expectedresults: Rule should be created, modified and deleted successfully
-        with given attributes.
+            with given attributes.
 
         :CaseImportance: Critical
         """
@@ -58,8 +58,12 @@ class DiscoveryRuleTestCase(APITestCase):
             loc = entities.Location().create()
             hostgroup = entities.HostGroup(organization=[org]).create()
             discovery_rule = entities.DiscoveryRule(
-                name=name, search_=search, hostname=hostname,
-                organization=[org], location=[loc], hostgroup=hostgroup
+                name=name,
+                search_=search,
+                hostname=hostname,
+                organization=[org],
+                location=[loc],
+                hostgroup=hostgroup,
             ).create()
             self.assertEqual(name, discovery_rule.name)
             self.assertEqual(hostname, discovery_rule.hostname)
@@ -84,8 +88,13 @@ class DiscoveryRuleTestCase(APITestCase):
             discovery_rule.enabled = enabled
             discovery_rule = discovery_rule.update(
                 [
-                    'name', 'organization', 'location', 'hostgroup',
-                    'search_', 'max_count', 'enabled'
+                    'name',
+                    'organization',
+                    'location',
+                    'hostgroup',
+                    'search_',
+                    'max_count',
+                    'enabled',
                 ]
             )
             self.assertEqual(name, discovery_rule.name)
--- a/tests/foreman/api/test_discoveryrule.py
+++ b/tests/foreman/api/test_discoveryrule.py
@@ -23,28 +23,24 @@ from requests.exceptions import HTTPError
 
 from robottelo.datafactory import valid_data_list
 from robottelo.decorators import tier1
-from robottelo.decorators import tier2
 from robottelo.test import APITestCase
 
 
 class DiscoveryRuleTestCase(APITestCase):
     """Tests for ``katello/api/v2/discovery_rules``."""
 
-    @classmethod
-    def setUpClass(cls):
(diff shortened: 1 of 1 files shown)


tests/foreman/api/test_discoveryrule.py
```diff
@@ -23,221 +23,101 @@
...
+            hostgroup = entities.HostGroup(organization=[org]).create()
+            discovery_rule = entities.DiscoveryRule(
+                name=name,
+                search_=search,
+                hostname=hostname,
+                organization=[org],
+                location=[loc],
+                hostgroup=hostgroup,
+            ).create()
+            self.assertEqual(name, discovery_rule.name)
```
omkarkhatavkar: One suggestion from my end tries to use python assert rather than self. assert, As we already in phase moving that from PR  https://github.com/SatelliteQE/robottelo/pull/7686

dvagala: Okay it's fixed, thanks for pointing that out.

tests/foreman/api/test_discoveryrule.py
```diff
@@ -23,221 +24,91 @@
...
 class DiscoveryRuleTestCase(APITestCase):
     """Tests for ``katello/api/v2/discovery_rules``."""
 
-    @classmethod
-    def setUpClass(cls):
-        """Create a hostgroup which can be re-used in tests."""
-        super(DiscoveryRuleTestCase, cls).setUpClass()
-        cls.hostgroup = entities.HostGroup().create()
+    @tier1
+    def test_positive_end_to_end(self):
```
mshriver: Consider using a more descriptive test name here. We can assume its a positive type test (that should be in the polarion meta anyway).

This appears to be a standard 'CRUD' type test, where the given test case is creating, updating, and deleting.

Consider `test_discovery_rule_crud`

tests/foreman/api/test_discoveryrule.py
```diff
@@ -23,221 +24,91 @@
...
-
-    @tier1
-    def test_positive_update_org_loc(self):
-        """Update org and location of selected discovery rule
-
-        :id: 0f8ec302-f9de-4713-87b7-0f1aca515149
-
-        :expectedresults: Rule was updated and with given org & location
-        """
+        assert name == discovery_rule.name
```
mshriver: I don't see any teardown of the created entities in this test pattern.  The test eventually calls delete on the discovery rule - but that only happens if all of the interim assertions pass.

I think after creating these entities, we should use `request.addfinalizer(entity.delete)` so that on teardown (success or failure within the case) the entities are cleaned up after the test is done.

If the creation of these entities (other than discovery rule, as that is the target of the test) is moved into a fixture, then you don't have to use `addfinalizer` and you can clean them up after yielding.

tests/foreman/api/test_discoveryrule.py
rplevka: I believe @dvagala wanted to refactor this test the old-fashioned way to maintain consistency.
He is planning to do the full transiiton to pytest test case later, so perhaps I would merge it as it is first. Would that work for you?

btw: the location of the test describes it pretty well: it's in `tests/discoveryrule.py` which reflects in the test name in the report.

tests/foreman/api/test_discoveryrule.py
rplevka: wouldn't `request.addfinalizer(entity.delete)` crash the whole pytest session in case the server fails to delete it? Or will it be treated same as if the fixture failed? just asking.

I prefer the cleanup to be put into a fixture.

tests/foreman/api/test_discoveryrule.py
mshriver: ack, its a suggestion.

tests/foreman/api/test_discoveryrule.py
rplevka: iirc there already were attempts to tear down the taxonomies after test execution, but due the way Satellite used to treat them*, the teardown failed too often.
It would definitely be worth revisiting this topic to explore what has changed since then.
Such inconsistencies in behaviour were reported and a lot of them got fixed eventually.
A lot of seemingly weird behaviour was also caused by using "admin" user account, which is not limited to selected subset of taxonomies. perhaps having a (non admin) session user for every test (created vy a fixture) might deliver the needed isolation as well as more accurate UX. This is a broad topic deserving its own issue ticket and should be handled at design level .
We probably don't want to block this PR.

*e.g. - a test1 created org1 for its own use. In the meantime, test2 created a Smart Proxy entity, which automatically got assigned to all existing orgs. After that, test1 tried to tear down the org1 which satellite refused to do as "it was used by the Smart Proxy".

rplevka: if you're happy with this and consider the comment resolved, please Merge the PR. it already got all the required Approvals.
thanks!

omkarkhatavkar: I do agree with @rplevka this is with the whole project and needs discussion and design implementation steps. We can plan a similar #7686 strategy to every where

PR #7681, opened 2020-03-08
Generate custom certs in setup

swadeley:
* Avoids problem of certs expiring
* Avoids tying certs to specific infrastucture servers

PR #7700, opened 2020-03-19
Test positive generate capsule certs

swadeley:
Tests for creating capsule certs using absolute and relative paths.

Bug 1466688 - capsule-certs-generate --certs-tar does not accept relative path

PR #7708, opened 2020-03-30
Remove unwanted swp file

swadeley:
In commit for #7681 I think I mistakenly pushed a swp file, yet it does not show in github PR #7681

Please check if you see it in master branch:
git log -- .robottelo.properties.swp


swadeley:
> It actually came in #7700 , but yes, shouldn't be here.
> 
oh right, its at the top when you view the comits, thanks

> For whatever reason our .gitignore does not specify swp files. You can add it to global gitignore to avoid similar mistake in the future:
> 
> ```shell
> $ echo '.*swp' > ~/.gitignore
> $ git config --global core.excludesfile ~/.gitignore
> ```

yep, done that :)   (forgot to set this up when I got my new laptop)

swadeley:
@mirekdlugosz 
> For whatever reason our .gitignore does not specify swp files. You can add it to global gitignore to avoid similar mistake in the future:
> 
it was added once here:
git show decabdbecf845633c62ed8f407137e704a745801

unfortunately, then removed:
git show 9a40f4687f3cc2afcf915d44f67e690b49a17586

swadeley:
thank you

PR #7710, opened 2020-04-01
Automated stubbed test cases of settings.

devendra104:
Automated 2 stubbed test cases one for positive and other for the negative scenario.
1- test_positive_update_login_page_footer_text_with_long_string
2- test_negative_settings_access_to_non_admin

### DIFF
--- a/tests/foreman/ui/test_settings.py
+++ b/tests/foreman/ui/test_settings.py
@@ -15,10 +15,12 @@
 :Upstream: No
 """
 import pytest
+from airgun.session import Session
 from fauxfactory import gen_url
 from nailgun import entities
 from pytest import raises
 
+from robottelo.cli.user import User
 from robottelo.datafactory import filtered_datapoint
 from robottelo.datafactory import gen_string
 from robottelo.decorators import fixture
@@ -209,11 +211,10 @@ def test_positive_selectors(session):
                 setting_cleanup(setting_attr, str(original_val.default))
 
 
-@stubbed()
 @tier3
-def test_positive_update_login_page_footer_text_with_long_string():
-    """Attempt to update parameter "Login_page_footer_text"
-        with long length string under General tab
+def test_positive_update_login_page_footer_text_with_long_string(session):
+    """Testing to update parameter "Login_page_footer_text with long length
+    string under General tab
 
     :id: b1a51594-43e6-49d8-918b-9bc306f3a1a2
 
@@ -228,13 +229,20 @@ def test_positive_update_login_page_footer_text_with_long_string():
 
     :CaseImportance: Medium
 
-    :CaseAutomation: notautomated
-
     :CaseLevel: Acceptance
     """
+    property_name = 'login_text'
+    property_value = entities.Setting().search(query={'search': f'name={property_name}'})[0]
+    login_text_data = gen_string('alpha', 270)
+    with session:
+        try:
+            session.settings.update(f"name={property_name}", f"{login_text_data}")
+            result = session.login.logout()
+            assert result["login_text"] == login_text_data
+        finally:
+            setting_cleanup(setting_name=property_name, setting_value=property_value.value)
 
 
-@stubbed()
 @tier3
 def test_negative_settings_access_to_non_admin():
     """Check non admin users can't access Administer -> Settings tab
@@ -242,20 +250,26 @@ def test_negative_settings_access_to_non_admin():
     :id: 34bb9376-c5fe-431a-ac0d-ef030c0ab50e
 
     :steps:
+
         1. Login with non admin user
         2. Check "Administer" tab is not present
         3. Navigate to /settings
         4. Check message permission denied is present
 
-    :expectedresults: Administer -> Settings tab should not be available to
-        non admin users
+    :expectedresults: Administer -> Settings tab should not be available to non admin users
 
     :CaseImportance: Medium
 
-    :CaseAutomation: notautomated
-
     :CaseLevel: Acceptance
     """
+    login = gen_string('alpha')
+    password = gen_string('alpha')
+    entities.User(admin=False, login=login, password=password).create()
+    try:
+        with Session(user=login, password=password) as session:
+            assert session.settings.browser.title == 'Permission denied'
+    finally:
+        User.delete({'login': login})
 
 
 @stubbed()


devendra104:
Test Result:
test_positive_update_login_page_footer_text_with_long_string

```
$ pytest tests/foreman/ui/test_settings.py::test_positive_update_login_page_footer_text_with_long_string

============================================================================================ 1 passed in 71.60 seconds ============================================================================================
```
test_negative_settings_access_to_non_admin

```
$ pytest tests/foreman/ui/test_settings.py::test_negative_settings_access_to_non_admin

============================================================================================ 1 passed in 36.38 seconds ============================================================================================
```

jameerpathan111:
@devendra104 travis failed.
Try running `#pre-commit run -a` then push changes again.

tests/foreman/ui/test_settings.py
```diff
@@ -209,53 +211,58 @@ def test_positive_selectors(session):
...
+            query={'search': 'name={}'.format(property_name)}
+        )
+    ][0]
+    login_text_data = gen_string('alpha', 270)
+    with session:
+        try:
+            session.settings.update("name={}".format(property_name),
+                                    "{}".format(login_text_data))
+            result = session.settings.read("name={}".format(property_name))
+            assert result["table"][0]["Value"] == login_text_data
```
omkarkhatavkar, line 240: Can you verify this on login page, you can use the logout method as well to navigate there.

tests/foreman/ui/test_settings.py
```diff
@@ -209,53 +211,58 @@ def test_positive_selectors(session):
...
     :CaseLevel: Acceptance
     """
+    property_name = 'login_text'
+    property_value = [
+        ele.value
+        for ele in entities.Setting().search(
+            query={'search': 'name={}'.format(property_name)}
+        )
+    ][0]
+    login_text_data = gen_string('alpha', 270)
```
omkarkhatavkar, line 234: anything about 270 char size?

tests/foreman/ui/test_settings.py
```diff
@@ -209,53 +211,58 @@ def test_positive_selectors(session):
...
-def test_positive_update_login_page_footer_text_with_long_string():
-    """Attempt to update parameter "Login_page_footer_text"
-        with long length string under General tab
+def test_positive_update_login_page_footer_text_with_long_string(session):
+    """Testing to update parameter "Login_page_footer_text with long length
+    string under General tab
 
     :id: b1a51594-43e6-49d8-918b-9bc306f3a1a2
 
-    :steps:
```
jameerpathan111, line 220: Shouldn't we keep these test steps?

tests/foreman/ui/test_settings.py
devendra104: The purpose of this test case is to check the "login page footer text" updated with long strings or not, and that's the only reason I used string length.

tests/foreman/ui/test_settings.py
devendra104: I think it is not required here, because the steps are very straight forward.
If you think it should be then definitely I will update it.

jameerpathan111: ok.

devendra104: fixed.

tests/foreman/ui/test_settings.py
mirekdlugosz: Yeah, I think we can skip checking if setting is updated, we are much more interested in footer on login page.

The problematic part is that while we expose `logout()` method of base view, it actually doesn't do anything. See https://github.com/SatelliteQE/airgun/pull/397 and discussion in it. I think logout should be modeled after `RHSSOLoginEntity::logout`, i.e. after logging out it should instantiate login page view and return `read()` on that.

Another problem is that our login view exposes very limited set of widgets, and footer is not one of them.

omkarkhatavkar: @mirekdlugosz RHSSOLoginEntity:: logout is special logout which navigates to intermediate page. We will need to merge SatelliteQE/airgun#397. This is waiting for a long time, we can avoid the to get used by all other test cases. @mirekdlugosz what's your view?

devendra104: > Yeah, I think we can skip checking if setting is updated, we are much more interested in footer on login page.
Agree.
> 
> The problematic part is that while we expose `logout()` method of base view, it actually doesn't do anything. See [SatelliteQE/airgun#397](https://github.com/SatelliteQE/airgun/pull/397) and discussion in it. I think logout should be modeled after `RHSSOLoginEntity::logout`, i.e. after logging out it should instantiate login page view and return `read()` on that.
> 
> Another problem is that our login view exposes very limited set of widgets, and footer is not one of them.

Yes, it is still in review state, once it gets merge then we will have two options to achieve it.

1- Raise the PR for view update(add the login_text locator in RhssoExternalLogoutView) and Use RHSSOLoginEntity:: logout

2- Update the omkar's logout PR.

I tested it in both ways but the login page is not properly loaded, Hence, We are unable to read the login view's content. It is possible if we modify the element mentioned in browser_wait_for_element and that is not the correct way. 
Can anybody tell me how we can wait for a few seconds in LoginEntity::logout/RHSSOLoginEntity::logout method before calling their login view?

devendra104: > I tested it in both ways but the login page is not properly loaded, Hence, We are unable to read the login view's content. It is possible if we modify the element mentioned in browser_wait_for_element and that is not the correct way.
> Can anybody tell me how we can wait for a few seconds in LoginEntity::logout/RHSSOLoginEntity::logout method before calling their login view?

Problem resolved.

devendra104: Raised Airgun PR https://github.com/SatelliteQE/airgun/pull/474

devendra104:
Please check, if something is pending.

tests/foreman/ui/test_settings.py
```diff
@@ -228,34 +229,50 @@ def test_positive_update_login_page_footer_text_with_long_string():
...
-    :CaseAutomation: notautomated
-
     :CaseLevel: Acceptance
     """
+    login = gen_string('alpha')
+    password = gen_string('alpha')
+    entities.User(admin=False, login=login, password=password).create()
+    try:
+        with Session(user=login, password=password) as session:
+            assert session.settings.session.browser.title == 'Permission denied'
```
mirekdlugosz, line 273:
```suggestion
            assert session.browser.title == 'Permission denied'
```

mirekdlugosz:
@SatelliteQE/quality-engineers can I have second ACK, so we can merge it today before I create 6.7.z branch?

tests/foreman/ui/test_settings.py
devendra104: Fixed, It should be session.settings.browser.title.

devendra104:
@omkarkhatavkar @jameerpathan111 Please review and merge.

tests/foreman/ui/test_settings.py
mirekdlugosz: Right, we expose `browser` in `session`, but first you must initialize it, which is done automatically when accessing any entity. :+1:

mirekdlugosz:
I created 6.7.z branch today. Unfortunately, this PR did not make it due to lack of review.

If you want this to run on 6.7.z as well, please raise cherry-pick PR against 6.7.z branch after this one is merged into master.

PR #7711, opened 2020-04-01
Use correct component for RHAI tests

mirekdlugosz:
There was a misunderstanding.

"Component Tiering Proposal" document I have used last time does not mention "Insights Plugin" at all - it only has "Insights Inventory Plugin". I thought they are the same thing.

Turns out, they are not. Insights Plugin is basically proxy between cloud.rh.com and insights-client on registered host, we collaborate with Insights QE on that. Insights Inventory Plugin is now renamed rh_cloud and is for uploading data about registered hosts to cloud.rh.com. We collaborate with Subscription Management QE on that.

Both components are available in bugzilla and Mojo, neither of them is deprecated.

Since this is metadata update, no tests are needed.

### DIFF
--- a/testimony.yaml
+++ b/testimony.yaml
@@ -58,6 +58,7 @@ CaseComponent:
     - Infobloxintegration
     - Infrastructure
     - InsightsInventoryPlugin
+    - InsightsPlugin
     - Installer
     - InterSatelliteSync
     - katello-agent
--- a/tests/foreman/rhai/test_rhai.py
+++ b/tests/foreman/rhai/test_rhai.py
@@ -6,7 +6,7 @@
 
 :CaseLevel: Acceptance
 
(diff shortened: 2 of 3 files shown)


testimony.yaml
```diff
@@ -58,6 +58,7 @@ CaseComponent:
     - Infobloxintegration
     - Infrastructure
     - InsightsInventoryPlugin
```
omkarkhatavkar, line 60: 'InsightsInventoryPlugin' => this is required because of Polarian?

mirekdlugosz: Both are required, as both are valid values - these are two different plugins, both supported.

We don't use `InsightsInventoryPlugin` anywhere now, as this is newer thing. Presumably I will be adding these tests one day in the future.

PR #7712, opened 2020-04-01
test manipulates with default settings and this is causing problems to tests which run in paralel

tstrych:
This was really, really hard to find.

My test was failing from time to time. ( unable to find host, but host info works 5 seconds ago ) this is really weird. 

I found all the test which runs in parallel during the time period when the my test was unable to take a look at host info, I needed to compare three runs to see this is the only test which can cause that. 

And the proof:
I add debug point to this test and then I run `hammer host list` on my sat, result:  it was empty. 
After the test ends, `hammer host list` shows the host I have.


rplevka:
very great catch. these are always fun :)
thank you

PR #7717, opened 2020-04-06
Report Portal Module

jyejare:
Report Portal Module in Robottelo.

Consists of:

ReportPortal:
------------------------------
- [x] rp = ReportPortal()
```
In [3]: rp                                                                                                                                                                                                           
Out[3]: <robottelo.report_portal.portal.ReportPortal at 0x7feebe552710>
```

- [x] rp.launches(sat_version='6.7.0', launch_type='satellite6')
```
Out[6]: 
{'6.7.0': {'20.0': {'owner': 'robottelo',
  'share': False,
  'description': 'downstream snap automation',
  'id': '5e8982e1149b3f0001090fd3',
  'name': 'Upgrades',
  'number': 68,
  'start_time': 1586041117000,
  'end_time': 1586069630387,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '394',
    'passed': '311',
    'failed': '71',
    'skipped': '12'},
   'defects': {'product_bug': {'total': 3, 'PB001': 3},
    'automation_bug': {'AB001': 3, 'total': 5, 'ab_1ibec4rs6jf3r': 2},
    'system_issue': {'total': 43, 'SI001': 43},
    'to_investigate': {'total': 19, 'TI001': 19},
    'no_defect': {'ND001': 13, 'total': 13}}},
  'tags': ['Satellite', 'rhel7', '6.7', '6.7.0-20.0'],
  'mode': 'DEFAULT',
  'isProcessing': False,
  'approximateDuration': 36407659.6,
  'hasRetries': False},
 {'owner': 'robottelo',
  'share': False,
.
.
.
```
- [x] rp.launch(sat_version='6.7.0', snap_version='20.0')
```
Out[7]: 
{'owner': 'robottelo',
 'share': False,
 'description': 'downstream snap automation',
 'id': '5e8982e1149b3f0001090fd3',
 'name': 'Upgrades',
 'number': 68,
 'start_time': 1586041117000,
 'end_time': 1586069630387,
 'status': 'FAILED',
 'statistics': {'executions': {'total': '394',
   'passed': '311',
   'failed': '71',
   'skipped': '12'},
  'defects': {'product_bug': {'total': 3, 'PB001': 3},
   'automation_bug': {'AB001': 3, 'total': 5, 'ab_1ibec4rs6jf3r': 2},
   'system_issue': {'total': 43, 'SI001': 43},
   'to_investigate': {'total': 19, 'TI001': 19},
   'no_defect': {'ND001': 13, 'total': 13}}},
 'tags': ['Satellite', 'rhel7', '6.7', '6.7.0-20.0'],
 'mode': 'DEFAULT',
 'isProcessing': False,
 'approximateDuration': 36407659.6,
 'hasRetries': False}
```
Tests:
-----------------------------
- [x] launch = rp.launches(sat_version='6.7.0', launch_type='satellite6').pop('20.0')
```
<__main__.Launch object at 0x7fdc5ba28e50>
```

- [x] launch.tests(status='failed')
```
Out[10]: 
['tests/foreman/api/test_contentmanagement.py::CapsuleContentManagementTestCase::test_positive_capsule_pub_url_accessible', 'tests/foreman/api/test_contentmanagement.py::CapsuleContentManagementTestCase::test_positive_uploaded_content_library_sync', 'tests/foreman/ui/test_ldap_authentication.py::test_single_sign_on_using_rhsso', 'tests/foreman/ui/test_ldap_authentication.py::test_external_logout_rhsso', 'tests/foreman/ui/test_ldap_authentication.py::test_external_new_user_login', ...]
```

- [x] launch.tests(defect_type='automation_bug')
```
Out[11]:                                            
['tests/foreman/api/test_contentmanagement.py::CapsuleContentManagementTestCase::test_positive_uploaded_content_library_sync', 'tests/foreman/cli/test_satellitesync.py::ContentViewSync::test_positive_export_import_redhat_cv_with_huge_contents', 'tests/foreman/ui/test_host.py::test_positive_bulk_delete_host', 'tests/foreman/cli/test_contentview.py::ContentViewTestCase::test_positive_republish_after_rh_content_removed', 'tests/foreman/ui/test_computeresource_vmware.py::test_positive_resource_vm_power_management', 'tests/foreman/ui/test_contentview.py::test_positive_publish_rh_content_with_errata_by_date_filter', 'tests/foreman/cli/test_contentview.py::ContentViewTestCase::test_positive_remove_cv_version_from_multi_env_capsule_scenario', 'tests/foreman/api/test_template.py::TemplateSyncTestCase::test_positive_export_json_output', 'tests/foreman/api/test_template.py::TemplateSyncTestCase::test_positive_export_filtered_templates_to_localdir', 'tests/foreman/ui/test_operatingsystem.py::test_positive_update_with_params']
```
- [x] launch.tests(status='failed', defect_type='to_investigate')
```
Out[15]: 
['tests/foreman/api/test_contentmanagement.py::CapsuleContentManagementTestCase::test_positive_uploaded_content_library_sync', 'tests/foreman/cli/test_satellitesync.py::ContentViewSync::test_positive_export_import_redhat_cv_with_huge_contents', 'tests/foreman/ui/test_host.py::test_positive_bulk_delete_host', 'tests/foreman/cli/test_contentview.py::ContentViewTestCase::test_positive_republish_after_rh_content_removed', 'tests/foreman/ui/test_computeresource_vmware.py::test_positive_resource_vm_power_management', 'tests/foreman/ui/test_contentview.py::test_positive_publish_rh_content_with_errata_by_date_filter', 'tests/foreman/cli/test_contentview.py::ContentViewTestCase::test_positive_remove_cv_version_from_multi_env_capsule_scenario', 'tests/foreman/api/test_template.py::TemplateSyncTestCase::test_positive_export_json_output', 'tests/foreman/api/test_template.py::TemplateSyncTestCase::test_positive_export_filtered_templates_to_localdir', 'tests/foreman/ui/test_operatingsystem.py::test_positive_update_with_params']
```

Attributes:
-----------------------------
- [x] launch.info
- [x] launch.name
- [x] launch.satellite_version
- [x] launch.snap_version
- [x] launch.statistics
- [x] rp.defect_types
- [x] rp.statuses

### DIFF
--- a/robottelo/report_portal/portal.py
+++ b/robottelo/report_portal/portal.py
@@ -1,0 +1,258 @@
+import logging
+import re
+
+import requests
+from tenacity import retry
+from tenacity import stop_after_attempt
+from tenacity import wait_fixed
+
+from robottelo.config import settings
+
+LOGGER = logging.getLogger(__name__)
+launch_types = ['satellite6', 'upgrades']
+
+
+class ReportPortal:
+    """Represents ReportPortal
+
+    This holds the properties and functions to interact with Report Portal properties and launches
+    """
+
+    defect_types = {
+        'product_bug': 'PB001',
+        'rhel_bug': 'pb_1ies7k2sxbsdh',
+        'satellite_bug': 'pb_vf6zbvnxv77d',
+        'automation_bug': 'AB001',
+        'robottelo_bug': 'ab_1ibec4rs6jf3r',
+        'nailgun_bug': 'ab_t4kkyw0yaaet',
+        'airgun_bug': 'ab_u7umbjyvti7a',
+        'system_issue': 'SI001',
+        'saucelabs_issue': 'si_rhufdawx03e8',
+        'to_investigate': 'TI001',
+        'no_defect': 'ND001',
+    }
+    statuses = ['failed', 'passed', 'skipped', 'interrupted', 'in_progress']
+
+    def __init__(self):
+        """Configure the settings and initiate report portal properties"""
+        if not settings.configured:
+            settings.configure()
+        self.rp_url = settings.report_portal.rp_url
+        self.rp_project = settings.report_portal.rp_project
+        self.rp_api_key = settings.report_portal.rp_key
+
+    @property
+    def api_url(self):
+        """Super url of report portal
+        :returns: Base url for API request
+        """
+        return f'{self.rp_url}/api/v1/{self.rp_project}/'
+
+    @property
+    def headers(self):
+        """The headers for Report Portal Requests.
+        :returns: header for API request
+        """
+        return {'Authorization': f'Bearer {self.rp_api_key}'}
+
+    def _format_launches(self, launches):
+        """The pretty formatter function that formats launches in a structured way
+
+        :param filter launches: Satellite or Upgrade Type launches
+
+        :returns dict: Launches, keyed with their snap_versions formatted as,
+            `{'snap_version1':launch_object1, 'snap_version2': launch_object2}`
+        """
+        formatted_launches = {}
+        for launch_info in launches:
+            launch = Launch(rp=self, launch_info=launch_info)
+            if hasattr(launch, 'satellite_version'):
+                sat_ver = launch.satellite_version
+                snap_dict = {launch.snap_version: launch}
+                if sat_ver not in formatted_launches:
+                    formatted_launches[sat_ver] = snap_dict
+                else:
+                    formatted_launches[sat_ver].update(snap_dict)
+        return formatted_launches
+
+    def _launch_requester(self):
+        """The launch GET requester to fetch the all available launches of ReportPortal
+
+        :returns dict: The json of all RP launches
+        """
+        params = {'page.page': 1, 'page.size': 500, 'page.sort': 'start_time'}
+        resp = requests.get(
+            url=f'{self.api_url}/launch', headers=self.headers, params=params, verify=False
+        )
+        resp.raise_for_status()
+        return resp.json()
+
+    def launches(self, sat_version=None, launch_type='satellite6'):
+        """Returns launches in Report Portal customized by sat_version, launch_type and
+        latest number of launches sorted by latest sat version/snap version.
+
+        This does not includes each tests data for all tests in all launches, but it includes
+        just an overview data for all and each launch in Report Portal
+
+        :param str sat_version: The satellite version
+            If its not specified, then latest count of launches of satellite versions returned
+        :param str launch_type: Either satellite6 or upgrades,
+            default returns only non-upgrade launches
+        :returns dict: The launches of Report portal.
+            if sat_version is given,
+            ```{'snap_version1':launch_object1, 'snap_version2':launch_object2}```
+            else,
+            ```{'sat_version1':{'snap_version1':launch_object1, ..}, 'sat_version2':{}}```
+        """
+        if launch_type not in launch_types:
+            raise ValueError(
+                f'Wrong Launch Type is given as {launch_type} should be one of {launch_types}'
+            )
+        data = self._launch_requester()
+        sorted_launches = sorted(
+            data['content'], key=lambda launch: launch['start_time'], reverse=True
+        )
+        typed_launches = filter(
+            lambda launch: launch['name'].lower() == launch_type, sorted_launches
+        )
+        launches_ = self._format_launches(typed_launches)
+        if sat_version:
+            if sat_version not in launches_:
+                raise ValueError(
+                    f'The given satellite version \'{sat_version}\' is not available in '
+                    f'Report portal project \'{self.rp_project}\''
+                )
+            else:
+                return launches_[sat_version]
+        return launches_
+
+    def launch(self, sat_version, snap_version=None, launch_type='satellite6'):
+        """Returns a specific launch data in Report Portal Project
+
+        This does not includes each tests data in launch
+
+        :param str sat_version: The satellite version
+        :param str snap_version: The snap version of a given satellite version
+            if None, the latest launch data of a given sat_version is returned
+        :param str launch_type: Either satellite6 or upgrades, default returns only
+            non-upgrade launches
+        :returns dict: The data directory of requested or latest launch
+        """
+        launches_ = self.launches(sat_version=sat_version, launch_type=launch_type)
+        if snap_version:
+            if snap_version not in launches_.keys():
+                raise ValueError(
+                    f'The given snap_version \'{snap_version}\' is not available '
+                    f'in satellite version \'{sat_version}\''
+                )
+        else:
+            # Get the latest launch of the given sat_version
+            snap_version = next(iter(launches_.keys()))
+        return launches_[snap_version]
+
+
+class Launch:
+    def __init__(self, rp, launch_info):
+        """Initiates the Launch and its properties of a specified snap version in
+        a specified satellite version
+
+        :param ReportPortal rp: A report portal object
+        :param dict launch_info: The dict information about a launch
+        """
+        self.report_portal = rp
+        self.info = launch_info
+        self.name = self.info['name']
+        self.statistics = self.info['statistics']['executions']
+        self._versions()
+
+    def _versions(self):
+        """Sets satellite and snap version attributes of a launch"""
+        version_compiler = re.compile(r'([\d\.]+)[\.-](\d+\.\d|[A-Z]+)')
+        launch_tags = self.info['tags']
+        try:
+            launch_name = next(filter(version_compiler.search, launch_tags))
+        except StopIteration:
+            LOGGER.debug(
+                'Invalid launch with no build name is detected. '
+                f'The launch has tags {launch_tags}'
+            )
+            return
+        self.satellite_version = re.search(version_compiler, launch_name).group(1)
+        self.snap_version = re.search(version_compiler, launch_name).group(2)
+
+    def _test_params(self, status, defect_type):
+        """Customise parameters for Test items API request
+
+        :returns dict: The parameters dict for API test items request
+        """
+        params = [
+            ('filter.eq.launch', self.info['id']),
+            ('page.page', 1),
+            ('page.size', 50),
+            ('page.sort', 'start_time'),
+        ]
+        rp_defect_types = ReportPortal.defect_types
+        if defect_type:
+            if defect_type in rp_defect_types:
+                params.insert(0, ('filter.eq.issue$issue_type', rp_defect_types[defect_type]))
+            else:
+                raise ValueError(
+                    f'Invalid value \'{defect_type}\' for defect type parameter, '
+                    f'should be one of {[*rp_defect_types.keys()]}'
+                )
+        rp_statuses = ReportPortal.statuses
+        if status:
+            if status in rp_statuses:
+                params.insert(2, ('filter.eq.status', status))
+            else:
+                raise ValueError(
+                    f'Invalid value \'{status}\' for status parameter, '
+                    f'should be one of {rp_statuses}'
+                )
+        return dict(params)
+
+    @retry(
+        stop=stop_after_attempt(4), wait=wait_fixed(10),
+    )
+    def _test_requester(self, params, page):
+        """The Test Items GET requester to fetch the data on a page
+
+        If any error and before Failing explicitly, it retries for 3 times with 10 seconds delay
+
+        :returns tuple (int, list): Total pages count and the list of tests along with
+            each tests properties in a page
+        """
+        params['page.page'] = page
+        resp = requests.get(
+            url=f'{self.report_portal.api_url}/item',
+            headers=self.report_portal.headers,
+            params=params,
+            verify=False,
+        )
+        resp.raise_for_status()
+        total_pages = resp.json()['page']['totalPages']
+        pagedata = resp.json().get('content')
+        return total_pages, pagedata
+
+    def tests(self, status=None, defect_type=None):
+        """Returns tests data customized by kwargs parameters.
+
+        This is a main function that will be called to retrieve the tests data
+        of a particular test status or/and defect_type
+
+        :param str status: Filter tests of a launch with tests `status`
+        :param str defect_type: Filter tests of a launch with tests `defect_type`
+        :returns dict: All filtered tests dict based on params data keyed by test name and test
+            properties as value, in format -
+            ```{'test_name1':test1_properties_dict, 'test_name2':test2_properties_dict}```
+        """
+        page = 1
+        params = self._test_params(status, defect_type)
+        total_pages, data = self._test_requester(params=params, page=page)
+        while page < total_pages:
+            page += 1
+            _, pagedata = self._test_requester(params=params, page=page)
+            data.extend(pagedata)
+        # formatting tests data to return tests data keyed by test name
+        tests_ = {test['name']: test for test in data}
+        return tests_
--- a/robottelo.properties.sample
+++ b/robottelo.properties.sample
@@ -575,3 +575,12 @@ rhsso_user=
 user_password=
 # RH-SSO Host Realm
 realm=
+
+# Report Portal Section for Rerunning failed tests
+[report_portal]
+# Portal URL
+# portal_url=
+# Project Name
+# project=satellite6
+# API key of an user
+# api_key=
--- a/robottelo/config/base.py
+++ b/robottelo/config/base.py
@@ -1282,6 +1282,31 @@ class VirtWhoSettings(FeatureSettings):
         return validation_errors
 
 
+class ReportPortalSettings(FeatureSettings):
+    """Report portal settings definitions."""
+
+    def __init__(self, *args, **kwargs):
+        super(ReportPortalSettings, self).__init__(*args, **kwargs)
+        self.rp_url = None
+        self.rp_project = None
+        self.rp_key = None
+
+    def read(self, reader):
+        """Read Report portal settings."""
+        self.rp_url = reader.get('report_portal', 'portal_url')
+        self.rp_project = reader.get('report_portal', 'project')
+        self.rp_key = reader.get('report_portal', 'api_key')
+
+    def validate(self):
+        """Validate Report portal settings."""
+        validation_errors = []
+        if not all(self.__dict__.values()):
+            validation_errors.append(
+                'All [report_portal] {} options must be provided'.format(self.__dict__.keys())
+            )
+        return validation_errors
+
+
 class Settings(object):
     """Robottelo's settings representation."""
 
@@ -1349,6 +1374,7 @@ class Settings(object):
         self.upgrade = UpgradeSettings()
         self.vmware = VmWareSettings()
         self.virtwho = VirtWhoSettings()
+        self.report_portal = ReportPortalSettings()
 
     def configure(self, settings_path=None):
         """Read the settings file and parse the configuration.


mirekdlugosz:
This is work in progress, or ready for review?

Could you add some more context? E.g. I have no idea what that is and what is it for. I imagine other reviewers feel the same.

jyejare:
@mirekdlugosz Most of the work is done, except The Pending things in Description.

I opened a PR for the people to review the 'code' who and all aware of what we are doing with ReportPortal APIs. But I should add docstrings soon for all to review it.

I have added DO_NOT_MERGE and In_Progress labels by now.

jyejare:
@mirekdlugosz @mshriver :

Here are the logs of the working Report Portal module and Proof to merge :)  :

```
In [7]: from robottelo.report_portal.portal import Tests                                                                                                                                                            

In [8]: latest_launch_tests = Tests(launch=1)                                                                                                                                                                       

In [9]: latest_launch_tests                                                                                                                                                                                         
Out[9]: <robottelo.report_portal.portal.Tests at 0x7fd5b0c91890>

In [10]: latest_launch_tests.tests(status='FAILED', defect_type='to_investigate')                                                                                                                                   
2020-04-06 23:10:08 - robottelo.report_portal.portal - INFO - Fetching Tests for Launch: ['Satellite', 'rhel7', '6.7', '6.7.0-20.0'] to select failed tests to rerun
2020-04-06 23:10:08 - robottelo.report_portal.portal - DEBUG - Fetching results for Page : 1
Out[10]: 
[{'id': '5e8982f2149b3f000109101d',
  'name': 'tests/foreman/cli/test_contentview.py::ContentViewTestCase::test_positive_remove_cv_version_from_multi_env_capsule_scenario',
  'description': 'test_positive_remove_cv_version_from_multi_env_capsule_scenario time: 1894.821',
  'tags': ['High', 'foreman', 'contentviews', 'ltran'],
  'type': 'STEP',
  'start_time': 1586052882000,
  'end_time': 1586054776821,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:6b678b1c87578089e7da6ed44b612d8f'},
 {'id': '5e898301149b3f0001091065',
  'name': 'tests/foreman/ui/test_contentview.py::test_positive_subscribe_system_with_rh_custom_spin',
  'description': 'test_positive_subscribe_system_with_rh_custom_spin time: 620.229',
  'tags': ['foreman', 'Low', 'contentviews', 'ltran'],
  'type': 'STEP',
  'start_time': 1586058936000,
  'end_time': 1586059556229,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:d749647583efbd629d5f1d8a6b536a15'},
 {'id': '5e898302149b3f0001091068',
  'name': 'tests/foreman/ui/test_dashboard.py::test_positive_user_access_with_host_filter',
  'description': 'test_positive_user_access_with_host_filter time: 570.874',
  'tags': ['High', 'foreman', 'ogajduse', 'dashboard'],
  'type': 'STEP',
  'start_time': 1586059559000,
  'end_time': 1586060129874,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:873c253fa3fc472ff038f1d02bdd2486'},
 {'id': '5e898302149b3f000109106b',
  'name': 'tests/foreman/ui/test_discoveredhost.py::test_positive_pxe_based_discovery',
  'description': 'test_positive_pxe_based_discovery time: 1564.067',
  'tags': ['rplevka', 'foreman', 'Critical', 'discoveryplugin'],
  'type': 'STEP',
  'start_time': 1586061372000,
  'end_time': 1586062936067,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:a80afd66f0809b71c8ba47c8e601097e'},
 {'id': '5e898303149b3f000109106e',
  'name': 'tests/foreman/ui/test_discoveredhost.py::test_positive_pxe_less_with_dhcp_unattended',
  'description': 'test_positive_pxe_less_with_dhcp_unattended time: 329.725',
  'tags': ['rplevka', 'foreman', 'Critical', 'discoveryplugin'],
  'type': 'STEP',
  'start_time': 1586061691000,
  'end_time': 1586062020725,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:cb0b4a75364bbff678d826cdfffb8fa0'},
 {'id': '5e898304149b3f0001091073',
  'name': 'tests/foreman/ui/test_discoveredhost.py::test_positive_auto_provision_host_with_rule',
  'description': 'test_positive_auto_provision_host_with_rule time: 10.287',
  'tags': ['High', 'rplevka', 'foreman', 'discoveryplugin'],
  'type': 'STEP',
  'start_time': 1586062162000,
  'end_time': 1586062172287,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:d5c9aed754fe7176bb158ee87f40ca49'},
 {'id': '5e89830a149b3f0001091090',
  'name': 'tests/foreman/ui/test_subscription.py::test_positive_access_with_non_admin_user_with_manifest',
  'description': 'test_positive_access_with_non_admin_user_with_manifest time: 63.399',
  'tags': ['foreman', 'subscriptionmanagement', 'Critical', 'chiggins'],
  'type': 'STEP',
  'start_time': 1586063592000,
  'end_time': 1586063655399,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:64c69108120415ff07848205dfdd4630'},
 {'id': '5e898331149b3f0001091148',
  'name': 'tests/foreman/cli/test_activationkey.py::ActivationKeyTestCase::test_positive_usage_limit',
  'description': 'test_positive_usage_limit time: 420.764',
  'tags': ['activationkeys', 'foreman', 'Critical', 'chiggins'],
  'type': 'STEP',
  'start_time': 1586064412000,
  'end_time': 1586064832764,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:4316ce48ddcfdb1a822f0f7c7b48630c'},
 {'id': '5e898362149b3f000109123c',
  'name': 'tests/foreman/ui/test_contentcredential.py::test_positive_end_to_end',
  'description': 'test_positive_end_to_end time: 82.986',
  'tags': ['High', 'jcallaha', 'foreman', 'gpgkeys'],
  'type': 'STEP',
  'start_time': 1586065991000,
  'end_time': 1586066073986,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:9f73cb81e16c3f2753482e9c385b1ea7'},
 {'id': '5e89836c149b3f000109126e',
  'name': 'tests/foreman/ui/test_computeresource_azurerm.py::test_positive_azurerm_host_provision_ud',
  'description': 'test_positive_azurerm_host_provision_ud time: 830.107',
  'tags': ['vijsingh', 'foreman', 'computeresources-azure', 'Critical'],
  'type': 'STEP',
  'start_time': 1586066435000,
  'end_time': 1586067265107,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:c25c2169ee9a551cc07055c37d52364a'},
 {'id': '5e898375149b3f0001091297',
  'name': 'tests/foreman/ui/test_hostcollection.py::test_positive_remove_package',
  'description': 'test_positive_remove_package time: 174.142',
  'tags': ['High', 'swadeley', 'foreman', 'hostcollections'],
  'type': 'STEP',
  'start_time': 1586067035000,
  'end_time': 1586067209142,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:b31f5f144d2fe5fbf0562611f051e326'},
 {'id': '5e898378149b3f00010912a9',
  'name': 'tests/foreman/ui/test_organization.py::test_positive_end_to_end',
  'description': 'test_positive_end_to_end time: 308.185',
  'tags': ['lvrtelov', 'foreman', 'Critical', 'organizationslocations'],
  'type': 'STEP',
  'start_time': 1586067365000,
  'end_time': 1586067673185,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:48573c831f2d2d9879b3778a17e28773'},
 {'id': '5e89837c149b3f00010912ba',
  'name': 'tests/foreman/ui/test_oscaptailoringfile.py::test_positive_end_to_end',
  'description': 'test_positive_end_to_end time: 175.295',
  'tags': ['High', 'foreman', 'scapplugin', 'jpathan'],
  'type': 'STEP',
  'start_time': 1586067675000,
  'end_time': 1586067850295,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:2b8619c303f58bc5cdd385e3578f0b5b'},
 {'id': '5e89838d149b3f0001091311',
  'name': 'tests/foreman/ui/test_user.py::test_positive_end_to_end',
  'description': 'test_positive_end_to_end time: 242.609',
  'tags': ['High', 'foreman', 'usersroles', 'pondrejk'],
  'type': 'STEP',
  'start_time': 1586068731000,
  'end_time': 1586068973609,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:a0b142034ba8a810381a0f1fca563c94'},
 {'id': '5e898336149b3f0001091161',
  'name': 'tests/foreman/cli/test_remoteexecution.py::TestAnsibleREX::test_positive_run_effective_user_job[rhel7]',
  'description': 'test_positive_run_effective_user_job[rhel7] time: 56.315',
  'tags': ['High', 'foreman', 'remoteexecution', 'pondrejk'],
  'type': 'STEP',
  'start_time': 1586070268568,
  'end_time': 1586070324884,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:fb8a761d97a75a7c5a53425df7767b45'},
 {'id': '5e898337149b3f0001091164',
  'name': 'tests/foreman/cli/test_remoteexecution.py::TestAnsibleREX::test_positive_run_reccuring_job[rhel7]',
  'description': 'test_positive_run_reccuring_job[rhel7] time: 0.001',
  'tags': ['High', 'foreman', 'remoteexecution', 'pondrejk'],
  'type': 'STEP',
  'start_time': 1586070268717,
  'end_time': 1586070268718,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:b8f5891fd4c4c7054e2300f1db543cec'},
 {'id': '5e898337149b3f0001091166',
  'name': 'tests/foreman/cli/test_remoteexecution.py::TestAnsibleREX::test_positive_run_packages_and_services_job[rhel7]',
  'description': 'test_positive_run_packages_and_services_job[rhel7] time: 0.001',
  'tags': ['High', 'foreman', 'remoteexecution', 'pondrejk'],
  'type': 'STEP',
  'start_time': 1586070268864,
  'end_time': 1586070268865,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:db254617c0e6a15491866f12e6dde008'},
 {'id': '5e898346149b3f00010911ae',
  'name': 'tests/foreman/cli/test_remoteexecution.py::TestRemoteExecution::test_positive_run_default_job_template_multiple_hosts_by_ip[rhel7]',
  'description': 'test_positive_run_default_job_template_multiple_hosts_by_ip[rhel7] time: 56.320',
  'tags': ['High', 'foreman', 'remoteexecution', 'pondrejk'],
  'type': 'STEP',
  'start_time': 1586070274294,
  'end_time': 1586070330614,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:b4fbe943466e14ec4f9a377cb5b4f547'},
 {'id': '5e89838e149b3f0001091314',
  'name': 'tests/foreman/longrun/test_oscap.py::OpenScapTestCase::test_positive_oscap_run_with_tailoring_file_with_ansible',
  'description': 'test_positive_oscap_run_with_tailoring_file_with_ansible time: 312.707',
  'tags': ['foreman', 'Critical', 'scapplugin', 'jpathan'],
  'type': 'STEP',
  'start_time': 1586070299301,
  'end_time': 1586070612008,
  'status': 'FAILED',
  'statistics': {'executions': {'total': '1',
    'passed': '0',
    'failed': '1',
    'skipped': '0'},
   'defects': {'product_bug': {'total': 0, 'PB001': 0},
    'automation_bug': {'AB001': 0, 'total': 0},
    'system_issue': {'total': 0, 'SI001': 0},
    'to_investigate': {'total': 1, 'TI001': 1},
    'no_defect': {'ND001': 0, 'total': 0}}},
  'path_names': {},
  'issue': {'issue_type': 'TI001',
   'autoAnalyzed': False,
   'ignoreAnalyzer': False},
  'has_childs': False,
  'launchId': '5e8982e1149b3f0001090fd3',
  'uniqueId': 'auto:e5e5adf8a098d59129f7dd71b0dfd1f5'}]

```

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,204 @@
...
+            launches = sorted(data['content'], key=lambda x: x['start_time'], reverse=True)
+            return launches
+
+    def launch(self, latest):
+        """Returns a specific launch overview data
+
+        This does not includes each tests data in launch
+
+        :param int latest: The launch place in Report portal
+            e.g To get the latest launch in Report Portal, ser latest = 1,
```
omkarkhatavkar, line 93: typo 'ser' should be 'set'

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,204 @@
...
+            'system_issue': 'SI001',
+            'to_investigate': 'TI001',
+            'no_defect': 'ND001',
+        }
+        return _defect_types
+
+    def _params(self, **kwargs):
+        """Parameters for Test items API request
+
+        :param dict kwargs: The tests `defect_type` and/or testss `status`
```
omkarkhatavkar, line 131: s/testss/tests/g

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,204 @@
...
+        This does not includes each tests data in launch
+
+        :param int latest: The launch place in Report portal
+            e.g To get the latest launch in Report Portal, ser latest = 1,
+                For second latest, set latest = 2
+
+        :returns dict: The launch overview data of requested launch
+        """
+        latest = int(latest) if isinstance(latest, str) else latest
+        return self.launches[int(latest) - 1]
```
omkarkhatavkar, line 99: No need to convert it to `return self.launches[int(latest) - 1]` as you already added check

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,204 @@
...
+        resp = requests.get(
+            self.launch_url, headers=self.headers, params=self.params, verify=False
+        )
+        resp.raise_for_status()
+        data = resp.json()
+        if data['content']:
+            launches = sorted(data['content'], key=lambda x: x['start_time'], reverse=True)
+            return launches
+
+    def launch(self, latest):
```
omkarkhatavkar, line 87: Do you think adding value as latest=1 as default value make sense?

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,204 @@
...
+            self._tests(params=params, data=data, page=page)
+        return data
+
+    def tests(self, **kwargs):
+        """Returns tests data adjusted by keyword arguments
+
+        This is a main function that will be called to retrieve the tests data
+        of a particular test status or/and defect_type
+
+        :param dict kwargs: The tests `defect_type` and/or testss `status`
```
omkarkhatavkar, line 195: s/testss/tests/g

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,204 @@
...
+            params.insert(2, ('filter.eq.status', test_status))
+        return dict(params)
+
+    @retry(
+        stop=stop_after_attempt(4), wait=wait_fixed(10),
+    )
+    def _requester(self, params, page):
+        """The Test Items GET requester to fetch the data on a page
+
+        If any error and before Failing explicitly, it retries for 3 times with 10 seconds delay
```
omkarkhatavkar, line 157: retry times are 3? we are applying 4.

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,204 @@
...
+            each tests properties in a page
+        """
+        params['page.page'] = page
+        resp = requests.get(self.tests_url, headers=self.headers, params=params, verify=False)
+        resp.raise_for_status()
+        total_pages = resp.json()['page']['totalPages']
+        pagedata = resp.json().get('content')
+        return total_pages, pagedata
+
+    def _tests(self, params, data, page=1):
```
omkarkhatavkar, line 172: we could have also initialized data=[] with a default value?

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,204 @@
...
+        """The Recursive Function that roll over on all pages to get all tests data in an API request
+
+        :param dict params: The parameters dictionary for API request
+        :param list data: The Empty dict to be filled during recursion with all tests data
```
omkarkhatavkar, line 176: Empty dict or list?  as _requester returns tuple with int and list

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,204 @@
...
+        :param page: The initial page which will be changed to next during recursion
```
omkarkhatavkar, line 177: `: param int  page`

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        params = [
+            ('filter.eq.launch', self.launch['id']),
+            ('page.page', 1),
+            ('page.size', 50),
+            ('page.sort', 'start_time'),
+        ]
+        defect_type = kwargs.get('defect_type')
+        test_status = kwargs.get('status')
+        if defect_type and defect_type in self.defect_types:
```
omkarkhatavkar, line 145: if defect_type and defect_type in self.defect_types:
```suggestion
        if defect_type in self.defect_types:
```

robottelo/report_portal/portal.py
mirekdlugosz: That's usually a bad idea. See e.g. [this stackoverflow question](https://stackoverflow.com/q/1132941/3552063) for problems it may cause. Answers and comments have many links to further material, if you want to learn more.

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,204 @@
...
+
+    @property
+    def defect_types(self):
+        """The various defect types in Report Report Portal for tests
+
+        This holds the codes for all the defect types, the tests will be marked with.
+
+        :returns dict: The dict of varios defect types and their codes
+        """
+        _defect_types = {
```
rplevka, line 119: there are more types, especially around AB00*
you might want to add those

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,204 @@
...
+            e.g To get the latest launch in Report Portal, ser latest = 1,
+                For second latest, set latest = 2
+
+        :returns dict: The launch overview data of requested launch
+        """
+        latest = int(latest) if isinstance(latest, str) else latest
+        return self.launches[int(latest) - 1]
+
+
+class Tests(Launch):
```
rplevka, line 102: let's call this `TestItems`, to stick with the RP nomenclature.

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        LOGGER.debug(f"Fetching results for Page : {page}")
+        total_pages, pagedata = self._requester(params=params, page=page)
+        data += pagedata
+        if page < total_pages:
+            page += 1
+            self._tests(params=params, data=data, page=page)
+        return data
+
+    def tests(self, **kwargs):
```
rplevka, line 189: `test_items`?

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,204 @@
...
+        self.rp_url = settings.report_portal.rp_url
+        self.rp_project = settings.report_portal.rp_project
+        self.rp_api_key = settings.report_portal.rp_key
+
+    @property
+    def api_url(self):
+        """Super url of report portal to be used as a base in all child classes
+        :returns Base url for API request
+        """
+        _api_url = f'{self.rp_url}/api/v1/{self.rp_project}/'
```
rplevka, line 32: for a sake of future compatibility, I'd parametrize the API version (put it into properties file).
`.../api/{self.rp_api}/...`

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,204 @@
...
+    This holds the properties and function to interact with the Report Portal
+    """
+
+    def __init__(self):
+        """Configure the settings and initiate report portal properties"""
+        if not settings.configured:
+            settings.configure()
+        self.rp_url = settings.report_portal.rp_url
+        self.rp_project = settings.report_portal.rp_project
+        self.rp_api_key = settings.report_portal.rp_key
```
rplevka, line 25: I'd add `self.verify_ssl` attribute here and use its value with `requests` methods across the code

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        _params = {
+            'page.page': 1,
+            'page.size': 500,
+            'page.sort': 'start_time',
+        }
+        return _params
+
+    @property
+    def launches(self):
```
rplevka, line 70: `s/launches/get_launches/`

robottelo/report_portal/portal.py
rplevka, line 87: `s/launch/get_launch`
- anyway, is this method required?
You could easily use the `get_launches` method if you made it accept custom parameters.
Passing custom `page.sort` and `page.size` could easily give you access to what you want. Isn't this too overengineered?

robottelo/report_portal/portal.py
jyejare: @omkarkhatavkar Thats correct omkar, the Docstring of stop_after_attempt says:
```Stop when the previous attempt >= max_attempt```

File: robottelo/report_portal/portal.py (hunk 7 above)
Comment (jyejare) [reply robottelo/report_portal/portal.py:172]: What @mirekdlugosz said.

File: robottelo/report_portal/portal.py (hunk 11 above)
Comment (jyejare) [reply robottelo/report_portal/portal.py:119]: Added, Thanks for bringing that up.

File: robottelo/report_portal/portal.py (hunk 12 above)
Comment (jyejare) [reply robottelo/report_portal/portal.py:102]: I think its none other than same. Lets make it useful for Robottelo and with robottelo terms of tests :)

File: robottelo/report_portal/portal.py (hunk 4 above)
Comment (jyejare) [reply robottelo/report_portal/portal.py:87]: Probably it might look over-engineered, but it's just for simplicity.

The clear thought behind that is to differentiate get `all launches` and single `launch`. Later both will be populated with their submodules, e.g :

`Launches`: mode, names, owners, tags
`Launch`: report (probably less likely to implement because it just exports the results for a launch, but we might have more detail in the report in just getting the launch from launches)

Comment (jyejare) [reply robottelo/report_portal/portal.py:87]: Also. I dont like using these verbs along with nouns. The docstrings are good enough to understand what the def is doing.

File: robottelo/report_portal/portal.py (hunk 16 above)
Comment (jyejare) [reply robottelo/report_portal/portal.py:70]: I dont like using these verbs along with nouns. The docstrings are good enough to understand what the def is doing.

File: robottelo/report_portal/portal.py (hunk 15 above)
Comment (jyejare) [reply robottelo/report_portal/portal.py:25]: Here, I am initiating only the parameters which are dynamic using config file.

Also:
`req = requests.get(url='', parameters={}, verify_ssl=False)` looks much better than `req = requests.get(url='', parameters={}, verify_ssl=self.verify_ssl)`.

I don't want to unnecessarily set the variables.

File: robottelo/report_portal/portal.py (hunk 7 above)
Comment (omkarkhatavkar) [reply robottelo/report_portal/portal.py:172]: @mirekdlugosz thanks

File: robottelo/report_portal/portal.py (hunk 14 above)
Comment (jyejare) [reply robottelo/report_portal/portal.py:32]: Can precisely explain ?

I don't think we should have an component wise API from properties file. We can tweak/append the per component in their own classes. As I am doing it here for Launch and Tests class.

File: robottelo/report_portal/portal.py (hunk 13 above)
Comment (jyejare) [reply robottelo/report_portal/portal.py:189]: same comment as above for making it compatible with robottelo terms.

File: robottelo/report_portal/portal.py (hunk 14 above)
Comment (JacobCallahan) [reply robottelo/report_portal/portal.py:32]: this is something that might be more easily addressed after robottelo's config is revamped. we can table this for now.

**Comment by jyejare** [comment]
@rplevka Do you wanna discuss more on PR or should we just merge it ?

File: robottelo/report_portal/portal.py (hunk 15 above)
Comment (mshriver) [reply robottelo/report_portal/portal.py:25]: @rplevka valid suggestion for sure, but something we can easily iterate on with this design.

File: robottelo/report_portal/portal.py (hunk 14 above)
Comment (mshriver) [reply robottelo/report_portal/portal.py:32]: Agreed @JacobCallahan, and again something we can iterate on with this design.

File: robottelo/report_portal/portal.py [hunk 17]
```diff
@@ -0,0 +1,216 @@
...
+        self.rp_project = settings.report_portal.rp_project
+        self.rp_api_key = settings.report_portal.rp_key
+
+    @property
+    def api_url(self):
+        """Super url of report portal to be used as a base in all child classes
+        :returns Base url for API request
+        """
+        _api_url = f'{self.rp_url}/api/v1/{self.rp_project}/'
+        return _api_url
```
mshriver, line 33: Consider just one-lining this property and not naming the string. A lot of these properties can be cached as well, as I don't think project/url/api_key will change while requesting launches. I'd _almost_ suggest creating a singleton instance in this module for import, but I don't think it'll really save us anything.

robottelo/report_portal/portal.py
mshriver: I think the nouns are fine without the `get_` prefix, especially in the case of properties.

That being said, see my larger comment above about the general structure here - and whether this request should be a property or a function.

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,216 @@
...
+    @property
+    def headers(self):
+        """The headers for Report Requests.
+        :returns header for API request
+        """
+        _headers = {'Authorization': f'Bearer {self.rp_api_key}'}
+        return _headers
+
+
+class Launch(ReportPortal):
```
mshriver, line 44: I don't think subclassing the Launch and tests from each other is the best approach here.

This might get called over-engineering it, but I think a different approach would better represent how these entities are related.

In my view, `launches` should be a function on the `ReportPortal` instance, not a property of the `Launch` class itself.

A report portal has launches, those launches have tests - this suggests a composited structure to me, not inherited.

`ReportPortal` instance attributes are correct as they are, and the `Launch` instance attribute is correct, but their relationship is not accurately represented to me, in this structure.

I'm not suggesting we go as far as concerning ourselves with whether we are collecting against slightly stale data, and completely over-engineering the request for launch data, but that we re-structure these a bit.

- `ReportPortal.launches` should be returning a data structure of Launch instances (keyed on some launch metadata?)
- `ReportPortal.launch` should be returning a specific launch based on passed arg (version + snap number)
- `Launch.tests` should be returning `Test` class instances
- `Test.launch` should be an instance attribute passed to its constructor from the Launch instance that it belongs to
- `Launch.report_portal` should be an instance attribute passed to its constructor from the ReportPortal instance it belongs to

I think that the current `Tests` class is misleading, given the url instance attribute is for an item.

We can simply subclass the `Test` class from AttrDict, instantiating it with the processed JSON from the report portal response, so that plugins using this ReportPortal tool are able to get a list of test items, and attributize the metadata comparisons needed for test collection.

This creates a pattern where in the plugin caller, we're doing something that looks like:
```
rp = ReportPortal()
current_version = pretend_with_me_method_that_gets_current_test_version()
# current_version  is something like '6.7.0-20.0' for snap 20
# assume pytest plugin hook, we have list of collected tests
test_items = complete_collected_tests
already_executed_tests = rp.launch(current_version).tests  # dict keyed on test id
tests_to_rerun = []
tests_that_passed = []
for test in complete_collected_tests:
    if already_executed_tests.get(test.id):
        if already_executed_tests[test.id].status.overall == 'passed':
            tests_that_passed.append(test)
        else:
            tests_to_rerun.append(test)
```

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,216 @@
...
+    Report Portal launches or specific launch
+    """
+
+    def __init__(self):
+        """Initiate report portal launch properties"""
+        super(Launch, self).__init__()
+        self.launch_url = f'{self.api_url}/launch'
+
+    @property
+    def params(self):
```
mshriver, line 57: Why is this a property? These look like static class attributes or module constants to me.

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,216 @@
...
+        if data['content']:
+            launches = sorted(data['content'], key=lambda x: x['start_time'], reverse=True)
+            return launches
+
+    def launch(self, latest=1):
+        """Returns a specific launch overview data in Report Portal Project
+
+        This does not includes each tests data in launch
+
+        :param int latest: The launch place in Report portal
```
mshriver, line 92: I don't think this is adequate or specific enough for our needs.

We need to be able to request a launch that matches the version that is being tested against - ideally querying that version from the target satellite host.

Which in turn means this method needs to filter the launches based on metadata so that we can identify the correct launch.

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,216 @@
...
+                For second latest, set latest = 2
+
+        :param int launch: The launch position in Report portal
+        """
+        super(Tests, self).__init__()
+        self.tests_url = f'{self.api_url}/item'
+        self.launch = self.launch(latest=launch)
+
+    @property
+    def defect_types(self):
```
mshriver, line 118: Again this looks like a static class attribute or a module constant to me, not an instance property.

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,216 @@
...
+        """
+        params['page.page'] = page
+        resp = requests.get(self.tests_url, headers=self.headers, params=params, verify=False)
+        resp.raise_for_status()
+        total_pages = resp.json()['page']['totalPages']
+        pagedata = resp.json().get('content')
+        return total_pages, pagedata
+
+    def _tests(self, params, data, page=1):
+        """The Recursive Function that roll over on all pages to get all tests data in an API request
```
mshriver, line 185: Is there a benefit here to this recursing?

It appears that this could be implemented with a non-recursing while statement.

```
total_pages, data = self._requester(params=params, page=page)
while page < total_pages:
    page += 1
    _, pagedata = self._requester(params=params, page=page)
    data.extend(pagedata)  # I'm not a fan of adding non-integers with +
return data
```

robottelo/report_portal/portal.py
mshriver: Potato/Potato - we'll be using this within pytest hooks, where the tests are just called `items`. Some tangential benefit to the method here being named a bit differently.

robottelo/report_portal/portal.py
jyejare: Yeah, while working with the collecter function, I observed some of the drawbacks of this design. Need to rework. I shall update the PR soon.

robottelo/report_portal/portal.py
jyejare: Yeah, hence I didn't think of caching them not going to save much time.

robottelo/report_portal/portal.py
jyejare: The main intention behind subclassing them is to independently import those classes elsewhere. like:
```from robottelo.report_portal.portal import Tests``` 
where I need not to instantite Reportportal class. I can go independentaly.

Comment (jyejare) [reply robottelo/report_portal/portal.py:44]: @mshriver , 

I am collecting tests to rerun like this (using current RP implmentation):
```
launch_tests_ = Tests(launch=launch_position)
rerunable_defect_types = ['no_defect', 'to_investigate', 'saucelabs_issue', 'system_issue']
for dtype in eligible_defect_types:
     rerunable_tests += [test['name'] for test in launch_tests_.tests(status='FAILED', defect_type=dtype)]
return rerunable_tests
```

I am only collecting tests (with their name) to rerun, Not sure why you are collecting passed tests above.

File: robottelo/report_portal/portal.py (hunk 19 above)
Comment (jyejare) [reply robottelo/report_portal/portal.py:57]: Yeah, make sense. I was copying this property from report-portal-tools repo. Fixed now.

File: robottelo/report_portal/portal.py (hunk 21 above)
Comment (jyejare) [reply robottelo/report_portal/portal.py:118]: Updated.

File: robottelo/report_portal/portal.py (hunk 20 above)
Comment (jyejare) [reply robottelo/report_portal/portal.py:92]: Yeah! Again I could sense the same while working on test_collector method. Updating it.

**Comment by mirekdlugosz** [comment]
@jyejare @mshriver Is this something that we want to have merged before branching off 6.7.z, or can we merge it after branching and cherry-pick if required?

Looks like there are a lot of open comments, so I don't see it being done in next day or two...

**Comment by jyejare** [comment]
@mirekdlugosz Quickly working on it to see if i can merge. There is a separate PR for 'collection_modify' and hence shouldn't take that many efforts.

File: robottelo/report_portal/portal.py (hunk 18 above)
Comment (mshriver) [reply robottelo/report_portal/portal.py:44]: I'm not seeing the advantage or intent when importing a `Launch` or `Test` class that is subclassed, and where you would not be instantiating a `ReportPortal` instance. Can you elaborate there?  I'd assume you will always need/want the ReportPortal instance for the other classes to be relevant.

We discussed the keep/discard test item collection intent in the call today, but to summarize:

By tracking both tests to keep in the collection and tests to discard, we can report to pytest collected cases which were discarded, which is useful in assessing test collection for a session.  We can also hook to report these discard statistics.

I also made the initial comment before reading your method for test listing, which includes direct filtering of results - so my example is unnecessarily iterating over the results to filter based on overall status.

File: robottelo/report_portal/portal.py (hunk 22 above)
Comment (jyejare) [reply robottelo/report_portal/portal.py:185]: I am not hard on using Recursion though I wanted to save some lines.

Comment (jyejare) [reply robottelo/report_portal/portal.py:185]: Changing it to simple while loop.

**Comment by jyejare** [comment]
@mshriver  The New Changes are updated as you suggested. As discussed on GMeet, removed Test class altogether as we dont need `Launch.tests returning Test class instances`.

Plus now we can filter launches by launch_type (especially called 'name' in RP) that is: `regular launches` or `upgrade launches`.

Check results updated in the Description of this PR.

File: robottelo/report_portal/portal.py [hunk 23]
```diff
@@ -0,0 +1,222 @@
...
+        return f'{self.rp_url}/api/v1/{self.rp_project}/'
+
+    @property
+    def headers(self):
+        """The headers for Report Portal Requests.
+        :returns: header for API request
+        """
+        return {'Authorization': f'Bearer {self.rp_api_key}'}
+
+    def launches(self, sat_version=None, launch_type='satellite6', latest=200):
```
mshriver, line 42: Don't mean to nitpick here, but two things about the `latest` kwarg:

1. Default of 200 seems very high for our intended use, is there a reason to default to so many launches?
2. The naming seems off, as we use the `latest` convention more like a boolean, to indicate we're getting the 'latest' version of something (latest rhel, latest satellite snap, etc).  Something like `count` would be more appropriate here I think, with the docblock explaining the count is considered from the latest chronological launch working backwards.

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,222 @@
...
+        :param int latest: The latest number of launches dict to return,
+            default returns latest 200 snaps else the max available
+        :returns dict: The directory of launches keyed by satellite version
+        """
+        launch_types = ['satellite6', 'upgrades']
+        if launch_type not in launch_types:
+            raise ValueError(
+                f'Wrong Launch Type is given as {launch_type} should be one of {launch_types}'
+            )
+        params = {'page.page': 1, 'page.size': 500, 'page.sort': 'start_time'}
```
mshriver, line 61: What's the max supported page size? Could we just pass `latest` here as the page size, so we don't have to further slice the response at L75?

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,222 @@
...
+        resp = requests.get(
+            url=f'{self.api_url}/launch', headers=self.headers, params=params, verify=False
+        )
+        resp.raise_for_status()
+        data = resp.json()
+        launches_ = {}
+        if data['content']:
```
mshriver, line 68: Any concern/potential for `content` to not be set in the response, when the response has a positive return code?

Can avoid the KeyError here with `if data.get('content')`

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,222 @@
...
+        :param str launch_type: Either satellite6 or upgrades,
+            default returns only non-upgrade launches
+        :param int latest: The latest number of launches dict to return,
+            default returns latest 200 snaps else the max available
+        :returns dict: The directory of launches keyed by satellite version
+        """
+        launch_types = ['satellite6', 'upgrades']
+        if launch_type not in launch_types:
+            raise ValueError(
+                f'Wrong Launch Type is given as {launch_type} should be one of {launch_types}'
```
mshriver, line 59: Consider storing these as constants in the module.

That way the caller doesn't have to "guess" at the strings, they can use the one defined in the module.

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,222 @@
...
+            url=f'{self.api_url}/launch', headers=self.headers, params=params, verify=False
+        )
+        resp.raise_for_status()
+        data = resp.json()
+        launches_ = {}
+        if data['content']:
+            all_launches = sorted(
+                data['content'], key=lambda launch: launch['start_time'], reverse=True
+            )
+            typed_launches = list(
```
mshriver, line 72: I think this is being cast directly to a list because of slicing it in the `for` statement. The iterator could be used directly if we aren't slicing, as mentioned above.

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,222 @@
...
+                filter(lambda launch: launch['name'].lower() == launch_type, all_launches)
+            )
+            for launch_info in typed_launches[:latest]:
+                version_compiler = re.compile(r'([\d\.]+)[\.-](\d+\.\d|[A-Z]+)')
+                launch_name = list(filter(version_compiler.search, launch_info['tags']))[0]
```
mshriver, line 77: You can use `next()` on the iterator instead of casting it to a list and indexing.

robottelo/report_portal/portal.py
jyejare: Actually 200 was defined when we didn't have typed and versioned launches. It can be cut down max 30 now. Thanks for bringing that up. Good catch.

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,222 @@
...
+                    f'The given snap_version \'{snap_version}\' is not available '
+                    f'in satellite version \'{sat_version}\''
+                )
+        else:
+            # Get the latest launch of the given sat_version
+            snap_version = list(launches_.keys())[0]
+        return launches_[snap_version]
+
+
+class Launch:
```
mshriver, line 119: What's your intent for the use of this class?

I would think that the `ReportPortal.launches|launch` methods would be creating instances of this class, not just returning raw data structures.

robottelo/report_portal/portal.py
jyejare: Even if rc is 200, it retuns `"content": [],` depending on the params. But its not necessary, removing it.

robottelo/report_portal/portal.py
jyejare: Not specific, but from your last comment `Launch.report_portal should be an instance attribute passed to its constructor from the ReportPortal instance it belongs to and as per our discussion to remove only Test class, I kept this. :)

I really don't think we should go into returning objects now, it's not our necessity right now. But having that will make it smarter.

What would you suggest ?

robottelo/report_portal/portal.py
jyejare: Not sure about the alternative to slice the filter object. Do you think of any?

robottelo/report_portal/portal.py
jyejare: @mshriver Also if we return Launch objects for launches and launch functions in ReportPortal, we will end up with maximum depth recursssion error and unless we create a separate functions that will return objects of launch. I also beleive that this is significantly add up to the time consumption.

robottelo/report_portal/portal.py
mshriver: gross, cool, thanks!

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,221 @@
...
+
+        :param ReportPortal rp: An object of a Report Portal Class
+        :param str sat_version: The satellite version
+        :param str snap_version: The snap version of a given satellite version
+            if None, the latest launch data of a given sat_version is considered
+        :param str launch_type: Either satellite6 or upgrades, default returns only
+            non-upgrade launches
+        """
+        self.report_portal = rp
+        self.defect_types = {
```
mshriver, line 131: This appears static, it can be a class attribute, not an instance attribute.  I think its also an attribute/property of ReportPortal, not of a launch. Its the same for all launches on a ReportPortal instance.

mshriver: Its also being redefined in #7758 in constants. Its relative to ReportPortal, I think it should be a class attribute there.

robottelo/report_portal/portal.py
mshriver: The recursion is only because you're trying to set a launch property on a launch class to the result of the function that should be creating the instance in the first place.

This `Launch` class should model a specific launch instance from Report Portal. It should not have a `launch` property on it, it IS a launch. This is seemingly trying to be a class that's an interface for a launch, but ReportPortal instance is our interface.

In `ReportPortal.launches` you have all of the info that should define a launch instance in report portal, and you return it as a dictionary within a list.  That should instead be taking all of that launch info, and returning a list of `Launch` instances.

A `Launch` should have attributes that uniquely define it, as given by the named `launch_info` variable in `ReportPortal.launches` - the name, sat version, and snap version.

The launch statistics could also be defined as a launch instance attribute, since they're available at init.

This is creating a needlessly layered pattern within your pytest fixture in #7758.  Its also assuming there's only one launch for a given sat version, which is not a safe assumption.

With what I'm suggesting, you should be able to reference the 'info' for a launch as instance attributes, and you shouldn't have to even directly instantiate a launch.

It would allow for the following pattern
```
rp = ReportPortal()
matching_launches = rp.Launches(sat_version=current_sat_version_under_test)
# maybe log that there were multiple launches, maybe composite them!
last_launch = matching_launches.pop()
# if statistics is an instance attribute, a dictionary:
last_pass_percent = round(100*(last_launch.statistics['passed']/last_launch.statistics['total']))
last_failed_tests = last_launch.tests(defect_type='to_investigate')
```

jyejare: ahh, Now I got it right! Or May be just back from summit to my regular work :)

Implemented !

robottelo/report_portal/portal.py
jyejare: Because it's more specific to launch tests I thought to keep it here. But still, I disagree with making it ReportPortal attributes. Let's say we cannot filter launches of Report portal with defect types but tests of launches can be.

The reason I wanted to keep it in constants as I am using them in rerun plugin in #7758 and that is even before RP class initialization.

If we don't want to define it as constants then its better to keep them as RP class attributes so that we can quickly fail in options sanitization in #7758 and that is even before intiating launches objects.

So doing it !

jyejare:
@mshriver : Some improvements made along with launches returning Launch objects:

- latest : `Is now removed, I don't find it very useful. Anyways we can slice on launches to get the latest.`
- sat_version, snap_version calculation: `Is now moved to Launch class as its specific to launch.`
- defect_types, statuses: `Both are now RP class attributes`
- _launch_requester: `To separate the clutter from launches`
- _format_launches: `That formats the launches keyed by sat_version and snap_version, ofcourse to remove clutter from launches function`
- Launch class only accepts launch_info now: `Removed sat_version and snap_version params which are now being calculated inside Launch class itself.`
- _versions : 'The def that sets sat_version and snap_version from Launch class itself instead of RP class'
- Better argument handling in tests def: `for defect_type and status params`
- tests now dont have **kwargs: `Direct defect_type and status params are set, only 2 params so removed **kwargs`

jyejare:
@mshriver The Description of PR is nor updated with the latest way of using it. :)

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,257 @@
...
+        :param ReportPortal rp: A report portal object
+        :param dict launch_info: The dict information about a launch
+        """
+        self.report_portal = rp
+        self.info = launch_info
+        self.name = self.info['name']
+        self.statistics = self.info['statistics']['executions']
+        self._versions(self.info)
+
+    def _versions(self, launch_info):
```
mshriver, line 168: Not a blocker - but `launch_info` doesn't have to be passed, it can be accessed with `self.info`

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,257 @@
...
+        sorted_launches = sorted(
+            data['content'], key=lambda launch: launch['start_time'], reverse=True
+        )
+        typed_launches = filter(
+            lambda launch: launch['name'].lower() == launch_type, sorted_launches
+        )
+        launches_ = self._format_launches(typed_launches)
+        if sat_version:
+            if sat_version not in launches_:
+                raise ValueError(
```
mshriver, line 121: I'm not sure this should raise a ValueError, instead of just returning an empty dict and just log/print this message instead.

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,257 @@
...
+        pagedata = resp.json().get('content')
+        return total_pages, pagedata
+
+    def tests(self, status=None, defect_type=None):
+        """Returns tests data customized by kwargs parameters.
+
+        This is a main function that will be called to retrieve the tests data
+        of a particular test status or/and defect_type
+
+        :param str status: Filter tests of a launch with tests `status`
```
mshriver, line 243: Can you provide some example or define some constants to be used here?  Its not clear if status could be `Failed`, `Fail`, `fail`, `failed`, `error`, etc.

robottelo/report_portal/portal.py
```diff
@@ -0,0 +1,257 @@
...
+            tests properties
+        """
+        page = 1
+        params = self._test_params(status, defect_type)
+        total_pages, data = self._test_requester(params=params, page=page)
+        while page < total_pages:
+            page += 1
+            _, pagedata = self._test_requester(params=params, page=page)
+            data.extend(pagedata)
+        # returning only test names
```
mshriver, line 255: Optional:

I know within the collection we only need the test names when we pass a status, but I think this function would be far more versatile if it returns more data for the tests. That way the caller can apply whatever  additional logic that they want.

We could generate a dictionary, keyed on the test name, so the caller can use `keys()`.

robottelo/report_portal/portal.py
jyejare: Make Sense, wonder how I missed :)

robottelo/report_portal/portal.py
jyejare: Raising error because sat_version is a user input to launches def.

Lets user enter 6.7 instead of 6.7.0, so if we don't raise an error in this situation and will return only empty dict, the user might think there is no data for 6.7.
If user knows there is a data for 6.7.0, he/she might end up diagnosing what's wrong with the module :)

'sat_version' is an optional parameter and if we dont provide a value it that would give the user a chance to see all the correct satellite version names as `launches` returns all sat versions  data.

robottelo/report_portal/portal.py
jyejare: Yes, same as `defect_types` class attribute of RP, the `status` attribute is also defined for Report Portal.

robottelo/report_portal/portal.py
jyejare: True! And earlier I was returning all the data for a test. But it was our(me&U) meeting which decides to return only test names.

Never mind I can improve in that regards.

PR #7725, opened 2020-04-08
Added settings "Sendmail" test case to verify the "Test Email" functionality. 

devendra104:
Airgun issue.
https://github.com/SatelliteQE/airgun/issues/476

### DIFF
--- a/tests/foreman/ui/test_settings.py
+++ b/tests/foreman/ui/test_settings.py
@@ -20,6 +20,7 @@ from fauxfactory import gen_url
 from nailgun import entities
 from pytest import raises
 
+from robottelo import ssh
 from robottelo.cli.user import User
 from robottelo.datafactory import filtered_datapoint
 from robottelo.datafactory import gen_string
@@ -342,9 +343,8 @@ def test_negative_update_email_delivery_method_smtp():
     """
 
 
-@stubbed()
 @tier3
-def test_positive_update_email_delivery_method_sendmail():
+def test_positive_update_email_delivery_method_sendmail(session):
     """Updating Sendmail params on Email tab
 
     :id: c774e713-9640-402d-8987-c3509e918eb6
@@ -367,9 +367,39 @@ def test_positive_update_email_delivery_method_sendmail():
     :CaseImportance: Critical
 
     :CaseLevel: Acceptance
-
-    :CaseAutomation: notautomated
     """
+    property_name = "Email"
+    mail_config_default_param = {
+        "delivery_method": "",
+        "email_reply_address": "",
+        "email_subject_prefix": "",
+        "sendmail_arguments": "",
+        "sendmail_location": "",
+        "send_welcome_email": "",
+    }
+    mail_config_default_param = {
+        content: entities.Setting().search(query={'search': f'name={content}'})[0]
+        for content in mail_config_default_param
+    }
+    mail_config_new_params = {
+        "delivery_method": "Sendmail",
+        "email_reply_address": f"root@{ssh.settings.server.hostname}",
+        "email_subject_prefix": [gen_string('alpha')],
+        "sendmail_location": "/usr/sbin/sendmail",
+        "send_welcome_email": "Yes",
+    }
+    command = "grep " + f'{mail_config_new_params["email_subject_prefix"]}' + " /var/mail/root"
+
+    with session:
+        try:
+            for mail_content, mail_content_value in mail_config_new_params.items():
+                session.settings.update(mail_content, mail_content_value)
+            test_mail_response = session.settings.send_test_mail(property_name)[0]
+            assert test_mail_response == "Email was sent successfully"
+            assert ssh.command(command).return_code == 0
+        finally:
+            for key, value in mail_config_default_param.items():
+                setting_cleanup(setting_name=key, setting_value=value.value)
 
 
 @stubbed()


tests/foreman/ui/test_settings.py
```diff
@@ -353,9 +353,44 @@ def test_positive_update_email_delivery_method_sendmail():
...
+
+    with session:
+        try:
+            for mail_content in mail_config_new_params:
+                session.settings.update(
+                    "{}".format(mail_content), "{}".format(mail_config_new_params[mail_content])
+                )
+            contents = session.settings.contents()
+            contents.email_button.click()
+            assert contents.notification_locator.text == "Email was sent successfully"
```
mirekdlugosz, line 387: Do we control where this email is sent? Could we set it up to send to local mailbox, and then SSH into box and confirm that it was received?

tests/foreman/ui/test_settings.py
```diff
@@ -353,9 +353,44 @@ def test_positive_update_email_delivery_method_sendmail():
...
+        "sendmail_location": "/usr/sbin/sendmail",
+        "send_welcome_email": "Yes",
+    }
+
+    with session:
+        try:
+            for mail_content in mail_config_new_params:
+                session.settings.update(
+                    "{}".format(mail_content), "{}".format(mail_config_new_params[mail_content])
+                )
```
mirekdlugosz, line 384: I presume there is no API/CLI way to send this email, so we need UI for that. But do we have to set all options using UI? I believe it would be much faster if we used API instead, and particular way we set config is probably not important overall.

tests/foreman/ui/test_settings.py
```diff
@@ -353,9 +353,44 @@ def test_positive_update_email_delivery_method_sendmail():
...
+            contents = session.settings.contents()
+            contents.email_button.click()
+            assert contents.notification_locator.text == "Email was sent successfully"
+        finally:
+            for mail_content in mail_config_default_param:
+                setting_cleanup(
+                    setting_name=mail_content,
+                    setting_value=mail_config_default_param[mail_content],
+                )
```
mirekdlugosz, line 393: Please use `items()` to iterate over key and value at once:
```suggestion
            for key, value in mail_config_default_param.items():
                setting_cleanup(setting_name=key, setting_value=value)
```

devendra104:
> You mentioned this depends on Airgun change, but link is to issue, not PR. Just so we are on the same page: you will work on Airgun change and send PR, or you expect one of more experienced people to change Airgun?
> 
> Either way, I think this needs DO NOT MERGE label, as in current form this test has no way of passing.

Raised PR#https://github.com/SatelliteQE/airgun/pull/477

tests/foreman/ui/test_settings.py
```diff
@@ -353,9 +353,44 @@ def test_positive_update_email_delivery_method_sendmail():
...
+        "email_subject_prefix": "",
+        "sendmail_arguments": "",
+        "sendmail_location": "",
+        "send_welcome_email": "",
+    }
+    for content in mail_config_default_param:
+        mail_config_default_param[content] = [
+            ele.value
+            for ele in entities.Setting().search(query={'search': 'name={}'.format(content)})
+        ][0]
```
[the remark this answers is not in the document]
devendra104: Yes, but faced serialization related issues(in tear down) hence, I picked the f-strings related changes only and kept the rest of things in their original state.

tests/foreman/ui/test_settings.py
devendra104: Fixed.

tests/foreman/ui/test_settings.py
devendra104: The intent behind this to check the email contents are updated or not through UI, Even though we can achieve the same from CLI. The complete update operation from UI would take 9 to 10 seconds(sometimes we can not quantify the automation performance over quality). And this is the only test case in UI(Settings, sendmail/smtp) where we are doing email related update, I still have some test cases related to Sendemail/smtp, I will prefer to use API there.

tests/foreman/ui/test_settings.py
devendra104: Fixed.

devendra104:
Test Result:
```
$ pytest  tests/foreman/ui/test_settings.py::test_positive_update_email_delivery_method_sendmail
platform linux -- Python 3.6.6, pytest-4.6.3, py-1.8.1, pluggy-0.13.1                                                                                                                                                      
===================================================================================== 1 passed in 103.46 seconds ======================================================================================
```

tests/foreman/ui/test_settings.py
mirekdlugosz: Personally I don't see much value in adding coverage for different settings, but if it takes only ~10 seconds, then there's also no reason to not use UI.

tests/foreman/ui/test_settings.py
```diff
@@ -353,9 +353,46 @@ def test_positive_update_email_delivery_method_sendmail():
...
+        "email_subject_prefix": [gen_string('alpha')],
+        "sendmail_location": "/usr/sbin/sendmail",
+        "send_welcome_email": "Yes",
+    }
+
+    with session:
+        try:
+            for mail_content in mail_config_new_params:
+                session.settings.update(
+                    "{}".format(mail_content), "{}".format(mail_config_new_params[mail_content])
```
mirekdlugosz, line 383: Haven't noticed that before:
1. Useless use of format. `mail_content` is string.
2. Iterating over keys and looking up value on each iteration is not *wrong*, but isn't very pythonic, either. Better to use `.items()` on dict and iterating over key/value pair:
```python
for mail_content, mail_content_value in mail_config_new_params.items():
    session.settings.update(mail_content, mail_content_value)  # is value a string? If not, we might need this .format() anyway
```

tests/foreman/ui/test_settings.py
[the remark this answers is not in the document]
mirekdlugosz: Could you elaborate? What was the error message?
Review is good time to catch problems like that and fix them.

tests/foreman/ui/test_settings.py
```diff
@@ -353,9 +353,43 @@ def test_positive_update_email_delivery_method_sendmail():
...
+        "sendmail_location": "",
+        "send_welcome_email": "",
+    }
+    mail_config_default_param = {
+        content: entities.Setting().search(query={'search': f'name={content}'})[0]
+        for content in mail_config_default_param
+    }
+    mail_config_new_params = {
+        "delivery_method": "Sendmail",
+        "email_reply_address": "root@{}".format(ssh.settings.server.hostname),
```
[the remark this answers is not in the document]
devendra104: Fixed.

devendra104:
Test Result:

```
$ pytest tests/foreman/ui/test_settings.py::test_positive_update_email_delivery_method_sendmail
platform linux -- Python 3.6.6, pytest-4.6.3, py-1.8.1, pluggy-0.13.1

===================================================================================== 1 passed 114.74 seconds ======================================================================================
```

tests/foreman/ui/test_settings.py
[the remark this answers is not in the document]
devendra104: It is working now, I fixed it by using dict property.

tests/foreman/ui/test_settings.py
```diff
@@ -353,9 +353,43 @@ def test_positive_update_email_delivery_method_sendmail():
...
+        try:
+            for mail_content, mail_content_value in mail_config_new_params.items():
+                session.settings.update(mail_content, mail_content_value)
+            assert session.settings.send_test_mail()[0] == "Email was sent successfully"
+            assert (
+                ssh.command(
+                    "grep {} /var/mail/root".format(mail_config_new_params["email_subject_prefix"])
+                ).return_code
+                == 0
+            )
```
mirekdlugosz, line 387: Not blocking, but this hanging comparison makes me uncomfortable. Perhaps break it down to multiple lines:
```python
            cmd = "grep {} /var/mail/root".format(mail_config_new_params["email_subject_prefix"])
            assert ssh.command(cmd).return_code == 0
```
In fact, this way it takes fewer lines of code than your current approach.

mirekdlugosz:
I created 6.7.z branch today. Unfortunately, this PR did not make it due to lack of review.

If you want this to run on 6.7.z as well, please raise cherry-pick PR against 6.7.z branch after this one is merged into master.

PR #7739, opened 2020-04-17
Cherry pick of PR #7725 #7710 in 6.7.z

devendra104:


mirekdlugosz:
Let's not cherry-pick until the thing we are cherry-picking is actually merged.
If changes to #7725 are requested and done, the same changes will have to be done here. Which is additional burden on you.
If I ACK now and then you commit changes to #7725, my ACK will be invalidated, but GitHub will not show that in any way. Which is additional burden on me and other reviewers.

Once #7725 is merged, feel free to remove this label and add comment reminding us to do review.

PR #7741, opened 2020-04-20
we used capsule repo for katello-agent installation

tstrych:
We installed katello host tools, with `Red Hat Satellite Capsule 6.6 (for RHEL 7 Server) (RPMs) ` repository. 

I change it to `Red Hat Satellite Tools 6.6 (for RHEL 7 Server) (RPMs)`.

test results for `HostSubscriptionTestCase`
```
============= 2 failed, 7 passed, 54 deselected in 1270.51 seconds =============
```

Failed tests are `test_negative_without_attach` and `test_negative_without_attach_with_lce`
These tests fails from different reasons.


tests/foreman/cli/test_host.py
```diff
@@ -2088,9 +2088,9 @@ def setUpClass(cls):
         # create a rh capsule content
```
jameerpathan111, line 2088:
```suggestion
        # setup rh satellite tools repository content
```

jameerpathan111: you can change comment to something like this

tests/foreman/cli/test_host.py
```diff
@@ -2088,9 +2088,9 @@ def setUpClass(cls):
...
         setup_org_for_a_rh_repo(
             {
-                'product': PRDS['rhsc'],
-                'repository-set': REPOSET['rhsc7'],
-                'repository': REPOS['rhsc7']['name'],
+                'product': PRDS['rhel'],
+                'repository-set': REPOSET['rhst7_66'],
```
jameerpathan111, line 2092: @tstrych should we use `Satellite Tools 6.7` in our tests instead of 6.6?

tstrych: @jameerpathan111 , not sure if satellite tools 6.7 are published, if it will work, but yes I can add them also to constants :)

tstrych:
```
=============== 2 failed, 7 passed, 54 deselected in 1257.45 seconds ================
```

tstrych:
I rather change only satellite tools as I am not sure about capsule, for example I was unable to find 6.7 capsule for rhel 6, I don't know if it is supported. But I think email about change of constant file should be send.

tests/foreman/cli/test_host.py
rplevka: @tstrych that's wrong. Even if they are not published, we're supposed to use the tested one - this would mean, using the one from the dogfood (defined as custom repository)

tstrych:
Test results:
`============= 2 failed, 7 passed, 54 deselected in 1311.64 seconds =============`

mirekdlugosz:
@rplevka comment ( https://github.com/SatelliteQE/robottelo/pull/7741#discussion_r411657691 ) looked kind of important... was that discussed outside of GitHub and properly resolved before merge?

tests/foreman/cli/test_host.py
swadeley: Hello

I think @rplevka means that in the time before the tools repo is published for the next release, which for some reason always happens late in the development cycle, tests related to packages in the tools repo should use the tools repo from dogfooding server.

I took a look to see where we pull from dogfood. I found two places which use 'settings.sattools_repo' from robottelo.properties file:

https://github.com/SatelliteQE/robottelo/blob/8907e0257eb0dd978624db7f87e647f026c76aa3/robottelo/cli/factory.py#L3419

https://github.com/SatelliteQE/robottelo/blob/8907e0257eb0dd978624db7f87e647f026c76aa3/robottelo/products.py#L647

swadeley:
Hello

Should this be cherry picked to 6.7 branch?

tstrych:
I ping @rplevka , about that, but probably the message was lost in the chat, I will ask him again. 

@swadeley  probably should.

tstrych:
@mirekdlugosz, @swadeley, @jameerpathan111 

I talked with @rplevka ,
He has a good point. We should run tests in 6.8 against 6.8 satelllite tools repository. 
But I used here constants which are 6.7, (from cdn, stable repo). Problem is when we will run 6.8 automation it will goes with 6.7 sattools. But it should run with 6.8 sattools. 

We should use dogfood repo from robotello properties called `sattools_repo` instead of  repositories in `constants` which we are using. I didn't know about that, and as you merge this you probably don't know that too.

@rplevka agreed it should be raised at automation meeting next week, to inform and discuss what is the right  repository test design, and what is the right approach for z-streams.

tstrych:
Till then I am not going to push this into z-stream.

swadeley:
Hello @tstrych 

Firstly, my apologies for merging too soon.

Your changes and additions were essentially correct. So now the question is, should  'rhst7' never have been set to 'Red Hat Satellite Tools 6.6' in the past, or tests should not be using that? I think a review of tests that use 'rhst7' would help make that question clear: Is the purpose of a test to test the packages in the tools repo, in which case 'settings.sattools_repo', or just to prove you can subscribe to a genuine RH repo? Because the rhel-7-server-satellite-tools repo is a small but genuine RH repo it is useful for some tests that just need a RH repo.

tstrych:
Yes depends what you are trying to achieve in the test. But review of tests  is not necessarily enough. Because people who wrote these tests maybe do not know the difference, or better said they didn't know about the other option. That's why we should discuss this. Ideally everyone after discussion should be aware of these two options and is able to choose right option for his new tests. Respectively for fixing the existing tests.

Issue #7734, opened 2020-04-16
Upgrading from 6.6 to 6.7 breaks the Tasks index page


jyejare:
#Customer Case

- [ ] Automate issue in upgrade Scenario
-OR-
- [ ] Increase coverage around tasks in upgrade tier tests, if not already

pondrejk:
I assume the issue is limited to UI, right? We heave one test that goes 
into to the tasks page, though it's kind of hidden in dashboard test class (test_positive_task_status), I'm going to mark it for upgrade inclusion, though I found some other fixes are needed as the test was failing in previous runs (I'm not the component owner).

If also cli tests would make sense to test this, we could include also tests/foreman/cli/test_remoteexecution.py test_positive_run_default_job_template_by_ip that lists tasks via cli

vijay8451:
@pondrejk thanks to adjusting the automation accordingly , Yeah it's seen on UI only ..

JacobCallahan:
Is this issue still open or has it been resolved?

PR #7744, opened 2020-04-21
dashboard task status fix

pondrejk:
fixing two tests against changed responses from server, test_positive_task_status added to upgrade set, context here https://github.com/SatelliteQE/robottelo/issues/7734
requires https://github.com/SatelliteQE/airgun/pull/480
needs cherrypick to 6.7

```
pytest tests/foreman/ui/test_dashboard.py -k test_positive_host_configuration_chart

tests/foreman/ui/test_dashboard.py .                                                                           [100%]

====================================== 1 passed, 3 deselected in 74.61 seconds =======================================
```
```
pytest tests/foreman/ui/test_dashboard.py -k test_positive_task_status

tests/foreman/ui/test_dashboard.py . [100%]

=================== 1 passed, 3 deselected in 48.67 seconds ====================
```

### DIFF
--- a/tests/foreman/ui/test_dashboard.py
+++ b/tests/foreman/ui/test_dashboard.py
@@ -131,14 +131,15 @@ def test_positive_host_configuration_chart(session):
         session.organization.select(org_name=org.name)
         session.location.select(loc_name=loc.name)
         dashboard_values = session.dashboard.read('HostConfigurationChart')
-        assert dashboard_values['chart']['No report'] == '100%'
+        assert dashboard_values['chart'][''] == '100%'
 
 
+@upgrade
 @run_in_one_thread
 @tier2
 def test_positive_task_status(session):
     """Check if the Task Status is working in the Dashboard UI and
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_dashboard.py
```diff
@@ -131,14 +131,15 @@ def test_positive_host_configuration_chart(session):
         session.organization.select(org_name=org.name)
         session.location.select(loc_name=loc.name)
         dashboard_values = session.dashboard.read('HostConfigurationChart')
-        assert dashboard_values['chart']['No report'] == '100%'
+        assert dashboard_values['chart'][''] == '100%'
```
omkarkhatavkar, line 134: this is weird we are not having the key with associated. 
`assert '100%' in test['chart'].values()`  can't we do like this to avoid the dependency on the key? what're your thoughts?

mirekdlugosz: Agreed with Omkar, this is puzzling. But not blocking.

PR #7751, opened 2020-04-23
Receptor upgrade

lhellebr:


tests/foreman/cli/test_remoteexecution.py
```diff
@@ -516,6 +517,12 @@ def test_positive_run_receptor_installer(self):
         )
         assert 'project-receptor.satellite_receptor_installer' in result
         assert 'Exit status: 0' in result
+        # check that there is one receptor conf file and it's only readable by the receptor user and root
+        result = ssh.command(
+            'ls -l /etc/receptor/*/receptor.conf | grep -- \'-r--------\\. 1 foreman-proxy\' && ls -l /etc/receptor/*/receptor.conf | wc -l | grep \'^1$\'',
```
mirekdlugosz, line 522: This line is too long, Travis complains about it.

For permissions and ownership, it's better to use `stats` instead of `ls`:
```bash
$ stat --format '%a:%U' /etc/fstab 
644:root
```

How expensive is ssh.command, anyway? Could we break it down to two commands and assert on their output?
There are some helpers in robottelo to re-use SSH session, might help speeding things up if time is a concern.

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -516,6 +517,12 @@ def test_positive_run_receptor_installer(self):
         )
         assert 'project-receptor.satellite_receptor_installer' in result
         assert 'Exit status: 0' in result
+        # check that there is one receptor conf file and it's only readable
+        # by the receptor user and root
+        result = ssh.command('stat /etc/receptor/*/receptor.conf --format "%a:%U"')
+        assert result.stdout[0] == '400:foreman-proxy'
```
mirekdlugosz, line 523: Now I think about this, if `*` wildcard expands to more than one dir, you will get multiple lines in output, so there would be no way for this assertion to pass. In other words, this assert implicitly contains the next one.

But "Explicit is better than implicit.", so that's OK.

PR #7756, opened 2020-04-27
Report Portal Failure: Capsule syncing issue

latran:
This failure has been ongoing and dev said it's slated to be fixed in 7.0.  For the time being, I'm not sure what to do with this test case.  I can either have this small commit which allows the test to pass or stub it out until 7.0.  Please check my email regarding the details on this test case.

### DIFF
--- a/tests/foreman/cli/test_contentview.py
+++ b/tests/foreman/cli/test_contentview.py
@@ -4156,7 +4156,7 @@ class ContentViewTestCase(CLITestCase):
                     }
                 )
             # Synchronize the capsule content
-            Capsule.content_synchronize({'id': capsule['id'], 'organization-id': org['id']})
+            # Capsule.content_synchronize({'id': capsule['id'], 'organization-id': org['id']})
             capsule_content_info = Capsule.content_info(
                 {'id': capsule['id'], 'organization-id': org['id']}
             )


mshriver:
@latran I'm going to set this PR to draft state since it seems to be forgotten/stale. Please check out @mirekdlugosz comment about updating the sync method to only sync if one is not in progress (and wait for result), which would be a great pattern to have for test safety.

JacobCallahan:
@latran from our discussion earlier, it would be best to change the content_synchronize behavior to check to see if a sync is already in progress. A user should also be able to override this behavior to force a sync regardless of whether a sync is ongoing or not. 
There should be a way to check this via the CLI (maybe https://github.com/SatelliteQE/robottelo/blob/61f527060a02dd6e0a31c4ff8c5f4d6a2506c3e8/robottelo/cli/capsule.py#L84) but if not, the check could be performed via API.

omkarkhatavkar:

tests/foreman/cli/test_contentview.py
```diff
@@ -4156,7 +4156,7 @@ def test_positive_remove_cv_version_from_multi_env_capsule_scenario(self):
                     }
                 )
             # Synchronize the capsule content
-            Capsule.content_synchronize({'id': capsule['id'], 'organization-id': org['id']})
+            # Capsule.content_synchronize({'id': capsule['id'], 'organization-id': org['id']})
```
mshriver, line 4159: Why not base whether this function is called on the BZ state associated with the failure that's projected to be fixed in a future version?

mshriver: Also, was there additional conversation about the suggestions from @JacobCallahan and @mirekdlugosz to actually update the `content_synchronize` method to base its actions on if a sync is already in progress?

latran: Conversation was already discussed with @JacobCallahan about this.  I will look into it and work with what was suggested.  I will update this PR once the pytest conversion PR is merged.

JacobCallahan: I think I just merged that pr

latran:
The `CapsuleVirtualMachine` class will be replaced and Jake is working on its replacement.  Till then, this is blocked.

Issue #7762, opened 2020-04-30
Intermitant failures to install packages on test VMs


swadeley:
I have many cases of tests failing during the setup stage because a package was not available. The logs in report portal just showed a connection to the test VM followed by tare down. Running tests locally for many days I noticed tests failing with "No package xxx available."

I suspected it was due to download policy "on demand", and being the unlucky first VM to ask for a package and yum timing out while waiting for the repo to pull it from upstream. So I added some extra steps to log the attempts and try three times.

However, it happened yesterday that out of two VMs as clients, the first one installed the package but the second one failed three times with "No package walrus-0.71-1.noarch available.". Logging into the VM to check logs, I noticed chronyd jumped the time just as the download was due to be attempted (based on log time stamps).

As some are against "debug code" in master branch, I will make a PR with just one line to set the time in 'robottelo/helpers.py' .

Thank you

Update: I have added more than one line, one command, but more than one line as I decided to change the code to use the ssh.get_connection session manager construct to save some time on setting up SSH connections.

swadeley:
Hello

I think this update will stop the time jumps:
 set UTC time to avoid time jumps #831

mirekdlugosz:
I'm pretty sure you meant https://github.com/SatelliteQE/automation-tools/pull/831 , as robottelo #831 was merged 6 years ago and is completely unrelated to VM clock.

swadeley:
> I'm pretty sure you meant [SatelliteQE/automation-tools#831](https://github.com/SatelliteQE/automation-tools/pull/831) , as robottelo #831 was merged 6 years ago and is completely unrelated to VM clock.

oh dear oh dear, I did check the preview, but failed to comprehend. I will try edit it.

thank you

swadeley:
Hello

New images with fix as per [1] were brought into use on 14 May 2020.

[1] https://github.com/SatelliteQE/automation-tools/pull/831

PR #7764, opened 2020-05-03
Add ntpdate command to avoid clock jumps (#7762)

swadeley:
Setting the time in the first few seconds of starting a VM will avoid chronyd jumping the clock some seconds later which can clash with yum transactions.


swadeley:
This does not work well. Consulted crony dev (mlichvar):

"Changing the clock with ntpdate while chronyd is running causes confusion. chronyd
reports there is an unexpected backward step."

Closing this and will try to fix in repo for snap guest.

PR #7767, opened 2020-05-08
Automation for stubbed test "test_negative_update_hostname_with_empty_fact"

akhil-jha:
Signed-off-by: Akhil Jha

### DIFF
--- a/tests/foreman/api/test_setting.py
+++ b/tests/foreman/api/test_setting.py
@@ -99,7 +99,6 @@ class SettingTestCase(APITestCase):
                 login = login.update(['value'])
                 self.assertEqual(login.value, login_text)
 
-    @stubbed()
     @tier2
     def test_negative_update_hostname_with_empty_fact(self):
         """Update the Hostname_facts settings without any string(empty values)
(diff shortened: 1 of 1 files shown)


tests/foreman/api/test_setting.py
```diff
@@ -108,9 +107,21 @@ def test_negative_update_hostname_with_empty_fact(self):
...
             hostname_facts setting
-
-        :CaseAutomation: notautomated
         """
+        hostname_facts_id = [
+            ele.id
+            for ele in entities.Setting().search(
+                query={"per_page": 200, 'search': 'name="discovery_hostname"'}
+            )
+        ][0]
```
mirekdlugosz, line 116: If you only going to use first value, you might as well not iterate over all of them:
```suggestion
        hostname_facts_id = entities.Setting().search(
            query={"per_page": 200, 'search': 'name="discovery_hostname"'}
        )[0].id
```

tests/foreman/api/test_setting.py
```diff
@@ -108,9 +107,21 @@ def test_negative_update_hostname_with_empty_fact(self):
...
+        facts = entities.Setting(id=hostname_facts_id).read()
+        original_value = facts.value
+        facts.value = ""
+        try:
+            facts = facts.update(['value'])
+            self.assertNotEqual(facts.value, "", msg="Empty string")
```
mirekdlugosz, line 122: I know it's here for consistency with other tests, but since we are switching to pytest, you might as well use `assert` here instead of unittest-provided method.

PR #7772, opened 2020-05-11
Code Coverage 1: low impact test case

latran:
IMO, this is a low impact test case that should be removed.  Let me know if you think otherwise.


latran:
Added a couple more removals.  Removed stubbed test cases as it's better covered in CLI or UI than API.

jameerpathan111:
> Added a couple more removals. Removed stubbed test cases as it's better covered in CLI or UI than API.

@latran @mirekdlugosz To which interface we give more priority/importance while adding test coverage for a feature?

latran:
@jameerpathan111 

That's a good question.  I think it's a case by case basis and what limitations does each test case have.  I try to do all 3, but typically CLI and UI are top priorities for me.

latran:
Can someone merge this please?

PR #7774, opened 2020-05-12
Change settings to Dynaconf [WIP]

### LINKED ISSUE #6756: Use YAML for configuration
I think we should look to change configuration file from using ConfigParser-style INI files to using YAML. 

I am filing this issue to solicit feedback and discuss things that need to be done to accomplish this

We have a fair amount of code that looks at the Settings objects, luckly we have abstracted a lot of the reading of the config parser into `INIReader`. I believe the fastest approach would be to create a new class like `YAMLReader` that has the same methods as `INIReader`. This would mean we wouldn't have to readwrite any of the `FeatureSettings` code to handle the new format.

There would be some other refactoring required, anywhere INIReader is called. 

We can create a script to convert existing robotello.properties to robotello.yaml.

Here are some question I would like some feedback on.

1. Do people likes this idea?

2. Would we want to support both the new YAML and old INI at the same time, or can we do a clean cut over (assuming we have an easy way to convert the new files. 

3. What should the file be called? 
-Do we want to call it `robotello.properties` but have it contain YAML, just so it will be compatible with existing CI/tooling? 
-We could probably support multiple names `robottelo.yaml`, 'robottelo.properties`, etc. 

4. Does anyone see any advantage to moving to using a config directory, with multiple files instead of one. Some like :`robotello.config.d`:
    ```
    $ ls  robottelo.config.d/
   compute-resource.yaml	server.yaml
    ```
    In some ways I think this would be better, but I go back and forth.

5. The INIReader compantable YAMLReader would be good for existing configs, since the INI format is only 2 levels. However YAML allows for much more complicated configs, what sorts of methods will we want to have for "newer" configuration features.

rochacbruno:
Fix #6756 

Most of the work have been done upstream in order to make Dynaconf compatible with robottelo:

- Envless mode: https://github.com/rochacbruno/dynaconf/pull/337
- Lowercase Keys: https://github.com/rochacbruno/dynaconf/pull/338 
- Auto Complete: https://github.com/rochacbruno/dynaconf/pull/339

Upstream issues still pending before 3.0.0 release https://github.com/rochacbruno/dynaconf/milestone/9

## Changes in this PR

- Renamed `settings` to `legacy_settings` (kept to be used to migrate old robottelo.properties to the new formats)
- Created a new Dynaconf instance in place of `settings` object
- Dynaconf is loaded in `envless_mode` (no layered separation for env such as default, production, testing etc)
- Dynaconf is loaded with `lowecase_read=True` to allow lowercase settings variable (while the default and recommended by dynaconf is uppercase for settings and constants)
- Settings files are split on `conf/` directory (still need to decide the better organization for settings files or if we are going with a single file)
- Added `to_dict` function to the legacy settings

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+# This file is here for dynaconf to set this as root level directory
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,14 @@ tests/foreman/pytest.ini
 
 # For `coverage xml` generated file.
 **/coverage.xml
+
+# Dynaconf local files
+/.secrets.yaml
+/.secrets_*.yaml
+/settings.local.yaml
+/settings*.local.yaml
+
+# I don't know where those 2 files come from
+# but they are always there.
+full_upgrade
+upgrade_highlights
--- a/conf/base.yaml
+++ b/conf/base.yaml
@@ -1,0 +1,40 @@
+screenshots_path: /tmp/robottelo/screenshots/
+locale: en_US.UTF-8
+verbosity: debug
+tmp_dir: /var/tmp
+artifacts_server: server.example.com
+log_driver_commands:
+  - newSession
+  - windowMaximize
+  - get
+  - findElement
+  - sendKeysToElement
+  - clickElement
+  - mouseMoveTo
+browser: selenium
+webdriver: chrome
+browseroptions:
+webdriver_binary: /usr/bin/firefox
+webdriver_desired_capabilities:
+  platform: Linux
+  browserName: chrome
+  unhandledPromptBehavior: dismiss
+  unexpectedAlertBehaviour: ignore
+  acceptInsecureCerts: true
+command_executor: http://127.0.0.1:4444/wd/hub
+saucelabs_user:
+saucelabs_key:
+cdn: true
+run_one_datapoint: false
+rhel6_repo: http://example.com/yum/repo_files/rhel6-updates.repo
+rhel7_repo: http://example.com/yum/repo_files/rhel7-updates.repo
+rhel6_os: http://example.com/yum/rhel6-os/
+rhel7_os: http://example.com/yum/rhel7-os/
+rhel8_os:
+  baseos: http://example.com/rhel-8/BaseOS/x86_64/os/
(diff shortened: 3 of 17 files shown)


rochacbruno:
Worked on it during my PTO time, VAlidators are now taking some time to convert, also depending on the release of Dynaconf 3.0.0

rochacbruno:
## TODO

- [x] Release [Dynaconf 3.0.0](https://github.com/rochacbruno/dynaconf/milestone/9)
  **DONE** but now there will be a 3.1.0
- [ ] Command to migrate robottelo.properties to YAML format
   
     The **legacy_settings**  object got a new method called `to_dict` https://github.com/rochacbruno/robottelo/blob/b9b9469fb222b9057b45de9e2ec7f7df7caa1677/robottelo/config/base.py#L1612
    This field is needed because on Robottelo legacy settings some fields are computed and parsed at runtime, so a static conversion would not work.

- [ ] Command `robottelo configure` interactive generate config file from sample

   Idea is to provide a new command on robottelo CLI to allow users to migrate their settings to the new format.
   a command can be added here https://github.com/rochacbruno/robottelo/blob/b9b9469fb222b9057b45de9e2ec7f7df7caa1677/robottelo/config/base.py#L1612 and then called like `manage command_name` in the root of robottelo.
   
- [ ] Rewrite `settings` helpers in a new location
     - Some doesn't belong to settings (this needs to be carefully looked), ex: does this belong to settings object? 
https://github.com/rochacbruno/robottelo/blob/b9b9469fb222b9057b45de9e2ec7f7df7caa1677/robottelo/config/base.py#L1551
   
     - Some should be created as a `Lazy` setting object - issue is opened https://github.com/rochacbruno/dynaconf/issues/377 

- [ ] Write all the validators
   All the `validate` methods should be rewritten in form of dynaconf.Validator like https://github.com/rochacbruno/robottelo/blob/b9b9469fb222b9057b45de9e2ec7f7df7caa1677/robottelo/config/validators.py#L7

- [ ] Document the usage
- [ ] Run the CI tests

mshriver:
Replaced by #8003, thanks so much for the proof of concept and getting this started @rochacbruno !

PR #7779, opened 2020-05-13
more storage size to gce tests

tstrych:
Found error in last automation run, I run it locally and I found another one with storage size explained here. 
After that I was able to run test successfully in my local environment. 

test resulsts:
`============ 2 passed, 28 deselected, 6 warnings in 1041.00 seconds ============`

Don't know when minimal storage size requirements were updated. I mention this beacause that means I do not know if this should go into 6.7.z. Needs to be tested on 6.7 and we will see.

### DIFF
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -1774,6 +1774,7 @@ def test_positive_gce_provision_end_to_end(
     hostname = '{0}.{1}'.format(name, gce_domain.name)
     gceapi_vmname = hostname.replace('.', '-')
     root_pwd = gen_string('alpha', 15)
+    storage = [{'size': 20}]
     with Session('gce_tests') as session:
         session.organization.select(org_name=module_org.name)
         session.location.select(loc_name=module_loc.name)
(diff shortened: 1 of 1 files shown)


tstrych:
@lvrtelov is owner of gce, but these tests are in the test_host. At least you should know about it Lucia.

tstrych:
Thanks for answers and for questions as always :) 

@mirekdlugosz  by default there was 10. 

Maybe @lvrtelov knows if the requirements changed. That's one of the reasons why I pinged her above. 

Error message from ui:
Screenshot from 2020-05-14 18-42-35

mirekdlugosz:
Thanks for reply.

I looked at foreman source code, and I'm pretty sure Satellite just displays message it received from compute resource provider (GCE, in this case). Error message comes from [`app/models/concerns/orchestration/compute.rb`](https://github.com/theforeman/foreman/blob/develop/app/models/concerns/orchestration/compute.rb), and function to create VM is defined in [`app/models/compute_resources/foreman/model/gce.rb`](https://github.com/theforeman/foreman/blob/develop/app/models/compute_resources/foreman/model/gce.rb). Neither of them has anything about disk size.

I'd say 6.7 is affected as well. So this is candidate for cherry-picking after merging.

PR #7799, opened 2020-05-27
Remove hammer content-host test

swadeley:
* hammer content-host command was removed in Sat6.1
* there is no single command replacement for errata count


swadeley:
thank you

PR #7803, opened 2020-06-01
fam package check

pondrejk:
result:
```pytest tests/foreman/sys/test_fam.py 

collected 1 item                                                                                                     

tests/foreman/sys/test_fam.py .                                                                                [100%]

============================================= 1 passed in 162.68 seconds =============================================
```


robottelo/constants.py
```diff
@@ -1797,3 +1797,59 @@
 }
 
 RHSSO_RESET_PASSWORD = {"temporary": "false", "type": "password", "value": ""}
+
+FOREMAN_ANSIBLE_MODULES = [
+    'foreman_architecture',
```
mirekdlugosz, line 1802: I know this is the way we have historically approached this, but I dislike how it mandates manual intervention to cover new modules as they appear.

Would it be possible to obtain list of all modules dynamically, from package itself or Satellite API?
This way we would cover all modules that there are automatically.

We can keep constant as "bare minimum that must exist". This will allow us to detect possibly unintentional removal of modules. We would remove it from constant only once we are certain this removal is intended.

tests/foreman/sys/test_fam.py
```diff
@@ -0,0 +1,42 @@
...
+"""
+from robottelo import ssh
+from robottelo.constants import FOREMAN_ANSIBLE_MODULES
+from robottelo.decorators import destructive
+from robottelo.decorators import run_in_one_thread
+
+
+@destructive
+@run_in_one_thread
+def test_positive_ansible_modules():
```
mirekdlugosz, line 23:
```suggestion
def test_positive_ansible_modules_installation():
```
?

At the end of the day, we don't check ansible modules themselves, we check if they can be installed and that there is something working after installation.

robottelo/constants.py
pondrejk: thanks, updated so that the list is obtained from the directory where the package installs is (more elegant solution not at hand atm)

rplevka:
ACK :) too late

PR #7804, opened 2020-06-01
Pluged in Pytest API fixtures

jyejare:
This PR contains:

- [x] Moved existing `robottelo/api/entity_fixtures.py` to `pytest_fixtures/api_fixtures.py`.
- [x] `pytest_fixtures/api_fixtures` is a new plugin to pytest that should remove all explicit calls/imports to API Entity fixtures in conftest.py.
- [x] Renaming of `module fixtures` to `default fixtures` as they were just reading the default/existing entity not creating a new one.
- [x] Additional: Tests in test modules now have renamed fixture names wherever they have been accessed. Hence handling the test failures due to renamed fixture names.

### DIFF
--- a/pytest_fixtures/api_fixtures.py
+++ b/pytest_fixtures/api_fixtures.py
@@ -23,6 +23,19 @@ from robottelo.test import settings
 
 # Global Satellite Entities
 
+if not settings.configured:
+    settings.configure()
+
+
+@pytest.fixture(scope='session')
+def default_lce():
+    return entities.LifecycleEnvironment().search(query={'search': 'name=Library'})[0]
+
+
+@pytest.fixture(scope='module')
+def module_lce(module_org):
+    return entities.LifecycleEnvironment(organization=module_org).create()
+
 
 @pytest.fixture(scope='module')
 def module_org():
@@ -34,21 +47,23 @@ def module_location(module_org):
     return entities.Location(organization=[module_org]).create()
 
 
-@pytest.fixture(scope='module')
-def module_lce(module_org):
-    return entities.LifecycleEnvironment(organization=module_org).create()
-
-
-@pytest.fixture(scope='module')
-def module_smart_proxy(module_location):
+@pytest.fixture(scope='session')
+def default_smart_proxy():
     smart_proxy = (
         entities.SmartProxy()
         .search(query={'search': 'name={0}'.format(settings.server.hostname)})[0]
         .read()
     )
-    smart_proxy.location.append(entities.Location(id=module_location.id))
-    smart_proxy.update(['location'])
     return entities.SmartProxy(id=smart_proxy.id).read()
+
+
+@pytest.fixture(scope='session')
+def default_domain(default_smart_proxy):
+    *_, domain_name = settings.server.hostname.partition('.')
+    dom = entities.Domain().search(query={'search': 'name={}'.format(domain_name)})[0]
+    dom.dns = default_smart_proxy
+    dom.update(['dns'])
+    return entities.Domain(id=dom.id).read()
 
 
 @pytest.fixture(scope='module')
@@ -57,34 +72,28 @@ def module_domain(module_org, module_location):
 
 
 @pytest.fixture(scope='module')
-def module_subnet(module_org, module_location, module_domain, module_smart_proxy):
+def module_subnet(module_org, module_location, default_domain, default_smart_proxy):
     network = settings.vlan_networking.subnet
     subnet = entities.Subnet(
         network=network,
         mask=settings.vlan_networking.netmask,
-        domain=[module_domain],
+        domain=[default_domain],
         location=[module_location],
         organization=[module_org],
-        dns=module_smart_proxy,
-        dhcp=module_smart_proxy,
-        tftp=module_smart_proxy,
-        discovery=module_smart_proxy,
+        dns=default_smart_proxy,
+        dhcp=default_smart_proxy,
+        tftp=default_smart_proxy,
+        discovery=default_smart_proxy,
         ipam='DHCP',
     ).create()
     return subnet
 
 
-@pytest.fixture(scope='module')
-def module_partiontable(module_org, module_location):
-    ptable = (
-        entities.PartitionTable()
-        .search(query={'search': 'name="{0}"'.format(DEFAULT_PTABLE)})[0]
-        .read()
-    )
-    ptable.location.append(module_location)
-    ptable.organization.append(module_org)
-    ptable.update(['location', 'organization'])
-    return entities.PartitionTable(id=ptable.id).read()
+@pytest.fixture(scope='session')
+def default_partitiontable():
+    ptables = entities.PartitionTable().search(query={'search': f'name="{DEFAULT_PTABLE}"'})
+    if ptables:
+        return ptables[0].read()
 
 
 @pytest.fixture(scope='module')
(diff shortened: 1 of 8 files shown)


pytest_fixtures/api_fixtures.py
```diff
@@ -19,6 +19,19 @@
 
 # Global Satellite Entities
 
+if not settings.configured:
+    settings.configure()
+
+
+@pytest.fixture(scope='module')
```
mshriver, line 26: session scope?

jyejare: Yeah ! I should do it for all the default fixtures!

pytest_fixtures/api_fixtures.py
```diff
@@ -47,34 +55,45 @@ def module_smart_proxy(module_location):
     return entities.SmartProxy(id=smart_proxy.id).read()
 
 
+@pytest.fixture(scope='module')
+def default_domain(module_org, module_location, default_smart_proxy):
```
mshriver, line 59: What is the intent here with the use of 'default' for some of these module scoped fixtures?

I think it introduces ambiguity - is it something that exists by default on the satellite test instance? Is it something provided by default in the build, or that robottelo adds to the instance that is a default for the session?

pytest_fixtures/api_fixtures.py
```diff
@@ -144,15 +163,15 @@ def module_os(
             entities.OperatingSystem()
             .search(
                 query={
-                    'search': 'family="Redhat" AND major="{0}" AND minor="{1}")'.format(
+                    u'search': u'family="Redhat" AND major="{0}" AND minor="{1}")'.format(
```
mshriver, line 166: These `u` should not be necessary, did you intentionally include them or did a formatting tool add them?

jyejare: Those are not necessary and I just forgot to `f-string` them.

pytest_fixtures/api_fixtures.py
```diff
@@ -31,12 +44,7 @@ def module_location(module_org):
 
 
 @pytest.fixture(scope='module')
-def module_lce(module_org):
-    return entities.LifecycleEnvironment(organization=module_org).create()
-
-
-@pytest.fixture(scope='module')
-def module_smart_proxy(module_location):
+def default_smart_proxy(module_location):
```
mshriver, line 47: Some docblocks on these would be incredibly helpful, especially considering this and `default_domain` are actually returning data structures (the result of `read()`), not entity instances.

pytest_fixtures/api_fixtures.py
jyejare: These are:

- Either, added to the instance by automation as pre-requisites before the pytest session. Like the domain/ subnet etc.
- Or, the defaults in satellite, like `x86_64` and `i386` are default architectures present in satellite. Same for default OS as `RedHat`.

mshriver: Noted - I think if we're going to use 'default' for fixture names in robottelo, it should be clear what a 'default' is, and I think its very ambiguous to use it for both of these situations.

To me, defaults provided by the build itself in satellite/foreman/etc, shouldn't necessarily be fixtures - there is no setup involved.

For things that the test session is going to use as a default, but are expected pre-requisites in the satellite instances under test, I think fixtures are appropriate to do the lookup - as they can skip the test if the expected entity isn't pre-configured, or can take action to configure it for the session.

JacobCallahan:
ACK pending Mike's comments being resolved.

pytest_fixtures/api_fixtures.py
jyejare: No, `Read` returns instance itself of an entity. Same with `Create` which inside returns `Read` only. 

All `default_` fixtures are reading the existing entities from satellite instead of creating them, but they are returning an instance itself.

pytest_fixtures/api_fixtures.py
jyejare: default fixtures are fixtured so that we will minimize the count/time of reading those fixtures (in short caching them).  There are also 

> For things that the test session is going to use as a default, but are expected pre-requisites in the satellite instances under test, I think fixtures are appropriate to do the lookup - as they can skip the test if the expected entity isn't pre-configured, or can take action to configure it for the session.

Do you prefer to have a fixture that will look for default(pre-requisite) first else create new or raise pytest.failure ?

jyejare:
@mshriver just replied to your comments and fixed one. Looking for your ACK.

Also, @omkarkhatavkar is waiting on this PR to be merged for further contribution.

pytest_fixtures/api_fixtures.py
mshriver: I'd prefer having it lookup and create, only failing if its unable to create.

tests/foreman/conftest.py
```diff
@@ -5,29 +5,6 @@
 
 import pytest
 
-from robottelo.api.entity_fixtures import azurerm_settings  # noqa: F401
```
mshriver, line 8: :tada:

vijay8451:
@jyejare below are related to fixtures as tests failing around same:
- `module_configtemaplate` is no more valid, reason  https://github.com/SatelliteQE/nailgun/pull/724
- Below calls are no more required as by default org/loc are part of same entities hence they would fail with error ` "error": {"message":"Validation failed: Taxonomy has already been taken"}`
```
        x.location.append(module_location)
        xorganization.append(module_org)
       x.update(['location', 'organization'])
```
Note: `x` means any entity which is doing the same under fixtures.

pytest_fixtures/api_fixtures.py
jyejare: @mshriver Actually we are giving a choice here! One can choose existing or create new! In tests, sometimes we need existing entity and sometimes we need to create new. By lookup and create, we will force users to use existing entity which is not the case for every test.

So I will leave it up to the QE:
1. To use existing -OR-
2. To use newly created -OR-
3. To check the existing and and create new

This can be handled from tests itself instead we forcing from fixture.

jyejare:
@vijay8451 Fixed configTemplate as per PR https://github.com/SatelliteQE/robottelo/pull/7822.

And all the occurrences of org and loc update for all default fixtures have been removed.

jyejare:
The Travis will be fixed once #7825 will be merged.

pytest_fixtures/api_fixtures.py
mshriver: We're talking about two different use cases within the tests then.

The entities which we are calling 'default', and which we expect to exist for the session and not be destroyed by the tests, should have a lookup+create pattern. This would allow robottelo to run on an instance pre-configured for test (what we could configure for our test targets internally) or on a satellite instance that's just been installed with no further configuration.

Tests which require a unique entity that they will be modifying or destroying should not use the default fixture, but a module or function scoped fixture that is separate from the default fixture.  We could control this with indirect parametrization, but I think its more clear to have a separate fixture.  Some of these would be defined local to the test module if they aren't relevant to other test modules, but fixtures which are relevant to multiple components should be defined globally, but still be separate from the 'default' component fixtures.

pytest_fixtures/api_fixtures.py
```diff
@@ -30,57 +43,53 @@ def module_location(module_org):
...
         .read()
     )
-    smart_proxy.location.append(entities.Location(id=module_location.id))
-    smart_proxy.update(['location'])
     return entities.SmartProxy(id=smart_proxy.id).read()
 
 
+@pytest.fixture(scope='session')
+def default_domain(default_smart_proxy):
+    _, _, domain_name = settings.server.hostname.partition('.')
```
JacobCallahan, line 58:
```suggestion
    *_, domain_name = settings.server.hostname.partition('.')
```

JacobCallahan:
Travis failures are unrelated. Merging

PR #7809, opened 2020-06-02
Create robottelo.utils, move issue_handler helpers

mshriver:
Create pytest_plugin.issue_handlers module
Modify behavior of bz-cache option to avoid unnecessarily annoying logic
Create version util
Move things from helpers into specific modules


robottelo/utils/issue_handlers/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Bugzilla: BZ:123456
+
+    Arguments:
+        issue {str} -- A string containing handler + number e.g: BZ:123465
+        data {dict} -- Issue data indexed by <handler>:<number> or None
+    """
+    # Handlers can be extended to support different issue trackers.
+    if str(issue).startswith(SUPPORTED_HANDLERS):
+        handler_code, _, _ = str(issue).partition(":")
```
JacobCallahan, line 22:
```suggestion
        handler_code, *_ = str(issue).partition(":")
```

pytest_fixtures/api_fixtures.py
```diff
@@ -30,57 +43,53 @@ def module_location(module_org):
...
         .read()
     )
-    smart_proxy.location.append(entities.Location(id=module_location.id))
-    smart_proxy.update(['location'])
     return entities.SmartProxy(id=smart_proxy.id).read()
 
 
+@pytest.fixture(scope='session')
+def default_domain(default_smart_proxy):
+    _, _, domain_name = settings.server.hostname.partition('.')
```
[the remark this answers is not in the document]
mshriver: This is actually from the changes in 7804, but I can make it here.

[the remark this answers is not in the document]
mshriver: I'm going to actually suggest changing this to index in my commit, because all we're binding is `domain_name` to the last item in the tuple. No need to shuffle around the syntax of arguments that we aren't using anyway.

Since we're using partition, and it always returns a three item tuple, I'd actually think explicit unpacking of all three items is better here.

robottelo/utils/issue_handlers/__init__.py
mshriver: Again just moved the original function here, but I'll specifically index.

pytest_fixtures/api_fixtures.py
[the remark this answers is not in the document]
JacobCallahan: the change I suggest combines the unpacking of right value with packing of undesired left values.
i believe jitendra already made this change in his PR though

pytest_fixtures/api_fixtures.py
```diff
@@ -59,7 +59,7 @@ def default_smart_proxy():
 
 @pytest.fixture(scope='session')
 def default_domain(default_smart_proxy):
-    *_, domain_name = settings.server.hostname.partition('.')
+    domain_name = settings.server.hostname.partition('.')[-1]
```
JacobCallahan, line 62: +1 this is better

pytest_plugins/issue_handlers.py
```diff
@@ -0,0 +1,294 @@
...
+            data[issue.strip()]['used_in'].append({'usage': usage, **kwargs})
+
+
+def _should_deselect(issue, data=None):
+    """Check if test should be deselected based on marked issue."""
+    # Handlers can be extended to support different issue trackers.
+    handlers = {'BZ': bugzilla.should_deselect_bz}
+    supported_handlers = tuple(f"{handler}:" for handler in handlers.keys())
+    if str(issue).startswith(supported_handlers):
+        handler_code, _, _ = str(issue).partition(":")
```
JacobCallahan, line 93:
```suggestion
        handler_code = str(issue).partition(":")[0]
```

pytest_plugins/issue_handlers.py
```diff
@@ -0,0 +1,294 @@
...
+                ...
+            }
+    """
+    settings.configure()
+    valid_markers = ["skip_if_open", "skip", "deselect"]
+    collected_data = defaultdict(lambda: {"data": {}, "used_in": []})
+
+    bz_cache_filename = config.getvalue('bz_cache')  # use existing json cache?
+    cached_data = None
+    if bz_cache_filename is not None:
```
JacobCallahan, line 144:
```suggestion
    if bz_cache_filename:
```
as long as bz_cache_filename doesn't have any valid "falsey" values

pytest_plugins/issue_handlers.py
```diff
@@ -0,0 +1,294 @@
...
+                cached_data = {
+                    k: search_version_key(k, v) for k, v in json.load(bz_cache_file).items()
+                }
+        except FileNotFoundError:
+            # no bz cache file exists
+            LOGGER.warning(
+                f'bz-cache option used, file [{bz_cache_filename}] not found, '
+                f'NOT using cached data.'
+            )
+            pass
```
JacobCallahan, line 157: could use pathlib here and check that the file exists first.
```python
bz_cache = Path(bz_cache_filename)
if bz_cache.exists():
    with bz_cache.open() as bz_cache_file:
        LOGGER.info(f'Using BZ cache file for issue collection: {bz_cache_filename}')
        cached_data = {
            k: search_version_key(k, v) for k, v in json.load(bz_cache_file).items()
        }
else:
    LOGGER.warning(
        f'bz-cache option used, file [{bz_cache_filename}] not found, NOT using cached data.'
    )
```

pytest_plugins/issue_handlers.py
```diff
@@ -0,0 +1,294 @@
...
+        component = None
+        bzs = None
+        importance = None
+
+        # register test module as processed
+        test_modules.add(item.module)
+
+        # Find matches from docstrings top-down from: module, class, function.
+        mod_cls_fun = (item.module, getattr(item, 'cls', None), item.function)
+        for docstring in [d for d in map(inspect.getdoc, mod_cls_fun) if d is not None]:
```
JacobCallahan, line 174: same falsey question for `is not None`

pytest_plugins/issue_handlers.py
mshriver: Its my preference to be explicit when checking against `None`, but I'm flexible if the shorter check is the syntax preference here in Robottelo.

pytest_plugins/issue_handlers.py
mshriver: Optional suggestion or would you rather see that pattern?  I prefer the try/except pattern in cases like this, maybe catching more than `FileNotFoundError` (like `OSError`) would make this more robust as it would account for many other file failures.

pytest_plugins/issue_handlers.py
mshriver: ack - here I definitely want to be explicit, as its possible the docblock is an empty string (dumb, but possible?). I'm not nearly as certain that the return from `getdoc` will be `None`, like I would be with the pytest CLI `config.getvalue` call and check above.

Would shorten the composition a bit - optional or required?

pytest_plugins/issue_handlers.py
JacobCallahan: do we expect an OSError? as a general rule, i try not to use try's! :D

pytest_plugins/issue_handlers.py
JacobCallahan: definitely optional

pytest_plugins/issue_handlers.py
mshriver: well, the file could exist, but not be readable, or writeable, etc - all of which fall under OSError.

pytest_plugins/markers.py
```diff
@@ -1,3 +1,6 @@
+# Custom markers for robottelo tests
```
jyejare, line 1: Worth removing them from `robottelo.decorators` ?

pytest_plugins/issue_handlers.py
```diff
@@ -0,0 +1,272 @@
...
+from robottelo.config import settings
+from robottelo.helpers import slugify_component
+from robottelo.utils.issue_handlers import add_workaround
+from robottelo.utils.issue_handlers import bugzilla
+from robottelo.utils.issue_handlers import is_open
+from robottelo.utils.issue_handlers import should_deselect
+from robottelo.utils.version import search_version_key
+from robottelo.utils.version import VersionEncoder
+
+LOGGER = logging.getLogger('robottelo')
```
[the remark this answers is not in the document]
mshriver: using `'issue_handlers_plugin'` to provide a bit more context - good call.

pytest_plugins/issue_handlers.py
```diff
@@ -0,0 +1,272 @@
...
+DEFAULT_BZ_CACHE_FILE = 'bz_cache.json'
+
+
+def pytest_addoption(parser):
+    """Adds custom options to pytest runner."""
+    parser.addoption(
+        "--bz-cache",
+        action='store_true',
+        help=f"Use an existing {DEFAULT_BZ_CACHE_FILE} instead of calling BZ API. "
+        "One will be created automatically by default, but not used by default",
```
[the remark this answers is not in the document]
mshriver: Will give it a shot at re-wording, should be in next commit pushed to PR, let me know what you think.

pytest_plugins/issue_handlers.py
```diff
@@ -0,0 +1,272 @@
...
+    )
+
+
+@pytest.hookimpl(trylast=True)
+def pytest_collection_modifyitems(session, items, config):
+    """Generate the issue collection (using bz cache via bugzilla issue handler util"""
+    # This collection includes pre-processed `is_open` status for each issue
```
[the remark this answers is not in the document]
mshriver: First one sure, but they're specific to the line following the comments, not the behavior of the function overall.

The call to `generate_issue_collection` is what is actually writing the file, and reading the option from config.

pytest_plugins/issue_handlers.py
```diff
@@ -0,0 +1,272 @@
...
+                            "component": "Repositories",
+                            "usage": "skip_if_open"
+                        },
+                        ...
+                    ]
+                },
+                ...
+            }
+    """
+    settings.configure()
```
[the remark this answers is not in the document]
mshriver: Can't assume any other plugins are going to load the settings, otherwise I'm setting this up for failure.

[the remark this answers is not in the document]
mshriver: This is more a failure of the overall design of the settings instance as it exists right now. No module should have to trigger configure, because no module can be certain it hasn't already been done - and no module should have to check whether the configuration file has been read yet.

pytest_plugins/issue_handlers.py
```diff
@@ -0,0 +1,272 @@
...
+                LOGGER.info(f'Using BZ cache file for issue collection: {DEFAULT_BZ_CACHE_FILE}')
+                cached_data = {
+                    k: search_version_key(k, v) for k, v in json.load(bz_cache_file).items()
+                }
+        except FileNotFoundError:
+            # no bz cache file exists
+            LOGGER.warning(
+                f'--bz-cache option used, cache file [{DEFAULT_BZ_CACHE_FILE}] not found'
+            )
+            pass
```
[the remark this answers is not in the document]
mshriver: true, not syntactically required here, personal habit of including them in such exception handling with no re-raise. will remove.

pytest_plugins/issue_handlers.py
```diff
@@ -0,0 +1,272 @@
...
+)
+
+BZ = re.compile(
+    # To match :BZ: 123456, 456789
+    r"\s*:BZ:\s*(?P<bz>.*\S*)",
+    re.IGNORECASE,
+)
+
+
+def generate_issue_collection(items, config):  # pragma: no cover
```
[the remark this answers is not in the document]
mshriver: Agreed on it being a bit too long, but I would like to avoid re-factoring the contents into pieces within this PR, given the scope. Would you agree its acceptable as is (copied and modified from original location), and that we can break it up further as we have the need?

pytest_plugins/markers.py
mshriver: Not sure exactly what you mean removing here - there are plenty of pytest marker re-naming I would like to remove from `decorators.__init__` and update all test modules to import and use them directly from pytest.  I don't think it falls into the scope of this PR.

This change is just actually registering the existing markers with pytest so we don't get warnings about unregistered markers.

pytest_plugins/issue_handlers.py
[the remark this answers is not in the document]
jyejare: @mshriver Agree, lets open another PR for it.

pytest_plugins/markers.py
jyejare: The same markers defined in your `pytest_plugins/markers.py` are defined in `robottelo.decorators.__init__.py`. So I am talking abt the same that u mentioned above comment.

I am ok to improve it from another PR.

pytest_plugins/issue_handlers.py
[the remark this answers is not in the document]
jyejare: @mshriver Agree on the failure of design, but it is how it is :)

Is DynaConf going to help here ?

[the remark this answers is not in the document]
mshriver: Note I did update the use to include the check first.

Dynaconf will indeed improve this, as modules reading settings will no longer have to worry about whether or not the config has been loaded yet.

pytest_plugins/markers.py
mshriver: Gotcha. I didn't want to add `robottelo.decorators.__init__` to conftest, but rather register them with pytest from a new markers module, and move marker definitions into this module.

PR #7821, opened 2020-06-08
VirtualMachine: fix overwriting distro with source image

lpramuk:
- VirtualMachine: fix overwirting distro with source image
- CapsuleVirtualMachine: omit RHEL7 conditions (there is no RHEL6 capsule since 6.3)

wrong *vm.py*:
```
L118: self.distro = settings.distro.image_el7 or distro
L143: source_image = "{self.distro}-base"
```
as you can see image_el7 -> **distro** -> source_image

vs.

fixed *vm.py*:
```
L118: self.distro = distro
L143: source_image ="{settings.distro.image_el7}-base"
```
image_el7 -> source_image

### DIFF
--- a/robottelo/vm.py
+++ b/robottelo/vm.py
@@ -78,7 +78,7 @@ class VirtualMachine(object):
         bridge=None,
         network=None,
     ):
-        distro_mapping = {
+        image_map = {
             DISTRO_RHEL6: settings.distro.image_el6,
             DISTRO_RHEL7: settings.distro.image_el7,
             DISTRO_RHEL8: settings.distro.image_el8,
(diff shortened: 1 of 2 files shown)


lpramuk:
Tests are not using vm.distro so none of them are affected. Why I did this PR? 

wrong:
```distro = rhel77-20190128.1 | rhel610 | rhel810-20200429.0```  < basically source_image
if distro[:5] = DISTRO_RHEL7


correct:
```distro = rhel7 | rhel6 | rhel8 ```
if distro == DISTRO_RHEL7

pondrejk:
Well, maybe one day some test will use vm.distro :) Not sure why make_docs fails in travis, seems unrelated

PR #7822, opened 2020-06-08
Fix AttributeError: module 'nailgun.entities' has no attribute 'ConfigTemplate'

lpramuk:
ConfigTemplate was removed in nailgun  - https://github.com/SatelliteQE/nailgun/pull/724 (and cherrypicked to stable)
But no action done for robottelo.

This broke pytest test collection:
```
==================================== ERRORS ====================================
__________ ERROR collecting tests/foreman/api/test_multiple_paths.py ___________
tests/foreman/api/test_multiple_paths.py:37: in <module>
    entities.ConfigTemplate,
E   AttributeError: module 'nailgun.entities' has no attribute 'ConfigTemplate'
```

Due to this error no sequential tests are being run in all tiers !!!

PR #7824, opened 2020-06-08
WIP: Fix filtered_datapoint() and improve parametrized()

lpramuk:


lpramuk:
> Did some search, in /ui/test_product.py there is an instance of `@parametrize('product_name', **valid_data_list('ui'))`, is that affected by this change?

Yeah I see https://github.com/SatelliteQE/robottelo/pull/7822 merged (another pytest collection error) so I will rebase on top of it and can try pytest collection thru all the tests

PR #7832, opened 2020-06-15
moving the fixtures so can be use across other class

omkarkhatavkar:
### PR Details 

- Currently, this PR is the segregate the fixture from UI LDAP component so it can be used Across another project.
- Idle session timeout verification test. Although this works well in 1 min idle timeout in browser but seems like there is some problem with selenium session browser takes some more time to get an idle timeout. Hence I have increased the idle timeout sleep in the test. If you have any suggestion that will work I'm open for that.
- Enhance the test to verify the login should be failed for deleted RHSSO User 

   
### Test Result
```
Testing started at 12:17 PM ...
/home/okhatavk/okhatavk/Satellite/robottelo/env/bin/python /home/okhatavk/Downloads/pycharm-community-2018.1.4/helpers/pycharm/_jb_pytest_runner.py --target test_ldap_authentication.py::test_session_expire_rhsso_idle_timeout
Launching py.test with arguments test_ldap_authentication.py::test_session_expire_rhsso_idle_timeout in /home/okhatavk/Satellite/robottelo/tests/foreman/ui

plugins: forked-1.1.3, xdist-1.31.0, cov-2.8.1, mock-1.10.4, services-1.3.12020-06-17 06:47:41 - conftest - DEBUG - Collected 1 test cases
collected 1 item

========================== 1 passed in 459.11 seconds ==========================
```


pytest_fixtures/authsource_fixtures.py
```diff
@@ -0,0 +1,207 @@
...
+        'ipa_otp_username': settings.ipa.otp_user,
+        'ldap_ipa_user_passwd': settings.ipa.password_ipa,
+        'ipa_base_dn': settings.ipa.basedn_ipa,
+        'ipa_group_base_dn': settings.ipa.grpbasedn_ipa,
+        'ldap_ipa_hostname': settings.ipa.hostname_ipa,
+        'time_based_secret': settings.ipa.time_based_secret,
+    }
+
+
+@fixture(scope='function')
```
jyejare, line 57: Keep function scoped fixtures in test_module else make it module-level scope fixture. So that it can be created every time per module.

Applies for all the non-module/non-session/non-class scoped fixtures.

I can be wrong though, @mshriver what do you think on keeping function level scoped fixtures here ?

pytest_fixtures/authsource_fixtures.py
```diff
@@ -0,0 +1,207 @@
+"""Test class for Active Directory Feature
```
mshriver, line 1: I don't think this metadata belongs in the fixture module, as this module is for defining related fixtures, to be used by various test modules.

Fixtures within this module may be used by test modules that are only reliant on LDAP as a prerequisite.

pytest_fixtures/authsource_fixtures.py
```diff
@@ -0,0 +1,207 @@
...
+from robottelo.constants import LDAP_SERVER_TYPE
+from robottelo.datafactory import gen_string
+from robottelo.decorators import fixture
+from robottelo.rhsso_utils import create_mapper
+from robottelo.rhsso_utils import get_rhsso_client_id
+from robottelo.rhsso_utils import run_command
+
+
+@fixture(scope='module')
+def ldap_data():
```
mshriver, line 34: I wouldn't necessarily consider this or `ipa_data` to be fixtures, as there is no setup involved, even within the framework. Its just creating a data structure from configuration data - this is something that could be defined as a module variable and simply used in the actual fixtures that achieve some configuration.

pytest_fixtures/authsource_fixtures.py
mshriver: I don't think there should be any restriction to the fixture scope used within these modules.

The fixture scope is relevant to its use within the framework and Satellite - there could be 10 test modules that contain test cases that need an authentication source configured on a function scoped level.

pytest_fixtures/authsource_fixtures.py
mshriver, line 1: I'd say its sufficient to call this module something like `pytest_fixtures/satellite_auth.py`, as the path contains 'fixtures' so we don't need to repeat it in the module name.

Being a bit more explicit in the naming of the module, it should be obvious these are fixtures related to authentication in Satellite - we could additionally later have `pytest_fixtures/virtwho_auth.py` for fixtures related to virt-who hypervisor authentication (as an example).

pytest_fixtures/authsource_fixtures.py
omkarkhatavkar: This is old code, thanks for raising. I agree on that no need makes as a fixture.

omkarkhatavkar:
@jyejare @mshriver Please have a second look on the PR.

pytest_fixtures/satellite_auth.py
```diff
@@ -0,0 +1,167 @@
...
+        location=[module_loc],
+    ).create()
+
+
+@fixture()
+def ldap_user_name(ldap_data, test_name):
+    """Add LDAP user to satellite by logging in, return username to test and delete the user (if
+    still exists) when test finishes.
+    """
+    with Session(test_name, ldap_data['ldap_user_name'], ldap_data['ldap_user_passwd']):
```
jyejare, line 67: For Session, instead of with context manager, just initiate Session and then manually close. Something like:
```
session = Session(......)
session.close()
```
Not sure abt the close function, just check whats the actual method to close it before you push this suggestion.

pytest_fixtures/satellite_auth.py
```diff
@@ -0,0 +1,167 @@
...
+        pass
```
jyejare, line 68: Not sure you really want this session to open and close without doing anything. Do you wanted to implement something here.

pytest_fixtures/satellite_auth.py
```diff
@@ -0,0 +1,167 @@
...
+    yield ldap_data['ldap_user_name']
+    users = entities.User().search(
+        query={'search': 'login="{}"'.format(ldap_data['ldap_user_name'])}
+    )
+    if users:
+        users[0].delete()
```
jyejare, line 74: If it's creating and deleting for each test then whats idea here for implementing it as a fixture? a normal function would have done same for you.

I understand that now this can be used other test modules as well, but these type of utilities we used to right as helper functions.

Need to discuss on whether we want to draw the line between helpers and fixtures.

@mshriver Thoughts ?

pytest_fixtures/satellite_auth.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+@fixture()
+def ldap_usergroup_name():
+    """Return some random usergroup name, and attempt to delete such usergroup when test finishes.
+    """
+    usergroup_name = gen_string('alphanumeric')
+    yield usergroup_name
+    user_groups = entities.UserGroup().search(query={'search': 'name="{}"'.format(usergroup_name)})
```
jyejare, line 83: I believe, the teardown part is taken care as an `extra` effort, whereas none of the fixtures or tests are doing a teardown of entities created during tests. We simply don't care about what has been created in tests unless required.

Does it block/fails your tests, if you don't teardown them?

pytest_fixtures/satellite_auth.py
```diff
@@ -0,0 +1,167 @@
...
+    """Return some random ldap name, and attempt to delete all ldap when test starts.
+    """
+    ldap = entities.AuthSourceLDAP().search()
+    for ldap_auth in range(len(ldap)):
+        users = entities.User(auth_source=ldap[ldap_auth]).search()
+        for user in range(len(users)):
+            users[user].delete()
+        ldap[ldap_auth].delete()
+    ldap_name = gen_string('alphanumeric')
+    yield ldap_name
```
jyejare, line 99:
```suggestion
    ldap = entities.AuthSourceLDAP().search()
    for ldap_auth in ldap:
        users = entities.User(auth_source=ldap_auth).search()
        for user in users:
            user.delete()
        ldap_auth.delete()
    ldap_name = gen_string('alphanumeric')
    yield ldap_name
```

pytest_fixtures/satellite_auth.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+@fixture(scope='module')
+def enroll_idm_and_configure_external_auth():
+    """Enroll the Satellite6 Server to an IDM Server."""
+    run_command(cmd='yum -y --disableplugin=foreman-protector install ipa-client ipa-admintools')
+
+    run_command(
+        cmd='echo {0} | kinit admin'.format(settings.ipa.password_ipa),
```
jyejare, line 108: Replace `.format()` with f-string formatting.
```suggestion
        cmd=f'echo {settings.ipa.password_ipa} | kinit admin',
```

jyejare: Do this for all over in your PR.

pytest_fixtures/satellite_auth.py
```diff
@@ -0,0 +1,167 @@
...
+        hostname=settings.ipa.hostname_ipa,
+    )
+    result = run_command(
+        cmd='ipa host-add --random {}'.format(settings.server.hostname),
+        hostname=settings.ipa.hostname_ipa,
+    )
+
+    for line in result:
+        if "Random password" in line:
+            _, password = line.split(': ', 2)
```
jyejare, line 118: What if the password has ':' in it, this is likely to retrive half passwrd. Need better password fetching machenism ?

pytest_fixtures/satellite_auth.py
```diff
@@ -0,0 +1,167 @@
...
+    for ldap_auth in range(len(ldap)):
+        users = entities.User(auth_source=ldap[ldap_auth]).search()
+        for user in range(len(users)):
+            users[user].delete()
+        ldap[ldap_auth].delete()
+    ldap_name = gen_string('alphanumeric')
+    yield ldap_name
+
+
+@fixture(scope='module')
```
jyejare, line 102: This can be `session` scoped so that once the enrollment is done, we need not to establish it again for different ldap test modules.

pytest_fixtures/satellite_auth.py
```diff
@@ -0,0 +1,167 @@
...
+    )
+    _, domain = settings.ipa.hostname_ipa.split('.', 1)
+    run_command(
+        cmd="ipa-client-install --password '{}' --domain {} --server {} --realm {} -U".format(
+            password, domain, settings.ipa.hostname_ipa, domain.upper()
+        )
+    )
+
+
+@fixture(scope='module')
```
jyejare, line 132: This can be `session` scoped so that once the enrollment is done, we need not to establish it again for different ldap test modules.

pytest_fixtures/satellite_auth.py
```diff
@@ -0,0 +1,167 @@
...
+    run_command(
+        cmd="satellite-installer --foreman-keycloak true "
+        "--foreman-keycloak-app-name 'foreman-openidc' "
+        "--foreman-keycloak-realm '{}' ".format(settings.rhsso.realm),
+        timeout=1000,
+    )
+    run_command(cmd="systemctl restart httpd")
+
+
+@fixture(scope='module')
```
jyejare, line 158: Same session scoped comment.

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1186,6 +1016,42 @@ def test_single_sign_on_using_rhsso(enable_external_auth_rhsso, session):
         update_rhsso_settings_in_satellite(revert=True)
 
 
+@upgrade
```
jyejare, line 1019: I think we don't run destructive tests for an upgrade yet. @devendra104 Thoughts ?

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1186,6 +1016,42 @@ def test_single_sign_on_using_rhsso(enable_external_auth_rhsso, session):
...
+
+    :id: 80247b30-a988-11ea-943c-d46d6dd3b5b2
+
+    :steps:
+        1. Change the idle timeout settings for the External Authentication
+        2. Login into Satellite using RHSSO login and wait for the idle timeout
+
+    :expectedresults: After completion of the idle timeout user session
+        should get expired
+    """
```
jyejare, line 1032: Do you want to reconsider the testImportance/caseLevel of this test ? By defaults its `high` I think.

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1186,6 +1016,42 @@ def test_single_sign_on_using_rhsso(enable_external_auth_rhsso, session):
...
+        property_name = 'idle_timeout'
+        setting_entity = entities.Setting().search(query={'search': f'name={property_name}'})[0]
+        setting_entity.value = 1
+        setting_entity.update({'value'})
+
+        with session(login=False):
+            session.rhsso_login.login(
+                {'username': settings.rhsso.rhsso_user, 'password': settings.rhsso.password}
+            )
+            sleep(360)
```
jyejare, line 1044: Instead of Hard wait time can we use `wait_for` ?

omkarkhatavkar: https://github.com/SatelliteQE/robottelo/pull/7832#issue-434630704 mentioned here about the details.

tests/foreman/ui/test_ldap_authentication.py
omkarkhatavkar: these are all important tests for keycloak feature

pytest_fixtures/satellite_auth.py
omkarkhatavkar: This is tested code, I have just moved from UI folder to common location

tests/foreman/ui/test_ldap_authentication.py
omkarkhatavkar: @devendra104 and @jyejare I can remove them if you guys think it is not needed?

pytest_fixtures/satellite_auth.py
```diff
@@ -0,0 +1,167 @@
...
+from robottelo.constants import LDAP_ATTR
+from robottelo.constants import LDAP_SERVER_TYPE
+from robottelo.datafactory import gen_string
+from robottelo.decorators import fixture
+from robottelo.rhsso_utils import create_mapper
+from robottelo.rhsso_utils import get_rhsso_client_id
+from robottelo.rhsso_utils import run_command
+
+
+@fixture(scope='function')
```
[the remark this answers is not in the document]
omkarkhatavkar: It is a required function, to pass each test

pytest_fixtures/satellite_auth.py
omkarkhatavkar: I don't know much intention about that, this is old code. My PR was about moving the fixture. I consider this unblocking?

tests/foreman/ui/test_ldap_authentication.py
devendra104: > I think we don't run destructive tests for an upgrade yet. @devendra104 Thoughts ?

We have some destructive test cases for upgrade and that executes in every snap.

omkarkhatavkar:
I messed up with git merge issues closing this PR, raise another with similar code.

pytest_fixtures/satellite_auth.py
omkarkhatavkar: Yes we can not have the multiple user groups with the same name across multiple user groups

PR #7844, opened 2020-06-22
Add test for entries per page in settings.

devendra104:
I have added a test case for settings to test the entries_per_page feature.

Validation Point:
- The value should be updated from UI.
- After the update changes should be applied in the Monitors task's page (per page entry and page count should be updated according to the new set value)

Test Result:
```
pytest tests/foreman/ui/test_settings.py::test_positive_entries_per_page

============================================================================================ 1 passed in 74.33 seconds ============================================================================================
```

### DIFF
--- a/tests/foreman/ui/test_settings.py
+++ b/tests/foreman/ui/test_settings.py
@@ -14,6 +14,8 @@
 
 :Upstream: No
 """
+import math
+
 import pytest
 from airgun.session import Session
 from fauxfactory import gen_url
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_settings.py
```diff
@@ -489,3 +489,41 @@ def test_negative_update_hostname_with_empty_fact(session):
...
+        0
+    ]
+    with session:
+        try:
+            session.settings.update(f"name={property_name}", property_value)
+            page_content = session.task.read_all(widget_names="Pagination")
+            assert str(property_value) in page_content["Pagination"]["per_page"]
+            total_pages = int(int(page_content["Pagination"]["total_items"]) / property_value) + (
+                1 if int(page_content["Pagination"]["total_items"]) % property_value else 0
+            )
```
mirekdlugosz, line 526:
```suggestion
            import math
            total_pages = math.ceil(page_content["Pagination"]["total_items"]) / property_value)
```

tests/foreman/ui/test_settings.py
```diff
@@ -489,3 +489,41 @@ def test_negative_update_hostname_with_empty_fact(session):
         finally:
             for key, value in default_hostname.items():
                 setting_cleanup(setting_name=key, setting_value=value.value)
+
+
+@tier3
+def test_positive_entries_per_page(session):
```
mirekdlugosz, line 495: Since you modify global setting, I'd advise to use `run_in_one_thread` decorator.

devendra104:
> Is there bugzilla for issue related to that setting? Did it ever cause any problems for us in any way?
> 
> I feel test like that would be much better suited to upstream unit tests, where it would take couple hundred milliseconds instead of entire minute. Writing to settings itself is so basic operation, that if it doesn't work for any reason, we will probably see multiple errors all over the suite.
> 
> On the other hand, this at least takes only a minute, so it's not _that_ bad from performance point of view.

> Is there bugzilla for issue related to that setting? Did it ever cause any problems for us in any way?
> 
> I feel test like that would be much better suited to upstream unit tests, where it would take couple hundred milliseconds instead of entire minute. Writing to settings itself is so basic operation, that if it doesn't work for any reason, we will probably see multiple errors all over the suite.
> 
> On the other hand, this at least takes only a minute, so it's not _that_ bad from performance point of view.


> Is there bugzilla for issue related to that setting? Did it ever cause any problems for us in any way?
> 
> I feel test like that would be much better suited to upstream unit tests, where it would take couple hundred milliseconds instead of entire minute. Writing to settings itself is so basic operation, that if it doesn't work for any reason, we will probably see multiple errors all over the suite.
> 
> On the other hand, this at least takes only a minute, so it's not _that_ bad from performance point of view.
One RFE is available(BZ#1746221) around it, I will add one more verification step in this test case once this RFE gets fixed.
I didn't see any problem related to this but for the code coverage point of view, this scenario is important.

tests/foreman/ui/test_settings.py
devendra104: Fixed.

tests/foreman/ui/test_settings.py
devendra104: Fixed.

tests/foreman/ui/test_settings.py
```diff
@@ -489,3 +490,42 @@ def test_negative_update_hostname_with_empty_fact(session):
...
+        3. GoTo Monitor > Tasks Table > Pagination
+        4. Check the new per page entry is updated in pagination list
+        5. Check the page count on the basis of the new updated entries per page.
+
+    :expectedresults: New set entry-per-page should be available in the pagination list and
+        page count should match according to the new setting
+
+    :CaseImportance: Medium
+
+    :CaseLevel: Acceptance
```
omkarkhatavkar, line 514: Can we please add the Bugzilla ID here for future reference

devendra104: Fixed

PR #7846, opened 2020-06-22
Fix travis-tox for fixture plugins PR

jyejare:
Fix Travis for PR #7804 .

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,9 +1,12 @@
 [tox]
-envlist = py36, testcov
+envlist = py36
 
 [testenv]
-setenv =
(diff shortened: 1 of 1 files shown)


rochacbruno:
Hi @jyejare 

 execute the following steps:

```
pip uninstall pycurl
export PYCURL_SSL_LIBRARY=openssl
pip install pycurl
```

jyejare:
@rochacbruno I was trying to set that env var using setenv tox configuration. But it seems it was not working. 

I tried exporting it plain bu as you mentioned abovet, tox was throwing an error for export command not availability. So I exported using sh now and it works now.

PR #7855, opened 2020-06-26
Document Report Portal Module and Pytest Rerun based on Report Portal results

jyejare:
PR includes:

- The ReportPortal module Documentation Preview: https://github.com/jyejare/robottelo/tree/document_rerun/robottelo/report_portal#report-portal-module
- The Pytest Pugin Rerun Failure Document Preview: https://github.com/jyejare/robottelo/tree/document_rerun/pytest_plugins/rerun_rp#pytest-plugin---rerun-failures-based-on-report-portal
- The Pytest plugin `uncollection` is now renamed to `rerun_rp` in separate `rerun_rp` folder.


jyejare:
The ReportPortal module Documentation Preview: https://github.com/jyejare/robottelo/tree/document_rerun/robottelo/report_portal#report-portal-module

conftest.py
```diff
@@ -2,7 +2,7 @@
 
 pytest_plugins = [
     # Plugins
-    "pytest_plugins.uncollector",
+    "pytest_plugins.rerun_rp.rerun_rp",
```
omkarkhatavkar, line 5: instead of `rerun_rp` can't we can keep it as `rp` or `report_portal`. making folder as rerun_rp we are limiting it single activity. We could have further functions from the report portal that have different functionality. Just a suggestion, non-blocking from my side

pytest_plugins/rerun_rp/README.adoc
```diff
@@ -0,0 +1,74 @@
+= Pytest Plugin - _Rerun Failures based On Report Portal_
+
+The `pytest_plugins/rerun_rp/reren_rp.py` module is designed as a plugin to the pytest in robottelo test framework. This plugin will run first before the BZ skipping plugin.
+
+This plugin helps https://github.com/SatelliteQE[SatelliteQE] to rerun _passed, failed or skipped_ tests in latest finished ReportPortal launch.
+
+.Some Checks
```
omkarkhatavkar, line 7: Highlight this Some Checks as bold, suggestion

conftest.py
jyejare: Basically, this plugin will only be hooked for rerunning the tests using RP data. If we will have something new to add for RP, it will be added in RP module and not in this plugin.

Even we have something else do to with RP results as pytest plugin, we will have a separate plugin for that. May be something like `block_rp` etc.

jyejare:
@swadeley Need you ACK here !

jyejare:
Two ACKs already. Merging it as its pending from long time and has zero chances of producing any regressions.

PR #7858, opened 2020-07-01
moving the fixtures from common location to global level

omkarkhatavkar:
### PR Details 

- Currently, this PR is the segregate the fixture from UI LDAP component so it can be used Across another project.
- Idle session timeout verification test. Although this works well in 1 min idle timeout in browser but seems like there is some problem with selenium session browser takes some more time to get an idle timeout. Hence I have increased the idle timeout sleep in the test. If you have any suggestion that will work I'm open for that.
- Enhance the test to verify the login should be failed for deleted RHSSO User 

   
### Test Result
```
Testing started at 12:17 PM ...
/home/okhatavk/okhatavk/Satellite/robottelo/env/bin/python /home/okhatavk/Downloads/pycharm-community-2018.1.4/helpers/pycharm/_jb_pytest_runner.py --target test_ldap_authentication.py::test_session_expire_rhsso_idle_timeout
Launching py.test with arguments test_ldap_authentication.py::test_session_expire_rhsso_idle_timeout in /home/okhatavk/Satellite/robottelo/tests/foreman/ui

plugins: forked-1.1.3, xdist-1.31.0, cov-2.8.1, mock-1.10.4, services-1.3.12020-06-17 06:47:41 - conftest - DEBUG - Collected 1 test cases
collected 1 item

========================== 1 passed in 459.11 seconds ==========================
```


pytest_fixtures/satellite_auth.py
```diff
@@ -0,0 +1,153 @@
...
+        tls=False,
+        port='389',
+        organization=[module_org],
+        location=[module_loc],
+    ).create()
+
+
+@fixture(scope='function')
+def auth_source_ipa(module_org, module_loc):
+    ipa_data = {
```
mshriver, line 48: Something to consider here for future updates, especially after the settings to dynaconf change.

We could flatten these two fixtures into a single fixture function that's indirectly parametrized on LDAP/IPA if we simplified the settings lookup defaults for the IPA options.

This would be simple if the settings instances for auth config data had consistent keys, but were nested under their relevant type, like `settings.ldap.username` and `settings.ipa.username` (instead of `settings.ipa.username_ipa`).

Little bit of logic to add ldap/ipa specific options when the parameter includes them and you've got a single flexible fixture.

pytest_fixtures/satellite_auth.py
```diff
@@ -0,0 +1,153 @@
...
+        attr_login=LDAP_ATTR['login_ad'],
+        server_type=LDAP_SERVER_TYPE['API']['ad'],
+        attr_mail=LDAP_ATTR['mail'],
+        name=gen_string('alpha'),
+        host=ldap_data['ldap_hostname'],
+        tls=False,
+        port='389',
+        organization=[module_org],
+        location=[module_loc],
+    ).create()
```
mshriver, line 43: Function scoped fixtures - but no cleanup of the auth source after configuration?

Please consider two things here:
1. Module scoping the fixture for optimization - how many of the auth tests need function scope? What's the cost of setup/teardown of the auth sources
2. Teardown of the auth source - how will these fixtures behave when we're running the entire auth test suite? Are we relying on a behavior in `nailgun.entities.create()` to resolve the fact that we might try to configure the same auth source more than once?

pytest_fixtures/satellite_auth.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+@fixture(scope='session')
+def ldap_auth_name():
+    """Return some random ldap name, and attempt to delete all ldap when test starts.
```
mshriver, line 79: I think this fixture could use a more relevant name to the actions that its taking.

In fact it appears to me to be two separate fixtures, one of which is questionable as a fixture:

1. `def delete_auth_source_users`

A fixture that deletes the users related to auth sources - does this need to be executed for each test that uses the  `auth_source` fixture?  Why only do this at the start of the test session?


2. Random alphanumeric string

This doesn't need to be a fixture, it looks like each of the tests is using this to then call a `create` - that random string can just be called within the test function.  IMO fixtures should be taking _actions_ to setup, not just providing randomized data or data from settings.

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -44,18 +42,33 @@
...
 from robottelo.rhsso_utils import get_rhsso_client_id
 from robottelo.rhsso_utils import run_command
 
-
 pytestmark = [run_in_one_thread]
 
-
 EXTERNAL_GROUP_NAME = 'foobargroup'
 
+ldap_data = {
```
mshriver, line 54: Consider putting these data structures outside of the test module for consistent import and use.  This same dictionary is used within `pytest_fixtures.satellite_auth` for the fixtures.

The fixture could actually return the entity instead of only setting it up, so that these fields can be referenced, or it could return the raw dictionary of values used for setup.

Alternatively these dictionaries could be defined in the fixture module outside of the fixture functions, so that it can be imported. If doing this, make sure that the module can still be read without exception when the settings are missing.

pytest_fixtures/satellite_auth.py
omkarkhatavkar: > Function scoped fixtures - but no cleanup of the auth source after configuration?
> 
> Please consider two things here:
> 
> 1. Module scoping the fixture for optimization - how many of the auth tests need function scope? What's the cost of setup/teardown of the auth sources
All of them need the function scope because we update the test with minor changes with same user from auth source. While if have already login with that user next test would use the same the auth source. This refactor has earlier similar issues because of that stability of tests wasn't there. Hence we added a check to remove everything based on test wise. Creating auth source is from API won't take that much time. 
If I wanted to mimic the same customer behaviour we have only a single auth source in the customer. If we make them both LDAP AD and IPA as module-level then end up with issue connecting two servers.         
      
> 2. Teardown of the auth source - how will these fixtures behave when we're running the entire auth test suite? Are we relying on behavior in `nailgun.entities.create()` to resolve the fact that we might try to configure the same auth source more than once?

This was the wrapper I added to make sure that everything gets clean up. If you think I should be placed below code some common location I can do that.    
https://github.com/omkarkhatavkar/robottelo/blob/cc306ff69f93737abb15c6a7b7dfc2498452efdb/tests/foreman/ui/test_ldap_authentication.py#L149-L168

pytest_fixtures/satellite_auth.py
omkarkhatavkar: I like that idea. It makes sense to have it single fixture.

pytest_fixtures/satellite_auth.py
omkarkhatavkar: > I think this fixture could use a more relevant name to the actions that its taking.
> 
> In fact it appears to me to be two separate fixtures, one of which is questionable as a fixture:
> 
> 1. `def delete_auth_source_users`
> 
> A fixture that deletes the users related to auth sources - does this need to be executed for each test that uses the `auth_source` fixture? Why only do this at the start of the test session?

Yeah, it was old code, Maybe It was started as the LDAP username after a while it was updated the function. I can update the name make sense to me.   
> 1. Random alphanumeric string
> 
> This doesn't need to be a fixture, it looks like each of the tests is using this to then call a `create` - that random string can just be called within the test function. IMO fixtures should be taking _actions_ to setup, not just providing randomized data or data from settings.

This is because we ldap_auth and users generated by the tier 2 tests from API or CLI can cause the data issue. To make sure that we wanted to have a clean set up before going further on UI tests

tests/foreman/ui/test_ldap_authentication.py
omkarkhatavkar: I was thinking the same and put the data structure into constants.py but seems like we can not use settings object in constant.py on that time.

pytest_fixtures/satellite_auth.py
mshriver: For #2 - I see. Yes, this should be in a function scoped fixture that looks like the following.  I would still consider making the function more specific to the ldap sources provided by the fixtures used for the test.  That way the test function stack is doing setup, test execution, and teardown all via linked fixtures.

Considering your responses on the session scoped ldap cleanup fixture below, sounds like you agree with making a change here in the direction I'm suggesting.

The session scoped ldap cleanup fixture below would only do user cleanup _before_ the tests run, not after or between any specific tests.

```
def ldap_cleanup():
    yield
    <iterate over auth sources and delete users>
```

tests/foreman/ui/test_ldap_authentication.py
mshriver: I wouldn't go as far as putting it in constants, but if its needed in the fixtures and the tests, it would be good to DRY the definition.

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -325,7 +170,7 @@ def test_wrapper(test_func, *args, **kwargs):
 
 @ldap_wrapper
```
mshriver, line 171: Defining and ldap cleanup fixture (as discussed above) would remove the need for this wrapper/decorator. Pytest already has the framework to execute the method with the pattern here - without having to define and use this decorator.

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1219,15 +1046,51 @@ def test_external_logout_rhsso(enable_external_auth_rhsso, session):
...
+        with session(login=False):
+            session.rhsso_login.login(
+                {'username': settings.rhsso.rhsso_user, 'password': settings.rhsso.password}
+            )
+            sleep(360)
+            with raises(NavigationTriesExceeded) as error:
+                session.task.read_all(widget_names="current_user")['current_user']
+            assert error.typename == "NavigationTriesExceeded"
+
+    finally:
```
mshriver, line 1078: Consider using `pytest.request.addfinalizer` to actually define this as cleanup, instead of in a try block within the test.

You can decorate a function defined within the test function at ~L1066 after `setting_entity` is defined.

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1254,13 +1117,19 @@ def test_external_new_user_login_and_check_count(session):
             session.browser.refresh()
             count = session.ldapauthentication.read_auth_source_counts('External')
             assert count == current_count + 1
+            # checking delete user can't login anymore
+            delete_rhsso_user(user_details['username'])
```
mshriver, line 1121: Its good to include a test for this behavior within the test function - but should the call to `delete_rhsso_user` be moved out from the `finally` block that it was in?

Previously, if there was an exception during any of the statements above, `delete_rhsso_user` was called - with this change it only gets called if all the statements above are successful.

If this action is going to be included in the test regardless of an exception, its a change to the test behavior on failure - will this potentially impact other tests in the suite?

Ideally the action is still included as a cleanup - via `pytest.request.addfinalizer`, not via this try/finally pattern.

pytest_fixtures/satellite_auth.py
omkarkhatavkar: did you mean to remove all that decorator logic and get back to the fixture which will do all that cleanup? that fixture would be at function scope as the requirement is for the function scope? @mshriver is this my understanding is correct of your comment ?

tests/foreman/ui/test_ldap_authentication.py
omkarkhatavkar: Okay got you here! ignore my above comment

tests/foreman/ui/test_ldap_authentication.py
omkarkhatavkar: Earlier `delete_rhsso_user(user_details['username'])  ` was not part of the test, but now it is . Let me see if I can add that as you mentioned in pytest.request.addfinalizer.

pytest_fixtures/satellite_auth.py
mshriver: Correct, the behavior is actually built into pytest, I think we should use those existing mechanisms, not write our own decorators that re-implement behavior that pytest supports.

tests/foreman/ui/test_ldap_authentication.py
omkarkhatavkar: @mshriver I'm facing issue putting them in constants. if you agree I can put them again in the fixture so that they can be used across all tests. (based on your first comment, let me know if my understanding is correct )

pytest_fixtures/satellite_auth.py
omkarkhatavkar: added here 
https://github.com/SatelliteQE/robottelo/pull/7858/commits/79f816ca7e15e8d2cec4733b767b0655621d4b56#diff-ecddf5d413467e9d1edbc2cd0c22c984R116

jyejare:
# Merged from Android Github.

PR #7862, opened 2020-07-06
Remove redundant and low priority bootstrap tests

swadeley:
Hello

Removing old stubbed tests that are low priority or do not cover some feature with recent bugs.


swadeley:
Thank you

PR #7867, opened 2020-07-13
[Azure-UI] Updated tests based on enhancements

vijay8451:
- Updated tests based on new enhancement 
- Test results:
```
tests/foreman/ui/test_computeresource_azurerm.py::test_positive_end_to_end_azurerm_ft_host_provision PASSED                                                           [ 50%]
tests/foreman/ui/test_computeresource_azurerm.py::test_positive_azurerm_host_provision_ud PASSED                                                                      [100%]

```

### DIFF
--- a/tests/foreman/ui/test_computeresource_azurerm.py
+++ b/tests/foreman/ui/test_computeresource_azurerm.py
@@ -36,13 +36,8 @@ if not setting_is_set('azurerm'):
 
 
 @pytest.fixture(scope='module')
-def module_org(module_azurerm_cr):
-    return module_azurerm_cr.organization[0].read()
-
-
-@pytest.fixture(scope='module')
-def module_loc(module_azurerm_cr):
-    return module_azurerm_cr.location[0].read()
+def module_org():
+    return entities.Organization().create()
 
 
 @pytest.fixture(scope='module')
@@ -61,6 +56,7 @@ def module_azure_cp_attrs(module_azurerm_cr, module_azurerm_finishimg):
             "platform": AZURERM_PLATFORM_DEFAULT,
             "script_command": "touch /var/tmp/text.txt",
             "script_uris": AZURERM_FILE_URI,
+            "volumes_attributes": {"0": {"disk_size_gb": "5"}},
             "interfaces_attributes": {
                 "0": {"public_ip": "Static", "private_ip": "false", "network": nw_id}
             },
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_computeresource_azurerm.py
```diff
@@ -61,6 +61,7 @@ def module_azure_cp_attrs(module_azurerm_cr, module_azurerm_finishimg):
             "platform": AZURERM_PLATFORM_DEFAULT,
             "script_command": "touch /var/tmp/text.txt",
             "script_uris": AZURERM_FILE_URI,
+            "volumes_attributes": {"0": {"disk_size_gb": "5"}},
```
vijay8451, line 64: added attr for additional disk enhancement

tests/foreman/ui/test_computeresource_azurerm.py
```diff
@@ -36,13 +36,13 @@
 
 
 @pytest.fixture(scope='module')
-def module_org(module_azurerm_cr):
-    return module_azurerm_cr.organization[0].read()
+def module_org():
+    return entities.Organization().create()
```
omkarkhatavkar, line 40: As we have already had the module scope fixture for this at https://github.com/SatelliteQE/robottelo/blob/43b38ade8c5f293c20afa43b120d178236767e19/pytest_fixtures/api_fixtures.py#L41 I would think we do not need this now. This is unblocking though.

tests/foreman/ui/test_computeresource_azurerm.py
```diff
@@ -36,13 +36,13 @@
...
 
 
 @pytest.fixture(scope='module')
-def module_loc(module_azurerm_cr):
-    return module_azurerm_cr.location[0].read()
+def module_location(module_org):
```
omkarkhatavkar, line 44: As we have already had the module scope fixture for this at https://github.com/SatelliteQE/robottelo/blob/43b38ade8c5f293c20afa43b120d178236767e19/pytest_fixtures/api_fixtures.py#L41 I would think we do not need this now. This is unblocking though.

tests/foreman/ui/test_computeresource_azurerm.py
vijay8451: Current this one overwriting it hence to avoid the same used module level 
https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/ui/conftest.py#L17

tests/foreman/ui/test_computeresource_azurerm.py
vijay8451: Good catch !!  this one not needed and removed in latest push.

PR #7872, opened 2020-07-15
Fix test cases in Polarion that are misrepresented

hkx303:
**Description**
Some virt-who plugin cases(API and CLI cases) in Polarion show Not Automated in Automation label, those are misrepresented, we have automated them.


hkx303:
> ACK

wow, thank you for your quick response!!

PR #7882, opened 2020-07-20
Add RH Cloud - Inventory E2E test

mirekdlugosz:
```
$ pytest -v tests/foreman/ui/test_rhcloud_inventory.py
platform linux -- Python 3.8.3, pytest-4.6.3, py-1.8.1, pluggy-0.13.1 -- /home/mzalewsk/.virtualenvs/robottelo/bin/python3
cachedir: .pytest_cache
rootdir: /home/mzalewsk/sources/robottelo
plugins: cov-2.10.0, mock-1.10.4, forked-1.2.0, services-1.3.1, xdist-1.32.0
collecting ... 2020-07-20 11:50:03 - conftest - DEBUG - Collected 1 test cases
collected 1 item                                                                                                                                                                              

tests/foreman/ui/test_rhcloud_inventory.py::test_rhcloud_inventory_e2e PASSED                                                                                                           [100%]

================================================================================= 1 passed in 327.00 seconds ==================================================================================
```

### DIFF
--- a/tests/foreman/ui/test_rhcloud_inventory.py
+++ b/tests/foreman/ui/test_rhcloud_inventory.py
@@ -1,0 +1,186 @@
+"""Tests for RH Cloud - Inventory, also known as Insights Inventory Upload
+
+:Requirement: RH Cloud - Inventory
+
+:CaseAutomation: Automated
+
+:CaseLevel: System
+
+:CaseComponent: RHCloud-Inventory
+
+:TestType: Functional
+
+:CaseImportance: High
+
+:Upstream: No
+"""
+import hashlib
+import json
+import os
+import tarfile
+
+import pytest
+from nailgun import entities
+
+from robottelo import manifests
+from robottelo import ssh
+from robottelo.api.utils import upload_manifest
+from robottelo.constants import DEFAULT_SUBSCRIPTION_NAME
+from robottelo.constants import DISTRO_RHEL7
+from robottelo.vm import VirtualMachine
+
+
+def get_host_counts(tarobj):
+    metadata_counts = {}
+    slices_counts = {}
+    for file_ in tarobj.getmembers():
+        file_name = os.path.basename(file_.name)
+        if not file_name.endswith(".json"):
+            continue
+        json_data = json.load(tarobj.extractfile(file_))
+        if file_name == "metadata.json":
+            metadata_counts = {
+                f"{key}.json": value['number_hosts']
+                for key, value in json_data['report_slices'].items()
+            }
+        else:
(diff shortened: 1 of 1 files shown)


mirekdlugosz:
@san7ket FYI, this is E2E test for Satellite. Compared to test plan outlined in "rh_cloud smoke test suite plan" document:
* I don't check if number of hosts in metadata matches total number of hosts in slice files. Is this something you check, or can check, on swatch side?
* I don't check if report is re-generated after 24 hours. One, it's awkward thing to do in Robottelo. Two, auto-upload defaults to off since plugin version x.0.9.

@eko999 @Junefen  FYI, this is test we have discussed earlier today. To generate report, all you really need is `session.cloudinventory.generate_report(org.name)`. You can then `session.cloudinventory.download_report(org.name)` to download report using Airgun, or just SSH into Satellite machine and fetch it. When using SSH, note that there are two paths that need to be checked - there used to be a bug when moving report file to `done` didn't work properly. See `get_remote_report_checksum` function.

san7ket:
@eko999 @Junefen @mirekdlugosz I also looked into the plugin documentation and looks like we have a way to trigger this via CLI.
Just tested on satellite server and seems to work fine. Hope this helps

Report can be generated via CLI on Satellite
```
export organization_id=1
export target=/var/lib/foreman/red_hat_inventory/generated_reports/
/usr/sbin/foreman-rake foreman_inventory_upload:report:generate
```

san7ket:
@mirekdlugosz  Validation of data in report vs hosts data in satellite would still be Satellite plugin tests. 
Did the hosts make it to report would be easier to test here rather than at later stage. As, if hosts didn't make it report, they would never make it to cloud anyway.

mirekdlugosz:
@san7ket I added couple new assertions, see bottom of file now. I check if JSON files in tar have the same names as metadata, and that number of hosts in each slice is the same as metadata says it should be. I believe this gets us covered, let me know if you would do it differently.

For E2E test on Satellite side, I prefer to stay with UI for interactions. We expect that customers will mainly interact with plugin through UI, or configure it and rely on automatic daily uploads. But rake task looks like good way for virt-who team, as it's faster and doesn't require pulling in and maintaining browser UI automation layer. Good find!

san7ket:
LGTM, Thanks for all the work on this, We just need a way now to export the report. Let's discuss on this more offline.

mirekdlugosz:
Merging, as it's been 2 weeks already and now it has both tier1 and tier2 reviewer ACK

san7ket:
@mirekdlugosz Can you tag me in cherry picks awell.

mirekdlugosz:
@san7ket sure, once I start working on them.

PR #7883, opened 2020-07-20
adding stubbed test case for SLES 6.8 feature

tstrych:
There should be 4 SLES distro versions, 
I want run `test_positive_report_package_installed_removed` for all of them. `[11.4,12.3,12.4,15.1]`.

If you need any additional info to approve this commit please let me know. 
I am definitely not sure by all of the tags that should be used here.

### DIFF
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -2555,3 +2555,62 @@ class EncDumpTestCase(CLITestCase):
         """
         hostname = ssh.command('hostname').stdout[0]
         self.assertIsInstance(Host.enc_dump({'name': hostname}), list)
+
+
+class SLESTestCase(CLITestCase):
+    """ Tests SUSE Linux Enterprise Server Client """
+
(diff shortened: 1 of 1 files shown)


tstrych:
The main part of it is the setup. The test it self is not importnat I just want to check if the registered host works properly. 
Anyway, once the setup will work correctly. I can create some general install package on client :)

JacobCallahan:
with upcoming improvements, we'll have to keep in mind different package managers to make installation and other actions platform-independent at the test level. If that is the case, then rolling these tests into existing content host tests would be the best option.

tstrych:
@JacobCallahan can you please bring more clearance to what should I change in these stubs ? 

Put these stubs into [contenthost](https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/ui/test_contenthost.py) ?  Only UI tests are directly differentiated to contenthost test.

tstrych:
I have two ack's so I am merging it :)

PR #7887, opened 2020-07-21
Include tests as parametrized in polarion

jyejare:
Good News!

Your Parametrized tests will now be included as `parametrized` in polarion. In Polarion's term, it would be called as `test with iterations`.

This PR includes:

- [x] A token `parametrized` for parametrizing tests in `api/test_user.py` that will make these tests iterable in polarion.

- [x] Port `idgen` parametrized tests to `**parametrized`.

For results:

1. Go to `Stage Polarion`.
2. Visit, Test Run `Satellite test_with_param - parallel - Tier 1`.
3. Verify you can see `8 parameterized results` from `2 tests`, only 1 of that test is parametrized.
4. Visit `test_positive_test_params_gadar` parametrized tests for iterations.

Hureyy!

### DIFF
--- a/testimony.yaml
+++ b/testimony.yaml
@@ -148,3 +148,10 @@ Upstream:
     - 'No'
     required: true
     type: choice
+Parametrized:
+    casesensitive: false
+    choices:
+    - 'Yes'
+    - 'No'
+    required: false
+    type: choice
--- a/tests/foreman/api/test_contentview.py
+++ b/tests/foreman/api/test_contentview.py
@@ -45,6 +45,7 @@ from robottelo.constants import PUPPET_MODULE_NTP_PUPPETLABS
 from robottelo.constants import REPOS
 from robottelo.constants import REPOSET
 from robottelo.datafactory import invalid_names_list
+from robottelo.datafactory import parametrized
 from robottelo.datafactory import valid_data_list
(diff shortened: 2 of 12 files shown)


testimony.yaml
```diff
@@ -148,3 +148,9 @@ Upstream:
     - 'No'
     required: true
     type: choice
+Parametrized:
+    casesensitive: false
+    choices:
+    - 'Yes'
```
lpramuk, line 154: `- 'No'`
isn't no missing ?

mshriver:
There were some offline conversations about the scope and impact of this change, and the short term needs.

I'm happy to accept this change if the PR includes token updates for all test cases that are parametrized within the framework.

Parametrization through fixtures will make detecting this and applying the token automatically will be challenging, @jyejare will continue to investigate how testimony or betelgeuse could be modified to detect and add this.

jyejare:
@mshriver @JacobCallahan @lpramuk Added token for all the test modules now. Please ACK and merge if no objection.

jyejare:
@mshriver @JacobCallahan @lpramuk Also all `idgen` are moved to `**parametrized`, so that the data dictionary keys will be appended to the test names in pytest collection/ polarion iterations.

PR #7891, opened 2020-07-22
rename medium to media

tstrych:
Making the names of test_media consistent between ui api and cli tests.


swadeley:
Hello

My first thought was why only change file name, but I see its not a simple case.

In `tests/foreman/api/test_media.py` I see:

    Tests for the ``media`` paths.

    :Requirement: Media

whereas in `tests/foreman/cli/test_medium.py` I see:

    Test for Medium  CLI

    :Requirement: Medium

Seems the name comes from the hammer commands, e.g. `hammer location add-medium`. So the reason for the change is not compelling for me.

tstrych:
Thanks for feedback, and you are right, I will ask developers for reasoning behind that. And if I am right I will create bug. Anyway I should close this, maybe reopen after that bug would be resolved.

tstrych:
Developers explain it: hammer uses singular names of entities.

PR #7894, opened 2020-07-23
change of CaseComponent from API and Navigation to Search

tstrych:
Reasoning behind this:

Navigation bugs are not directly connected to bookmarks.
The nearest component to bookmarks is Search. 

It doesn't make sense to have two different components for bookmark.
Better solution is to have same component for all tests (api/cli/ui/upgrades). 

Please let me know if you have different view on this problem.

I would be happy to get agreement from @lvrtelov as she is owner of Search component.


lvrtelov:
ACK :)

PR #7896, opened 2020-07-23
Dynflow 6.8 feature test

lhellebr:


tests/foreman/sys/test_dynflow.py
```diff
@@ -0,0 +1,41 @@
...
+        :expectedresults: Correct dynflow processes are running, respecting settings
+
+        :CaseImportance: High
+        """
+        ssh.command(
+            "cd /etc/foreman/dynflow/ && cp worker-hosts-queue.yml test.yml && sed -i s/5/6/"
+            "test.yml && systemctl restart 'dynflow-sidekiq@test' && res=1 && while"
+            "[ $res -eq 1 ]; do systemctl status 'dynflow-sidekiq@test' -l | "
+            "grep ' of 6 busy' && res=$(echo $?); done"
+        )  # if thread count is not respected or the process is not running, this should timeout
```
omkarkhatavkar, line 41: No Python Assertion? dont you think good to check the return code? you are managing everything in the shell itself.

lhellebr: I think there is nothing to check - the correct thread count either is there and the script doesn't time out or it is not there after certain waiting (depends on how long ssh.command is set up to wait) and the script times out.

tests/foreman/sys/test_dynflow.py
```diff
@@ -0,0 +1,41 @@
...
+
+        :expectedresults: Correct dynflow processes are running, respecting settings
+
+        :CaseImportance: High
+        """
+        ssh.command(
+            "cd /etc/foreman/dynflow/ && cp worker-hosts-queue.yml test.yml && sed -i s/5/6/"
+            "test.yml && systemctl restart 'dynflow-sidekiq@test' && res=1 && while"
+            "[ $res -eq 1 ]; do systemctl status 'dynflow-sidekiq@test' -l | "
+            "grep ' of 6 busy' && res=$(echo $?); done"
```
mirekdlugosz, line 40: This kind of loop can be simplified to:
```
while ! systemctl status 'dynflow-sidekiq@test' -l | grep -q ' of 6 busy' ; do sleep 0.5 ; done
```

Shouldn't you clean up after yourself, i.e. remove test.yml file?

Could we put individual commands into list and just join them on '&&' when calling? I think it will be little easier to read. Something like:
```python
commands = [
    "cd /etc/foreman/dynflow/",
    "cp worker-hosts-queue.yml test.yml"
]
ssh.command(" && ".join(commands))
```

lhellebr: Good points, I'll do that

lhellebr: Done

lhellebr:
@omkarkhatavkar , can you ACK this pls?

omkarkhatavkar (approved):
ACK, sorry late reply here. somehow missed this PR thread

PR #7899, opened 2020-07-24
canned role fix

pondrejk:
Fixes intermittent key error affecting canned role testclass
```
pytest tests/foreman/api/test_role.py::CannedRoleTestCases
platform linux -- Python 3.7.7, pytest-4.6.3, py-1.7.0, pluggy-0.12.0
rootdir: /home/pondrejk/Documents/robottelo
plugins: repeat-0.8.0, cov-2.10.0, mock-1.10.4, forked-0.2, services-1.3.1, xdist-1.33.0
collecting ... 2020-07-24 09:46:33 - conftest - DEBUG - Collected 42 test cases
collected 42 items                                                                                                   

tests/foreman/api/test_role.py ss....s...F.....s........s................                                      [100%]

```
One failure unrelated

### DIFF
--- a/tests/foreman/api/test_role.py
+++ b/tests/foreman/api/test_role.py
@@ -112,6 +112,8 @@ class CannedRoleTestCases(APITestCase):
                 'role': {'name': name, 'organization_ids': orgs or [], 'location_ids': locs or []}
             }
         )
+        if 'role' in org_admin:
+            return entities.Role(id=org_admin['role']['id']).read()
         return entities.Role(id=org_admin['id']).read()
 
     def create_org_admin_user(self, role_taxos, user_taxos, same_taxos):


tests/foreman/api/test_role.py
```diff
@@ -112,6 +112,8 @@ def create_org_admin_role(self, name=None, orgs=None, locs=None):
                 'role': {'name': name, 'organization_ids': orgs or [], 'location_ids': locs or []}
             }
         )
+        if 'role' in org_admin:
+            return entities.Role(id=org_admin['role']['id']).read()
```
mirekdlugosz, line 116: Do we know why `role` key is present only sometimes? What does it depend on?

lhellebr: If it's not consistent, it seems like either a product bug or Nailgun bug. Won't this fix just hide it?

pondrejk: @mirekdlugosz @lhellebr sorry for delay, it is hard to reproduce this outside of our automation environment, though there it occurs regularly. Fwict the inconsistency is connected to multi-threaded run. Alternative candidate solution would be wrapping the class into @run_in_one_thread

mirekdlugosz: @pondrejk I think I would prefer `@run_in_one_thread` in this case. Seems to be cleaner solution.
I would like to wait for @lhellebr opinion, since he also commented. If Lukas is OK with current implementation, please go ahead and merge. I know these kind of failures are hard to track down, reproduce and fix.

mirekdlugosz (approved):

PR #7905, opened 2020-07-28
[Pytest Way] Stubbed tests as skipped or passed tests

jyejare:
Earlier we used to Return `UnitTest.skip` to skip the test which has a `stubbed` marker. That was a UnitTest way.

1. Now, This PR changes it to pyTest way of doing the same and as a pytest plugin to skip/pass the Test.

- The option `mark-manuals-skipped` to run a plugin mark all stubbed tests as `skipped` and its by default True.

- The option `mark-manuals-passed` to run a plugin mark all stubbed tests as` passed` and its by default False.

2. This PR also adds stubbed marked as a config part of plugin.

- So removed it from the decorator's module.

3. So, all the test modules have replaced `@stubbed` marker from the decorators module with `@pytest.mark.stubbed` from pytest module.

### DIFF
--- a/pytest_plugins/manual_skipped.py
+++ b/pytest_plugins/manual_skipped.py
@@ -1,0 +1,42 @@
+import logging
+
+import pytest
+
+from robottelo.constants import NOT_IMPLEMENTED
+
+LOGGER = logging.getLogger(__name__)
+
+
+def pytest_configure(config):
+    """Register custom marker for stubbed test cases."""
+    marker = "stubbed: Tests that are not automated yet or manual only."
+    config.addinivalue_line("markers", marker)
+
+
+def pytest_collection_modifyitems(session, config, items):
+    """
+    Collects tests based on pytest option to select tests marked as failed on Report Portal
+    """
+    mark_passed = config.getvalue('mark_manuals_passed')
+    mark_skipped = config.getvalue('mark_manuals_skipped')
+    if mark_passed or (not mark_skipped):
+        return
+    for item in items:
+        if item.get_closest_marker(name='stubbed'):
+            item.add_marker(marker=pytest.mark.skip(reason=NOT_IMPLEMENTED))
+
+
+def pytest_addoption(parser):
+    """Add options to pytest to modify manual tests as skipped"""
+    parser.addoption(
+        "--mark-manuals-skipped",
+        action='store_true',
+        default=True,
+        help='Mark stubbed tests as skipped tests.',
+    )
+    parser.addoption(
+        "--mark-manuals-passed",
+        action='store_true',
+        default=False,
+        help='Mark stubbed tests as passed tests.',
+    )
--- a/tests/foreman/api/test_activationkey.py
+++ b/tests/foreman/api/test_activationkey.py
@@ -16,6 +16,7 @@
 """
 import http
 
+import pytest
 from fauxfactory import gen_integer
 from fauxfactory import gen_string
 from nailgun import client
@@ -34,7 +35,6 @@ from robottelo.datafactory import invalid_names_list
 from robottelo.datafactory import valid_data_list
 from robottelo.decorators import run_in_one_thread
 from robottelo.decorators import skip_if_not_set
-from robottelo.decorators import stubbed
 from robottelo.decorators import tier1
 from robottelo.decorators import tier2
 from robottelo.decorators import tier3
@@ -483,7 +483,7 @@ class ActivationKeyTestCase(APITestCase):
     @upgrade
     @skip_if_not_set('fake_manifest')
     @tier2
-    @stubbed()
+    @pytest.mark.stubbed
     def test_positive_add_future_subscription(self):
         """Add a future-dated subscription to an activation key.
 
--- a/conftest.py
+++ b/conftest.py
@@ -5,6 +5,7 @@ pytest_plugins = [
     "pytest_plugins.rerun_rp.rerun_rp",
     "pytest_plugins.markers",
     "pytest_plugins.issue_handlers",
+    "pytest_plugins.manual_skipped",
     # Fixtures
     "pytest_fixtures.api_fixtures",
     # Component Fixtures
--- a/pytest_plugins/markers.py
+++ b/pytest_plugins/markers.py
@@ -4,7 +4,6 @@
 def pytest_configure(config):
     """Register custom markers to avoid warnings."""
     markers = [
-        "stubbed: Tests that are not automated yet.",
         "deselect(reason=None): Mark test to be removed from collection.",
         "skip_if_open(issue): Skip test based on issue status.",
         "tier1: Tier 1 tests",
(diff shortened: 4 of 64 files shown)


pytest_plugins/manual_skipped.py
```diff
@@ -0,0 +1,26 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+
+def pytest_collection_modifyitems(items, config):
+    """
+    Collects tests based on pytest option to select tests marked as failed on Report Portal
+    """
+    manualskip_args = config.getvalue('mark_manuals_skipped')
+    if not manualskip_args:
```
devendra104, line 15: Just suggestion, Can we do it this way?
```
if config.getvalue('mark_manuals_skipped'):
   for item in items:
        if item.get_closest_marker(name='stubbed'):
           item.add_marker(marker=pytest.mark.skip(reason=NOT_IMPLEMENTED))
else:
     return
```

pytest_plugins/manual_skipped.py
```diff
@@ -0,0 +1,26 @@
...
+
+from robottelo.constants import NOT_IMPLEMENTED
+
+LOGGER = logging.getLogger(__name__)
+
+
+def pytest_collection_modifyitems(items, config):
+    """
+    Collects tests based on pytest option to select tests marked as failed on Report Portal
+    """
```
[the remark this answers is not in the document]
mshriver: This is actually dictated by pytest, as its a hook implementation and the function won't be called directly by any part of the framework. I don't think its necessary to document the arg definitions here.

I would prefer that we maintain the function signature to the API ref, even if we aren't using the `session`. 

```python
def pytest_collection_modifyitems(session, config, items):
```

pytest_plugins/manual_skipped.py
jyejare: I think both its the same and the suggestion won't be helping much in any manner.

But the current implementation quickly exits out of code if the marker is not mentioned in the pytest. Putting else loop unnecessarily increasing the conditioning.

But still I am buying your removal of stubbed var idea.

jyejare:
@devendra104 @mshriver Please revisit, comments has been addressed.

pytest_plugins/manual_skipped.py
```diff
@@ -0,0 +1,25 @@
...
+        return
+    for item in items:
+        if item.get_closest_marker(name='stubbed'):
+            item.add_marker(marker=pytest.mark.skip(reason=NOT_IMPLEMENTED))
+
+
+def pytest_addoption(parser):
+    """Add options to pytest to modify manual tests as skipped"""
+    help_script = 'Mark stubbed tests as skipped tests.'
+    parser.addoption("--mark-manuals-skipped", nargs='?', const=True, help=help_script)
```
mshriver, line 25: Why use `nargs` and `const` here, instead of just `store_true` to create a simple flag?

mshriver:
This implementation will cause warnings because the mark is unregistered, and is replacing the marker name itself:

```
.robottelo/lib64/python3.6/site-packages/_pytest/mark/structures.py:337
  /home/setup/repos/robottelo/.robottelo/lib64/python3.6/site-packages/_pytest/mark/structures.py:337: PytestUnknownMarkWarning: Unknown pytest.mark.This is a Manual test! - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html
    PytestUnknownMarkWarning,

```

jyejare:
@mshriver Thanks for the comments. Comments addressed.

jyejare:
@mshriver @devendra104 @JacobCallahan All the infected files by `@stubbed()` marker replaced with `@stubbed`.

jyejare:
@mshriver These comments are similar to what we are getting for issue handlers plugin and we can simply ignore them.

mshriver:
@jyejare can you be more specific in your response? What messages are we getting from issue_handlers plugin?

I don't think we should be ignoring unknown mark warnings, our markers should be predictable and registered with pytest.

pytest_plugins/manual_skipped.py
```diff
@@ -0,0 +1,31 @@
...
+        return
+    for item in items:
+        if item.get_closest_marker(name='stubbed'):
+            item.add_marker(marker=pytest.mark.skip(reason=NOT_IMPLEMENTED))
+
+
+def pytest_addoption(parser):
+    """Add options to pytest to modify manual tests as skipped"""
+    help_script = 'Mark stubbed tests as skipped tests.'
+    parser.addoption("--mark-manuals-skipped", action='store_true', default=True, help=help_script)
```
mshriver, line 31: Discussed with @jyejare updating this option to allow for controlling the stubbed case results between passing and skipping.  Default should be skipping, as that's the behavior that we want when these cases are included in collection by default.  The option here should reflect that by having a default boolean that will skip when there is no option passed. Passing the option should trigger the non-default behavior of `passed` results for stubbed cases.

jyejare: @mshriver Done! 

All the files are also replaced to use `stubbed` marker from pytest directly instead of from `robottelo.decorators`

jyejare:
@mshriver @devendra104 ,

Results:

Manuals Skipped:
```
$ pytest -m stubbed tests/foreman/api/test_role.py --mark-manuals-skipped --disable-warnings
=============== test session starts ================
platform linux -- Python 3.7.6, pytest-4.6.3, py-1.8.0, pluggy-0.12.0
plugins: forked-1.0.2, picked-0.4.4, services-1.3.1, cov-2.9.0, mock-1.10.4, xdist-1.32.0
collecting ... 2020-08-18 08:09:23 - conftest - DEBUG - Collected 47 test cases
collected 47 items / 41 deselected / 6 selected                                                                                                                                                                    

tests/foreman/api/test_role.py ssssss                                                                                                                                                                        [100%]

===================== 6 skipped, 41 deselected  in 13.28 seconds =================
```

Manuals Passed:
```
$ pytest -m stubbed tests/foreman/api/test_role.py --mark-manuals-passed --disable-warnings
===================== test session starts ==========================
platform linux -- Python 3.7.6, pytest-4.6.3, py-1.8.0, pluggy-0.12.0
plugins: forked-1.0.2, picked-0.4.4, services-1.3.1, cov-2.9.0, mock-1.10.4, xdist-1.32.0
collecting ... 2020-08-18 08:10:43 - conftest - DEBUG - Collected 47 test cases
collected 47 items / 41 deselected / 6 selected                                                         

tests/foreman/api/test_role.py ......                                                             [100%]

=================== 6 passed, 41 deselected in 45.48 seconds ================
```

tests/foreman/api/test_activationkey.py
```diff
@@ -483,7 +483,7 @@ def test_positive_fetch_product_content(self):
     @upgrade
     @skip_if_not_set('fake_manifest')
     @tier2
-    @stubbed()
+    @pytest.mark.stubbed
```
mshriver, line 486: :clap: I prefer this direct marking syntax over obfuscated marks that have to be imported, and don't provide any wrapper function!

devendra104 (approved):
ACK

PR #7908, opened 2020-07-29
Fix test_positive_assign_http_proxy_to_products

jameerpathan111:
Fixes tests/foreman/api/test_product.py::ProductTestCase::test_positive_assign_http_proxy_to_products

### DIFF
--- a/tests/foreman/api/test_product.py
+++ b/tests/foreman/api/test_product.py
@@ -28,7 +28,6 @@ from nailgun import entities
 from requests.exceptions import HTTPError
 
 from robottelo import manifests
-from robottelo import ssh
 from robottelo.api.utils import upload_manifest
 from robottelo.constants import FAKE_1_PUPPET_REPO
 from robottelo.constants import FAKE_1_YUM_REPO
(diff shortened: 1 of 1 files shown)


mirekdlugosz:
@jameerpathan111 flake8 in travis complains about imported but unused module.

Also, can you explain how it fixes the test? Did Satellite stop putting fqdn in log during sync? Was that expected change? Is there another log file we should maybe look into instead?

jameerpathan111:
> Also, can you explain how it fixes the test? Did Satellite stop putting fqdn in log during sync? Was that expected change? Is there another log file we should maybe look into instead?

The thing is, the test passes locally and I can see the fqdn in log as well but on automation m/c it fails for unknown reason.
And I think checking log for fqdn is not that important and we can skip that part till we find root cause. we have other tests which use real proxy to sync content so this change shouldn't cause any problem.

@mirekdlugosz what do you think we should do?

jameerpathan111:
@mirekdlugosz as per our discussion offline, I will try to check what's wrong with automation m/c and see if I can fix the issue.

jyejare:
@jameerpathan111 Whats ur plan on this ? Is the test still failing ? if yes, would you like to continue on this PR ?

Else lets close it ?

JacobCallahan:
@jameerpathan111 can you revisit this PR and see if the issue persists in latest automation? If so, can you update the PR? Thanks

jameerpathan111:
@JacobCallahan I no longer see this issue in latest automation run.

JacobCallahan:
> @JacobCallahan I no longer see this issue in latest automation run.

Perfect, thanks for closing!

PR #7941, opened 2020-08-19
Refactor BZ Markers

mshriver:
Register the BZ marker, and update issue_handlers to pass it a list of all the BZ's marked through skip_if_open decorator.

Add a --BZ option to maintain the current support for test case filtering on specific BZs. The argument can be specified multiple times to filter for multiple BZs.

Verbosely log when a testcase is deselected based on these filters and BZ marks so its obvious why a test is or isn't included in collection, and what marks apply.

Tested with existing test cases that have singular and multiple skip_if_open mark decorators applied.
Tested with singular and multiple BZ filters passed on CLI.
Collection and logging were as expected.

Further refactoring could simplify the use of the `skip_if_open` decorator, building the skip behavior into a direct BZ marker.  The tests would have to be updated with a new mark decorator, and the hook updated to not apply BZ marks, but read them to apply skipif marks, which it already is doing.

### DIFF
--- a/pytest_plugins/issue_handlers.py
+++ b/pytest_plugins/issue_handlers.py
@@ -16,13 +16,17 @@ from robottelo.utils.issue_handlers import should_deselect
 from robottelo.utils.version import search_version_key
 from robottelo.utils.version import VersionEncoder
 
-LOGGER = logging.getLogger('issue_handlers_plugin')
+LOGGER = logging.getLogger('robottelo')
 
 DEFAULT_BZ_CACHE_FILE = 'bz_cache.json'
 
 
 def pytest_addoption(parser):
-    """Adds custom options to pytest runner."""
+    """ Add CLI options related to issue handlers
+
+    Add a --bz-cache option to control use of a BZ cache from a local JSON file
+    Add a --BZ option for filtering BZ marked testcases
+    """
     parser.addoption(
         "--bz-cache",
         action='store_true',
@@ -30,6 +34,26 @@ def pytest_addoption(parser):
         f"Without this flag a cache file will be created with the name {DEFAULT_BZ_CACHE_FILE}."
         "This default cache file can be used on subsequent runs by including this flag",
     )
+    parser.addoption(
+        "--BZ",
+        help="Filter test collection based on BZ markers. "
+        "BZ markers are applied automatically by skip_if_open marks on tests. "
+        "Comma separated list to include multiple BZs for collection."
+        " Example: `--BZ 123456,456789`",
+    )
+
+
+def pytest_configure(config):
+    """Register custom markers to avoid warnings."""
+    markers = [
+        "skip_if_open(issue): Skip test based on issue status.",
+        (
+            "BZ(number): Bugzillas related to the testcase, "
+            "for use with `--BZ <number>` option for collection"
+        ),
+    ]
+    for marker in markers:
+        config.addinivalue_line("markers", marker)
 
 
 @pytest.hookimpl(trylast=True)
@@ -42,13 +66,39 @@ def pytest_collection_modifyitems(session, items, config):
 
     pytest.issue_data = generate_issue_collection(items, config)
 
-    # No item deselect, just adding skipif markers
+    # Add skipif markers, and modify collection based on --bz option
+    bz_filters = config.getoption('BZ', None)
+    if bz_filters:
+        bz_filters = bz_filters.split(',')
+    selected = []
+    deselected = []
     for item in items:
+        # Add a skipif marker for the issues
         skip_if_open = item.get_closest_marker('skip_if_open')
         if skip_if_open:
             # marker must have `BZ:123456` as argument.
             issue = skip_if_open.kwargs.get('reason') or skip_if_open.args[0]
             item.add_marker(pytest.mark.skipif(is_open(issue), reason=issue))
+
+        # remove items from collection
+        if bz_filters:
+            # Only include items which have BZ mark that includes any of the filtered bz numbers
+            item_bz_marks = set(getattr(item.get_closest_marker('BZ', None), 'args', []))
+            if bool(set(bz_filters) & item_bz_marks):
+                selected.append(item)
+            else:
+                LOGGER.debug(
+                    f'Deselected test [{item.nodeid}] due to BZ filter {bz_filters} '
+                    f'and available marks {item_bz_marks}'
+                )
+                deselected.append(item)
+                continue  # move to the next item
+            # append only when there is a BZ in common between the lists
+        else:
+            selected.append(item)
+
+    config.hook.pytest_deselected(items=deselected)
+    items[:] = selected
 
 
 IS_OPEN = re.compile(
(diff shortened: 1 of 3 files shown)


pytest_plugins/issue_handlers.py
```diff
@@ -16,20 +16,44 @@
 from robottelo.utils.version import search_version_key
 from robottelo.utils.version import VersionEncoder
 
-LOGGER = logging.getLogger('issue_handlers_plugin')
+LOGGER = logging.getLogger('robottelo')
```
mshriver, line 19: There is no specific configuration in logging.yaml for a logger for this file, and I don't necessarily think it needs its own?

This results in the log messages I added in the PR being console and file printed during execution, after the pytest warnings, with the default logging config. I could understand not wanting this change to logging verbosity during collection, but it only logs when BZ filtering is active.

pytest_plugins/issue_handlers.py
```diff
@@ -42,14 +66,39 @@ def pytest_collection_modifyitems(session, items, config):
 
     pytest.issue_data = generate_issue_collection(items, config)
 
-    # No item deselect, just adding skipif markers
+    # Add skipif markers, and modify collection if there were BZ options passed
```
jyejare, line 69: s/passed/provided ?

Or


```suggestion
    # Add skipif markers, and modify collection based on BZ option.
```

pytest_plugins/issue_handlers.py
```diff
@@ -42,14 +66,39 @@ def pytest_collection_modifyitems(session, items, config):
...
+    bz_filters = config.getoption('BZ', None)
```
jyejare, line 70: I m sure by default it should  be returning None, else we can set the default to none from pytest_addoption itself.

pytest_plugins/issue_handlers.py
```diff
@@ -42,14 +66,39 @@ def pytest_collection_modifyitems(session, items, config):
...
         skip_if_open = item.get_closest_marker('skip_if_open')
         if skip_if_open:
             # marker must have `BZ:123456` as argument.
             issue = skip_if_open.kwargs.get('reason') or skip_if_open.args[0]
             item.add_marker(pytest.mark.skipif(is_open(issue), reason=issue))
 
+        # remove items from collection
+        if bz_filters:
+            # Only include items which have BZ mark that includes any of the filtered bz numbers
+            item_marked_bzs = getattr(item.get_closest_marker('BZ', None), 'args', [])
```
jyejare, line 84: Can we rename this var to something better ?

Something  like bz_marked_items ?

pytest_plugins/issue_handlers.py
```diff
@@ -42,14 +66,39 @@ def pytest_collection_modifyitems(session, items, config):
...
     for item in items:
+        # Add a skipif marker for the issues
         skip_if_open = item.get_closest_marker('skip_if_open')
         if skip_if_open:
             # marker must have `BZ:123456` as argument.
             issue = skip_if_open.kwargs.get('reason') or skip_if_open.args[0]
             item.add_marker(pytest.mark.skipif(is_open(issue), reason=issue))
 
+        # remove items from collection
+        if bz_filters:
```
jyejare, line 82:
```suggestion
        if bzys:
```

Bz filters sounds like something to do with bugzilla.

pytest_plugins/issue_handlers.py
```diff
@@ -42,14 +66,39 @@ def pytest_collection_modifyitems(session, items, config):
...
+            # Only include items which have BZ mark that includes any of the filtered bz numbers
+            item_marked_bzs = getattr(item.get_closest_marker('BZ', None), 'args', [])
+            filter_and_mark_match = bool(set(bz_filters) & set(item_marked_bzs))
+            if item_marked_bzs and filter_and_mark_match:
```
jyejare, line 86: I think we need not to check for item_marked_bzs condition again here  since we have to done that in previous stage.

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -64,7 +64,7 @@ def pytest_collection_modifyitems(items, config):
     """
     Collects tests based on pytest option to select tests marked as failed on Report Portal
     """
-    fail_args = config.getvalue('only_failed')
+    fail_args = config.getoption('only_failed', False)
```
jyejare, line 67: Is this really required to set False. Isnt it auto understood ?

Since I have tested this plugin with and without  option, it works great for me without setting it to False.

jyejare (commented):
Nic improvements overall.

Still, I think --bz should take multiple bz's(comma separated) to filter the tests instead of using `append`. Since mentioning that option for each bz and on cli would be painful. Not a blocking comment though.

Ack with pending  comments/ questions.

pytest_plugins/issue_handlers.py
```diff
@@ -187,8 +236,8 @@ def generate_issue_collection(items, config):  # pragma: no cover
 
                 # Store issue key to lookup in the deselection process
                 deselect_data[item.location] = issue_key
-                # Add issue as a marker to enable filtering e.g: "-m BZ_123456"
-                item.add_marker(getattr(pytest.mark, issue_key.replace(':', '_')))
+                # Add issue as a marker to enable filtering e.g: "-BZ 123456"
```
JacobCallahan, line 239:
```suggestion
                # Add issue as a marker to enable filtering e.g: "--BZ 123456"
```

JacobCallahan (changes requested):

pytest_plugins/issue_handlers.py
mshriver: It could, but I'd like to keep the current pattern instead of providing a default for `--BZ`.  The option is only set if the user provides it.

pytest_plugins/issue_handlers.py
mshriver: `bz_marked_items` is misleading, as its not a list of items with BZ markers. Its a list of the BZs marked on the given item.  Changing to `item_bz_marks`.

pytest_plugins/issue_handlers.py
mshriver: Is `bzys` an acronym for something?  its a test collection filter, as described in the pytest option definition.

pytest_plugins/issue_handlers.py
mshriver: Good call, this can be simplified.

pytest_plugins/rerun_rp/rerun_rp.py
mshriver: Being consistent and explicit with the the pytest API here. So that its not dependent on the default of the option definition.  `config.getoption` will raise ValueError if no default is passed or specified by the option definition.

https://docs.pytest.org/en/stable/_modules/_pytest/config.html#Config.getoption

pytest_plugins/issue_handlers.py
jyejare: Generally Bzys => Bugzillas. But lets stick to what u have defined.

jyejare (approved):
LGTM now! ACK!

No blocker comments pending!

JacobCallahan (approved):
ACK

PR #7952, opened 2020-08-26
Adding skip due to BZ:1850934

akhil-jha:
The tests are failing because of unsuccessful `satellite-installer`. Adding skip until the BZ#1850934 is fixed.

### DIFF
--- a/tests/foreman/api/test_computeresource_azurerm.py
+++ b/tests/foreman/api/test_computeresource_azurerm.py
@@ -242,6 +242,7 @@ class TestAzureRMHostProvisioningTestCase:
 
         return azurermclient.get_vm(name=class_host_ft.name.split('.')[0])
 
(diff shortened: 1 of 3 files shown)


omkarkhatavkar (approved):
ACK

pondrejk (changes requested):
Hi @akhil-jha can you add the BZ number also to docstrings? It is useful to have that reference when  skips are eventually removed.

akhil-jha:
> Hi @akhil-jha can you add the BZ number also to docstrings? It is useful to have that reference when skips are eventually removed.

Done.

JacobCallahan:
@akhil-jha can you update your commit description to describe why these are being skipped?

akhil-jha:
> @akhil-jha can you update your commit description to describe why these are being skipped?

Ah. My bad. I have updated it now.

pondrejk (approved):

Issue #7958, opened 2020-08-27
Revert merge of Filter: Resourch Type changes for roles #7928


swadeley:
Hello @latran @mirekdlugosz 

this change needs to be reverted:
https://github.com/SatelliteQE/robottelo/pull/7928

git commit in master is:

    `robottelo]$ git show 8496a8c0439fe37baf478e94fe08aa66c8bf994e
    commit 8496a8c0439fe37baf478e94fe08aa66c8bf994e
    Author: ltran <ltran@redhat.com>
    Date:   Mon Aug 10 19:37:49 2020 -0400

    Filter: Resourch Type changes for roles
    Resource Type does not have blank selection anymore.  It's changed to '(Miscellaneous)'`

These tests fail with the change in master:
tests/foreman/ui/test_lifecycleenvironment.py::test_positive_custom_user_view_lce
tests/foreman/ui/test_repository.py::test_positive_create_as_non_admin_user_with_cv_published
tests/foreman/ui/test_repository.py::test_positive_create_as_non_admin_user

In all cases you will see "results": []

    `DEBUG    nailgun.client:client.py:121 Received HTTP 200 response: {
                                        "total": 312,
                                        "subtotal": 0,
                                        "page": 1,
                                        "per_page": 350,
                                        "search": "resource_type=\"(Miscellaneous)\"",
                                        "sort": {
                                          "by": null,
                                          "order": null
                                        },
                                        "results": []
                                      }
`

swadeley:
If I revert the change:

    robottelo]$ git revert 8496a8c0439fe37baf478e94fe08aa66c8bf994e
    Auto-merging tests/foreman/ui/test_repository.py
    [master 4feb2bf2] Revert "Filter: Resourch Type changes for roles"
     3 files changed, 4 insertions(+), 4 deletions(-)


then all  three tests pass:

robottelo]$ pytest tests/foreman/ui/test_lifecycleenvironment.py::test_positive_custom_user_view_lce
============================================== test session starts ===============================================
platform linux -- Python 3.7.8, pytest-4.6.3, py-1.9.0, pluggy-0.13.1
shared_function enabled - OFF - scope:  - storage: file
rootdir: /home/sjw/sat6/robottelo_1/robottelo
plugins: xdist-2.0.0, services-1.3.1, mock-1.10.4, forked-1.3.0, cov-2.10.1
collecting ... 2020-08-27 08:22:34 - conftest - DEBUG - Collected 1 test cases
[...]

PR #7928, opened 2020-08-10
Filter: Resourch Type changes for roles

latran:
Resource Type does not have blank selection anymore in the role filters page.  It's changed to '(Miscellaneous)' when creating a role for a user for `access_dashboard`.

`test_positive_custom_user_view_lce`
```
$ pytest tests/foreman/ui/test_lifecycleenvironment.py::test_positive_custom_user_view_lce
============= test session starts ================
platform linux -- Python 3.6.8, pytest-4.6.3, py-1.8.1, pluggy-0.13.1
shared_function enabled - OFF - scope:  - storage: file
rootdir: /home/ltran/Projects/robottelo
plugins: services-1.3.1, mock-1.10.4
collecting ... 2020-08-10 23:30:50 - conftest - DEBUG - Collected 1 test cases
collected 1 item                                                                                                                                                                                                                                                           

tests/foreman/ui/test_lifecycleenvironment.py .                                                                                                                                                                                                                      [100%]

======================== warnings summary =========================
/home/ltran/Projects/p3env/lib/python3.6/site-packages/_pytest/mark/structures.py:337
  /home/ltran/Projects/p3env/lib/python3.6/site-packages/_pytest/mark/structures.py:337: PytestUnknownMarkWarning: Unknown pytest.mark.lifecycleenvironments - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html
    PytestUnknownMarkWarning,

/home/ltran/Projects/p3env/lib/python3.6/site-packages/_pytest/mark/structures.py:337
  /home/ltran/Projects/p3env/lib/python3.6/site-packages/_pytest/mark/structures.py:337: PytestUnknownMarkWarning: Unknown pytest.mark.BZ_1420511 - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html
    PytestUnknownMarkWarning,

/home/ltran/Projects/p3env/lib/python3.6/site-packages/_pytest/mark/structures.py:337
  /home/ltran/Projects/p3env/lib/python3.6/site-packages/_pytest/mark/structures.py:337: PytestUnknownMarkWarning: Unknown pytest.mark.high - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html
    PytestUnknownMarkWarning,

-- Docs: https://docs.pytest.org/en/latest/warnings.html
============== 1 passed, 3 warnings in 105.63 seconds =================================
[...]

PR #7963, opened 2020-08-28
Revert "Filter: Resourch Type changes for roles" (#7958)

swadeley:
Reverts SatelliteQE/robottelo#7928

See https://github.com/SatelliteQE/robottelo/issues/7958


mirekdlugosz (approved):
As promised, ACKing if that helps.

pondrejk (approved):
ack, will let @latran have a say

PR #7967, opened 2020-08-31
some delay in check after org creation

pondrejk:
Newly created organization is not immediately available to hammer info, causing test failure in `UserTestCase::test_positive_last_login_for_new_user`. Adding wait_for to allow for the new org to get registered

```
pytest tests/foreman/cli/test_user.py::UserTestCase::test_positive_last_login_for_new_user
========================================================= test session starts ==========================================================
platform linux -- Python 3.7.8, pytest-4.6.3, py-1.7.0, pluggy-0.12.0
rootdir: /home/pondrejk/Documents/robottelo
plugins: repeat-0.8.0, cov-2.10.0, mock-1.10.4, forked-0.2, services-1.3.1, xdist-1.33.0
collecting ... 2020-08-31 06:56:02 - conftest - DEBUG - Collected 1 test cases
collected 1 item                                                                                                                       

tests/foreman/cli/test_user.py .                                                                                                 [100%]
```

### DIFF
--- a/robottelo/cli/base.py
+++ b/robottelo/cli/base.py
@@ -2,6 +2,8 @@
 import logging
 import re
 
+from wait_for import wait_for
+
 from robottelo import ssh
 from robottelo.cli import hammer
 from robottelo.config import settings
@@ -193,7 +195,18 @@ class Base(object):
                     raise CLIError(tmpl.format(cls.__name__))
                 info_options['organization-id'] = options['organization-id']
 
-            new_obj = cls.info(info_options)
+            # organization creation can take some time
+            if cls.command_base == 'organization':
+                new_obj, _ = wait_for(
+                    lambda: cls.info(info_options),
+                    timeout=300,
+                    delay=5,
+                    silent_failure=True,
+                    handle_exception=True,
+                )
+            else:
+                new_obj = cls.info(info_options)
+
             # stdout should be a dictionary containing the object
             if len(new_obj) > 0:
                 result = new_obj


robottelo/cli/base.py
```diff
@@ -193,7 +195,18 @@ def create(cls, options=None, timeout=None):
                     raise CLIError(tmpl.format(cls.__name__))
                 info_options['organization-id'] = options['organization-id']
 
-            new_obj = cls.info(info_options)
+            # organization creation can take some time
+            if cls.command_base == 'organization':
+                new_obj = wait_for(
```
mirekdlugosz, line 200: `wait_for` returns namedtuple (which is like non-mutable dict). the way it's written now, this function returns different types depending on whether it creates organization or anything else.

I think you need to unwrap return value from `WaitForResult`, and assign that to `new_obj`.

robottelo/cli/base.py
```diff
@@ -193,7 +195,18 @@ def create(cls, options=None, timeout=None):
...
+                    lambda: cls.info(info_options),
+                    timeout=300,
+                    delay=15,
```
mirekdlugosz, line 203: Isn't that a little too extreme? I would expect "some time" to take 2-3 seconds, not 15. Maybe change that to `5`?

Also, isn't nailgun synchronous? And shouldn't Satellite return data only after it is stored and can be used? Kind of looks like workaround for bug in product, might be worth having discussion with devs.

mirekdlugosz (changes requested):
Incoherent return types are blocking for me. This will probably cause large number of tests to fail.

Second comment is more in suggestion space.

robottelo/cli/base.py
JacobCallahan: I'd like to know the resolution of that conversation.

robottelo/cli/base.py
pondrejk: uh, I wish that thing had documentation

robottelo/cli/base.py
pondrejk: Well things aren't that dreary on real-life Satellite, it's just that having hundreds of orgs in place makes things slower. I don't see this as a bug, I mean if the org is not ready, why should it appear in hammer info? Should hammer info wait until org is created? What if the task gets stuck? Filing would imho just feed backlog

mirekdlugosz: > I mean if the org is not ready, why should it appear in hammer info?

I see it rather as: If the org is not ready, why did `hammer create` finish and return its id, and why this id can't be used immediately?

Either way, that was deep in suggestion and speculation territory. I'm OK with not following up with dev.

mirekdlugosz (approved):
ACK

robottelo/cli/base.py
JacobCallahan: Mirek, that is a good point and a good potential bug if things truly aren't ready when create exists.

JacobCallahan (approved):
ACK

PR #7971, opened 2020-09-02
[qe_test-coverage] - added qe coverage for bugzilla 1873439

omkarkhatavkar:
### PR Desription 
Verify the failure of login for the external rhsso user in case same username internal user exists.

### Test Result 
```
plugins: forked-1.1.3, xdist-1.31.0, cov-2.8.1, mock-1.10.4, services-1.3.12020-09-02 12:54:14 - conftest - DEBUG - Collected 1 test cases
collected 1 item

```

### DIFF
--- a/tests/foreman/ui/test_ldap_authentication.py
+++ b/tests/foreman/ui/test_ldap_authentication.py
@@ -18,6 +18,7 @@ import os
 from time import sleep
 
 import pyotp
+import pytest
 from airgun.session import Session
 from fauxfactory import gen_url
 from nailgun import entities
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1152,6 +1153,48 @@ def test_external_new_user_login_and_check_count_rhsso(
...
+    :BZ: 1873439
+
+    :CaseImportance: High
+
+    :steps:
+        1. create an internal user
+        2. create a rhsso user with same username mentioned in internal user
+        3. update the satellite to use rhsso and now try login using external rhsso user
+
+    :expectedresults: external rhsso user should not able to login with same username as internal
```
mirekdlugosz, line 1175: Are we sure this is what will happen in this scenario?

I only skimmed over bz and upstream fix, but I'm under impression that the way it will be fixed and exact details on how it will work are still being discussed.

Of course we can still merge and if fix turns out to be different, then raise another PR with fix.

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1152,6 +1153,48 @@ def test_external_new_user_login_and_check_count_rhsso(
...
+    ).create()
+    external_rhsso_user = create_new_rhsso_user(client_id, username=username)
+    login_details = {
+        'username': external_rhsso_user['username'],
+        'password': settings.rhsso.password,
+    }
+    with Session(login=False) as rhsso_session:
+        with raises(NavigationTriesExceeded) as error:
+            rhsso_session.rhsso_login.login(login_details)
+            rhsso_session.task.read_all()
```
mirekdlugosz, line 1194: I feel this part would look much nicer if written somewhat like:

```python
rhsso_session.rhsso_login.login(login_details)
login_page = rhsso_session.rhsso_login.read()
assert login_page['error_message']
assert "I'm not letting you in!" in login_page['error_message']
assert not login_page['logged_in']
# etc., you get the idea
```

So - try to login, call `read` on login page and run some explicit assertions, instead of asserting by-product of failed login, which is exception raised by navmazing.

Since that would require change in Airgun first, I don't consider it blocking for PR itself. But something to think about.

mirekdlugosz (approved):
ACK with question and suggestion

tests/foreman/ui/test_ldap_authentication.py
JacobCallahan: @mirekdlugosz what level of effort would be needed to make that change in airgun?

omkarkhatavkar: @mirekdlugosz I thought about that and will required to wait till upstream PR get merged, due to importance of this BZ I just wanted to add it So I dont miss this and anyways this will be skip till and when the BZ get resolved we could see this test is up and running. 

Anyone can ask me why to do this double work ? As I already have all the sauces/required helper functions, this writing test did not took not more than 20 mins hence I did it. 

@mirekdlugosz As this is unblocking I will add airgun issue linkinf with this PR so we will not miss your suggestion.

tests/foreman/ui/test_ldap_authentication.py
omkarkhatavkar: Yeah, you are correct based on the upstream PR, We are ahead of with our testing. We can raise fix when we see this test fails and tracked too.

omkarkhatavkar: As per my understanding the User should not able to login with message "User is already exist with same Username" in other LDAP Auth Sources.

tests/foreman/ui/test_ldap_authentication.py
mirekdlugosz: @JacobCallahan it's impossible to give accurate estimate without looking at pages themselves, but I imagine it should break down to adding couple of widgets to view and new `read()` method to entity. And thinking whether it makes more sense to call `read()` explicitly, or at the end of `login()` method, or both.

The bigger problem is, Omkar is only other active reviewer in Airgun, so if he raises PR there, we will probably wait weeks to have it merged. That's why I don't consider this suggestion blocking.

@omkarkhatavkar adding ticket to Airgun so we can get back to this later works for me. Thanks.

omkarkhatavkar: raised here https://github.com/SatelliteQE/airgun/issues/513

JacobCallahan (approved):
ACK

PR #7975, opened 2020-09-05
Test: onthefly functionality

akhil-jha:
```
pytest tests/foreman/ui/test_ldap_authentication.py -k test_onthefly_functionality
platform linux -- Python 3.6.8, pytest-4.6.3, py-1.9.0, pluggy-0.13.1
plugins: services-1.3.1, mock-1.10.4, cov-2.10.1, forked-1.3.0, xdist-2.1.0
collecting ... 2020-09-05 13:09:37 - conftest - DEBUG - Collected 30 test cases
collected 30 items / 29 deselected / 1 selected                                                                                                                             

tests/foreman/ui/test_ldap_authentication.py .                                                                                                                        [100%]

================================================================= 1 passed, 29 deselected in 79.99 seconds ==================================================================
```

### DIFF
--- a/tests/foreman/ui/test_ldap_authentication.py
+++ b/tests/foreman/ui/test_ldap_authentication.py
@@ -1324,3 +1324,40 @@ def test_deleted_idm_user_should_not_be_able_to_login(auth_source_ipa, ldap_tear
         with raises(NavigationTriesExceeded) as error:
             ldapsession.user.search('')
         assert error.typename == "NavigationTriesExceeded"
+
+
+@tier2
+def test_onthefly_functionality(session, ipa_data, ldap_tear_down):
+    """IDM user will not be created automatically in Satellite if onthefly is
+    disabled
+
+    :id: 6998de30-ef77-11ea-a0ce-0c7a158cbff4
+
+    :Steps:
+        1. Create an auth source (IDM) with onthefly disabled
+        2. Try login with a user from auth source
+
+    :expectedresults: Login fails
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1324,3 +1324,40 @@ def test_deleted_idm_user_should_not_be_able_to_login(auth_source_ipa, ldap_tear
         with raises(NavigationTriesExceeded) as error:
             ldapsession.user.search('')
         assert error.typename == "NavigationTriesExceeded"
+
+
+@tier2
+def test_onthefly_functionality(session, ipa_data, ldap_tear_down):
+    """IDM user will not be created automatically in Satellite if onthefly is
+    disabled
```
mirekdlugosz, line 1332: In test, you check that user can't log in. Would it make sense to explicitly check if user is not created?

mirekdlugosz (approved):
Code looks OK, left one question.
ACK

tests/foreman/ui/test_ldap_authentication.py
akhil-jha: The user will only be created in satellite after the user has been authenticated from LDAP(here IDM). So being said that, the user creation depends on the successful login.
Even if we `'account.onthefly_register': True`, until the user logs in, the account will not be created.

omkarkhatavkar (approved):
ACK

PR #7989, opened 2020-09-14
added CAC card manual tests related to rhsso

omkarkhatavkar:
### PR description 
These are the manual tests needed to run with CAC card.


mirekdlugosz (approved):
Consider adding "CaseAutomation" metadata field.

`CaseAutomation: NotAutomated` is for tests you plan to automate in the future, `CaseAutomation: ManualOnly` is for tests that can't or shouldn't be automated.
I think these test probably will be `ManualOnly`, since you won't be able to insert and eject CAC card automatically.

omkarkhatavkar:
> CaseA

Thanks for that, I was using wrong keyword

jameerpathan111 (approved):
ACK

PR #7640, opened 2020-02-12
Add test syspurpose end to end

swadeley:
A new test to create a host using an activation key with system purpose values set, then change them

PR #7996, opened 2020-09-15
test syspurpose fix

tstrych:
The activation key has no subscriptions added. This is causing error as there is unknown value. 

So it's wrong to check it inside the test as there is just one random one added by auto-attach, and this one is changing as manifest is changing.

I want to ask @swadeley  the creator of this test if I understand this correctly. And I want to know what was the purpose of checking  `host_subscriptions` inside syspurpose_test at first.

platform linux -- Python 3.7.9, pytest-4.6.3, py-1.8.1, pluggy-0.13.1 --automation_v3/bin/python
cachedir: .pytest_cache
rootdir: /home/tstrych/Documents/work/automation
plugins: forked-1.1.3, services-1.3.1, cov-2.10.1, mock-1.10.4, xdist-2.1.0
collecting ... 2020-09-15 13:27:04 - conftest - DEBUG - Collected 64 test cases
collected 64 items / 63 deselected / 1 selected

test_host.py::HostSubscriptionTestCase::test_syspurpose_end_to_end 2020-09-15 15:34:54 - robottelo - DEBUG - Finished Test: HostSubscriptionTestCase/test_syspurpose_end_to_end
2020-09-15 15:34:54 - robottelo - INFO - Started tearDownClass: tests.foreman.cli.test_host/HostSubscriptionTestCase

warnings ommitted
============ 1 passed, 63 deselected, 2 warnings in 470.08 seconds =============

### DIFF
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -2452,6 +2452,13 @@ class HostSubscriptionTestCase(CLITestCase):
                 'content-view-id': self.content_view['id'],
             }
         )
+        ActivationKey.add_subscription(
+            {
+                'organization-id': self.org['id'],
+                'id': activation_key['id'],
+                'subscription-id': self.default_subscription_id,
+            }
+        )
         # Register a host using the activation key
         self._register_client(activation_key=activation_key, enable_repo=True, auto_attach=True)
         self.assertTrue(self.client.subscribed)
@@ -2494,13 +2501,12 @@ class HostSubscriptionTestCase(CLITestCase):
         self.assertEqual(
             host['subscription-information']['system-purpose']['service-level'], "Self-Support2"
         )
-        # Assert subscriptions present
         host_subscriptions = ActivationKey.subscriptions(
             {'organization-id': self.org['id'], 'id': activation_key['id'], 'host-id': host['id']},
             output_format='json',
         )
         self.assertGreater(len(host_subscriptions), 0)
-        self.assertEqual(self.subscription_name, host_subscriptions[0]['name'])
+        self.assertEqual(host_subscriptions[0]['name'], self.subscription_name)
         # Unregister host
         Host.subscription_unregister({'host': self.client.hostname})
         with self.assertRaises(CLIReturnCodeError):


swadeley:
Hello @tstrych 

From the bug [Bug 1698158 - [RFE] Activation Keys should support setting system purpose](https://bugzilla.redhat.com/show_bug.cgi?id=1698158) I can see that the host must be registered, but no mention of subscribed.

Give me some time to dig through notes I made at the time.

Thank you

swadeley:
Hello @tstrych 

from my IRC logs: 
me: I have written another test which firsts creates empty AK then tries to update syspurpose values. It fails Service level "Self-Support" is not available to units of organization <blah>
dev: correct as in you need to import a subscription thru a manifest
me :adding custom product did not help, adding repo to custom product did not help.
me: adding a manifest did help

So its the values you see in the web UI that are related to the subscription which in turn comes from manifest. Today I tested by created a "test_org" with no manifest, registered a VM, and I can still see three options under **Service Level (SLA)**.  Anyway, that is why there is a subscription added in the test. As to the assert, this was based on feedback:
https://github.com/SatelliteQE/robottelo/pull/7640#discussion_r379021999

You can see there, at that link, I originally had:
`self.assertGreater(len(host_subscriptions), 0)`

Is that all understandable?

tstrych:
Yes makes sense, thanks. 

But in setupclass or test itself, there no subscription added to activation key.
So IMO, checking if there expected subscription is wrong. It seems there is just one 




result from host which is registered:
```
subscription-manager status
+-------------------------------------------+
   System Status Details
+-------------------------------------------+
Overall Status: Insufficient

Employee SKU:
- Guest has not been reported on any host and is using a temporary unmapped guest subscription. For more
  information, please see https://access.redhat.com/solutions/1592573

System Purpose Status: Mismatched
```

 And this subscription is now `Employee SKU` but two weeks before it was `Red Hat Satellite Infrastructure Subscription` and even before[ jhutar changed](https://github.com/SatelliteQE/robottelo/commit/a4353a1a22b27affc20d68d13032610be881215d) it from `Red Hat Satellite Employee Subscription`. 
So definitely checking constant leads us to fixing now and then, so it's better to not use it in this case. 

IMO this test is not focused on subscriptions. We can test a manifest on satellite if that was important. Manifest is the important thing the stephen mentioned if I underestand correctly to the problem. 

Probably the assert with `self.assertGreater(len(host_subscriptions), 0)` can stay but I think it's more clean without it. 

Please @swadeley  let me know what you think, and if I do not understand a problem please correct me :)

JacobCallahan:
Assuming there is no custom product added to the key, then the assertion should be: `assert host_subscriptions` since we are moving away from unittest.

swadeley:
Hello @tstrych 

I agree to a certain extent the subscription name does not matter, but I think having a RHEL subscription in the organization is required (and that is what `setup_org_for_a_rh_repo` in the class `HostSubscriptionTestCase` does). That setup function comes from [robottelo/cli/factory.py](https://github.com/SatelliteQE/robottelo/blob/12481caa060497f2f8a21a2c8c12012e8d5325ec/robottelo/cli/factory.py#L1922), and there is a lot of error handling in there. 

To get this PR approve more quickly, I will approve your changes now and leave it up to the tier 2 reviewer to decide if the test must assert a subscription exists.

BTW, we should not use Employee Subscription for testing.

Thank you

swadeley (approved):
ACK

tstrych:
@JacobCallahan [unittest2pytest](https://github.com/pytest-dev/unittest2pytest) will do that anyway, when I will convert whole `cli/test_host`.  (next sprint ) 

Thank you @swadeley for all info. 
`setup_org_for_a_rh_repo` indeed enables red hat satellite tools ` 'repository-set': REPOSET['rhst7'], 'repository': REPOS['rhst7']['name'],`
But it enables them only on the satellite not on the host. There is no subscription on host, as there is neither product or subscriptions added to activation key. 

I think you are right we should check somehow `subscription` if we want to place this test inside `HostSubscriptionTestCase`. 
So I added subscription. This should solve our problem. Now we do know which subscription to test.

tstrych:
Test results.
```
collected 64 items / 63 deselected / 1 selected

test_host.py::HostSubscriptionTestCase::test_syspurpose_end_to_end 2020-09-18 17:23:01 - robottelo - DEBUG - Finished Test: HostSubscriptionTestCase/test_syspurpose_end_to_end

============ 1 passed, 63 deselected, 2 warnings in 347.57 seconds =============
```

pondrejk (approved):

PR #7997, opened 2020-09-15
Allow each xdist worker to have its own Satellite instance

JacobCallahan:
This is phase 1 of this process, meaning Satellites must be available
and added to the config before tests are ran.
Any overflowing workers will be directed at the main Satellite hostname:
`settings.server.hostname`

### DIFF
--- a/pytest_fixtures/xdist.py
+++ b/pytest_fixtures/xdist.py
@@ -1,0 +1,13 @@
+"""Fixtures specific to or relating to pytest's xdist plugin"""
+import pytest
+
+from robottelo.config import settings
+
+
+@pytest.fixture(scope="session", autouse=True)
+def align_xdist_satellites(worker_id):
+    """Set a different Satellite per worker when available in robottelo's config"""
+    settings.configure()
+    settings.server.hostname = settings.server.get_hostname(worker_id)
+    settings._configure_entities()
+    settings._configure_airgun()
--- a/robottelo.properties.sample
+++ b/robottelo.properties.sample
@@ -2,6 +2,10 @@
 [server]
 # Server hostname
 hostname=
+# Additional satellite hostnames for xdist workers
+# gw0=
+# gw1=
+# gw2=
 
 # Path to private ssh key to be used when connecting via SSH.
 ssh_key=
--- a/robottelo/config/base.py
+++ b/robottelo/config/base.py
@@ -178,6 +178,10 @@ class ServerSettings(FeatureSettings):
         """
         return (self.admin_username, self.admin_password)
 
+    def get_hostname(self, key="hostname"):
+        reader = INIReader(os.path.join(get_project_root(), SETTINGS_FILE_NAME))
+        return reader.get('server', key, self.hostname)
+
     def get_url(self):
         """Return the base URL of the Foreman deployment being tested.
 
--- a/conftest.py
+++ b/conftest.py
@@ -8,6 +8,7 @@ pytest_plugins = [
     "pytest_plugins.manual_skipped",
     # Fixtures
     "pytest_fixtures.api_fixtures",
+    "pytest_fixtures.xdist",
     # Component Fixtures
     "pytest_fixtures.satellite_auth",
     "pytest_fixtures.templatesync_fixtures",
--- a/requirements-optional.txt
+++ b/requirements-optional.txt
@@ -2,7 +2,6 @@
 codecov
 flake8
 pytest-cov
-pytest-xdist
 redis
 tox
 pre-commit
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,6 +14,7 @@ productmd==1.26
 pytest==4.6.3
 pytest-services==1.3.1
 pytest-mock==1.10.4
+pytest-xdist==1.34.0
 selenium==3.141.0
 requests==2.22.0
 testimony==2.1.0


JacobCallahan:
**Test results**
```
--- Before Change ---

# pytest tests/foreman/api/test_activationkey.py 
==================================================================== test session starts =====================================================================
platform linux -- Python 3.8.5, pytest-4.6.3, py-1.9.0, pluggy-0.13.1
rootdir: /home/jake/Programming/robottelo
plugins: cov-2.10.0, services-1.3.1, mock-1.10.4, forked-1.3.0, xdist-1.34.0
collecting ... 2020-09-14 19:53:04 - conftest - DEBUG - Collected 22 test cases
collected 22 items                                                                                                                                           

tests/foreman/api/test_activationkey.py ......s...............                                                                                         [100%]

====================================================================== warnings summary ======================================================================
---
===================================================== 21 passed, 1 skipped, 7 warnings in 448.62 seconds =====================================================


--- After Change ---

# pytest tests/foreman/api/test_activationkey.py 
==================================================================== test session starts =====================================================================
platform linux -- Python 3.8.5, pytest-4.6.3, py-1.9.0, pluggy-0.13.1
rootdir: /home/jake/Programming/robottelo
plugins: cov-2.10.0, services-1.3.1, mock-1.10.4, forked-1.3.0, xdist-1.34.0
collecting ... 2020-09-14 20:45:32 - conftest - DEBUG - Collected 22 test cases
collected 22 items                                                                                                                                           

tests/foreman/api/test_activationkey.py ......s...............                                                                                         [100%]

====================================================================== warnings summary ======================================================================
---
===================================================== 21 passed, 1 skipped, 7 warnings in 495.71 seconds =====================================================

# pytest -n 4 tests/foreman/api/test_activationkey.py 
==================================================================== test session starts =====================================================================
platform linux -- Python 3.8.5, pytest-4.6.3, py-1.9.0, pluggy-0.13.1
rootdir: /home/jake/Programming/robottelo
plugins: cov-2.10.0, services-1.3.1, mock-1.10.4, forked-1.3.0, xdist-1.34.0
gw0 [22] / gw1 [22] / gw2 [22] / gw3 [22]
.s....................                                                                                                                                 [100%]
====================================================================== warnings summary ======================================================================
---
==================================================== 21 passed, 1 skipped, 28 warnings in 225.23 seconds =====================================================
```

JacobCallahan:
Moved pytest-xdist from requirements-optional to requirements as this is now no longer optional.

tstrych:
The code result looks nice :) 
What will happen if you will run the test with more workers than you have ?
Is this already working with ui tests ?

JacobCallahan:
@tstrych the final run i did was with 3 available satellites, but 4 xdist workers. Overflowing workers are given the main satellite. This does call the airgun config during it, so should have the same behavior. If you have some stable airgun tests you'd like me to run, I can.

tstrych:
Thanks @JacobCallahan 
I just wanted to know if it works, your word is enough ;) 

Just checking if I understand correctly.
So in theory 
3 satellites and 4'th worker is the satellite itself, 
So If I will run it with 6 workers the 4'th 5'th and 6'th will run on main satellite ? So in ideal world there will be speed-up x times, where x is number of workers different from main satellite ? (in our little example 3-times faster)

JacobCallahan:
@tstrych yeah so any worker that doesn't have a satellite assigned in robottelo.properties will align to `settings.server.hostname`. A future version could attempt to do balancing, where it may choose from another xdist's satellite. And even further in the future (once broker support is merged and other lab pieces are in place), this fixture can be updated to spin up a Satellite for any xdist worker that doesn't have one set in the config!

tstrych:
Thanks for explanation :)

pondrejk:
So is this a replacement of multi-threaded execution or is it possible to run several threads on these workers? I assume tests result collection will be adjusted to the new situation.

pytest_fixtures/xdist.py
```diff
@@ -0,0 +1,12 @@
+"""Fixtures specific to or relating to pytest's xdist plugin"""
+import pytest
+
+from robottelo.config import settings
+
+
+@pytest.fixture(scope="session", autouse=True)
+def align_xdist_satellites(worker_id):
+    """Set a different Satellite per worker when available in robottelo's config"""
+    settings.server.hostname = settings.server.get_hostname(worker_id)
```
jyejare, line 10: You need to check and set the configuration first before accessing the `settings` from `robottelo.properties` by :

```
if not settings.configured:
     settings.configure()
```
You are able to access those settings because other plugins above `pytest_fixtures.xdist` in `conftest.py` are setting it for you. But tomorrow they might swap the place or probably configurable whether to call or not. Hence its good to set it locally.

You can also choose to set it at the module level instead in the fixture function.

The plan is to fix it in future by setting it once for all the required modules.

robottelo.properties.sample
```diff
@@ -2,6 +2,10 @@
 [server]
 # Server hostname
 hostname=
+# Additional satellite hostnames for xdist workers
+# gw0=
+# gw1=
+# gw2=
```
jyejare, line 8: Is it capped to only 3 additional workers , or it can be any numbers of workers ?

If yes, Can we atleast define upto `gw7` here ? OR inform the availability of any number of workers in the comment.

pytest_fixtures/xdist.py
```diff
@@ -0,0 +1,12 @@
...
+    settings._configure_entities()
```
jyejare, line 11: The `settings.configure()` above will automatically take care of `_configure_entities()` and `_configure_airgun()`. So we are good to remove it from here.

jyejare (commented):
ACK pending comments.

robottelo.properties.sample
mshriver: Its unlimited in the behavior, you could have 100 xdist workers if you've got 100 instances to list here. But the machine running pytest better have a _lot_ of cores and memory :D

mshriver (approved):
Approved, @jyejare has a good point about needing to pre-configure settings before using them.

JacobCallahan:
@jyejare @mshriver I've made a minor update to the PR. This added the settings.configure() call as per Jitendra's suggestion. Unfortunately, configuring airgun/nailgun still needs to be done manually as loading the rest of the config overrides the change in this fixture.

pytest_fixtures/xdist.py
```diff
@@ -0,0 +1,13 @@
+"""Fixtures specific to or relating to pytest's xdist plugin"""
+import pytest
+
+from robottelo.config import settings
+
+
+@pytest.fixture(scope="session", autouse=True)
+def align_xdist_satellites(worker_id):
+    """Set a different Satellite per worker when available in robottelo's config"""
+    settings.configure()
```
jyejare, line 10: Please check if its already configured, if these is an inbuilt check fr it then i m all good fr it.

JacobCallahan: settings.configure has an internal check for this

robottelo/config/base.py
```diff
@@ -178,6 +178,10 @@ def get_credentials(self):
         """
         return (self.admin_username, self.admin_password)
 
+    def get_hostname(self, key="hostname"):
+        reader = INIReader(os.path.join(get_project_root(), SETTINGS_FILE_NAME))
+        return reader.get('server', key, self.hostname)
```
lpramuk, line 183: why using INIReader again (re-reading ini) if you have `settings.configure()`ed already? 
```
def get_hostname(self, key="hostname"):
    return settings.server.__dict__[key] or settings.server.hostname
```

lpramuk (changes requested):
why do you grab for low level INIReader when you have all settings read already

robottelo/config/base.py
JacobCallahan: that's an option, but I try to avoid using dunder attributes within robottelo

robottelo.properties.sample
```diff
@@ -2,6 +2,10 @@
 [server]
 # Server hostname
 hostname=
+# Additional satellite hostnames for xdist workers
+# gw0=
```
mirekdlugosz, line 6: Wouldn't it be cleaner to make `hostname` list of available hostnames?

Then helper function can extract number from worker name, get modulo of that on list length and use that as index:

```python
>>> worker_id = "gw5"
>>> hosts = [f'dhcp{num}.example.com' for num in range(4)]
>>> hosts
['dhcp0.example.com', 'dhcp1.example.com', 'dhcp2.example.com', 'dhcp3.example.com']
>>> worker_num = int(worker_id[2:])  # or something more clever, if workers aren't always named `gw<int>`
>>> hosts[ worker_num % len(hosts) ]
'dhcp2.example.com'
```

This helper should also account for `settings.server.hostname` being either string or a list.

robottelo/config/base.py
mirekdlugosz: You could use `getattr`, then:
```python
def get_hostname(self, key="hostname"):
    return getattr(self.server, key, settings.server.hostname)
```

As discussed in #8003 , settings module is not ideal place for helpers like that. I would like to avoid introducing new one only to have it yanked in few months.

robottelo.properties.sample
JacobCallahan: that was similar to my initial thought, but this is more clear for ini-based config files.

jyejare:
@JacobCallahan Thanks for configuring before using settings.

Considering the *lesser time* it takes to configure the airgun and nailgun twice (before and after get hostname), Acking this PR with pending comments from other reviewers.

jyejare (approved):

JacobCallahan:
@lpramuk @mirekdlugosz are both of you good with this?

robottelo.properties.sample
mirekdlugosz: Wouldn't say one is clearer than another:

```
[server]
hostname=dhcp0.example.com,dhcp1.example.com,dhcp2.example.com
```

```
[server]
hostname=dhcp0.example.com
gw0=dhcp1.example.com
gw1=dhcp2.example.com
```

But I would say having single setting that is list of values is preferable to leaking internal pytest-xdist identifiers to settings module.

This is ultimately up to you, but we could have solution that is clean and nicely integrated with dynaconf from day one.

mirekdlugosz:
@JacobCallahan neither of my comments is resolved, but if it's critical to have it merged, this can be certainly done without my endorsement. I haven't NACKed, after all.

JacobCallahan:
@mirekdlugosz I am going to try your getattr suggestion later today, but won't be re-formatting the ini at this time. that is something i'd like to do once we have dynaconf integration in, as yaml is much better suited for nested k-v entries.

robottelo/config/base.py
JacobCallahan: @mirekdlugosz So this won't work, since the worker id key-value pairs aren't read into the server class. Adding the helper method is the cleanest way to handle this until we can revisit after dynaconf integration is complete.

lpramuk: time being hack, OK   yet how many times yet :-)

lpramuk (approved):
Ouch!

PR #8003, opened 2020-09-21
Dynaconf integration - phase 1

mirekdlugosz:
This PR continues and includes work done in https://github.com/SatelliteQE/robottelo/pull/7774 . Main goal is to use dynaconf as Robottelo configuration library.

This PR introduces `ConfigProxy` facade that shall serve as single point for all configuration values retrieval across Robottelo. Main features are:

1. Contains references to both dynaconf settings store and legacy robottelo settings store. Values are looked up in dynaconf first.  This allows to defer hard cut-off date of migration.
2. Caching of configuration values, including properties and methods. Properties are calculated only once, methods are pre-computed and thin wrappers are returned in their place (calling method again will return pre-calculated value)
3. Singleton. Multiple `ConfigProxy()` instances share same underlying cache and references to settings store.
4. Settings are automatically configured on first import. No need for `if settings.configured` conditionals and explicit `settings.configure()` calls.
5. Fully transparent for Python code. Callers should not notice that new objects are returned in place of direct configuration options
    * When debugging using f-strings, do note that `f'{var=}'` goes for string representation (`var.__repr__`) directly. Cast to string if you want to see setting value: `f'{var=!s}'`
    * some of external libraries, like ssh module, are picky about type of variable they receive on input. Facade maintains list of exceptions where it should return configuration value instead of proxy, but that list might be incomplete.

## Why draft?

* While unit tests all pass locally, I have not exercised facade in actual tests. This needs to be done before we merge this PR.
* I am looking for some discussion around Dynaconf invocation (single settings.yaml vs multiple yaml files in conf/), and general design direction.
* It makes sense for me to rebase and clean up commit history before final merge.
* It might be good idea to include documentation and unit tests for facade.

## Remaining work

There are some problems with current approach that should be solved before final switch to Dynaconf and depreciation of internal settings module:

1. `Settings` class is responsible for setting up logging, configuring Nailgun and Airgun. Logging should be decoupled into standalone module, *guns configuration should be moved to facade or `robottelo.settings` module
2. `Settings` class contains helpers methods, introduced in https://github.com/SatelliteQE/robottelo/pull/2869 . While I somewhat understand the reason, I strongly believe that helper functions have no place in configuration module, even if helpers are thin wrappers around specific configuration options. Dynaconf shares that belief. These helpers should be moved back to helpers.py, and callers need to be updated.
3. Tests skipping logic depends on settings module internals, such as `settings.all_features` (list of all "features") and `settings.<section>.validate()` (method validating specific section). Dynaconf has neither, so it either needs to be monkey-patched, or skipping logic should be refactored. Dynaconf validates all settings when `settings.validators.validate()` is called, ideally during settings module initialization - way before any `skip` method is called.
4. `robottelo.yaml` file and `settings.server.version` aren't really necessary with dynaconf around, and should be refactored.
5. settings are imported in context of Satellite CI (where all configuration options are provided), and in context of Robottelo CI (where none of configuration options are provided). This difference might have been reason behind introducing `settings.configure()` method, but it proved awkward to work with (unnecessary call in python shell, `robottelo.config.base.SSHClientSettings` `command_timeout` and `connection_timeout` properties, VM class unit tests). Some other solutions should be considered.

## Migration plan

For me, the most important thing is for migration to be transparent to users. Where it is not possible, cut-off dates should be decided irrespective of any specific PR, with possibility of being announced months in advance.

1. Introduce dynaconf as optional dependency. robottelo.properties is still required, but dynaconf settings take precedence.
2. Refactor robottelo internals to allow complete switch to dynaconf (see "Remaining work" above).
3. Port existing robottelo.properties to dynaconf yaml files for CI.
4. Drop usage of legacy settings module. Before this point, everyone in the team needs to use new dynaconf yaml file(s) instead of legacy robottelo.properties.


settings.sample.yaml
```diff
@@ -0,0 +1,16 @@
+# copy this file as `settings.local.yaml` and override any config provided under `conf/*.yaml`
+# or preferably use environment variables instead of this file
+# example:
+# `export SATQE_SERVER__HOSTNAME=myserver.redhat.com`
+---
+server:
```
mshriver, line 6: @JacobCallahan should keep an eye on this for #7997

robottelo/config/__init__.py
```diff
@@ -1,4 +1,48 @@
...
+from robottelo.config.base import Settings as LegacySettings
+from robottelo.config.proxy import ConfigNodeWrapper
+from robottelo.config.proxy import ConfigProxy
+
+logger = logging.getLogger('robottelo.config')
+
+legacy_settings = LegacySettings()
+
+dynaconf_settings = LazySettings(
+    envvar_prefix="SATQE",
```
lpramuk, line 17: SATQE? when this is ROBOTTELO

mirekdlugosz: I guess Bruno's intent was that Nailgun and Airgun will switch to Dynaconf as well one day, and they would all take config from single namespace.

I'll let others chime in before making change. Personally, I'm fine either way.

.env
```diff
@@ -0,0 +1 @@
+# This file is here for dynaconf to set this as root level directory
```
lpramuk, line 1: bad wording
```
You can put in the root of your application a file called .env to automatically load some env variables by 

settings = Dynaconf(load_dotenv=True)
```

mirekdlugosz: I don't understand. Your proposed wording has completely different meaning than the current one.

Clearly, the intent behind this file is to tell dynaconf where root of the project is. So relative paths used in `LazySettings()` instantiation are resolved relative to directory with this file. It's not there as yet another source of settings.

robottelo/config/__init__.py
JacobCallahan: I'd prefer ROBOTTELO as we would likely have repo-specific settings.

robottelo/config/base.py
```diff
@@ -1620,6 +1620,21 @@ def _configure_third_party_logging(self):
         for logger in loggers:
             logging.getLogger(logger).setLevel(logging.WARNING)
 
+    def to_dict(self):
```
JacobCallahan, line 1623: we do have some nested values that might trip this up since it isn't recursive.

robottelo/config/proxy.py
```diff
@@ -0,0 +1,238 @@
...
+        except KeyError:
+            pass
+
+        try:
+            return self._dispatch_computed_value(full_path)
+        except KeyError:
+            pass
+
+        value = self._get_from_configs(full_path)
+        return value
```
JacobCallahan, line 228: untested, but looks better imo
```suggestion
        return (
            self._from_cache.get (full_path) or 
            self._dispatch_computed_value.get(full_path) or 
            self._get_from_configs(full_path)
        )
```

JacobCallahan (changes requested):

robottelo/config/proxy.py
```diff
@@ -0,0 +1,238 @@
...
+            version = get_sat_version()
+
+        return version
+
+    def __capsule_hostname(self):
+        try:
+            instance_name = self.get('capsule.instance_name')
+            domain = self.get('capsule.domain')
+            return f"{instance_name}.{domain}"
+        except KeyError():
```
jyejare, line 154:
```suggestion
        except KeyError:
```

robottelo/config/validators.py
```diff
@@ -0,0 +1,251 @@
+from dynaconf import Validator
+
+from robottelo.constants import AZURERM_VALID_REGIONS
+from robottelo.constants import VALID_GCE_ZONES
+
+
+validators = dict(
+    server=[
+        Validator("server.hostname", must_exist=True),
+        Validator("SERVER.ssh_key", must_exist=True)
```
jyejare, line 10: s/SERVER/server/ ? 

and below as well ?

jyejare (commented):
LGTM. Have left some comments and questions.

I expect the Author to create a separate commit for new additions in future in this PR.

robottelo/config/__init__.py
lpramuk: I'm going to object to myself :-)

if we integrate dynaconf to our projects I can imagine situation where we are using different prefixes for differrent projects, but then we want one env var to represent the same entity in multiple projects we need to set multiple env vars....

Isn't better to just use one prefix everywhere, so now for all the projects better suites `SATQE` (than `ROBOTTELO`)

lpramuk: To comply with https://github.com/SatelliteQE/robottelo/pull/7451 you have to use `ROBOTTELO_` prefix

This is already supported https://github.com/SatelliteQE/robottelo/blob/97c6845dd3bd86c8e034320736f7030125be6f0a/robottelo/config/base.py#L80
```
 value = os.environ.get(f'ROBOTTELO_{section.upper()}_{option.upper()}')
```

mirekdlugosz: I switched to ROBOTTELO, since that's people here seem to prefer.

@lpramuk #7451 is re-implementation of one of dynaconf functionalities, we only have it because we didn't had dynaconf. Once dynaconf is integrated, that PR should be reverted.
Also, does anybody use that feature? Jenkins sets these variables anywhere?

robottelo/config/base.py
mirekdlugosz: Bruno's idea was to introduce properties-to-yaml translation tool, and that method was needed for that tool. Since I don't intent to implement such tool (I introduce facade so old new settings can co-exist and we can port them gradually), I simply dropped this method.

robottelo/config/proxy.py
mirekdlugosz: Don't forget that falsy values are completely normal and valid - `self._from_cache(full_path)` can return False, None, empty string, 0 (zero) or empty list. How would you differentiate between None-as-valid-value and None-as-could-not-find?

I could use sentinel value and direct comparison between it and whatever function returned, but at that point I just use `if` instead of `try/except`. Some people swear that `try/except` is not flow control, but in Python we believe in EAFP.

.gitignore
```diff
@@ -59,3 +59,14 @@ tests/foreman/pytest.ini
 
 # For `coverage xml` generated file.
 **/coverage.xml
+
+# Dynaconf local files
```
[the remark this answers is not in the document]
mirekdlugosz: Dynaconf doesn't generate any data. These are local as in they are read by Dynaconf, but should never be committed.

Usually, you would have some default settings that are safe to store in public git, like name of cloud provider or PyPI address; and then you would have additional file with username and password or API key, which obviously should never be made public.

In Robottelo, you would be able to just copy YAML files from jenkins-configs and then overwrite selected values in local file, like server FQDN and path to SSH key.

robottelo/config/__init__.py
```diff
@@ -1,4 +1,48 @@
...
+    )
+
+try:
+    dynaconf_settings.validators.validate()
+except ValidationError:
+    logger.error(
+        "Dynaconf validation failed, continuing for the sake of unit tests", exc_info=True
+    )
+
+settings_proxy = ConfigProxy()
```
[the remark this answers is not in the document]
mirekdlugosz: 1. As Robottelo user and developer, you are never expected to interact with that object and class name directly. The whole point is that it's transparent for users within application.
2. ["Proxy" is established name of the pattern](https://en.wikipedia.org/wiki/Proxy_pattern). "Proxy" is used as synonym for "substitute" in many different contexts, also outside of programming. I can understand how it can be confusing, but that's just effect of term being overloaded.

What name would you propose? Would `SettingsFacade` be better?

robottelo/config/validators.py
mirekdlugosz: Guess that was added before Bruno introduced case-insensitive keys into dynaconf. I believe now dynaconf understands both, so it doesn't matter for it.
Changed, thanks.

robottelo/config/__init__.py
lpramuk: Ah. OK. Jenkins didn't yet adopt it

mirekdlugosz:
I have included suggestions from comments and rebased PR. I re-run tier1 tests against updated branch. Full results are in $JENKINS/job/satellite6-standalone-automation/2054/

Compared to snap 19 tier1 results, I observed 52 new failures. Out of these, 26 passed when run locally and 15 failed because SatLab machines don't have `fuser` binary available (and all tests using fake proxy depend on it). Then there were 9 that also failed locally with message "name has already been taken". There might be something wrong with `valid_data_list()` helper, but I haven't investigated further.

That leaves us with only 2 new failures. I couldn't associate them with any change in settings module - they seem to be caused by environment misconfiguration, or are intermittent.

@lpramuk @jyejare @mshriver @JacobCallahan 
please see new commits and my replies to your comments. Please resolve conversations you feel are done now.

jyejare (approved):
LGTM

robottelo/config/__init__.py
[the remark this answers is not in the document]
mirekdlugosz: As you wish, renamed in 87d37aee9a7457d632bba18802f55207ba0a11cc :
* `ConfigProxy` -> `SettingsFacade`
* `ConfigNodeWrapper` -> `SettingsNodeWrapper`

robottelo/config/facade.py
```diff
@@ -0,0 +1,311 @@
...
+    'server.scheme',
+    'azurerm.azure_region',
+    'azurerm.client_id',
+    'azurerm.client_secret',
+    'azurerm.password',
+    'azurerm.ssh_pub_key',
+    'azurerm.subscription_id',
+    'azurerm.tenant_id',
+    'azurerm.username',
+    'compute_resources.libvirt_hostname',
```
JacobCallahan, line 29: with the addition of a new broker setting, you'll like need to add that in here.

robottelo/config/validators.py
```diff
@@ -0,0 +1,219 @@
...
+            "azurerm.azure_region",
+            "azurerm.ssh_pub_key",
+            "azurerm.username",
+            "azurerm.password",
+            "azurerm.azure_subnet",
+            must_exist=True,
+        ),
+        Validator("azurerm.azure_region", is_in=AZURERM_VALID_REGIONS),
+    ],
+    capsule=[Validator("capsule.instance_name", must_exist=True)],
```
JacobCallahan, line 28: you'll likely need to add a validator for broker as well

settings.sample.yaml
JacobCallahan: we can make this change later. current settings would likely work from robottelo.properties

JacobCallahan (changes requested):
Hey Mirek, since this PR was opened, broker has been integrated. With that, a new setting has been added for the user's broker directory. Can you pull in those changes and see if they work with this PR? Thanks!

robottelo/config/facade.py
mirekdlugosz: From what I see, `broker_directory` is read from settings and immediately put in environmental variable. Nothing in robottelo tries to read anything from `settings.broker` object - I assume broker uses this environmental variable.

This reading and variable setting is done inside legacy Settings class, before facade is constructed. Since nothing tries to fetch this value through facade, there is no need for facade to treat this setting as exception.

And even then - facade needs exceptions only for settings that are serialized to JSON (as Python's `json` module can serialize only basic data structures and does not allow objects to define how they should be serialized), or are passed to some picky modules that look into actual variable type. No need to put every single setting into exceptions list.

robottelo/config/validators.py
mirekdlugosz: Considering that right now validator exceptions are logged and ignored (we must do that until we actually mandate dynaconf on every robottelo user, that is until around phase 3), I'm leaning towards separate, smaller PR that will add all missing validators and remove these that are not needed. Such PR would be submitted after this one is merged.
How does that sound?

mirekdlugosz:
@JacobCallahan I extended exceptions list with some new settings added since last rebase in 3524735f6202320aba3297d9da8fb394fddfd7e4 , but broker directory is not on this list. Please see https://github.com/SatelliteQE/robottelo/pull/8003#discussion_r516879408 for longer explanation of rationale behind that decision. Let me know if I'm missing something.

JacobCallahan (approved):
ACK, thanks!

PR #8023, opened 2020-09-30
test fix test_negative_without_attach_with_lce

tstrych:
I am trying to firstly fix this issue before I unittest2pytest conversion.  I will rewrite the tests all at once. 

Why the change was necessary:

[test_negative_without_attach_with_lce](https://github.com/tstrych/robottelo/blob/5322c1f8da76187832ee9ebc356b2ce882ab4cf1/tests/foreman/cli/test_host.py#L2369) should fail with [registration](https://github.com/tstrych/robottelo/blob/5322c1f8da76187832ee9ebc356b2ce882ab4cf1/tests/foreman/cli/test_host.py#L2211) of `self.repository_id ` which is set in [setupclass to rhst7](https://github.com/tstrych/robottelo/blob/5322c1f8da76187832ee9ebc356b2ce882ab4cf1/tests/foreman/cli/test_host.py#L2098)

But in 6.8 the test was failing as it should,
because in the setupclass in [setup_org_for_a_rh_repo](https://github.com/tstrych/robottelo/blob/5322c1f8da76187832ee9ebc356b2ce882ab4cf1/tests/foreman/cli/test_host.py#L2083), and then in [_setup_org_for_a_rh_repo ](https://github.com/SatelliteQE/robottelo/blob/8257be03b9364545492ca05cdf975c22004e6659/robottelo/cli/factory.py#L1826) there is repository-set enabled. And this repository-set is also `rhst7`. So you should be able to register with that client as it already has the repository-set enabled, synced. So I just disable the repository-set. To achieve  that content-view has to be cleaned and removed. 

Now the interesting part, in Satellite6 #217  which was for  Satellite rhel7 6.7.0-21.0 6.7  it [passed]( Satellite rhel7 6.7.0-21.0 6.7 ).
I found where the problem is, in that case in was in setupclass which was done [here in #217 ](https://reportportal-sat-qe.cloud.paas.psi.redhat.com/ui/#satellite6/launches/all%7Cpage.page=1&page.size=50&filter.has.tags=rhel7&page.sort=start_time,number%2CDESC/5f3cc16952d69c000169fda1%7Cpage.page=1&page.size=50&filter.cnt.name=test_negative_without_attach&page.sort=start_time%2CASC?page.page=1&page.size=50&filter.cnt.name=test_negative_without_attach&page.sort=start_time%2CASC&log.item=5f3cc1b752d69c000169fe72)
where Red Hat Satellite Capsule was enabled instead of Red Hat Satellite Tools. 

`attach_to_default` can be found only in `test_negative_without_attach_with_lce`

Test result:
```
collected 64 items / 63 deselected / 1 selected

test_host.py::HostSubscriptionTestCase::test_negative_without_attach_with_lce 2020-09-30 14:31:25 - robottelo - DEBUG - Finished Test: HostSubscriptionTestCase/test_negative_without_attach_with_lce
2020-09-30 14:31:25 - robottelo - INFO - Started tearDownClass: tests.foreman.cli.test_host/HostSubscriptionTestCase
--warnings ommitted 
============ 1 passed, 63 deselected, 2 warnings in 333.19 seconds =========

### DIFF
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -2122,12 +2122,7 @@ class HostSubscriptionTestCase(CLITestCase):
         self.client.install_katello_ca()
 
     def _register_client(
-        self,
-        activation_key=None,
-        lce=False,
-        enable_repo=False,
-        auto_attach=False,
-        attach_to_default=False,
+        self, activation_key=None, lce=False, enable_repo=False, auto_attach=False,
     ):
         """Register the client as a content host consumer
 
@@ -2139,13 +2134,8 @@ class HostSubscriptionTestCase(CLITestCase):
         :param auto_attach: boolean to indicate whether to register with
             auto-attach option, in case of registration with activation key a
             command is launched
-        :param attach_to_default: boolean to indicate whether to attach to
-            plain RHEL subsctiption
         :return: the registration result
         """
-        assert (
-            not auto_attach or not attach_to_default
-        ), 'Only one of auto_attach or attach_to_default must be set'
 
         if activation_key is None:
             activation_key = self.activation_key
@@ -2162,43 +2152,6 @@ class HostSubscriptionTestCase(CLITestCase):
             )
             if auto_attach and self.client.subscribed:
                 result = self.client.run('subscription-manager attach --auto')
-
-        if attach_to_default:
-
-            # Remove repository-set from SetUpClass to have no repositories available
-            # Firstly the content view needs to be removed
-            default_cv = ContentView.info({'name': DEFAULT_CV, 'organization-id': self.org['id']})
-            default_lce = LifecycleEnvironment.info(
-                {'name': ENVIRONMENT, 'organization-id': self.org['id']}
-            )
-            ContentView.remove(
-                {
-                    'id': self.content_view['id'],
-                    'lifecycle-environments': ",".join(
-                        [ENVIRONMENT, self.env['name'], self.hosts_env['name']]
-                    ),
-                    'organization-id': self.org['id'],
-                    'system-content-view-id': default_cv['id'],
-                    'system-environment-id': default_lce['id'],
-                    'key-content-view-id': default_cv['id'],
-                    'key-environment-id': default_lce['id'],
-                }
-            )
-            ContentView.delete({'id': self.content_view['id']})
-            RepositorySet.disable(
-                {
-                    'basearch': 'x86_64',
-                    'name': REPOSET['rhst7'],
-                    'organization-id': self.org['id'],
-                    'product': PRDS['rhel'],
-                }
-            )
-            result = self.client.run(
-                'subscription-manager list --available --matches "%s" --pool-only'
-                % DEFAULT_SUBSCRIPTION_NAME
-            )
-            pool_id = result.stdout[0]
-            result = self.client.run('subscription-manager attach --pool "%s"' % pool_id)
 
         if self.client.subscribed and enable_repo:
             self.client.enable_repo(self.repository_id)
@@ -2357,6 +2310,7 @@ class HostSubscriptionTestCase(CLITestCase):
     def test_negative_without_attach_with_lce(self):
         """Attempt to enable a repository of a subscription that was not
         attached to a host
+        This test is not using the setUpClass Entities except subscription_name and repository_id
 
         :id: fc469e70-a7cb-4fca-b0ea-3c9e3dfff849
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: tests/foreman/cli/test_host.py [hunk 1]
```diff
@@ -2163,6 +2164,35 @@ def _register_client(
...
+            default_cv = ContentView.info({'name': DEFAULT_CV, 'organization-id': self.org['id']})
+            default_lce = LifecycleEnvironment.info(
+                {'name': ENVIRONMENT, 'organization-id': self.org['id']}
+            )
+            ContentView.remove(
+                {
+                    'id': self.content_view['id'],
+                    'lifecycle-environments': f"{ENVIRONMENT},"
+                    + f"{self.env['name']},"
+                    + f"{self.hosts_env['name']}",
```
mirekdlugosz, line 2179: Could you move that to variable that is set before `ContentView.remove`? It's really hard to read.

```python
lcenvs = ",".join([ENVIRONMENT, self.env['name'], self.hosts_env['name'] ])
```

tests/foreman/cli/test_host.py
```diff
@@ -2163,6 +2164,35 @@ def _register_client(
...
 
         if attach_to_default:
+
+            # Remove repository-set from SetUpClass to have no repositories available
+            # Firstly the content view needs to be removed
+            default_cv = ContentView.info({'name': DEFAULT_CV, 'organization-id': self.org['id']})
+            default_lce = LifecycleEnvironment.info(
+                {'name': ENVIRONMENT, 'organization-id': self.org['id']}
+            )
+            ContentView.remove(
```
mirekdlugosz, line 2174: Is it smart to remove contentview and disable repository set? Don't you modify global config this way? What about next test that uses `_register_client`, but with `attach_to_default` set to `False`? Would it still pass?

mirekdlugosz (commented):
My main concern is that this test modifies global Satellite setting and while this particular test may pass, it does so at expense of subsequent tests.

If you can reassure us this is not the case, I'll be happy to ACK.

tests/foreman/cli/test_host.py
tstrych: I was told to use f string, but I agree it's hard to read it. 
Yes I should use my mind, and should choosethe different approach when the f string is on multiple lines :+1:

tests/foreman/cli/test_host.py
tstrych: As I mentioned in description of this PR,  there is only one test which use attach_to_default. 
And as each test creates his own [host ](https://github.com/SatelliteQE/robottelo/pull/8023/files#diff-c5074666cd0572f014a9882591a5cce5L2115). I think there should be no problem with that. 

But what I do not like is that in _register client is an option that is used only for one test, so in the future when another test will be added, another work would be required.

I will run multiple test and I will provide the result.

tstrych: @mirekdlugosz  ^^

mirekdlugosz: OK, this reassurance is all I needed.

mirekdlugosz (approved):
ACK

tstrych:
@mirekdlugosz you are right, the problem is not with host but with life-cycle environment. As it's removed in one test and another test is using it.

tstrych:
I am still wondering if this could be done with less resources. As right know the test itself takes 8 minutes on my local machine and there is a lot of entities,  sync and publish that has to be done again. 

Test results. 
```
collected 64 items / 55 deselected / 9 selected

test_host.py::HostSubscriptionTestCase::test_negative_without_attach 2020-10-08 19:27:40 - robottelo - DEBUG - Finished Test: HostSubscriptionTestCase/test_negative_without_attach

test_host.py::HostSubscriptionTestCase::test_negative_without_attach_with_lce 2020-10-08 19:35:49 - robottelo - DEBUG - Finished Test: HostSubscriptionTestCase/test_negative_without_attach_with_lce

test_host.py::HostSubscriptionTestCase::test_positive_attach 2020-10-08 19:39:14 - robottelo - DEBUG - Finished Test: HostSubscriptionTestCase/test_positive_attach

test_host.py::HostSubscriptionTestCase::test_positive_attach_with_lce 2020-10-08 19:42:02 - robottelo - DEBUG - Finished Test: HostSubscriptionTestCase/test_positive_attach_with_lce

test_host.py::HostSubscriptionTestCase::test_positive_auto_attach 2020-10-08 19:44:46 - robottelo - DEBUG - Finished Test: HostSubscriptionTestCase/test_positive_auto_attach

test_host.py::HostSubscriptionTestCase::test_positive_register 2020-10-08 19:46:37 - robottelo - DEBUG - Finished Test: HostSubscriptionTestCase/test_positive_register

test_host.py::HostSubscriptionTestCase::test_positive_remove 2020-10-08 19:48:59 - robottelo - DEBUG - Finished Test: HostSubscriptionTestCase/test_positive_remove

test_host.py::HostSubscriptionTestCase::test_positive_unregister 2020-10-08 19:51:21 - robottelo - DEBUG - Finished Test: HostSubscriptionTestCase/test_positive_unregister

test_host.py::HostSubscriptionTestCase::test_syspurpose_end_to_end 2020-10-08 19:53:45 - robottelo - DEBUG - Finished Test: HostSubscriptionTestCase/test_syspurpose_end_to_end
2020-10-08 19:53:45 - robottelo - INFO - Started tearDownClass: tests.foreman.cli.test_host/HostSubscriptionTestCase
============ 9 passed, 55 deselected, 3 warnings in 1996.95 seconds ============
```

rplevka (approved):
ack

PR #8027, opened 2020-10-01
qe_test_coverage for BZ:1670397

akhil-jha:
When the user is created manually(i.e., onthefly disabled) from an authsource, the attributes such as first name, last name and email should be synced with LDAP.

### DIFF
--- a/tests/foreman/ui/test_ldap_authentication.py
+++ b/tests/foreman/ui/test_ldap_authentication.py
@@ -1401,3 +1401,59 @@ def test_timeout_and_cac_card_ejection():
 
     :expectedresults: Satellite should terminate the session after mentioned timeout in setting
     """
+
+
+@tier2
+@pytest.mark.skip_if_open("BZ:1670397")
+def test_verify_attribute_of_users_are_updated(session, ldap_data, ldap_tear_down):
+    """Verify if attributes of LDAP user are updated upon first login when
+    onthefly is disabled
+
+    :id: 163b346c-03be-11eb-acb9-0c7a158cbff4
+
+    :Steps:
+        1. Create authsource with onthefly disabled
+        2. Create a user manually and select the authsource created
+        3. Attributes of the user (like names and email) should be synced.
+
+    :BZ: 1670397
+
+    :expectedresults: The attributes should be synced.
+    """
+    ldap_auth_name = gen_string('alphanumeric')
+    auth_source_name = 'LDAP-' + ldap_auth_name
+    with session:
+        session.ldapauthentication.create(
+            {
+                'ldap_server.name': ldap_auth_name,
+                'ldap_server.host': ldap_data['ldap_hostname'],
+                'ldap_server.server_type': LDAP_SERVER_TYPE['UI']['ad'],
+                'account.account_name': ldap_data['ldap_user_name'],
+                'account.password': ldap_data['ldap_user_passwd'],
+                'account.base_dn': ldap_data['base_dn'],
+                'account.groups_base_dn': ldap_data['group_base_dn'],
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1401,3 +1401,59 @@ def test_timeout_and_cac_card_ejection():
...
+    :BZ: 1670397
+
+    :expectedresults: The attributes should be synced.
+    """
+    ldap_auth_name = gen_string('alphanumeric')
+    auth_source_name = 'LDAP-' + ldap_auth_name
+    with session:
+        session.ldapauthentication.create(
+            {
+                'ldap_server.name': ldap_auth_name,
```
mirekdlugosz, line 1428: I don't see "onthefly" being disabled here. Looks like this test depends on this particular setting default value, and I can't say I'm happy about this.

Could you set that option explicitly here?

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1401,3 +1401,59 @@ def test_timeout_and_cac_card_ejection():
...
+    ) as ldapsession:
+        with raises(NavigationTriesExceeded) as error:
+            ldapsession.user.search('')
+        assert error.typename == "NavigationTriesExceeded"
+    with session:
+        user_values = session.user.read(ldap_data['ldap_user_name'])
+        assert user_values['user']['login'] == ldap_data['ldap_user_name']
+        assert user_values['user']['firstname'] is not None
+        assert user_values['user']['lastname'] is not None
+        assert user_values['user']['mail'] is not None
```
mirekdlugosz, line 1459: Do we have control over these values? Could we use stronger assertion than "there is some value"?

tests/foreman/ui/test_ldap_authentication.py
akhil-jha: onthefly is disabled by default, anyway, I mentioned it explicitly.

tests/foreman/ui/test_ldap_authentication.py
akhil-jha: I have updated the code. Earlier the last name was not present in the AD. I have updated it.

mirekdlugosz (approved):
Code looks OK, thanks for including changes from comments.
Please add test result in comment.

akhil-jha:
The bug is still open(New state), the test will be skipped.
```
pytest tests/foreman/ui/test_ldap_authentication.py -k test_verify_attribute_of_users_are_updated
Test session starts (platform: linux, Python 3.8.5, pytest 4.6.3, pytest-sugar 0.9.4)
rootdir: /home/akjha/satelliteqe/robottelo
plugins: mock-1.10.4, xdist-1.34.0, services-1.3.1, forked-1.3.0, sugar-0.9.4, cov-2.10.1
collecting ... 2020-10-06 12:43:40 - conftest - DEBUG - Collected 33 test cases

 tests/foreman/ui/test_ldap_authentication.py s                                         100% ██████████
========================================= warnings summary ============================================
Results (0.17s):
       1 skipped
      32 deselected
```

mirekdlugosz:
> The bug is still open(New state), the test will be skipped.

Ah, right, there's `skip_if_open` call. Sorry for the noise.

rplevka (approved):
I'm going to ack this, but I have some comments - 
- the bug doesn't seem to be bound to UI so I would proably use API for the test. It would take fraction of the time the UI test will take and provide much more consistent reults (there are occassional errors caused by slow UI, etc.)
- I'd move the part where you create the auth source to the fixture as it is not really a part of the test (this is probably more of a formality)
- you could provide the test results with the bz decorator removed. At least we could see the test has no other issue and fails at the assertion as expected

PR #8028, opened 2020-10-02
Improve and simplify SCA test

swadeley:
Simple Content Access test was very slow
  Use only CDN Tools repo for SCA tests
  Use RPMs from that repo to trigger errata

### DIFF
--- a/tests/foreman/cli/test_contentaccess.py
+++ b/tests/foreman/cli/test_contentaccess.py
@@ -24,7 +24,6 @@ from robottelo.cli.factory import setup_cdn_and_custom_repositories
 from robottelo.cli.factory import setup_virtual_machine
 from robottelo.cli.host import Host
 from robottelo.cli.package import Package
-from robottelo.config import settings
 from robottelo.constants import DISTRO_RHEL7
 from robottelo.constants import ENVIRONMENT
 from robottelo.constants import PRDS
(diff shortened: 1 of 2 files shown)


swadeley:
robottelo]$ git status
On branch improve_sca_test
nothing to commit, working tree clean
robottelo]$ pytest --show-capture=no tests/foreman/cli/test_contentaccess.py 
==================================== test session starts =======================================
platform linux -- Python 3.7.9, pytest-4.6.3, py-1.9.0, pluggy-0.13.1
rootdir: /sat6/robottelo
plugins: xdist-2.1.0, services-1.3.1, mock-1.10.4, forked-1.3.0, cov-2.10.1
collecting ... 2020-10-02 08:49:16 - conftest - DEBUG - Collected 4 test cases
collected 4 items                                                                                                

tests/foreman/cli/test_contentaccess.py ....                                                               [100%]
======================== 4 passed, 6 warnings in 528.50 seconds =============================

tests/foreman/cli/test_contentaccess.py
```diff
@@ -92,40 +91,28 @@ def setUpClass(cls):
         )
         # Create repositories
         cls.repos = [
-            # Red Hat Enterprise Linux 7
```
mirekdlugosz, line 95: Since `repos` is one-element list, I guess it would be good to refactor it little more - rename to `repo` and change all callers to treat it as dict with data, not as list of dicts.

mirekdlugosz (approved):
ACK with suggestion

tests/foreman/cli/test_contentaccess.py
mshriver: I like keeping this as plural, as the called function is written for plural, and it allows for that use instead of only supporting singular.

Another consideration for future efforts @swadeley is actually using entity models in the tests and functions used here, instead of a dictionary to represent a repository.  Instead of key references and dictionaries defined within the tests, actually use a `Repository` class to model the entity, and create an instance of it here for the test to use and pass to other functions.

mshriver (approved):

tests/foreman/cli/test_contentaccess.py
swadeley: @mirekdlugosz , the `setup_cdn_and_custom_repositories` function wants a "a list of dict repositories options", so I think I cannot just change `cls.repos = [ {` to `cls.repos = {` with just one dict.
Re-factor to use simpler `setup_org_for_a_rh_repo` is an option.

What do you think?

Thank you

mirekdlugosz: @swadeley I was thinking about internal uses in test here. You can always wrap in list when passing to helper.
Although it seems that this helper is used only handful of times in robottelo, and in other instances it gets one-element lists.
Using simpler helper sounds like a good approach, too.

swadeley: Thanks for suggestions

BTW I posted my comment above without having seen Mike's comment (as I had the this open and did not reload).

PR #8031, opened 2020-10-02
Robottelo Pytest Package Update to 5.4.3

ColeHiggins2:
Updating pytest from 4.6.3 to 5.4.3.

After running several tests in the API, CLI, and UI, It does not look like any of the failures are related to the pytest changes. Please review the standalone Jenkins jobs for automation results. Jobs are 2010, 2013, and 2027 in Jenkins. 

If you think anything else should be updated, Please comment below

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ mock==3.0.5
 navmazing==1.1.6
 paramiko==2.5.0
 productmd==1.26
-pytest==4.6.3
+pytest==5.4.3
 pytest-services==1.3.1
 pytest-mock==1.10.4
 selenium==3.141.0
--- a/tests/foreman/conftest.py
+++ b/tests/foreman/conftest.py
@@ -40,7 +40,7 @@ def pytest_report_header(config):
         if not scope:
             scope = ''
         storage = settings.shared_function.storage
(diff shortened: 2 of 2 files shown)


lpramuk:
https://docs.pytest.org/en/stable/changelog.html#pytest-5-1-0-2019-08-15

- []  pytest.raises and pytest.warns no longer support strings as the second argument.

lpramuk (changes requested):
Add more changes to be able to update

mshriver (approved):

lpramuk (approved):

jyejare:
@JacobCallahan @mshriver @mirekdlugosz There are many such PRs opened by py-up bot. To merge those PRs I was looking for there use in tests and helpers in robottelo. Else we should make some mechanism to auto-test the related tests wrt to those python modules. Do you anything already exists like that e.g genealogist. 

Thoughts ?

mshriver:
I'm not aware of existing tools to do such a targeted search/test for python packages. A general pull request tester (planned) could be combined with identification of package changes in the PR, and tests using that package. With the layers of function calls its certainly not a trivial problem to solve.

PR #8041, opened 2020-10-06
Convert test_organization from unittest to pytest.

tpapaioa:
This PR converts the tests in `tests/foreman/cli/test_organization.py` from unittest to pytest. Changes include:

- The `OrganizationTestCase` class has been removed, and all tests moved to the module level.
- Proxy setup and teardown is moved to a fixture named `proxy`.
- The class-level `org` attribute is now implemented with the existing module-scoped `module_org` fixture. Because `module_org` uses nailgun instead of cli methods, it is a nailgun entity instead of a python dictionary, so key/value access of the type `org['name']` had to be changed to attribute access: `org.name`.
- Unittest-style assertions are now regular `assert` statements.

```
$ pytest tests/foreman/cli/test_organization.py

[...]

collected 36 items

tests/foreman/cli/test_organization.py ........s........................ [ 91%]
...                                                                      [100%]

============ 35 passed, 1 skipped, 4 warnings in 667.94s (0:11:07) =============
```

### DIFF
--- a/tests/foreman/cli/test_organization.py
+++ b/tests/foreman/cli/test_organization.py
@@ -19,6 +19,7 @@ from fauxfactory import gen_string
 
 from robottelo.cleanup import capsule_cleanup
 from robottelo.cli.base import CLIReturnCodeError
+from robottelo.cli.factory import CLIFactoryError
 from robottelo.cli.factory import make_compute_resource
 from robottelo.cli.factory import make_domain
 from robottelo.cli.factory import make_hostgroup
@@ -37,10 +38,10 @@ from robottelo.config import settings
 from robottelo.constants import FOREMAN_PROVIDERS
 from robottelo.datafactory import filtered_datapoint
 from robottelo.datafactory import invalid_values_list
+from robottelo.datafactory import parametrized
 from robottelo.datafactory import valid_data_list
 from robottelo.datafactory import valid_org_names_list
 from robottelo.decorators import skip_if_not_set
-from robottelo.test import CLITestCase
 
 
 @filtered_datapoint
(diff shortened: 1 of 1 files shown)


latran:
@tpapaioa Please add test result in commit.

tpapaioa:
Waiting on PR https://github.com/SatelliteQE/robottelo/pull/8070 to fix `get_available_capsule_port()`, which is used by `make_proxy`.

tests/foreman/cli/test_organization.py
```diff
@@ -55,26 +55,31 @@ def valid_labels_list():
...
-        '{}_{}'.format(gen_string('alpha', 5), gen_string('alpha', 5)),
+        f"{gen_string('alpha', 5)}-{gen_string('alpha', 5)}",
+        f"{gen_string('alpha', 5)}-{gen_string('alpha', 5)}",
     ]
 
 
-class OrganizationTestCase(CLITestCase):
-    """Tests for Organizations via Hammer CLI"""
+@pytest.fixture
+def _make_proxy(request):
```
mshriver, line 64: I'd remove the leading `_`, as fixtures aren't imported between modules anyway.

mshriver: future TODO - use the core pytest fixtures for proxy and org creation, instead of CLI factory methods.

mshriver (approved):
Looks great, thanks @tpapaioa. Consider updating the fixture names, but its not required.

tests/foreman/cli/test_organization.py
```diff
@@ -55,26 +55,31 @@ def valid_labels_list():
...
+    yield org
+    org_cleanup(org['id'])
 
-    @classmethod
-    def setUpClass(cls):
-        """Create an organization."""
-        super().setUpClass()
-        cls.org = make_org()
+
+class TestOrganization:
```
JacobCallahan, line 81: any reason to keep the class?

JacobCallahan (changes requested):

tests/foreman/cli/test_organization.py
JacobCallahan: he has to keep the `_` because `make_proxy` is already int he namespace because of an import

mshriver: Can just use a different name for the fixture.

tpapaioa: Is there a problem with the name I chose?

tests/foreman/cli/test_organization.py
tpapaioa: Any reason to remove it? More work to remove the class for no benefit that I see, but if we're going to make it a rule to remove classes, I can do it.

tests/foreman/cli/test_organization.py
mshriver: I guess you could call it a personal problem. Its an odd pattern for a fixture name, as `_method` implies a 'private' method, which is implicit to defining a pytest fixture within a test module.

tpapaioa: Changed `_make_org` and `_make_proxy` to `org` and `proxy`.

tests/foreman/cli/test_organization.py
tpapaioa: Done.

tests/foreman/cli/test_organization.py
```diff
@@ -55,630 +56,601 @@ def valid_labels_list():
         gen_string('alpha'),
         gen_string('alphanumeric'),
         gen_string('numeric'),
-        '{}-{}'.format(gen_string('alpha', 5), gen_string('alpha', 5)),
-        '{}_{}'.format(gen_string('alpha', 5), gen_string('alpha', 5)),
+        f"{gen_string('alpha', 5)}-{gen_string('alpha', 5)}",
+        f"{gen_string('alpha', 5)}-{gen_string('alpha', 5)}",
```
lvrtelov, line 60: I think the change from "_" to "-" is oversight here. Please fix this. But everything else LGTM.

lvrtelov (changes requested):

tests/foreman/cli/test_organization.py
tpapaioa: Fixed.

lvrtelov (approved):

mshriver:
@JacobCallahan concern about keeping the test class has been addressed, and the module has been flattened.

PR #8042, opened 2020-10-06
Convert test_contentmanagement from unittest to pytest

tpapaioa:
The conversion of `tests/foreman/api/test_contentmanagement.py` to pytest makes the following changes:

1.) Change class names, remove inheritance from unittest class `APITestCase`.

`ContentManagementTestCase` -> `TestSatelliteContentManagement`
`CapsuleContentManagementTestCase` -> `TestCapsuleContentManagement`

2.) Change unittest `assert*` method calls to regular `assert` calls.
3.) Move capsule VM setup and teardown from unittest `setUpClass` and `tearDownClass` class methods to a class-scoped pytest fixture named `capsule_vm`. Use the fixture instead of a test class/instance attribute for tests that need it.

4.) Move `skip_if_not_set` decorators from the deleted `setUpClass` method to each test directly.
5.) Use f-strings.

Test output:

```
# pytest tests/foreman/api/test_contentmanagement.py
tests/foreman/api/test_contentmanagement.py s......s.....                [100%]

[...]
============ 11 passed, 2 skipped, 6 warnings in 2078.71s (0:34:38) ============
```

The two skipped tests are due to one unconditional `pytest.mark.skip` and one `pytest.mark.stubbed` decorator.

### DIFF
--- a/tests/foreman/api/test_contentmanagement.py
+++ b/tests/foreman/api/test_contentmanagement.py
@@ -21,7 +21,6 @@ import pytest
 from fauxfactory import gen_string
 from nailgun import client
 from nailgun import entities
-from nailgun.entity_mixins import TaskFailedError
 
 from robottelo import manifests
 from robottelo import ssh
@@ -56,13 +55,25 @@ from robottelo.helpers import get_data_file
 from robottelo.helpers import md5_by_url
 from robottelo.host_info import get_repo_files
 from robottelo.host_info import get_repomd_revision
-from robottelo.test import APITestCase
 from robottelo.utils.issue_handlers import is_open
 from robottelo.vm import VirtualMachine
 from robottelo.vm_capsule import CapsuleVirtualMachine
 
 
-class ContentManagementTestCase(APITestCase):
+@pytest.fixture(scope='class')
+def capsule_vm():
+    """Create a standalone capsule for tests"""
+    vm = CapsuleVirtualMachine()
+    vm.create()
+
+    vm._capsule = entities.Capsule().search(query={'search': f'name={vm.hostname}'})[0]
+
+    yield vm
+
+    vm.destroy()
+
+
+class TestSatelliteContentManagement:
     """Content Management related tests, which exercise katello with pulp
     interactions.
     """
@@ -88,8 +99,8 @@ class ContentManagementTestCase(APITestCase):
         for _ in range(2):
             product = entities.Product(organization=org).create()
             repo = entities.Repository(product=product, url=FAKE_7_YUM_REPO).create()
-            with self.assertNotRaises(TaskFailedError):
-                repo.sync()
+            response = repo.sync()
+            assert response, f"Repository {repo} failed to sync."
 
     @pytest.mark.tier2
     @pytest.mark.skipif((not settings.repos_hosting_url), reason='Missing repos_hosting_url')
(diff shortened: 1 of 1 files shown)


tests/foreman/api/test_contentmanagement.py
```diff
@@ -480,42 +501,23 @@ def test_positive_capsule_sync(self):
         sync_status = capsule.content_get_sync()
         last_sync_time = sync_status['last_sync_time']
 
-        # If BZ1439691 is open, need to sync repo once more, as repodata
-        # will change on second attempt even with no changes in repo
-        if is_open('BZ:1439691'):
```
mirekdlugosz, line 485: I see this BZ didn't get enough traction and was CLOSED WONTFIX as stale. Can we really remove this entire branch? Shouldn't we rather make it unconditional, since behavior we considered a bug is just part of how Satellite works?

mirekdlugosz (changes requested):
I like that you used that as opportunity to also do migration from `.format` to f-strings.

Travis failure is blocking.

As part of our PR review process, we require submitters to provide some kind of demonstration that provided changes do work. Usually it takes form of running some tests and pasting test results. Could you run all changed tests (this entire file, I guess) and provide results here?

tests/foreman/api/test_contentmanagement.py
```diff
@@ -765,6 +812,7 @@ def test_positive_mirror_on_sync(self):
 
         :CaseLevel: System
         """
+        breakpoint()
```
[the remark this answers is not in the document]
tpapaioa: Fixed.

tests/foreman/api/test_contentmanagement.py
tpapaioa: Fixed.

mirekdlugosz (approved):
Thanks for addressing all comments, and sorry for not replying here sooner!

ACK

tests/foreman/api/test_contentmanagement.py
```diff
@@ -88,8 +99,8 @@ def test_positive_sync_repos_with_large_errata(self):
         for _ in range(2):
             product = entities.Product(organization=org).create()
             repo = entities.Repository(product=product, url=FAKE_7_YUM_REPO).create()
-            with self.assertNotRaises(TaskFailedError):
-                repo.sync()
+            response = repo.sync()
```
mshriver, line 102: Consider using try/except here to call `pytest.fail` should the call to repo.sync raise TaskFailedError. That way the test result is failure, not error.

tests/foreman/api/test_contentmanagement.py
```diff
@@ -311,37 +307,43 @@ def test_positive_checksum_sync(self):
...
-        org = entities.Organization(smart_proxy=[self.capsule_id]).create()
+        org = entities.Organization(smart_proxy=[capsule_vm._capsule.id]).create()
         product = entities.Product(organization=org).create()
         repo = entities.Repository(
             product=product, checksum_type='sha256', download_policy='immediate'
         ).create()
         lce = entities.LifecycleEnvironment(organization=org).create()
         # Associate the lifecycle environment with the capsule
-        capsule = entities.Capsule(id=self.capsule_id).read()
+        capsule = entities.Capsule(id=capsule_vm._capsule.id).read()
```
mshriver, line 320: `entities.Capsule(id=capsule_vm._capsule.id` should be equivalent to `capsule_vm._capsule`, I don't think there's an actual need to re-create the entity instance here.  But you're just continuing the pattern, changing this is optional.

tests/foreman/api/test_contentmanagement.py
mshriver: There is actually a recommendation from pytest docs for consistently dealing with this type of behavior by creating a dummy context manger. 

https://docs.pytest.org/en/stable/example/parametrize.html#parametrizing-conditional-raising

Not for the scope of this PR, but something we could consider adding for all test modules to use.

mshriver (approved):

PR #8048, opened 2020-10-07
Remove confusingly named test case in cv

latran:
This test case is confusing, unnecessary, low priority, and plainly makes no sense.  The main goal is to subscribe to a repo, but it seems to be more focused on testing filters.  Both are captured else where already so I'm deleting this test case.


mirekdlugosz (approved):
I guess at one point in the past gofer/katello-agent was hard requirement for host subscription? Yeah, we don't do that anymore.

ACK

JacobCallahan:
@mirekdlugosz that is the strange part about this test, katello-agent was never a requirement for registration...

JacobCallahan (approved):
ACK

swadeley (commented):
ACK

PR #8049, opened 2020-10-08
Let's update the system before capsule installation.

vsedmik:
The test passes locally even without the update. However, the system should be up to date before installation.
I've also replaced wget by curl as wget is not preinstalled on satlab, making this fail while debugging there.

```
Launching pytest with arguments tests/foreman/cli/test_capsule_installer.py::CapsuleInstallerTestCase::test_positive_reinstall_on_same_node_after_remove in /home/vsedmik/PycharmProjects/robottelo

platform linux -- Python 3.7.6, pytest-4.6.3, py-1.8.0, pluggy-0.13.1 -- /home/vsedmik/PycharmProjects/robottelo/venv/bin/python
cachedir: .pytest_cache
rootdir: /home/vsedmik/PycharmProjects/robottelo
plugins: services-1.3.1, mock-1.10.4
collecting ... 2020-10-07 13:21:11 - conftest - DEBUG - Collected 1 test cases
collected 1 item

tests/foreman/cli/test_capsule_installer.py::CapsuleInstallerTestCase::test_positive_reinstall_on_same_node_after_remove 

venv/lib64/python3.7/site-packages/_pytest/mark/structures.py:337
  /home/vsedmik/PycharmProjects/robottelo/venv/lib64/python3.7/site-packages/_pytest/mark/structures.py:337: PytestUnknownMarkWarning: Unknown pytest.mark.capsule - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html
    PytestUnknownMarkWarning,

venv/lib64/python3.7/site-packages/_pytest/mark/structures.py:337
  /home/vsedmik/PycharmProjects/robottelo/venv/lib64/python3.7/site-packages/_pytest/mark/structures.py:337: PytestUnknownMarkWarning: Unknown pytest.mark.BZ_1327442 - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html
    PytestUnknownMarkWarning,

venv/lib64/python3.7/site-packages/_pytest/mark/structures.py:337
  /home/vsedmik/PycharmProjects/robottelo/venv/lib64/python3.7/site-packages/_pytest/mark/structures.py:337: PytestUnknownMarkWarning: Unknown pytest.mark.high - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html
    PytestUnknownMarkWarning,

=================== 1 passed, 3 warnings in 2133.95 seconds ====================
```

### DIFF
--- a/robottelo/vm_capsule.py
+++ b/robottelo/vm_capsule.py
@@ -9,6 +9,7 @@ from fauxfactory import gen_alphanumeric
 from robottelo import ssh
 from robottelo.cli.capsule import Capsule
 from robottelo.cli.host import Host
+from robottelo.cli.settings import Settings
 from robottelo.config import settings
 from robottelo.constants import SATELLITE_FIREWALL_SERVICE_NAME
 from robottelo.decorators import setting_is_set
@@ -231,12 +232,21 @@ class CapsuleVirtualMachine(VirtualMachine):
         )
         self.configure_rhel_repo(settings.__dict__[self.distro + '_repo'])
         self.run('yum repolist')
+        self.run('yum -y update')
         self.run('yum -y install satellite-capsule', timeout=1200)
         result = self.run('rpm -q satellite-capsule')
         if result.return_code != 0:
             raise CapsuleVirtualMachineError(
                 'Failed to install satellite-capsule package\n{}'.format(result.stderr)
             )
+        # update http proxy except list
+        result = Settings.list({'search': 'http_proxy_except_list'})[0]
+        if result["value"] == "[]":
+            except_list = '[{0}]'.format(self.hostname)
+        else:
+            except_list = result["value"][:-1] + ', {0}]'.format(self.hostname)
+        Settings.set({'name': 'http_proxy_except_list', 'value': except_list})
+        # generate certificate
         cert_file_path = '/root/{0}-certs.tar'.format(self.hostname)
         certs_gen = ssh.command(
             'capsule-certs-generate '
--- a/robottelo/vm.py
+++ b/robottelo/vm.py
@@ -539,7 +539,7 @@ gpgcheck=0'''.format(
                 password = settings.server.admin_password
 
             cmd += ' --consumerid {0} --username {1} --password {2}'.format(
-                consumerid, username, password,
+                consumerid, username, password
             )
             if auto_attach:
                 cmd += ' --auto-attach'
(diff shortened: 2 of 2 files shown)


mirekdlugosz (commented):
I guess it makes sense and updating is something we tell customers, but I'm curious about test run time impact of that change. How long does update itself take, right now?

I see this test is already at the longer side (35 minutes!), so maybe skipping update to save minute or two is not worth it... But I'd like to have that conversation anyway.

vsedmik:
The update itself take 4m15s when running against vm on ginger.

mirekdlugosz:
So, that's increase of execution time of around 13%. That's more than I personally would like...  
I'll let others voice their opinion. If we don't get that resolved in week or so, please ping me off Github.

Code-wise, that PR is OK.

vsedmik:
I do understand, it's really long. Apart from the `test_positive_reinstall_on_same_node_after_remove` test, the instances of `CapsuleVirtualMachine` class are used in setup of test_contentmanagement.py and in one test of test_contentview.py. Which means plus ~8m30s for automation run as it is today, so it's even worse.

On the other hand, we have already bumped into issues when satellite/capsule was installed on a non-updated system, so I would rather stick to the proper way. Also, if an updated rhel image is used by CI, the output of yum update would be 'nothing to do' with a very little time impact.

Last but not least - this particular test checks a feature which is no longer supported and hacks it by acquiring katello-remove from the artifacts server. So there is a way to potentially save 2134 seconds as easy as snapping a finger and still having shorter total run time than we have today.

swadeley:
> So, that's increase of execution time of around 13%. That's more than I personally would like...

but I think some of that time will be saved by the installer not causing package updates

+1 to the use of `curl` not `wget`

swadeley (approved):
ACK

swadeley:
I wonder how much time adding this line before the yum update would save:
`yum install -y deltarpm`

vsedmik:
It seems the deltarpm installation hasn't affected the update duration significantly.

swadeley:
> It seems the deltarpm installation hasn't affected the update duration significantly.

thanks for testing, it could be worth testing it where tests are pulling from CDN (cause that will have delta RPMs)

robottelo/vm_capsule.py
```diff
@@ -231,12 +231,17 @@ def _setup_capsule(self):
...
         self.run('yum -y install satellite-capsule', timeout=1200)
         result = self.run('rpm -q satellite-capsule')
         if result.return_code != 0:
             raise CapsuleVirtualMachineError(
                 'Failed to install satellite-capsule package\n{}'.format(result.stderr)
             )
+        ssh.command(
+            'hammer setting set --name http_proxy_except_list --value [{0}]'.format(self.hostname),
+            hostname=settings.server.hostname,
+        )
```
jameerpathan111, line 244: @vsedmik I looked into this and noticed that we can import Settings from robottelo.cli.settings here as well, so we can change code to:

```
result = Settings.list({"search": 'name=http_proxy_except_list'})[0]
if result["value"] == "[]":
    except_list = '[{0}]'.format(self.hostname)
else:
    except_list = result["value"][:-1] + ', {0}]'.format(self.hostname)
Settings.set({'name': 'http_proxy_except_list', 'value': except_list})
```

jameerpathan111 (approved):
LGTM

mirekdlugosz (approved):
Since nobody else raises concern about performance, I'm ACKing and merging, as promised last week.

PR #8051, opened 2020-10-08
Add ibutsu-pytest to requirements

mshriver:
Will allow for pushing results to ibutsu.

Adding this to CI so that we can push results, not so that component owners can use them.


JacobCallahan (approved):
ACK

lhellebr (approved):
LGTM.
Just tried to install the package without any issues so this can't hurt.

jyejare:
@mshriver Any long term goal behind pushing the results to ibutsu ?

PR #8055, opened 2020-10-09
[6.6.z] Pin azure (wrapanapi dep.) version to work-around installation failure

lpramuk:
Backports https://github.com/SatelliteQE/robottelo/pull/7730 to `6.6.z`
Fixes stylecheck `E225` and `F523`
```
wrapanapi requires azure>=3.0.0, but 5.0.0 introduced
backwards-incompatible change of dropping "azure" package. Specific
subpackages should be used instead.

Fix should be applied in wrapanapi setup file. Then it would only be
matter of pulling in new version of it as Robottelo dependency. The problem
is, we currently depend on 3.2.0, and fixed package would be at least 3.5.7.

To allow us to still use old version of wrapanapi, I'm introducing
direct dependency on correctly versioned "azure" package and abuse the
fact that pip lacks proper dependency solver - it will pull in our
azure, and once it comes to wrapanapi, it will consider its dependency on
azure to be met.
```


lpramuk:
Fixes the install error:
```
Collecting azure>=3.0.0

  Downloading azure-5.0.0.zip (4.6 kB)

    ERROR: Command errored out with exit status 1:

     command: /home/travis/virtualenv/python3.6.7/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-tqsmrxoj/azure/setup.py'"'"'; __file__='"'"'/tmp/pip-install-tqsmrxoj/azure/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-oxf8vblb

         cwd: /tmp/pip-install-tqsmrxoj/azure/

    Complete output (24 lines):

      File "/tmp/pip-install-tqsmrxoj/azure/setup.py", line 60, in <module>

        raise RuntimeError(message)

    RuntimeError:

    

    Starting with v5.0.0, the 'azure' meta-package is deprecated and cannot be installed anymore.

    Please install the service specific packages prefixed by `azure` needed for your application.
```

jyejare (approved):
ACK

mshriver (approved):

JacobCallahan (approved):
ACK

PR #8059, opened 2020-10-14
Fixed satellite installer services check test case

devendra104:
Removed "foreman" from the services list in the test_positive_check_installer_services test case.


mirekdlugosz (approved):
Expected results in line 1385 still refer `foreman-tasks` service. Could you remove that reference as well?

LGTM otherwise

devendra104:
> Expected results in line 1385 still refer `foreman-tasks` service. Could you remove that reference as well?
> 
> LGTM otherwise

Fixed.

omkarkhatavkar (approved):
ACK

omkarkhatavkar:
@devendra104 travis is not happy, can you make it happy?

PR #8061, opened 2020-10-15
cherry pick - test syspurpose fix

tstrych:
#7996

PR #8062, opened 2020-10-15
cherry pick - satellite tools constants changed to 6.7

tstrych:
#7741


tstrych:
I forgot to cp this into 6.7, now I found it. 
Also cherry picking #8061  #8063 .
Changes in mentioned cp are based on this one. 

Travis failure seems to be not connected to this PR. 
```
Warning, treated as error:
/home/travis/build/SatelliteQE/robottelo/docs/autoapi/robottelo/cli/factory/index.rst:87:<autosummary>:2:Literal block expected; none found.
Makefile:45: recipe for target 'html' failed
make[1]: *** [html] Error 2
make[1]: Leaving directory '/home/travis/build/SatelliteQE/robottelo/docs'
Makefile:60: recipe for target 'docs' failed
make: *** [docs] Error 2
The command "make docs" exited with 2.
```

mirekdlugosz (approved):
ACK

> Would be glad if you can give me an advice for future.

We don't have established policy or guideline for cherry-picks into z-streams.

It's important to remember that `git cherry-pick` works on single **commit**. Usually one commit == one PR, but sometimes there are multiple commits in PR. In that case, the key is to cherry-pick all commits. They can be squashed before push to maintain "single commit (cherry-pick) in z-stream == single PR in master".

You can create separate PR for each cherry-pick / PR from master. However, if these cherry-picks are related and it doesn't make sense to pull one without others (especially if one depends on others and it's important to maintain order of merging), it might make more sense to submit them all in single PR. In that case, it's best to add comment for reviewers that they shouldn't use "Squash and merge", as we want to maintain relationships between commits and PRs.

At the end of the day, only important thing is whether next person will be able to understand what happened and why. As long as you put original commit ids somewhere (in GitHub comment or in commit message), that goal should be achieved.

rplevka (approved):

PR #8069, opened 2020-10-19
added the user-group sync test with external auth rhsso

omkarkhatavkar:
### Test Result 

```
Testing started at 9:14 PM ...
/home/okhatavk/okhatavk/Satellite/robottelo/env/bin/python /home/okhatavk/Downloads/pycharm-community-2018.1.4/helpers/pycharm/_jb_pytest_runner.py --target test_ldap_authentication.py::test_user_permissions_rhsso_user_multiple_group
Launching py.test with arguments test_ldap_authentication.py::test_user_permissions_rhsso_user_multiple_group in /home/okhatavk/Satellite/robottelo/tests/foreman/ui

plugins: forked-1.1.3, cov-2.8.1, mock-1.10.4, xdist-1.34.0, lazy-fixture-0.6.3, services-1.3.12020-10-19 15:44:12 - conftest - DEBUG - Collected 1 test cases
collected 1 item [100%]



=================== 1 passed, 2 warnings in 1072.56 seconds ====================
```

### DIFF
--- a/robottelo/constants/__init__.py
+++ b/robottelo/constants/__init__.py
@@ -1761,6 +1761,12 @@ RHSSO_NEW_USER = {
     "username": "random_name",
 }
 
+RHSSO_USER_UPDATE = {'realm': "realm_name", "userId": "user_id"}
+
+RHSSO_NEW_GROUP = {
+    "name": "group_name",
+}
+
 RHSSO_RESET_PASSWORD = {"temporary": "false", "type": "password", "value": ""}
 
 FOREMAN_ANSIBLE_MODULES = [
--- a/tests/foreman/ui/test_ldap_authentication.py
+++ b/tests/foreman/ui/test_ldap_authentication.py
@@ -28,6 +28,9 @@ from pytest import skip
 
 from robottelo import ssh
 from robottelo.api.utils import create_role_permissions
+from robottelo.cli.factory import make_usergroup
+from robottelo.cli.factory import make_usergroup_external
+from robottelo.cli.ldapauthsource import ExternalAuthSource
 from robottelo.config import settings
 from robottelo.constants import CERT_PATH
 from robottelo.constants import LDAP_ATTR
@@ -42,10 +45,14 @@ from robottelo.decorators import skip_if_not_set
 from robottelo.decorators import tier2
 from robottelo.decorators import upgrade
 from robottelo.helpers import file_downloader
+from robottelo.rhsso_utils import create_group
 from robottelo.rhsso_utils import create_new_rhsso_user
+from robottelo.rhsso_utils import delete_rhsso_group
 from robottelo.rhsso_utils import delete_rhsso_user
 from robottelo.rhsso_utils import get_rhsso_client_id
 from robottelo.rhsso_utils import run_command
+from robottelo.rhsso_utils import update_rhsso_user
+
 
 pytestmark = [run_in_one_thread]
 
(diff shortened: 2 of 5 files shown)


robottelo/constants/__init__.py
```diff
@@ -1761,6 +1761,12 @@
     "username": "random_name",
 }
 
+UPDATE_USER = {'realm': "realm_name", "userId": "user_id"}
```
mirekdlugosz, line 1764: Does that need to be in constants?
Seems you only use dict keys, as you replace values in fixture.

Either way, this name is too generic and it's hard to make connection between name and actual content. Perhaps something like `RHSSO_USER_KEYS_TO_UPDATE` would be more fitting.

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1143,6 +1168,66 @@ def test_login_failure_rhsso_user_if_internal_user_exist(
...
+    for group_name, argument in zip(group_names, arguments):
+        # adding/creating rhsso groups
+        create_group(group_name=group_name)
+        update_rhsso_user(username, group_name=group_name)
+        argument['name'] = group_name
+
+        # creating satellite external groups
+        user_group = make_usergroup(argument)
+        make_usergroup_external(
+            {'auth-source-id': '3', 'user-group-id': user_group['id'], 'name': group_name}
```
mirekdlugosz, line 1219: Is it always going to be `'3'`? Could we obtain this value dynamically?
Otherwise, that might be good candidate for constant.

mirekdlugosz (commented):
Why is this PR Draft? Code looks complete, and you even have test results. Could you edit initial message with explanation why is that drafted?

Code looks OK, except for small comments. I would have ACKed it.

tests/foreman/ui/test_ldap_authentication.py
omkarkhatavkar: Yes I think we can get it, will update based on that

omkarkhatavkar:
> ode looks OK, except for small comments. I would have ACKed it.

I wanted to add more group tests

robottelo/constants/__init__.py
omkarkhatavkar: Just to make easier I making it as `RHSSO_USER_UPDATE` not too long and not generic as well

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1143,6 +1169,131 @@ def test_login_failure_rhsso_user_if_internal_user_exist(
...
+    :CaseImportance: Medium
+
+    :steps:
+        1. create usergroup with admin permissions respectively
+        2. assigned that group to rhsso user
+        3. verify the permission of the rhsso user in Satellite
+        4. delete the rhsso group
+
+    :expectedresults: external rhsso user's permissions should get revoked after external rhsso
+        group deletion.
```
mirekdlugosz, line 1190: How does that work for user that is already logged in? Will log him out / revoke permissions on next action, or only on next login after session timeout?

That would be stronger assertion and worth checking, too.

mirekdlugosz (approved):
ACK with question/suggestion

tests/foreman/ui/test_ldap_authentication.py
omkarkhatavkar: This is after session timeout and logout I have tested, I have already test case around timeout. I will once again cross check if it on the browser refresh. If yes will update the test case, otherwise I don't see any change needs to be made

omkarkhatavkar: @mirekdlugosz I think whatever I have added it make sense. We already have timeout test case, if we include this then the test execution time will increase. We can not use logout here because in RHSSO there nothing called hard logout. If you still have query we can discuss this offline.

mirekdlugosz (approved):

pondrejk (approved):

PR #8070, opened 2020-10-19
Fix get_available_capsule_port() and usage of valid_data_list() in Capsule tests

tpapaioa:
The helper method `get_available_capsule_port()` was calling the `fuser` command on the Satellite to get unused ports within a given range. This command is not installed by default in recent versions of Satellite, so I've changed the method to use `ss` (provided by the `iproute` rpm) instead.

Also, some capsule tests were iterating over the keys of the `dict` returned by `valid_data_list()`, instead of its values. Those tests have been fixed.

Code coverage checks required a unit test for `get_available_capsule_port()`, so I've added one in `tests/robottelo/test_helpers.py`.

Test results:

- Helper method unit test:

```
$ pytest -k test_return_port tests/robottelo/test_helpers.py
platform linux -- Python 3.8.5, pytest-4.6.3, py-1.9.0, pluggy-0.13.1
rootdir: /home/tpapaioa/PycharmProjects/robottelo
plugins: services-1.3.1, forked-1.3.0, xdist-1.34.0, mock-1.10.4
collected 15 items / 14 deselected / 1 selected                                                                                                                                                                                              

tests/robottelo/test_helpers.py .                                                                                                                                                                                                      [100%]

================================================================================================== 1 passed, 14 deselected in 0.05 seconds ===================================================================================================
```

- Capsule tests (skipped tests are all marked with `pytest.mark.stubbed`):

```
$ pytest tests/foreman/cli/test_capsule.py
platform linux -- Python 3.8.5, pytest-4.6.3, py-1.9.0, pluggy-0.13.1
rootdir: /home/tpapaioa/PycharmProjects/robottelo
plugins: services-1.3.1, forked-1.3.0, xdist-1.34.0, mock-1.10.4
collecting ... 2020-10-19 18:04:01 - conftest - DEBUG - Collected 16 test cases
collected 16 items                                                                                                                                                                                                                           

tests/foreman/cli/test_capsule.py .......sssssssss                                                                                                                                                                                     [100%]

[...]

============================================================================================= 
7 passed, 9 skipped, 10 warnings in 407.32 seconds =============================================================================================
```

### DIFF
--- a/robottelo/helpers.py
+++ b/robottelo/helpers.py
@@ -328,8 +328,8 @@ def add_remote_execution_ssh_key(hostname, key_path=None, proxy_hostname=None, *
 
 def get_available_capsule_port(port_pool=None):
     """returns a list of unused ports dedicated for fake capsules
-    This calls a fuser command on the server prompting for a port range. fuser
-    commands returns a list of ports which have a PID assigned (a list of ports
+    This calls an ss command on the server prompting for a port range. ss
+    returns a list of ports which have a PID assigned (a list of ports
     which are already used). This function then substracts unavailable ports
     from the other ones and returns one of available ones randomly.
 
@@ -350,21 +350,22 @@ def get_available_capsule_port(port_pool=None):
                 f'got {type(port_pool_range)} instead'
             )
     # returns a list of strings
-    fuser_cmd = ssh.command(
-        f'fuser -n tcp {{{port_pool[0]}..{port_pool[1]}}} 2>&1 | awk -F/ \'{{print$1}}\''
+    ss_cmd = ssh.command(
+        f"ss -tnaH sport ge {port_pool[0]} sport le {port_pool[-1]}"
+        " | awk '{n=split($4, p, \":\"); print p[n]}' | sort -u"
     )
-    if fuser_cmd.stderr:
+    if ss_cmd.stderr:
         raise CapsuleTunnelError(
-            f'Failed to create ssh tunnel: Error getting port status: {fuser_cmd.stderr}'
+            f'Failed to create ssh tunnel: Error getting port status: {ss_cmd.stderr}'
         )
     # converts a List of strings to a List of integers
     try:
-        print(fuser_cmd)
-        used_ports = map(int, [val for val in fuser_cmd.stdout[:-1] if val != 'Cannot stat file '])
+        print(ss_cmd)
+        used_ports = map(int, [val for val in ss_cmd.stdout[:-1] if val != 'Cannot stat file '])
 
     except ValueError:
         raise CapsuleTunnelError(
-            f'Failed parsing the port numbers from stdout: {fuser_cmd.stdout[:-1]}'
+            f'Failed parsing the port numbers from stdout: {ss_cmd.stdout[:-1]}'
         )
     try:
         # take the list of available ports and return randomly selected one
--- a/tests/robottelo/test_helpers.py
+++ b/tests/robottelo/test_helpers.py
@@ -4,6 +4,7 @@ from unittest import mock
 import pytest
 
 from robottelo.helpers import escape_search
+from robottelo.helpers import get_available_capsule_port
 from robottelo.helpers import get_host_info
 from robottelo.helpers import get_server_version
 from robottelo.helpers import HostInfoError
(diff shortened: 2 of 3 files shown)


tests/robottelo/test_helpers.py
```diff
@@ -118,3 +119,15 @@ def test_slugify_component():
...
+
+
+class TestGetAvailableCapsulePort:
+    """Tests for method ``get_available_capsule_port``."""
+
+    @mock.patch('robottelo.helpers.ssh')
+    def test_return_port(self, ssh):
+        """get_available_capsule_port returns a port number.
+        """
+        ssh.command = mock.MagicMock(return_value=FakeSSHResult(['""'], 0))
```
jameerpathan111, line 131: Can someone explain what's happening here? and how it's relevant to `get_available_capsule_port`?

robottelo/helpers.py
```diff
@@ -359,23 +359,22 @@ def get_available_capsule_port(port_pool=None):
                 )
             )
     # returns a list of strings
-    fuser_cmd = ssh.command(
-        'fuser -n tcp {{{0}..{1}}} 2>&1 | awk -F/ \'{{print$1}}\''.format(
-            port_pool[0], port_pool[-1]
-        )
+    ss_cmd = ssh.command(
+        f"ss -tnaH sport ge {port_pool[0]} sport le {port_pool[-1]}"
+        " | awk '{print $4}' | awk -F ':' '{print $NF}' | sort | uniq"
```
mirekdlugosz, line 364: Every time I see command output piped to same command, I wonder if there is a way to do everything on single command invocation. Usually there is.
```
" | awk '{n=split($4, p, ":"); print p[n]}' | ...
```

Also, if you don't pass any flag to `uniq`, you can tell `sort` to skip duplicates:
```
| sort -u
```
That flag has been there since at least 1980s, if compatibility would be your concern.

mirekdlugosz (approved):
Good fix, and thanks for adding unit test!
ACK

Technically, the problem is not "Satellite stopped shipping fuser", as it never did. It was being added by `satellite6-installer` Jenkins job. SatLab team uses different installation procedure and they never added `fuser` explicitly. But otherwise you are correct, `fuser` might not be present on Satellite we use for testing, and that will cause tests to fail.
Overall, binaries from base installation should always be preferred.

tests/robottelo/test_helpers.py
mirekdlugosz: `ssh.command` function is replaced with mock, that will always return object specified in `return_value` when original function is called. This way `get_available_capsule_port` will not try to connect to SSH server, something that must fail on travis and is just too costly in regular unit test run.

If you are looking for general discussion about mocks (what they are, what purpose they serve, what are their downsides), you can give following two articles a try:
* [How to mock in Python? – (almost) definitive guide](https://breadcrumbscollector.tech/how-to-mock-in-python-almost-definitive-guide/)
* [Beware of chicken testing! (or mocks overuse)](https://breadcrumbscollector.tech/beware-of-chicken-testing-or-mocks-overuse/)

If you are interested in `MagickMock` class specifically, you can read [`unittest.mock`](https://docs.python.org/3/library/unittest.mock.html) documentation.

robottelo/helpers.py
tpapaioa: Updated the command to use `awk` only once, and to use `sort -u` instead of `sort | uniq`. Thanks!

Re-ran tests, and the test output is the same as above.

robottelo/helpers.py
```diff
@@ -359,23 +359,22 @@ def get_available_capsule_port(port_pool=None):
...
     try:
-        print(fuser_cmd)
-        used_ports = map(int, [val for val in fuser_cmd.stdout[:-1] if val != 'Cannot stat file '])
+        print(ss_cmd)
+        used_ports = map(int, [val for val in ss_cmd.stdout[:-1] if val != 'Cannot stat file '])
 
     except ValueError:
         raise CapsuleTunnelError(
-            'Failed parsing the port numbers from stdout: {0}'.format(fuser_cmd.stdout[:-1])
+            'Failed parsing the port numbers from stdout: {0}'.format(ss_cmd.stdout[:-1])
```
JacobCallahan, line 377: can you make these f-strings?

JacobCallahan (changes requested):

robottelo/helpers.py
tpapaioa: Done.

JacobCallahan (approved):
ACK, thanks!

PR #8072, opened 2020-10-20
Added test case for custom repo download policies

devendra104:
The purpose of this test case is to validate the default download policy of the custom repository, The default download policies value of the new creating repository gets changed according to the set value in the settings.

In this test case, we are validating both available policies.

- [ ] Immediate

- [ ] On Demand

Test Results:

```
$ pytest tests/foreman/ui/test_settings.py::test_positive_custom_repo_download_policy
platform linux -- Python 3.6.6, pytest-4.6.3, py-1.8.1, pluggy-0.13.1                                                                                                                                                                           

================================================================================================ 
===================================================================================== 1 passed in 299.02 seconds ======================================================================================
```

### DIFF
--- a/tests/foreman/api/test_setting.py
+++ b/tests/foreman/api/test_setting.py
@@ -201,3 +201,45 @@ class SettingTestCase(APITestCase):
 
         :CaseAutomation: notautomated
         """
+
+    @run_in_one_thread
+    @tier2
+    def test_positive_custom_repo_download_policy(self):
+        """ Check the set custom repository download policy for newly created custom repository.
+
+        :id: d5150cce-ba85-4ea0-a8d1-6a54d0d29571
+
+        :Steps:
+            1. Create a product, Organization
+            2. Update the Default Custom Repository download policy in the setting.
+            3. Create a custom repo under the created organization.
+            4. Check the set policy of new created repository.
+            5. Repeat steps 2 to 4 for both download policy.
+
+        :expectedresults: The set download policy should be the default policy in the newly created
+         repository.
+
+        :CaseImportance: Medium
+
+        :CaseLevel: Acceptance
+        """
+        download_policy_id = "default_download_policy"
+        download_policies = ["immediate", "on_demand"]
+        download_policy_property = entities.Setting().search(
+            query={'search': f'name={download_policy_id}'}
+        )[0]
+        default_property_value = download_policy_property.value
+        org = entities.Organization().create()
+        prod = entities.Product(organization=org).create()
+        try:
+            for download_policy in download_policies:
+                download_policy_property.value = download_policy
+                download_policy_property.update({'value'})
+                repo = entities.Repository(
+                    product=prod, content_type='yum', organization=org
+                ).create()
+                assert repo.download_policy == download_policy
+                repo.delete()
+        finally:
+            prod.delete()
+            setting_cleanup(setting_name=download_policy_id, setting_value=default_property_value)


tests/foreman/ui/test_settings.py
```diff
@@ -532,3 +532,52 @@ def test_positive_entries_per_page(session):
             assert str(total_pages) == page_content["Pagination"]["pages"]
         finally:
             setting_cleanup(setting_name=property_name, setting_value=default_property_value.value)
+
+
+@run_in_one_thread
+@tier2
```
jameerpathan111, line 538: should we parameterize this kind of test?
`@pytest.mark.parametrize('download_policies', **parametrized(["Immediate", "On Demand"]))`
But this could add over head to test like creating org, product, deleting product and restoring setting.
No idea if it's worth doing in this case, still wanted to put it here.
Devendra, if you choose to parameterize it, don't forget to add `:parametrized: yes` docstring token. :)

mirekdlugosz (commented):
Is UI a right level for that? In general, we try to limit number of UI tests and try to make them as broad as possible.

Looks like there is nothing UI-specific in this test, and it basically boils down to couple API calls. It would be 5-10 times faster on API/CLI level, and would provide exactly the same coverage.

devendra104:
> Is UI a right level for that? 

I think yes, the intention behind adding this test case in UI is to verify the policy-related changes from UI and the same with their validation(newly created repository should contain the same changes)

>In general, we try to limit number of UI tests and try to make them as broad as possible.

Yes, you are right.

> Looks like there is nothing UI-specific in this test, and it basically boils down to couple API calls. It would be 5-10 times faster on API/CLI level, and would provide exactly the same coverage.

As I say, The purpose is only to tests whether the changes are work from UI or not, I completely agree with you CLI and API will provide the code coverage and faster execution but it will not give the assurance of this feature from UI. 

@mirekdlugosz Anyway, I haven't seen any recent issue related to this, so I think I can move this test from UI to CLI/API, what you say?

mirekdlugosz:
> @mirekdlugosz Anyway, I haven't seen any recent issue related to this, so I think I can move this test from UI to CLI/API, what you say?

I think it shouldn't be in UI unless there is good reason for it to be in UI. Something like "this workflow works when using CLI, but is broken on UI". So I think this test should be moved to CLI/API.

As for which one - earlier this year there was thread with subject "To which interface we give more priority/importance while adding test coverage for a feature?" on satqe-list. People gave various arguments where you should put your tests first, API or CLI. There's no consensus or guideline, up to you to decide what you find more compelling and why.

tests/foreman/ui/test_settings.py
devendra104: @jameerpathan111 Thanks for your input,  For now, I shifted this test case from UI to API, As you know the settings API test module is not a pytest, I am working on the conversion process and there I will definitely include your suggestion.

devendra104:
> > @mirekdlugosz Anyway, I haven't seen any recent issue related to this, so I think I can move this test from UI to CLI/API, what you say?
> 
> I think it shouldn't be in UI unless there is good reason for it to be in UI. Something like "this workflow works when using CLI, but is broken on UI". So I think this test should be moved to CLI/API.
> 

Your statement is valid here because we have other alternatives(API/CLI) to achieve this and no customer reported defect against it.

devendra104:
Test Results:

```
#pytest tests/foreman/api/test_setting.py::SettingTestCase::test_positive_custom_repo_download_policy
platform linux -- Python 3.6.6, pytest-4.6.3, py-1.8.1, pluggy-0.13.1

tests/foreman/api/test_setting.py .                                                                                                                                                                         [100%]

====================================================================================== 1 passed in 56.46 seconds ======================================================================================

```

jameerpathan111:
There are two tests which try to verify if repository is using default_download_policy. But they don't validate both policies. 
Should we remove/update those tests now that we have this one?
`tests/foreman/api/test_repository.py::test_positive_create_with_default_download_policy`
`tests/foreman/cli/test_repository.py::test_positive_create_with_default_download_policy`

jameerpathan111:
Also in settings description I see we have  'inherit', immediate', 'on_demand', or 'background' options available. 
Do we need to check rest of them?

devendra104:
> There are two tests which try to verify if repository is using default_download_policy. But they don't validate both policies.
> Should we remove/update those tests now that we have this one?
> `tests/foreman/api/test_repository.py::test_positive_create_with_default_download_policy`
> `tests/foreman/cli/test_repository.py::test_positive_create_with_default_download_policy`

I aware of these data and there we are only validating the available settings for the custom created repo, but here we change the policy in the settings and check whether the changes are reflected properly or not in the custom created repository.

devendra104:
> Also in settings description I see we have 'inherit', immediate', 'on_demand', or 'background' options available.
> Do we need to check rest of them?

There are three methods in which 1 is deprecated that's why I didn't include that one.

-  On-Demand
-  Immediate
- Background (depricated)

mirekdlugosz (approved):
Thanks for moving this to API. This way we reduced execution time by 80%.
ACK

jameerpathan111 (approved):
ACK

PR #8074, opened 2020-10-20
Fix parametrized tokens for polarion

mshriver:
Missing token results in mismatched polarion result uploading.

### DIFF
--- a/tests/foreman/api/test_organization.py
+++ b/tests/foreman/api/test_organization.py
@@ -92,6 +92,8 @@ class TestOrganization:
             auto-generated label.
 
         :CaseImportance: Critical
+
+        :parametrized: yes
         """
         org = entities.Organization(name=name, description=name).create()
         assert org.name == name
@@ -111,6 +113,7 @@ class TestOrganization:
 
         :expectedresults: The organization cannot be created.
 
+        :parametrized: yes
         """
         with pytest.raises(HTTPError):
             entities.Organization(name=name).create()
@@ -202,6 +205,8 @@ class TestOrganizationUpdate:
         :expectedresults: The organization's name is updated.
 
         :CaseImportance: Critical
+
+        :parametrized: yes
         """
         setattr(module_org, 'name', name)
         module_org = module_org.update(['name'])
@@ -217,6 +222,8 @@ class TestOrganizationUpdate:
         :expectedresults: The organization's description is updated.
 
         :CaseImportance: Critical
+
+        :parametrized: yes
         """
         setattr(module_org, 'description', desc)
         module_org = module_org.update(['description'])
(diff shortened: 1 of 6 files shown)


tests/foreman/cli/test_role.py
```diff
@@ -188,7 +188,7 @@ def make_role_with_permissions(self):
 
     @tier1
     @upgrade
-    @pytest.mark.parametrize('per_page', **parametrized([1, 5, 20]))
```
mshriver, line 191: this method isn't doing anything useful for this parameter. It just results in ID's of 0, 1, 2 instead of 1, 5, 20, which are pretty meaningless.

tests/foreman/api/test_organization.py
```diff
@@ -217,6 +222,8 @@ def test_positive_update_description(self, module_org, desc):
         :expectedresults: The organization's description is updated.
 
         :CaseImportance: Critical
+
+        :parametrized: yes
         """
         setattr(module_org, 'description', desc)
         module_org = module_org.update(['description'])
```
latran, line 229: There's a parametrization on line 326 in the file that should include a docstring that's missing.

mshriver: great catch, thanks! Added.

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -122,6 +122,8 @@ def test_positive_run_default_job_template_by_ip(self, fixture_vmsetup, fixture_
             and task can be listed by name and ID
 
         :BZ: 1647582
+
+        :parametrized: yes
```
omkarkhatavkar, line 126: Why we are adding `:parametrized: yes` here in all this file. I dont see markup has been used in the test case level

omkarkhatavkar (approved):
ACK pending some comments

tests/foreman/cli/test_remoteexecution.py
mshriver: All of the tests using the `fixture_vmsetup` fixture are parametrized, because the fixture is parametrized.

mirekdlugosz: betelgeuse is supposed to read metadata from module and class before reading function. So if there is some metadata value common for everything in class or file, there is no reason to duplicate it for each individual test.

mshriver: Its not required on all tests.

It has been added to the tests using the `fixture_vmsetup` fixture, which itself is parametrized. Putting it at the module would apply it to test cases that are not parametrized.

mirekdlugosz: > It has been added to the tests using the `fixture_vmsetup` fixture

Which are all but one test inside `TestRemoteExecution` class. If that test could be moved outside of class, `parametrized` metadata could be moved to class level.

Although I never considered metadata duplication blocking. That was just FYI.

> Putting it at the module would apply it to test cases that are not parametrized.

That is correct.

mirekdlugosz (approved):

PR #7924, opened 2020-08-07
Introducing broker as a VM manager

JacobCallahan:
Added basic fixture examples
Added a new ContentHost class to subclass broker's Host class
Converted test_positive_candlepin_events_processed_by_STOMP to use a broker host
Added a setting to robottelo.properties and accomanying config
Added the broker dependency to requirements.txt

Issue #3611, opened 2016-07-18
Error handling for negative tests in hammer


sghai:
This is the issue specifically for negative scenarios. Currently, we are checking the returncode however, we should actually check the expected error message too. This is required because in hammer if we pass `incorrect_id` of an entity as an option then some cmd fails like:

``` bash
]# hammer  -u admin -p changeme hostgroup create --name text --subnet-id 12345
Could not create the hostgroup:
  ERROR:  insert or update on table "hostgroups" violates foreign key constraint "hostgroups_subnet_id_fk"
  DETAIL:  Key (subnet_id)=(12345) is not present in table "subnets".

```

Since we are just validating `returncode` which is not sufficient like below where we are just checking returncode, because never know if cmd fails with expected error message or some db error

``` python
def test_negative_delete_by_id(self):
        """Create HostGroup then delete it by wrong ID

        @id: 047c9f1a-4dd6-4fdc-b7ed-37cc725c68d3

        @assert: HostGroup is not deleted
        """
        for entity_id in invalid_id_list():
            with self.subTest(entity_id):
                with self.assertRaises(CLIReturnCodeError):
                    HostGroup.delete({'id': entity_id})

```

This can be done like below:

``` python
        subnet_id = gen_string('numeric', 4)
        with self.assertRaises(CLIReturnCodeError) as exception:
            HostGroup.create({
                'name': gen_string('alpha'),
                'subnet-id': subnet_id
            })
        self.assertIs(
            exception.exception.stderr,
            'Could not find subnet {0}'.format(subnet_id)

```

renzon:
Once it impacts almost all files from cli, I am gonna create issue for each file so PR are smaller and easier to review

rplevka:
@renzon - just keep this issue as a tracker, create a list of files with checkboxes so we can keep track of those you've already addressed. e.g.:
- [ ] test_organization.py
- [ ] test_user.py

elyezer:
One thing to have in mind is if we start to do localization testing this will all break since the messages are going to be translated. If localization tests is something to have in the future then this should be considered before starting working on this issue.

pondrejk:
Adding a tracking checklist (for lack of a better place):

- [ ] test_abrt.py -- no negative cases
- [ ] test_activationkey.py -- #3716
- [ ] test_architecture.py -- #3747 
- [ ] test_bootstrap_script.py
- [ ] test_capsule.py
- [ ] test_capsule_installer.py
[...]

PR #8075, opened 2020-10-21
Convert cli syncplan unittest to pytest

swadeley:
Hello

Convert cli syncplan unittest to pytest

Thank you

### DIFF
--- a/tests/foreman/cli/test_syncplan.py
+++ b/tests/foreman/cli/test_syncplan.py
@@ -15,6 +15,7 @@
 
 :Upstream: No
 """
+import logging
 from datetime import datetime
 from datetime import timedelta
 from time import sleep
@@ -41,6 +42,7 @@ from robottelo.constants import REPOS
 from robottelo.constants import REPOSET
 from robottelo.datafactory import filtered_datapoint
 from robottelo.datafactory import invalid_values_list
+from robottelo.datafactory import parametrized
 from robottelo.datafactory import valid_data_list
 from robottelo.decorators import run_in_one_thread
 from robottelo.decorators import tier1
@@ -49,7 +51,6 @@ from robottelo.decorators import tier3
 from robottelo.decorators import tier4
 from robottelo.decorators import upgrade
 from robottelo.ssh import upload_file
-from robottelo.test import CLITestCase
 
 
 @filtered_datapoint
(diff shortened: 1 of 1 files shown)


swadeley:
Hello

there is a text reference to this old issue:
https://github.com/SatelliteQE/robottelo/issues/3611

I will remove that as there are no more `assert_error_msg` checks in this file.

tests/foreman/cli/test_syncplan.py
```diff
@@ -155,24 +154,25 @@ def validate_repo_content(self, repo, content_types, after_sync=True):
...
         :expectedresults: Sync plan is created and has random name
 
         :CaseImportance: Critical
         """
-        for name in valid_data_list():
-            with self.subTest(name):
-                new_sync_plan = self._make_sync_plan({'name': name})
-                self.assertEqual(new_sync_plan['name'], name)
+        sync_plan = self._make_sync_plan({'name': name})
+        assert sync_plan['name'] == name
```
mshriver, line 175: Is this actually asserting against the name that was set on the entity within Satellite, or is it asserting against the _object_ returned by this function?  Its an important distinction here, and the test should be fetching the entity after it is created to strongly assert that the attribute was correct on Satellite itself.

tests/foreman/cli/test_syncplan.py
```diff
@@ -187,7 +187,7 @@ def test_positive_create_with_description(self):
         for desc in valid_data_list():
             with self.subTest(desc):
```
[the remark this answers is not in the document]
swadeley: Hi @mshriver , yes, I needed to get the class and definitions in the class working (ported to fixtures as @mirekdlugosz pointed ouit) so I could test the changes I was making to the subTest structures as I went along. I will go through the whole test again and review it. Thank you.

tests/foreman/cli/test_syncplan.py
swadeley: Will check

swadeley: @mshriver I fixed the first three test by adding `result = SyncPlan.info` to read back the Sync plan before making an assert.

mshriver: Great! That's a much stronger assertion pattern for this test. We don't want to just be testing our entity class constructor :)

swadeley:
Hello

I had to change LOGGER back to `logger`

    -    LOGGER = logging.getLogger('robottelo')
   +    logger = logging.getLogger('robottelo')

robottelo]$ pytest --show-capture=no tests/foreman/cli/test_syncplan.py
================================= test session starts =========================================
platform linux -- Python 3.7.9, pytest-4.6.3, py-1.9.0, pluggy-0.13.1
rootdir: /sat6/robottelo_1/robottelo
plugins: xdist-2.0.0, services-1.3.1, mock-1.10.4, forked-1.3.0, cov-2.10.1
collecting 86 items                                                                                              
2020-10-23 12:25:11 - conftest - DEBUG - Collected 86 test cases
collected 86 items                                                                                               

tests/foreman/cli/test_syncplan.py ....................................................................... [ 82%]
.......s.......                                                                                            [100%]

===================== 85 passed, 1 skipped, 3 warnings in 2768.21 seconds =========================

pondrejk (approved):

tests/foreman/cli/test_syncplan.py
```diff
@@ -102,17 +103,18 @@ def valid_name_interval_update_tests():
...
+@pytest.mark.usefixtures("setup_org_for_syncplan")
+class TestSyncPlan:
     """Sync Plan CLI tests."""
 
     org = None
+    logger = logging.getLogger('robottelo')
 
-    def setUp(self):
+    @pytest.fixture(scope='class')
+    def setup_org_for_syncplan(self):
```
mirekdlugosz, line 114: I feel we are stopping step too early here.

It's better for fixture to return object instead of modifying specific class definition. Modifying class is available and often necessary step during unittest -> pytest, but here I don't see the need for it.

`cached=True` parameter seems to put object in cache, so next time same factory function is called, it will return already existing object. With module/class scoped fixture, is there really any need for this?

I suggest to re-use existing `module_org` fixture. Injecting it to all tests with `pytest.mark.usefixtures` is good idea.

mirekdlugosz (approved):
ACK with suggestion. Let me know if you plan to work on this, or prefer to merge it as-is. Or you can merge it yourself, this is second ACK.

tests/foreman/cli/test_syncplan.py
swadeley: OK, I will try that, thanks for the review

swadeley:
Hello

I removed the class and the org setup function, made changes to use `module_org` fixture.

robottelo]$ pytest --show-capture=no tests/foreman/cli/test_syncplan.py 
=================================== test session starts ======================================
platform linux -- Python 3.7.9, pytest-4.6.3, py-1.9.0, pluggy-0.13.1
rootdir: /home/sjw/sat6/robottelo_1/robottelo
plugins: xdist-2.0.0, services-1.3.1, mock-1.10.4, forked-1.3.0, cov-2.10.1
collecting 86 items
2020-10-30 09:54:10 - conftest - DEBUG - Collected 86 test cases
collected 86 items

tests/foreman/cli/test_syncplan.py ....................................................................... [ 82%]
.......s.......                                                                                            [100%]

====================== 85 passed, 1 skipped, 3 warnings in 2740.11 seconds ==================

tests/foreman/cli/test_syncplan.py
swadeley: @mirekdlugosz , changes made, please review once more.

I added `module_org` to list of arguments, replacing `self` rather than use `pytest.mark.usefixtures` above every test.

Thank you

mirekdlugosz (approved):
> platform linux -- Python 3.7.9, pytest-4.6.3, py-1.9.0, pluggy-0.13.1

Since #7924 Robottelo depends on broker, which is exclusively Python 3.8+. Since #8031 we also use pytest 5.x. You might want to take a moment to remove your existing virtualenv and rebuild it using Python 3.8, to avoid potential trouble down the road.

If your OS doesn't provide Python 3.8 or newer, take a look at [pyenv](https://github.com/pyenv/pyenv).

mirekdlugosz:
@swadeley also, we branched off 6.8.z earlier this week. If you want these changes to apply to automation for 6.8 z-stream releases, please submit new PR with cherry-pick.

swadeley:
> > platform linux -- Python 3.7.9, pytest-4.6.3, py-1.9.0, pluggy-0.13.1
> 
> Since #7924 Robottelo depends on broker, which is exclusively Python 3.8+. Since #8031 we also use pytest 5.x. You might want to take a moment to remove your existing virtualenv and rebuild it using Python 3.8, to avoid potential trouble down the road.
> 
> If your OS doesn't provide Python 3.8 or newer, take a look at [pyenv](https://github.com/pyenv/pyenv).

Thanks for the tip, I see updating to Fed32 is another option
https://fedoraproject.org/wiki/Releases/32/ChangeSet#Python_3.8

swadeley:
Thank you to all reviewers

PR #8082, opened 2020-10-22
Fix automation failure where newly added organizations appeared for existing report templates. + New asserts.

lhellebr:


omkarkhatavkar:
@lhellebr can you please add test result and details about PR why this was required?

lhellebr:
```
$ pytest tests/foreman/ui/test_reporttemplates.py::test_positive_end_to_end
[...]
====================================== 1 passed, 14 warnings in 246.34 seconds ======================================
```

This was required because tests run in parallel and when a report template is marked as Default, newly added organizations are added to it. It fixes failures of this test in automation.

tests/foreman/ui/test_reporttemplates.py
```diff
@@ -210,7 +210,18 @@ def test_positive_end_to_end(session, module_org, module_loc):
         session.reporttemplate.update(name, {'template.name': new_name, 'type.snippet': True})
         rt = session.reporttemplate.read(new_name)
         assert rt['template']['name'] == new_name
+        assert rt['template']['default'] is False
+        assert rt['template']['template_editor']['editor'] == content
+        assert rt['inputs'][0]['name'] == template_input[0]['name']
```
omkarkhatavkar, line 215: I did not understand why template_input was kept in list with single object. may be adding as dict would have been simpler ? non-blocking comment and not regarding this PR though. But Just a Thought.

omkarkhatavkar (approved):
ACK pending some comments

tests/foreman/ui/test_reporttemplates.py
mirekdlugosz: Because Airgun uses `RemovableWidgetsItemsListView` there. Haven't looked at code, but judging by name it expects list of items to fill. So you would have to add more complexity to Airgun, so it can handle dict passed in directly. Or you can just wrap your dict in list on test side.

mirekdlugosz (approved):
LGTM

PR #8085, opened 2020-10-22
sles is no longer supported

tstrych:
there will be no test in automation in future 
removing the stubbed case


mirekdlugosz (approved):
ACK

pondrejk (approved):

PR #8089, opened 2020-10-26
Add report to default org before using it. Some tests remove it. Fixes automation.

lhellebr:


lhellebr:
Travis failure seems unrelated

tests/foreman/cli/test_reporttemplates.py
```diff
@@ -659,6 +659,14 @@ def test_positive_generate_with_name_and_org(self):
         host_name = gen_string('alpha')
         host = make_fake_host({'name': host_name})
 
+        # make sure the template is in the default org
+        org_names = ReportTemplate.info({'name': 'Host - Statuses'})['organizations']
+        orgs = [Org.info({'name': org_name})['id'] for org_name in org_names]
+        default_org_id = Org.info({'name': DEFAULT_ORG})['id']
+        if default_org_id not in orgs:
+            orgs.append(default_org_id)
+        ReportTemplate.update({'name': 'Host - Statuses', 'organization-ids': orgs})
```
mirekdlugosz, line 668: That probably isn't going to have very visible impact on performance, but I feel we are doing unnecessary things here.

First, `update()` call should be indented, so it's only called when default org is not in template org. Right now we might update template with exactly same value, assuming template is in default org.

Second, shouldn't we just check for default_org_name in org_names, and skip all invocations of `Org.info()` in scenario when template already is in default org? If template is assigned to many organizations, but not default org, this will save us a lot of `hammer` calls.

```python
        org_names = ReportTemplate.info({'name': 'Host - Statuses'})['organizations']
        default_org = Org.info({'name': DEFAULT_ORG})
        if default_org['name'] not in org_names:
            org_ids = [Org.info({'name': org_name})['id'] for org_name in org_names]
            org_ids.append(default_org['id'])
            ReportTemplate.update({'name': 'Host - Statuses', 'organization-ids': org_ids})        
```

I think that's also a little more understandable, as you have one less variable to keep track of before getting to conditional.

mirekdlugosz (commented):
Please add test result after this change.

tests/foreman/cli/test_reporttemplates.py
lhellebr: Right! I'll fix that.

lhellebr:
```
$ pytest tests/foreman/cli/test_reporttemplates.py::ReportTemplateTestCase::test_positive_generate_with_name_and_org
[...]
=================================================================================================== 1 passed, 2 warnings in 187.05 seconds ===================================================================================================
```

omkarkhatavkar (approved):
ACK

mirekdlugosz (approved):

mirekdlugosz:
@lhellebr I forgot to mention, we branched off 6.8.z earlier this week, so that change will apply only to 6.9 automation. If you want it also in 6.8 z-stream releases, please submit new PR with cherry-pick.

PR #8092, opened 2020-10-27
Move all decorator marks to direct pytest marks

mshriver:
I don't believe there is value in defining these names, for individual import and decoration, when nearly every test module will need to import pytest regardless, and the marks can be accessed directly.

Marks are defined in `pytest_plugins.markers`, and we can include any contextual comments about the marks there.  There is no need to doubly define what marks are registered for use by test modules.

We can run the standalone automation job against this if there is concern.

This includes a few other items that I caught along the way:

* Convert some unit test cases that had no setup methods in use.
* Includes import of pytest on many modules
* Various test cleanup concerning inconsistent use of existing markers
* Remove file encoding comments, utf-8 is the default in py3

I'm going to do my best to keep this up-to-date with master, but as people merge pytest conversion PRs it makes for some large rebasing efforts. Please still consider the overall changes for review, and I'll make sure there are no conflicts and travis is passing cleanly before merge.

### DIFF
--- a/robottelo/decorators/__init__.py
+++ b/robottelo/decorators/__init__.py
@@ -2,34 +2,12 @@
 import logging
 from functools import wraps
 
-import pytest
 import unittest2
 
 from robottelo.config import settings
 
 LOGGER = logging.getLogger(__name__)
 OBJECT_CACHE = {}
-
-# Test Tier Decorators
-# CRUD tests
-tier1 = pytest.mark.tier1
-# Association tests
-tier2 = pytest.mark.tier2
-# Systems integration tests
-tier3 = pytest.mark.tier3
-# Long running tests
-tier4 = pytest.mark.tier4
-# Destructive tests
-destructive = pytest.mark.destructive
-# Upgrade
-upgrade = pytest.mark.upgrade
-# Tests to be executed in 1 thread
-run_in_one_thread = pytest.mark.run_in_one_thread
-
-# Shortcuts for pytest methods
-parametrize = pytest.mark.parametrize
-fixture = pytest.fixture
-skipif = pytest.mark.skipif
 
 
 def setting_is_set(option):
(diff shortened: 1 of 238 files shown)


robottelo/decorators/__init__.py
```diff
@@ -2,35 +2,13 @@
...
 
 from robottelo.config import settings
 
 LOGGER = logging.getLogger(__name__)
 OBJECT_CACHE = {}
 
-# Test Tier Decorators
-# CRUD tests
-tier1 = pytest.mark.tier1
-# Association tests
```
jameerpathan111, line 16: @mshriver we should also update `docs/features/decorators.rst` to reflect this change.
And do we need to change anything in robottelo-ci because of this?

jyejare (commented):
Since there changes with the pytest markers specially the Tier marks. I would love to see the Betelgeuse rightly uploads the test cases where Tier parameter for Test on polarion is set.

You might want to use stage jenkins for it.

jyejare:
@mshriver I am also really existed to know how do you this bug refactor of 313 files. Any special tool ?

robottelo/decorators/__init__.py
mshriver: There should be any changes or impact to any tools using these, as functionally they are identical

mshriver:
@jyejare Just using find+replace tools in your IDE of choice. Its all pretty consistent, so it didn't take very long at all.

mshriver:
@jyejare I have no concern about external tool integration with this, as the marks are identical. All that decorators was doing was naming the mark - which when we weren't using pytest, might have provided some small benefit to imports. Since we're using one line per module imports now, there's really no benefit to having this additional layer in the markers.  All the test modules will be importing pytest, so I think its far better to be explicit and direct in our pytest markers.  `@tierX` becomes ambiguous when you have to go look at the import, and another module to see whether `tierX` marker is the regular marker, or if some additional behavior is wrapped into it. In this case there is no additional behaviors to these marks. You see an explicit pytest mark decorator on the test case, and it should be easy to find that mark definition using `pytest --markers`, or looking at `robottelo/pytest_plugins/markers.py` to see what that mark does, if anything.

jyejare:
@mshriver Yes ! I have no doubt behind the intention of this change. I have very much welcome the change by moving this to one liners. But my only concern is the dependent tool which does the test cases uploading to RP and Polarion shouldn't be broken :)

With Betelgeuse, I think we are already having some concerns over there in uploading the iterative test cases with very difficult traces to find out the reason. So just wanted to confirm if this style works or there any hidden conf that needs to be changed !

robottelo/decorators/__init__.py
jameerpathan111: ok. 
and I can raise a pr to update `docs/features/decorators.rst` later.

jameerpathan111 (approved):
LGTM

jyejare:
@mshriver Merging the PR with changes by keeping an eye on Polarion :)

PR #8094, opened 2020-10-27
Remove test_positive_reinstall_on_same_node_after_remove

vsedmik:
According to the customer portal [1] the katello-remove utility has been deprecated in 6.7 and the recommended way of server reinstall is to re-provision the operating system and install it again. As the katello-remove is no longer available in capsule too, I assume the same applies to the capsule server.

This being said, I suggest to retire the test, as it verifies feature which is no longer supported and takes over 35 minutes to complete. The installer functionality remains covered in robottelo/vm_capsule.py [2] and remains used by several tests.

[1] https://access.redhat.com/solutions/5048661
[2] https://github.com/SatelliteQE/robottelo/blob/ad6daf6cdc26790274a79982a6e35f08206e9e65/robottelo/vm_capsule.py#L222

### DIFF
--- a/tests/foreman/cli/test_capsule_installer.py
+++ b/tests/foreman/cli/test_capsule_installer.py
@@ -15,22 +15,9 @@
 
 :Upstream: No
 """
-import os
-from tempfile import mkstemp
-
 import pytest
 
-from robottelo import ssh
-from robottelo.cli.base import CLIReturnCodeError
-from robottelo.cli.capsule import Capsule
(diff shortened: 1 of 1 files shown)


pondrejk (approved):

mirekdlugosz (approved):
ACK

6.8.z branch was created earlier today. If you want to retire this test from 6.8 z-stream releases as well, please submit new PR with cherry-pick.

PR #8096, opened 2020-10-27
no cleanup on fake capsules

pondrejk:
The idea is to clean up only if it actually frees up resources

### DIFF
--- a/tests/foreman/cli/test_realm.py
+++ b/tests/foreman/cli/test_realm.py
@@ -20,7 +20,6 @@ import random
 import pytest
 from fauxfactory import gen_string
 
-from robottelo.cleanup import capsule_cleanup
 from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.factory import CLIFactoryError
 from robottelo.cli.factory import make_proxy
@@ -36,7 +35,6 @@ def _make_proxy(options=None):
     """Create a Proxy and register the cleanup function"""
     proxy = make_proxy(options=options)
     yield proxy
-    capsule_cleanup(proxy['id'])
 
 
 @run_in_one_thread


omkarkhatavkar (approved):
ACK, Travis issue is untreated this PR

mirekdlugosz (commented):
Could you elaborate on this change a little? Why call cleanup only if options were not passed? What is the relationship between options and "actually freeing up resources"? I admit I don't really understand what you are trying to achieve and how this change achieves that.

How do you intend to pass `options` to fixture? It's fixture, and we usually pass fixture references as test arguments, we don't call them directly.

Also, demonstration that this change may work is missing. I guess we would need test results for test that passes `options`, and one for test that does not.

mshriver:
Moving to draft due to travis failures and no response to @mirekdlugosz comment above about the behavior of this function at its core.

These methods should ultimately get refactored as pytest fixtures, and options would have to be indirectly parametrized.  That aside, you're getting at the core create/delete behavior that our fixtures should adhere to, which is great!

In the past my team has used an `entity.delete_if_exists` method to cover similar situations, where we wanted to clean up test resources 'safely' during teardown, without failure if the entity was already deleted by the test.

I think you should turn your contribution for this specific test module into a larger effort across the framework, and we create a consistent pattern for cleaning up test resources. This should come after the pytest conversions are complete, so we're not dealing with a mix of unittest and pytest setup and teardown hooks.

pondrejk:
@mirekdlugosz  @mshriver sorry for delay.
Thinking about this again I just want to get rid of the cleanup in this module as it causes intermittent timeouts in tests and doesn't bring much value. I'm OK with discussing an over-reaching cleanup policy, but for now, would you agree with just removing it here?

mirekdlugosz:
> Thinking about this again I just want to get rid of the cleanup in this module as it causes intermittent timeouts in tests and doesn't bring much value.

Sounds fair to me.

But first please remove import on line 23. Now imported function is never used, and that causes flake8 to complain, which in turn causes CI to fail.

mirekdlugosz (approved):
ACK
Merging, as CI passes now.

PR #8100, opened 2020-10-28
Pytest conversion for tests/foreman/cli/test_repository.py

tpapaioa:
This PR converts `tests/foreman/cli/test_repository.py` to pytest. The main changes are:

1.) Add `repo_cleanup`, `product_cleanup`, `gpg_key_cleanup` methods to `robottelo.cleanup` for use in fixture finalizers.

2.) Move the specific repo parameters for the tests into a module-level dictionary, `REPO_OPTIONS`. Each individual dictionary entry is a list of one or more dictionaries which contain the actual repository settings that will be used by the repo fixture. The corresponding `REPO_PARAMS` dictionary contains the same information, but each value of `REPO_OPTIONS` is passed through the `parametrized()` helper method to return the parameter values and ids in a form that can be passed to `pytest.mark.parametrize()`. A similar, but much smaller, pair of dictionaries, `REPO_CRED_OPTIONS` and `REPO_CRED_PARAMS`, has also been added, for use with tests that create authenticated repos. If the test also needs to use the credential record separately, then `REPO_CRED_PARAMS` passes a tuple containing the repo options and the credential record.

3.) Rename the test classes from `*TestCase` to `Test*`, and remove their inheritance from unittest classes.

4.) Remove `setUpClass` and other org, product, and repo setup methods from the classes.

5.) Create `org`, `repo`, `gpg_key` fixtures. Each of these calls a helper method that creates the corresponding record and also adds a finalizer to delete the record when done.

6.) Add a `repo_options` fixture that will return the repo options when `pytest.mark.parametrize()` parametrizes it via the `indirect=True` keyword arg. The `repo` fixture can then use the `repo_options` fixture to perform repo creation, and the test case can also use `repo_options` if it needs to read these options directly.

7.) All tests that previously created repos within the test are now parametrized to use the `repo_options` and `repo` fixtures. Some tests that expect repo creation to fail, or that otherwise need to perform org, product, repo, etc. creation within the test, now use the helper methods `_org()`, `_product()`, `_repository()` etc., to ensure that record deletion occurs during teardown.

8.) All unittest-style assert methods are changed to regular `assert` calls.

Test results:

```
# pytest tests/foreman/cli/test_repository.py

[...]

collected 276 items

tests/foreman/cli/test_repository.py ................................... [ 12%]
........................................................................ [ 38%]
........................................................................ [ 64%]
...................................................................ss... [ 90%]
s..ssssssssssssssss.s....                                                [100%]

[...]

========== 256 passed, 20 skipped, 3 warnings in 13897.37s (3:51:37) ===========
```

### DIFF
--- a/tests/foreman/cli/test_repository.py
+++ b/tests/foreman/cli/test_repository.py
@@ -14,6 +14,8 @@
 
 :Upstream: No
 """
+import logging
+
 import pytest
 from fauxfactory import gen_alphanumeric
 from fauxfactory import gen_string
@@ -30,7 +32,6 @@ from robottelo.cli.factory import make_gpg_key
 from robottelo.cli.factory import make_lifecycle_environment
 from robottelo.cli.factory import make_org
 from robottelo.cli.factory import make_product
-from robottelo.cli.factory import make_product_wait
 from robottelo.cli.factory import make_repository
 from robottelo.cli.factory import make_role
 from robottelo.cli.factory import make_user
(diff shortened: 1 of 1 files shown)


mirekdlugosz:
@tpapaioa there's now conflict that prevents merge of this PR. Please fix it.

tpapaioa:
Rebased, applied fixes from `pre-commit` checks.

robottelo/cleanup.py
```diff
@@ -34,6 +34,21 @@ def host_cleanup(host_id=None):
     entities.Host(id=host_id).delete()
 
 
+def repo_cleanup(repo_id=None):
```
mshriver, line 37: I think this change is appropriate within the scope of this PR, but it makes me think we can create a simple cleanup method that takes an entity type and ID to implement this cleanup pattern across every entity.

tests/foreman/cli/test_repository.py
```diff
@@ -75,68 +83,332 @@
...
-            > 0,
-            timeout=30,
-            delay=2,
-            logger=self.logger,
-        )
-        return self._get_image_tags_count(repo=repo)
+    @request.addfinalizer
+    def _cleanup():
+        if Org.exists(search=('id', org['id'])):
+            org_cleanup(org['id'])
```
mshriver, line 333: @rplevka has raised concerns about deleting organizations, because creating entities as admin user puts the entity in the scope of every organization.  Deleting the organization could have impact on running tests, especially in the current CI structure where xdist is used against a single satellite target. We will likely have to leave this off for now, and fix the behavior at a fundamental level before starting to clean up organizations during test execution.

tests/foreman/cli/test_repository.py
```diff
@@ -75,68 +83,332 @@
...
+
+@pytest.fixture
+def repo_options(request):
+    """Return the options that were passed as indirect parameters."""
+    yield getattr(request, 'param', {}).copy()
+
+
+@pytest.fixture
+def org(request):
+    yield _org(request)
```
mshriver, line 385: Because the cleanup is registered with `request` fixture, these fixtures can just `return`. Functionally equivalent, but yield implies generator behavior or additional statements.

tests/foreman/cli/test_repository.py
```diff
@@ -75,68 +83,332 @@
...
-    """Repository CLI tests."""
+YUM_REPOS = (FAKE_0_YUM_REPO, FAKE_1_YUM_REPO, FAKE_2_YUM_REPO, FAKE_3_YUM_REPO, FAKE_4_YUM_REPO)
+PUPPET_REPOS = (
+    FAKE_1_PUPPET_REPO,
+    FAKE_2_PUPPET_REPO,
+    FAKE_3_PUPPET_REPO,
+    FAKE_4_PUPPET_REPO,
+    FAKE_5_PUPPET_REPO,
+)
+INVALID_CREDS = invalid_http_credentials()
```
mshriver, line 102: I appreciate the effort to make these parameters programmatic and consistent, which we should keep.

I would like to avoid calling these functions at module import, as well as the call to `parametrized` at L294. Let's find a strategy to delay these calls to test generation/collection.

mshriver (changes requested):
Want to discuss organization cleanup, and calling parametrization functions on module import. Thanks @tpapaioa !

tests/foreman/cli/test_repository.py
rplevka: just for the record - this is not a case of every entity. IIRC, this was happening with e.g. capsules.

tests/foreman/cli/test_repository.py
tpapaioa: `invalid_http_credentials()`, `parametrized()`, etc., have to be called during module import by the test method decoration, but I've moved them into each individual test's `@pytest.mark.parametrize()`, because otherwise duplicate repo names causes problems during repo fixture setup, now that the repo cleanup code has been removed.

tests/foreman/cli/test_repository.py
tpapaioa: Done.

tests/foreman/cli/test_repository.py
tpapaioa: Removed cleanup code pending further discussion.

robottelo/cleanup.py
tpapaioa: Removed cleanup code for now.

tests/foreman/cli/test_repository.py
```diff
@@ -75,981 +76,1249 @@
...
+        'repo_options',
+        [
+            {
+                'content-type': 'docker',
+                'name': gen_string('alpha'),
+                'docker-upstream-name': 'fedora/rabbitmq',
+            }
+        ],
+        indirect=True,
+    )
```
mshriver, line 153: Consider specifying a meaningful ID for the parameter.

mshriver (approved):

swadeley:
Hello, why no test results after last commit?

PR #8108, opened 2020-10-30
Use source_display_name param to randomize source name

lhellebr:
```
$ pytest tests/foreman/cli/test_remoteexecution.py::TestRemoteExecution::test_positive_run_receptor_installer
[...]
================================================================================================= 1 passed, 2 warnings in 187.34s (0:03:07) ==================================================================================================

```

### DIFF
--- a/tests/foreman/cli/test_remoteexecution.py
+++ b/tests/foreman/cli/test_remoteexecution.py
@@ -511,6 +511,14 @@ class TestRemoteExecution:
 
         :BZ: 1818076
         """
+        # Set Host parameter source_display_name to something random.
+        # To avoid 'name has already been taken' error when run multiple times
+        # on a machine with the same hostname.
+        host_id = Host.info({'name': settings.server.hostname})['id']
+        Host.set_parameter(
+            {'host-id': host_id, 'name': 'source_display_name', 'value': gen_string('alpha')}
+        )
+
         template_name = 'Configure Cloud Connector'
         invocation = make_job_invocation(
             {


tests/foreman/cli/test_remoteexecution.py
```diff
@@ -511,6 +511,13 @@ def test_positive_run_receptor_installer(self):
 
         :BZ: 1818076
         """
+        # Set Host parameter source_display_name to something random.
+        # To avoid 'name has already been taken' error when run multiple times
+        # on a machine with the same hostname.
+        host_id = Host.info({'name': settings.server.hostname})['id']
+        Host.set_parameter({'host-id': host_id, 'name': 'source_display_name',
+                            'value': gen_string('alpha')})
```
mirekdlugosz, line 519: black doesn't like this syntax and fails travis. See travis log for details.
Please fix.

mirekdlugosz (approved):
I admit I don't fully understand this.
I mean - I understand that right now if you try to run this test twice against the same Satellite, it will fail. And with that change, it won't fail.
But aren't Ansible playbooks supposed to be idempotent? Aren't we hiding legitimate product bug here?

I would appreciate a little more explanation, but that's not blocking, so ACK.

lhellebr:
Mirek, there's a crucial difference between these two cases:
1) I run a playbook. Then, on the same machine, I run the playbook again.
2) I run a playbook. Get rid of the machine. Then, I get another machine that gets the same hostname and I run the playbook.

In the first case, if this running a playbook failed, I would consider it a product bug.
In the second case, the Receptor has never been installed on the machine which the playbook is being run on. It tries to register a source, using the default value (hostname) as Source name. It then fails because some other machine previously created a Source with the same name. I thing this is legitimate behavior.

mirekdlugosz:
@lhellebr thanks for explanation.
Personally, I think CRC/Satellite/Receptor/playbook should handle second case as well (they should use machine-id from subscription-manager or Satellite id to uniquely identify source, instead of using FQDN), but I can understand why someone would not consider it a bug. If you haven't already, I would suggest to submit BZ for this behavior.
But I wouldn't block this PR until this BZ is resolved.

lhellebr:
When default source name (=hostname) is already taken, you can manually specify another name. If the default name is taken, it either shows that you use two different machines with the same hostname, or that you manually specified the name before. I think it's ok to fail in that case. @adamruzicka, what is your opinion?

adamruzicka:
> they should use machine-id from subscription-manager or Satellite id to uniquely identify source, instead of using FQDN)

We had this in an early version and IIRC received a pushback, because people looking at sources had trouble identifying that one satellite source they were looking for.

I'm with @lhellebr on this one, imo it is better to crash hard and let the user know there's something fishy going on rather than just overwriting whatever is with sources.

mshriver (approved):

PR #8109, opened 2020-10-30
Fix Reruning all failed tests check

jyejare:
A quick fix where running `all` failed tests check was throwing an exception that one of the `defect_type` options should be provided.

Since the check for failed_option flag was checking for `all` in list instead of string.

### DIFF
--- a/pytest_plugins/rerun_rp/rerun_rp.py
+++ b/pytest_plugins/rerun_rp/rerun_rp.py
@@ -126,7 +126,7 @@ def pytest_collection_modifyitems(items, config):
     test_args = {}
     if fail_args:
         test_args['status'] = 'failed'
-        if not fail_args == ['all']:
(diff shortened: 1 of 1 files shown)


JacobCallahan (approved):
ACK

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -126,7 +126,7 @@ def pytest_collection_modifyitems(items, config):
     test_args = {}
     if fail_args:
         test_args['status'] = 'failed'
-        if not fail_args == ['all']:
+        if not fail_args == 'all':
```
mshriver, line 129: `if fail_args != 'all'` would probably be a bit better syntax here, but this works.

mshriver (approved):

Issue #7722, opened 2020-04-07
refactor api.discovered_host test cases


rplevka:
there are certain areas that can be improved,
e.g. sourcing the default settings in one request,
enhance the assertions for discovered hosts, etc.

PR #8112, opened 2020-10-30
pytest conversion - api.test_discoveredhost

rplevka:
- implement read_log method (the tests can do assertions based on the server-side live logs)
- VM network in `libvirt_discovery` module no longer hardcoded to `bridge` - this allows the tests to be executed in custom (local) environments
- pytest conversion - api/test_discovered_host (+ some parametrization added)
- test coverage - api/test_discovered_host
- fixes https://github.com/SatelliteQE/robottelo/issues/7722

- [ ] relies on: https://github.com/SatelliteQE/nailgun/pull/762

test_results

### DIFF
--- a/robottelo/libvirt_discovery.py
+++ b/robottelo/libvirt_discovery.py
@@ -56,7 +56,8 @@ class LibvirtGuest:
         libvirt_server=None,
         image_dir=None,
         mac=None,
-        bridge=None,
+        network=None,
+        network_type=None,
     ):
         self.cpu = cpu
         self.ram = ram
@@ -77,16 +78,22 @@ class LibvirtGuest:
             self.image_dir = image_dir
         if mac is None:
             self.mac = _gen_mac_for_libvirt()
-        if bridge is None:
-            self.bridge = settings.vlan_networking.bridge
+        if network is None:
+            if settings.vlan_networking.bridge or settings.vlan_networking.network:
+                if settings.vlan_networking.bridge:
+                    self.network_type = 'bridge'
+                    self.network = settings.vlan_networking.bridge
+                else:
+                    self.network_type = 'network'
+                    self.network = settings.vlan_networking.network
         else:
-            self.bridge = bridge
-        if not self.bridge:
+            self.network_type = network_type
+            self.network = network
+        if not self.network:
             raise LibvirtGuestError(
-                'A bridge name must be provided. Make sure to fill '
-                '"bridge" on vlan_networking section of your robottelo '
-                'configuration. Or provide a not None bridge '
-                'argument.'
+                'A network name must be provided. Make sure to fill "bridge" or "network" on '
+                'vlan_networking section of your robottelo configuration.'
+                'Or provide a not None network and network_type arguments.'
             )
         self.boot_iso = boot_iso
         self.extra_nic = extra_nic
@@ -109,7 +116,7 @@ class LibvirtGuest:
         command_args = [
             'virt-install',
             '--hvm',
-            '--network=bridge:{vm_bridge}',
+            '--network={vm_nw_type}:{vm_network}',
             '--mac={vm_mac}',
             '--name={vm_name}',
             '--ram={vm_ram}',
@@ -118,11 +125,13 @@ class LibvirtGuest:
             '--os-variant=rhel7',
             '--disk path={image_name},size=8',
             '--noautoconsole',
+            '--graphics spice,listen=0.0.0.0',
         ]
 
         if not self.boot_iso:
             # Required for PXE-based host discovery
-            command_args.append('--pxe')
+            command_args.append('--boot network')
+
         else:
             # Required for PXE-less host discovery, where we boot the host
             # with bootable discovery ISO
@@ -132,7 +141,7 @@ class LibvirtGuest:
 
         if self.extra_nic:
             nic_mac = _gen_mac_for_libvirt()
-            command_args.append('--network=bridge:{vm_bridge}')
+            command_args.append('--network={vm_nw_type}:{vm_network}')
             command_args.append(f'--mac={nic_mac}')
 
         if self._domain is None:
@@ -145,7 +154,8 @@ class LibvirtGuest:
 
         self.hostname = f'{self.guest_name}.{self._domain}'
         command = ' '.join(command_args).format(
-            vm_bridge=self.bridge,
+            vm_nw_type=self.network_type,
+            vm_network=self.network,
             vm_mac=self.mac,
             vm_name=self.hostname,
             vm_ram=self.ram,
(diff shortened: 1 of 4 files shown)


robottelo/libvirt_discovery.py
```diff
@@ -118,11 +118,14 @@ def create(self):
             '--os-variant=rhel7',
             '--disk path={image_name},size=8',
             '--noautoconsole',
+            '--graphics spice,listen=0.0.0.0',
```
rplevka, line 121: this helps to remotely attach to the console via VNC while debugging

tests/foreman/api/test_discoveryrule.py
```diff
@@ -116,3 +116,55 @@ def test_negative_create_with_invalid_host_limit_and_priority():
         entities.DiscoveryRule(max_count=gen_string('alpha')).create()
     with pytest.raises(HTTPError):
         entities.DiscoveryRule(priority=gen_string('alpha')).create()
+
```
rplevka, line 119: moved the following to `discovery_rule` component

robottelo/libvirt_discovery.py
```diff
@@ -77,16 +78,22 @@ def __init__(
             self.image_dir = image_dir
         if mac is None:
             self.mac = _gen_mac_for_libvirt()
-        if bridge is None:
-            self.bridge = settings.vlan_networking.bridge
+        if network is None:
+            if settings.vlan_networking.bridge or settings.vlan_networking.network:
```
mirekdlugosz, line 82: If `network` is None (which is default), and neither of settings are defined (which shouldn't happen, but technically can), then we will thrown AttributeError on line 92, when trying to check if `self.network` is falsy. That's probably not what you wanted.

These three nested ifs make me think that there's probably better way to write this. Maybe helper function/method? Something that would take `network` and `network_type` as arguments, and return `self.network_type`  and `self.network` values back? Untested proposal below:

```python
def get_network_settings(network, network_type):
    if network is not None:
        return (network, network type)
    if settings.vlan_networking.bridge:
        return ('bridge', settings.vlan_networking.bridge)
    if settings.vlan_networking.network:
        return ('network', settings.vlan_networking.network)
    return (None, None)  # return None tuple, so tuple unpacking and assignment always works
                         # we bail out in caller if `self.network` is falsy
```

robottelo/libvirt_discovery.py
```diff
@@ -118,11 +125,14 @@ def create(self):
...
             '--disk path={image_name},size=8',
             '--noautoconsole',
+            '--graphics spice,listen=0.0.0.0',
         ]
 
         if not self.boot_iso:
             # Required for PXE-based host discovery
-            command_args.append('--pxe')
+            # command_args.append('--pxe')
+            command_args.append('--boot network')
```
mirekdlugosz, line 134: Could you please add some explanation to this change? Is it backwards compatible?

tests/foreman/api/test_discoveredhost.py
```diff
@@ -67,561 +77,484 @@ def _create_discovered_host(name=None, ipaddress=None, macaddress=None):
...
+    groups = re.search('DHCPACK on (\\d.+) to', dhcp_pxe.out)
+    assert len(groups.groups()) == 1, 'Unable to parse bootloader ip address'
+    pxe_ip = groups.groups()[0]
+
+    # assert that server retrieves pxelinux config over TFTP
+    for pattern in [
+        (f'Client {pxe_ip} finished pxelinux.0', 'pxelinux.0'),
+        (f'Client {pxe_ip} finished pxelinux.cfg/default', 'pxelinux.cfg/default'),
+        (f'Client {pxe_ip} finished boot/fdi-image/vmlinuz0', 'fdi-image/vmlinuz0'),
+        (f'Client {pxe_ip} finished boot/fdi-image/initrd0.img', 'fdi-image/initrd0.img'),
```
mirekdlugosz, line 144: Hm, could you get away with only iterating over file names, and constructing `pattern[0]` inside loop? Looks like they all follow same pattern, you only need to pass `pxe_ip` and file name itself.

tests/foreman/api/test_discoveredhost.py
```diff
@@ -67,561 +77,484 @@ def _create_discovered_host(name=None, ipaddress=None, macaddress=None):
...
+    try:
+        _wait_for_log(channel, f'\\[I\\|app\\|{req_id}\\] Completed 201 Created')
+    except TimedOutError:
+        # raise assertion error
+        raise AssertionError(f'Timed out waiting for "/facts" 201 response')
+
+    default_config = entity_mixins.DEFAULT_SERVER_CONFIG
+    for _ in range(30):
+        discovered_host = entities.DiscoveredHost(user_config or default_config).search(
+            query={'search': f'name={host.guest_name}'}
```
mirekdlugosz, line 193: Would it make sense to rewrite this loop using `wait_for`? I saw you using it elsewhere.

mirekdlugosz (commented):
Great work! That's most challenging PR I've reviewed in a while.
I'm mostly concerned about changes in `LibvirtGuest` class. Other comments are suggestions.

robottelo/libvirt_discovery.py
rplevka: Sure, here's the description from `virt-install` manpage:

```
--pxe
           Install from PXE. This just tells the VM to boot off the network for the first boot.
```
```
--boot BOOTOPTS
           Optionally specify the post-install VM boot configuration. This option allows specifying a boot device
           order, permanently booting off kernel/initrd with option kernel arguments, and enabling a BIOS boot menu
           (requires libvirt 0.8.3 or later)

           --boot cdrom,fd,hd,network
               Set the boot device priority as first cdrom, first floppy, first harddisk, network PXE boot.

```

tl;dr - the usage of `--pxe` has been wrong since the beginning - it is used for 1-time boot and such machine won't survive reboots.

the `--boot network` on the other hand properly sets the persistent boot device (order) so the VM uses it even after reboot - since we also test VM behavior after reboot, this is a necessity.

Btw, i should probably remove the commented line

tests/foreman/api/test_discoveredhost.py
rplevka: Absolutely, this isn't my code (I just added a bunch of new assertion before it).
Will do

robottelo/libvirt_discovery.py
rplevka: I'm counting on [settings.verify()](https://github.com/SatelliteQE/robottelo/blob/master/robottelo/config/base.py#L1148), which accepts exactly 1 of bridge or network in `[vlan_networking]`

tests/foreman/api/test_discoveredhost.py
rplevka: I certainly could, but I didn't want to sacrifice readability. I think this way anybody could easily see what we're up for.
What do you think?

tests/foreman/api/test_discoveredhost.py
rplevka: - comment addressed.

pondrejk (approved):

mirekdlugosz (approved):
Let's not hold this any longer. ACK and merge

PR #8116, opened 2020-11-02
Pytest conversion for installer test cases

devendra104:
The purpose of this PR to convert installers test cases from unit-test to pytest.

**Test Results:**

```
$ pytest tests/foreman/installer/test_installer.py
platform linux -- Python 3.6.6, pytest-6.1.1, py-1.9.0, pluggy-0.13.1                                                                                                                                                                                                 
tests/foreman/installer/test_installer.py ...sss                                                                                                                                                            [100%]

================================================================================================ warnings summary =================================================================================================

============================================================================== 3 passed, 3 skipped, 6 warnings in 126.61s (0:02:06) =============================================================================== 
```

### DIFF
--- a/tests/foreman/installer/test_installer.py
+++ b/tests/foreman/installer/test_installer.py
@@ -26,7 +26,6 @@ from robottelo.decorators import tier1
 from robottelo.decorators import tier3
 from robottelo.decorators import upgrade
 from robottelo.helpers import get_host_info
-from robottelo.test import TestCase
 
 PREVIOUS_INSTALLER_OPTIONS = {
     '--[no-]colors',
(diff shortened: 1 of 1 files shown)


omkarkhatavkar (approved):
ACK

tests/foreman/installer/test_installer.py
```diff
@@ -1345,112 +1344,107 @@
...
-                result = ssh.command(status_format.format(service))
-                self.assertEqual(result.return_code, 0)
-                self.assertEqual(len(result.stderr), 0)
+    :id: 85fd4388-6d94-42f5-bed2-24be38e9f104
 
-        # check status reported by hammer ping command
-        result = ssh.command(
-            'hammer -u {0[0]} -p {0[1]} ping'.format(settings.server.get_credentials())
-        )
+    :expectedresults: All services {'elasticsearch', 'foreman-proxy', 'httpd',
```
mirekdlugosz, line 1415: Given that this list of services in docstring is already outdated, would it make sense to remove it? Just say "All services are started".
Non-blocking comment.

tests/foreman/installer/test_installer.py
```diff
@@ -1345,112 +1344,107 @@
...
+    'pulp_workers', 'qdrouterd', 'qpidd', 'tomcat'} are started
+    """
+    major_version = get_host_info()[1]
+    service_name = 'tomcat6' if major_version == RHEL_6_MAJOR_VERSION else 'tomcat'
+    SATELLITE_SERVICES.add(service_name)
+    for service in SATELLITE_SERVICES:
+        if major_version >= RHEL_7_MAJOR_VERSION:
+            result = ssh.command(f'systemctl status {service}')
+        else:
+            result = ssh.command(f'service {service} status')
```
mirekdlugosz, line 1428: Personally, I liked previous version better. Satellite RHEL version is not going to change during test execution, so it makes sense to extract `if` outside of the loop - we don't waste resources on condition that is the same in every iteration.

It's not going to have visible impact on test execution time, so it's another non-blocking comment.

mirekdlugosz (approved):
ACK.

I left two suggestions. They aren't blocking, so if you don't want to implement them, feel free to merge this yourself (or add comment that you won't include these changes and I'll merge). I'm not merging now to give you opportunity to reply to my suggestions.

tests/foreman/installer/test_installer.py
devendra104: you are right, "All services are started" is looks more appropriate than mentioning each service in the docstring.

tests/foreman/installer/test_installer.py
devendra104: Fixed.

mirekdlugosz (approved):
Thanks for including suggested changes.
ACK and merge.
Travis failure is unrelated, caused by #8003 merge. Fixed in #8121

PR #8122, opened 2020-11-04
record start_time property workaround

rplevka:
This is a workaround for:
https://github.com/pytest-dev/pytest/issues/7767

where the pre-defined `record_testsuite_timestamp_xml` fixture doesn't work with `xdist` yet.
This commit keeps the fixture, but sets `autouse=False`

and implements a manual property recording inside the `report_header` hook

### DIFF
--- a/tests/foreman/conftest.py
+++ b/tests/foreman/conftest.py
@@ -8,6 +8,7 @@ try:
     from pytest_reportportal import RPLogger, RPLogHandler
 except ImportError:
     pass
+from _pytest.junitxml import xml_key
 from robottelo.config import settings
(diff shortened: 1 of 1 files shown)


tstrych (approved):
LGTM
ACK

rplevka:
thanks @tstrych for discussing the `.now()` vs `.utcnow()` after reviewing the code I realised we actually use `.utcnow()` in the logger too, so I reverted it back just as you suggested.

lpramuk (approved):

jyejare (approved):

mshriver (approved):

PR #8125, opened 2020-11-04
dynaconf validators - add missing settings and default values

mirekdlugosz:
Dynaconf validators improvements:
* Add settings that were added to Robottelo between #8003 was opened and merged
* Add default values to dynaconf, mimic current behavior of Settings class

No tests run or required, as we don't have dynaconf settings file yet.


jaryn (approved):

mshriver (approved):

JacobCallahan (approved):
ACK

PR #8126, opened 2020-11-04
fix the test suite start_time format

rplevka:
In the recent patch (https://github.com/SatelliteQE/robottelo/pull/8122) I forgot to format the datetime string to be acceptable for the `rp_tools` script down the pipe.

old format:
`2020-11-04 17:32:13.066895`

```
    <testsuites>
        <testsuite errors="0" failures="0" ...">
            <properties>
                <property name="start_time" value="2020-11-04T19:40:54"/>
            </properties>
            <testcase ... file="tests/foreman/api/test_user.py" ...>
                <properties>
                    <property name="start_time" value="2020-11-04T19:40:58"/>
                </properties>
           ...
     ...
```

### DIFF
--- a/tests/foreman/conftest.py
+++ b/tests/foreman/conftest.py
@@ -12,6 +12,7 @@ from _pytest.junitxml import xml_key
 from robottelo.config import settings
 from robottelo.decorators import setting_is_set
 
+FMT_XUNIT_TIME = "%Y-%m-%dT%H:%M:%S"
 
 def log(message, level="DEBUG"):
     """Pytest has a limitation to use logging.logger from conftest.py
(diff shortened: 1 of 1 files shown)


tests/foreman/conftest.py
```diff
@@ -51,7 +51,7 @@ def pytest_report_header(config):
         now = datetime.datetime.utcnow()
         xml = config._store.get(xml_key, None)
         if xml:
-            xml.add_global_property('start_time', now)
+            xml.add_global_property('start_time', now.strftime("%Y-%m-%dT%H:%M:%S"))
```
mshriver, line 54: @rplevka Since we're hitting this inconsistency, please define a module constant like `FMT_XUNIT_TIME` for this format string, and use it both here and at L150 and L156 where the workers will be calling the fixtures to add these timestamps.

rplevka: thanks for the suggestion, I addressed your comment

mshriver (approved):

PR #8132, opened 2020-11-05
add robottelo test to assert timestamps in generated XML junit reports

rplevka:
Implements a test for the junit-reporting configuration defined in `conftest.py`.

this adds 2 tests ~~and registers 1 new marker (`dummy`)~~:

~~- one test is marked as `dummy` as it is not meant to be executed standalone. It is supposed to be executed by the other test~~
- the actual test which uses the robottelo `conftest.py` and `pytest.ini`, executes the dummy test and performs assertions on the generated report.

This is supposed to be executed by `tox` as a part of regression testing on new PRs

### DIFF
--- a/tests/robottelo/test_report.py
+++ b/tests/robottelo/test_report.py
@@ -1,0 +1,82 @@
+import datetime
+import glob
+import os
+from pathlib import Path
+from shutil import copyfile
+from tempfile import NamedTemporaryFile
+
+import pytest
+import xmltodict
+from fauxfactory import gen_string
+
+XUNIT_TIME_FORMAT = "%Y-%m-%dT%H:%M:%S"
+dummy_test_count = 2
+dummy_test = f'''import pytest
+
+@pytest.mark.parametrize('param', list(range(0, {dummy_test_count})))
+def test_dummy(param):
+    """A dummy test used by test_junit_timestamps.
+    Not to be run as a standalone test
+    """
+    print(f'running test {{param}}')
+    pass
+'''
+
+paths = [
+    ["['testsuites']['testsuite']['properties']['property']"],
+    [
+        f"['testsuites']['testsuite']['testcase'][{i}]['properties']['property']"
+        for i in range(dummy_test_count)
+    ],
+]
+
+
+@pytest.fixture(scope="session")
+def config():
+    path = Path(__file__).parent
+    conf_path = path / "../foreman"
+    copyfile(str(conf_path / '../foreman/conftest.py'), f'{str(path)}/conftest.py')
+    yield
+    os.remove(f'{str(path)}/conftest.py')
+
+
+@pytest.fixture(scope="session")
+def exec_test(config):
+    test_dir = str(Path(__file__).parent)
+    report_file = f'report_{gen_string("alphanumeric")}.xml'
+    with NamedTemporaryFile(dir=test_dir, mode='w', prefix='test_', suffix='.py') as f:
+        f.seek(0)
+        f.write(dummy_test)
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -10,4 +10,4 @@ deps =
     -rrequirements.txt
     pytest-cov
     pycurl
-commands = py.test --cov --cov-config=.coveragerc tests/robottelo
+commands = py.test --cov --cov-config=.coveragerc tests/robottelo -m "not dummy"
```
mshriver, line 13: I'm not a fan of this pattern, having a marker used for the unit tests that we then have to explicitly exclude from running the unit tests.

Is there really harm to the 'dummy' test running during a normal unit test execution? Just the time to execute it?

It does feel like overkill for this use case, but I much prefer marker implementations that follow a pattern of modifying test collection unless an option is included to run them when using a 'negative' marker like this. In other words, a test marked with `dummy` does not run in normal test collection, and is only included in collection when `--dummy` is passed.

mshriver (changes requested):
This is a great unit test to have, but I'd like to discuss how the marker is defined and used.

tox.ini
rplevka: Not a huge fan of this solution either. If you're okay with running the "empty" test, I'm more than happy to remove the marker.

Another possible approach could be to actually store the dummy test definition in a non-py plaintext file (or multiline string) and dump it to a `test_foo.py` file as a part of the fixture; and remove it for a cleanup.

mshriver: I think I like that second idea, you should be able to use `tempfile.NamedTemporaryFile` in a context, and on linux the `pytest.main` call should be able to read the written file. That will cleanly handle creation, potential name conflicts, and deletion on the context close.

rplevka:
> This is a great unit test to have, but I'd like to discuss how the marker is defined and used.

@mshriver I updated the test case as we discussed. Please, re-review.

tox.ini
mshriver: Looks like this mark isn't needed anymore for tox collection, with your updates to the unit test?

rplevka: ah, sorry

pytest.ini
```diff
@@ -1,2 +1,4 @@
 [pytest]
 junit_logging = all
+markers =
+	dummy: marks dummy tests to be executed only by different tests
```
lpramuk, line 4: `... only by robottelo internal tests`  ?

tests/robottelo/test_report.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+@pytest.fixture(scope="session")
+def exec_test(config):
+    report_file = f'report_{gen_string("alphanumeric")}.xml'
+    pytest.main(
+        [
+            '-n2',
+            f'--junit-xml={report_file}',
+            f'{str(Path(__file__).parent)}/test_report.py::test_dummy',
```
lpramuk, line 39: `f'{str(Path(__file__))}::test_dummy'`   <- isn't this better?

why using `.parent` when you descend then to `__file__`

pytest.ini
rplevka: sorry, this is an outdated PoC. This got completely removed

mshriver (approved):

lpramuk (approved):
Nice!  Pytest Matryoshka

mshriver:
@rplevka I definitely want to steal the pattern you have in this unit test for the pytest plugin unit tests for issue_handlers and the testimony_markers introduced in #8163

PR #8133, opened 2020-11-05
Fix skip_if_open marker use

mshriver:
I broke this in #8092, missed in review/CI.

### DIFF
--- a/tests/foreman/virtwho/cli/test_kubevirt.py
+++ b/tests/foreman/virtwho/cli/test_kubevirt.py
@@ -51,7 +51,7 @@ def virtwho_config(form_data):
     return VirtWhoConfig.create(form_data)['general-information']
 
 
-@pytest.mark.skip_if_open('BZ:1735540', reason='We have not supported kubevirt hypervisor yet')
+@pytest.mark.skip_if_open('BZ:1735540')
 class TestVirtWhoConfigforKubevirt:
     @pytest.mark.tier2
     def test_positive_deploy_configure_by_id(self, form_data, virtwho_config):
--- a/tests/foreman/virtwho/api/test_kubevirt.py
+++ b/tests/foreman/virtwho/api/test_kubevirt.py
@@ -20,7 +20,6 @@ from nailgun import entities
 from wait_for import wait_for
 
 from robottelo.config import settings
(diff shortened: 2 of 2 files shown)


JacobCallahan (approved):
ACK

tests/foreman/virtwho/cli/test_kubevirt.py
```diff
@@ -51,7 +51,7 @@ def virtwho_config(form_data):
     return VirtWhoConfig.create(form_data)['general-information']
 
 
-@pytest.mark.skip_if_open('BZ:1735540', reason='We have not supported kubevirt hypervisor yet')
+@pytest.mark.skip_if_open('BZ:1735540')
```
hkx303, line 54: Thanks Mike, but I prefer to use the `skipif` to keep it in line with https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/virtwho/api/test_kubevirt.py#L58
```
@pytest.mark.skipif(
    condition=(is_open('BZ:1735540')),
    reason='We have not supported kubevirt hypervisor yet',
)
```

hkx303: Or we should also modify this in https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/virtwho/api/test_kubevirt.py#L58

mshriver: I would actually prefer to use skip_if_open, as BZ skipping is the specific use for this marker. I can update api/test_kubevirt to use it as well?  Going this direction also means we don't have to import the is_open method into test modules.

mshriver: @hkx303 commit updated with both uses being consistent with `skip_if_open`, thanks for being flexible!

lpramuk (approved):

hkx303 (commented):
Then we should also delete this line :)
https://github.com/SatelliteQE/robottelo/blob/49f47af7d01094b740839aef14ec28548ad1ed83/tests/foreman/virtwho/api/test_kubevirt.py#L23

hkx303 (approved):
Give my ACK as I hope we can merge it today (US time), thank you!

hkx303 (approved):

PR #8136, opened 2020-11-05
Call robottelo by name for logger

swadeley:
Plinko fails if logger called using  __name__


JacobCallahan (approved):
ACK, thanks

ColeHiggins2 (approved):
ACK

pondrejk (approved):
poor Plinko

Issue #8138, opened 2020-11-06
Need virtwho.properties.sample in Robottelo repo


Gauravtalreja1:
For collecting virtwho-tests from `./tests/foreman/virtwho/*`, a `virtwho.properties` file is required, so there should be a sample file that exists in the robottelo repo.

Error message:
```
_________________________________________ ERROR collecting tests/foreman/virtwho/ui/test_*.py_________________________________________
tests/foreman/virtwho/ui/test_*.py:20: in <module>
    from robottelo.virtwho_utils import deploy_configure_by_command
robottelo/virtwho_utils.py:18: in <module>
    virtwho.configure()
robottelo/config/virtwho.py:351: in configure
    raise ImproperlyConfigured(f'Not able to find settings file at {settings_path}')
E   robottelo.config.base.ImproperlyConfigured: Not able to find settings file at /home/gtalreja/sat_workspace/robottelo/virtwho.properties
```

hkx303:
Thank you, will do

PR #8140, opened 2020-11-06
Add Github Action workflow for Robottelo

Gauravtalreja1:
**Proposal for:**

- [x] Adding a Github Action workflow for Robottelo.

- [x] Dropping/Decommissioning Travis-CI from Robottelo.

- [x] Adding a `broker_directory` in `robottelo.properties.sample`.

- [x] Adding a blank `virtwho.properties.sample` file to workaround https://github.com/SatelliteQE/robottelo/issues/8138 .

**Results**
Github Action run will look like:
https://github.com/Gauravtalreja1/robottelo/actions/runs/349472736
https://github.com/Gauravtalreja1/robottelo/pull/1/checks

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -1,0 +1,61 @@
+name: Robottelo - CI
+
+on:
+  push:
+  pull_request:
+    types: ["opened", "synchronize", "reopened"]
+
+env:
+    PYCURL_SSL_LIBRARY: gnutls
+
+jobs:
+  codechecks:
+    name: Code Quality
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8, 3.9]
+    steps:
+      - name: Checkout Robottelo
+        uses: actions/checkout@v2
+
+      - name: Set Up Python-${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install Dependencies
+        run: |
+          sudo apt-get install -y libgnutls28-dev libcurl4-openssl-dev libssl-dev
+          wget https://raw.githubusercontent.com/SatelliteQE/broker/master/broker_settings.yaml.example
+          pip install -U --no-binary=pycurl -r requirements.txt -r requirements-optional.txt
+          cp robottelo.properties.sample robottelo.properties
+          cp broker_settings.yaml.example broker_settings.yaml
+          cp virtwho.properties.sample virtwho.properties
+
+      - name: Pre Commit Checks
+        uses: pre-commit/action@v2.0.0
+
+      - name: Collect Tests
+        run: |
+          pytest -n 8 --setup-plan --disable-pytest-warnings tests/foreman/ tests/robottelo/
+          pytest -n 8 --setup-plan --disable-pytest-warnings -m pre_upgrade tests/upgrades/
+          pytest -n 8 --setup-plan --disable-pytest-warnings -m post_upgrade tests/upgrades/
+
+      - name: Test Robottelo Coverage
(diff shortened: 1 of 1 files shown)


.github/workflows/action.yml
```diff
@@ -0,0 +1,61 @@
...
+  push:
+  pull_request:
+    types: ["opened", "synchronize", "reopened"]
+
+env:
+    PYCURL_SSL_LIBRARY: gnutls
+
+jobs:
+  codechecks:
+    name: 📃 Code Quality
```
omkarkhatavkar, line 13: Looks like the icon are getting detected here

omkarkhatavkar:
@Gauravtalreja1 Let's keep the Travis file for some more time, we can remove later @mshirver thoughts?

.github/workflows/action.yml
mshriver: I'd prefer to not have emotes in our pipeline names.

mshriver:
https://github.com/SatelliteQE/robottelo/pull/8144

You'll see some failures in the test collection until this is merged.

JacobCallahan:
The travis config could be renamed instead of straight up removed, just to make it easy to bring it back (just in case).
If things go well for a while, it can be removed completely.

.github/workflows/action.yml
```diff
@@ -0,0 +1,61 @@
...
+    runs-on: ubuntu-latest
```
JacobCallahan, line 14: Can we change this to a fedora image?

.github/workflows/action.yml
```diff
@@ -0,0 +1,61 @@
...
+        uses: actions/checkout@v2
+
+      - name: 🐍 Set Up Python-${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install Dependencies
+        run: |
+          sudo apt-get install -y libgnutls28-dev libcurl4-openssl-dev libssl-dev
```
JacobCallahan, line 29: If we can do fedora, this will need to be changed

JacobCallahan (changes requested):
If possible, Fedora is preferred.

.github/workflows/action.yml
omkarkhatavkar: Looks like they dont support it out of the box, I saw this thread https://github.com/actions/virtual-environments/issues/1443, Any specific reasons we have to use this with fedora image? because we were using xenial Travis which is ubuntu distribution. I don't think we should be blocking this against fedora image.

.github/workflows/action.yml
omkarkhatavkar: If you want to support fedora we have matrix kind of thing. but I guess not that necessary.

.github/workflows/action.yml
mshriver: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

Not without changing the pipeline to be a docker image strategy. Github offers ubuntu VMs for these pipelines. Same as in travis, we're running on ubuntu.

Gauravtalreja1:
**New Changes:**
1. As per suggestion from @lpramuk , updated workflow to use `--setup-plan` in test collection rather than simple `--collect-only`.
2. Keeping a Travis config `.travis.yml` for some more time, for some more PR's.
3. Removed all emoticons from pipeline/task names.
3. Discarding changes done in `robottelo.properties.sample` as it was already there in that file.

**Latest Run:** https://github.com/Gauravtalreja1/robottelo/pull/1/checks

omkarkhatavkar (approved):
ACK

JacobCallahan (approved):
ACK

PR #8141, opened 2020-11-06
Add reason for pytest.skipif

mshriver:
When using a boolean as a condition, the reason kwarg is required.

Followup to #8092

### DIFF
--- a/tests/foreman/cli/test_puppetclass.py
+++ b/tests/foreman/cli/test_puppetclass.py
@@ -37,10 +37,7 @@ def make_puppet():
 
 @pytest.mark.tier2
 @pytest.mark.upgrade
-@pytest.mark.skipif(
-    not settings.repos_hosting_url,
-    reason="repos_hosting_url is not defined in robottelo.properties",
-)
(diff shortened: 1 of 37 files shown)


JacobCallahan (approved):
ACK

lpramuk (approved):

tests/foreman/cli/test_puppetclass.py
```diff
@@ -37,10 +37,7 @@ def make_puppet():
 
 @pytest.mark.tier2
 @pytest.mark.upgrade
-@pytest.mark.skipif(
-    not settings.repos_hosting_url,
-    reason="repos_hosting_url is not defined in robottelo.properties",
-)
+@pytest.mark.skipif((not settings.repos_hosting_url), reason="Missing repos_hosting_url")
```
swadeley, line 40: reason="Missing repos_hosting_url" 
This one has double quotation marks, all the others single.  Does it matter?

swadeley (commented):
ACK

tests/foreman/cli/test_puppetclass.py
swadeley: oops sorry, I was too late

PR #8144, opened 2020-11-06
Fix fixture resolution on a typo

mshriver:
This was causing a few tests to fail. Typo on the fixture name.


lpramuk (approved):

lpramuk:
`--setup-plan` is good check for all fixtures :-)
Nice you fixed this as well.

PR #8147, opened 2020-11-09
[Upgrade Customer Scenario]Pytest conversion for capsule test cases

devendra104:
The purpose of this PR to convert the capsule's unit test(Customer upgrade scenario) to pytest and fix some earlier failures that happened due to the orphaned process.

Test Results:

**Pre-Upgrade**

```
$ pytest -s -m "pre_upgrade" tests/upgrades/test_capsule.py
platform linux -- Python 3.6.6, pytest-6.1.1, py-1.9.0, pluggy-0.13.1                                                                                                                                                           

tests/upgrades/test_capsule.py .
================================================================================================ warnings summary =================================================================================================
============================================================================= 1 passed, 3 deselected, 7 warnings in 141.43s (0:02:21) =============================================================================
```

**Post-Upgrade** 
```
$ pytest -s -m "post_upgrade" tests/upgrades/test_capsule.py
platform linux -- Python 3.6.6, pytest-6.1.1, py-1.9.0, pluggy-0.13.1
plugins: rerunfailures-9.1.1, forked-1.1.3, mock-1.10.4, xdist-1.34.0, services-1.3.1, cov-2.8.1
collected 4 items / 1 deselected / 3 selected                                                                                                                                                                     

tests/upgrades/test_capsule.py ...
================================================================================================ warnings summary =================================================================================================

============================================================================ 3 passed, 1 deselected, 9 warnings in 1154.30s (0:19:14) =============================================================================

```

### DIFF
--- a/tests/upgrades/test_capsule.py
+++ b/tests/upgrades/test_capsule.py
@@ -1,6 +1,6 @@
 """Test for Capsule related Upgrade Scenario's
 
-:Requirement: Upgraded Satellite
+:Requirement: Upgraded Satellite & Capsule
 
 :CaseAutomation: Automated
 
@@ -20,8 +20,6 @@ from nailgun import entities
 from upgrade.helpers.tasks import wait_untill_capsule_sync
 from upgrade_tests import post_upgrade
 from upgrade_tests import pre_upgrade
-from upgrade_tests.helpers.scenarios import create_dict
-from upgrade_tests.helpers.scenarios import get_entity_data
 from upgrade_tests.helpers.scenarios import rpm1
 from upgrade_tests.helpers.scenarios import rpm2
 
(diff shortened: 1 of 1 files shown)


tests/upgrades/test_capsule.py
```diff
@@ -20,256 +20,233 @@
...
+    rpm_name = rpm2.split('/')[-1]
+    org = entities.Organization().search(query={'search': f'name="{DEFAULT_ORG}"'})[0]
+    activation_key = settings.upgrade.rhev_capsule_ak or settings.upgrade.capsule_ak
+    cap_host = settings.upgrade.rhev_cap_host or settings.upgrade.capsule_hostname
+    ak = entities.ActivationKey(organization=org.id).search(
+        query={'search': f'name={activation_key}'}
+    )[0]
+    ak_env = ak.environment.read()
+    repo_url = f'http://{settings.server.hostname}/pub/{repo_name}/'
+    # product = entities.Product(name=f'{prod_name}', organization=org.id).create()
```
pondrejk, line 176: intended?

tests/upgrades/test_capsule.py
```diff
@@ -20,256 +20,233 @@
...
+    product = entities.Product(organization=org.id).create()
+    repo_path = f'/var/www/html/pub/{repo_name}/'
+    create_repo(rpm2, repo_path)
+
+    repo = entities.Repository(product=product.id, name=f'{repo_name}', url=repo_url).create()
+    repo.sync()
+    # content_view = entities.ContentView(name=f'{cv_name}', organization=org.id).create()
```
pondrejk, line 183: same here

tests/upgrades/test_capsule.py
devendra104: No, it was left by mistake.

tests/upgrades/test_capsule.py
devendra104: No, it was left by mistake.

tests/upgrades/test_capsule.py
```diff
@@ -20,256 +20,231 @@
...
+    result = execute(
+        lambda: run(
+            f'[ -f /var/lib/pulp/published/yum/http/repos/'
+            f'{org_name}/{env_name.name}/{content_view.name}/custom/{pre_test_name}_prod/'
+            f'{pre_test_name}_repo/Packages/b/{rpm_name} ]; echo $?'
+        ),
+        host=cap_host,
+    )[cap_host]
+    assert result == '0'
+    __cleanup(content_view, repo, product)
```
mshriver, line 143: There is an important pattern here that should be used in pytest. 

A failure at any statement or assertion in the test function will immediately stop execution of the test function, and transition to `teardown` phase.

This means that calling `__cleanup` at the end of the test is ineffective if the test fails in any way - and we want to run cleanup regardless.

There are two approaches to dealing with this behavior.  I'd say either is acceptable and are good patterns.  One of these is easily applicable to your current test structure, the other would require more refactoring around handling the upgrade `depends_on` marker, but I'll describe both.

1.  Use a yielding fixture. Ideal behavior is the entities that need to be cleaned up are parameters of the fixture, and are provided by other fixtures, or created within the fixture function itself.  After `yield`, you can add statements to the fixture that will execute during `teardown` phase. The fixture function does not have to do anything before calling `yield` - so you can write a fixture function that only does cleanup.  This would require more refactoring for a clean pattern in your tests.

2. Use `request.addfinalizer` to register the function to be executed during `teardown` phase.  This should be called as early as possible within the test, to avoid any failures preventing cleanup from running.  `addfinalizer` takes a callable, and you can use a lambda to pass the arguments.  

The request pattern looks like: `request.addfinalizer(lambda: __cleanup(content_view, repo, product))`, and in this particular function I would add that line at L126, immediately after `content_view`, `repo`, and `product` entities are created.

mshriver (changes requested):
Great conversion, but the use of the `--cleanup` method needs to be adjusted to make sure cleanup is run even if the test case fails.

tests/upgrades/test_capsule.py
devendra104: Thanks, Mike for your valuable inputs, it really helps to clean the environment even if some failure happens in mid of the execution, the second approach is more appropriate for these test cases, Similar kind of thing I included in the new pytest conversion PR https://github.com/SatelliteQE/robottelo/pull/8186.

tests/upgrades/test_capsule.py
```diff
@@ -112,57 +111,51 @@ def test_post_user_scenario_capsule_sync(self):
             2. The repos/rpms from satellite should be synced to satellite
 
         """
-        env_name = get_entity_data(self.__class__.__name__)['env_name']
-        org_name = entities.Organization().search(query={'search': f'id={self.org_id}'})[0].label
-        capsule = entities.SmartProxy().search(query={'search': f'name={self.cap_host}'})[0]
+        request.addfinalizer(lambda: __cleanup(content_view, repo, product))
```
mshriver, line 114: Does the lambda work correctly here even though `content_view`, `repo`, and `product` aren't yet defined?

tests/upgrades/test_capsule.py
```diff
@@ -112,57 +111,51 @@ def test_post_user_scenario_capsule_sync(self):
...
-            host=self.cap_host,
-        )[self.cap_host]
-        self.assertEqual('0', result)
+            host=cap_host,
+        )[cap_host]
+        assert result == '0'
 
 
-class Scenario_capsule_sync_2(APITestCase):
+class TestCapsuleSync2:
```
mshriver, line 148: If you're going to keep classes in the test module, their names should be more meaningful. Maybe `TestCapsuleSyncNewRepo` - or just drop the classes altogether.

mshriver (commented):
I'd like to discuss the use of `TestClass` / `TestClass2` / `TestClass2` pattern in this module. Since its only keeping the original class and test naming pattern, I wouldn't consider it a blocking comment for the conversion.

One outstanding question about the use of the `__cleanup` method through `request.addfinalizer` before the passed entities are actually instantiated.

tests/upgrades/test_capsule.py
JacobCallahan: My intuition leans heavily toward it not working, but I'd like to see it tested and verified that those entities are properly cleaned up.

tests/upgrades/test_capsule.py
JacobCallahan: I'm not convinced that this file needs classes, or at least not this many.

tests/upgrades/test_capsule.py
```diff
@@ -265,11 +252,11 @@ def test_post_user_scenario_capsule_sync_3(self):
...
         # Run a Capsule sync
-        capsule = entities.SmartProxy().search(query={'search': f'name={self.cap_host}'})[0]
-        call_entity_method_with_timeout(entities.Capsule(id=capsule.id).content_sync, timeout=1500)
+        capsule = entities.SmartProxy().search(query={'search': f'name={cap_host}'})[0]
+        call_entity_method_with_timeout(entities.Capsule(id=capsule.id).content_sync, timeout=3600)
         sync_status = entities.Capsule(id=capsule.id).content_get_sync()
-        self.assertEqual(len(sync_status['active_sync_tasks']), 0)
-        self.assertEqual(len(sync_status['last_failed_sync_tasks']), 0)
+        assert len(sync_status['active_sync_tasks']) == 0
+        assert len(sync_status['last_failed_sync_tasks']) == 0
```
JacobCallahan, line 262: non-blocking, but looks better imo
```suggestion
        assert sync_status['active_sync_tasks']
        assert sync_status['last_failed_sync_tasks']
```

JacobCallahan (changes requested):

tests/upgrades/test_capsule.py
devendra104: @mshriver @JacobCallahan   Agree, I will change the name and bind the two post-upgrade scenarios in a single class as there are no pre-upgrade test cases for that.
 
Class is required here, It binds the pre and post-upgrade test cases in a single scenario and that will help others to understand what this scenario does.

tests/upgrades/test_capsule.py
devendra104: @mshriver @JacobCallahan It works here, I have re-executed the test cases multiple times, and no extra cleanup has required for that. here we are calling the function and that is already defined, the order only matters when we call some method that is not defined.

In the below example if we call ```request.addfinalizer(org.delete)``` before the ```org``` definition then it will get failed because we don't have any ```org``` definition,  So in this case, we need to call the ```request.addfinalizer``` after the definition.

org = entities.Organization().search(query={'search': f'name={pre_test_name}_org'})[0]
request.addfinalizer(org.delete)

tests/upgrades/test_capsule.py
devendra104: fixed.

devendra104:
**Test Results:**

**Pre-Upgrade Test Result**
```
pytest -s -m pre_upgrade tests/upgrades/test_capsule.py::TestCapsuleSync::test_pre_user_scenario_capsule_sync
plugins: rerunfailures-9.1.1, ibutsu-1.12, services-2.2.1, forked-1.1.3, mock-3.3.1, xdist-1.34.0, cov-2.8.1
collected 1 item                     
.

================================================================================================ warnings summary =================================================================================================
==================================================================================== 1 passed, 2 warnings in 141.75s (0:02:21) ====================================================================================
```

**Post-Upgrade Test Results**

```
$ pytest -s -m post_upgrade tests/upgrades/test_capsule.py::TestCapsuleSync
plugins: rerunfailures-9.1.1, ibutsu-1.12, services-2.2.1, forked-1.1.3, mock-3.3.1, xdist-1.34.0, cov-2.8.1
collected 2 items / 1 deselected / 1 selected             
================================================================================================ warnings summary =================================================================================================
============================================================================= 1 passed, 1 deselected, 2 warnings in 236.46s (0:03:56) =============================================================================

$ pytest -s -m post_upgrade tests/upgrades/test_capsule.py::TestCapsuleSyncNewRepo
plugins: rerunfailures-9.1.1, ibutsu-1.12, services-2.2.1, forked-1.1.3, mock-3.3.1, xdist-1.34.0, cov-2.8.1
collected 2 items         
..
================================================================================================ warnings summary =================================================================================================
==================================================================================== 2 passed, 2 warnings in 758.69s (0:12:38) ====================================================================================

     

```

mshriver (approved):

PR #8150, opened 2020-11-09
Add workflow for updating robottelo image on Quay.

jameerpathan111:
What this does?
This workflow is mainly responsible for updating Robottelo image on Quay container registry.
It also performs code checks before pushing new image to registry.

How this works?
It uses [build-and-push-docker-images](https://github.com/marketplace/actions/build-and-push-docker-images) github action by Docker to build and push robottelo image to Quay.

I have separated codechecks workflow for push and pull_request event.  
It will be triggered whenever there is any change in robottelo repository.
There are two jobs as follows:
1. codechecks
2. robottelo_container: This depends on codechecks job. It will be responsible for pushing new image to Quay.

I have also added a status badge to display build status of workflow.

Test run for this change:
https://github.com/jameerpathan111/robottelo/actions/runs/357673664

Note:
For this to work, we first need to create following [github secrets](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets).
- REGISTRY_SERVER:quay.io
- REGISTRY_USERNAME:satqe-username
- REGISTRY_PASSWORD:satqe-password
- REGISTRY_NAMESPACE:satqe-something

### DIFF
--- a/.github/workflows/update_robottelo_image.yml
+++ b/.github/workflows/update_robottelo_image.yml
@@ -1,11 +1,6 @@
-name: Check Code Quality and Update Robottelo container image on Docker Hub.
+name: update_robottelo_image
 
-on:
-  push:
-    branches: master
-
-env:
-    PYCURL_SSL_LIBRARY: gnutls
+on: [push]
 
 jobs:
   codechecks:
@@ -14,6 +9,9 @@ jobs:
     strategy:
       matrix:
         python-version: [3.8, 3.9]
+    env:
+      PYCURL_SSL_LIBRARY: gnutls
+      SATELLITE_VERSION: 6.9
     steps:
       - name: Checkout Robottelo
         uses: actions/checkout@v2
@@ -31,39 +29,43 @@ jobs:
           cp robottelo.properties.sample robottelo.properties
           cp broker_settings.yaml.example broker_settings.yaml
           cp virtwho.properties.sample virtwho.properties
+
       - name: Pre Commit Checks
         uses: pre-commit/action@v2.0.0
 
       - name: Collect Tests
         run: |
-          pytest --collect-only tests/robottelo tests/foreman
-          pytest --collect-only tests/upgrades -m pre_upgrade
-          pytest --collect-only tests/upgrades -m post_upgrade
-      - name: Test Robottelo Coverage
-        run: pytest --cov --cov-config=.coveragerc --cov-report=xml tests/robottelo
+          pytest -n 8 --setup-plan --disable-pytest-warnings tests/foreman/ tests/robottelo/
+          pytest -n 8 --setup-plan --disable-pytest-warnings -m pre_upgrade tests/upgrades/
+          pytest -n 8 --setup-plan --disable-pytest-warnings -m post_upgrade tests/upgrades/
 
       - name: Make Docs
         run: |
           make test-docstrings
           make docs
+
       - name: Analysis (git diff)
         if: failure()
         run: git diff
 
-      - name: Upload Codecov Coverage
-        uses: codecov/codecov-action@v1.0.13
-        with:
-          file: coverage.xml
-          name: ${{ github.run_id }}-py-${{ matrix.python-version }}
-
   robottelo_container:
     needs: codechecks
-    name: Update Robottelo container image on Docker Hub.
+    name: Update Robottelo container image on Quay.
     runs-on: ubuntu-latest
 
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+
+      - name: Get image tag
+        id: image_tag
+        run: |
+          echo -n ::set-output name=IMAGE_TAG::
+          TAG="${GITHUB_REF##*/}"
+          if [ "${TAG}" == "master" ]; then
+              TAG="latest"
+          fi
+          echo "${TAG}"
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1
@@ -71,16 +73,17 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
 
-      - name: Login to DockerHub
+      - name: Login to Quay Container Registry
         uses: docker/login-action@v1
         with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
+          registry: ${{ secrets.QUAY_SERVER }}
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
 
-      - name: Build and push
+      - name: Build and push image to Quay
         uses: docker/build-push-action@v2
         with:
           context: .
           file: ./Dockerfile
           push: true
-          tags: satelliteqe/robottelo:latest
+          tags: ${{ secrets.QUAY_SERVER }}/${{ secrets.QUAY_NAMESPACE }}/robottelo:${{ steps.image_tag.outputs.IMAGE_TAG }}
--- a/README.rst
+++ b/README.rst
@@ -16,6 +16,10 @@ Robottelo
 .. image:: https://readthedocs.org/projects/robottelo/badge/?version=latest
   :target: https://robottelo.readthedocs.io/en/latest/?badge=latest
 
+.. image:: https://github.com/SatelliteQE/robottelo/workflows/update_robottelo_image/badge.svg
+   :alt: update_robottelo_image build status on GitHub Actions
+   :target: https://github.com/SatelliteQE/robottelo/actions
+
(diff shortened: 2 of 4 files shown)


JacobCallahan (changes requested):
We can keep the Dockerhub setup as-is for now. How about you change this PR to be the one that introduces automated quay builds?

jameerpathan111:
@JacobCallahan made the required changes. Now it will work for Quay.

JacobCallahan (approved):

jameerpathan111:
@JacobCallahan someone needs to create those secrets before merging this pr.

jameerpathan111:
Do not merge till github secrets are created.

.github/workflows/update_robottelo_image.yml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Pre Commit Checks
+        uses: pre-commit/action@v2.0.0
+
+      - name: Collect Tests
+        run: |
+          pytest -n 8 --setup-plan --disable-pytest-warnings tests/foreman/ tests/robottelo/
+          pytest -n 8 --setup-plan --disable-pytest-warnings -m pre_upgrade tests/upgrades/
+          pytest -n 8 --setup-plan --disable-pytest-warnings -m post_upgrade tests/upgrades/
+
+      - name: Test Robottelo Coverage
```
jyejare, line 41: Do we really need this check for pushing the robottelo image to Quay ?

The code-coverage shouldnt have anything to do with image , I would remove it.

.github/workflows/update_robottelo_image.yml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Make Docs
+        run: |
+          make test-docstrings
+          make docs
+
+      - name: Analysis (git diff)
+        if: failure()
+        run: git diff
+
+      - name: Upload Codecov Coverage
```
jyejare, line 53: Same for this as above.

jyejare (approved):
ACK pending non blocking comments / questions.

.github/workflows/update_robottelo_image.yml
```diff
@@ -0,0 +1,97 @@
...
+          cp virtwho.properties.sample virtwho.properties
+
+      - name: Pre Commit Checks
+        uses: pre-commit/action@v2.0.0
+
+      - name: Collect Tests
+        run: |
+          pytest -n 8 --setup-plan --disable-pytest-warnings tests/foreman/ tests/robottelo/
+          pytest -n 8 --setup-plan --disable-pytest-warnings -m pre_upgrade tests/upgrades/
+          pytest -n 8 --setup-plan --disable-pytest-warnings -m post_upgrade tests/upgrades/
```
lpramuk, line 39: might be useful? or required? to `export SATELLITE_VERSION=6.9` to avoid ssh checks for version as there is no SAT instance involved. 

robottelo tests are decorated by BZs - the decorator needs to know SAT version for its own logic

lpramuk (changes requested):
Please export SATELLITE_VERSION

.github/workflows/update_robottelo_image.yml
jameerpathan111: @jyejare probably we don't need it. It's just that I have separated the workflow which we currently have, that's why you are seeing this here. If you want I can remove this from here but then should I keep the push event which I removed? 
Currently the main robottelo_container job depends on codechecks job, it's upto us if we want it this way.

@JacobCallahan any thoughts?

.github/workflows/update_robottelo_image.yml
mshriver: I'd actually like to fix this behavior in robottelo, so that it will default to some version automatically without needing to set this env var. If there's no host to connect to, just pick a version. Downside being we would have to periodically update it (maybe once a year)

.github/workflows/update_robottelo_image.yml
```diff
@@ -0,0 +1,97 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to Quay Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ${{ secrets.REGISTRY_SERVER }}
+          username: ${{ secrets.REGISTRY_USERNAME }}
+          password: ${{ secrets.REGISTRY_PASSWORD }}
```
JacobCallahan, line 89:
```suggestion
          registry: ${{ secrets.QUAY_SERVER }}
          username: ${{ secrets.QUAY_USERNAME }}
          password: ${{ secrets.QUAY_PASSWORD }}
```

JacobCallahan (changes requested):

.github/workflows/update_robottelo_image.yml
JacobCallahan: I agree. this is something I was initially thinking as well.

.github/workflows/update_robottelo_image.yml
jameerpathan111: updated.

.github/workflows/update_robottelo_image.yml
jameerpathan111: removed

.github/workflows/update_robottelo_image.yml
jameerpathan111: removed

JacobCallahan (approved):

.github/workflows/update_robottelo_image.yml
jameerpathan111: Mike, so I shouldn't add it here, right?

mshriver: its fine to have it here for now

.github/workflows/update_robottelo_image.yml
```diff
@@ -0,0 +1,88 @@
+name: update_robottelo_image
+
+on: [push]
```
mshriver, line 3: Since we accept PRs to mutliple branches, I would be more specific here that this is on master branch only.

I think we're going to want separate image labels for the z-stream branches, and a separate workflow for them.

```
on:
  push:
    branches:
      - master
```

README.rst
```diff
@@ -16,6 +16,10 @@ Robottelo
 .. image:: https://readthedocs.org/projects/robottelo/badge/?version=latest
   :target: https://robottelo.readthedocs.io/en/latest/?badge=latest
 
+.. image:: https://github.com/SatelliteQE/robottelo/workflows/update_robottelo_image/badge.svg
```
mshriver, line 19: :100:

.github/workflows/update_robottelo_image.yml
mshriver: Actually I see you've handled this with the git ref setting the image tag, well done.

jameerpathan111: yes

.github/workflows/update_robottelo_image.yml
jameerpathan111: Added the SATELLITE_VERSION env variable [here](https://github.com/SatelliteQE/robottelo/pull/8150/files#diff-ee633393f1eca923b9e701f896601dcb1a7a1cd0c55238b8c0bb37c076459898R14).

README.rst
jameerpathan111: :)

.github/workflows/update_robottelo_image.yml
```diff
@@ -0,0 +1,89 @@
...
+          cp robottelo.properties.sample robottelo.properties
+          cp broker_settings.yaml.example broker_settings.yaml
+          cp virtwho.properties.sample virtwho.properties
+
+      - name: Pre Commit Checks
+        uses: pre-commit/action@v2.0.0
+
+      - name: Collect Tests
+        run: |
+          pytest -n 8 --setup-plan --disable-pytest-warnings tests/foreman/ tests/robottelo/
```
mshriver, line 38: Two things that aren't blockers but I think merit discussion.

1. I would not disable pytest warnings.  Currently we see many of those, but I'm working on fixing that, #8163  is part of such effort.  We should not be getting any pytest warnings, but people would have to come here to see them.

2. The output of the workflow is a bit difficult to parse with all of the output being in one expandable block.  Setup-plan is great for debugging when working with pytest and its fixtures, but I think the output is needlessly verbose here. Collection alone satisfies what we want to test and provides a clear list of everything collected.

mshriver (approved):

.github/workflows/update_robottelo_image.yml
jameerpathan111: @JacobCallahan @lpramuk @jyejare any thoughts?

lpramuk (approved):

jameerpathan111:
Thanks everyone :)

PR #8159, opened 2020-11-11
fixing the correct test type

omkarkhatavkar:
These are touching installer can cause an adverse effect on satellite can cause failures in other tests hence marking them as destructive


jyejare (approved):
Since these are not actually destructing the satellite and should be purely handled by try_ catch.

But after a conversation with @omkarkhatavkar , I understood that the tests are updating the satellite conf by interfering with installer and other tests are suffering if one of them fails.

I would ACK for now with leaving scope of improvement here in future.

jameerpathan111 (approved):
ACK

PR #8162, opened 2020-11-11
[6.9 PreSnap] Fix AzureRM cases, failing in Template search

jyejare:
Fixes 11 AzureRM test cases while searching for provisioning template to update it. 

Test Results for some Test cases as follows:
```
# Azure RM VM on_off_test:
== 1 passed, 1 deselected in 959.02s (0:15:59) ====

# test_positive_azurerm_host_provisioned:
===== 1 passed, 1 deselected in 517.16s (0:08:37) ==========

# pytest tests/foreman/api/test_computeresource_azurerm.py::TestAzureRM_UserData_Provisioning --only-failed
=============== 2 passed 504.97s (0:08:24) =======


# pytest tests/foreman/cli/test_computeresource_azurerm.py --only-failed (The error in one case below is a product bug):
======== 2 passed, 8 deselected, 1 error in 988.15s (0:16:28) ==========
```

### DIFF
--- a/robottelo/api/utils.py
+++ b/robottelo/api/utils.py
@@ -833,7 +833,7 @@ def update_provisioning_template(name=None, old=None, new=None):
     """
     temp = (
         entities.ProvisioningTemplate()
-        .search(query={'per_page': 1000, 'search': f'name="{name}"'})[0]
+        .search(query={'per_page': '1000', 'search': f'name="{name}"'})[0]
         .read()
     )
     if old in temp.template:


lpramuk (approved):

pondrejk (approved):

robottelo/api/utils.py
```diff
@@ -833,7 +833,7 @@ def update_provisioning_template(name=None, old=None, new=None):
     """
     temp = (
         entities.ProvisioningTemplate()
-        .search(query={'per_page': 1000, 'search': f'name="{name}"'})[0]
+        .search(query={'per_page': '1000', 'search': f'name="{name}"'})[0]
```
devendra104, line 836: Does "per_page" is required?

jyejare: @devendra104 Yes, very much since by default it searches in first 20 templates and then returns None. We have to either iterate over pages or use the higher number like 1000. 

Even though its 1000 the actual number of templates should be around 180, but it might get higher if any test adds the templates manually.

devendra104: Thanks, @jyejare for clarifying it, I think to fix the entries in per_page is not a good approach because count varies from time to time, if we use a loop then we can avoid the per_page related failure in the future.

jyejare: @devendra104 I know running through the loop would be more dynamic but the 1000 count is such high. I am sure that we ll never exceed that limit. :)

Also, lookup operation from Satellite API in 1000 entities is faster than nailgun looping search. I choose the faster way :)

PR #8163, opened 2020-11-11
Move component and importance marks to custom plugin

mshriver:
These were added during issue_handlers, and could not be used with -m in order to filter test selection
Apply marks automatically based on testimony tokens
Change to importance and component markers with the name as an arg
Add explicit CLI options to filter these, like --BZ

Tested with singular and multiple options for both arguments, as well as them in combination.

By my count the example below is correct, 29 tests in `tests/foreman/api` have component (ActivationKeys OR Search) AND importance (Critical OR Low).

```
(fix-caseimportance-marking) $ SAT_VERSION=6.8 pytest --collect-only --component ActivationKeys,Search --importance Critical,Low tests/foreman/api/
=========================================================== test session starts ===========================================================
rootdir: /home/setup/repos/robottelo, inifile: pytest.ini
plugins: forked-1.3.0, services-1.3.1, xdist-1.34.0, cov-2.10.1, mock-3.3.1, ibutsu-1.1.1
collecting 1261 items                                                                                                                     2020-11-12 03:04:29 - conftest - DEBUG - Collected 29 test cases
collected 1345 items / 1316 deselected / 29 selected                                                                                      
<Package /home/setup/repos/robottelo/tests/foreman/api>
  <Module test_activationkey.py>
    <UnitTestCase ActivationKeyTestCase>
      <TestCaseFunction test_negative_create_with_invalid_host_limit>
      <TestCaseFunction test_negative_create_with_invalid_name>
      <TestCaseFunction test_negative_create_with_no_host_limit>
      <TestCaseFunction test_negative_update_limit>
      <TestCaseFunction test_negative_update_max_hosts>
      <TestCaseFunction test_negative_update_name>
      <TestCaseFunction test_positive_add_future_subscription>
      <TestCaseFunction test_positive_add_host_collections>
      <TestCaseFunction test_positive_create_limited_hosts>
      <TestCaseFunction test_positive_create_unlimited_hosts>
      <TestCaseFunction test_positive_create_with_name>
      <TestCaseFunction test_positive_delete>
      <TestCaseFunction test_positive_fetch_product_content>
      <TestCaseFunction test_positive_remove_host_collection>
      <TestCaseFunction test_positive_update_auto_attach>
    <UnitTestCase ActivationKeySearchTestCase>
      <TestCaseFunction test_positive_search_by_org>
  <Module test_bookmarks.py>
    <UnitTestCase BookmarkTestCase>
      <TestCaseFunction test_negative_create_empty_query>
      <TestCaseFunction test_negative_create_null_public>
      <TestCaseFunction test_negative_create_same_name>
      <TestCaseFunction test_negative_create_with_invalid_name>
      <TestCaseFunction test_negative_update_empty_query>
      <TestCaseFunction test_negative_update_invalid_name>
      <TestCaseFunction test_negative_update_same_name>
      <TestCaseFunction test_positive_create_public>
      <TestCaseFunction test_positive_create_with_name>
      <TestCaseFunction test_positive_create_with_query>
      <TestCaseFunction test_positive_update_name>
      <TestCaseFunction test_positive_update_public>
      <TestCaseFunction test_positive_update_query>


```

### DIFF
--- a/docs/features/decorators.rst
+++ b/docs/features/decorators.rst
@@ -62,43 +62,6 @@ This decorator is used to avoid false failures when an feature is supported
 only on one os version. For example, ostree repository is available
 in RHEL7 but not in RHEL6.
 
-tier[n]
----------
-
-This set of decorators defines test levels:
-
-* ``tier1`` marks component level functional tests (that verify defined functional requirements using a range of normal and erroneous input data). Example::
-
-    import pytest
-
-    @pytest.mark.tier1
-    def test_positive_create_with_username(self):
-        """Create User for all variations of Username"""
-
-* ``tier2`` marks integration level functional tests and may include basic non-functional tests (security, performance regression, installation, compose validation). Example::
-
-    import pytest
-
-    @pytest.mark.tier2
-    def test_positive_view_cve(self):
-        """View CVE number(s) in Errata Details page"""
-
-* ``tier3`` marks system level tests::
-
-    import pytest
-
-    @pytest.mark.tier3
-    def test_positive_sync_with_enabled_notification(self):
-        """Receive email after every sync operation"""
-
-* ``tier4`` marks complex and long running tests. Example::
-
-    import pytest
-
-    @pytest.mark.tier4
-    def test_positive_upload_to_satellite(self):
-        """Perform end to end oscap test and upload reports"""
-
 skip_if_not_set
 ---------------
 
--- a/logging.conf
+++ b/logging.conf
@@ -1,8 +1,8 @@
 [loggers]
-keys=nailgun,root,robottelo,robozilla,robottelo_config
+keys=nailgun,root,robottelo,robozilla,robottelo_config,robottelo_collection
 
 [handlers]
-keys=consoleHandler,fileHandler
+keys=consoleHandler,fileHandler,fileHandlerCollection
 
 [formatters]
 keys=simpleFormatter
@@ -25,6 +25,12 @@ level=ERROR
 handlers=fileHandler
 qualname=robottelo.config
 
+[logger_robottelo_collection]
+# For test collection logging
+level=INFO
+handlers=fileHandlerCollection
+qualname=robottelo.collection
+
 [logger_robozilla]
 level=DEBUG
 handlers=fileHandler
@@ -41,6 +47,12 @@ class=FileHandler
 level=DEBUG
 formatter=simpleFormatter
 args=('robottelo.log', 'a')
+
+[handler_fileHandlerCollection]
+class=FileHandler
+level=DEBUG
+formatter=simpleFormatter
+args=('robottelo_collection.log', 'a')
 
 [formatter_simpleFormatter]
 format=%(asctime)s - %(name)s - %(levelname)s - %(message)s
--- a/pytest_plugins/issue_handlers.py
+++ b/pytest_plugins/issue_handlers.py
@@ -16,7 +16,7 @@ from robottelo.utils.issue_handlers import should_deselect
 from robottelo.utils.version import search_version_key
 from robottelo.utils.version import VersionEncoder
 
-LOGGER = logging.getLogger('robottelo')
+LOGGER = logging.getLogger('robottelo.collection')
 
 DEFAULT_BZ_CACHE_FILE = 'bz_cache.json'
 
--- a/pytest_plugins/issue_handlers.py
+++ b/pytest_plugins/issue_handlers.py
@@ -62,8 +62,8 @@ def pytest_collection_modifyitems(session, items, config):
     This collection includes pre-processed `is_open` status for each issue
 
     """
-    # generate_issue_collection will save a file, set by --bz-cache value
 
+    # generate_issue_collection will save a file, set by --bz-cache value
     pytest.issue_data = generate_issue_collection(items, config)
 
     # Add skipif markers, and modify collection based on --bz option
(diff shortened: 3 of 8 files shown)


docs/features/decorators.rst
```diff
@@ -34,71 +34,6 @@ Please note that ''stubbed'' is a decorator generator, and cannot be used as a
 "classic" Python decorator - it must be ''\@stubbed()'', not ''\@stubbed''
 (note the parenthesis).
 
-skip_if_os
```
mshriver, line 37: Not really related but cleaned them up as I saw them, I can remove this from the commit if we want to not include it in this PR.

pytest_plugins/testimony_markers.py
```diff
@@ -0,0 +1,106 @@
...
+)
+
+importance_regex = re.compile(
+    # To match :CaseImportance: Critical
+    r"\s*:CaseImportance:\s*(?P<importance>\S*)",
+    re.IGNORECASE,
+)
+
+
+@pytest.hookimpl(tryfirst=True)
```
jyejare, line 47: We have two plugins with tryfirst (rerun_rp) , which one will be used first ? As per the sequence in conftest list ?

pytest_plugins/testimony_markers.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # apply the marks for component and importance
+        # Find matches from docstrings starting at smallest scope
+        item_docstrings = [
+            d
+            for d in map(inspect.getdoc, (item.function, getattr(item, 'cls', None), item.module))
+            if d is not None
+        ]
+        for docstring in item_docstrings:
+            item_mark_names = [m.name for m in item.iter_markers()]
```
jyejare, line 70: I think this can be fetched above and outside of this for loop. Since the marks for item are not going to change for an item docstring. Saving the time a bit.

pytest_plugins/testimony_markers.py
```diff
@@ -0,0 +1,106 @@
...
+            # Filter test collection based on CLI options for filtering
+            # filters should be applied together
+            # such that --component Repository --importance Critical
+            # only collects tests which have both of these marks
+
+            # https://github.com/pytest-dev/pytest/issues/1373  Will make this way easier
+            # testimony requires both importance and component, this will blow up if its forgotten
+            importance_marker = item.get_closest_marker('importance').args[0]
+            if importance and importance_marker not in importance:
+                LOGGER.debug(
```
jyejare, line 89: I think the logfile will be overloaded with these logs ! Since it runs throught 2000 tests, so atlease 1000-1400 logs will be logged if we choose any importance :)

Lets avoid it ? Or overall choose a different logger instead of robottelo.

pytest_plugins/testimony_markers.py
```diff
@@ -0,0 +1,106 @@
...
+                    f'Deselected test {item.nodeid} due to "--importance {importance}",'
+                    f'test has importance mark: {importance_marker}'
+                )
+                deselected.append(item)
+                continue
+            component_marker = item.get_closest_marker('component').args[0]
+            if component and component_marker not in component:
+                LOGGER.debug(
+                    f'Deselected test {item.nodeid} due to "--component {component}",'
```
jyejare, line 98: Same here as above.

tests/foreman/api/test_activationkey.py
```diff
@@ -188,8 +184,6 @@ def test_positive_update_name(self):
 
         :expectedresults: Activation key is created, and its name can be
             updated.
-
-        :CaseImportance: High
```
jyejare, line 192: Since the `CaseImportance` is defined in a global scope, you are removing it from here ??

jyejare (commented):
ACK with minor improvement comments.

pytest_plugins/testimony_markers.py
```diff
@@ -0,0 +1,106 @@
...
+        help="Comma separated list of component names to include in collection",
+    )
+
+
+def pytest_configure(config):
+    """Register markers related to testimony tokens"""
+    for marker in [
+        'importance: CaseImportance testimony token, use --importance to filter',
+        'component: Component testimony token, use --component to filter',
+        # TODO: components read from testimony.yaml
```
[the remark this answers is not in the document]
mshriver: I'm actually going to remove this, as its no longer relevant. I started this with keeping the direct component name markers, then moved them to mark args.

pytest_plugins/testimony_markers.py
mshriver: There's no guarantee within pytest what order the hooks will be executed in, but these two hooks should be fine regardless of the order that they run in. There will be more `items` loop iterations in whichever runs first, and fewer `items` loop iterations in whichever runs second.

tests/foreman/api/test_activationkey.py
mshriver: Its defined at the module scope, L13

pytest_plugins/testimony_markers.py
mshriver: You know I went back and forth myself on including this log message, and where.

In the past I have run into problems with identifying the reason behind some test collection result because of a lack of logging when items were being deselected, which was my overall motivation behind including this.

I have some issues with how robottelo defaults to DEBUG logging, and how the pytest logger dumps everything to the console - with or without this particular message. This is a separate issue to address and resolve the behavior of.

I've added a new logger configuration for `robottelo.collection`, and have updated both this module and `pytest_plugins/issue_handlers.py`. Due to behavior of robottelo file/console handler, the messages will continue to be dumped to the console and robottelo.log. This is the exact same behavior as the existing `robottelo.config` log handler.

logging.conf
```diff
@@ -25,6 +25,12 @@ level=ERROR
 handlers=fileHandler
 qualname=robottelo.config
 
+[logger_robottelo_collection]
```
mshriver, line 28: @jyejare Check this out, added custom logger for collection issues.

docs/features/decorators.rst
lhellebr: It's fine to do it in this PR, but why are you doing it? What has made skip_if_os obsolete?

docs/features/decorators.rst
```diff
@@ -34,71 +34,6 @@ Please note that ''stubbed'' is a decorator generator, and cannot be used as a
...
-RHEL6.1.z version and so on.
-
-Arbitrary number versions can be passed as parameters. On second test both RHEL
-5 and 6 families would be skipped.
-
-This decorator is used to avoid false failures when an feature is supported
-only on one os version. For example, ostree repository is available
-in RHEL7 but not in RHEL6.
-
-tier[n]
```
lhellebr, line 65: I think we still use tier marks, why remove it?

pytest_plugins/issue_handlers.py
```diff
@@ -197,28 +197,29 @@ def generate_issue_collection(items, config):  # pragma: no cover
 
     # --- Build the issue marked usage collection ---
     for item in items:
-        component = None
-        bz_marks_to_add = []
-        importance = None
+        if item.nodeid.startswith('tests/robottelo/'):
+            # Unit test, no bz processing
+            # TODO: We don't need to be including component and importance in collected_data
```
lhellebr, line 202: I'm confused. Is this one TODO, or three different TODOs? And if we don't need to include component and importance, why do we?

lhellebr (approved):
Three comments/questions, otherwise LGTM.

pytest_plugins/issue_handlers.py
mshriver: Its one TODO, but couldn't fit on one line. I'll reformat and reword a bit.

mshriver: As far as why do we - couldn't say, but it shouldn't be necessary.  From what I can tell, there is similar behavior here between BZ markers and the testimony tokens, so the functionality was just added into the same function.

I would want to further verify that the issue handlers output data isn't somehow actually using importance or component before removing that data from the dictionary in this function, which is why I'm not including removing it in the PR.

docs/features/decorators.rst
mshriver: Good catch @lhellebr, skip_if_os was not part of the marks move, its in `decorators/host.py`, though listed here as if it were under `decorators/__init__.py`.

I'll update.

docs/features/decorators.rst
mshriver: They are no longer part of `decorators` module, and don't belong in this section of the docs.

Registered marks are seen with `pytest --markers`, and our mark registration should include a help text string for documentation.

jyejare (approved):

logging.conf
jyejare: Looks Good ! ACKed!

JacobCallahan (approved):
ACK

PR #8164, opened 2020-11-12
Remove two host-collection stubbed tests

swadeley:
There is no hammer command for adding and removing subscriptions
  to a host-collection.

The preferred workflow is to use activation keys to add subscriptions during provisioning. Although there is a ` hammer host subscription` command, now that we have SCA, it seems unlikely we will have a ` hammer host-collection subscription` command in the near future.

Thank you


omkarkhatavkar (approved):
ACK

mirekdlugosz (approved):
ACK

swadeley:
Thank you

PR #8169, opened 2020-11-12
Convert {api, cli}/test_product.py to pytest

tpapaioa:
This PR converts `tests/foreman/api/test_product.py` and `tests/foreman/cli/test_product.py` from unittest to pytest.

Test output:
```
# pytest tests/foreman/cli/test_product.py

[...]

collected 16 items                                  

tests/foreman/cli/test_product.py ................    [100%]

[...]

16 passed, 3 warnings in 468.39s (0:07:48)


# pytest tests/foreman/api/test_product.py

collected 67 items

tests/foreman/api/test_product.py ............................................s......................                                                                                                                                  [100%]

66 passed, 1 skipped, 3 warnings in 238.19s (0:03:58) 
```

### DIFF
--- a/tests/foreman/api/test_product.py
+++ b/tests/foreman/api/test_product.py
@@ -35,407 +35,424 @@ from robottelo.constants import VALID_GPG_KEY_FILE
 from robottelo.constants.repos import FAKE_1_PUPPET_REPO
 from robottelo.constants.repos import FAKE_1_YUM_REPO
 from robottelo.datafactory import invalid_values_list
+from robottelo.datafactory import parametrized
 from robottelo.datafactory import valid_data_list
 from robottelo.helpers import read_data_file
-from robottelo.test import APITestCase
 
 
-class ProductTestCase(APITestCase):
-    """Tests for ``katello/api/v2/products``."""
+@pytest.mark.tier1
+@pytest.mark.parametrize('name', **parametrized(valid_data_list()))
+def test_positive_create_with_name(request, name, module_org):
+    """Create a product providing different valid names
 
-    @classmethod
-    def setUpClass(cls):
-        """Set up organization for tests."""
-        super().setUpClass()
-        cls.org = entities.Organization().create()
+    :id: 3d873b73-6919-4fda-84df-0e26bdf0c1dc
 
-    @pytest.mark.tier1
-    def test_positive_create_with_name(self):
-        """Create a product providing different valid names
+    :parametrized: yes
 
-        :id: 3d873b73-6919-4fda-84df-0e26bdf0c1dc
+    :expectedresults: A product is created with expected name.
 
-        :expectedresults: A product is created with expected name.
+    :CaseImportance: Critical
+    """
+    product = entities.Product(name=name, organization=module_org).create()
+    assert product.name == name
 
-        :CaseImportance: Critical
-        """
-        for name in valid_data_list():
-            with self.subTest(name):
-                product = entities.Product(name=name, organization=self.org).create()
-                self.assertEqual(name, product.name)
 
-    @pytest.mark.tier1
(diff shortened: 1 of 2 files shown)


swadeley:
Hello

I see you are you adding cleanup functions. Normally we don't do that because the Satellite is created just for the test run, and any extra tasks just creates loads on the server and increases test run time.

tests/foreman/api/test_product.py
```diff
@@ -35,22 +36,87 @@
...
+    return http_proxy
 
-    @classmethod
-    def setUpClass(cls):
-        """Set up organization for tests."""
-        super().setUpClass()
-        cls.org = entities.Organization().create()
+
+@pytest.fixture(scope='module')
+def org(request):
```
mshriver, line 110: I see in a few of your PR's you have created such a structure for creating and deleting entities within fixtures.

There are some common entities already defined in `pytest_fixtures/api_fixtures.py`, which I think you should use here, keeping this conversion simpler.

As @swadeley has raised on this PR, there is a separate discussion to be had around which entities we should cleanup, and which should be left alone.

mshriver (changes requested):
The test conversions themselves are just fine, but the PR introduces a lot more concerning cleanup and entity fixtures. there are some common fixtures that you can start using now, and this PR can be simplified. Cleanup, when implemented, should be done in the common framework fixtures so that all modules benefit from consistent entity behavior.

tests/foreman/api/test_product.py
tpapaioa: Done.

tests/foreman/api/test_product.py
```diff
@@ -35,407 +35,410 @@
...
+@pytest.mark.parametrize('name', **parametrized(valid_data_list()))
+def test_positive_create_with_name(request, name, module_org):
+    """Create a product providing different valid names
 
-    @classmethod
-    def setUpClass(cls):
-        """Set up organization for tests."""
-        super().setUpClass()
-        cls.org = entities.Organization().create()
+    :id: 3d873b73-6919-4fda-84df-0e26bdf0c1dc
```
mshriver, line 48: You'll need to include a testimony token here, because the test is parametrized.

`:parametrized: yes`

I'm going to try to add this to testimony validation so its part of CI checks (or pre-commit).

tpapaioa: Done.

mshriver (approved):

tests/foreman/cli/test_product.py
```diff
@@ -36,274 +36,278 @@
...
+        result = ssh.command('hammer repository list')
+        assert default_product_name in "".join(result.stdout)
+
+        # verify that defaults setting should not affect other entities
+        product_list = Product.list({'organization-id': non_default_org['id']})
+        assert non_default_product_name == product_list[0]['name']
+        repository_list = Repository.list({'organization-id': non_default_org['id']})
+        assert non_default_product_name == repository_list[0]['product']
+
+    finally:
```
devendra104, line 219: This is just my suggestion, I think we can achieve it without using try and finally like below.
```
def __cleanup(org_id):
    Defaults.delete({'param-name': 'organization_id'})
    result = ssh.command('hammer defaults list')
    assert org_id in "".join(result.stdout) 
```
and call request.addfinalizer(lambda: __cleanup(org_id)) in the begining of the test case.

devendra104 (approved):
left one non-blocking suggestion.

tests/foreman/cli/test_product.py
tpapaioa: I left the original `try / finally` in place, because the `assert` after removing the default parameters should be part of the test.

devendra104 (approved):
ACK

PR #8172, opened 2020-11-13
Bump repo versions to 6.8

lhellebr:
There are still old versions of some repos - most notably Tools - being used.

Note that there were some even older versions than 6.7, e.g. Capsule 6.6 - I don't know why so I might break something fixing this, but I think it's more likely that someone just forgot to bump them before.

Tried one test that uses `rhst7`:

```
$ pytest tests/foreman/api/test_errata.py::ErrataTestCase::test_positive_install_in_hc
================================================ test session starts ================================================
[...]
===================================== 1 passed, 2 warnings in 555.63s (0:09:15) =====================================
```


robottelo/constants/__init__.py
```diff
@@ -480,8 +490,8 @@
         'key': 'rhva6',
     },
     'rhva65': {
-        'name': ('Red Hat Enterprise Virtualization Agents for RHEL 6 Server RPMs x86_64 6.5'),
-        'version': '6.5',
+        'name': ('Red Hat Enterprise Virtualization Agents for RHEL 6 Server RPMs x86_64 6.8'),
+        'version': '6.8',
```
jameerpathan111, line 494: @lhellebr is this right? I am guessing, `rhva65` is referring to rhel version here and not satellite tools. Need to check.

robottelo/constants/__init__.py
```diff
@@ -539,10 +549,10 @@
         'rhva6',
         'rhva65',
         'repo_name',
-        'Red Hat Enterprise Virtualization Agents for RHEL 6 Server RPMs x86_64 6.5',
+        'Red Hat Enterprise Virtualization Agents for RHEL 6 Server RPMs x86_64 6.8',
```
jameerpathan111, line 552: probably here as well. we should check.

robottelo/constants/__init__.py
mirekdlugosz: I think Jameer is right, this refers to RHEL version, not Satellite. Newest RHEL 6 is 6.10.

Also, variable name `rhva65` should probably be changed as well (along with everything that might use this name...).

lhellebr: Yes, this refers to RHEL version. Restored 6.5 and added entries for 6.10.

jameerpathan111 (approved):
LGTM

mshriver (approved):

PR #8177, opened 2020-11-13
Introducing Capsule and Satellite host classes

JacobCallahan:
These classes are an additional abstraction layer around the existing
ContentHost class.
These are meant for interacting with hosts provided by Broker.
Usage is similar to the current ContentHost class. However,
the Satellite class also wraps nailgun, airgun, and robottelo's hammer
library under self.api, self.ui_session, and self.cli respectively.

### DIFF
--- a/pytest_fixtures/broker.py
+++ b/pytest_fixtures/broker.py
@@ -2,7 +2,9 @@ import pytest
 from broker.broker import VMBroker
 
 from robottelo.constants import BROKER_RHEL77
+from robottelo.hosts import Capsule
 from robottelo.hosts import ContentHost
+from robottelo.hosts import Satellite
 
 
 @pytest.fixture
@@ -39,3 +41,17 @@ def rhel77_contenthost_class(request):
     with VMBroker(host_classes={'host': ContentHost}, **BROKER_RHEL77) as host:
         request.cls.content_host = host
         yield
+
+
+@pytest.fixture
+def satellite_latest():
+    """A fixture that provides a latest Satellite"""
+    with VMBroker(host_classes={'host': Satellite}, workflow='deploy-sat-lite') as sat:
+        yield sat
+
+
+@pytest.fixture
+def capsule_latest():
+    """A fixture that provides an unconfigured latest Capsule"""
+    with VMBroker(host_classes={'host': Capsule}, workflow='deploy-sat-capsule') as cap:
+        yield cap
--- a/robottelo/helpers.py
+++ b/robottelo/helpers.py
@@ -232,7 +232,7 @@ def read_data_file(filename):
         return file_contents.read()
 
 
-def install_katello_ca(hostname=None):
+def install_katello_ca(hostname=None, sat_hostname=None):
     """Downloads and installs katello-ca rpm
 
     :param str hostname: Hostname or IP address of the remote host. If
@@ -241,10 +241,15 @@ def install_katello_ca(hostname=None):
     :raises: AssertionError: If katello-ca wasn't installed.
 
     """
-    ssh.command(f'rpm -Uvh {settings.server.get_cert_rpm_url()}', hostname)
+    if sat_hostname:
+        cert_rpm_url = f'http://{sat_hostname}/pub/katello-ca-consumer-latest.noarch.rpm'
+    else:
+        sat_hostname = settings.server.hostname
+        cert_rpm_url = settings.server.get_cert_rpm_url()
+    ssh.command(f'rpm -Uvh {cert_rpm_url}', hostname)
     # Not checking the return_code here, as rpm could be installed before
     # and installation may fail
-    result = ssh.command(f'rpm -q katello-ca-consumer-{settings.server.hostname}', hostname)
+    result = ssh.command(f'rpm -q katello-ca-consumer-{sat_hostname}', hostname)
     # Checking the return_code here to verify katello-ca rpm is actually
     # present in the system
     if result.return_code != 0:
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -1,4 +1,6 @@
 import logging
+from functools import cached_property
+from pathlib import Path
 from urllib.parse import urljoin
 from urllib.parse import urlunsplit
 
@@ -12,6 +14,32 @@ from robottelo.helpers import install_katello_ca
 from robottelo.helpers import remove_katello_ca
 
 logger = logging.getLogger('robottelo')
+
+
+def setup_capsule(satellite, capsule, registration_args=None, installation_args=None):
+    """Given satellite and capsule instances, run the commands needed to set up the capsule
+
+    Note: This does not perform content setup actions on the Satellite
+
+    :param satellite: An instance of this module's Satellite class
+    :param capsule: An instance of this module's Capsule class
+    :param registration_args: A dictionary mapping argument: value pairs for registration
+    :param installation_args: A dictionary mapping argument: value pairs for installation
+    :return: An ssh2-python result object for the installation command.
+
+    """
+    if not registration_args:
+        registration_args = {}
+    file, cmd_args = satellite.capsule_certs_generate(capsule)
+    if installation_args:
+        cmd_args.update(installation_args)
+    satellite.execute(
+        f'sshpass -p "{capsule.password}" scp -o "StrictHostKeyChecking no" '
+        f'{file} root@{capsule.hostname}:{file}'
+    )
+    capsule.install_katello_ca(sat_hostname=satellite.hostname)
+    capsule.register_contenthost(**registration_args)
+    return capsule.install(**cmd_args)
 
 
 class ContentHostError(Exception):
@@ -101,19 +129,19 @@ class ContentHost(Host):
         if not result.status:
             raise ContentHostError('Failed to install katello-host-tools')
 
-    def install_katello_ca(self):
+    def install_katello_ca(self, sat_hostname=None):
         """Downloads and installs katello-ca rpm on the broker virtual machine.
 
         Uses common helper `install_katello_ca(hostname=None)`, but passes
         `self.hostname` instead of the hostname as we are using fake hostnames
-        forbroker virtual machines.
+        for broker virtual machines.
 
         :return: None.
         :raises robottelo.hosts.ContentHostError: If katello-ca wasn't
             installed.
         """
         try:
-            install_katello_ca(hostname=self.hostname)
+            install_katello_ca(hostname=self.hostname, sat_hostname=sat_hostname)
         except AssertionError:
             raise ContentHostError('Failed to download and install the katello-ca rpm')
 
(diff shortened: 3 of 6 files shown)


pytest_fixtures/broker.py
```diff
@@ -39,3 +39,25 @@ def rhel77_contenthost_class(request):
...
         yield
+
+
+@pytest.fixture
+def satellite68_snap20():
+    """A fixture that provides a 6.8 Snap 20 Satellite"""
+    with VMBroker(
+        host_classes={'host': Satellite},
+        workflow='deploy-sat-lite',
+        tempalte='tpl-sat-lite-6.8.0-20.0-rhel-7',
```
JacobCallahan, line 50: these are currently hard-coded to snap 20, but after some changes in satlab, we should be able to specify GA.

mshriver: typo in template.

I think these should be read from settings, not hard coded in the fixture.

robottelo/hosts.py
```diff
@@ -359,3 +366,121 @@ def patch_os_release_version(self, distro=DISTRO_RHEL7):
...
+
+
+class Satellite(Capsule):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._init_nailgun()
+        self._init_cli()
+        self._init_airgun()
+
+    def _init_nailgun(self):
```
mshriver, line 402: These could be cached_property decorated properties, supplying `self.api`, `self.cli`, and `self.ui_session`

pondrejk (commented):
Though I see nothing suspicious I don't feel I understand the intended usage thoroughly enough to say ack. Maybe for changes like this you could schedule some explanatory demo (apologies if something like that already happened). We could have a group Q&A there and then possibly conclude with a celebratory merge :)

JacobCallahan:
@pondrejk a demo was provided in one of our previous meetings. I can link the meeting recording for you if you'd like.

JacobCallahan:
I updated one of the existing destructive tests to use the new satellite fixture. Since it is currently skipped due to an open bz, I temporarily removed the skip and verified that the test reproduces the issue found in the bug [1]. Additionally, this all happens on the isolated satellite since the one in my config doesn't exist.
```
>           os_list_after_default = satellite69_latest.cli.OperatingSys.list()

tests/foreman/cli/test_operatingsystem.py:376: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
robottelo/cli/base.py:317: in list
    result = cls.execute(cls._construct_command(options), output_format=output_format)
robottelo/cli/base.py:253: in execute
    return cls._handle_response(response, ignore_stderr=ignore_stderr)
_ __ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <class 'robottelo.cli.operatingsys.OperatingSys'>
response = SSHCommandResult(stdout=b'', stderr='Association not found for organization\n', return_code=70, output_format='csv'), ignore_stderr = None
```
[1] - https://bugzilla.redhat.com/show_bug.cgi?id=1649011

swadeley (approved):
ACK

pytest_fixtures/broker.py
```diff
@@ -39,3 +41,17 @@ def rhel77_contenthost_class(request):
     with VMBroker(host_classes={'host': ContentHost}, **BROKER_RHEL77) as host:
         request.cls.content_host = host
         yield
+
+
+@pytest.fixture
+def satellite_latest():
+    """A fixture that provides a latest Satellite"""
+    with VMBroker(host_classes={'host': Satellite}, workflow='deploy-sat-lite') as sat:
```
mshriver, line 49: These workflow names should be pulled from configuration instead of hard coded. Not a blocker for this initial design, but moving forward we need to keep the broker provider dependencies decoupled from the testing framework.

pytest_fixtures/broker.py
```diff
@@ -39,3 +41,17 @@ def rhel77_contenthost_class(request):
...
+        yield sat
+
+
+@pytest.fixture
+def capsule_latest():
+    """A fixture that provides an unconfigured latest Capsule"""
+    with VMBroker(host_classes={'host': Capsule}, workflow='deploy-sat-lite') as sat:
```
mshriver, line 56: I think you mean `deploy-sat-capsule` here for the workflow.

robottelo/hosts.py
```diff
@@ -359,3 +366,121 @@ def patch_os_release_version(self, distro=DISTRO_RHEL7):
...
+            return DecClass
+
+        # set the server configuration to point to this satellite
+        self.nailgun_cfg = ServerConfig(
+            auth=(settings.server.admin_username, settings.server.admin_password),
+            url=f'https://{self.hostname}',
+            verify=False,
+        )
+        # add each nailgun entity to self.api, injecting our server config
+        self.api = lambda: None
```
mshriver, line 424: Not a fan of this pattern, as the class attribute is now a function object that we're going to call __getattr__ on.  Why not use an empty class, or AttrDict/Box type instance to store the entity classes?

robottelo/hosts.py
```diff
@@ -359,3 +366,121 @@ def patch_os_release_version(self, distro=DISTRO_RHEL7):
...
+                # not everything has an mro method, we don't care about them
+                pass
+
+    def _init_cli(self):
+        """Import all robottelo cli entities and wrap them under self.cli"""
+        import importlib
+        from robottelo.cli.base import Base
+
+        self.cli = lambda: None
+        for file in Path('robottelo/cli/').iterdir():
```
mshriver, line 439: Consider using entrypoints to explicitly define what classes should be loaded, instead of everything in the cli directory.

robottelo/hosts.py
```diff
@@ -14,6 +15,12 @@
 logger = logging.getLogger('robottelo')
 
 
+def setup_capsule(satellite, capsule):
```
mshriver, line 18: Why create this method that has to be imported and doesn't return the result, instead of directly calling `satellite.capsule_certs_generate` where needed?

robottelo/hosts.py
```diff
@@ -359,3 +366,121 @@ def patch_os_release_version(self, distro=DISTRO_RHEL7):
...
+                    return frame.function
+
+        self.ui_session = Session(
+            session_name=get_caller(),
+            user=settings.server.admin_username,
+            password=settings.server.admin_password,
+            hostname=self.hostname,
+        )
+
+    def capsule_certs_generate(self, capsule_fqdn, **extra_kwargs):
```
mshriver, line 470: Consider taking the capsule instance as an arg, and referencing the relevant attribute, instead of making the caller resolve the attribute.

pytest_fixtures/broker.py
```diff
@@ -39,3 +41,17 @@ def rhel77_contenthost_class(request):
     with VMBroker(host_classes={'host': ContentHost}, **BROKER_RHEL77) as host:
         request.cls.content_host = host
         yield
+
+
+@pytest.fixture
+def satellite_latest():
```
mshriver, line 47: This is problematic, as the test session may not be running against a satellite that matches the version that will be resolved by a broker call that doesn't specify a version.  Even if this isn't backported to the 6.8.z branch, it will become a problem with 6.9.z branch is created.

tests/foreman/cli/test_operatingsystem.py
```diff
@@ -358,31 +356,34 @@ def test_positive_update_parameters_attributes(self):
...
-            assert result.return_code == 0
-            assert DEFAULT_ORG not in "".join(result.stdout)
+    :BZ: 1649011
+
+    :expectedresults: os list should list operating systems when the
+        default organization is set
+    """
+    satellite_latest.api.OperatingSystem().create()
+    os_list_before_default = satellite_latest.cli.OperatingSys.list()
+    assert len(os_list_before_default) > 0
```
mshriver, line 374: > 0 is extraneous here, `len()` will always be positive, and 0 is falsy.

tests/foreman/cli/test_operatingsystem.py
```diff
@@ -358,31 +356,34 @@ def test_positive_update_parameters_attributes(self):
...
 
-    @pytest.mark.destructive
-    @pytest.mark.skip_if_open("BZ:1649011")
-    def test_positive_os_list_with_default_organization_set(self):
-        """list operating systems when the default organization is set
 
-        :id: 2c1ba416-a5d5-4031-b154-54794569a85b
+@pytest.mark.tier2
+@pytest.mark.skip_if_open("BZ:1649011")
+def test_positive_os_list_with_default_organization_set(satellite_latest):
```
mshriver, line 362: Loving this pattern of explicitly interacting with the satellite instance resolved for this test function!

robottelo/hosts.py
JacobCallahan: I don't believe that will work with the way we populate api and cli.

pytest_fixtures/broker.py
JacobCallahan: that's a good idea, but let's wait until after dynaconf phase 2, as i'd like to nest some of those.

pytest_fixtures/broker.py
JacobCallahan: ha yes!!

robottelo/hosts.py
JacobCallahan: This was a todo that shouldn't have made it into the pr.

pytest_fixtures/broker.py
JacobCallahan: For the time being, this is safe (works in non-Z runs). But we can revisit after dynaconf phase 2.

tests/foreman/cli/test_operatingsystem.py
JacobCallahan: yeah that's from the previous test code, but I can make it better. `assert os_list_before_default` should do fine.

robottelo/helpers.py
```diff
@@ -241,13 +241,18 @@ def install_katello_ca(hostname=None):
...
+        cert_rpm_url = settings.server.get_cert_rpm_url()
+    ssh.command(f'rpm -Uvh {cert_rpm_url}', hostname)
     # Not checking the return_code here, as rpm could be installed before
     # and installation may fail
-    result = ssh.command(f'rpm -q katello-ca-consumer-{settings.server.hostname}', hostname)
+    result = ssh.command(f'rpm -q katello-ca-consumer-{sat_hostname}', hostname)
     # Checking the return_code here to verify katello-ca rpm is actually
     # present in the system
-    if result.return_code != 0:
+    if getattr(result, "return_code", getattr(result, "status", 1)) != 0:
```
tpapaioa, line 255: Can you clarify what this change is doing? What is `result.status`? I don't see in `SSHCommandResult` where a `status` attribute is ever set.

robottelo/helpers.py
```diff
@@ -241,13 +241,18 @@ def install_katello_ca(hostname=None):
...
+        cert_rpm_url = settings.server.get_cert_rpm_url()
+    ssh.command(f'rpm -Uvh {cert_rpm_url}', hostname)
     # Not checking the return_code here, as rpm could be installed before
     # and installation may fail
-    result = ssh.command(f'rpm -q katello-ca-consumer-{settings.server.hostname}', hostname)
+    result = ssh.command(f'rpm -q katello-ca-consumer-{sat_hostname}', hostname)
     # Checking the return_code here to verify katello-ca rpm is actually
     # present in the system
-    if result.return_code != 0:
+    if getattr(result, 'return_code') != 0:
```
mshriver, line 255: Using `getattr` with no default will still raise AttributeError, why change this syntax?

mshriver (approved):

JacobCallahan:
@mshriver updated and ready!

PR #8180, opened 2020-11-18
Convert cli Host-Collection unittest to pytest

swadeley:
Hello

Converting CLI test_host_collection , unittest to pytest

### DIFF
--- a/tests/foreman/cli/test_host_collection.py
+++ b/tests/foreman/cli/test_host_collection.py
@@ -28,258 +28,241 @@ from robottelo.cli.lifecycleenvironment import LifecycleEnvironment
 from robottelo.constants import DEFAULT_CV
 from robottelo.constants import ENVIRONMENT
 from robottelo.datafactory import invalid_values_list
+from robottelo.datafactory import parametrized
 from robottelo.datafactory import valid_data_list
-from robottelo.test import CLITestCase
 
 
-class HostCollectionTestCase(CLITestCase):
-    """Tests for Host Collections via Hammer CLI"""
+def _make_fake_host_helper(module_org):
+    """Make a new fake host"""
+    library = LifecycleEnvironment.info({'organization-id': module_org.id, 'name': ENVIRONMENT})
+    default_cv = ContentView.info({'organization-id': module_org.id, 'name': DEFAULT_CV})
+    return make_fake_host(
+        {
+            'content-view-id': default_cv['id'],
+            'lifecycle-environment-id': library['id'],
+            'name': gen_string('alpha', 15),
+            'organization-id': module_org.id,
+        }
+    )
 
-    @classmethod
-    def setUpClass(cls):
-        """Prepare some data to be used in tests"""
-        super().setUpClass()
-        cls.organization = make_org()
-        cls.library = LifecycleEnvironment.info(
-            {'organization-id': cls.organization['id'], 'name': ENVIRONMENT}
-        )
-        cls.default_cv = ContentView.info(
-            {'organization-id': cls.organization['id'], 'name': DEFAULT_CV}
-        )
-        make_host_collection({'organization-id': cls.organization['id']})
 
-    def _make_fake_host_helper(self):
-        """Make a new fake host"""
-        return make_fake_host(
+@pytest.mark.upgrade
+@pytest.mark.tier2
+def test_positive_end_to_end(module_org):
+    """Check if host collection can be created with name and description,
+    content host can be added and removed, host collection can be listed,
+    updated and deleted
+
+    :id: 2d3b718e-6f57-4c83-aedb-15604cc8a4bd
+
+    :expectedresults: Host collection is created and has expected name and
+        description, content-host is added and removed, host collection is
+        updated and deleted.
+
+    :CaseImportance: Critical
+    """
+    name = list(valid_data_list().values())[0]
+    desc = list(valid_data_list().values())[0]
+    new_host_col = make_host_collection(
+        {'description': desc, 'name': name, 'organization-id': module_org.id}
+    )
+    assert new_host_col['name'] == name
+    assert new_host_col['description'] == desc
+
+    # add host
+    new_system = _make_fake_host_helper(module_org)
+    no_of_content_host = new_host_col['total-hosts']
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_host_collection.py
```diff
@@ -29,257 +30,256 @@
...
+        {
+            'content-view-id': default_cv['id'],
+            'lifecycle-environment-id': library['id'],
+            'name': gen_string('alpha', 15),
+            'organization-id': module_org.id,
+        }
+    )
+
+
+@pytest.mark.build_sanity
```
mshriver, line 58: What's the runtime of this testcase?

swadeley: robottelo]$ pytest  --show-capture=no tests/foreman/cli/test_host_collection.py -m build_sanity
=============================== test session starts ========================================
rootdir: /sat6/robottelo, inifile: pytest.ini
plugins: ibutsu-1.12, mock-3.3.1, forked-1.3.0, services-2.2.1, cov-2.10.1, xdist-1.34.0
collecting ... 2020-11-18 19:58:01 - conftest - DEBUG - Collected 17 test cases
collected 17 items / 16 deselected / 1 selected                                                                  

tests/foreman/cli/test_host_collection.py .                                                                [100%]

===== 1 passed, 16 deselected, 2 warnings in 156.95s (0:02:36) ======

mshriver: A little slow for what we're looking for, though on the edge of acceptable.

Would `test_positive_create_with_limit` be a sufficient option, to at least validate we can create HostCollections? If we can't update or delete a host collection, would that block us from testing the build, or require a re-build of a snap?

tests/foreman/cli/test_host_collection.py
```diff
@@ -28,258 +28,247 @@
...
-    def setUpClass(cls):
-        """Prepare some data to be used in tests"""
-        super().setUpClass()
-        cls.organization = make_org()
-        cls.library = LifecycleEnvironment.info(
-            {'organization-id': cls.organization['id'], 'name': ENVIRONMENT}
+@pytest.fixture(scope='module', autouse=True)
+def set_up_a_host_collection(module_org):
+    """ Make a host collection for use in tests"""
+    make_host_collection({'organization-id': module_org.id})
```
mshriver, line 38: Why is this fixture necessary for every test function? From what I can tell, every test is calling `make_host_collection` itself, and no tests are actually using or dependent on the host collection created here?

tests/foreman/cli/test_host_collection.py
```diff
@@ -28,258 +28,247 @@
...
+
+
+def _make_fake_host_helper(module_org):
```
mshriver, line 41: Since this is only dependent on the module_org, I believe it can be defined as a function scoped fixture. 

You can indirectly parametrize the 'count' of hosts, for the test case where you need two instead of one. The fixture can yield a list of host(s) for consistency. It could also delete the fake host after yielding, to clean up after itself.

This is an **optional** change - its borderline whether you'd call this 'setup' for the test. Its also questionable whether its worth the indirect parametrization instead of just calling the function in a loop when you need multiple hosts.

tests/foreman/cli/test_host_collection.py
swadeley: > A little slow for what we're looking for, though on the edge of acceptable.
> 
> Would `test_positive_create_with_limit` be a sufficient option, to at least validate we can create HostCollections? 
Yes, will move pytest mark to use that

>If we can't update or delete a host collection, would that block us from testing the build, or require a re-build of a snap?
I guess not

tests/foreman/cli/test_host_collection.py
swadeley: I see only the end to end test is using two hosts collections. I think there is no need for two, I will remove it.

tests/foreman/cli/test_host_collection.py
swadeley: @mshriver 

> Since this is only dependent on the module_org, I believe it can be defined as a function scoped fixture.
> 
That works OK for the end to end test.

I cannot call a fixture twice from one test and could not find an easy way around this.

> You can indirectly parametrize the 'count' of hosts, for the test case where you need two instead of one. 

By "indirectly parametrize" do you mean something like:
@pytest.mark.parametrize('fake_host', [2], indirect=['fake_host'])

The fixture can yield a list of host(s) for consistency. It could also delete the fake host after yielding, to clean up after itself.
> 
> This is an **optional** change - its borderline whether you'd call this 'setup' for the test. Its also questionable whether its worth the indirect parametrization instead of just calling the function in a loop when you need multiple hosts.

Only two tests in this module use the ` _make_fake_host_helper`,  so its not really "setup for tests"; we do not need it to be run in advance.

Not sure if this is worth it.

Thank you

mshriver (approved):

pondrejk (approved):

pondrejk:
Hi @swadeley, changes look good, could paste some test results just to be sure we can merge?

swadeley:
> Hi @swadeley, changes look good, could paste some test results just to be sure we can merge?

Hello @pondrejk , this is a problem with test result due to Nailgun defect; hosts are not being updated.
See https://github.com/SatelliteQE/nailgun/pull/777

swadeley:
Hello @pondrejk 

robottelo]$ git status
On branch convert_cli_host-collection_unittest_to_pytest

robottelo]$ pytest tests/foreman/cli/test_host_collection.py
========================= test session starts ===============================================
rootdir: /home/sjw/sat6/robottelo, inifile: pytest.ini
plugins: ibutsu-1.12, mock-3.3.1, forked-1.3.0, services-2.2.1, cov-2.10.1, xdist-1.34.0
collecting ... 2020-12-04 15:59:25 - conftest - DEBUG - Collected 17 test cases
collected 17 items                                                                                               

tests/foreman/cli/test_host_collection.py .................                                                [100%]

============== 17 passed, 3 warnings in 551.67s (0:09:11) =================

Thank you

swadeley:
Thank you

PR #8183, opened 2020-11-19
Update urllib3 to 1.26.2

mshriver:
requests 2.25 has a minimum version requirement that results in a warning
urllib3 1.25 won't get automatically force-upgraded, so I'm adding a specific dependency to clear the warning


pondrejk (approved):

mshriver:
I think a better strategy for addressing these warnings would be to use the eager upgrade strategy for pip, instead of the default, which is what triggers this behavior.

PR #8187, opened 2020-11-20
Fixes failing scap tests and adds realm cleanup

jameerpathan111:
This PR fixes tests failing because of Ansible variables not present. And adds realm cleanup so that capsule created can be deleted without any error.
For more info, see Satellite 6.9 tier1 build 8(snap 2)
```
$ pytest tests/foreman/ui/test_oscappolicy.py 
Test results:
collected 2 items

tests/foreman/ui/test_oscappolicy.py ..                                  [100%]
================== 2 passed, 3 warnings in 545.06s (0:09:05) ===================
```

### DIFF
--- a/tests/foreman/cli/test_realm.py
+++ b/tests/foreman/cli/test_realm.py
@@ -19,6 +19,8 @@ import random
 import pytest
 from fauxfactory import gen_string
 
+from robottelo.cleanup import capsule_cleanup
+from robottelo.cleanup import realm_cleanup
 from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.factory import CLIFactoryError
 from robottelo.cli.factory import make_proxy
@@ -27,10 +29,11 @@ from robottelo.cli.realm import Realm
 
 
 @pytest.fixture(scope='module')
-def _make_proxy(options=None):
+def _make_proxy(request):
     """Create a Proxy and register the cleanup function"""
-    proxy = make_proxy(options=options)
-    yield proxy
+    proxy = make_proxy()
+    request.addfinalizer(lambda: capsule_cleanup(proxy['id']))
+    return proxy
 
 
 @pytest.mark.run_in_one_thread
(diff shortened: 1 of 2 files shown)


pondrejk (approved):

tests/foreman/cli/test_realm.py
```diff
@@ -35,6 +36,14 @@ def _make_proxy(options=None):
     capsule_cleanup(proxy['id'])
 
 
+@pytest.fixture(scope='function')
+def _make_realm(options=None):
```
mshriver, line 40: This is not a proper pytest fixture pattern - the fixture does not take arguments, and is not called directly by the test.

This is calling `make_realm` with no options, and yielding the entity instance returned by that function.  The test function is then calling `_make_realm` again, with options passed this time, creating the realm entity twice.

Parameters in the function signature are themselves other fixture names, that pytest will resolve and execute.

If you want to add realm cleanup, you can do so with `request.addfinalizer(realm_cleanup(realm['id'])` within the test functions, after the realm has been created.  Include the `request` fixture in the test function parameters.

There is a valid fixture use pattern where the fixture returns a function object, which can then be called within the test, but that's not necessary here and is best reserved for more complex usage.

mshriver (changes requested):
The fixture pattern for realm cleanup isn't quite right, and `request.addfinalizer()` is a more appropriate solution for the desired effect in the tests.

tests/foreman/cli/test_realm.py
jameerpathan111: updated the pr.

jameerpathan111:
@mshriver updated the pr.

jameerpathan111:
Somehow I missed this discussion: https://github.com/SatelliteQE/robottelo/pull/8096#issuecomment-734247014
I have re-based the pr and added back the capsule cleanup part. It's required because this will cause failure in other tests, in my case it causes failure in tailoring files tests.

jyejare (approved):

tests/foreman/cli/test_realm.py
```diff
@@ -82,12 +85,13 @@ def test_positive_realm_info_name(self, _make_proxy):
                 'locations': proxy['locations'],
             }
         )
+        request.addfinalizer(realm_cleanup(realm['id']))
```
mshriver, line 88: This is actually executing the `realm_cleanup` method before passing the result of that method execution to `request.addfinalizer`. When using `request.addfinalizer`, you need to wrap the function call in a lambda.

mshriver (approved):

PR #8188, opened 2020-11-20
Fixed micro(or .z) version and zStream Rerun, also removed older way of version retrieval 

jyejare:
The version property of a server does not retrieves the `micro version` of a satellite version and hence the retrieval of zStream launches were not possible ! This PR fixes it !

### DIFF
--- a/robottelo/host_info.py
+++ b/robottelo/host_info.py
@@ -64,7 +64,7 @@ def _extract_sat_version(ssh_cmd):
     ssh_result = ssh.command(ssh_cmd)
     if ssh_result.stdout:
         version_description = ssh_result.stdout[0]
-        version_re = r'[^\d]*(?P<version>\d(\.\d){1})'
+        version_re = r'[^\d]*(?P<version>\d(\.\d){2})'
         result = re.search(version_re, version_description)
         if result:
             host_sat_version = result.group('version')
--- a/robottelo/host_info.py
+++ b/robottelo/host_info.py
@@ -31,9 +31,7 @@ def get_host_os_version():
     return 'Not Available'
 
 
-_SAT_6_2_VERSION_COMMAND = 'rpm -q satellite'
-
-_SAT_6_1_VERSION_COMMAND = 'grep "VERSION" /usr/share/foreman/lib/satellite/version.rb'
+_SAT_6_VERSION_COMMAND = 'rpm -q satellite'
 
 
 @functools.lru_cache(maxsize=1)
@@ -42,14 +40,10 @@ def get_host_sat_version():
     :return: Satellite version
     :rtype: version
     """
-    commands = (
-        _extract_sat_version(c) for c in (_SAT_6_2_VERSION_COMMAND, _SAT_6_1_VERSION_COMMAND)
-    )
-    for version, ssh_result in commands:
-        if version != 'Not Available':
-            LOGGER.debug('Host Satellite version: {}'.format(version))
-            return version
-
+    version, ssh_result = _extract_sat_version(_SAT_6_VERSION_COMMAND)
+    if version != 'Not Available':
+        LOGGER.debug('Host Satellite version: {}'.format(version))
+        return version
     LOGGER.warning('Host Satellite version not available: {!r}'.format(ssh_result))
     return version
 
--- a/pytest_plugins/rerun_rp/rerun_rp.py
+++ b/pytest_plugins/rerun_rp/rerun_rp.py
@@ -119,7 +119,7 @@ def pytest_collection_modifyitems(items, config):
         return
     rp = ReportPortal()
     version = settings.server.version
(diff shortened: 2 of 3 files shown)


robottelo/host_info.py
```diff
@@ -64,7 +64,7 @@ def _extract_sat_version(ssh_cmd):
     ssh_result = ssh.command(ssh_cmd)
     if ssh_result.stdout:
         version_description = ssh_result.stdout[0]
-        version_re = r'[^\d]*(?P<version>\d(\.\d){1})'
+        version_re = r'[^\d]*(?P<version>\d(\.\d){2})'
```
mirekdlugosz, line 67: That code is executed only if `SATELLITE_VERSION` environmental variable is not set, and AFAIK it will always be set on current Jenkins pipeline. What's worse - current Jenkins pipeline assume `SATELLITE_VERSION` to be `major.minor`, without z-stream number included.

This effectively causes difference in version content depending on environment robottelo is run on. version will usually include z-stream number locally, but will never include it on Jenkins. I'm certain this difference is going to bite us one day.

But I think you don't really need this, as change in `pytest_plugins/rerun_rp/rerun_rp.py` already discards z-stream number?

mshriver: @mirekdlugosz great point, and a pattern that I believe we are eliminating in the new CI pipelines. @lpramuk Thoughts on this behavior in the new CI pipelines - The current open MR from @jawatts is not setting SATELLITE_VERSION env var.

Overall I think the version pulled from the target satellite should override the envvar, because that's the _real_ version that we're running against. SATELLITE_VERSION should only be used if there's no target satellite, or because it can't be reached.

jyejare: @mirekdlugosz If the Jenkins pipeline assumes the Satellite Version as major.minor then should either be fixed by considering the `micro` appending to it or Jenkins chops out micro if its being used from this function. And I am sure the jenkins pipeline isnt fetching the version from this function. Its the static satellite version there.

As far as this PR concerns, I checked the usage of this function all over robottelo and it doesn't seems to bite anywhere in robottelo. Also, the rerun_rp isnt discarding the `z-stream`, I am removing `version.epoch` as it always returns `0` for zstream and ystream satellite.

jyejare: @mshriver Good point out there. We should use actual extracted version from Satellite. Also, Whats the way @jawatts / @lpramuk is using the to extract sat version from sat ?

jyejare: Overall, This PR helps us in unblocking zStream rerun and hence we have to weigh this PR with Jenkins changes and get it in.

mshriver: @jyejare right now the pipelines aren't pulling satellite version from running instances, has no need to. Its receiving specific UMB messages with x.y.z versions to trigger jobs.  

That being said, I think we should start to be more specific with the version read from running Satellites, and adjust pipelines and run environments as needed.

The pipelines are build for X.Y versions, so there are going to be points where we'll have to deal with some version processing. @jyejare I trust you'll be accountable to dealing with those changes, and updating the team on venv version setting.

mshriver:
@jyejare you may want to separate the use of epoch in rerun_rp plugin and the change to reading z-stream version in the host version helper.

Please correct CI failure, and rebase + push to trigger a run of the new github pipeline.

robottelo/host_info.py
jyejare: @mshriver As per discussion on GMeet which in short describes that, 'We will be shortly moving to new Pipelines where we dont sets SATELLITE_VERSION and the plan is to read the satellite version as 6.y.z. And you have ACKed that this PR is on track of that.'.

As to prove this PR doesn't breaks the existing pipeline, I have ran Standalone Automation Job (#2097 with Tier1 tests)  and I dint see any unusual failure caused due to this PR. 

Also, I have removed the Satellite 6.1 way of retrieving Satellite Version from sat by looking into version.rb file as its no longer required in recent releases (rpm -q satellite is enough). Also the version.rb file is removed from recent releases and I am not aware any other file which can be replaced here.

The Travis in now fixed.

jyejare:
@mshriver @mirekdlugosz Please revisit, long waiting PR
 !

mshriver (approved):

mshriver:
@jyejare Approved, not that long waiting, you pushed new commits only a few hours ago.

mshriver (approved):
@jyejare I think you should bring this change to `master` branch as well.

jyejare:
@mshriver yes, once its merged, I ll cherrypick.

jyejare:
@mirekdlugosz Looking for your input / second review.

jyejare:
@SatelliteQE/tier-2-reviewers Please review this long pending PR.

lhellebr (approved):

PR #8189, opened 2020-11-20
Pytest conversion for {api, cli}/test_operatingsystem.py

tpapaioa:
This PR converts `tests/foreman/api/test_operatingsystem.py` and `tests/foreman/cli/test_operatingsystem.py` from unittest to pytest. The fixture `module_org` is used, and subtests are converted to parametrized tests.

```
# pytest tests/foreman/api/test_operatingsystem.py tests/foreman/cli/test_operatingsystem.py

[...]

collected 154 items

tests/foreman/api/test_operatingsystem.py ...........s.................. [ 19%]
...................................................................      [ 62%]
tests/foreman/cli/test_operatingsystem.py .............................. [ 82%]
..........................s                                              [100%]

[...]

============ 152 passed, 2 skipped, 3 warnings in 944.71s (0:15:44) ============
```

### DIFF
--- a/tests/foreman/api/test_operatingsystem.py
+++ b/tests/foreman/api/test_operatingsystem.py
@@ -30,12 +30,12 @@ from requests.exceptions import HTTPError
 
 from robottelo.constants import OPERATING_SYSTEMS
 from robottelo.datafactory import invalid_values_list
+from robottelo.datafactory import parametrized
 from robottelo.datafactory import valid_data_list
-from robottelo.test import APITestCase
 from robottelo.utils.issue_handlers import is_open
 
 
-class OperatingSystemParameterTestCase(APITestCase):
+class TestOperatingSystemParameter:
     """Tests for operating system parameters."""
 
     @pytest.mark.tier1
@@ -52,7 +52,7 @@ class OperatingSystemParameterTestCase(APITestCase):
         # Check whether OS 1 exists.
         os1 = entities.OperatingSystem(id=1).read_raw()
         if os1.status_code == NOT_FOUND and entities.OperatingSystem().create().id != 1:
-            self.skipTest('Cannot execute test, as operating system 1 is not available.')
+            pytest.skip('Cannot execute test, as operating system 1 is not available.')
 
         # Create and read a parameter for operating system 1. The purpose of
         # this test is to make sure an HTTP 422 is not returned, but we're also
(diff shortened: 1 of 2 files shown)


tests/foreman/api/test_operatingsystem.py
```diff
@@ -63,21 +63,16 @@ def test_verify_bugzilla_1114640(self):
...
 
-    @classmethod
-    def setUpClass(cls):
-        """Create an organization which can be re-used in tests."""
-        super().setUpClass()
-        cls.org = entities.Organization().create()
-
     @pytest.mark.tier1
-    def test_positive_create_with_name(self):
+    @pytest.mark.parametrize('name', **parametrized(valid_data_list()))
```
mshriver, line 74: Please add `:parametrized: yes` to test function docblocks when they are parametrized.

mshriver (changes requested):
Testimony tokens are required for parametrization

tests/foreman/api/test_operatingsystem.py
tpapaioa: Done.

tests/foreman/api/test_operatingsystem.py
```diff
@@ -142,41 +137,43 @@ def test_positive_read_minor_version_as_string(self):
...
         """
         name = gen_string('utf8')
-        for desc in valid_data_list():
-            with self.subTest(desc):
-                os = entities.OperatingSystem(name=name, description=desc).create()
-                self.assertEqual(os.name, name)
-                self.assertEqual(os.description, desc)
+        os = entities.OperatingSystem(name=name, description=desc).create()
+        assert os.name == name
+        assert os.description == desc
```
devendra104, line 159: Add the cleanup.

tests/foreman/api/test_operatingsystem.py
```diff
@@ -63,55 +63,50 @@ def test_verify_bugzilla_1114640(self):
...
             name=name, operatingsystem=1, value=value
         ).create()
-        self.assertEqual(os_param.name, name)
-        self.assertEqual(os_param.value, value)
+        assert os_param.name == name
+        assert os_param.value == value
 
 
-class OperatingSystemTestCase(APITestCase):
+class TestOperatingSystem:
```
devendra104, line 70: I think class is not required.

tests/foreman/api/test_operatingsystem.py
```diff
@@ -63,55 +63,50 @@ def test_verify_bugzilla_1114640(self):
...
         :expectedresults: Operating system entity is created and has proper
             name
 
         :CaseImportance: Critical
         """
-        for name in valid_data_list():
-            with self.subTest(name):
-                os = entities.OperatingSystem(name=name).create()
-                self.assertEqual(os.name, name)
+        os = entities.OperatingSystem(name=name).create()
```
devendra104, line 87: Add the cleanup.

devendra104 (commented):
Left two common comments, one for cleanup and the other for class.

tests/foreman/api/test_operatingsystem.py
tpapaioa: There is currently no cleanup done, so I'm not adding any as part of the pytest conversion. Based on discussion with @mshriver I am holding off on adding any cleanup until the team has discussed the pros and cons involved.

tests/foreman/api/test_operatingsystem.py
tpapaioa: Same comment as above.

tests/foreman/api/test_operatingsystem.py
tpapaioa: There are currently two different classes in the same module. I'm keeping the classes in place for now, so that the tests are kept logically distinct.

devendra104: yes, you are right, but the same thing we can also achieve by putting the test details inside the test case doc.  This is just my suggestion, I don't think the class is actually required here.

devendra104 (approved):
left one non-blocking comment.

mshriver (approved):

PR #8194, opened 2020-11-24
setup_org_for_a_rh_repo fixed for OS repos

lhellebr:
This adds support for OS repos to be added as custom repos, without the need to upload manifest.


mirekdlugosz (approved):
Can you update docstring, so it mentions that this functions supports base RHEL repos as well?

Otherwise, ACK

jameerpathan111 (approved):
LGTM

jameerpathan111:
@mirekdlugosz Is there anything we need to do about reduced code coverage?

pondrejk (approved):

PR #8173, opened 2020-11-13
Test chost errata pagination

swadeley:
Hello

Closed loop test for:
[BZ#1662254 – Per page value does not work for Content Host Errata page on Satellite WebUI](https://bugzilla.redhat.com/show_bug.cgi?id=1662254)

[Bug 1846670 - Unable to apply all Errata via Remote Execution on Web UI with "Select All" ](https://bugzilla.redhat.com/show_bug.cgi?id=1846670)

This test depends on these PRs:
[ Replace yum3 repo #7 ](https://github.com/SatelliteQE/fedorapeople-repos/pull/7)
[ Add pagination support on CHost's errata tab #532 ](https://github.com/SatelliteQE/airgun/pull/532)
[ Add support to install all CHost's errata #535 ](https://github.com/SatelliteQE/airgun/pull/535)

Test to follow soon.

PR #8198, opened 2020-11-25
Downgrade wt.core due to regression with airgun

mshriver:
https://github.com/SatelliteQE/robottelo/pull/8173#issuecomment-731166431

https://github.com/SatelliteQE/airgun/issues/536

Once 536 has been resolved, upgrade to latest wt.core

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,7 +20,7 @@ requests==2.25.0
 testimony==2.1.0
 unittest2==1.1.0
 wait-for==1.1.4
-widgetastic.core==0.61
+widgetastic.core==0.51
 widgetastic.patternfly==1.3.2
 wrapanapi==3.5.7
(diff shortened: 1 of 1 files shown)


jameerpathan111 (approved):
ACK

swadeley (approved):
ACK

mirekdlugosz:
As mentioned in https://github.com/SatelliteQE/robottelo/pull/8173#issuecomment-731166431 , one of the problems of older widgetastic is that it uses API that has been removed in Python 3.9. So with that change, we effectively force everyone to be on Python 3.8 (you can't use lower version, because broker needs at least 3.8; you can't use higher version, because widgetastic needs at most 3.8).

I trust our team members can handle this little inconvenience, but I would like to confirm that Jenkins will not explode.

PR #8211, opened 2020-11-30
Update to RHBA-2020:4367 for rhst7

swadeley:
Hello

  This change is for  Satellite 6.8 Satellite Tools Release

OLD https://access.redhat.com/errata/RHBA-2020:1455.html
NEW https://access.redhat.com/errata/RHBA-2020:4367.html


lhellebr (commented):
This is only valid for 6.8 Tools so at first, I thought this should go directly to 6.8z branch.
On the other hand though, 6.8 Tools are sometimes used even when testing 6.9 snaps since 6.9 Tools are not yet in CDN. Therefore, I think this is ok.
Thoughts?

swadeley:
> This is only valid for 6.8 Tools so at first, I thought this should go directly to 6.8z branch.
> On the other hand though, 6.8 Tools are sometimes used even when testing 6.9 snaps since 6.9 Tools are not yet in CDN. Therefore, I think this is ok.
> Thoughts?
@lhellebr Correct, we use tools repo from a GA release in some tests of a pre-beta snap where we just want a small RHEL repo from the CDN. For example, where you want a properly subscribed host.

Thank you

lhellebr (approved):

mirekdlugosz (approved):
> This is only valid for 6.8 Tools so at first, I thought this should go directly to 6.8z branch.
> On the other hand though, 6.8 Tools are sometimes used even when testing 6.9 snaps since 6.9 Tools are not yet in CDN. Therefore, I think this is ok.
> Thoughts?

As far as I remember, Tools repo is created near GA, so there is no 6.9 tools repo to use during 6.9 development. Your options are 6.8 tools repo and 6.7 tools repo (and all earlier releases tools repos), out of which 6.8 seems to be best match.

ACK

swadeley:
Thank you

PR #8217, opened 2020-12-03
[Rerun RP Pytest Plugin Improvement] Rerun Upgrade Tests

jyejare:
Until now/today, The Rerun RP plugin was helpful to rerun regular tier tests, but with this improvement, one can also rerun upgrade failed/skipped tests on an upgraded setup.

The usage is simple and just adds one more option `upgrades-rerun` to this plugin.

Usage:
```
# pytest tests/foreman/ --only-failed --upgrades-rerun
```

### DIFF
--- a/pytest_plugins/rerun_rp/rerun_rp.py
+++ b/pytest_plugins/rerun_rp/rerun_rp.py
@@ -83,25 +83,28 @@ def pytest_addoption(parser):
 
         Defect_types:
             Collect failed tests marked with defect types: {[*ReportPortal.defect_types.keys()]}
-        '''
+    '''
     parser.addoption("--only-failed", const='all', nargs='?', help=help_text)
     help_text = '''
         Collects only skipped tests in Report Portal to run only skipped tests.
 
         Usage: --only-skipped
-        '''
+    '''
     parser.addoption("--only-skipped", action='store_true', default=False, help=help_text)
     help_text = '''
-            Collects user tests from report portal.
+        Collects user tests from report portal.
 
-            Usage: --user [value]
+        Usage: --user [value]
 
-            Value: [ Report_portal_username ]
-
-            Defect_types:
-                Collect tests of specific user from Report Portal
-        '''
+        Value: [ Report_portal_username ]
+    '''
     parser.addoption("--user", nargs='?', help=help_text)
+    help_text = '''
+        Rerun Upgrade launch tests instead of Regular Tier launch from Report Portal,
+
+        Usage: --upgrades-rerun
+    '''
+    parser.addoption("--upgrades-rerun", action='store_true', default=False, help=help_text)
 
 
 @pytest.hookimpl(tryfirst=True)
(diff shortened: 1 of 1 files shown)


pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -83,25 +83,28 @@ def pytest_addoption(parser):
...
-            Usage: --user [value]
+        Usage: --user [value]
 
-            Value: [ Report_portal_username ]
-
-            Defect_types:
-                Collect tests of specific user from Report Portal
-        '''
+        Value: [ Report_portal_username ]
+    '''
```
jyejare, line 100: This change is not related to the upgrade rerun improvement. This is just a nitpick. Just FYI to avoid the confusion.

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -83,25 +83,28 @@ def pytest_addoption(parser):
...
     parser.addoption("--user", nargs='?', help=help_text)
+    help_text = '''
+        Rerun Upgrade launch tests instead of Regular Tier launch from Report Portal,
+
+        Usage: --upgrade-rerun
```
jameerpathan111, line 105:
```suggestion
        Usage: --upgrades-rerun
```
typo?

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -83,25 +83,28 @@ def pytest_addoption(parser):
...
+    '''
+    parser.addoption("--upgrades-rerun", action='store_true', default=False, help=help_text)
```
devendra104, line 107: "upgrade-tests-rerun" is more appropriate than "upgrades-rerun"

jyejare: @devendra104 As pytest and rerun plugin is used for running tests and this plugin specially plays with test collection, so I think I need not to explicitly add `test` word there.

So I am good with whats it now.

jyejare:
@jameerpathan111 @devendra104 Can you merge this! I have addressed your comments !

devendra104 (approved):
ACK

jameerpathan111 (approved):
ACK

jyejare:
@SatelliteQE/airgun-tier-2-reviewers VERY VERY LONG pending PR. Please review :)

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -113,13 +116,20 @@ def pytest_collection_modifyitems(items, config):
...
         return
     rp = ReportPortal()
     version = settings.server.version
     sat_version = f'{version.base_version}.{version.epoch}'
     LOGGER.info(f'Fetching Report Portal launches for target Satellite version: {sat_version}')
-    launch = next(iter(rp.launches(sat_version=sat_version).values()))
+    launch = next(
+        iter(
+            rp.launches(
+                sat_version=sat_version, launch_type='upgrades' if upgrades_rerun else 'satellite6'
```
mshriver, line 129: This hardcoding (and use of) `'satellite6'` is something we need to move away from. The launch type should be pulled from report portal configuration, or directly passable.

mshriver (approved):

pytest_plugins/rerun_rp/rerun_rp.py
jyejare: @mshriver by configuration you mean from current robottelo.properties (future dynaconf) ...  that should be doable and need since we are moving to DynaConf.

jyejare: We have that already available in RP, but thats basically for Report Portal Module. For plugin I just wanted to make it easier by `--upgrades_rerun` option.

PR #8170, opened 2020-11-12
Fix test_api_endtoend, use deepdiff for comparison

mshriver:
Deepdiff was already installed as a sub-dependency, make it directly required
Use deepdiff to provide rich feedback in the test failure of what the difference is between the api path data structures

There appears to be a regression of https://github.com/Katello/katello-selinux/pull/24 that is causing candlepin_events service to fail to start. This causes the `EndToEndTestCase::test_positive_ping` testcase to fail on 6.9 snap 1

```
setup@localhost:~/repos/robottelo (fix-api-endtoend-test*%=) % pytest -v -m build_sanity tests/foreman/endtoend/test_api_endtoend.py
==================================================================================== test session starts =====================================================================================
platform linux -- Python 3.8.6, pytest-5.4.3, py-1.9.0, pluggy-0.13.1 -- /home/setup/repos/robottelo/.robottelo/bin/python3.8
cachedir: .pytest_cache
shared_function enabled - OFF - scope:  - storage: file
rootdir: /home/setup/repos/robottelo, inifile: pytest.ini
plugins: forked-1.3.0, services-1.3.1, xdist-1.34.0, cov-2.10.1, mock-3.3.1, ibutsu-1.1.1
collecting ... 2020-11-12 18:05:42 - conftest - DEBUG - Collected 7 test cases
collected 7 items / 1 deselected / 6 selected                                                                                                                                                

tests/foreman/endtoend/test_api_endtoend.py::AvailableURLsTestCase::test_positive_get_links PASSED                                                                                     [ 16%]
tests/foreman/endtoend/test_api_endtoend.py::AvailableURLsTestCase::test_positive_get_status_code PASSED                                                                               [ 33%]
tests/foreman/endtoend/test_api_endtoend.py::EndToEndTestCase::test_positive_find_admin_user PASSED                                                                                    [ 50%]
tests/foreman/endtoend/test_api_endtoend.py::EndToEndTestCase::test_positive_find_default_loc PASSED                                                                                   [ 66%]
tests/foreman/endtoend/test_api_endtoend.py::EndToEndTestCase::test_positive_find_default_org PASSED                                                                                   [ 83%]
tests/foreman/endtoend/test_api_endtoend.py::EndToEndTestCase::test_positive_ping FAILED                                                                                               [100%]

```

- bulk system_purpose: https://bugzilla.redhat.com/show_bug.cgi?id=1894159
- simple content access: https://bugzilla.redhat.com/show_bug.cgi?id=1890658
[...]

PR #8224, opened 2020-12-07
Remove AvailableURLsTestCase::test_positive_get_links from build_sanity

lpramuk:
We cannot to maintain the API URIs list prior new releases.
Removing from build_sanity (moving the mark from class to other test we want to include)

### DIFF
--- a/tests/foreman/endtoend/test_api_endtoend.py
+++ b/tests/foreman/endtoend/test_api_endtoend.py
@@ -895,7 +895,6 @@ API_PATHS = {
 }
 
 
-@pytest.mark.build_sanity
 class AvailableURLsTestCase(TestCase):
     """Tests for ``api/v2``."""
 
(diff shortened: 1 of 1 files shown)


mshriver:
I'm not sure I agree here, as the test should be straightforward to maintain.  If its not providing value when we get new snaps, when is the test providing value?

Given your comments in the description, I'd remove the test entirely, or keep it in build_sanity and rely on the test owner to upgrade the test when it fails because the API endpoints changed.

lpramuk:
This is going to fail whenever there is change in API.
Such fail has nothing to do with build sanity.

This test should be in regression testing, when you can check with Devs on the API changes,  expected? or not? ....

mshriver:
@lpramuk I could agree with that assessment of value in regression.

I fixed it for a short time with #8170, but the API continues to change with early release snaps.

mshriver (approved):

mirekdlugosz (approved):
LGTM

PR #8230, opened 2020-12-11
Do not delete as part of teardown

swadeley:
Unecessary deletions create load on Satellite
    Deleting org can cause race conditions

### DIFF
--- a/pytest_fixtures/api_fixtures.py
+++ b/pytest_fixtures/api_fixtures.py
@@ -518,8 +518,6 @@ def repo_setup():
     lce = entities.LifecycleEnvironment(organization=org).create()
     details = {'org': org, 'product': product, 'repo': repo, 'lce': lce}
     yield details
-    for property_name in ['lce', 'repo', 'product', 'org']:
-        details[property_name].delete()
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


devendra104 (commented):
@swadeley One fixture is used by the customer upgrade scenarios and another is used by the "run_in_one_thread" marker test case, I don't think it will create any load or race condition.

swadeley:
> @swadeley One fixture is used by the customer upgrade scenarios and another is used by the "run_in_one_thread" marker test case, I don't think it will create any load or race condition.

@devendra104 If the teardown is essential to your test then it should be part of your test. Any fixture in this `pytest_fixtures/api_fixtures.py file could be used by anyone in their test in the future, and they have a right to expect it is safe to use.

devendra104:
> > @swadeley One fixture is used by the customer upgrade scenarios and another is used by the "run_in_one_thread" marker test case, I don't think it will create any load or race condition.
> 
> @devendra104 If the teardown is essential to your test then it should be part of your test. Any fixture in this `pytest_fixtures/api_fixtures.py file could be used by anyone in their test in the future, and they have a right to expect it is safe to use.

I think it can be used by anyone and it will not create any load or race condition, If it is mandatory to use teardown type of fixtures in the test file then I will definitely move this inside the test file or create separate fixtures file.

mshriver (approved):

devendra104 (approved):
ACK

mshriver:
CI pipeline failing due to ubuntu requirement not available, unrelated to change here.

swadeley:
Thank you

PR #8240, opened 2020-12-15
Update action.yml

JacobCallahan:
This change should resolve the current issues we're seeing with failed package downloads.


omkarkhatavkar (approved):
ACK

rplevka (approved):
ACK, ...`apt update` can't harm

PR #8246, opened 2020-12-21
[POC] Initial DynaConf YAMLs - The robottelo copy of Gitlab SateQE Jenkins

jyejare:
The initial DynaConf YAMLs to replace the `robottelo.properties` - robottelo and server sections.

### DIFF
--- a/conf/robottelo.yaml
+++ b/conf/robottelo.yaml
@@ -10,7 +10,7 @@ ROBOTTELO:
   # Directory for temporary files
   TMP_DIR: /var/tmp
   # Web Server to provide various test artifacts
-  ARTIFACTS_SERVER: replace-with-artifacts-server-hostname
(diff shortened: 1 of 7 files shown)


JacobCallahan:
I do still prefer commented out fields over ones that exist, but have a "replace-with-*" value

mshriver:
@jyejare I think it would be good to name these files `<name>.yaml.template`, so that its clear these are template files, and should be copied/updated before use.

jyejare:
@mshriver Yeah with `.template` extension, the Robottelo DockerFile in CI wont complain while copying conf local YAMLs to robottelo. :)

There is force copy though :)

jyejare:
@mshriver @JacobCallahan Closing this PR as I have merged this PR changes to #8300 as one place to hold the changes required for DynaConf !

PR #8247, opened 2020-12-21
Convert test_rhai_client.py to pytest

tpapaioa:
This PR converts `test_rhai_client` to pytest. Setup code for org and activation key creation and manifest upload has been moved to fixtures. The test class has been deleted, and the provisioned client VM now uses `DISTRO_DEFAULT` (currently equal to `DISTRO_RHEL7`) instead of `DISTRO_RHEL6`.

Because `insights-client --test-connection` currently fails due to the bug described in

Bug 1892405 - insights-client --test-connection fails "Upload URL" test on host registered to Satellite 6
https://bugzilla.redhat.com/show_bug.cgi?id=1892405

the test is marked with `@pytest.mark.skip_if_open('BZ:1892405')`.

```
$ pytest tests/foreman/rhai/test_rhai_client.py
[...]
collected 1 item                                                                                                                                                                                                                             

tests/foreman/rhai/test_rhai_client.py s
[...]
=== 1 skipped, 3 warnings in 1.14s ===
```

### DIFF
--- a/tests/foreman/rhai/test_rhai_client.py
+++ b/tests/foreman/rhai/test_rhai_client.py
@@ -15,69 +15,72 @@
 :Upstream: No
 """
 import pytest
-from fauxfactory import gen_string
+from fauxfactory import gen_alpha
 from nailgun import entities
 
 from robottelo import manifests
 from robottelo.api.utils import upload_manifest
 from robottelo.constants import DEFAULT_SUBSCRIPTION_NAME
-from robottelo.constants import DISTRO_RHEL6
+from robottelo.constants import DISTRO_DEFAULT
 from robottelo.decorators import skip_if_not_set
-from robottelo.test import TestCase
 from robottelo.vm import VirtualMachine
 
 
+@pytest.fixture(scope='module')
+def org():
+    """Create a new organization with name prefix 'insights_'"""
+    return entities.Organization(name=gen_alpha(15, start='insights_')).create()
+
+
+@pytest.fixture(scope='module')
+def manifest_org(org):
+    """Upload manifest to organization."""
+    with manifests.clone() as manifest:
+        upload_manifest(org.id, manifest.content)
+    return org
+
+
+@pytest.fixture(scope='module')
+def activation_key(manifest_org):
+    """Create activation key using default CV and library environment."""
+    activation_key = entities.ActivationKey(
+        auto_attach=True,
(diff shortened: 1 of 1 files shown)


tests/foreman/rhai/test_rhai_client.py
```diff
@@ -14,70 +14,73 @@
...
+    :id: 167758c9-cbfa-4a81-9a11-27f88aaf9118
+
+    :expectedresults: 'insights-client --test-connection' should return 0 on a client that is
+    successfully registered to RHAI through Satellite.
+    """
+    with VirtualMachine(distro=DISTRO_DEFAULT) as vm:
+        vm.configure_rhai_client(activation_key.name, org.label, DISTRO_DEFAULT)
+        test_connection = vm.run('insights-client --test-connection')
+        logger.info(f'insights-client --test-connection returned: {test_connection.return_code}')
+        assert test_connection.return_code == 0, 'insights-client --test-connection failed'
```
JacobCallahan, line 86: it could be useful to also include stderr

tests/foreman/rhai/test_rhai_client.py
```diff
@@ -14,70 +14,73 @@
...
+def org():
+    """Create a new organization with name prefix 'insights_'"""
+    return entities.Organization(name=gen_alpha(15, start='insights_')).create()
+
+
+@pytest.fixture(scope='module')
+def manifest(org):
+    """Upload manifest to organization."""
+    with manifests.clone() as manifest:
+        upload_manifest(org.id, manifest.content)
```
JacobCallahan, line 44: could change this to something like `manifest_org` and return the org

JacobCallahan (changes requested):

tests/foreman/rhai/test_rhai_client.py
tpapaioa: I've updated the `assert` to print `return_code`, `stdout`, and `stderr`. I've removed the use of `logger`, since the `robottelo.ssh` module already logs all of this to `robottelo.log` when the command is run.

tests/foreman/rhai/test_rhai_client.py
tpapaioa: Done.

JacobCallahan (approved):
ACK, thanks for the updates!

mshriver (approved):

PR #8252, opened 2020-12-22
disable repo sync after test completion

swadeley:
Hello

After our discussion about reducing load on Satellite under test, it was on my to-do list to make PR for disabling sync plans . I found that none of the tests in CLI and API files have real yum repos, so nothing is being downloaded upon repo sync. So the load is negligible but I have made the changes anyway. The UI tests already delete sync plans after the test, so no change required in UI file.

EIDT: I found one UI test that used API call to set up the sync plan and then did not delete it like the other purely GUI tests did. Now added disable step to that on too.

Thank you

### DIFF
--- a/tests/foreman/api/test_syncplan.py
+++ b/tests/foreman/api/test_syncplan.py
@@ -760,7 +760,8 @@ def test_positive_synchronize_custom_products_future_sync_date(module_org, reque
 
     :BZ: 1695733
     """
-    delay = 4 * 60  # delay for sync date in seconds
+    # Test with multiple products and multiple repos needs more delay.
+    delay = 8 * 60  # delay for sync date in seconds
     products = [entities.Product(organization=module_org).create() for _ in range(2)]
     repos = [
         entities.Repository(product=product).create() for product in products for _ in range(2)
--- a/tests/foreman/api/test_syncplan.py
+++ b/tests/foreman/api/test_syncplan.py
@@ -108,6 +108,17 @@ def validate_repo_content(repo, content_types, after_sync=True):
             ], 'Repository contains invalid number of content entities.'
 
 
+def disable_syncplan(sync_plan):
+    """
+    Disable sync plans after a test to reduce distracting task events, logs, and load on Satellite.
+    Note that only a Sync Plan with a repo would create a noticeable load.
+    You can also create sync plans in a disabled state where it is unlikely to impact the test.
+    """
+    sync_plan.enabled = False
+    sync_plan = sync_plan.update(['enabled'])
+    assert sync_plan.enabled is False
+
+
 @pytest.mark.tier1
 def test_positive_get_routes():
     """Issue an HTTP GET response to both available routes.
@@ -156,6 +167,8 @@ def test_positive_create_enabled_disabled(module_org, enabled):
     sync_plan = entities.SyncPlan(enabled=enabled, organization=module_org).create()
     sync_plan = sync_plan.read()
     assert sync_plan.enabled == enabled
+    # disable sync plan after test
+    disable_syncplan(sync_plan)
 
 
 @pytest.mark.parametrize('name', **parametrized(valid_data_list()))
@@ -171,7 +184,7 @@ def test_positive_create_with_name(module_org, name):
 
     :CaseImportance: Critical
     """
-    sync_plan = entities.SyncPlan(name=name, organization=module_org).create()
+    sync_plan = entities.SyncPlan(enabled=False, name=name, organization=module_org).create()
     sync_plan = sync_plan.read()
     assert sync_plan.name == name
 
@@ -190,7 +203,9 @@ def test_positive_create_with_description(module_org, description):
 
     :CaseImportance: Critical
     """
-    sync_plan = entities.SyncPlan(description=description, organization=module_org).create()
+    sync_plan = entities.SyncPlan(
+        enabled=False, description=description, organization=module_org
+    ).create()
     sync_plan = sync_plan.read()
     assert sync_plan.description == description
 
@@ -209,7 +224,7 @@ def test_positive_create_with_interval(module_org, interval):
     :CaseImportance: Critical
     """
     sync_plan = entities.SyncPlan(
-        description=gen_string('alpha'), organization=module_org, interval=interval
+        enabled=False, description=gen_string('alpha'), organization=module_org, interval=interval
     )
     if interval == SYNC_INTERVAL['custom']:
         sync_plan.cron_expression = gen_choice(valid_cron_expressions())
@@ -232,7 +247,9 @@ def test_positive_create_with_sync_date(module_org, sync_delta):
     :CaseImportance: Critical
     """
     sync_date = datetime.now() + timedelta(seconds=sync_delta)
-    sync_plan = entities.SyncPlan(organization=module_org, sync_date=sync_date).create()
+    sync_plan = entities.SyncPlan(
+        enabled=False, organization=module_org, sync_date=sync_date
+    ).create()
     sync_plan = sync_plan.read()
     assert sync_date.strftime('%Y-%m-%d %H:%M:%S UTC') == sync_plan.sync_date
 
@@ -309,6 +326,8 @@ def test_positive_update_enabled(module_org, enabled):
     sync_plan.update(['enabled'])
     sync_plan = sync_plan.read()
     assert sync_plan.enabled == enabled
+    # disable sync plan after test
+    disable_syncplan(sync_plan)
 
 
 @pytest.mark.parametrize('name', **parametrized(valid_data_list()))
@@ -325,7 +344,7 @@ def test_positive_update_name(module_org, name):
 
     :CaseImportance: Critical
     """
-    sync_plan = entities.SyncPlan(organization=module_org).create()
+    sync_plan = entities.SyncPlan(enabled=False, organization=module_org).create()
     sync_plan.name = name
     sync_plan.update(['name'])
     sync_plan = sync_plan.read()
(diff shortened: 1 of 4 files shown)


tests/foreman/api/test_syncplan.py
```diff
@@ -639,6 +639,10 @@ def test_negative_synchronize_custom_product_past_sync_date(module_org):
     with pytest.raises(AssertionError):
         validate_task_status(repo.id, max_tries=2)
     validate_repo_content(repo, ['erratum', 'package', 'package_group'], after_sync=False)
+    # Disable sync plan after the test to reduce load on Satellite
+    sync_plan.enabled = False
+    sync_plan = sync_plan.update(['enabled'])
+    assert sync_plan.enabled is False
```
jameerpathan111, line 645: Stephen this LGTM.
one query, should we create a separate sync_plan_disable function(teardown) to handle this?

swadeley: I don't know what benefit that would have, e.g. would it be faster or easier to follow? It not really a teardown, so I reluctant to change it.

swadeley:
Another advantage of this change is that there will be less messages about tasks.

tests/foreman/api/test_syncplan.py
omkarkhatavkar: @swadeley I agree with @jameerpathan111 you have used that 4 line of code to disable the sync plan which is not part of test and used multiple times. It  can be taken as teardown or fixture that take care of disabling the sync plan

mshriver: @jameerpathan111 @omkarkhatavkar I would agree, its better to not repeat this block so many times. I would definitely consider it cleanup/teardown, because its not related to the test coverage, and should be part of the pytest TEARDOWN phase.  With it in the CALL phase like this, an assert would look like a test failure, not a failure in disabling the sync_plan after the test coverage is complete.  One additional consideration is that the test may fail before these statements are reached, and the sync_plan will be left on the system, enabled.

Given the sync_plans are created within the tests here, I think you should define a function that includes these three statements, and takes `sync_plan` as the argument.

These test functions should call `request.addfinalizer()` after creating the sync plan, so that it gets disabled at the proper scope, after the test call phase finishes.

mshriver (changes requested):
Let's use a pattern where we're not repeating the statements necessary to disable the sync plan.  Thanks for adding this enhancement though, addressing some of the latent system load the tests are currently leaving.

omkarkhatavkar:

swadeley:
trigger: test-robottelo
pytest: tests/foreman/cli/test_syncplan.py tests/foreman/api/test_syncplan.py

swadeley:
trigger: test-robottelo
pytest: tests/foreman/api/test_syncplan.py

tests/foreman/cli/test_syncplan.py
```diff
@@ -135,6 +135,17 @@ def validate_repo_content(repo, content_types, after_sync=True):
             assert not int(repo['content-counts'][content])
 
 
+def disable_syncplan(sync_plan):
```
jyejare, line 138: This is a teardown and you need not to perform it from CLI. You can use the api utils that you have (if) created as per above suggestion.

tests/foreman/api/test_syncplan.py
```diff
@@ -196,9 +209,13 @@ def test_positive_create_with_description(module_org, description):
...
     :CaseImportance: Critical
     """
-    sync_plan = entities.SyncPlan(description=description, organization=module_org).create()
+    sync_plan = entities.SyncPlan(
+        enabled=False, description=description, organization=module_org
+    ).create()
     sync_plan = sync_plan.read()
     assert sync_plan.description == description
+    # disable sync plan after test
+    disable_syncplan(sync_plan)
```
jyejare, line 218: This is already `enabled=False` Syncplan, need not to disable again.

jyejare (commented):
ACK pending comments.

mshriver:
@swadeley @jyejare I've moved this to draft because there are some comments that look like they are required for approval. @jyejare  and 'ack pending comments' should be a rejected review + draft state, so that its clear to @swadeley as a contributor that his PR needs updates to get approved/merged.

tests/foreman/api/test_syncplan.py
swadeley: OK, fixed

tests/foreman/cli/test_syncplan.py
swadeley: How would you call the API version of `disable_syncplan()` helper from CLI test? The structure of sync_plan in CLI is different to that in the API.

jyejare: @swadeley Since the teardown is different from the actual test steps / test object, this is not set in stone to teardown the test fron CLI, you should be able to choose the easiest/faster way to do it. And API is always faster.

jyejare:
@swadeley Any updates on this ? Pending from long ?

swadeley:
As Mike as moved this back to draft I will close this for now

mshriver:
@swadeley Moving the PR to draft state is part of the review process due to the required comments - it shouldn't be taken as an indication that the PR should be closed, or isn't relevant anymore.

The intent of this PR is sound - the patterns used to implement it just need some work.  I'd be happy to discuss these patterns in more depth if you would like some guidance on the implementation.

JacobCallahan:
@swadeley has said that he would like to revisit this soon.

tests/foreman/cli/test_syncplan.py
swadeley: @jyejare I have moved the helper from the API module to the api/utils file.

To get the CLI module to be able to use the API helper I added a line to read the sync plan using the nailgun code so that a sync plan object can be passed form the CLI module to the API helper in the expected format. I am sure there may be other ways to do this, e.g. pass org annd sync plan ID to the helper and do the read there, but then you are adding an extra read step in the case of API test.

Thank you.

swadeley:
trigger: test-robottelo
pytest: tests/foreman/api/test_syncplan.py tests/foreman/cli/test_syncplan.py

swadeley:
Hi, in PRT test 928 failed on:
tests.foreman.cli.test_syncplan.test_positive_synchronize_rh_product_future_sync_date  
tests.foreman.cli.test_syncplan.test_positive_synchronize_rh_product_past_sync_date
because `AttributeError: 'dict' object has no attribute 'id'`

These two tests do not use module_org and so format for org ID was different. Now fixed.

PRT test 928 also failed on tests.foreman.api.test_syncplan.test_positive_synchronize_custom_products_future_sync_date but that is an intermittent problem, something I need to investigate further.

swadeley:
trigger: test-robottelo
pytest: tests/foreman/api/test_syncplan.py tests/foreman/cli/test_syncplan.py

swadeley:
trigger: test-robottelo
pytest: tests/foreman/api/test_syncplan.py tests/foreman/cli/test_syncplan.py tests/foreman/ui/test_syncplan.py

swadeley:
I have squashed some commits while removing one redundant step.

swadeley:
trigger: test-robottelo
pytest: tests/foreman/api/test_syncplan.py tests/foreman/cli/test_syncplan.py tests/foreman/ui/test_syncplan.py

tests/foreman/api/test_syncplan.py
```diff
@@ -156,6 +157,8 @@ def test_positive_create_enabled_disabled(module_org, enabled):
     sync_plan = entities.SyncPlan(enabled=enabled, organization=module_org).create()
     sync_plan = sync_plan.read()
     assert sync_plan.enabled == enabled
+    # disable sync plan after test
+    disable_syncplan(sync_plan)
```
mshriver, line 161: So, consider what happens if the assert fails, or if Nailgun throws an exception during the read - this line will never get executed, so the syncplan will stay enabled.

In order to make this a stronger 'cleanup, I would move these calls into the test teardown by using `request.addfinalizer`.

Here this would look like: 

```
def test_positive_create_enabled_disabled(module_org, enabled, request, target_sat):
    sync_plan = target_sat.api.SyncPlan(enabled=enabled, organization=module_org).create()
    request.addfinalizer(lambda: disable_syncplan(sync_plan))
    sync_plan = sync_plan.read()
    assert sync_plan.enabled == enabled
```

mshriver: This applies to anywhere that you're directly calling `disable_syncplan` within the test.

It should be called via `request.addfinalizer()` right after it gets created.  The function won't actually execute until pytest is in the teardown phase.

mshriver (approved):

tests/foreman/api/test_syncplan.py
swadeley: I have added request.addfinalizer step to API, CLI, and UI modules.

tests/foreman/cli/test_syncplan.py
swadeley: I have now added request.addfinalizer step to API, CLI, and UI modules.

swadeley:
Hi, in PRT #946, test failed on:
tests.foreman.api.test_syncplan.test_positive_synchronize_custom_products_future_sync_date

Test passes locally. The test has two products and two repos per product and I think just needs more time between events so I am increasing the delay from 4 to 8 mins.

tests.foreman.ui.test_syncplan.test_positive_end_to_end 
tests.foreman.ui.test_syncplan.test_positive_end_to_end_custom_cron

These two UI tests pass locally. Possibly an airgun issue in reading the time.

swadeley:
trigger: test-robottelo
pytest:  tests/foreman/api/test_syncplan.py::test_positive_synchronize_custom_products_future_sync_date tests/foreman/ui/test_syncplan.py -k end_to_end

swadeley:
Thank you

PR #8261, opened 2020-12-30
[6.9 - Super Simple Installer] Stubs tests and Automated installer sections test

jyejare:
Satellite 6.9 Super Simple Installer features:

- [x] Stubs
- [x] An installer help sections automated test

### DIFF
--- a/tests/foreman/installer/test_installer.py
+++ b/tests/foreman/installer/test_installer.py
@@ -694,6 +694,7 @@ PREVIOUS_INSTALLER_OPTIONS = {
     '--puppet-user',
     '--puppet-vardir',
     '--puppet-version',
+    '--register-with-insights',
     '--reset-certs-ca-common-name',
     '--reset-certs-ca-expiration',
     '--reset-certs-city',
@@ -1315,6 +1316,31 @@ PREVIOUS_INSTALLER_OPTIONS = {
     '-v',
 }
 
+LAST_SAVED_SECTIONS = {
+    '= Generic:',
+    '= Module certs:',
+    '= Module foreman:',
+    '= Module foreman_cli:',
+    '= Module foreman_compute_ec2:',
+    '= Module foreman_compute_gce:',
+    '= Module foreman_compute_libvirt:',
+    '= Module foreman_compute_openstack:',
+    '= Module foreman_compute_ovirt:',
+    '= Module foreman_compute_vmware:',
+    '= Module foreman_plugin_tasks:',
+    '= Module foreman_proxy:',
+    '= Module foreman_proxy_content:',
+    '= Module foreman_proxy_plugin_ansible:',
+    '= Module foreman_proxy_plugin_dhcp_infoblox:',
+    '= Module foreman_proxy_plugin_dhcp_remote_isc:',
+    '= Module foreman_proxy_plugin_discovery:',
+    '= Module foreman_proxy_plugin_dns_infoblox:',
+    '= Module foreman_proxy_plugin_openscap:',
+    '= Module foreman_proxy_plugin_pulp:',
+    '= Module foreman_proxy_plugin_remote_execution_ssh:',
+    '= Module katello:',
+    '= Module puppet:',
+}
 
 SATELLITE_SERVICES = {
     'dynflow-sidekiq@orchestrator',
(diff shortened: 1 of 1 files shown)


tests/foreman/installer/test_installer.py
```diff
@@ -1446,6 +1484,34 @@ def test_installer_options_and_flags():
...
+def test_installer_help_sections():
+    """Look for changes on installer sections
+
+    :id: c49a0a40-530f-40f9-afb8-aa5f61868c8f
+
+    :Steps:
+        1. parse installer sections
+        2. compare with last sections
+
+    :expectedresults: Ideally sectionss should not change on zstreams.
```
omkarkhatavkar, line 1498: typo `sectionss`

omkarkhatavkar (approved):
ACK

tests/foreman/installer/test_installer.py
```diff
@@ -1315,6 +1315,31 @@
...
+    '= Module foreman_proxy_plugin_dhcp_infoblox:',
+    '= Module foreman_proxy_plugin_dhcp_remote_isc:',
+    '= Module foreman_proxy_plugin_discovery:',
+    '= Module foreman_proxy_plugin_dns_infoblox:',
+    '= Module foreman_proxy_plugin_openscap:',
+    '= Module foreman_proxy_plugin_pulp:',
+    '= Module foreman_proxy_plugin_remote_execution_ssh:',
+    '= Module katello:',
+    '= Module puppet:',
+}
```
devendra104, line 1342: Subsections verification is missing, Please add subsection verification along with sections.

tests/foreman/installer/test_installer.py
```diff
@@ -1353,6 +1378,19 @@ def extract_params(lst):
                 yield token.replace(',', '')
 
 
+def extract_sections(lst):
```
devendra104, line 1381: No need to create a separate function to extract sections, we can generalize the existing extract params function so it will work for all the params(params, sections, subsections).

tests/foreman/installer/test_installer.py
```diff
@@ -1446,6 +1484,34 @@ def test_installer_options_and_flags():
     assert PREVIOUS_INSTALLER_OPTIONS == current_installer_options, msg
 
 
+@pytest.mark.upgrade
+@pytest.mark.tier3
+def test_installer_help_sections():
+    """Look for changes on installer sections
```
devendra104, line 1490: No need to add a separate test case here most of the things are common, just add a parameterized fixtures to verify all params, sections, and subsections.

```
@pytest.mark.upgrade
@pytest.mark.tier3
@pytest.mark.parametrize("type", ['params', "sections", "subsections"])
def test_installer_options_and_flags(type):
```

tests/foreman/installer/test_installer.py
jyejare: Since we are verifying each and every option separately, I didnt emphasize on subsections. Do you still fill that we should still verify subsections. The gain is very less though.

tests/foreman/installer/test_installer.py
jyejare: Written/Modified `extract_params(lst)` to `extract_help(what)`.

tests/foreman/installer/test_installer.py
jyejare: Thanks for the nice suggestion here. Implemented !

jyejare:
@devendra104 Addressed your last comments and replied to 1.

tests/foreman/installer/test_installer.py
```diff
@@ -1337,20 +1362,25 @@
 }
 
 
-def extract_params(lst):
-    """Generator function to extract satellite installer params from lst.
+def extract_help(filter):
```
mshriver, line 1365: Given the use of `filter` in the function, it should be an optional parameter.

tests/foreman/installer/test_installer.py
```diff
@@ -1337,20 +1362,25 @@
...
-            if token[0] == '-':
-                yield token.replace(',', '')
+    stdout = ssh.command('satellite-installer --full-help').stdout
+    for line in (stdout or []):
+        if filter == 'sections':
+            line = line.strip()
+            if line.startswith('= '):
+                yield line
+        else:
+            first_2_tokens = line.split()[:2]
```
mshriver, line 1380: This is no longer calling `strip()` before the split, as that is only done when filter is specified.

tests/foreman/installer/test_installer.py
```diff
@@ -1419,31 +1449,33 @@ def test_positive_check_installer_services():
...
 @pytest.mark.tier3
-def test_installer_options_and_flags():
-    """Look for changes on installer options and flags
+@pytest.mark.parametrize('filter', ['params', 'sections'])
+def test_installer_options_and_sections(filter):
+    """Look for changes on installer sections and options/flags
 
     :id: a51d3b9f-f347-4a96-a31a-770349db08c7
 
+    :parametrized: true
```
mshriver, line 1458: The PR is failing the GHA because of formatting in the function above, but it will also fail testimony validation - this field is a choice of 'yes' or 'no'.

https://github.com/SatelliteQE/robottelo/blob/master/testimony.yaml#L154

mshriver (changes requested):
Couple things to address

tests/foreman/installer/test_installer.py
jyejare: OOPS ! FIXED!

jyejare:
@mshriver Fixed ! Please merge !

devendra104 (approved):
ACK

mshriver (approved):

PR #8264, opened 2021-01-04
Convert cli errata unittest to pytest

swadeley:
Convert CLI test_errata.py from unittest to pytest

### DIFF
--- a/tests/foreman/cli/test_errata.py
+++ b/tests/foreman/cli/test_errata.py
@@ -14,28 +14,25 @@
 
 :Upstream: No
 """
-import datetime
+from datetime import datetime
+from datetime import timedelta
 from operator import itemgetter
 
 import pytest
 from fauxfactory import gen_string
+from nailgun import entities
 
-from robottelo import manifests
 from robottelo import ssh
-from robottelo.cleanup import vm_cleanup
+from robottelo.api.utils import wait_for_tasks
 from robottelo.cli.activationkey import ActivationKey
 from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.contentview import ContentView
 from robottelo.cli.contentview import ContentViewFilter
 from robottelo.cli.erratum import Erratum
-from robottelo.cli.factory import make_activation_key
-from robottelo.cli.factory import make_content_view
 from robottelo.cli.factory import make_content_view_filter
 from robottelo.cli.factory import make_content_view_filter_rule
 from robottelo.cli.factory import make_filter
 from robottelo.cli.factory import make_host_collection
-from robottelo.cli.factory import make_lifecycle_environment
-from robottelo.cli.factory import make_org
 from robottelo.cli.factory import make_product
 from robottelo.cli.factory import make_repository
 from robottelo.cli.factory import make_role
@@ -48,7 +45,6 @@ from robottelo.cli.hostcollection import HostCollection
 from robottelo.cli.org import Org
 from robottelo.cli.repository import Repository
 from robottelo.cli.repository_set import RepositorySet
-from robottelo.cli.subscription import Subscription
 from robottelo.cli.task import Task
 from robottelo.cli.user import User
 from robottelo.config import settings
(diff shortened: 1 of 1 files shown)


swadeley:
Hello

At this stage still with classes. The main changes are in replacing the set-up function.

robottelo]$ pytest --show-capture=no tests/foreman/cli/test_errata.py 
========================= test session starts =====================================================
platform linux -- Python 3.8.7, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
rootdir: /sat6/robottelo_2/robottelo
plugins: ibutsu-1.12, mock-3.3.1, forked-1.3.0, services-2.2.1, cov-2.10.1, xdist-1.34.0
collecting ... 2021-01-15 16:45:03 - conftest - DEBUG - Collected 10 test cases
collected 10 items                                                                                                                                                                                

tests/foreman/cli/test_errata.py ..........                                                                                                                                                 [100%]

========================== 10 passed in 1031.41s (0:17:11) ============================================

mshriver:
@swadeley this is failing code quality checks due to `Black` formatting - ensure pre-commit hooks are running in your environment to avoid these types of failures. I'm going to move it to draft because its unmergable, but will still give a preliminary review.

pytest_fixtures/api_fixtures.py
```diff
@@ -46,6 +46,12 @@ def module_org():
     return entities.Organization().create()
 
 
+@pytest.fixture(scope='module')
+def orgs(request):
```
mshriver, line 50: To maintain consistency in the fixture naming, I would go with `module_orgs`

tests/foreman/cli/test_errata.py
```diff
@@ -80,113 +74,107 @@
...
 ERRATUM_MAX_IDS_INFO = 10
 
+CUSTOM_REPO_URL = FAKE_9_YUM_REPO
+CUSTOM_PACKAGE = FAKE_1_CUSTOM_PACKAGE
+CUSTOM_ERRATA_ID = FAKE_2_ERRATA_ID
+CUSTOM_PACKAGE_ERRATA_APPLIED = FAKE_2_CUSTOM_PACKAGE
+
 
+@skip_if_not_set('clients', 'fake_manifest')
+@pytest.mark.skipif((not settings.repos_hosting_url), reason='Missing repos_hosting_url')
```
mshriver, line 89: These markers are doing the same thing, but in two different ways. Let's consolidate them to one mark/decorator.

@JacobCallahan thoughts on which way we should go, that would lend itself toward future plans for handling test collection/skipping based on available configuration?

tests/foreman/cli/test_errata.py
```diff
@@ -80,113 +74,107 @@
...
 @pytest.mark.run_in_one_thread
-class HostCollectionErrataInstallTestCase(CLITestCase):
+class TestHostCollectionErrataInstallTest:
```
mshriver, line 91: I think you can remove the `'Test'` at the end of this class name.

tests/foreman/cli/test_errata.py
```diff
@@ -80,113 +74,107 @@
...
-    CUSTOM_PACKAGE_ERRATA_APPLIED = FAKE_2_CUSTOM_PACKAGE
-    VIRTUAL_MACHINES_COUNT = 2
 
-    @classmethod
-    @skip_if_not_set('clients', 'fake_manifest')
-    @pytest.mark.skipif((not settings.repos_hosting_url), reason='Missing repos_hosting_url')
-    def setUpClass(cls):
-        """Create Org, Lifecycle Environment, Content View, Activation key,
-        Host, Host-Collection
+@pytest.fixture(scope='module')
```
mshriver, line 95: Maybe its just the diff, but these fixture definitions appear to not be indented, and they follow the statement defining the test class.

The fixtures can be defined outside of the class, but they should be defined before the test class.  If you want to limit the fixture resolution to within the class, the fixture functions should be indented to be class methods.

pytest_fixtures/api_fixtures.py
swadeley: git blame suggests that is not part of this PR:
7737854b1 robottelo/api/entity_fixtures.py (Jitendra Yejare 2020-02-17 15:00:23 +0530  45) def module_org():

It creates one org and its widely used in that file.

mshriver: Miscommunication.

The current fixtures for organization creation are called `module_org` for the module scoped fixture, and `default_org` for the session scoped fixture - as `default_*` was widely used across the test suite for session scoped entities.

Entity fixtures (with some early adopter exceptions that should be cleaned up) follow a naming pattern of `<scope>_<entity>` - `module_org`, `module_location`, `function_role`, etc.

To be consistent with existing fixture names, and this pattern, your new fixture should be called `module_orgs`, a pluralization of the current `module_org` fixture.

mshriver: Think I actually misread your response. You didn't intend to define this fixture, and haven't committed it to your branch?

You need to rebase and clean up your commits in that case, if you don't intend to add this fixture in this PR.

swadeley: oh dear, this bit should not have been committed:
    +def orgs(request):
    +    num = getattr(request, 'param', 2)
    +    return [entities.Organization().create() for i in range(num)]

It was an idea that I did not use in the end. I will remove and test again.

tests/foreman/cli/test_errata.py
```diff
@@ -80,113 +74,107 @@
...
-            HostCollection.add_host(
-                {
-                    'id': self.host_collection['id'],
-                    'organization-id': self.org['id'],
-                    'host-ids': host['id'],
-                }
-            )
-        ActivationKey.add_host_collection(
+@pytest.fixture(scope='module')
+def setUp(module_org, module_activation_key):
```
mshriver, line 127: Please use a descriptive name for the fixture function.  It doesn't appear to return anything, so maybe a name like `register_host_collection`.

Indentation wise - this is not indented, but defined under the class definition. Same as above, please make it class scoped and define within the class, or move it's definition out of the class.

tests/foreman/cli/test_errata.py
```diff
@@ -80,113 +74,107 @@
...
+    """Create and setup host collection, hosts and virtual machines."""
+    virtual_machines = []
+    VIRTUAL_MACHINES_COUNT = 2
```
mshriver, line 130: I don't see any benefit to defining this attribute in this way.  I appreciate and like the idea of not just hard-coding a 'magic' number in the for loop, but it doesn't need to be capitalized like a constant.

To turn this into a 'feature' of the fixture, tie this vm count to `request.param`, so that tests can modify it with indirect parametrization.

tests/foreman/cli/test_errata.py
```diff
@@ -80,113 +74,107 @@
...
+
+    def _is_errata_package_installed(virtual_machine):
         """Check whether errata package is installed.
 
         :type virtual_machine: robottelo.vm.VirtualMachine
         :rtype: bool
         """
-        result = virtual_machine.run(f'rpm -q {self.CUSTOM_PACKAGE_ERRATA_APPLIED}')
+        result = virtual_machine.run(f'rpm -q {CUSTOM_PACKAGE_ERRATA_APPLIED}')
         return True if result.return_code == 0 else False
```
mshriver, line 174: Good opportunity to clean this up a bit. `return result.return_code == 0`

tests/foreman/cli/test_errata.py
```diff
@@ -80,113 +74,107 @@
...
 
     @pytest.mark.tier3
-    def test_positive_install_by_hc_id_and_org_id(self):
+    def test_positive_install_by_hc_id_and_org_id(module_org, setup):
```
mshriver, line 177: I think this is supposed to be the `setUp` fixture defined above - capitalization matters.

mshriver, line 177: Even if `self` isn't used within the test function, it is a good practice for every test function defined within a pytest class to take `self` as the first argument, before fixtures.

tests/foreman/cli/test_errata.py
```diff
@@ -204,17 +192,17 @@ def test_positive_install_by_hc_id_and_org_id(self):
         """
         install_task = HostCollection.erratum_install(
             {
-                'id': self.host_collection['id'],
-                'organization-id': self.org['id'],
-                'errata': [self.CUSTOM_ERRATA_ID],
+                'id': host_collection['id'],
```
mshriver, line 195: Where does `host_collection` come from in this use?

Looks like the fixture called `setUp` needs to return some entities for the tests to use.

tests/foreman/cli/test_errata.py
```diff
@@ -204,17 +192,17 @@ def test_positive_install_by_hc_id_and_org_id(self):
...
+                'organization-id': module_org.id,
+                'errata': [CUSTOM_ERRATA_ID],
             }
         )
         Task.progress({'id': install_task[0]['id']})
-        for virtual_machine in self.virtual_machines:
-            self.assertTrue(self._is_errata_package_installed(virtual_machine))
+        for virtual_machine in virtual_machines:
```
mshriver, line 201: Same with `virtual_machines` - this is part of the fixture, but isn't returned by the fixture and is ambiguous in use here. I'm surprised by the fact that this doesn't result in an UnboundLocal exception on execution, as it hasn't been assigned within the function scope.

tests/foreman/cli/test_errata.py
```diff
@@ -464,15 +452,15 @@ def test_negative_install_by_hc_name_without_org_info(self):
 
         :CaseLevel: System
         """
-        with self.assertRaises(CLIReturnCodeError) as context:
+        with pytest.raises(CLIReturnCodeError) as context:
```
mshriver, line 455: https://docs.pytest.org/en/stable/reference.html#pytest-raises

`pytest.raises` takes a kwarg of `match`, which I think you should take advantage of here.  It looks like CLIReturnCodeError will include stderr in its `__str__` / `__repr__`, and `match` should work for your use case.

Then you don't have to separately assert outside of the pytest.raises context.

tests/foreman/cli/test_errata.py
```diff
@@ -714,757 +700,768 @@ def test_host_errata_search_commands(self):
...
-        """
-        product_erratum_list = Erratum.list(
+    )
+    Repository.synchronize({'id': repo['id']})
+    return org1
+
+
+@pytest.fixture(scope='class')
+def org2():
+    """an org with one custom product & repository"""
```
mshriver, line 877: Please use the docblocks to highlight the differences between `org1`, `org2`, etc fixtures

tests/foreman/cli/test_errata.py
```diff
@@ -714,757 +700,768 @@ def test_host_errata_search_commands(self):
...
-            {'product-id': self.org_product['id'], 'per-page': 1000}
-        )
-        org_product_errata_ids = {errata['errata-id'] for errata in org_product_erratum_list}
-        errata_org_product_erratum_list = Erratum.list(
-            {'product-id': self.errata_org_product['id'], 'per-page': 1000}
-        )
-        errata_org_product_errata_ids = {
-            errata['errata-id'] for errata in errata_org_product_erratum_list
+@pytest.fixture(scope='class')
+def org1():
```
mshriver, line 859: Again please check your indentation here, there are fixtures defined under a class, but not indented.

tests/foreman/cli/test_errata.py
```diff
@@ -714,757 +700,768 @@ def test_host_errata_search_commands(self):
...
+        erratum_info_list.append(Erratum.info(options={'id': errata_id}, output_format='json'))
+    sorted_erratum_info_list = sorted(
+        erratum_info_list, key=itemgetter(sort_by), reverse=sort_reversed
+    )
+    return sorted_erratum_info_list
+
+
+@pytest.mark.tier3
+@pytest.mark.upgrade
+def test_positive_list_sort_by_issued_date(module_org):
```
mshriver, line 947: These test functions are defined after `TestErrata` class, but aren't indented under it. Earlier comments about including `self` apply here as well if these test cases should be inside the class.

tests/foreman/cli/test_errata.py
```diff
@@ -714,757 +700,768 @@ def test_host_errata_search_commands(self):
...
+        # build a list of errata-id field value in the same order as
+        # received from the detailed sorted errata info list
+        sorted_errata_ids = [errata['errata-id'] for errata in sorted_errata_info_list]
+        # ensure that the errata ids received by Erratum.list is sorted
+        # as needed
+        assert errata_ids == sorted_errata_ids
+
+
+@pytest.mark.tier3
+def test_positive_list_filter_by_org_id_and_sort_by_updated_date(module_org, rh_repo, custom_repo):
```
mshriver, line 995: There is a large amount of repeated code throughout these functions, with the same statements being executed in several test cases with only small differences. The test case names here are unwieldy because of their length.

It looks like the org field could be parametrized against a single test function, iterating over `id, name, label` for the org field, and `updated, issued` for the sorting.  Even further, org/product could be parametrized.

This will build a test function that implements a functionality to test, with a matrix of parameters that apply to that functionality.  The current test structure forces a high maintenance cost for any changes or fixes for the tests, and seemingly little value when it comes to test complexity.

mshriver (changes requested):
Some patterns to discuss, a few of which are fairly significant test restructuring. I can appreciate separating the initial pytest refactor from larger test improvement.

omkarkhatavkar:

tests/foreman/cli/test_errata.py
swadeley: @mshriver they are not the same, the @skip_if_not_set('clients' was to ensure you had configured the clients section of robottelo properties. As we are moving away from ini file type robottelo properties the point is now moot.

tests/foreman/cli/test_errata.py
swadeley: right, and I have removed the classes now too

tests/foreman/cli/test_errata.py
swadeley: I have removed the classes and reformatted the file, added new fixtures to break up the larger ones

tests/foreman/cli/test_errata.py
swadeley: I have removed the classes and reformatted the file, added new fixtures to break up the larger ones

tests/foreman/cli/test_errata.py
swadeley: Hello

I have done that, please review once more
Thank you

tests/foreman/cli/test_errata.py
swadeley: OK

tests/foreman/cli/test_errata.py
swadeley: Hi

I have reworked the fixtures, please review again

swadeley: Hi

I have removed the classes
I have reworked the fixtures, please review again

tests/foreman/cli/test_errata.py
swadeley: I have reworked the fixtures, please review again

tests/foreman/cli/test_errata.py
swadeley: I have reworked the fixtures, please review again

tests/foreman/cli/test_errata.py
swadeley: Hi

I have removed the classes
I have reworked the fixtures, please review again

tests/foreman/cli/test_errata.py
swadeley: Hi

I am sorry I did not have time to refactor this test, converting it to pytest is a big job in itself

tests/foreman/cli/test_errata.py
swadeley: I have improved docblocks a bit, please review once more

tests/foreman/cli/test_errata.py
swadeley: Hi

I have removed the classes
I have reworked the fixtures, please review again

tests/foreman/cli/test_errata.py
swadeley: Still ToDO

tpapaioa:
The code quality checks are failing:

```
tests/foreman/cli/test_errata.py:92:1: E304 blank lines found after function decorator
tests/foreman/cli/test_errata.py:95:1: E304 blank lines found after function decorator
```
I assume these lines are meant to be applied to all tests in the module:
```
@pytest.mark.skipif((not settings.repos_hosting_url), reason='Missing repos_hosting_url')
```
The way to do that is at the module level is with the `pytestmark` variable, which is a list of decorator markers (without the  `@`):

```
pytestmark = [
    pytest.mark.skipif((not settings.repos_hosting_url), reason='Missing repos_hosting_url'),
    pytest.mark.run_in_one_thread,
]
```

swadeley:
@tpapaioa Thank you

tests/foreman/cli/test_errata.py
swadeley: Changes made

swadeley:
trigger: test-robottelo
pytest: tests/foreman/cli/test_errata.py

swadeley:
trigger: test-robottelo
pytest: tests/foreman/cli/test_errata.py

tests/foreman/cli/test_errata.py
mshriver: Those skip markers still apply, and will continue to apply until the config facade layer is removed - at which point they'll be converted in bulk if a new syntax is necessary. So feel free to continue using them to control test behavior.

tests/foreman/cli/test_errata.py
```diff
@@ -79,1392 +75,1432 @@
...
+        1. erratum list --organization=<org name> --order 'updated ASC'
+        2. erratum list --organization=<org name> --order 'updated DESC'
+
+    :expectedresults: Errata is filtered by org name and sorted by updated
+        date.
+    """
+    sort_data = [('updated', 'ASC'), ('updated', 'DESC')]
+    for sort_field, sort_order in sort_data:
+        sort_text = f'{sort_field} {sort_order}'
+        sort_reversed = True if sort_order == 'DESC' else False
```
mshriver, line 1111: Doesn't need to be changed, but note that these ternarys are overkill when you're assigning a boolean.

`sort_reversed = sort_order == 'DESC'`

mshriver (approved):

PR #8268, opened 2021-01-05
Update github actions to no longer us gnutls

JacobCallahan:
This should eliminate the current issues we're seeing with conflicting
backends.


ColeHiggins2 (approved):
ACK

mshriver (approved):
Good riddance!

latran (approved):
ACK

PR #8271, opened 2021-01-06
Fix test failure

akhil-jha:


tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -854,9 +854,7 @@ def test_positive_login_ad_user_no_roles(auth_source, test_name, ldap_tear_down,
         functional areas of UI
     """
     with Session(test_name, ad_data['ldap_user_name'], ad_data['ldap_user_passwd']) as ldapsession:
-        with pytest.raises(NavigationTriesExceeded):
-            ldapsession.user.search('')
-        assert ad_data['ldap_user_name'] in ldapsession.task.read_all()['current_user']
+        ldapsession.task.read_all()
```
omkarkhatavkar, line 857: I think we need to assert it `assert ldapsession.task.read_all()`

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1838,9 +1834,7 @@ def test_verify_ldap_filters_ipa(session, ipa_add_user, auth_source_ipa, ipa_dat
     # 'test_user' able to login before the filter is applied.
     test_user = ipa_add_user
     with Session(user=test_user, password=ipa_data['ldap_user_passwd']) as ldapsession:
-        with pytest.raises(NavigationTriesExceeded):
-            ldapsession.user.search('')
-        assert test_user in ldapsession.task.read_all()['current_user']
+        ldapsession.task.read_all()
```
omkarkhatavkar, line 1837: same here

omkarkhatavkar (approved):
ACK minor points

tests/foreman/ui/test_ldap_authentication.py
mshriver: Agreed, I think this is a blocking comment/question though, not sure why you approved with this unanswered. @akhil-jha Thoughts or explanation of why this isn't asserted against in some way?

mshriver (changes requested):
Without a PR description or in-line comments, its not clear what exactly is being fixed here, and why the test is reading from ldapsession without asserting against the read data at all.

tests/foreman/ui/test_ldap_authentication.py
akhil-jha: Since the user has not been given any permission other than the default(also the fixtures are default) which is just the bookmark and task, I believe, checking for any access is not worth it. Also on the assert keyword, since the user has no access to any area, we cannot assert it against any value. So if `ldapsession.task.read_all()` fails, that means the user was not able to login and the test will fail. If `ldapsession.task.read_all()` is successful, putting assert or not doesn't matter.

akhil-jha:
> Without a PR description or in-line comments, its not clear what exactly is being fixed here, and why the test is reading from ldapsession without asserting against the read data at all.

I honestly don't remember why I didn't put description and test result. :/
My bad. I will update it.

mshriver (approved):

PR #8245, opened 2020-12-17
Avoid 'Taxonomy has already been taken'. Explicitly cast rhel7_os to str so it won't fail being converted to JSON.

lhellebr:

PR #8278, opened 2021-01-08
Rewrite to pytest

lhellebr:


lhellebr:
Results after applying https://github.com/Katello/katello/pull/9084 and https://github.com/SatelliteQE/robottelo/pull/8245:

```
$ pytest tests/foreman/longrun/test_provisioning_computeresource.py
============================================== test session starts ==============================================
rootdir: /home/lhellebr/git/robottelo
plugins: services-1.3.1, ibutsu-1.1.1, xdist-1.34.0, cov-2.10.0, mock-3.3.1, forked-1.2.0
collected 2 items                                                                                               

tests/foreman/longrun/test_provisioning_computeresource.py ..                                             [100%]

========================================= 2 passed in 558.87s (0:09:18) =========================================

```

Depends on https://github.com/SatelliteQE/robottelo/pull/8245.

tests/foreman/longrun/test_provisioning_computeresource.py
```diff
@@ -255,12 +266,11 @@ def test_positive_provision_vmware_with_host_group(self):
                 'provision-method': 'build',
             }
         )
-        hostname = '{}.{}'.format(host_name, self.config_env['domain'])
-        self.assertEquals(hostname, host['name'])
+        hostname = '{}.{}'.format(host_name, provisioning['config_env']['domain'])
```
omkarkhatavkar, line 269: f' strings ?

tests/foreman/longrun/test_provisioning_computeresource.py
```diff
@@ -133,55 +145,54 @@ def test_positive_provision_rhev_with_host_group(self):
...
                 'volume': "size_gb=10,"
                 "storage_domain={},"
-                "bootable=True".format(self.storage_id),
+                "bootable=True".format(rhev['storage_id']),
                 'provision-method': 'build',
             }
         )
-        hostname = '{}.{}'.format(host_name, self.config_env['domain'])
-        self.assertEquals(hostname, host['name'])
+        hostname = '{}.{}'.format(host_name, provisioning['config_env']['domain'])
```
omkarkhatavkar, line 181: f strings

omkarkhatavkar (approved):
ACK

tests/foreman/longrun/test_provisioning_computeresource.py
lhellebr: I just rewrote to pytest, perhaps I'll mass-port to f-strings later.

swadeley: Hi, take a look at https://github.com/SatelliteQE/robottelo/pull/8281

tests/foreman/longrun/test_provisioning_computeresource.py
```diff
@@ -23,88 +22,101 @@
...
 from robottelo.helpers import host_provisioning_check
-from robottelo.helpers import ProvisioningCheckError
-from robottelo.test import CLITestCase
 from robottelo.utils.issue_handlers import is_open
 
 
-class ComputeResourceHostTestCase(CLITestCase):
-    """RHEVComputeResource CLI tests."""
+@pytest.fixture(scope="module")
+def bridge():
```
mshriver, line 30: This is overkill usage of fixtures, no need to wrap access to `settings` into fixtures.

tests/foreman/longrun/test_provisioning_computeresource.py
```diff
@@ -23,88 +22,101 @@
...
+        compute=True, org=module_org, loc=module_location, os=os
+    )
+    provisioning['os_name'] = provisioning['config_env']['os']
+    return provisioning
+
+
+class TestComputeResourceHost:
+    """RHEVComputeResource CLI tests."""
+
+    def tearDown(self, provisioning):
```
mshriver, line 112: I don't actually see this being called anywhere in your refactor - in unittest, this function was called automatically at the function scope.

I think you should define `tear_down` as a function scoped fixture, that yields and then calls the relevant three lines to delete hosts.

tests/foreman/longrun/test_provisioning_computeresource.py
```diff
@@ -23,88 +22,101 @@
...
+    provisioning['org_name'] = module_org.name
+    provisioning['loc_name'] = module_location.name
+    provisioning['config_env'] = configure_provisioning(
+        compute=True, org=module_org, loc=module_location, os=os
+    )
+    provisioning['os_name'] = provisioning['config_env']['os']
+    return provisioning
+
+
+class TestComputeResourceHost:
```
mshriver, line 109: I don't believe the test class is providing much value here, since all fixtures have been module scoped, and there don't appear to be any other test classes in this module.

Consider flattening the test module, and defining the test functions without a class.

tests/foreman/longrun/test_provisioning_computeresource.py
```diff
@@ -23,88 +22,101 @@
...
-            compute=True, org=cls.org, loc=cls.loc, os=cls.rhev_img_os
-        )
-        cls.os_name = cls.config_env['os']
 
-    def tearDown(self):
-        """Delete the host to free the resources"""
-        super().tearDown()
-        hosts = Host.list({'organization': self.org_name})
+@pytest.fixture(scope="module")
+def rhev(bridge):
```
mshriver, line 35: While at this time its a fair amount of syntax refactoring, I think the original test had a better pattern with attributized reference to these key/value pairs.

You can do this very simply by defining a `collections.namedtuple` type, and use it to store data from `settings`.

Its not clear to me why all of these key/value pairs have to be 'translated', and stored with `rhev_<key>`, when the test appears to be building dictionaries that reference the original `<key>` without any prefix.  It appears the test could directly reference `settings.rhev`, and these fixtures would be far less verbose.

mshriver (changes requested):
A few things to discuss, and tear down behavior to resolve.

mshriver:
Please squash out the merge commit that is in this branch. Branches should be rebased against master, instead of using `pull` or `merge`

tests/foreman/longrun/test_provisioning_computeresource.py
```diff
@@ -26,68 +28,100 @@
...
-    rhev['rhev_vm_name'] = settings.rhev.vm_name
-    rhev['rhev_storage_domain'] = settings.rhev.storage_domain
-    rhev['rhv_api'] = RHEVMSystem(
-        hostname=rhev['rhev_url'].split('/')[2],
-        username=rhev['rhev_username'],
-        password=rhev['rhev_password'],
+@pytest.fixture(scope="function")
+def rhev():
+    bridge = settings.vlan_networking.bridge
+    rhev = namedtuple(
```
mshriver, line 34: Thanks for taking the time to make this switch so we have attributized access! 

There is a bit more that needs done here - using `namedtuple` is kind of a two step process.

First, an instance of `namedtuple` is created, which acts like a class definition.  Then that is used to create individual instances of that type.

```
RHEVSettings = namedtuple('RHEVSettings', [<field names>])
rhev = RHEVSettings(settings.rhev.hostname, settings.rhev.password, <etc>)
```

Any thoughts on ways to simplify these blocks, so that we're not having to map all of these fields?

tests/foreman/longrun/test_provisioning_computeresource.py
```diff
@@ -96,181 +130,182 @@ def provisioning(module_org, module_location):
...
-        hostname = '{}.{}'.format(host_name, provisioning['config_env']['domain'])
-        assert hostname == host['name']
-        # Check on Vmware, if VM exists
-        assert vmware['vmware_api'].does_vm_exist(hostname)
-        host_info = Host.info({'name': hostname})
-        host_ip = host_info.get('network').get('ipv4-address')
-        # Start to run a ping check if network was established on VM
-        host_provisioning_check(ip_addr=host_ip)
+@pytest.fixture(scope="function")
+def tear_down(provisioning):
```
mshriver, line 146: Still not seeing this fixture used within  the test functions?

lhellebr:
@mshriver , is there anything else that requires changes?

omkarkhatavkar:

rplevka (approved):

lhellebr:
@mshriver

mshriver:
Trusting @rplevka second review here, thanks @lhellebr!

PR #8285, opened 2021-01-14
Add polarion test case and test run scripts into robottelo

lpramuk:
- added polarion test case upload script
- added polarion test run upload script

### DIFF
--- a/scripts/polarion-test-case-upload.sh
+++ b/scripts/polarion-test-case-upload.sh
@@ -1,0 +1,81 @@
+#!/bin/bash
+
+# requires curl and betelgeuse pypi package installed
+# https://betelgeuse.readthedocs.io/en/latest/
+set -e
+
+usage() {
+cat <<EOM
+
+usage: $0 options
+
+Tool to upload Test Case data into Polarion
+
+OPTIONS:
+  -h, --help
+         Show this message
+  -s, --server <polarion URL>
+         Upload to the specified polarion instance
+  -u, --user <username>
+         Specify the user name to use for authentication
+  -p, --password <password>
+         Specify the password to use for authentication
+         You can pass POLARION_PASSWORD env variable instead of this option
+  -P, --project <projectname>
+         Specifiy the polarion project to use
+
+
+EXAMPLE:
+  $0 -s http://polarion.example.com/polarion/ -u Polar -p *** -P PolarStuff
+
+EOM
+}
+
+if ! options=$(getopt -o hs:u:p:P: -l help,server:,user:,password:,project: -- "$@"); then
+  exit 1
+fi
+
+eval set -- $options
+
+while [ $# -gt 0 ]; do
+    case $1 in
+    -s|--server) POLARION_URL="$2" ; shift ;;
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -2,6 +2,7 @@
 # Version updates managed by pyup.io
 
 attrdict==2.0.1
+betelgeuse==1.7.0
```
mshriver, line 5: I think this should go into `requirements-optional.txt` (Ideally we refactor anyway to use setuptools extra_requires groups),  as its not directly related to actually executing pytest from robottelo.

scripts/polarion-test-case-upload.sh
```diff
@@ -0,0 +1,33 @@
+# requires betelgeuse pypi package
```
mshriver, line 1: Please provide more thorough docblocks for both scripts explaining their use, arguments, defaults, etc.

scripts/polarion-test-case-upload.sh
```diff
@@ -0,0 +1,33 @@
...
+
+TC_PATH='tests/foreman/api tests/foreman/cli tests/foreman/ui tests/foreman/longrun tests/foreman/sys tests/foreman/installer tests/foreman/rhai tests/foreman/endtoend tests/foreman/virtwho'
```
mshriver, line 3: Can we instead exclude the paths this should not be run against, or at least comment what we're avoiding here? Looks like all of `tests/foreman` could be included.

scripts/polarion-test-run-upload.sh
```diff
@@ -0,0 +1,27 @@
...
+    --response-property "${POLARION_SELECTOR}" \
+    --test-run-title "${TEST_RUN_ID}" \
+    --test-run-id "${TEST_RUN_ID}" \
+    --test-run-group-id "${TEST_RUN_GROUP_ID}" \
+    --status inprogress \
+    "./${ENDPOINT}-results.xml" \
+    tests/foreman \
+    "${POLARION_USERNAME}" \
+    "${POLARION_PROJECT}" \
+    "sat-${IMPORTANCE}-results.xml"
```
mshriver, line 24: I think this (and the ENDPOINT-results.xml on L20) needs to be more configurable. Its relevant to how we're writing CI, but these envvars won't be relevant or set for component owners trying to upload results.

lpramuk: oops I think I already amended the line, watch me repushing PR

scripts/polarion-test-case-upload.sh
lpramuk: Refactored.
it has no sense

omkarkhatavkar:

requirements.txt
lpramuk: removing from `requirements.txt`

mshriver (approved):
Approving because the addition of these scripts don't impact the framework itself, but with some comments.

1. This should be implemented as a robottelo-cli command group in `Click`
2. The betelgeuse requirement should be included in requirements-optional.txt, as its required to run this script.

JacobCallahan (approved):

PR #8290, opened 2021-01-19
Pull GHA workflow yamls from master branch

mshriver:
There are a few commits involving GHA changes in master branch from when
these files were created, but those commits include extraneous content.

I don't think there is any need to preserve specific commit history for
these files and cherry-pick those specific commits from master to 6.y.z
branches. The GHA configuration should be consistent across our actively
developed branches.

### DIFF
--- a/.github/workflows/merge_to_master.yml
+++ b/.github/workflows/merge_to_master.yml
@@ -1,6 +1,10 @@
+# CI stages to execute against master branch on PR merge
 name: update_robottelo_image
 
 on: [push]
+
+env:
+    PYCURL_SSL_LIBRARY: openssl
 
 jobs:
   codechecks:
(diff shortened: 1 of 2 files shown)


swadeley (approved):
ACK

jyejare (approved):
ACK !

But do you think we should rename `.github/workflows/merge_to_68z.yml` or `.github/workflows/merge_to_zStream.yml` instead of `.github/workflows/merge_to_master.yml`  ?

mshriver:
@jyejare I think the unit tests do still have value, and some contributors have been updating them when they modify framework components. They've helped me with some of my PRs when it comes to testing, and I think its worth keeping them as part of our PR CI process, and encouraging, if not enforcing, that the unit tests have to be updated or added when working on core framework.

PR #8291, opened 2021-01-19
Add pytest to create coverage report on master

mshriver:
Grabbed the coverage upload in #8289, but missed adding the pytest run that generates the report.

The pull request and merge workflows have an effectively identical job definition, but GHA doesn't provide a great way of reducing that duplication. The `on` attribute of the workflow only applies to the workflow itself, not any specific jobs within the workflow. It does not appear that this `on` attribute is accessible through any of the available contextual metadata within the job, so that they can be dynamically executed.

### DIFF
--- a/.github/workflows/merge_to_master.yml
+++ b/.github/workflows/merge_to_master.yml
@@ -48,6 +48,9 @@ jobs:
           pytest -n 8 --setup-plan --disable-pytest-warnings -m pre_upgrade tests/upgrades/
           pytest -n 8 --setup-plan --disable-pytest-warnings -m post_upgrade tests/upgrades/
 
+      - name: Test Robottelo Coverage
+        run: pytest --cov --cov-config=.coveragerc --cov-report=xml tests/robottelo
+
       - name: Make Docs
         run: |
           make test-docstrings


swadeley (approved):
ACK, but note that when I ran this locally there were a few `no such table: coverage_schema` errors at the end.

jyejare (approved):
ACK,

TBH we have never make use of this codecov and people are not adding any tests here. This will always be a constant until we setup a process to add tests by a contributor. Else lets remove this annoying thing completely from Robottelo.

Not sure how much `Test the Test Suite` definition applies here with a broken/half coverage with no one interested to improve it.

mshriver:
@swadeley interesting, I'll be reaching out if that happens in master too. The job step is verbatim copied from the one that runs in the PR workflows.  Codecov has integration between GH and codecov.io, so those errors might be related to locally missing files/content that GH has in the workflow environment

mshriver:
@jyejare I think the unit tests do still have value, and some contributors have been updating them when they modify framework components. They've helped me with some of my PRs when it comes to testing, and I think its worth keeping them as part of our PR CI process, and encouraging, if not enforcing, that the unit tests have to be updated or added when working on core framework.

PR #8296, opened 2021-01-20
Fix test_positive_last_checkin_status test case

hkx303:
**Description**
Now, the time format for the Last Checkin in the contest host page has been changed from `Jan 19, 11:42 PM` to `January 20, 2021, 12:32 PM`. Need to update the test case to fix the format error.


mshriver (approved):

swadeley (approved):
ACK

PR #8300, opened 2021-01-21
Dynaconf CI changes including native configuration for Airgun and Nailgun

jyejare:
This PR includes:
--------------------------
0. Merged PR #8246 in this PR. which contains:
    - Additions of DynaConf YAML templates for 2 sections- Robottelo and Server. 
    - Inclusion of missing repos in Robottelo from Gitlab and 
    - Updates to tests wrt repo changes
1. Configuration for Nailgun and Airgun moved SettingsNodeWrapper which fetches settings from Dynaconf / Legacy settings using settings Facade.
2. Calling Nailgun and Airgun settings funcs post settings are set from facade (Obiviously) :)
3. Removed Nailgun and Airgun configuration from legacy settings object, due to (1).
4. New Robottelo dispatch value func which returns data using settings object for DynaConf as `setings.screenshots_path` instead of `settings.robottelo.screenshots_path`. This notation is being used by all the tests in robottelo, so this saves from big refactoring.
5. Separate repos.yaml and swapped the places of existing settings in robottelo and server YAMLs.
6. Logging updated for facade based settings object. robottelo and server settings base config removed.

Backporting Commit:
-----------------------------
The back-porting commit https://github.com/SatelliteQE/robottelo/commit/3fdb58ff63f68500d7f106d71695d31502b7c747 is added to make PR / facade compatible with existing / old pipeline. The commit should be removed as soon as we move to new pipeline. See below.

TO-DOs for future (OR when we start using new Pipeline):
-------------------------------------------------------------------------------
1. Uncommit the commit https://github.com/SatelliteQE/robottelo/commit/3fdb58ff63f68500d7f106d71695d31502b7c747 and rebase with that times Master.

TO-DOs for future (OR when we translate all robottelo.settings sections to DynaConf settings):
--------------------------------------------------------------------------------------------------------------------------------
1. Remove facade and all its methods.
2. Remove all the methods from `robottelo.config.base` of legacy settings.
3. Update Unit-tests for removals of legacy settings and additions of dynaconf settings.


JacobCallahan (approved):
ACK. Merge pending some API/UI test results

robottelo/config/facade.py
```diff
@@ -119,6 +123,71 @@ def __repr__(self):
             self.__wrapped__,
         )
 
+    def get_project_root(self):
```
mshriver, line 126: This is now defined here and in `robottelo.config.base`. I think you should import the already defined function, and move it if that causes a circular import.

robottelo/config/facade.py
```diff
@@ -284,6 +353,21 @@ def _dispatch_computed_value(self, key):
         self._add_to_cache(key, value)
         return value
 
+    def _dispatch_robottelo_value(self, key):
```
mshriver, line 356: Maybe I'm missing something obvious here, but its not clear to me why this method is actually needed given the changes you've made?  I'm not seeing what the existing use of `_dispatch_computed_value` or `_get_from_configs` would miss that this is handling.

At a minimum include a docblock here explaining the necessity.

jyejare: @mshriver I have listed the reason of this description on the Description of this PR. Still copying here :
```
New Robottelo dispatch value func which returns data using settings object for DynaConf as `setings.screenshots_path` instead of settings.robottelo object as `settings.robottelo.screenshots_path`. This notation is being used by all the tests in robottelo, and hence we need a big refactor for the tests/helpers using settings from robottelo. So this saves from big refactoring.
```

jyejare: I didnt include docblock as none of other funcs have it :)

mshriver: That's not a good pattern to follow, and violates our code standards, which dictate all classes, methods, and functions will have docstrings.

mshriver: Thanks for explaining the need here.

I think ultimately its a pattern that _should_ be refactored throughout the framework. Access to settings values should be explicit - if I'm working on a test and referencing `settings.screenshot_path` instead of `settings.robottelo.screenshot_path`, I've lost the context of where my settings are defined. This layer obfuscates that for very minimal convenience.

Due to the incremental transition to using dynaconf, this refactor doesn't have to occur in this PR.

I think the docblock for this method should include a TODO to refactor `settings` references to `settings.robottelo` where appropriate for the new config.  Its a very easy find+replace task.

jyejare: Sure! and Make Sense, since the PR is raised as a workaround for unblocking the DynaConf CI, my plan is to point the settings to settings.robottelo in a separate PR where we will also remove the INIReader reference altogether, even we will remove the facade implementation itself in that PR, so the need of this will be auto removed.

robottelo/config/facade.py
jyejare: Fixed! Using it from base right now.

jyejare:
@mshriver @JacobCallahan Merged #8246 here for the reason mentioned in that PR as close comment.

robottelo/config/facade.py
```diff
@@ -284,6 +347,21 @@ def _dispatch_computed_value(self, key):
...
+        robottelo_keys = [setting.lower() for setting in self._configs[0].robottelo.keys()]
+        top_key = key.split('.')[0] if '.' in key else key
```
JacobCallahan, line 352:
```suggestion
        top_key = key.split('.')[0]
```

JacobCallahan (changes requested):
One minor change

robottelo/config/facade.py
jyejare: Its required when key == `satttools.rhel6` but not when key == `capsule_repo` which doesnt have `.`.

In short, it is required for 4 level of a setting. E.g `settings.robottelo.sattools.rhel6`.

jyejare: @JacobCallahan you can check Robottelo.yaml for those kind of settings.

JacobCallahan: From a pure python perspective, this change isn't necessary. If no period is in they key, then split will result in a single element list. Since you just want that first element, it is functionally the same.

jyejare: Yes ! Updated.

JacobCallahan (approved):
Thanks for your hard work on this one!

omkarkhatavkar:

mshriver:
@jyejare Please investigate the code quality checks

jyejare:
@mshriver yup, investigating.

conf/robottelo.yaml.template
```diff
@@ -0,0 +1,92 @@
...
+  # webdriver_binary=/usr/bin/chromedriver
+  # webdriver_binary=C:\\Program Files (x86)\\Microsoft Web Driver\\MicrosoftWebDriver.exe
+  WEBDRIVER_BINARY: /usr/bin/chromedriver
+  # Zalenium command_executor
+  # COMMAND_EXECUTOR: http://127.0.0.1:4444/wd/hub
+  # CDN sync
+  CDN: true
+  # Run one datapoint or multiple datapoints for tests
+  RUN_ONE_DATAPOINT: false
+  # Provide link to rhel6/7/8 repo here, as puppet rpm would require packages from
```
mshriver, line 63: I think we should consider starting a separate file for all of the repository fields from here to the end. They aren't necessarily core framework configuration, but test target configuration

conf/server.yaml.template
```diff
@@ -0,0 +1,24 @@
+SERVER:
+  # Server hostname
+  # HOSTNAME: replace-with-satellite-hostname
+  # Path to private ssh key to be used when connecting via SSH.
+  # SSH_KEY:
```
mshriver, line 5: Please include `SSH_KEY_STRING` here as well

conf/server.yaml.template
```diff
@@ -0,0 +1,24 @@
...
+  # HTTP scheme when building the server URL
+  # Suggested values for "scheme" are "http" and "https".
+  SCHEME: https
+  # Server port
+  PORT: 443
+  # SSH username
+  SSH_USERNAME: root
+  # SSH password if not using a SSH key to connect
+  # SSH_PASSWORD: replace-with-ssh-passwd
```
mshriver, line 14: Nitpick, but it would be good to organize these a bit better - keep the `SSH_*` keys together, and the `HOSTNAME`/`SCHEME`/`PORT` keys together

conf/server.yaml.template
```diff
@@ -0,0 +1,24 @@
...
+  # Admin username when accessing API and UI
+  ADMIN_USERNAME: admin
+  # Admin password when accessing API and UI
+  ADMIN_PASSWORD: changeme
+
+SSH_CLIENT:
```
mshriver, line 20: A TODO once we've dropped the legacy configuration layer - bring `SSH_*` keys into this block since they're all directly related. Its odd to have some SSH client configuration outside of the SSH_CLIENT block.

conf/robottelo.yaml.template
jyejare: Added a new `repos.yaml.template` with all repos keys.

conf/server.yaml.template
jyejare: I was not sure how one can provide a plain SSH Key String, `specially with new line chat or without` even I didnt try by myself in robottelo. But in CI it picks very well.

Still added for component owners on the basis of Go `find yourself`. :)

conf/server.yaml.template
jyejare: Done , with comment on top of them :

`# Use either of SSH_PASSWORD or SSH_KEY or SSH_KEY_STRING below`

conf/server.yaml.template
jyejare: Yes, makes very sense ! For now we can handle it in separate PR as we need to update dependent tests for that in this PR. and which we dont want.

jyejare:
@mshriver Fixed your comments, we should be good to get this in.

Travis is also fixed and passing.

mshriver (approved):
Excellent work @jyejare !

jyejare:
@mshriver @JacobCallahan 

Tested Tier1 tests with new commit https://github.com/SatelliteQE/robottelo/pull/8300/commits/3fdb58ff63f68500d7f106d71695d31502b7c747 for backporting the changes for old pipeline. The results are satisfactory and without any side-effects.

The comparative tests results analysis is done between satellite-jenkins/DynaConf-Robottelo-Tier1/#3 (54 fails) and satellite-jenkins/automation-6.9-tier1-rhel7/#18 (51 fails) test runs. The extra failures are not at all related to this PR changes.

So we are to get this in.

conf/server.yaml.template
mshriver: The key string will likely be set via envvar, not directly in the file - either way, it can contain newlines as private keys should.

mshriver:
test-robottelo

mshriver:
test-robottelo

PR #8301, opened 2021-01-22
Add requirements-optional to pyup

mshriver:
Pin broker to 0.1.8, pyup should provide regular updates on release

I added a configuration field to pyupbot to apply a label for us, but it requires private key permissions even on public repos.  I'm not sure who currently owns the API key used by pyup-bot, but we may need to update that for the configuration to take effect.

Also thought it was worthwhile to have our 'optional' requirements installed as well.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 # Version updates managed by pyup.io
 
 attrdict==2.0.1
-broker>=0.1.3
+broker==0.1.8
 cryptography==3.3.1
 deepdiff==5.2.2
 dynaconf==3.1.2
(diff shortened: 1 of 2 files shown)


JacobCallahan (approved):

requirements.txt
```diff
@@ -2,7 +2,7 @@
 # Version updates managed by pyup.io
 
 attrdict==2.0.1
-broker>=0.1.3
+broker==0.1.8
```
mshriver, line 5: We've gone back and forth on this a bit, but using `>=` means pyup-bot won't ever update the minimum version for us automatically.

Since pip's default is not an eager update, I'm moving this back to pinned, and pyup-bot will automatically keep it up-to-date for us.

PR #8303, opened 2021-01-22
Add the ability for Robottelo to dynamically request test targets

JacobCallahan:
This change introduces the ability for robottelo to fill Satellite test
target requirements based on criteria defined in the new server settings
file.
One notable configuration change is that hostname is now hostnames,
which requires a list of one or more hostnames.
If one or more Satellites are required to meet xdist worker demands,
then the additional settings will control on-demand behavior.

### DIFF
--- a/conf/server.yaml.template
+++ b/conf/server.yaml.template
@@ -1,6 +1,31 @@
 SERVER:
-  # Server hostname
-  # HOSTNAME: replace-with-satellite-hostname
+  # Server hostnames, a list of one or more dns-resolvable hostnames
+  HOSTNAMES:
+    - null
+  #  - replace.with.satellite.hostname
+  #  - replace.with.satellite.hostname
+  VERSION:
+    # The full release version (6.9.2)
+    # RELEASE: populate with satellite version
+    # The snap version currently testing (if applicable)
+    # SNAP:
+    # The source of Satellite packages. Can be one of:
+    # internal, ga, beta
+    SOURCE: "internal"
+  # run-on-one - All xdist runners default to the first satellite
+  # balance - xdist runners will be split between available satellites
+  # on-demand - any xdist runner without a satellite will have a new one provisioned.
+  # if a new satellite is required, test execution will wait until one is received.
+  XDIST_BEHAVIOR: "run-on-one"
+  # Before attempting a new broker checkout, check the local inventory
+  FROM_INVENTORY: False
+  # A broker filter specifying which hosts can be used
+  INVENTORY_FILTER: "name<sat"
+  # If one or more Satellites are provisioned,
+  # this setting determines if they will be automatically checked in
+  AUTO_CHECKIN: False
+  # The Ansible Tower workflow used to deploy a satellite
+  DEPLOY_WORKFLOW: "deploy-sat-jenkins"
   # HTTP scheme when building the server URL
   # Suggested values for "scheme" are "http" and "https".
   SCHEME: https
--- a/pytest_fixtures/broker.py
+++ b/pytest_fixtures/broker.py
@@ -1,10 +1,44 @@
 import pytest
 from broker.broker import VMBroker
+from wait_for import wait_for
 
+from robottelo.config import settings
 from robottelo.constants import BROKER_RHEL77
 from robottelo.hosts import Capsule
 from robottelo.hosts import ContentHost
 from robottelo.hosts import Satellite
+
+
+@pytest.fixture(scope='session')
+def satellite_factory():
+    def factory(retry_limit=3, delay=300, **broker_args):
+        vmb = VMBroker(
+            host_classes={'host': Satellite},
+            workflow=settings.server.deploy_workflow,
+            **broker_args
+        )
+        timeout = (1200 + delay) * retry_limit
+        sat = wait_for(
+            vmb.checkout, func_kwargs=broker_args, timeout=timeout, delay=delay, fail_condition=[]
+        )
+        return sat.out
+
+    return factory
+
+
+@pytest.fixture(scope='session')
+def capsule_factory():
+    def factory(retry_limit=3, delay=300, **broker_args):
+        vmb = VMBroker(
+            host_classes={'host': Satellite}, workflow='deploy-sat-capsule', **broker_args
+        )
+        timeout = (1200 + delay) * retry_limit
+        cap = wait_for(
+            vmb.checkout, func_kwargs=broker_args, timeout=timeout, delay=delay, fail_condition=[]
+        )
+        return cap.out
+
+    return factory
 
 
 @pytest.fixture
@@ -46,7 +80,9 @@ def rhel77_contenthost_class(request):
 @pytest.fixture
 def satellite_latest():
     """A fixture that provides a latest Satellite"""
-    with VMBroker(host_classes={'host': Satellite}, workflow='deploy-sat-lite') as sat:
+    with VMBroker(
+        host_classes={'host': Satellite}, workflow=settings.server.deploy_workflow
+    ) as sat:
         yield sat
 
 
--- a/pytest_fixtures/xdist.py
+++ b/pytest_fixtures/xdist.py
@@ -1,13 +1,52 @@
 """Fixtures specific to or relating to pytest's xdist plugin"""
-import pytest
+import logging
+import random
 
+import pytest
+from broker.broker import VMBroker
+
+import robottelo
 from robottelo.config import settings
+
+logger = logging.getLogger('robottelo')
 
 
 @pytest.fixture(scope="session", autouse=True)
-def align_xdist_satellites(worker_id):
-    """Set a different Satellite per worker when available in robottelo's config"""
-    settings.configure()
-    settings.server.hostname = settings.server.get_hostname(worker_id)
-    settings.configure_nailgun()
+def align_to_satellite(worker_id, satellite_factory):
+    """Attempt to align a Satellite to the current xdist worker"""
+    cache_proxy = robottelo.config.settings_proxy._cache
+    # clear any hostname that may have been previously set
+    cache_proxy['server.hostname'] = on_demand_sat = None
+
+    if worker_id in ['master', 'local']:
+        worker_pos = 0
+    else:
+        worker_pos = int(worker_id.replace('gw', ''))
+
+    # attempt to add potential satellites from the broker inventory file
+    if settings.server.inventory_filter:
+        settings.server.hostnames  # need to prime the cache
+        hosts = VMBroker().from_inventory(filter=settings.server.inventory_filter)
+        cache_proxy['server.hostnames'].extend([host.hostname for host in hosts])
+
+    # attempt to align a worker to a satellite
+    if settings.server.xdist_behavior == 'run-on-one' and settings.server.hostnames:
+        cache_proxy['server.hostname'] = settings.server.hostnames[0]
+    elif settings.server.hostnames and worker_pos < len(settings.server.hostnames):
+        cache_proxy['server.hostname'] = settings.server.hostnames[worker_pos]
+    elif settings.server.xdist_behavior == 'balance' and settings.server.hostnames:
+        cache_proxy['server.hostname'] = random.choice(settings.server.hostnames)
+    # get current satellite information
+    elif settings.server.xdist_behavior == "on-demand":
+        on_demand_sat = satellite_factory()
+        if on_demand_sat.hostname:
+            cache_proxy['server.hostname'] = on_demand_sat.hostname
+        # if no satellite was received, fallback to balance
+        if not settings.server.hostname:
+            cache_proxy['server.hostname'] = random.choice(settings.server.hostnames)
+    logger.info(f"xdist worker {worker_id} " f"was assigned hostname {settings.server.hostname}")
     settings.configure_airgun()
+    settings.configure_nailgun()
+    yield
+    if on_demand_sat and settings.server.auto_checkin:
+        VMBroker().checkin(host=on_demand_sat)
--- a/robottelo/config/validators.py
+++ b/robottelo/config/validators.py
@@ -6,7 +6,14 @@ from robottelo.constants import VALID_GCE_ZONES
 
 validators = dict(
     server=[
-        Validator("server.hostname", must_exist=True),
+        Validator("server.hostname", must_exist=False),
+        Validator("server.hostnames", must_exist=True, is_type_of=list),
+        Validator("server.version.release", must_exist=True),
+        Validator("server.version.source", must_exist=True),
+        Validator(
+            "server.xdist_behavior", must_exist=True, is_in=['run-on-one', 'balance', 'on-demand']
+        ),
+        Validator("server.auto_checkin", default=False, is_type_of=bool),
         (
             Validator("server.ssh_key", must_exist=True)
             | Validator("server.ssh_password", must_exist=True)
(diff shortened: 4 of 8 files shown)


JacobCallahan:
This depends on #8246 and #8300 
Once those are finalized, changes to this PR are likely needed.

omkarkhatavkar:

tests/upgrades/conftest.py
```diff
@@ -259,7 +259,10 @@ def __initiate(config):
         for upgrade_mark in (PRE_UPGRADE_MARK, POST_UPGRADE_MARK)
         if upgrade_mark in config.option.markexpr
     ]:
-        raise OptionMarksError('options error: pre_upgrade or post_upgrade marks must be selected')
+        pytest.skip(
+            'options error: pre_upgrade or post_upgrade marks must be selected',
+            allow_module_level=True,
+        )
```
JacobCallahan, line 265: optional, but more of a way to solve annoyances with local testing.

conf/server.yaml.template
```diff
@@ -1,6 +1,29 @@
...
+    # RELEASE: populate with satellite version
+    # The snap version currently testing (if applicable)
+    # SNAP:
+    # The source of Satellite packages. Can be one of:
+    # internal, ga, beta
+    SOURCE: "internal"
+  # run-on-one - All xdist runners default to one satellite
+  # balance - xdist runners will be split between available satellites
+  # on-demand - any xdist runner without a satellite will have a new one provisioned.
+  XDIST-BEHAVIOR: "run-on-one"
```
mshriver, line 18: I believe all other dynaconf yamls have been `_` delineated, and we should stay consistent.

mshriver: Interestingly, dynaconf seems to convert this automatically anyway. hmmmm I'm not as sure that we need to be consistent.

pytest_fixtures/broker.py
```diff
@@ -1,12 +1,43 @@
...
 
 from robottelo.constants import BROKER_RHEL77
 from robottelo.hosts import Capsule
 from robottelo.hosts import ContentHost
 from robottelo.hosts import Satellite
 
 
+@pytest.fixture(scope="session")
+def satellite_factory():
+    def factory(retry_limit=3, delay=300, **broker_args):
```
mshriver, line 13: I think the function signature and behavior can be simplified to just take a `timeout` and pass it to wait_for.

pytest_fixtures/broker.py
```diff
@@ -1,12 +1,43 @@
...
 from broker.broker import VMBroker
+from wait_for import wait_for
 
 from robottelo.constants import BROKER_RHEL77
 from robottelo.hosts import Capsule
 from robottelo.hosts import ContentHost
 from robottelo.hosts import Satellite
 
 
+@pytest.fixture(scope="session")
```
mshriver, line 11: What's the benefit of wrapping this function in a fixture? It seems like a regular function could be imported and used just the same.

pytest_fixtures/broker.py
```diff
@@ -1,12 +1,43 @@
...
+
+
+@pytest.fixture(scope="session")
+def capsule_factory():
+    def factory(retry_limit=3, sleep=300, **broker_args):
+        vmb = VMBroker(
+            host_classes={'host': Capsule}, workflow='deploy-sat-capsule', **broker_args
+        )
+        retries = 0
+        while not (cap := vmb.checkout()) and (retries < retry_limit):
```
mshriver, line 31: I'd use same wait_for logic in both factory functions here.

Given the similarity, maybe this can be a single function with the first arg being `Satellite | Capsule`?

robottelo/config/validators.py
```diff
@@ -6,7 +6,15 @@
 
 validators = dict(
     server=[
-        Validator("server.hostname", must_exist=True),
+        Validator("server.hostname", must_exist=False),
+        Validator("server.hostnames", must_exist=True, is_type_of=list),
+        Validator("server.version.release", must_exist=True),
```
mshriver, line 11: I question this being must_exist=True.

If you have a host, it should just get pulled from that in order to deploy more sat/cap instances. 

It's only required if you're starting with no hosts and using on-demand, and even then could be considered optional, as you can call the workflows without providing any release version or snap.

robottelo/config/validators.py
```diff
@@ -6,7 +6,15 @@
...
+        Validator("server.version.source", must_exist=True),
+        Validator(
+            "server.xdist_behavior", must_exist=True, is_in=['run-on-one', 'balance', 'on-demand']
+        ),
+        Validator("server.on_demand", must_exist=True, is_type_of=bool),
```
mshriver, line 16: I think this field is out of date with the design now?

robottelo/config/validators.py
```diff
@@ -6,7 +6,15 @@
...
+        Validator("server.auto_checkin", must_exist=True, is_type_of=bool),
```
mshriver, line 17: I think this could be defaulted instead of must_exist?

mshriver (changes requested):
Couple items to discuss and review, looking very solid though!

conf/server.yaml.template
```diff
@@ -1,6 +1,29 @@
...
+  #  - replace.with.satellite.hostname
+  VERSION:
+    # The full release version (6.9.2)
+    # RELEASE: populate with satellite version
+    # The snap version currently testing (if applicable)
+    # SNAP:
+    # The source of Satellite packages. Can be one of:
+    # internal, ga, beta
+    SOURCE: "internal"
+  # run-on-one - All xdist runners default to one satellite
```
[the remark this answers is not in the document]
JacobCallahan: good clarification

conf/server.yaml.template
JacobCallahan: easy enough

pytest_fixtures/broker.py
JacobCallahan: I can go either way with this one. The question for me is more of whether I want users to have to keep in mind how long it takes to check out a Satellite.

pytest_fixtures/broker.py
JacobCallahan: It *might* be slightly more efficient, but that's it. If you want this line gone, I've got no problems with that.

pytest_fixtures/broker.py
JacobCallahan: YOU FOUND THE "HIDDEN" REQUIRED CHANGE!!!!!!!

robottelo/config/validators.py
JacobCallahan: This is more so required for bugzilla checks than satellite checkouts.

robottelo/config/validators.py
JacobCallahan: ahhhh you're right!

robottelo/config/validators.py
JacobCallahan: Yeah I could do that. I'd likely set the default to False, but this would need to be updated by CI to True. We can add that additional logic later for setting auto_checkin to False if x% of tests fail in a session.

conf/server.yaml.template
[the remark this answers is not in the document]
JacobCallahan: fixed

conf/server.yaml.template
JacobCallahan: fixed

pytest_fixtures/broker.py
JacobCallahan: fixed

pytest_fixtures/broker.py
JacobCallahan: fixed

pytest_fixtures/broker.py
JacobCallahan: fixed

robottelo/config/validators.py
JacobCallahan: fixed

robottelo/config/validators.py
JacobCallahan: fixed

pytest_fixtures/broker.py
mshriver: See I pay attention!

mshriver (approved):

pytest_fixtures/xdist.py
```diff
@@ -1,13 +1,52 @@
...
-    """Set a different Satellite per worker when available in robottelo's config"""
-    settings.configure()
-    settings.server.hostname = settings.server.get_hostname(worker_id)
-    settings.configure_nailgun()
+def align_to_satellite(worker_id, satellite_factory):
+    """Attempt to align a Satellite to the current xdist worker"""
+    cache_proxy = robottelo.config.settings_proxy._cache
+    # clear any hostname that may have been previously set
+    cache_proxy['server.hostname'], on_demand_sat = None, None
+    import random
```
jyejare, line 19: All imports at the top of the module plz.

pytest_fixtures/xdist.py
```diff
@@ -1,13 +1,52 @@
...
+
+    if worker_id == 'master':
+        worker_pos = 0
+    else:
+        worker_pos = int(worker_id.replace('gw', ''))
```
jyejare, line 24: Can be a one liner :

```suggestion
    worker_pos = 0 if worker_id == 'master' else int(worker_id.replace('gw', ''))
```

jyejare (approved):
ACK with some non blocking comments OR improvement suggestions !

conf/server.yaml.template
```diff
@@ -1,6 +1,29 @@
...
-  # Server hostname
-  # HOSTNAME: replace-with-satellite-hostname
+  # Server hostnames, a list of one or more dns-resolvable hostnames
+  HOSTNAMES:
+    - null
+  #  - replace.with.satellite.hostname
+  #  - replace.with.satellite.hostname
+  VERSION:
+    # The full release version (6.9.2)
+    # RELEASE: populate with satellite version
```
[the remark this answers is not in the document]
JacobCallahan: since this isn't an actual value, i'll leave it with the suggestion text as-is

conf/server.yaml.template
```diff
@@ -1,6 +1,29 @@
...
+    # The snap version currently testing (if applicable)
+    # SNAP:
+    # The source of Satellite packages. Can be one of:
+    # internal, ga, beta
+    SOURCE: "internal"
+  # run-on-one - All xdist runners default to the first satellite
+  # balance - xdist runners will be split between available satellites
+  # on-demand - any xdist runner without a satellite will have a new one provisioned.
```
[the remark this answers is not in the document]
JacobCallahan: It will indeed delay execution until a satellite is received (if a new one is needed) I can add a line stating this but might be overkill for a setting explanation.

conf/server.yaml.template
```diff
@@ -1,6 +1,29 @@
...
+  XDIST_BEHAVIOR: "run-on-one"
+  # Before attempting a new broker checkout, check the local inventory
+  FROM_INVENTORY: False
+  # A broker filter specifying which hosts can be used
+  INVENTORY_FILTER: "type=satellite"
```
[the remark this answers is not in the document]
JacobCallahan: changed it to a more accurate example.

conf/server.yaml.template
```diff
@@ -1,6 +1,29 @@
...
+  # If one or more Satellites are provisioned,
+  # this setting determines if it will be automatically checked in
+  AUTO_CHECKIN: False
```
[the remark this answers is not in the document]
JacobCallahan: All of these only affect a single user. So the default behavior will remain false in order to lessen the overall load and wait times.
Robottelo can only check-in hosts that were checked out in the same xdist session.
Fully explaining this behavior is too much for this file.

conf/server.yaml.template
```diff
@@ -1,6 +1,29 @@
...
+  # Path to private ssh key to be used when connecting via SSH.
```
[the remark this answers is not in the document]
JacobCallahan: Good catch! This was originally based on your draft template and it looks like the actual setting was moved further down in the file. Removing this line.

pytest_fixtures/xdist.py
```diff
@@ -1,13 +1,52 @@
...
-def align_xdist_satellites(worker_id):
-    """Set a different Satellite per worker when available in robottelo's config"""
-    settings.configure()
-    settings.server.hostname = settings.server.get_hostname(worker_id)
-    settings.configure_nailgun()
+def align_to_satellite(worker_id, satellite_factory):
+    """Attempt to align a Satellite to the current xdist worker"""
+    cache_proxy = robottelo.config.settings_proxy._cache
+    # clear any hostname that may have been previously set
+    cache_proxy['server.hostname'], on_demand_sat = None, None
```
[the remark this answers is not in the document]
JacobCallahan: That's an option. Stylistically I prefer the first, but they're functionally the same.

pytest_fixtures/xdist.py
JacobCallahan: I'll leave as-is for readability.

JacobCallahan:
@jyejare all comments addressed, thanks for the review!

conf/server.yaml.template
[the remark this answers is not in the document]
mshriver: execution only for that xdist worker will be halted. I'd agree its too much to describe here in the in-line comments. Readme or docblock for the hook should describe this behavior in more detail.

pytest_fixtures/broker.py
mshriver: I don't think they should have to consider that or that should be baked into this function's argument structure somehow. It should just be configurable for the amount of time it will wait.

The way you have it now makes an assumption about the amount of time it takes to get a satellite, and would have to be adjusted based on the performance of the broker provider, or things like resource pools cutting down deployment time.

The current default delay is very long, there is no need to wait 5 minutes between attempts to call checkout.

Callers of this function should be able to just pass along the arguments that wait_for uses.

pytest_fixtures/broker.py
```diff
@@ -1,12 +1,41 @@
...
 from robottelo.constants import BROKER_RHEL77
 from robottelo.hosts import Capsule
 from robottelo.hosts import ContentHost
 from robottelo.hosts import Satellite
 
 
+@pytest.fixture(scope='session')
+def satellite_factory():
+    def factory(retry_limit=3, delay=300, **broker_args):
+        vmb = VMBroker(host_classes={'host': Satellite}, workflow='deploy-sat-lite', **broker_args)
```
mshriver, line 14: The sat-lite template is going away, so this should be deploy-sat-jenkins.

That aside, it shouldn't be hard-coded here, this should be read from settings.

mshriver (changes requested):

pytest_fixtures/broker.py
JacobCallahan: I'm going to disagree. It is an easier use model in the current form as the user doesn't need to determine their own estimate for average checkout times.
Adding a config for average checkout times is overkill for this.
A 5 minute delay may be overkill, but since there should only be one reason (SLA limit) for this to fail, 5 minutes may give enough time for another session to complete and perform a checkin.

JacobCallahan:
@mshriver @jyejare all comments should be addressed and code updated where needed.

mshriver (approved):
Agreeing to disagree on some points, but regardless this is a great step forward for robottelo's system under test dynamic control!

jyejare:
@JacobCallahan Merging this one !

PR #8319, opened 2021-02-01
fixing "Taxonomy has already been taken" in various UI host tests

pondrejk:
The core of the problem is that the newly created orgs and locs in satellite are automatically added to some default elements like provisioning templates or smart proxy. Apparently this didn't happen before and our test try to add these taxonomies where they already exist resulting in errors.


omkarkhatavkar:

lpramuk (approved):

mshriver (approved):

PR #8324, opened 2021-02-01
Add temporary reportportal-client requirement

mshriver:
enum34 broke package installs in py38 when pytest-reportportal was added
reportportal-client 5.0.8 fixes enum issue

our direct dependency is pytest-reportportal, reportportal-client should
be installed through it.

This works around pytest-reportportal not yet updating its dependencies
or merging the following PR:
https://github.com/reportportal/agent-python-pytest/pull/247/files

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,6 +17,9 @@ pytest==6.2.2
 pytest-services==2.2.1
 pytest-mock==3.5.1
 pytest-reportportal==5.0.7
+# TODO remove rp-client requirement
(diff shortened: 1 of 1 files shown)


swadeley (approved):
ACK

rplevka (approved):
As discussed on the chat with @mshriver, this might not be really necessary as the requirement version is conditioned by `>=` - that means, `pip` would automatically pull the newest version in a clean environment.

We can still merge this and pin that dependency though to keep control over the version that is being installed

PR #8328, opened 2021-02-03
rex stubs cleanup

pondrejk:
Removing some rusty stubs, some of these probably won't be testable via CLI any time soon (time span/concurrency), some are low priority use cases (puppet/power), some are a combination of high infra requirements and low additional coverage (provisioned hosts)


swadeley (approved):
ACK

mshriver (approved):

PR #8336, opened 2021-02-08
Parameterize ldap auth

akhil-jha:
Result Updated: 23 Feb 2021

Screenshot from 2021-02-22 22-43-23

NOTE:
1) RHSSO tests not included as they are environment depended.
2) HTTPS ldap will fail while executing PRT because of improper certs.

### DIFF
--- a/pytest_fixtures/satellite_auth.py
+++ b/pytest_fixtures/satellite_auth.py
@@ -20,6 +20,7 @@ from robottelo.rhsso_utils import set_the_redirect_uri
 def ad_data():
     return {
         'ldap_user_name': settings.ldap.username,
+        'ldap_user_cn': settings.ldap.username,
         'ldap_user_passwd': settings.ldap.password,
         'base_dn': settings.ldap.basedn,
         'group_base_dn': settings.ldap.grpbasedn,
@@ -46,7 +47,14 @@ def ipa_data():
 
 @pytest.fixture(scope='session')
 def open_ldap_data():
-    return settings.open_ldap
+    return {
+        'ldap_user_name': settings.open_ldap.open_ldap_user,
+        'ldap_user_cn': settings.open_ldap.username,
+        'ldap_hostname': settings.open_ldap.hostname,
+        'ldap_user_passwd': settings.open_ldap.password,
+        'base_dn': settings.open_ldap.base_dn,
+        'group_base_dn': settings.open_ldap.group_base_dn,
+    }
 
 
 @pytest.fixture(scope='function')
@@ -93,8 +101,30 @@ def auth_source_ipa(module_org, module_loc, ipa_data):
     ).create()
 
 
+@pytest.fixture(scope='function')
+def auth_source_open_ldap(module_org, module_loc, open_ldap_data):
+    return entities.AuthSourceLDAP(
+        onthefly_register=True,
+        account=open_ldap_data['ldap_user_cn'],
+        account_password=open_ldap_data['ldap_user_passwd'],
+        base_dn=open_ldap_data['base_dn'],
+        groups_base=open_ldap_data['group_base_dn'],
+        attr_firstname=LDAP_ATTR['firstname'],
+        attr_lastname=LDAP_ATTR['surname'],
+        attr_login=LDAP_ATTR['login'],
+        server_type=LDAP_SERVER_TYPE['API']['posix'],
+        attr_mail=LDAP_ATTR['mail'],
+        name=gen_string('alpha'),
+        host=open_ldap_data['ldap_hostname'],
+        tls=False,
+        port='389',
+        organization=[module_org],
+        location=[module_loc],
+    ).create()
+
+
 @pytest.fixture
-def ldap_auth_source(request, module_org, module_location, ad_data, ipa_data):
+def ldap_auth_source(request, module_org, module_location, ad_data, ipa_data, open_ldap_data):
     if request.param.lower() == 'ad':
         # entity create with AD settings
         entities.AuthSourceLDAP(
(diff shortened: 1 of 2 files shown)


pytest_fixtures/satellite_auth.py
```diff
@@ -20,6 +20,7 @@
 def ad_data():
     return {
         'ldap_user_name': settings.ldap.username,
+        'ldap_user_cn': settings.ldap.username,
```
omkarkhatavkar, line 23: Why we need to duplicate same value ? In case if you still want to use it make the name as  `ldap_account_name`

pytest_fixtures/satellite_auth.py
```diff
@@ -46,7 +47,14 @@ def ipa_data():
 
 @pytest.fixture(scope='session')
 def open_ldap_data():
-    return settings.open_ldap
+    return {
+        'ldap_user_name': settings.open_ldap.open_ldap_user,
+        'ldap_user_cn': settings.open_ldap.username,
```
omkarkhatavkar, line 52: Same as above

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -47,6 +46,8 @@
 from robottelo.rhsso_utils import update_rhsso_user
 from robottelo.utils.issue_handlers import is_open
 
+# from robottelo.decorators import skip_if_not_set
```
omkarkhatavkar, line 49: seems like this is not required

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -184,20 +186,31 @@ def test_positive_end_to_end_ad(session, ldap_tear_down, ad_data):
...
     """
+    if ldap_auth_source['auth_type'] == 'ipa':
+        ldap_server_type = LDAP_SERVER_TYPE['UI']['ipa']
+        attr_login = LDAP_ATTR['login']
+    elif ldap_auth_source['auth_type'] == 'ad':
+        ldap_server_type = LDAP_SERVER_TYPE['UI']['ad']
+        attr_login = LDAP_ATTR['login_ad']
+    else:
+        ldap_server_type = LDAP_SERVER_TYPE['UI']['posix']
+        attr_login = LDAP_ATTR['login']
```
omkarkhatavkar, line 200: Above code is repeatable in the multiple tests, it would be good if can place this in the ldap_auth_source fixture itself. 
e.g.
```python
        if request.param.lower() == 'ipa':
            ldap_data['ldap_server_type'] = LDAP_SERVER_TYPE['UI']['ipa']
```

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -423,31 +349,36 @@ def test_positive_create_with_ad_https(session, test_name, ldap_tear_down, ad_da
...
-            ldapsession.usergroup.search('')
-        assert ldapsession.task.read_all()['current_user'] == settings.ldap.username
+            ldapsession.user.search('')
+    users = entities.User().search(
+        query={'search': 'login="{}"'.format(ldap_auth_source['ldap_user_name'])}
+    )
+    assert users[0].login == ldap_auth_source['ldap_user_name']
 
 
+@pytest.mark.parametrize("ldap_auth_source", ["AD"], indirect=True)
```
omkarkhatavkar, line 363: we can use IPA here as well ?

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -465,19 +396,22 @@ def test_positive_add_katello_role(
...
         with pytest.raises(NavigationTriesExceeded):
             session.architecture.search('')
         session.activationkey.create({'name': ak_name})
         assert session.activationkey.search(ak_name)[0]['Name'] == ak_name
         current_user = session.activationkey.read(ak_name, 'current_user')['current_user']
-        assert ad_data['ldap_user_name'] in current_user
+        assert ldap_auth_source['ldap_user_name'] in current_user
 
 
+@pytest.mark.parametrize("ldap_auth_source", ["AD"], indirect=True)
```
omkarkhatavkar, line 410: Can we use IPA here as well ?

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -586,18 +527,21 @@ def test_positive_delete_external_roles(
...
         )
-    with Session(test_name, ad_data['ldap_user_name'], ad_data['ldap_user_passwd']) as ldapsession:
+    with Session(
+        test_name, ldap_auth_source['ldap_user_name'], ldap_auth_source['ldap_user_passwd']
+    ) as ldapsession:
         with pytest.raises(NavigationTriesExceeded):
             ldapsession.location.create({'name': gen_string('alpha')})
 
 
+@pytest.mark.parametrize("ldap_auth_source", ["AD"], indirect=True)
```
omkarkhatavkar, line 541: Can we use IPA here as well ?

omkarkhatavkar (changes requested):
Given some minor comments, let me know your thoughts

pytest_fixtures/satellite_auth.py
akhil-jha: For ad, username and account name are the same but for openldap and ipa it's not. So while making the tests parameterize, different values were getting passed than required. So in order to make it same I duplicated the value.

pytest_fixtures/satellite_auth.py
akhil-jha: They are not same.

tests/foreman/ui/test_ldap_authentication.py
akhil-jha: I could remove that but I as it was used in the file and was essential, I thought of keeping it commented.

tests/foreman/ui/test_ldap_authentication.py
akhil-jha: Done

tests/foreman/ui/test_ldap_authentication.py
akhil-jha: Done

tests/foreman/ui/test_ldap_authentication.py
akhil-jha: Done.

tests/foreman/ui/test_ldap_authentication.py
akhil-jha: Done.

tests/foreman/ui/test_ldap_authentication.py
mshriver: No value to leaving commented imported statements, please remove.  These are providing no context as to how/where/why they _might_ be needed.

pytest_fixtures/satellite_auth.py
```diff
@@ -93,8 +101,30 @@ def auth_source_ipa(module_org, module_loc, ipa_data):
...
+        tls=False,
+        port='389',
+        organization=[module_org],
+        location=[module_loc],
+    ).create()
+
+
 @pytest.fixture
-def ldap_auth_source(request, module_org, module_location, ad_data, ipa_data):
+def ldap_auth_source(request, module_org, module_location, ad_data, ipa_data, open_ldap_data):
```
mshriver, line 127: It looks like every test function using `ldap_auth_source` is also using `ldap_tear_down` fixture too.  I would include it here, instead of the functions having to include it.  Enforce, through this fixture's use, that auth servers are torn down after the test using them is complete - at the scope that this fixture is configuring them in the first place.

mshriver (approved):
I have some extensive experience parametrizing authentication provider test coverage, so there are some larger structural conversations that I'd like to have outside the scope of this PR.

That being said, what's being added here looks pretty good, and I just had one nitpick comment about fixture use.

akhil-jha:
New results: 
Screenshot from 2021-03-03 21-33-05

mshriver:
@akhil-jha thanks for sharing a results screenshot! I'd encourage you to also use robottelo-runner to execute those tests in the new CI.

Please see internal docs for how to trigger. You'll make a comment that looks like what's quoted below, but I'm not putting the exact trigger phrase so that this comment isn't picked up.

```
trigger: test-rob-ottelo (remove the dash)
pytest: tests/foreman/path/to_module.py -k some_filter_if_you_want <other options>
```

pytest_fixtures/satellite_auth.py
akhil-jha: So for those test which use `ldap_auth_source`, we could for sure include `ldap_tear_down`there. Since some of the tests do not use `ldap_auth_source`  e.g `test_email_of_the_user_should_be_copied` which rely on the fixture `ldap_tear_down` which is in the top of the file, I rather make a new PR and modify there.

PS I tried moving the fixture from the test_ldap_authentication.py to pytest_fixtures/satellite_auth.py. Got index error and ergo I got irritated. :/

akhil-jha:
> @akhil-jha thanks for sharing a results screenshot! I'd encourage you to also use robottelo-runner to execute those tests in the new CI.
> 
> Please see internal docs for how to trigger. You'll make a comment that looks like what's quoted below, but I'm not putting the exact trigger phrase so that this comment isn't picked up.
> 
> ```
> trigger: test-rob-ottelo (remove the dash)
> pytest: tests/foreman/path/to_module.py -k some_filter_if_you_want <other options>
> ```

The problem with that is HTTPS ldap will fail because of improper certs, and I rather not skip those tests as they are end-to-end ones. Also RHSSO test is environment dependent, that will also fail.

omkarkhatavkar (approved):
ACK, Nice Work Akhil !

mshriver (approved):

PR #8340, opened 2021-02-09
Disable Report Portal parameter logging

rplevka:
Due the way the parametrization of some robottelo tests is implemented, we have to apply a workaround to change the report portal logger plugin behavior, so it doesn't log the used parameter in the metadata.

This used to result in various parameters to be logged across reruns and was causing RP not to match some tests and not marking them as re-runs.

rp_scrshot

### DIFF
--- a/pytest_plugins/disable_rp_params.py
+++ b/pytest_plugins/disable_rp_params.py
@@ -1,0 +1,13 @@
+import pytest
+
+
+@pytest.hookimpl(trylast=True)
+def pytest_configure(config):
+    """Override Report Portal (RP) service objects attribute
+    to prevent it from posting the parameter values of the tests.
+    This is to make RP to correctly match and handle the test re-runs.
+
+    We can remove this once we implement a compatible parametrization
+    """
+    if hasattr(config, 'py_test_service'):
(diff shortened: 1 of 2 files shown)


pytest_plugins/disable_rp_params.py
```diff
@@ -0,0 +1,13 @@
...
+@pytest.hookimpl(trylast=True)
+def pytest_configure(config):
+    '''Override Report Portal (RP) service objects attribute
+    to prevent it from posting the parameter values of the tests.
+    This is to make RP to correctly match and handle the test re-runs.
+
+    We can remove this once we implement a compatible parametrization
+    '''
+    if hasattr(config, 'py_test_service'):
+        config.py_test_service.rp_supports_parameters = False
```
mshriver, line 13: _sneaky sneaky sneaky_

mshriver (approved):

mshriver:
Approved... but fix formatting.

pytest_plugins/disable_rp_params.py
JacobCallahan: i love it!!!

JacobCallahan (approved):

PR #8344, opened 2021-02-10
Trying to fix this with selenium native method

omkarkhatavkar:
Trying to fix this with selenium native method. Currently, this is failing with CI and this PR intention to fix this and test with PRT so no more CI issue. 

Failed in CI with below error but passes locally 
```

19:08:23  >               assert ldapsession.location.search(location_name)[0]['Name'] == location_name
19:08:23  E               IndexError: list index out of range
19:08:23  
19:08:23  tests/foreman/ui/test_ldap_authentication.py:587: IndexError
```

### DIFF
--- a/tests/foreman/ui/test_ldap_authentication.py
+++ b/tests/foreman/ui/test_ldap_authentication.py
@@ -523,8 +523,11 @@ def test_positive_update_external_roles(
                 ldapsession.architecture.search('')
             ldapsession.location.create({'name': location_name})
             if is_open('BZ:1851905'):
-                ldapsession.browser.execute_script("window.history.go(-1)")
-            assert ldapsession.location.search(location_name)[0]['Name'] == location_name
+                current_url = ldapsession.browser.url
+                base_url = current_url.rsplit('/', 3)[0]
+                ldapsession.browser.selenium.get(base_url)
+            location = ldapsession.location.search(location_name)
+            assert location[0]['Name'] == location_name
             current_user = ldapsession.location.read(location_name, 'current_user')['current_user']
             assert ad_data['ldap_user_name'] in current_user
         session.usergroup.update(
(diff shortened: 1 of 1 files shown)


omkarkhatavkar:
trigger: test-robottelo
pytest: -k 'test_positive_update_external_roles or test_positive_delete_external_roles' tests/foreman/ui/

omkarkhatavkar:
trigger: test-robottelo
pytest: -k 'test_positive_update_external_roles or test_positive_delete_external_roles' tests/foreman/ui/

mshriver:
trigger: test-robottelo
pytest: -k 'test_positive_update_external_roles or test_positive_delete_external_roles' tests/foreman/ui/

omkarkhatavkar:
trigger: test-robottelo
pytest: -k 'test_positive_update_external_roles or test_positive_delete_external_roles' tests/foreman/ui/

omkarkhatavkar:
trigger: test-robottelo
pytest: -k 'test_positive_update_external_roles or test_positive_delete_external_roles' tests/foreman/ui/

tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -583,7 +584,8 @@ def test_positive_delete_external_roles(
                 ldapsession.architecture.search('')
             ldapsession.location.create({'name': location_name})
             if is_open('BZ:1851905'):
-                ldapsession.browser.execute_script("window.history.go(-1)")
+                ldapsession.browser.selenium.back()
```
mshriver, line 587: We can find a better approach to working around BZ 1851905.

I would suggest navigating to the page you expect to be on with `navigate_to` and assert against that fresh view.

It doesn't appear that there is a need to specifically test and replicate a workaround that's within the BZ and expected to be used.

The test is failing in PRT - with nothing returned by `search(location_name)` on the next line, so I don't believe this is resolving the failure.

mshriver (changes requested):
I don't believe this change fixes the test failure, see individual comment.

tests/foreman/ui/test_ldap_authentication.py
omkarkhatavkar: Yeah, I'm looking for the fix and trying multiple options. Thanks for shifting draft. Once the PRT give okay this will be good

omkarkhatavkar: @mshriver `navigate_to` itself will not work because we are on the page which is blank and has no way it will get navigate it's own because there not the menu on the left side. I think the possible solution would be navigate using url like to bas url and then start again.

omkarkhatavkar:
trigger: test-robottelo
pytest: -k 'test_positive_delete_external_roles' tests/foreman/ui/

omkarkhatavkar:
trigger: test-robottelo
pytest: -k 'test_positive_update_external_roles or test_positive_delete_external_roles' tests/foreman/ui/

omkarkhatavkar:
trigger: test-robottelo
pytest: -k 'test_positive_update_external_roles or test_positive_delete_external_roles' tests/foreman/ui/

PR #8352, opened 2021-02-15
skip rename on different bz

pondrejk:
rename issue received a spin-off bug, modified skip accordingly


swadeley (approved):
ACK, but I think also add UUID to QA Whiteboard not just path to test

mshriver (approved):

Issue #2138, opened 2015-04-28
Automation failure : testnames are given below


tkolhar:
This tests is failing for nightly:
tests.foreman.cli.test_host_collection.TestHostCollection.test_positive_update_1
tests.foreman.cli.test_host_collection.TestHostCollection.test_positive_update_2
tests.foreman.cli.test_host_collection.TestHostCollection.test_positive_update_3

 hammer -v -u admin -p changeme --output=csv host-collection update --max-content-hosts="17" --id="21" --organization-id="213"
[ERROR 2015-04-26 21:55:32 Exception] Error: Unrecognised option '--organization-id'

See: 'hammer host-collection update --help'
Could not update the the host collection:
  Error: Unrecognised option '--organization-id'
  
  See: 'hammer host-collection update --help'
[ERROR 2015-04-26 21:55:32 Exception] 

Clamp::UsageError (Unrecognised option '--organization-id'):
    /usr/lib/ruby/gems/1.8/gems/clamp-0.6.2/lib/clamp/option/parsing.rb:62:in `find_option'
    /usr/lib/ruby/gems/1.8/gems/clamp-0.6.2/lib/clamp/option/parsing.rb:28:in `parse_options'
    /usr/lib/ruby/gems/1.8/gems/clamp-0.6.2/lib/clamp/command.rb:53:in `parse'
    /usr/lib/ruby/gems/1.8/gems/hammer_cli-0.1.4/lib/hammer_cli/./apipie/../abstract.rb:31:in `parse'
    /usr/lib/ruby/gems/1.8/gems/clamp-0.6.2/lib/clamp/command.rb:67:in `run'
    /usr/lib/ruby/gems/1.8/gems/hammer_cli-0.1.4/lib/hammer_cli/./apipie/../abstract.rb:23:in `run'
    /usr/lib/ruby/gems/1.8/gems/clamp-0.6.2/lib/clamp/subcommand/execution.rb:11:in `execute'
    /usr/lib/ruby/gems/1.8/gems/clamp-0.6.2/lib/clamp/command.rb:68:in `run'
    /usr/lib/ruby/gems/1.8/gems/hammer_cli-0.1.4/lib/hammer_cli/./apipie/../abstract.rb:23:in `run'
    /usr/lib/ruby/gems/1.8/gems/clamp-0.6.2/lib/clamp/subcommand/execution.rb:11:in `execute'
[...]

PR #8365, opened 2021-02-18
Pytest conversion for content-view cli

latran:
Converted contentview view file for cli.  There will be a few failures due to some test cases not being fixed properly, but the goal here is ensure that the pytest conversion is correct.  I will open separate PRs for any failed test cases that follows after this PR is merged since this a very large file to review.  I will add test results on here soon.

### DIFF
--- a/tests/foreman/cli/test_contentview.py
+++ b/tests/foreman/cli/test_contentview.py
@@ -24,7 +24,6 @@ from fauxfactory import gen_alphanumeric
 from fauxfactory import gen_string
 from nailgun import entities
 
-from robottelo import manifests
 from robottelo import ssh
 from robottelo.api.utils import create_sync_custom_repo
 from robottelo.cli.base import CLIReturnCodeError
@@ -53,7 +52,6 @@ from robottelo.cli.puppetmodule import PuppetModule
 from robottelo.cli.repository import Repository
 from robottelo.cli.repository_set import RepositorySet
 from robottelo.cli.role import Role
-from robottelo.cli.subscription import Subscription
 from robottelo.cli.user import User
 from robottelo.config import settings
 from robottelo.constants import DEFAULT_CV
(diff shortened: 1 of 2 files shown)


tests/foreman/cli/test_contentview.py
```diff
@@ -77,122 +75,159 @@
...
-            )
-        except CLIReturnCodeError:
-            # Make sure to reset rhel_content_org and let the exception
-            # propagate.
-            ContentViewTestCase.rhel_content_org = None
-            raise
+def _get_content_view_version_lce_names_set(content_view_id, version_id):
+    """returns a set of content view version lifecycle environment names
+    :rtype: set
+    that it belongs under
```
mshriver, line 201: unintentional newline here

tests/foreman/cli/test_contentview.py
```diff
@@ -1310,17 +1297,17 @@ def test_positive_add_puppet_module(self):
         )
         # Check output of `content-view info` subcommand
         content_view = ContentView.info({'id': content_view['id']})
-        self.assertGreater(len(content_view['puppet-modules']), 0)
-        self.assertEqual(puppet_module['name'], content_view['puppet-modules'][0]['name'])
+        assert len(content_view['puppet-modules']) > 0
```
mshriver, line 1300: Using `assert len() > 0 ` is redundant, because `len` will return 0 (false) or greater than zero (true)

tests/foreman/cli/test_contentview.py
```diff
@@ -4809,74 +4653,64 @@ def test_positive_publish_promote_with_RH_ostree_and_other(self):
...
 
-    @classmethod
-    def setUpClass(cls):
-        """Create a product and an org which can be re-used in tests."""
-        super().setUpClass()
-        cls.org = make_org()
-        cls.product = make_product({'organization-id': cls.org['id']})
-
-    def _make_file_repository_upload_contents(self, options=None):
+    def _make_file_repository_upload_contents(self, module_org, module_product, options=None):
```
mshriver, line 4708: Given that these are pytest classes and not library classes, you don't necessarily need to name this as if it were a private method.

tests/foreman/cli/test_contentview.py
```diff
@@ -4917,27 +4751,27 @@ def test_positive_arbitrary_file_repo_addition(self):
 
         :expectedresults: Check FR is added to CV
 
-        :CaseAutomation: NotAutomated
+        :CaseAutomation: notautomated
```
mshriver, line 4754: Double check your testimony tokens here. The case is now enforced, and is causing failures in the GH workflow.

Second - this one is wrong! Not sure when these were automated and this was originally missed, but now's as good a time as ever to correct them.

I would recommend setting `:CaseAutomation: Automated` at the module level, and setting `NotAutomated` specifically on stubbed cases.

mshriver (commented):
Conversion looks great overall. A few nitpick/optional comments, the biggest issue is resolving the testimony tokens and auditing them for accuracy.

mshriver:
trigger: test-robottelo
pytest: tests/foreman/cli/test_contentview.py

tests/foreman/cli/test_contentview.py
latran: @mshriver so this should just be removed?

tests/foreman/cli/test_contentview.py
latran: True, but I'm just keeping what was originally there and making as little changes to what was there already.  I can still change this.

tests/foreman/cli/test_contentview.py
mshriver: Its optional and up to you, but worth mentioning. 

The types in python have a truthy-ness to them. If `content_view['puppet-modules']` is a list, you don't even have to call `len()`, you can assert directly against the list.  If the list is empty the assertion will fail, as an empty list is 'falsy'. If there are any elements in the list at all, the assertion will pass.  Going that far makes an assumption that `content_view['puppet-modules']` is a list, but the same assertion logic applies if its a dictionary, or a string - an empty dict or string will fail the direct assert, a dictionary with anything in it, or a string of any length (even whitespace) will pass the assertion.

You may find it more valuable to have a fully explicit assertion here against the length, but personally I would write my assertions to be direct here.

```In [6]: assert [], 'list was empty'                                             
AssertionError                            Traceback (most recent call last)
<ipython-input-6-d507d26ab449> in <module>
----> 1 assert [], 'list was empty'

AssertionError: list was empty
```

```
In [9]: assert [False], 'list was empty' 
```

tests/foreman/cli/test_contentview.py
mshriver: Its optional, but was worth mentioning.

latran:
Here's the result run and a few key points are worth mentioning below.

`=========================== short test summary info ============================
FAILED tests/foreman/cli/test_contentview.py::TestContentView::test_positive_create_with_name[12]
FAILED tests/foreman/cli/test_contentview.py::TestContentView::test_positive_update_name_by_id[12]
FAILED tests/foreman/cli/test_contentview.py::TestContentView::test_positive_update_name_by_name[12]
FAILED tests/foreman/cli/test_contentview.py::TestContentView::test_positive_remove_cv_version_from_multi_env_capsule_scenario
= 4 failed, 129 passed, 9 skipped, 6 deselected, 176 warnings in 6401.31s (1:46:41) =
Build step 'Virtualenv Builder' marked build as failure
Set build name.
New build name is '#2138 '
Deleting 1 temporary files
Recording test results
Archiving artifacts
Started calculate disk usage of build
Finished Calculation of disk usage of build in 0 seconds
Started calculate disk usage of workspace
Finished Calculation of disk usage of workspace in 0 seconds
Notifying upstream projects of job completion
Finished: FAILURE
`

So this failed because of the bash escape character for quotes.  See below.  This is a minor issue that is a headache to resolve that I haven't found a suitable solution to.  It's good to be aware of this issue though.  You can find the offending data entry here: https://github.com/SatelliteQE/robottelo/blob/master/robottelo/datafactory.py#L396
```
E           robottelo.cli.factory.CLIFactoryError: Failed to create ContentView with data:
E           { 'component-ids': None,
E             'composite': False,
E             'description': None,
E             'label': None,
E             'name': 'bar+{}|"?hi 뷏徃죌𞹼諸𐦤𖥾햦갥扐',
E             'organization': None,
E             'organization-id': 123,
E             'organization-label': None,
E             'product': None,
E             'product-id': None,
E             'repositories': None,
E             'repository-ids': None}
E           Command "content-view create" finished with return_code 1
E           stderr contains:
E           bash: -c: line 0: unexpected EOF while looking for matching `"'
E           bash: -c: line 1: syntax error: unexpected end of file
```

The other failure is `tests/foreman/cli/test_contentview.py::TestContentView::test_positive_remove_cv_version_from_multi_env_capsule_scenario` which has always failed due to capsule issues and I have a PR associated with it: #7756

Those are the only two failures that shouldn't be the deciding factor in not merging this PR if my pytest conversion looks good overall.  The rest of the tests are passing.

latran:
> Conversion looks great overall. A few nitpick/optional comments, the biggest issue is resolving the testimony tokens and auditing them for accuracy.

@mshriver should be resolved now.

mshriver (approved):

tests/foreman/cli/test_contentview.py
```diff
@@ -5017,20 +4851,16 @@ def test_positive_arbitrary_file_repo_promotion(self):
         :CaseLevel: Integration
 
         :CaseImportance: High
+
+        :parametrized: yes
```
JacobCallahan, line 4855: i don't see any parameterization here

tests/foreman/cli/test_contentview.py
```diff
@@ -2925,24 +2845,24 @@ def test_positive_sub_host_with_restricted_user_perm_at_custom_loc(self):
...
         # assert that this is the same content view
-        self.assertEqual(content_view['name'], user_content_view['name'])
+        assert content_view['name'] == user_content_view['name']
         # create a client host and register it with the created user
         with VirtualMachine(distro=DISTRO_RHEL7) as host_client:
             host_client.install_katello_ca()
             host_client.register_contenthost(
                 org['label'],
-                lce=f'{env["name"]}/{content_view["name"]}',
+                lce='{}/{}'.format(env['name'], content_view['name']),
```
JacobCallahan, line 2854: can you keep this as an f-string?

tests/foreman/cli/test_contentview.py
```diff
@@ -79,122 +77,159 @@
...
-from robottelo.ssh import upload_file
-from robottelo.test import CLITestCase
 from robottelo.vm import VirtualMachine
 from robottelo.vm_capsule import CapsuleVirtualMachine
 
 
-class ContentViewTestCase(CLITestCase):
-    """Content View CLI tests"""
+@pytest.fixture(scope='module')
+def module_rhel_content(module_manifest_org):
```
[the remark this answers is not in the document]
latran: Global as in put it under the `pytest_fixture` file?  If so, then not really since this is only specific to cv.

Global to just this file?  If so, then not really since only the class `Test ContentView` uses it.  In the future, if we add more code coverage that requires this in another class, then sure.

tests/foreman/cli/test_contentview.py
latran: My bad.  Should be fixed now.

tests/foreman/cli/test_contentview.py
latran: I could...

And I did.

tests/foreman/cli/test_contentview.py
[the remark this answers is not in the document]
JacobCallahan: A few other components could use a fixture that provides a sync'd repo. I do think it is globally useful :)

JacobCallahan (approved):

PR #8368, opened 2021-02-19
Fix xdist worker logging

mshriver:
Validated against xdist and single worker session, logs are now correctly written, and capture all of the nailgun/airgun client logging.

Default show-capture=no, because everything is written to logs. This was causing absurdly verbose logging to the console at the end of the test session.

### DIFF
--- a/tests/foreman/conftest.py
+++ b/tests/foreman/conftest.py
@@ -54,34 +54,23 @@ def pytest_report_header(config):
 
 
 @pytest.fixture(scope="session")
-def worker_id(request):
-    """Gets the worker ID when running in multi-threading with xdist"""
-    if hasattr(request.config, 'slaveinput'):
-        # return gw+(0..n)
-        return request.config.slaveinput['slaveid']
-    else:
-        return 'master'
-
-
-@pytest.fixture(scope="session")
 def configured_settings():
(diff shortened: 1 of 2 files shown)


tests/foreman/conftest.py
```diff
@@ -53,35 +53,24 @@ def pytest_report_header(config):
     return messages
 
 
-@pytest.fixture(scope="session")
-def worker_id(request):
```
mshriver, line 57: This is provided by xdist, we don't have to implement it.

tests/foreman/conftest.py
```diff
@@ -53,35 +53,24 @@ def pytest_report_header(config):
...
-
 @pytest.fixture(scope="session")
 def configured_settings():
     if not settings.configured:
         settings.configure()
     return settings
 
 
-@pytest.fixture(autouse=True, scope='module')
+@pytest.fixture(autouse=True, scope='session')
```
mshriver, line 63: Simply changing this fixture scope resolved the fact that this was not being executed on the xdist workers.

mshriver:
trigger: test-robottelo

JacobCallahan (approved):
ACK

PR #8369, opened 2021-02-19
Add fixture to disable BFA

mshriver:
Remove unused test_name fixture as well.

This used to be done via automation_tools against test targets before robottelo was executed.

Its necessary for robottelo to run effectively, and should be accomplished within the test session.

The scope on this blew up, the more I dug, the more I saw that should be changed.

### DIFF
--- a/tests/foreman/conftest.py
+++ b/tests/foreman/conftest.py
@@ -109,7 +109,6 @@ def test_name(request):
 
         tests.foreman.ui.test_activationkey::test_positive_create
         tests.foreman.api.test_errata::TestErrata::test_positive_list
-
     """
     return request.node._nodeid
 
--- a/conftest.py
+++ b/conftest.py
@@ -10,15 +10,15 @@ pytest_plugins = [
     "pytest_plugins.disable_rp_params",
     # Fixtures
     "pytest_fixtures.api_fixtures",
-    "pytest_fixtures.xdist",
     "pytest_fixtures.broker",
+    "pytest_fixtures.xdist",
     # Component Fixtures
-    "pytest_fixtures.satellite_auth",
-    "pytest_fixtures.sys_fixtures",
-    "pytest_fixtures.templatesync_fixtures",
     "pytest_fixtures.ansible_fixtures",
     "pytest_fixtures.oscap_fixtures",
+    "pytest_fixtures.satellite_auth",
+    "pytest_fixtures.sys_fixtures",
     "pytest_fixtures.smartproxy_fixtures",
+    "pytest_fixtures.templatesync_fixtures",
     "pytest_fixtures.user_fixtures",
     "pytest_fixtures.upgrade_fixtures",
 ]
--- a/pytest_fixtures/sys_fixtures.py
+++ b/pytest_fixtures/sys_fixtures.py
@@ -1,6 +1,8 @@
 import pytest
 
+from robottelo.config import settings
 from robottelo.rhsso_utils import run_command
+from robottelo.ssh import command as ssh_command
 
 
 @pytest.fixture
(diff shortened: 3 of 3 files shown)


mshriver:
trigger: test-robottelo

tests/foreman/conftest.py
```diff
@@ -114,18 +115,16 @@ def log_test_execution(robottelo_logger, test_name):
     robottelo_logger.debug(f'Finished Test: {test_name}')
 
 
-@pytest.fixture
-def test_name(request):
```
mshriver, line 118: This was unused, and not helpful, as its a one line call that can be made by any test or fixture requiring it.

tests/foreman/conftest.py
```diff
@@ -114,18 +115,16 @@ def log_test_execution(robottelo_logger, test_name):
...
-        tests.foreman.ui.test_activationkey::test_positive_create
-        tests.foreman.api.test_errata::TestErrata::test_positive_list
+@pytest.fixture(autouse=True, scope="session")
+def relax_bfa():
+    """Relax BFA protection against failed login attempts
 
+    The robottelo.ssh.command should resolve the hostname for the current xdist worker
     """
-    return request.node._nodeid
+    ssh_command(
```
mshriver, line 124: Worth changing it back after the session?

mshriver:
trigger: test-robottelo

lpramuk (approved):

tests/foreman/conftest.py
mshriver: not quite unused... making some updates to where this was used. Initial search didn't find anything... thanks pycharm.

mshriver:
trigger: test-robottelo

pytest_fixtures/user_fixtures.py
```diff
@@ -20,3 +25,33 @@ def default_viewer_role(module_org, default_location):
...
+def module_user(request, module_org, module_loc):
+    """Creates admin user with default org set to module org and shares that
+    user for all tests in the same test module. User's login contains test
+    module name as a prefix.
+
+    :rtype: :class:`nailgun.entities.Organization`
+    """
+    # take only "module" from "tests.ui.test_module"
+    test_module_name = request.module.__name__.split('.')[-1].split('_', 1)[-1]
+    login = f'{test_module_name}_{gen_string("alphanumeric")}'
```
JacobCallahan, line 40: what do you think the risk is if API, CLI, and UI tests potentially run on the same satellite?

pytest_fixtures/user_fixtures.py
```diff
@@ -20,3 +25,33 @@ def default_viewer_role(module_org, default_location):
...
+    password = gen_string('alphanumeric')
+    LOGGER.debug('Creating session user %r', login)
+    user = entities.User(
+        admin=True,
+        default_organization=module_org,
+        default_location=module_loc,
+        description=f'created automatically by airgun for module "{test_module_name}"',
```
JacobCallahan, line 47: This is created by nailgun. Or do you mean created for airgun?

tests/foreman/conftest.py
JacobCallahan: what is the risk if it isn't?

JacobCallahan (changes requested):

pytest_fixtures/user_fixtures.py
mshriver: this is straight copied from the original function, so the original function was wrong in stating this was done via airgun

pytest_fixtures/user_fixtures.py
mshriver: I don't see any, the name includes a randomized component.  The user is also deleted once the module is done.

There is a `module_user` fixture also defined in `pytest_fixtures/api_fixtures.py`, but it isn't customized at all and simply creates a user under org/location scoped for the module as well.

Some test modules are also defining their own `module_user` fixtures.

tests/foreman/conftest.py
mshriver: For CI, none as the instances are immediately deleted.

For individuals running the tests ... not much, they just can't lock themselves out of their instance with failed logins anymore.  If they're trying to trigger that condition it won't.

pytest_fixtures/user_fixtures.py
JacobCallahan: ahhh ok. yeah unless i missed an import due to git context, this is nailgun

mshriver: Fixed

mshriver:
trigger: test-robottelo

mshriver:
I can't reproduce the unit test failure in test_report.py locally, trying to figure out what's going on in the GHA.

mshriver:
trigger: test-robottelo

mshriver:
trigger: test-robottelo

JacobCallahan (approved):

PR #8370, opened 2021-02-19
Remove xdist from collect-only in GHA

mshriver:
Due to randomization of parameter values in robottelo, xdist collection
comparison sometimes fails.

This is causing GHA workflow failures on PRs that should otherwise pass,
causing maintainers to have to re-run workflows frequently


JacobCallahan (approved):
ACK

PR #8375, opened 2021-02-22
Move junit property fixtures and reporting hook

mshriver:
This was in conftest, which was getting copied for the unit test.

Since the file with the hooks/fixtures needs to be copied during the
unit test execution, split the reporting fixtures into their own file
for more specific unit testing.

This allows for content in foreman/conftest.py to not have to be unit
test safe, when the module is not intended for use in unit tests


mshriver:
trigger: test-robottelo

pytest_fixtures/reporting_fixtures.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def pytest_report_header(config):
+    """Called when pytest session starts"""
+    messages = []
+
+    shared_function_enabled = 'OFF'
+    scope = ''
+    storage = 'file'
+    if setting_is_set('shared_function'):
```
JacobCallahan, line 21: is there a case in which this wouldn't be set? I believe the facade layer runs through all settings before this point.

pytest_fixtures/reporting_fixtures.py
```diff
@@ -0,0 +1,58 @@
...
+
+import pytest
+from _pytest.junitxml import xml_key
+from xdist import get_xdist_worker_id
+
+from robottelo.config import settings
+from robottelo.decorators import setting_is_set
+
+
+FMT_XUNIT_TIME = "%Y-%m-%dT%H:%M:%S"
```
JacobCallahan, line 11: single quotes :)

pytest_fixtures/reporting_fixtures.py
```diff
@@ -0,0 +1,58 @@
...
+    return messages
+
+
+def pytest_sessionstart(session):
+    """Workaround for https://github.com/pytest-dev/pytest/issues/7767
+
+    remove if resolved and set autouse=True for record_testsuite_timestamp_xml fixture
+    """
+    if get_xdist_worker_id(session) == 'master':
+        if session.config.pluginmanager.hasplugin("junitxml"):
```
JacobCallahan, line 41: single quotes :)
other instances in this file as well

JacobCallahan (changes requested):
small things

pytest_fixtures/reporting_fixtures.py
mshriver: This is unchanged from the existing functionality, I just moved the function.  I can update it, but would prefer to keep the scope related to how the unit test operates and whether the hook runs on xdist workers.

pytest_fixtures/reporting_fixtures.py
mshriver: replaced quotes in this module and the unit test module.

JacobCallahan (approved):
ACK

JacobCallahan:
@rplevka can you check this out? Thanks!

rplevka (approved):

PR #8380, opened 2021-02-24
Remove unused ostree settings

tpapaioa:
This PR is part of the conversion dynaconf. Because the `ostree` section of `robottelo.properties` is no longer used in any tests, this PR simply removes it instead of converting it.


mshriver (approved):
Agreed, while there are framework components and tests for ostree, this setting is never used.

If its needed, we can define a new validator and dynaconf file.

JacobCallahan (approved):
Ha! Even better!!!

PR #8387, opened 2021-02-25
Update test_organization to use shorter params

mshriver:
generate random strings within the test, instead of a 256 char parameter value

This is actually a limitation of pytest-reportportal's live reporting feature, which throws exceptions with parameter values greater than 128 chars.


mshriver:
trigger: test-robottelo 
pytest: tests/foreman/api/test_organization.py  -k negative_update

swadeley (approved):
ACK

pondrejk (approved):

PR #8389, opened 2021-02-25
Update repos.yaml.template, add dynaconf variable formatting

mshriver:
There is a need to compose settings values from various components at runtime, take advantage of dynaconf dynamic variables through simple string formatting.

I think there is a bigger opportunity here to actually reformat how repos are stored, which will also require updating all references to the relevant settings in the test framework.  There is a lot of repetition related to active RHEL versions that could be better handled with a different structure, even leveraging jinja2 instead of basic formatting in order to iteratively create settings maps.

Throw a few lines into gitignore for some commonly found local files.


mshriver:
trigger: test-robottelo

mshriver:
trigger: test-robottelo

pondrejk (approved):

omkarkhatavkar (approved):
ACK

robottelo/config/validators.py
```diff
@@ -197,6 +197,30 @@
         ),
         Validator("report_portal.fail_threshold", default=20),
     ],
+    repos=[
+        Validator(
+            'repos.rhel6_repo',
+            'repos.rhel7_repo',
+            'repos.rhel8_repo',
```
jyejare, line 204: Add `repos.rhel_repo_host` the most important one :)

or are you leaving it upto the user to fill out the URLs directly instead of depending on repos.rhel_repo_host ?

robottelo/config/validators.py
```diff
@@ -197,6 +197,30 @@
...
+            'repos.rhel8_os.baseos',
+            'repos.rhel8_os.appstream',
+            'repos.sattools_repo.rhel6',
+            'repos.sattools_repo.rhel7',
+            'repos.sattools_repo.rhel8',
+            'repos.satmaintenance_repo',
+            'repos.rhscl_repo',
+            'repos.ansible_repo',
+            'repos.swid_tools_repo',
+            must_exist=False,
```
jyejare, line 220: Why its False ?

jyejare: Sattools repos are one on the important repos of all time.

jyejare (commented):
ACK pending questions.

otherwise, LGTM.

conf/robottelo.yaml.template
```diff
@@ -60,3 +60,7 @@ ROBOTTELO:
   CDN: true
   # Run one datapoint or multiple datapoints for tests
   RUN_ONE_DATAPOINT: false
+  # Satellite version supported by this branch
+  # UNDR version is used for some URL composition
+  SATELLITE_VERSION: "6.9"
```
jyejare, line 65: Why we are hard-coding it to 6.9, we have utils to fetch the satellite version from satellite and being used everywhere in tests. What if user provides a wrong satellite version here in robottelo ? Do we have any plans to use this setting in robottelo ? 

Thoughts ?

mshriver: Because the master branch of robottelo only supports a single satellite y-stream version, at the moment that's 6.9.  This would get updated to 6.10 when the 6.9.z branch is created.

That being said, I recognize your comments about setting it from env, and we can do that in CI. 

The utils to fetch satellite version from test aren't of any use here, this is used to format dynaconf vars when they're read.  We also need to be able to collect tests without having a satellite reachable.

This is part of a larger plan to deduce reliance on `get_satellite_version` in the pytest session when you haven't necessarily checked out a satellite or don't have one available yet.

@JacobCallahan PR #8303 is establishing an xdist pattern for instance checkout on demand. It will be using this setting (he's actually looking at creating a new VERSION section for this), in order to default broker args for the checkout.

robottelo/config/validators.py
mshriver: The validator wasn't validating any of these before, and I do not plan on making these fields required with the current validation strategy.

When we move to validation being dynamically executed when relevant tests are collected, then we can make these fields `must_exist`.

Right now, we don't want to make sattools_repo `must_exist` when you might not be running tests that use the field at all.

robottelo/config/validators.py
mshriver: Think about it this way - there is formatting in the example, but the only fields that really need validated are the actual constructed repo URLs.  Tests/framework doesn't actually use `repos.rhel_repo_host`, and if its missing you'll get an immediate KeyError when trying to load settings.

JacobCallahan (approved):
ACK

PR #8396, opened 2021-03-02
Add assignee testimony tokens

tpapaioa:
This PR adds a new testimony token, `Assignee`, to all tests with ` CaseComponent`. The ability to filter on test assignee has also been added as a pytest option, alongside the existing importance and component filters.

```
  --assignee=ASSIGNEE   Comma separated list of assignees to include in test
                        collection
```

### DIFF
--- a/tests/upgrades/test_capsule.py
+++ b/tests/upgrades/test_capsule.py
@@ -6,7 +6,9 @@
 
 :CaseLevel: Acceptance
 
-:CaseComponent: CLI
+:CaseComponent: Capsule
+
+:Assignee: vsedmik
 
 :TestType: Functional
 
--- a/pytest_plugins/testimony_markers.py
+++ b/pytest_plugins/testimony_markers.py
@@ -17,7 +17,11 @@ def pytest_addoption(parser):
     )
     parser.addoption(
         '--component',
-        help="Comma separated list of component names to include in test collection",
+        help='Comma separated list of component names to include in test collection',
+    )
+    parser.addoption(
+        '--assignee',
+        help='Comma separated list of assignees to include in test collection',
     )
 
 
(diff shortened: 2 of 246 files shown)


pondrejk:
Shouldn't this be set per file rather than per test? 
I wonder if we should include the assignee information in tests at all, these assignments tend to change quite often, will we have cycles to keep these up to date? I guess some mapping from case components to owners stored at one place could reduce future pain

mshriver:
@pondrejk storing this data separately from robottelo and trying to inject it has caused plenty of headaches for us, and requires a bunch of custom tooling.  An internal confluence page exists that outlines what the current process was for adding assignees - when Mojo was deprecated that job started breaking, and it has been injecting the wrong assignees since (Q4 2020).

Assignees don't change that often, and its an easy find/replace to update names.

Having this metadata in robottelo allows for test collection based on the metadata, and without any outside tooling provides additional metadata to any test result tool that we use (Report Portal, Polarion). We just have to turn tokens into markers, just like we do with component and importance. 

Yes, assignees should be set at the highest scope (class, module) that we can so that we're not repeating testimony tokens, same as any of the current tokens.

tpapaioa:
It is set per file, when only one `CaseComponent` is present at the module level. Some files have a mix of tests with different components set at the class or test level though.

pondrejk:
Well, this would work for a while, but I'm sure we'd get to a messy state pretty soon. If an associate drops a component to a colleague, someone will have to find all the component's tests and update names. If this doesn't happen, we don't have tools to find out that metadata are out of date, right? Could we instead have a single mapping file in robottelo, and then set the assignee based on component when creating markers?

Anyway, I'd say this PR would warrant a wider discussion as it potentially adds responsibilities for component owners for years to come

mshriver:
Creating markers alone from a mapping file doesn't make the data available to betelgeuse, which is only reading testimony tokens. It also doesn't allow for assigning a subset of tests from a component to a different assignee.

A wider discussion was opened up to the team and feedback solicited, and this solution described, with no feedback.  Not to say that it can't be heard now, but it isn't coming out of nowhere.

Part of this effort also includes considering tooling to ease making the assignee updates, but we maintained test case assignees this way for years in the ManageIQ/integration_tests project and it never became a maintenance burden.   

You can run pytest collection to see all of the test cases assigned to someone, as well as observe the metadata in RP/Polarion.  If its wrong in Polarion, there's a clear path to where and why it was set.  Test collection of `--assignee someone --component SomeComponent` and `--component SomeComponent` should be equal for all components owned by a given assignee.

There's a clear mapping in robottelo for component tests - both in the module organization, and in the CaseComponent field.

tests/upgrades/test_capsule.py
```diff
@@ -6,7 +6,9 @@
 
 :CaseLevel: Acceptance
 
-:CaseComponent: CLI
+:CaseComponent: Capsule
```
JacobCallahan, line 9: good catch

JacobCallahan (approved):
ACK

mshriver (approved):

PR #8401, opened 2021-03-05
convert api/role to pytest

peterdragun:
All test passed, except this one. Not sure what is causing this problem, but it is also failing in current automation. Test behavior wasn't changed by this PR. Needs some more investigation.
```
===================================== short test summary info =====================================
FAILED tests/foreman/api/test_role.py::TestCannedRole::test_negative_create_taxonomies_by_org_admin
======================= 1 failed, 42 passed, 7 skipped in 659.58s (0:10:59) =======================
```

### DIFF
--- a/tests/foreman/api/test_role.py
+++ b/tests/foreman/api/test_role.py
@@ -26,66 +26,43 @@ from requests.exceptions import HTTPError
 from robottelo.config import settings
 from robottelo.datafactory import gen_string
 from robottelo.datafactory import generate_strings_list
-from robottelo.test import APITestCase
+from robottelo.datafactory import parametrized
 from robottelo.utils.issue_handlers import is_open
 
 
-class RoleTestCase(APITestCase):
+class TestRole:
     """Tests for ``api/v2/roles``."""
 
     @pytest.mark.tier1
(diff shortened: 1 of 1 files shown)


omkarkhatavkar:

tests/foreman/api/test_role.py
```diff
@@ -186,24 +153,28 @@ def user_config(self, user):
...
+            'loc': entities.Location().create(),
+        }
+
+    @pytest.fixture
+    def filter_taxonomies(self):
+        """Create filter taxonomies"""
+        return {
+            'org': entities.Organization().create(),
+            'loc': entities.Location().create(),
+        }
```
omkarkhatavkar, line 174: Why we creating similar object fixtures here ? can't we keep same fixture for both  `role_taxonomies` and `filter_taxonomies` with some common name as 'taxonomies' uses based on the scope of fixture ?

omkarkhatavkar: seeing below code I can understand there are times you require both fixture simultaneously. ignore it

omkarkhatavkar (approved):
ACK

pondrejk (approved):

PR #8409, opened 2021-03-09
test subnet u2p conv

tstrych:
`=========== 40 passed, 7 skipped, 115 warnings in 106.04s (0:01:46) ============`

### DIFF
--- a/tests/foreman/api/test_subnet.py
+++ b/tests/foreman/api/test_subnet.py
@@ -21,6 +21,8 @@ http://theforeman.org/api/apidoc/v2/1.15.html
 
 :Upstream: No
 """
+import re
+
 import pytest
 from nailgun import entities
 from requests.exceptions import HTTPError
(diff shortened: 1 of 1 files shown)


tests/foreman/api/test_subnet.py
```diff
@@ -21,528 +21,524 @@
...
+    :expectedresults: The Subnet is created with parameter
+    """
+    parameter = [{'name': gen_string('alpha'), 'value': gen_string('alpha')}]
+    subnet = entities.Subnet(subnet_parameters_attributes=parameter).create()
+    assert subnet.subnet_parameters_attributes[0]['name'] == parameter[0]['name']
+    assert subnet.subnet_parameters_attributes[0]['value'] == parameter[0]['value']
+
+
+@pytest.mark.tier1
+@pytest.mark.parametrize('name', **parametrized(generate_strings_list()))
```
JacobCallahan, line 54: Male sure you add docstring metadata stating that this test is parameterized.

JacobCallahan (changes requested):

tests/foreman/api/test_subnet.py
tstrych: Male sure :D that is sexism :smile: :smiley:

tests/foreman/api/test_subnet.py
```diff
@@ -21,528 +21,532 @@
...
+    subnet_param = entities.Parameter(name=name, subnet=subnet.id, value=values).create()
+    assert subnet_param.name == name
+    assert subnet_param.value == values
+
+
+@pytest.mark.tier1
+@pytest.mark.parametrize(
+    'name',
+    **parametrized(
+        [f'{separator}'.join(generate_strings_list()) for separator in [',', '/', '-', '|']]
```
mshriver, line 107: As a pattern, its better to parametrize the simpler value here - the separator.

Then in the test, generate the random string using the separator parameter. This way the parameter value isn't randomized itself.

mshriver (approved):

mshriver (changes requested):
Turning my optional comment to a required one at @JacobCallahan request

tests/foreman/api/test_subnet.py
```diff
@@ -21,528 +21,528 @@
...
+    subnet = entities.Subnet().create()
+    name = gen_string('alpha')
+    values = ', '.join(generate_strings_list())
+    subnet_param = entities.Parameter(name=name, subnet=subnet.id, value=values).create()
+    assert subnet_param.name == name
+    assert subnet_param.value == values
+
+
+@pytest.mark.tier1
+@pytest.mark.parametrize('separator', **parametrized([',', '/', '-', '|']))
```
tstrych, line 104: here we go :) @JacobCallahan

JacobCallahan (approved):

tests/foreman/api/test_subnet.py
mshriver: This works, thanks!

When you pass a dictionary to `parametrized` method, it will use the dict keys as parameter ids - not necessary here, but optional.

@pytest.mark.parametrize('separator', **parametrized({'comma': ',', 'slash': '/', 'dash': '-', 'pipe': '|'}))

tests/foreman/api/test_subnet.py
```diff
@@ -21,528 +21,530 @@
...
+    name = gen_string('alpha')
+    values = ', '.join(generate_strings_list())
+    subnet_param = entities.Parameter(name=name, subnet=subnet.id, value=values).create()
+    assert subnet_param.name == name
+    assert subnet_param.value == values
+
+
+@pytest.mark.tier1
+@pytest.mark.parametrize(
+    'separator', **parametrized({'comma': ',', 'slash': '/', 'dash': '-', 'pipe': '|'})
```
tstrych, line 105: done :)

tests/foreman/api/test_subnet.py
tstrych: I mean does it really matter ? Sure it can save you some time, once you will see the test failed with parameter dash. 
But I bet this test will not fail. 

I mark the change below :)

tests/foreman/api/test_subnet.py
tstrych: @mshriver still have status with change requested

tstrych: mike gave me ack via gchat

mshriver: Some miscommunication here, I didn't approve in chat, I need to do a quick review before approving.

mshriver (approved):

PR #8412, opened 2021-03-11
old VM style for global registartion test

pondrejk:
I need this to run now (current ci, current snaps)

PR #8419, opened 2021-03-12
more fixes to globreg test with old VM

pondrejk:
adding changes I didn't realize were needed in  #8412

### DIFF
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -1377,7 +1377,7 @@ def test_positive_global_registration_end_to_end(
     GlobalParameter().set({'name': 'host_registration_insights', 'value': 1})
     GlobalParameter().set({'name': 'host_registration_remote_execution', 'value': 1})
     # rex interface
-    iface = 'eth0'
+    iface = 'ens3'
     # fill in the global registration form
     with session:
         cmd = session.host.get_register_command(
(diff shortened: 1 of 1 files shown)


JacobCallahan:
This should go direct into the new 6.9-legacy branch.

tests/foreman/ui/test_host.py
```diff
@@ -1377,7 +1377,7 @@ def test_positive_global_registration_end_to_end(
     GlobalParameter().set({'name': 'host_registration_insights', 'value': 1})
     GlobalParameter().set({'name': 'host_registration_remote_execution', 'value': 1})
     # rex interface
-    iface = 'eth0'
+    iface = 'ens3'
```
JacobCallahan, line 1380: Is this interface applicable to the test.only when using the virtual machine class?

JacobCallahan (changes requested):

tests/foreman/ui/test_host.py
pondrejk: That's what I figured from the automation logs. Successful PRT run in 8356 tells me eth0 is ok for the master branch.

pondrejk:
@JacobCallahan yes, this PR is open for the 6.9-legacy branch ;)

JacobCallahan:
> @JacobCallahan yes, this PR is open for the 6.9-legacy branch ;)

Ha! You're right!

tests/foreman/ui/test_host.py
JacobCallahan: Perfect, thanks for looking into that

JacobCallahan (approved):

omkarkhatavkar (approved):
ACK

PR #8440, opened 2021-03-19
Give upgrade tests to their rightful owners, API is incorrect component

lhellebr:
Everyone involved should review


JacobCallahan (approved):

mshriver (approved):
Thanks for correctly aligning these cases @lhellebr, I just cross referenced against Bugzilla.

PR #8450, opened 2021-03-22
Add new broker settings file for robottelo

JacobCallahan:
Currently just the single value as all other config values reside
in broker's own settings file.


swadeley (approved):
ACK

pondrejk (approved):

PR #8468, opened 2021-03-25
api bulk destroy test added

tstrych:
Depends on [nailgun PR](https://github.com/SatelliteQE/nailgun/pull/783)

`================ 1 passed, 61 deselected, 47 warnings in 52.49s ================`

While I was writing this test I found BZ. So I think it was worth the time.

### DIFF
--- a/tests/foreman/api/test_host.py
+++ b/tests/foreman/api/test_host.py
@@ -1438,3 +1438,47 @@ class TestHostInterface:
             host.read()
         except HTTPError:
             pytest.fail("HTTPError 404 raised unexpectedly!")
+
+
+class TestHostBulkAction:
+    """ Tests for host bulk actions. """
+
+    @pytest.mark.tier2
+    def test_positive_bulk_destroy(self, module_org):
+        """Destroy multiple hosts make sure that hosts were removed,
+        or were not removed when host is excluded from the list.
+
+        :id: 06d63376-8bf6-11eb-ab9f-98fa9b6ecd5a
+
+        :expectedresults: Included list of hosts,
+            that are not part of excluded list of host, are removed.
+
+        :CaseImportance: Medium
+        """
+
+        host_ids = []
+        for _ in range(3):
+            name = gen_choice(valid_hosts_list())
+            host = entities.Host(name=name, organization=module_org).create()
+            host_ids.append(host.id)
+
+        entities.Host().bulk_destroy(
+            data={
+                'organization_id': module_org.id,
+                'included': {'ids': host_ids},
+                'excluded': {'ids': host_ids[:-1]},
+            }
+        )
+        for host_id in host_ids[:-1]:
+            result = entities.Host(id=host_id).read()
+            assert result.id == host_id
+
+        with pytest.raises(HTTPError):
+            entities.Host(id=host_ids[-1]).read()
+
+        entities.Host().bulk_destroy(
+            data={'organization_id': module_org.id, 'included': {'ids': host_ids[:-1]}}
+        )
+        for host_id in host_ids[:-1]:
+            with pytest.raises(HTTPError):
+                entities.Host(id=host_id).read()


tstrych:
trigger: test-robottelo
pytest: -k test_positive_bulk_destroy tests/foreman/api/test_host.py
nailgun: 783

JacobCallahan:
Can you investigate the prt failure?

omkarkhatavkar:
trigger: test-robottelo
pytest: -k test_positive_bulk_destroy tests/foreman/api/test_host.py
nailgun: 783

tests/foreman/api/test_host.py
```diff
@@ -1438,3 +1438,45 @@ def test_positive_delete_and_check_host(self):
...
+
+        entities.Host().bulk_destroy(
+            data={
+                'organization_id': module_org.id,
+                'included': {'ids': host_ids},
+                'excluded': {'ids': host_ids[:-1]},
+            }
+        )
+        for host_id in host_ids[:-1]:
+            entities.Host(id=host_id).read()
```
pondrejk, line 1473: maybe assert for result return code, or for something in the result content

tstrych:
PRT failure was investigated, it took code coverage comment instead the proper one. We can discuss the problems on scheduled meeting. You can see new results passed.

tests/foreman/api/test_host.py
tstrych: This action will pass only if the hosts were excluded from the list.  I did not care about the read result. 

Please write me more, what would you expect to be there and why ?

tests/foreman/api/test_host.py
```diff
@@ -1438,3 +1438,45 @@ def test_positive_delete_and_check_host(self):
...
+        for _ in range(3):
+            name = gen_choice(valid_hosts_list())
+            host = entities.Host(name=name, organization=module_org).create()
+            host_ids.append(host.id)
+
+        entities.Host().bulk_destroy(
+            data={
+                'organization_id': module_org.id,
+                'included': {'ids': host_ids},
+                'excluded': {'ids': host_ids[:-1]},
```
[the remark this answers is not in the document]
tstrych: included and excluded are required parameters. 
You need to specify which hosts you want to run it on. 
And you can pass empty list of excluded hosts if you want to run it only on included.

So if I understand correctly to this behaviour: mathematically,  union of Included - union of excluded =  result union
Both included and and excluded are required so this is not about priority :smile:

tests/foreman/api/test_host.py
JacobCallahan: without some kind of check for the read contents or read success, what is the purpose of this loop?

tstrych: the read checks whether the excluded hosts were really excluded, 
read would fail if not

JacobCallahan: Without the read raising an exception, the test wouldn't know that the read failed.

tests/foreman/api/test_host.py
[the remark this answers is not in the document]
pondrejk: Hm I was wondering why would someone want to spell out included hosts and then marked them as excluded... but I see there is also included[search] so that makes sense when you have a broad search but want to be sure not to remove a specific host, maybe a topic for extension PR

tests/foreman/api/test_host.py
pondrejk: @tstrych we want to make these tests future proof, so some additional asserts could help us to catch issues that could go unnoticed otherwise (like some different code returned, or read passing successfully but returning wrong host), something like this (not guaranteeing these are the exact api calls):
```
result = entities.Host(id=host_id).read()
assert result.status == 0
assert result.stdout.id == host_id
```

JacobCallahan: you could remove the status check in the above example and rely solely on checking the host id

tstrych: thank you for  your explanation :+1:

tests/foreman/api/test_host.py
```diff
@@ -1438,3 +1438,47 @@ def test_positive_delete_and_check_host(self):
...
+            }
+        )
+        for host_id in host_ids[:-1]:
+            result = entities.Host(id=host_id).read()
+            assert result.id == host_id
+
```
tstrych, line 1475: there is no status or stdout field
but host id is checked as proposed

pondrejk (approved):

PR #8479, opened 2021-04-05
auth status is fails to give error on expiration bugzilla 1903693

omkarkhatavkar:
Based on the Bugzilla 1903693 it is failing on the first occurrence of auth status, it was passing earlier in the test because after any API call it starts passing as an expected message. I have to change the assertion location to validate this Bugzilla. We will hold this PR till Bugzilla get valid confirmation

### DIFF
--- a/tests/foreman/cli/test_auth.py
+++ b/tests/foreman/cli/test_auth.py
@@ -31,6 +31,7 @@ from robottelo.cli.settings import Settings
 from robottelo.cli.user import User
 from robottelo.config import settings
 from robottelo.constants import HAMMER_CONFIG
+from robottelo.utils.issue_handlers import is_open
 
 LOGEDIN_MSG = "Session exists, currently logged in as '{0}'"
 LOGEDOFF_MSG = "Using sessions, you are currently not logged in"
@@ -272,7 +273,7 @@ def test_positive_session_preceeds_saved_credentials(
 
     :id: e4277298-1c24-494b-84a6-22f45f96e144
 
-    :BZ: 1471099
+    :BZ: 1471099, 1903693
 
     :Steps:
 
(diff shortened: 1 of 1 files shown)


JacobCallahan (approved):
Changes are simple enough. What's your reason for the do not merge label?

omkarkhatavkar:
@JacobCallahan Just waiting for the response from Dev in the Bugzilla, Personally I think it is a valid bugzilla.

mshriver (approved):

tests/foreman/cli/test_auth.py
```diff
@@ -296,10 +296,10 @@ def test_positive_session_preceeds_saved_credentials(
         assert LOGEDIN_MSG.format(admin_user['login']) in result[0]['message']
         # list organizations without supplying credentials
         sleep(70)
-        with pytest.raises(CLIReturnCodeError):
-            Org.with_user().list()
         result = Auth.with_user().status()
         assert LOGEDOFF_MSG.format(admin_user['login']) in result[0]['message']
+        with pytest.raises(CLIReturnCodeError):
```
mshriver, line 301: The BZ has been triaged, but doesn't look like it has been targeted for any release yet.

Since adding the BZ to the metadata will cause the testcase to skip execution if the bug is open, I think it would be a better approach here to execute this `with` block based on the BZ status.  There are examples in the framework of doing this - you just build an `if` statement that checks the BZ blocking status.

Then, until this BZ is fixed, the test can still run, and just won't check that a CLIReturnCodeError is raised when you try to list orgs without a valid token.

mshriver:
Moving this to draft state to clearly show it is not waiting review or action by maintainers, but an action by the author.

mshriver (approved):

PR #8481, opened 2021-04-06
migrate discovery settings to dynaconf

rplevka:


conf/discovery.yaml.template
```diff
@@ -0,0 +1,3 @@
+DISCOVERY:
```
ColeHiggins2, line 1: Looks good. I would however add a description of when the discovery iso is required and for what kinds of tests, For documentation purposes.

mshriver: I'm not going to block merge on this, as I think there are other places for this to be documented or found.

mshriver (approved):

PR #8483, opened 2021-04-06
cli ostreeebranch to pytest

tstrych:
u2p conversion, change in robottelo constants for working ostree

Tested with commented `@pytest.mark.skip_if_open("BZ:1625783")`
=================== 6 passed, 1 warning in 108.41s (0:01:48) ===================

Don't know why, but when I test it without commented @pytest.mark.skip_if_open("BZ:1625783"). 
It should not work, but those tests do run. Any ideas ?

### DIFF
--- a/tests/foreman/cli/test_ostreebranch.py
+++ b/tests/foreman/cli/test_ostreebranch.py
@@ -19,138 +19,149 @@
 import random
 
 import pytest
+from nailgun import entities
 
 from robottelo.cli.contentview import ContentView
 from robottelo.cli.factory import make_content_view
 from robottelo.cli.factory import make_org_with_credentials
 from robottelo.cli.factory import make_product_with_credentials
 from robottelo.cli.factory import make_repository_with_credentials
-from robottelo.cli.factory import make_user
 from robottelo.cli.ostreebranch import OstreeBranch
 from robottelo.cli.repository import Repository
 from robottelo.config import settings
-from robottelo.constants.repos import FEDORA27_OSTREE_REPO
+from robottelo.constants.repos import OSTREE_REPO
 from robottelo.decorators.host import skip_if_os
-from robottelo.test import CLITestCase
+
+pytestmark = [
+    pytest.mark.skipif((not settings.repos_hosting_url), reason='Missing repos_hosting_url'),
+    pytest.mark.tier3,
+]
 
 
-@pytest.mark.skip_if_open("BZ:1625783")
+@pytest.fixture(scope='module')
+def ostree_user_credentials():
+    password = 'password'
+    user = entities.User(admin=True, password=password).create()
+    return user.login, password
+
+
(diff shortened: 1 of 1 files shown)


tstrych:
trigger: test-robottelo
pytest: tests/foreman/cli/test_ostreebranch.py

tstrych:
trigger: test-robottelo
pytest: tests/foreman/cli/test_ostreebranch.py

latran (commented):
Regarding @pytest.mark.skip_if_open("BZ:1625783") from your comment.  I think it's at a fixture level and that's why it maybe doesn't get skipped.  I don't think I've seen it at a fixture level before so I'm not 100% sure.  

If you add a class with the skip, then I think it will skip.  Let's see what other reviewers say, but overall the pytest conversion itself looks good overall.  I'll approve it once your skip issue is addressed.

mshriver:
@latran you're correct about the behavior of `skip_if_open` (and other) pytest marks.  Good observation!

tstrych:
I know there was strong influence to remove classes if they are not necessary. I will wait if somebody else will add any other comment and I will probably fix it tomorrow.

latran:
@tstrych there's always that debate, but since there's not multiple classes in this file, I would do what Mike suggested.  Please make changes and I'll ack.

latran (approved):
ACK

tests/foreman/cli/test_ostreebranch.py
```diff
@@ -19,138 +19,152 @@
...
+
+@pytest.fixture(scope='module')
+def ostree_user_credentials():
+    password = 'password'
+    user = entities.User(admin=True, password=password).create()
+    return user.login, password
+
+
+@pytest.fixture(scope='module')
+@pytest.mark.skipif((not settings.repos_hosting_url), reason='Missing repos_hosting_url')
```
mshriver, line 44: This was discussed, and I believe was already suggested to be updated, but an approval has been made with this active.

AFAIK, you cannot mark fixtures like this and have it apply to the tests.  Can you demonstrate that the tests will actually skip when this mark is applied to the fixture instead of to the test functions, module, or class?

mshriver: If you want to apply this without adding it to every function in the module, define this a the module level:

`pytestmark = [pytest.mark.skipif(yada yada yada)]`

mshriver (changes requested):
Marks need to be applied to test functions, classes, or the module.

tests/foreman/cli/test_ostreebranch.py
tstrych: yea definitely, I just didn't think... :expressionless:

mshriver (approved):

mshriver:
trigger: test-robottelo
pytest: tests/foreman/cli/test_ostreebranch.py

PR #8441, opened 2021-03-19
Refactor cli/test_reporttemplates, pytest fixtures

mshriver:
The assertions had largely already been rewritten.

Remove the class and dependence on CLITestCase.
Re-structure the unittest class setup function into several fixtures.

Rewrite one assertion pattern for CLI output checking

## Local testing

I have 3 failures locally because I don't have a provisioning host configured. Using robottelo-runnner to run it.

```
setup@localhost:~/repos/robottelo (pytest-reporttemplates-13991$%=) % pytest -v tests/foreman/cli/test_reporttemplates.py 
========================================================================================== test session starts ==========================================================================================
platform linux -- Python 3.8.8, pytest-6.2.2, py-1.10.0, pluggy-0.13.1 -- /home/setup/repos/robottelo/.robottelo/bin/python3.8
cachedir: .pytest_cache
shared_function enabled - OFF - scope:  - storage: file
rootdir: /home/setup/repos/robottelo, configfile: pyproject.toml
plugins: forked-1.3.0, ibutsu-1.14.1, xdist-2.2.1, mock-3.5.1, reportportal-5.0.8, cov-2.11.1, services-2.2.1
collected 21 items / 3 deselected / 18 selected                                                                                                                                                         

tests/foreman/cli/test_reporttemplates.py::test_positive_report_help PASSED                                                                                                                       [  5%]
tests/foreman/cli/test_reporttemplates.py::test_positive_end_to_end_crud_and_list PASSED                                                                                                          [ 11%]
tests/foreman/cli/test_reporttemplates.py::test_positive_generate_report_nofilter_and_with_filter PASSED                                                                                          [ 16%]
tests/foreman/cli/test_reporttemplates.py::test_positive_lock_and_unlock_report PASSED                                                                                                            [ 22%]
tests/foreman/cli/test_reporttemplates.py::test_positive_report_add_userinput PASSED                                                                                                              [ 27%]
tests/foreman/cli/test_reporttemplates.py::test_positive_dump_report PASSED                                                                                                                       [ 33%]
tests/foreman/cli/test_reporttemplates.py::test_positive_clone_locked_report PASSED                                                                                                               [ 38%]
tests/foreman/cli/test_reporttemplates.py::test_positive_generate_report_sanitized PASSED                                                                                                         [ 44%]
tests/foreman/cli/test_reporttemplates.py::test_negative_create_report_without_name PASSED                                                                                                        [ 50%]
[...]

PR #8486, opened 2021-04-07
Restructure tests/foreman/longrun/test_scap.py

jameerpathan111:


tests/foreman/longrun/test_oscap.py
```diff
@@ -42,39 +42,15 @@
...
-from robottelo.decorators import skip_if_not_set
 from robottelo.helpers import add_remote_execution_ssh_key
 from robottelo.helpers import file_downloader
 from robottelo.helpers import ProxyError
-from robottelo.test import CLITestCase
 from robottelo.vm import VirtualMachine
 
 
-class OpenScapTestCase(CLITestCase):
+class TestOpenScap:
```
JacobCallahan, line 51: if there is just one class in this file, is the class truly necessary?

tests/foreman/longrun/test_oscap.py
```diff
@@ -94,86 +70,118 @@ def fetch_scap_and_profile_id(cls, scap_name, scap_profile):
         ]
         return scap_id, scap_profile_ids
 
-    @classmethod
-    def configure_puppet_test(cls):
+    @pytest.fixture(scope="class", autouse=True)
```
JacobCallahan, line 73: Please switch these to single-quote strings.

tests/foreman/longrun/test_oscap.py
```diff
@@ -94,86 +70,118 @@ def fetch_scap_and_profile_id(cls, scap_name, scap_profile):
...
-            'ak_name': {'rhel8': ak_name_8, 'rhel7': ak_name_7, 'rhel6': ak_name_6},
-            'env_name': env.name,
-        }
-
     @pytest.mark.tier4
     @pytest.mark.libvirt_content_host
     @pytest.mark.upgrade
-    def test_positive_upload_to_satellite(self):
+    @pytest.mark.parametrize(
+        "rhel_repo, content, profile1, profile2, profile3, distro",
```
JacobCallahan, line 147: Please switch these to single-quote strings.

tests/foreman/longrun/test_oscap.py
```diff
@@ -409,81 +342,99 @@ def test_positive_oscap_run_with_tailoring_file_and_capsule(self):
...
+            arf_report = Arfreport.list({'search': f'host={vm.hostname.lower()}', 'per-page': 1})
+            assert arf_report is not None
+            Arfreport.delete({'id': arf_report[0].get('id')})
 
     @pytest.mark.upgrade
     @pytest.mark.libvirt_content_host
     @pytest.mark.tier4
-    def test_positive_oscap_run_with_tailoring_file_with_ansible(self):
+    @pytest.mark.parametrize(
+        "rhel_repo, content, profile, distro",
```
JacobCallahan, line 401: Please switch these to single-quote strings.

JacobCallahan (changes requested):

tests/foreman/longrun/test_oscap.py
jameerpathan111: updated the pr.

tests/foreman/longrun/test_oscap.py
jameerpathan111: updated the pr.

jameerpathan111:
updated test result:
```
$ pytest -n 3 -Wignore --include-libvirt tests/foreman/longrun/test_oscap.py
========================================================================== test session starts ===========================================================================
platform linux -- Python 3.8.6, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
rootdir: /home/jpathan/projects/robottelo, configfile: pyproject.toml
plugins: ibutsu-1.13, services-2.2.1, mock-3.5.1, reportportal-5.0.7, xdist-2.2.1, forked-1.3.0
gw0 [6] / gw1 [6] / gw2 [6]
......                                                                                                                                                             [100%]
===================================================================== 6 passed in 2376.55s (0:39:36) =====================================================================
WARNING:robottelo.config:Dynaconf validation failed, continuing for the sake of unit tests
```

tests/foreman/longrun/test_oscap.py
```diff
@@ -42,670 +42,593 @@
...
+        Scapcontent.update({'title': content['title'], 'organization-ids': organization_ids})
+
+    return request
+
+
+@pytest.mark.tier4
+@pytest.mark.libvirt_content_host
+@pytest.mark.upgrade
+@pytest.mark.parametrize(
+    'rhel_repo, content, profile1, profile2, profile3, distro',
```
omkarkhatavkar, line 143: instead `profile1, profile2, profile3` making as `dsr_profile, pci_profile, osp_profile` would help in code readability ?

this is non-blocker

omkarkhatavkar (approved):
ACK, one minor comment

tests/foreman/longrun/test_oscap.py
jameerpathan111: yeah, it would make code more readable but the profiles can change and I think it's better to keep it general. For example, currently profile3 name for rhel6 is `usgcb` and not `ospp` as it's for rhel7 and rhel8. So I can't use `osp_profile` because of this exception.

tests/foreman/longrun/test_oscap.py
```diff
@@ -42,670 +42,593 @@
...
+
+
+@pytest.fixture(scope='module')
+def scap_setup(request, module_org):
+    """Sets up the whole provisioning environment needed for Puppet based
+    end-to-end tests like OSCAP etc
+
+    :returns: A dict of entities to help with provisioning
+    """
+    request.puppet_classes = ['foreman_scap_client::params', 'foreman_scap_client']
```
mshriver, line 77: I don't think this should be tied to the request object, its static. It can be a module level constant.

tests/foreman/longrun/test_oscap.py
```diff
@@ -42,670 +42,593 @@
...
+    rhel6_content = OSCAP_DEFAULT_CONTENT['rhel6_content']
+    rhel7_content = OSCAP_DEFAULT_CONTENT['rhel7_content']
+    rhel8_content = OSCAP_DEFAULT_CONTENT['rhel8_content']
+    sat6_hostname = settings.server.hostname
+    proxy = Proxy.list({'search': sat6_hostname})[0]
+    p_features = set(proxy.get('features').split(', '))
+    if {'Puppet', 'Ansible', 'Openscap'}.issubset(p_features):
+        request.proxy_id = proxy.get('id')
```
mshriver, line 85: I don't think this is a pattern that should be used. the `request` fixture in pytest is special, and is not a generic object that we should be attaching attributes to.

You will see some use of setting `request.cls` attributes, which is different. In this case there is no longer a class.

The pattern that you should follow here is defining separate fixtures for the entities that are created during setup, and are utilized by the test.

There should be a `proxy_module` fixture defined, which actually _creates_ a proxy to use, not just assumes that there is already one configured.  It should then return that proxy entity, or its ID if that is all that is needed.

There should also be fixtures defined for the use of `request.ak_name`, `request.org`, `request.puppet_env`, `request.env`, and all others below.

Please see this PR as an example of how to scope and structure the fixtures:
https://github.com/SatelliteQE/robottelo/pull/8441/files

The test cases then include what fixtures they need to directly access, and use them directly, instead of this generic `setup` fixture returning the request instance which has been mangled by the fixture itself.

mshriver (changes requested):
The pattern of adding attributes to the request fixture instance is not a good one to follow, there should be separate fixtures defined for the multitude of entities that are defined and modified within this test setup.

tests/foreman/longrun/test_oscap.py
mshriver: Another option for a pattern to follow here if you would really prefer to have one setup fixture is to define a `namedtuple` that can be used to return all of the relevant entities through one fixture return.

```
from collections import namedtuple
SCAPEntities = namedtuple('SCAPEntities', ['proxy', 'ak_name', 'org'])  # short example, include all fields

@pytest.fixture
def scap_setup(module_org):
    proxy=entities.list()[0]
    ak_name={'rhel': 'random-string'}
    return SCAPEntities(proxy, ak_name, module_org)
   
def test_something(scap_setup):
    logger.info(f'The Proxy used for scap testing is: {scap_setup.proxy}')
    # reference scap_setup.<field> for all of the things attached to the request instance as currently returned from scap_setup   
```

tests/foreman/longrun/test_oscap.py
```diff
@@ -42,670 +44,617 @@
...
+    """ Update default scap content"""
+    for content in rhel8_content, rhel7_content, rhel6_content:
+        content = Scapcontent.info({'title': content}, output_format='json')
+        organization_ids = [content_org['id'] for content_org in content.get('organizations', [])]
+        organization_ids.append(module_org.id)
+        Scapcontent.update({'title': content['title'], 'organization-ids': organization_ids})
+
+
+@pytest.fixture(scope='module', autouse=True)
+def scap_setup(module_org, default_proxy, content_view, lifecycle_env, puppet_env):
```
mshriver, line 160: The structure above looks a lot better! But I think you have `autouse=True` and don't need it, in your case the tests are using `scap_setup`, which in turn will call all of the listed fixtures without the need for autouse.

I would add this fixture to tests that need it but aren't currently listing it instead of using autouse here.

JacobCallahan: correct. in general, autouse should be for fixtures that aren't direct dependencies of other tests but need to run anyway.

jameerpathan111: added it there by mistake, it was not intended. :P

jameerpathan111: updated pr

tests/foreman/longrun/test_oscap.py
jameerpathan111: updated pr

tests/foreman/longrun/test_oscap.py
jameerpathan111: updated pr

tests/foreman/longrun/test_oscap.py
```diff
@@ -42,670 +42,621 @@
...
-            # uploads report to Internal Capsule.
-            vm.execute_foreman_scap_client()
-            # Assert whether oscap reports are uploaded to
-            # Satellite6.
-            result = Arfreport.list({'search': f'host={vm.hostname.lower()}'})
-            assert result is not None
+        job_id = Host.ansible_roles_play({'name': vm.hostname.lower()})[0].get('id')
+        wait_for_tasks(
+            f"resource_type = JobInvocation and resource_id = {job_id} and "
+            "action ~ \"hosts job\""
```
JacobCallahan, line 525:
```suggestion
            f'resource_type = JobInvocation and resource_id = {job_id} and '
            'action ~ "hosts job"'
```

JacobCallahan (changes requested):

tests/foreman/longrun/test_oscap.py
jameerpathan111: updated pr.

JacobCallahan (approved):

JacobCallahan:
@mshriver have your comments been addressed to your satisfaction?

mshriver (approved):

mshriver:
Fixtures look much better now - but @jameerpathan111 needs to re-run pre-commit and amend his last commit to fix `Black` formatting.

jameerpathan111:
updated test result:

```
$ pytest -Wignore --include-libvirt tests/foreman/longrun/test_oscap.py | tee scap.log
platform linux -- Python 3.8.6, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
rootdir: /home/jpathan/projects/robottelo, configfile: pyproject.toml
plugins: ibutsu-1.13, services-2.2.1, mock-3.5.1, reportportal-5.0.7, xdist-2.2.1, forked-1.3.0
collected 12 items / 6 deselected / 6 selected

tests/foreman/longrun/test_oscap.py ......                               [100%]

================= 6 passed, 6 deselected in 5029.47s (1:23:49) =================
```

PR #8488, opened 2021-04-07
jobtemplate pytest conversion

pondrejk:
```
pytest tests/foreman/cli/test_jobtemplate.py 
======================================================================= test session starts ========================================================================                                                                                                                                                 

tests/foreman/cli/test_jobtemplate.py .......                                                                                                                [100%]
============================================================ 7 passed, 2 warnings in 178.98s (0:02:58) ==============
```

### DIFF
--- a/tests/foreman/cli/test_jobtemplate.py
+++ b/tests/foreman/cli/test_jobtemplate.py
@@ -21,190 +21,156 @@ from fauxfactory import gen_string
 
 from robottelo import ssh
 from robottelo.cli.base import CLIReturnCodeError
-from robottelo.cli.defaults import Defaults
 from robottelo.cli.factory import CLIFactoryError
 from robottelo.cli.factory import make_job_template
-from robottelo.cli.factory import make_location
-from robottelo.cli.factory import make_org
 from robottelo.cli.job_template import JobTemplate
 from robottelo.datafactory import invalid_values_list
-from robottelo.test import CLITestCase
+from robottelo.datafactory import parametrized
 
 TEMPLATE_FILE = 'template_file.txt'
 TEMPLATE_FILE_EMPTY = 'template_file_empty.txt'
 
 
-class JobTemplateTestCase(CLITestCase):
-    """Implements job template tests in CLI."""
+@pytest.fixture(scope="module", autouse=True)
+def module_template():
+    ssh.command(f'''echo '<%= input("command") %>' > {TEMPLATE_FILE}''')
+    ssh.command(f'touch {TEMPLATE_FILE_EMPTY}')
 
-    @classmethod
-    def setUpClass(cls):
-        """Create an organization to be reused in tests."""
-        super().setUpClass()
-        cls.organization = make_org()
-        ssh.command(f'''echo '<%= input("command") %>' > {TEMPLATE_FILE}''')
-        ssh.command(f'touch {TEMPLATE_FILE_EMPTY}')
 
-    @pytest.mark.tier1
-    def test_positive_create_job_template(self):
-        """Create a simple Job Template
+@pytest.mark.tier1
+def test_positive_create_job_template(module_org):
+    """Create a simple Job Template
 
-        :id: a5a67b10-61b0-4362-b671-9d9f095c452c
+    :id: a5a67b10-61b0-4362-b671-9d9f095c452c
 
-        :expectedresults: The job template was successfully created
+    :expectedresults: The job template was successfully created
 
-        :CaseImportance: Critical
-        """
-        template_name = gen_string('alpha', 7)
+    :CaseImportance: Critical
+    """
+    template_name = gen_string('alpha', 7)
+    make_job_template(
+        {
+            'organizations': module_org.name,
+            'name': template_name,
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_jobtemplate.py
```diff
@@ -28,183 +28,178 @@
...
-    @pytest.mark.tier1
-    def test_negative_create_job_template_with_invalid_name(self):
-        """Create Job Template with invalid name
-
-        :id: eb51afd4-e7b3-42c3-81c3-6e18ef3d7efe
-
-        :expectedresults: Job Template with invalid name cannot be created and
-            error is raised
+@pytest.mark.tier1
+def test_negative_create_job_template_with_invalid_name(module_org):
```
jameerpathan111, line 60: @pondrejk should we parameterize this test?
`@pytest.mark.parametrize('name', **parametrized(invalid_values_list()))`

tests/foreman/cli/test_jobtemplate.py
```diff
@@ -28,183 +28,178 @@
...
+        {
+            'organizations': module_org.name,
+            'name': template_name,
+            'file': TEMPLATE_FILE,
+        }
+    )
+    templates = JobTemplate.list({'organization-id': module_org.id})
+    assert len(templates) >= 1
+    Defaults.add({'param-name': 'organization_id', 'param-value': module_org.id})
+    Defaults.add({'param-name': 'location_id', 'param-value': location['id']})
```
omkarkhatavkar, line 178: I think the try block should be from the Default add, in case if the default add get fails on location_id then the finally won't be run

tests/foreman/cli/test_jobtemplate.py
mshriver: I would agree, so that one name failure doesn't block testing of the rest.

JacobCallahan:
It looks like you have a flake failure to resolve.

tpapaioa:
Here's the flake8 error for reference:

```tests/foreman/cli/test_jobtemplate.py:28:1: F401 'robottelo.cli.factory.make_org' imported but unused```

Also, the ssh commands at the top of the module should be moved into a fixture.

File: tests/foreman/cli/test_jobtemplate.py (hunk 2 above)
Comment (mshriver) [reply tests/foreman/cli/test_jobtemplate.py:178]: @omkarkhatavkar you raise a good point about the test flow on failure - will the call to `Defaults.delete` potentially raise an exception themselves.

I think this is a bit of an abused case of try/finally, when the pytest best practice for the intended behavior is the use of `request.addfinalizer` in the form of a function call or decoration.

If there is an exception in the call to `Defaults.add`, the test will result in an error, but that also means the finally block will never run anyway.

If there is a silent failure mode in calling `Defaults.add`, the test will continue, and hit an exception either within the try block or when `Defaults.delete` is called.

I think the best course of action here would be to follow a pattern as outlined below. The teardown phase of the pytest test loop will be responsible for handling the delete of the `Default` entities, and will be safe from any silent failures.  Its a pattern that also is applicable to test functions where entities are sometimes deleted during the course of the test (parametrization, general use fixtures creating entities).

```python
# test function signature updated to add request
# job templates listing, test content up to L176
Defaults.add({'param-name': 'organization_id', 'param-value': module_org.id})
Defaults.add({'param-name': 'location_id', 'param-value': location['id']})

@request.addfinalizer
def defaults_teardown():
    try:
        Defaults.delete({'param-name': 'organization_id'})
    except:
        # maybe log that teardown failed here, since the fixture scope in the framework is unique to this test
        pass
    try:
        Defaults.delete({'param-name': 'location_id'})
    except:
        pass

# Now get on with test content, the above function will execute on teardown
templates = JobTemplate.list()
assert len(templates) >= 1 

```

omkarkhatavkar: @pondrejk make sense if that unwanted test. I gave you a suggestion on the code. If it is not a valid test step then you should be good!

omkarkhatavkar (approved):
ACK

tests/foreman/cli/test_jobtemplate.py
```diff
@@ -21,190 +21,156 @@
...
+from robottelo.datafactory import parametrized
 
 TEMPLATE_FILE = 'template_file.txt'
 TEMPLATE_FILE_EMPTY = 'template_file_empty.txt'
 
 
-class JobTemplateTestCase(CLITestCase):
-    """Implements job template tests in CLI."""
+@pytest.fixture(scope="module")
+def module_template():
```
tpapaioa, line 35: You've moved the `ssh` commands into a fixture, but the fixture isn't being used by any of the tests. If it just needs to run once for all the tests, you could use `pytestmark` at the top of the module:

`pytestmark = [pytest.mark.usefixtures('module_template')]`

mshriver: This is a required change, @pondrejk please add the `usefixtures` mark on the module as described, or I believe `autouse=True` would work well here too.

mshriver (approved):

PR #8493, opened 2021-04-08
tests/robottelo/test_func_shared.py fix and pytest conversion

ppochop:
There was an issue with generating different values where ensuring the values were different was done by `self.assertNotEqual`. This could randomly cause the tests to fail. While the chances of that are generally minimal (e.g. 1/10,000). However, in `test_simple_shared_counter` the odds of that happening are 1/100 which is not that low and it actually failed in one of my PRs. This was fixed by simply removing the assertion statements. To ensure the values are different, `while` statements were added. This may not be necessary because again, the odds of that happening are low but I made a judgement call and put it in.

Because the file was not in pytest, I also converted it. Since this is my first attempt at pytest conversion, please don't spare any comments/suggestions :) 

```
pytest tests/robottelo/test_func_shared.py
collected 15 items                                                             

tests/robottelo/test_func_shared.py ...............                      [100%]

============================= 15 passed in 15.13s ==============================
```

### DIFF
--- a/tests/robottelo/test_func_shared.py
+++ b/tests/robottelo/test_func_shared.py
@@ -2,9 +2,9 @@ import multiprocessing
 import os
 import time
 
+import pytest
 from fauxfactory import gen_integer
 from fauxfactory import gen_string
-from unittest2 import TestCase
 
 from robottelo.decorators.func_shared.file_storage import get_temp_dir
 from robottelo.decorators.func_shared.file_storage import TEMP_FUNC_SHARED_DIR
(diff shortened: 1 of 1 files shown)


omkarkhatavkar:

mshriver (approved):

tests/robottelo/test_func_shared.py
```diff
@@ -154,142 +154,170 @@ def simple_shared_counter_with_exception_not_restored(index=0):
     raise NotRestorableException('error', "I'am not restorable")
 
 
-class FunctionSharedTestCase(TestCase):
-    @classmethod
-    def initiate_namespace_scope(cls):
-        # each_time generate a new name space
+class TestFuncShared:
+    @pytest.fixture(scope="class")
```
JacobCallahan, line 158: Please switch these to single-quote strings.
It looks like there are multiple instances of this throughout the file as well.

JacobCallahan (changes requested):

lhellebr (approved):
LGTM

PR #8502, opened 2021-04-12
[RFE] Upgrade Scenarios sharing workers between pre and post part

jyejare:
-- This is an RFE/request to share the same worker/satellite between the pre and post part of upgrade scenario:

- Due to which the entities created before the upgrade will be accessible on the same satellite post upgrade.

-- This PR adds a plugin to the tests/upgrades/conftest.py
-- This removes the dependency of universal Configuration from __init__.py and configures it as per worker hostname in plugin itself.

### DIFF
--- a/tests/upgrades/scenario_workers.py
+++ b/tests/upgrades/scenario_workers.py
@@ -1,0 +1,58 @@
+import json
+import os
+
+import pytest
+from automation_tools.satellite6.hammer import set_hammer_config
+from fabric.api import env
+
+import robottelo
+from robottelo.config import settings
+
+
+_json_file = 'upgrade_workers.json'
+
+
+def save_worker_id(test_name, worker_id):
+    data = {}
+    if os.path.exists(_json_file):
+        with open(_json_file) as rjson:
+            data = json.load(rjson)
+    with open(_json_file, 'w') as wjson:
+        # Removing the parameter name from test name before save
+        test_name = test_name.split('[')[0] if '[' in test_name else test_name
+        data.update({test_name: worker_id})
+        wjson.seek(0)
+        json.dump(data, wjson)
+
+
+def get_worker_id(test_name):
+    with open(_json_file) as rjson:
+        data = json.load(rjson)
+        return data.get(test_name)
+
+
+@pytest.fixture(autouse=True)
+def save_pretest_worker_id(request, worker_id):
+    """Saves the worker id of pre_upgrade test in json"""
+    if request.node.get_closest_marker('pre_upgrade'):
+        save_worker_id(request.node.name, worker_id)
+
+
+@pytest.fixture(autouse=True)
+def set_posttest_worker_id(request):
+    """Retrieves worker_id of pre_upgrade and sets the same to run post_upgrade"""
+    post_upgrade = request.node.get_closest_marker('post_upgrade')
+    if post_upgrade:
+        depend_test = post_upgrade.kwargs.get('depend_on')
+        if depend_test:
+            pre_test_name = depend_test.__name__
+            # settings.configure()
+            cache_proxy = robottelo.config.settings_proxy._cache
+            # Read the worker id of pre test name
+            worker_int = int(get_worker_id(pre_test_name).replace('gw', ''))
+            cache_proxy['server.hostname'] = settings.server.hostnames[worker_int]
+            settings.configure_nailgun()
+            settings.configure_airgun()
+            env.host_string = settings.server.hostname
+            env.user = 'root'
+            set_hammer_config()
--- a/tests/upgrades/scenario_workers.py
+++ b/tests/upgrades/scenario_workers.py
@@ -1,5 +1,5 @@
 import json
-import os
+from pathlib import Path
 
 import pytest
 from automation_tools.satellite6.hammer import set_hammer_config
@@ -12,47 +12,52 @@ from robottelo.config import settings
 _json_file = 'upgrade_workers.json'
 
 
-def save_worker_id(test_name, worker_id):
+def save_worker_hostname(test_name):
     data = {}
-    if os.path.exists(_json_file):
-        with open(_json_file) as rjson:
-            data = json.load(rjson)
-    with open(_json_file, 'w') as wjson:
-        # Removing the parameter name from test name before save
-        test_name = test_name.split('[')[0] if '[' in test_name else test_name
-        data.update({test_name: worker_id})
-        wjson.seek(0)
-        json.dump(data, wjson)
+    json_file = Path(_json_file)
+    if json_file.exists():
+        data = json.loads(json_file.read_text())
+    # Removing the parameter name from test name before save
+    test_name = test_name.split('[')[0] if '[' in test_name else test_name
+    data.update({test_name: settings.server.hostname})
+    json_file.write_text(json.dumps(data))
 
 
-def get_worker_id(test_name):
+def get_worker_hostname(test_name):
     with open(_json_file) as rjson:
         data = json.load(rjson)
         return data.get(test_name)
 
 
 @pytest.fixture(autouse=True)
-def save_pretest_worker_id(request, worker_id):
+def save_pre_upgrade_worker_hostname(request):
     """Saves the worker id of pre_upgrade test in json"""
     if request.node.get_closest_marker('pre_upgrade'):
-        save_worker_id(request.node.name, worker_id)
+        save_worker_hostname(request.node.name)
 
 
 @pytest.fixture(autouse=True)
-def set_posttest_worker_id(request):
-    """Retrieves worker_id of pre_upgrade and sets the same to run post_upgrade"""
+def set_post_upgrade_hostname(request):
+    """Retrieves worker hostname of pre_upgrade and sets the same to run post_upgrade
+
+    The limitation of this implementation is that the XDIST_BEHAVIOR won't be observed,
+    and tests running under xdist will not be isolated.
+    """
     post_upgrade = request.node.get_closest_marker('post_upgrade')
     if post_upgrade:
         depend_test = post_upgrade.kwargs.get('depend_on')
         if depend_test:
             pre_test_name = depend_test.__name__
-            # settings.configure()
+            settings.configure()
             cache_proxy = robottelo.config.settings_proxy._cache
+            pre_hostname = get_worker_hostname(pre_test_name)
+            if (pre_hostname is None) or (pre_hostname not in settings.server.hostnames):
+                pytest.skip(
+                    'Skipping the post_upgrade test as the pre_upgrade hostname is not found!'
+                )
             # Read the worker id of pre test name
-            worker_int = int(get_worker_id(pre_test_name).replace('gw', ''))
-            cache_proxy['server.hostname'] = settings.server.hostnames[worker_int]
+            cache_proxy['server.hostname'] = env.host_string = pre_hostname
+            env.user = 'root'
             settings.configure_nailgun()
             settings.configure_airgun()
-            env.host_string = settings.server.hostname
-            env.user = 'root'
             set_hammer_config()
--- a/tests/upgrades/scenario_workers.py
+++ b/tests/upgrades/scenario_workers.py
@@ -5,16 +5,17 @@ import pytest
 from automation_tools.satellite6.hammer import set_hammer_config
 from fabric.api import env
 
-import robottelo
+from robottelo.config import configure_airgun
+from robottelo.config import configure_nailgun
 from robottelo.config import settings
 
 
 _json_file = 'upgrade_workers.json'
+json_file = Path(_json_file)
 
 
 def save_worker_hostname(test_name):
     data = {}
-    json_file = Path(_json_file)
     if json_file.exists():
         data = json.loads(json_file.read_text())
     # Removing the parameter name from test name before save
@@ -23,10 +24,14 @@ def save_worker_hostname(test_name):
     json_file.write_text(json.dumps(data))
 
 
-def get_worker_hostname(test_name):
-    with open(_json_file) as rjson:
-        data = json.load(rjson)
-        return data.get(test_name)
+@pytest.fixture(scope='session')
+def shared_workers():
+    if json_file.exists():
+        return json.loads(json_file.read_text())
+
+
+def get_worker_hostname_from_testname(test_name, shared_workers):
+    return shared_workers.get(test_name)
 
 
 @pytest.fixture(autouse=True)
@@ -37,7 +42,7 @@ def save_pre_upgrade_worker_hostname(request):
 
 
 @pytest.fixture(autouse=True)
-def set_post_upgrade_hostname(request):
+def set_post_upgrade_hostname(request, shared_workers):
     """Retrieves worker hostname of pre_upgrade and sets the same to run post_upgrade
 
     The limitation of this implementation is that the XDIST_BEHAVIOR won't be observed,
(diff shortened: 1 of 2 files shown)


tests/upgrades/scenario_workers.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def save_worker_id(test_name, worker_id):
+    data = {}
+    if os.path.exists(_json_file):
+        with open(_json_file) as rjson:
+            data = json.load(rjson)
+    with open(_json_file, 'w') as wjson:
+        # Removing the parameter name from test name before save
+        test_name = test_name.split('[')[0] if '[' in test_name else test_name
```
JacobCallahan, line 21:
```suggestion
        test_name = test_name.split('[')[0]
```

tests/upgrades/scenario_workers.py
```diff
@@ -0,0 +1,55 @@
...
+@pytest.fixture(autouse=True)
+def set_posttest_worker_id(request):
+    """Retrieves worker_id of pre_upgrade and sets the same to run post_upgrade"""
+    post_upgrade = request.node.get_closest_marker('post_upgrade')
+    if post_upgrade:
+        depend_test = post_upgrade.kwargs.get('depend_on')
+        if depend_test:
+            pre_test_name = depend_test.__name__
+            settings.configure()
+            worker_int = int(get_worker_id(pre_test_name).replace('gw', ''))
```
JacobCallahan, line 49: you can add `worker_id` as an argument to this fixture and use that instead of running get_worker_id.

tests/upgrades/scenario_workers.py
```diff
@@ -0,0 +1,55 @@
...
+            settings.server.hostname = settings.server.hostnames[worker_int]
```
JacobCallahan, line 50: this direct assignment won't reliably work because the facade caches accessed values.

JacobCallahan (changes requested):

tests/upgrades/scenario_workers.py
jyejare: The get_worker_id is required to pick the same worker id as pre part of the scenario has used from saved `upgrade_workers.json`

tests/upgrades/scenario_workers.py
jyejare: I have tested this by creating a spcial test module and it works :)

Also, I am reinitiating the configuration on line number 48, so that refresh everything.

JacobCallahan: I don't believe configure does anything anymore. The facade is mapping that call to a lambda that returns None.

https://github.com/SatelliteQE/robottelo/blob/41f3c4863f32c89195853f4bd07834523b47108c/robottelo/config/facade.py#L399

jyejare: Though it's caching the accessed values we are alternating the server name and it works fine!

Here is the Test Module I created:

```
from upgrade_tests import post_upgrade
from upgrade_tests import pre_upgrade
from nailgun import entities

org_id = 1


class Testscene1:

    prod_name = 'test11'

    @pre_upgrade
    def test_pre_1(self):
        prod = entities.Product(name=self.prod_name, organization=org_id).create()
        assert prod.name == self.prod_name

    @post_upgrade(depend_on=test_pre_1)
    def test_post_1(self):
        prod = entities.Product(organization=org_id).search(query={'search': f'name={self.prod_name}'})
        assert prod[0].name == self.prod_name


class Testscene2:

    prod_name = 'test22'

    @pre_upgrade
    def test_pre_2(self):
        prod = entities.Product(name=self.prod_name, organization=org_id).create()
        assert prod.name == self.prod_name

    @post_upgrade(depend_on=test_pre_2)
    def test_post_2(self):
        prod = entities.Product(organization=org_id).search(query={'search': f'name={self.prod_name}'})
        assert prod[0].name == self.prod_name


class Testscene3:

    prod_name = 'test33'

    @pre_upgrade
    def test_pre_3(self):
        prod = entities.Product(name=self.prod_name, organization=org_id).create()
        assert prod.name == self.prod_name

    @post_upgrade(depend_on=test_pre_3)
    def test_post_3(self):
        prod = entities.Product(organization=org_id).search(query={'search': f'name={self.prod_name}'})
        assert prod[0].name == self.prod_name


class Testscene4:

    prod_name = 'test44'

    @pre_upgrade
    def test_pre_4(self):
        prod = entities.Product(name=self.prod_name, organization=org_id).create()
        assert prod.name == self.prod_name

    @post_upgrade(depend_on=test_pre_4)
    def test_post_4(self):
        prod = entities.Product(organization=org_id).search(query={'search': f'name={self.prod_name}'})
        assert prod[0].name == self.prod_name

```

jyejare: @JacobCallahan Do you have any full-proof way to set the server name?

JacobCallahan: To make a lasting change, I had to modify the value stored in the proxy's cache itself.
https://github.com/SatelliteQE/robottelo/blob/master/pytest_fixtures/xdist.py#L17-L19

jyejare: @JacobCallahan The suggested solution doesn't seem working for me, but the existing approach works well! I did negative testing as well and it shows the failures as expected.

Do you think I should implmemt the same allignment you have done ? Also, the facade is going away soon, I think we can merge this without wasting much effort on a thing that will move away. What do you think?

jyejare: I didn't try to set the server hostname to proxy btw, just tried to reset the cache and kept the existing way of setting the server hostname! Also, I am setting one sat for each worker so I don't need alignments that you have done in xdist.py.

@mshriver Thoughts ?

JacobCallahan: I think you might be lucking out with timing in your tests, but I can't reliably say for sure that this will work when used in conjunction with everything else.

I won't hold on that, but please note the behavior in automation runs after this is merged.

@mshriver can still weigh in.

jyejare: @mshriver Ping !

jyejare: @mshriver @JacobCallahan Pings!!!

JacobCallahan: Pong 🙃
We'll roll in Mike tomorrow

jyejare: @JacobCallahan Fixed this ! Now setting the server hostname in cache ! Also, tested this and its working great !

tests/upgrades/scenario_workers.py
```diff
@@ -0,0 +1,58 @@
...
+import robottelo
+from robottelo.config import settings
+
+
+_json_file = 'upgrade_workers.json'
+
+
+def save_worker_id(test_name, worker_id):
+    data = {}
+    if os.path.exists(_json_file):
```
mshriver, line 17: I'd encourage you to start using pathlib to handle these file interactions.

```
from pathlib import Path

json_file = Path(_json_file)
if json_file.exists():
    data = json.load(json_file.read_text())
data.update({(test_name.split('[')[0] if '[' in test_name else test_name): worker_id})
json_file.write_text(json.dump(data))
```

tests/upgrades/scenario_workers.py
```diff
@@ -0,0 +1,58 @@
...
+    if request.node.get_closest_marker('pre_upgrade'):
+        save_worker_id(request.node.name, worker_id)
+
+
+@pytest.fixture(autouse=True)
+def set_posttest_worker_id(request):
+    """Retrieves worker_id of pre_upgrade and sets the same to run post_upgrade"""
+    post_upgrade = request.node.get_closest_marker('post_upgrade')
+    if post_upgrade:
+        depend_test = post_upgrade.kwargs.get('depend_on')
```
mshriver, line 46: I think this should handle when depend_test isn't found - otherwise no change will be made to `settings_proxy._cache`, and the test will run against whatever host this xdist worker has been aligned with.

The test should likely be skipped, since you can't identify what satellite target the test should run against.

tests/upgrades/scenario_workers.py
```diff
@@ -0,0 +1,58 @@
...
+        if depend_test:
+            pre_test_name = depend_test.__name__
+            # settings.configure()
+            cache_proxy = robottelo.config.settings_proxy._cache
+            # Read the worker id of pre test name
+            worker_int = int(get_worker_id(pre_test_name).replace('gw', ''))
```
mshriver, line 52: The return from `get_worker_id` could be `None`, and you should handle that situation here and skip the test because you can't identify what satellite target the test should be executed against.

tests/upgrades/scenario_workers.py
```diff
@@ -0,0 +1,58 @@
...
+@pytest.fixture(autouse=True)
+def save_pretest_worker_id(request, worker_id):
+    """Saves the worker id of pre_upgrade test in json"""
+    if request.node.get_closest_marker('pre_upgrade'):
+        save_worker_id(request.node.name, worker_id)
+
+
+@pytest.fixture(autouse=True)
+def set_posttest_worker_id(request):
+    """Retrieves worker_id of pre_upgrade and sets the same to run post_upgrade"""
```
mshriver, line 43: This docblock and function naming is misleading.

The function doesn't change the worker ID, it changes the target server hostname for the worker where this fixture is executing.

This fixture also breaks isolation of the satellite targets on their xdist workers, implicitly changing to a `balance` XDIST_BEHAVIOR, since multiple workers could end up setting the same `server.hostname` as a target.

The docblock needs to clearly state the limitation of this implementation, that XDIST_BEHAVIOR won't be observed, and tests running under xdist will not be isolated.

tests/upgrades/scenario_workers.py
```diff
@@ -0,0 +1,58 @@
...
+    post_upgrade = request.node.get_closest_marker('post_upgrade')
+    if post_upgrade:
+        depend_test = post_upgrade.kwargs.get('depend_on')
+        if depend_test:
+            pre_test_name = depend_test.__name__
+            # settings.configure()
+            cache_proxy = robottelo.config.settings_proxy._cache
+            # Read the worker id of pre test name
+            worker_int = int(get_worker_id(pre_test_name).replace('gw', ''))
+            cache_proxy['server.hostname'] = settings.server.hostnames[worker_int]
```
mshriver, line 53: There are some massive assumptions made in this implementation that concern me.

You need to align the test cases with the satellite target where depend_test actually ran - not the worker that it ran on.  If XDIST_BEHAVIOR was `balance` in pre_upgrade testing, or if any worker restarted and fetched a new on-demand satellite, then this mapping will be off.

The fixture implementing on-demand behavior to align xdist workers also runs in parallel on multiple threads, and its possible that worker `gw8` doesn't execute against the host from `server.hostnames` at index #8.

In fact, `server.hostnames` is only populated automatically when an inventory filter is supplied and a broker inventory is read.

mshriver: @jyejare let's discuss more tomorrow.  I think a fundamental change here would just be to store the hostname in pre_upgrade tests, instead of storing the worker ID. Then in post_upgrade, resolve the target hostname from the `depend_test`, instead of resolving a worker ID and then trying to infer a satellite hostname from that worker ID.

tests/upgrades/scenario_workers.py
jyejare: @mshriver No, I think thats what is intended. 

If dependent pre-test is not their for any `post` test, we are absolutely ok to run it on any available satellite provided by xdist.

tests/upgrades/scenario_workers.py
jyejare: Made the changes as expected.

tests/upgrades/scenario_workers.py
jyejare: This code doesnt works for me ! Looking ..

tests/upgrades/scenario_workers.py
jyejare: Fixed!

tests/upgrades/scenario_workers.py
jyejare: Please check the code is updated.

tests/upgrades/scenario_workers.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+_json_file = 'upgrade_workers.json'
+
+
+def save_worker_hostname(test_name):
+    data = {}
+    json_file = Path(_json_file)
+    if json_file.exists():
+        data = json.loads(json_file.read_text())
```
mshriver, line 19: :clap: Isn't pathlib great?

jyejare: It is, ofcourse :)

tests/upgrades/scenario_workers.py
```diff
@@ -0,0 +1,63 @@
...
+    # Removing the parameter name from test name before save
+    test_name = test_name.split('[')[0] if '[' in test_name else test_name
+    data.update({test_name: settings.server.hostname})
+    json_file.write_text(json.dumps(data))
+
+
+def get_worker_hostname(test_name):
+    with open(_json_file) as rjson:
+        data = json.load(rjson)
```
mshriver, line 28: Consider opening an reading this file at the module level, or in a session scoped fixture. (if it exists, safely)

That way you don't open and JSON parse the file once for every test function in the post_upgrade session.

tests/upgrades/scenario_workers.py
```diff
@@ -0,0 +1,63 @@
...
+
+    The limitation of this implementation is that the XDIST_BEHAVIOR won't be observed,
+    and tests running under xdist will not be isolated.
+    """
+    post_upgrade = request.node.get_closest_marker('post_upgrade')
+    if post_upgrade:
+        depend_test = post_upgrade.kwargs.get('depend_on')
+        if depend_test:
+            pre_test_name = depend_test.__name__
+            settings.configure()
```
mshriver, line 51: @JacobCallahan Considering #8662, this call to `settings.configure()` should no longer be necessary here right?

tests/upgrades/scenario_workers.py
```diff
@@ -0,0 +1,63 @@
...
+            cache_proxy = robottelo.config.settings_proxy._cache
+            pre_hostname = get_worker_hostname(pre_test_name)
+            if (pre_hostname is None) or (pre_hostname not in settings.server.hostnames):
+                pytest.skip(
+                    'Skipping the post_upgrade test as the pre_upgrade hostname is not found!'
+                )
+            # Read the worker id of pre test name
+            cache_proxy['server.hostname'] = env.host_string = pre_hostname
```
mshriver, line 59: Let's coordinate this with #8662, because the cache_proxy is going away, and this should be written for post-8662 behavior.

tests/upgrades/scenario_workers.py
JacobCallahan: Correct. Settings are just settings now and values load automatically

JacobCallahan:
Let's revisit the parts Mike mentioned post-8662 merge

jyejare:
@JacobCallahan ACK !

mshriver:
@jyejare #8662 is in, let's get this merged tomorrow.

mshriver (approved):

tests/upgrades/scenario_workers.py
```diff
@@ -0,0 +1,68 @@
...
+    data.update({test_name: settings.server.hostname})
+    json_file.write_text(json.dumps(data))
+
+
+@pytest.fixture(scope='session')
+def shared_workers():
+    json_file = Path(_json_file)
+    if json_file.exists():
+        with open(_json_file) as rjson:
+            return json.load(rjson)
```
JacobCallahan, line 32:
```suggestion
        return json.loads(json_file.read_text())
```
same pattern as above

jyejare:
@JacobCallahan Please revisit ! Fixed the final comment !

JacobCallahan (approved):

PR #8508, opened 2021-04-13
update RP portal parameter names to match those in the settings file

rplevka:


mshriver (approved):

mshriver:
The report portal tool for failure collection is using these, but needs a bit of an overhaul anyway because the launch names were hardcoded an no longer match.

JacobCallahan (approved):

PR #8510, opened 2021-04-13
Add skip to modulestream test case due to bz

latran:
There's a regression issue that's failing this test case.  I'm adding a skip until the bz is fixed.

### DIFF
--- a/tests/foreman/ui/test_modulestreams.py
+++ b/tests/foreman/ui/test_modulestreams.py
@@ -45,6 +45,7 @@ def module_yum_repo(module_product):
     return yum_repo
 
 
+@pytest.mark.skip_if_open("BZ:1948758")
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_modulestreams.py
```diff
@@ -45,6 +45,7 @@ def module_yum_repo(module_product):
     return yum_repo
 
 
+@pytest.mark.skip_if_open("BZ:1948758")
 @pytest.mark.tier2
 def test_positive_module_stream_details_search_in_repo(session, module_org, module_yum_repo):
```
omkarkhatavkar, line 50: `:BZ: 1948758` add this line as well

mshriver (approved):

PR #8511, opened 2021-04-14
convert cli/test_discoveredhost to pytest

peterdragun:
Test results:
```====== 4 passed, 11 deselected in 725.37s (0:12:05) =======```

### DIFF
--- a/tests/foreman/cli/test_discoveredhost.py
+++ b/tests/foreman/cli/test_discoveredhost.py
@@ -30,501 +30,508 @@ from robottelo.cli.template import Template
 from robottelo.datafactory import gen_string
 from robottelo.decorators import skip_if_not_set
 from robottelo.libvirt_discovery import LibvirtGuest
-from robottelo.test import CLITestCase
+
+pytestmark = [pytest.mark.run_in_one_thread]
 
 
-@pytest.mark.run_in_one_thread
-class DiscoveredTestCase(CLITestCase):
-    """Implements Foreman discovery CLI tests."""
+def _assertdiscoveredhost(self, hostname):
+    """Check if host is discovered and information about it can be
+    retrieved back
 
-    def _assertdiscoveredhost(self, hostname):
-        """Check if host is discovered and information about it can be
-        retrieved back
+    Introduced a delay of 300secs by polling every 10 secs to get expected
+    host
+    """
+    for _ in range(30):
+        try:
+            discovered_host = DiscoveredHost.info({'name': hostname})
+        except CLIReturnCodeError:
+            sleep(10)
+            continue
+        return discovered_host
 
-        Introduced a delay of 300secs by polling every 10 secs to get expected
-        host
-        """
-        for _ in range(30):
-            try:
-                discovered_host = DiscoveredHost.info({'name': hostname})
-            except CLIReturnCodeError:
-                sleep(10)
-                continue
-            return discovered_host
 
-    @classmethod
-    @skip_if_not_set('vlan_networking')
-    def setUpClass(cls):
-        """Steps to Configure foreman discovery
+@skip_if_not_set('vlan_networking')
+@pytest.fixture(scope='class')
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by omkarkhatavkar** [comment]

**Review by mshriver — APPROVED** [review]

File: tests/foreman/cli/test_discoveredhost.py [hunk 1]
```diff
@@ -52,9 +51,9 @@ def _assertdiscoveredhost(self, hostname):
                 continue
             return discovered_host
 
-    @classmethod
     @skip_if_not_set('vlan_networking')
-    def setUpClass(cls):
+    @pytest.fixture(scope="class")
```
JacobCallahan, line 55: Please switch these to single-quote strings.

tests/foreman/cli/test_discoveredhost.py
```diff
@@ -79,36 +76,34 @@ def setUpClass(cls):
...
 
         # Get default settings values
-        cls.default_discovery_loc = Settings.list({'search': 'name=%s' % 'discovery_location'})[0]
-        cls.default_discovery_org = Settings.list(
-            {'search': 'name=%s' % 'discovery_organization'}
-        )[0]
-        cls.default_discovery_auto = Settings.list({'search': 'name=%s' % 'discovery_auto'})[0]
+        default_discovery_loc = Settings.list({'search': 'name=%s' % 'discovery_location'})[0]
+        default_discovery_org = Settings.list({'search': 'name=%s' % 'discovery_organization'})[0]
+        default_discovery_auto = Settings.list({'search': 'name=%s' % 'discovery_auto'})[0]
```
JacobCallahan, line 85: use f-strings here

tests/foreman/cli/test_discoveredhost.py
```diff
@@ -30,11 +30,10 @@
 from robottelo.datafactory import gen_string
 from robottelo.decorators import skip_if_not_set
 from robottelo.libvirt_discovery import LibvirtGuest
-from robottelo.test import CLITestCase
 
 
 @pytest.mark.run_in_one_thread
-class DiscoveredTestCase(CLITestCase):
+class TestDiscoveredHost:
```
JacobCallahan, line 36: is there a need for this to be a class instead of module-level tests?

JacobCallahan (changes requested):

tests/foreman/cli/test_discoveredhost.py
peterdragun: There is not, but if I used module-level tests I would need to mark every single test to run in one thread. Personally I prefer marking whole class, but I can change that.

tests/foreman/cli/test_discoveredhost.py
peterdragun: f-strings? :thinking: this is just merging 2 strings. I could use one string instead, but I think someone tried to make it more  readable by separating it.

tests/foreman/cli/test_discoveredhost.py
mshriver: You can mark the module, which will apply the mark to all test cases.

```pytestmark=[pytest.mark.run_in_one_thread]```  At the top of the module.

peterdragun: Thanks, didn't know about this option.

mshriver: Yep! It's not a perfect solution, if there is any test case where you _don't_ want that mark, you're SOL, there's no clean way to remove the mark on that single test.

tests/foreman/cli/test_discoveredhost.py
mshriver: It became part of our coding standards, and most of these uses cases have been converted. Its a great time to convert this one as well, I don't think it will reduce the readability since an f-string isn't even needed here.

`Settings.list({'search': 'name=discovery_location'})[0]`

PR #8516, opened 2021-04-14
changed assignee name for ContentVV

ColeHiggins2:
The assignee's name was incorrect. Changed to the official owner.


JacobCallahan (approved):

latran (approved):
ack

PR #8517, opened 2021-04-14
Replace logging with LogZero

mshriver:
This grew out of the desire to decouple the logging configuration from the settings module - a need described by @mirekdlugosz  in #8300.

The logging module's file config mechanism is being replaced by a dictionary config.  I was going to replace our file config with dictionary config, but with that large of a change we might as well consider changing logging modules altogether.

Moved get_project_root function to robottelo.config and updated callers

Add robottelo/logging.py, where logzero loggers are defined
loggers should be created for all of the loggers from logging.conf

New logging.yaml file is read directly by robottelo.logging in order to
be able to locally modify logging config without modifying the code.

Third party logging level is still being set, but the configuration pattern now covers widgetastic, navmazing, airgun, nailgun, and robozilla as third party instead of with their own handlers.

I have tested CLI/API/UI tests locally, seeing the expected log level messages on the console and in `logs/robottelo.log` from robottelo, robottelo.config, robottelo.collection, paramiko, airgun, widgetastic, navmazing, nailgun, and other third party loggers.


robottelo/config/__init__.py
```diff
@@ -11,6 +12,16 @@
...
 logger = logging.getLogger('robottelo.config')
 
+
+def get_project_root():
+    """Return the path to the Robottelo project root directory.
+
+    :return: A directory path.
+    :rtype: str
+    """
+    return os.path.realpath(os.path.join(os.path.dirname(__file__), os.pardir, os.pardir))
```
JacobCallahan, line 22: if you used pathlib's Path class, you can do the same with the below. that is unless I'm confused about exactly what's going on here.
```suggestion
    return str(Path().absolute())
```
if you need the absolute path of this file `str(Path(__file__).absolute())`

robottelo/logging.py
```diff
@@ -0,0 +1,54 @@
+import os
+from logging import Formatter
+
+import yaml
+from attrdict import AttrDict
+from logzero import setup_logger
+
+from robottelo.config import get_project_root
+
+ROBOTTELO = 'robottelo'
```
JacobCallahan, line 10: :) why?

robottelo/logging.py
```diff
@@ -0,0 +1,54 @@
...
+
+logging_yaml = AttrDict(yaml.load_all(os.path.join(get_project_root(), 'logging.yaml')))
```
JacobCallahan, line 12: if fet_project_root returned a pathlib object..
```suggestion
logging_yaml = AttrDict(yaml.load_all(get_project_root().joinpath('logging.yaml')))
```

JacobCallahan:
I love where this is going! imo pathlib would make some things easier, but not a requirement.

robottelo/config/__init__.py
mshriver: I'm down to use pathlib throughout this.

robottelo/logging.py
mshriver: You'll see! I was considering modifying the default logzero logger instance, so that the default for the framework is robottelo.log without it having to be specified by name.

robottelo/config/__init__.py
mshriver: Here it is resolving a grand-parent directory, which should be the root robottelo directory relative to the file.

robottelo/logging.py
JacobCallahan: ahhh so maybe something like this could be done?
```python
from robottelo import logging
logging.ROBOTTELO = f"robottelo_{xdist_id}"
logging.setup...
```

robottelo/config/__init__.py
```diff
@@ -1,15 +1,19 @@
...
 from robottelo.config.facade import SettingsFacade
 from robottelo.config.facade import SettingsNodeWrapper
+from robottelo.errors import ImproperlyConfigured
+from robottelo.logging import config_logger as logger
 
-logger = logging.getLogger('robottelo.config')
+
+# Path instance for the Robottelo project root directory
+robottelo_root_dir = Path().resolve().parent.parent
+robottelo_log_dir = robottelo_root_dir.joinpath('logs')
```
JacobCallahan, line 16: +100!

robottelo/decorators/func_locker.py
```diff
@@ -41,16 +41,14 @@ def test_that_conflict_with_test_to_lock(self)
...
 import os
 import tempfile
 from contextlib import contextmanager
 
 from pytest_services.locks import file_lock
 
 from robottelo.config import settings
-
-logger = logging.getLogger('robottelo')
+from robottelo.logging import logger
```
JacobCallahan, line 51: should this be a logzero import?

robottelo/config/__init__.py
JacobCallahan: i see these are also defined in robottelo.logging. could you use the same values defined there in place of these as well?

mshriver: Because `robottelo.config.__init__` and `robottelo.config.facade` and `robottelo.logging` import from each other, trying to import them turns on 'circular import mode' where everything blows up.

Maybe there is a better third place for these, or we can consolidate them when the facade gets removed and imports aren't quite so circular.

robottelo/decorators/func_locker.py
mshriver: I'd prefer for modules within the framework to import the logger directly from `robottelo.logging` - that way if any further changes are made to the logger, or it gets replaced, the framework imports don't change.

robottelo/config/__init__.py
JacobCallahan: I don't see any robottelo imports in robottelo.logging, so that might be the best place for the calculation and config could take the values from there. Ultimately, the computational impact is low, so this isn't close to a blocker.

robottelo/decorators/func_locker.py
JacobCallahan: good call

robottelo/logging.py
```diff
@@ -19,50 +22,68 @@
...
+configure_third_party_logging()
+
+
+# collection_logfile = robottelo_log_dir.joinpath(f'{ROBOTTELO}_collection.log')
+# collection_logfile.parent.mkdir(parents=True, exist_ok=True)
 collection_logger = logzero.setup_logger(
     name=f'{ROBOTTELO}.collection',
     level=logging_yaml.collection.level,
-    logfile=collection_logfile,
+    logfile=str(robottelo_log_file),
```
JacobCallahan, line 70: do you need the explicit string conversion here?

mshriver: I don't think its _needed_, but its following the docblock for `logzero.setup_logger`, where this parameter is supposed to be a string file path.

tests/robottelo/test_config_casts.py
```diff
@@ -60,6 +58,8 @@ def cast_logging_level(self):
         return casts.LoggingLevel()
 
     def test_cast_logging_level(self, cast_logging_level):
+        import logging
+
```
Gauravtalreja1, line 62: should this be a log zero import? like `from robottelo.logging import logger`

mshriver: Not in this case, because we're using the log level constants defined in the built-in `logging` module. Good eye though!

mshriver:
trigger: test-robottelo
pytest: -n 4 --importance Critical -m tier1 tests/foreman/cli tests/foreman/api tests/foreman/ui

mshriver:
trigger: test-robottelo
pytest: --importance Critical -m tier1 tests/foreman/cli

mshriver:
trigger: test-robottelo
pytest: --importance Critical -m tier1 tests/foreman/cli

tests/foreman/conftest.py
```diff
@@ -85,7 +88,7 @@ def pytest_collection_modifyitems(session, items, config):
     the items in-place.
     """
 
-    log(f"Collected {len(items)} test cases")
+    logger.debug(f"Collected {len(items)} test cases")
```
JacobCallahan, line 91: Please switch these to single-quote strings. ;)

tests/robottelo/test_config_casts.py
Gauravtalreja1: Okay gotcha, just a question, can't we just replace this with `logzero.loglevel` ? (optional)

pytest_plugins/logging_hooks.py
```diff
@@ -0,0 +1,69 @@
...
+            for handler in handlers_to_remove:
+                logger.removeHandler(handler)
+            file_handler = logging.FileHandler(
+                robottelo_log_dir.joinpath(f'robottelo_{worker_id}.log')
+            )
+            file_handler.set_name(f'robottelo.{worker_id}')
+            file_handler.setFormatter(worker_formatter)
+            logger.addHandler(file_handler)
+
+            if use_rp_logger:
```
mshriver, line 56: Still testing this piece, having a stability issue with our RP instance.

pytest_plugins/manual_skipped.py
```diff
@@ -28,11 +25,16 @@ def pytest_collection_modifyitems(session, config, items):
...
             if mark_skipped:
+                logger.debug(f'Marking collected stubbed test "{item.nodeid}" to skip')
                 item.add_marker(marker=pytest.mark.skip(reason=NOT_IMPLEMENTED))
             continue
 
         # The test case is stubbed, but --include-stubbed was NOT passed, deselect the item
         if stub_marked and not include_stubbed:
+            logger.debug(
+                f'Deselecting stubbed test {item.nodeid}, '
+                f'use --include-stubbed to include in collection'
```
JacobCallahan, line 36:
```suggestion
                'use --include-stubbed to include in collection'
```

tests/foreman/conftest.py
```diff
@@ -85,7 +30,7 @@ def pytest_collection_modifyitems(session, items, config):
     the items in-place.
     """
 
-    log(f"Collected {len(items)} test cases")
+    collection_logger.debug(f"Collected {len(items)} test cases")
```
JacobCallahan, line 33:
```suggestion
    collection_logger.debug(f'Collected {len(items)} test cases')
```

tests/foreman/conftest.py
```diff
@@ -97,7 +42,7 @@ def pytest_collection_modifyitems(session, items, config):
         if deselect:
             deselected_items.append(item)
             reason = deselect.kwargs.get('reason', deselect.args)
-            log(f"Deselected test '{item.name}' reason: {reason}")
+            collection_logger.debug(f"Deselected test '{item.name}' reason: {reason}")
```
JacobCallahan, line 45:
```suggestion
            collection_logger.debug(f'Deselected test "{item.name}" reason: {reason}')
```

JacobCallahan (changes requested):
small things. otherwise, this is looking great!

JacobCallahan (approved):
ACK on current edition. Merge pending updates and testing sign-off

pytest_plugins/logging_hooks.py
mshriver: I've got RP live logging working correctly now. I wasn't able to do it in a `pytest_sessionstart` hook, I had to do it in a fixture. But the logger doesn't have to be used from the fixture - the root logger imported through `robottelo.logging` will log to RP just fine.

pytest_plugins/logging_hooks.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    worker_formatter = logzero.LogFormatter(
+        fmt=f'%(asctime)s - {worker_id} - %(name)s - %(levelname)s - %(message)s',
+        datefmt=DEFAULT_DATE_FORMAT,
+    )
+    use_rp_logger = hasattr(request.node.config, 'py_test_service')
+    if use_rp_logger:
+        logging.setLoggerClass(RPLogger)
+
+    if is_xdist_worker(request):
```
mshriver, line 37: This restricts RP live logging to only run when using xdist. This is the same logic that was applied before - but it doesn't _have_ to be this way.  We generally will only be using RP result upload when using xdist in CI pipelines, which I think is why its setup this way. If you run without xdist and with RP options specified, it will upload results (in a tree, even if you set hierarchy to False), you just won't get any logs included.

pytest_plugins/logging_hooks.py
```diff
@@ -0,0 +1,69 @@
...
+            )
+            worker_handler.set_name(f'{worker_id}')
+            worker_handler.setFormatter(worker_formatter)
+            worker_handler.setLevel(worker_log_level)
+            logger.addHandler(worker_handler)
+
+            if use_rp_logger:
+                rp_handler = RPLogHandler(request.node.config.py_test_service)
+                rp_handler.setFormatter(worker_formatter)
+                # logger.addHandler(rp_handler)
```
mshriver, line 61: Yeah, so, weird right?  Well, I don't quite understand _why_, but RP's own documentation never adds the handler to the logger.  Calling `logging.setLoggerClass(RPLogger)` up above injects a hook into the core logger that forces RPLogHandler.emit to get called, even without adding the handler to the logger. Why they did this instead of just letting people control the handlers.... I don't know.

mshriver: It also appears to completely ignore the formatter that is set on the `RPLogHandler` instance, and `RPLogHandler.emit` only emits the message, unformatted. This is no change in behavior compared to the current RP logging hook in robottelo.

robottelo/config/__init__.py
```diff
@@ -49,8 +48,6 @@
 settings = SettingsNodeWrapper(settings_proxy)
 settings.configure_nailgun()
 settings.configure_airgun()
-settings.configure_logging()
```
mshriver, line 52: @JacobCallahan This makes your dynaconf end game PR just a little bit smaller :)

robottelo/config/facade.py
```diff
@@ -185,14 +182,18 @@ def configure_nailgun(self):
         def patched_gpgkey_init(self, server_config=None, **kwargs):
             """Set a default value on the ``content`` field."""
             gpgkey_init(self, server_config, **kwargs)
+            from robottelo.config import robottelo_root_dir
```
mshriver, line 185: This code won't be living for very long.

robottelo/logging.py
```diff
@@ -0,0 +1,89 @@
...
+with robottelo_root_dir.joinpath('logging.yaml').open() as f:
+    logging_yaml = AttrDict(yaml.load(f, yaml.FullLoader))
+
+DEFAULT_DATE_FORMAT = '%Y-%m-%d %H:%M:%S'
+
+defaultFormatter = logzero.LogFormatter(
+    fmt='%(asctime)s - %(name)s - %(levelname)s - %(message)s', datefmt=DEFAULT_DATE_FORMAT
+)
+
+logger = logzero.setup_logger(
```
mshriver, line 24: I ran into a lot of issues trying to use the logzero root logger for robottelo, and instead opted to create a specific logger.

pytest_plugins/logging_hooks.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def pytest_runtest_logstart(nodeid, location):
```
mshriver, line 64: The only issue I see with the logging behavior in this PR are these two hooks.

When running in xdist, the messages get written to the `robottelo_gw#.log` file - but they also get written to `robottelo.log`, resulting in robottelo.log just having all of the start/finish test lines with no other content when running in xdist.

The logger at this scope should be the xdist worker logger, but for some reason it also writes to the original file handler that was removed in the logging configuration hook/fixture.  I suspect its pytest capturing the output from these fixtures specifically?

I'm not too concerned about the behavior, but it's not as clean as it could be.

robottelo/config/facade.py
JacobCallahan: the entire facade :)

pytest_plugins/logging_hooks.py
JacobCallahan: i'm fine with it logging to the main file as well, could be better for people who want to see everything in one place.

mshriver: Well - it _only_ logs the messages from these hooks. _Everything_ else only goes to the xdist worker log file.

JacobCallahan (approved):

docs/conf.py
```diff
@@ -40,9 +40,9 @@ def setup(app):
 autoapi_dirs = ['../robottelo']
 source_suffix = '.rst'
 master_doc = 'index'
-exclude_patterns = ['_build', 'pytest/*']
+exclude_patterns = ['_build', 'pytest/*', 'logs/']
 nitpicky = False
-autodoc_default_options = {'members': None, 'undoc-members': None}
+autodoc_default_options = {'members': None, 'undoc-members': None, 'autoapi_keep_files': True}
```
mshriver, line 45: Gonna leave this in, because it's very useful when you're trying to debug docs failures on files that get deleted when autoapi fails.

docs/index.rst
```diff
@@ -220,114 +212,44 @@ $ pytest tests/foreman/cli/test_host.py -n 4
 
 For more information about Python's `pytest`_ module, read the documentation.
 
-Testing With Unittest
```
mshriver, line 223: Most of the changes in this file were to resolve docs build failures, but Unittest/Nosetest could have been removed with the final pytest PR, so removing them here.

docs/index.rst
```diff
@@ -220,114 +212,44 @@ $ pytest tests/foreman/cli/test_host.py -n 4
...
 
 It is possible to run UI tests within a docker container. To do this:
 
-* Install docker. It is provided by the ``docker`` package on Fedora and Red
-Hat. Be aware that the package may call ``docker-io`` on old OS releases.
-* Make sure that docker is up and running and the user that will run robottelo
-has permission to run docker commands. For more information check the docker
-installation guide https://docs.docker.com/engine/installation/.
+* Install docker. It is provided by the ``docker`` package on Fedora and Red Hat. Be aware that the package may call ``docker-io`` on old OS releases.
+* Make sure that docker running and your user has permission to run docker. For more information check https://docs.docker.com/engine/installation/.
```
swadeley, line 222:
```suggestion
* Make sure that docker is running and your user has permission to run docker. For more information, check https://docs.docker.com/engine/installation/.
```

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -51,7 +48,7 @@ def _get_tests(launch, **kwargs):
 def _get_test_collection(selectable_tests, items):
     """Returns the selected and deselected items"""
     # Select test item if its in failed tests else deselect
-    LOGGER.debug('Selecting/Deselecting tests based on latest launch test results..')
+    logger.debug('Selecting/Deselecting tests based on latest launch test results..')
```
swadeley, line 51:
```suggestion
    logger.debug('Selecting/Deselecting tests based on latest launch test results.')
```

robottelo/test.py
```diff
@@ -0,0 +1,323 @@
...
+        'exception' attribute. This allows you to inspect the exception after
+        the assertion::
+
+               with self.assertNotRaises(SomeException) as cm:
+                   do_something()
+               the_exception = cm.exception
+               self.assertEqual(the_exception.error_code, 1)
+
+        In addition, optional 'http_status_code' or 'cli_return_code' arg may
+        be passed. This allows to specify exact HTTP status code or CLI return
```
swadeley, line 251:
```suggestion
        be passed. This enables specifying exact HTTP status code or CLI return
```

robottelo/test.py
```diff
@@ -0,0 +1,323 @@
...
+class CLITestCase(TestCase):
+    """Test case for CLI tests."""
+
+    _default_interface = INTERFACE_CLI
+    _default_notraises_value_handler = CLINotRaisesValueHandler
+    _multiprocess_can_split_ = True
+
+    def assert_error_msg(self, raise_ctx, *contents):
+        """Checking error msg present on Raise Context Exception
+        Raise assertion error if any of contents are not present on error msg
```
swadeley, line 315:
```suggestion
        Raise assertion error if any of contents are not present in error msg
```

robottelo/test.py
```diff
@@ -0,0 +1,323 @@
...
+
+        :param raise_ctx: Raise Context
+        :param contents: contents which must be present on message
```
swadeley, line 318:
```suggestion
        :param contents: contents which must be present in message
```

robottelo/test.py
mshriver: Ha!! This entire module should not be here. This branch was created before the PRs went in removing it, and since I had changes to it the rebase resolved to keep the entire thing.

swadeley (approved):
ACK

PR #8518, opened 2021-04-15
Removed redundant for cycle (iterating already done in parametrization)

ppochop:
I removed the `for` cycle which did exactly what the parametrization does.
```
$ pytest tests/foreman/api/test_computeprofile.py::test_positive_delete
collected 7 items                                                              

tests/foreman/api/test_computeprofile.py .......                         [100%]
======================= 7 passed, 18 warnings in 31.44s ========================
```

### DIFF
--- a/tests/foreman/api/test_computeprofile.py
+++ b/tests/foreman/api/test_computeprofile.py
@@ -121,8 +121,7 @@ def test_positive_delete(new_name):
 
     :parametrized: yes
     """
-    for new_name in valid_data_list():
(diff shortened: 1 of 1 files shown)


omkarkhatavkar:

lhellebr (approved):
This indeed seems like fixing something with unnecessary quadratic complexity and furthermore, giving wrong results (since all the test instances actually ran with the same data). ACK, good catch.

JacobCallahan (approved):

PR #8184, opened 2020-11-19
cli test host rewritten to pytest

tstrych:
I added markers to those tests, to be able to run all test_host create. I did not use the class for that as I was told in past that I should remove the class. 
It make my life easier when I want to run the group of tests with same setup :) 
So I added them into  [pytest/markerspy](https://github.com/SatelliteQE/robottelo/blob/master/pytest_plugins/markers.py) to omit pytest warnings. 

I reused api fixtures and I build up the new ones based on them. 
Not sure if I should create new file only for these fixtures, I think it is fine this way.

Another think I was not sure about is `class HostSubscription`.
There are methods which were used in the setupclass methods and they are kind of complicated, so I though this way it can stay the same way as before so it will be good idea. 
Fixture `module_host_subscription` creates an object with all the data used in methods and tests later on. 
And setup is now in fixture `host_subscription_client`  and this client is set up in the beginning of the each test. 
I would be happy for your feedback about this and I am open to your ideas ;) 

I renamed the module_lce_search to [module_lce_library](https://github.com/SatelliteQE/robottelo/pull/8184/files#diff-28c9bb9ff38b04fba3cde391134e531da9e6eec5a991871745a28e2506af7189R475)

I changed the scope and the search query of [default_contentview](https://github.com/SatelliteQE/robottelo/pull/8184/files#diff-28c9bb9ff38b04fba3cde391134e531da9e6eec5a991871745a28e2506af7189L426) - I didn't find any other usage of it so I was not afraid of scope change.  And it didn't work so I change the search query. 

Tests which were using `make_proxy()` or better said fixture `function_proxy` were failing too. Dirty trick with `setenforce 0` helps here. 

Test results:
```
=========================== short test summary info ============================
FAILED test_host.py::test_positive_create_using_libvirt_without_mac - robotte...
FAILED test_host.py::test_hammer_host_info_output - AssertionError: assert '6...
FAILED test_host.py::test_negative_edit_parameter_by_non_admin_user - socket....
====== 3 failed, 41 passed, 6 skipped, 49 warnings in 3777.17s (1:02:57) =======
```

test_hammer_host_info_output
I changed [assert](https://github.com/SatelliteQE/robottelo/compare/master...tstrych:cli_test_host_unittest2pytest?expand=1#diff-f6232b85a7b80359d84a679b22645cdbeafbd09ffa6d4b08ecef98bb45b6cac2R1023) because they are sort by name and if the name starts with number it's before `admin` and it takes first index, and I run it again 
```
[...]

PR #8536, opened 2021-04-20
vm2br cli/test_host

tstrych:
cleanup with request object was added to `register_contenthost`.
Cleanup is preventing from possible organization mismatch error. This was discussed previously in automation chat. 
This would require other people to change their name definitions and method definitions when register_contenthost is called. 

Note: I am component owner

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -210,7 +210,7 @@ class ContentHost(Host):
         self,
         org='Default_Organization',
         activation_key=None,
-        lce=None,
+        lce='Library',
         consumerid=None,
         force=True,
         releasever=None,
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -210,7 +210,7 @@ class ContentHost(Host):
         self,
         org='Default_Organization',
         activation_key=None,
-        lce='Library',
+        lce=None,
         consumerid=None,
         force=True,
         releasever=None,
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -1999,6 +1999,7 @@ def test_negative_without_attach(request, module_host_subscription, host_subscri
     host = Host.info({'name': module_host_subscription.client.hostname})
     module_host_subscription.client.register_contenthost(
         module_host_subscription.org.name,
+        lce=None,  # required, to jump into right branch in register_contenthost method
         consumerid=host['subscription-information']['uuid'],
         force=False,
     )
@@ -2031,8 +2032,6 @@ def test_negative_without_attach_with_lce(module_host_subscription, host_subscri
         environment=lce,
         organization=org,
     ).create()
-    # entities.ContentView().search(
-    #     query={'search': f'name={DEFAULT_CV}', 'organization_id': f'{module_org.id}'}
     setup_org_for_a_rh_repo(
         {
             'product': PRDS['rhel'],
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -18,6 +18,7 @@
 """
 import time
 from datetime import datetime
+from datetime import timedelta
 from random import choice
 
 import pytest
@@ -1515,18 +1516,17 @@ def katello_host_tools_repos():
 @pytest.mark.skip_if_not_set('clients')
 @pytest.fixture(scope="function")
 def katello_host_tools_client(katello_host_tools_repos, rhel7_contenthost):
-    client = rhel7_contenthost
-    client.install_katello_ca()
+    rhel7_contenthost.install_katello_ca()
     # Register content host and install katello-host-tools
-    client.register_contenthost(
+    rhel7_contenthost.register_contenthost(
         katello_host_tools_repos['org'].label,
         katello_host_tools_repos['ak'].name,
     )
-    assert client.subscribed
-    host_info = Host.info({'name': client.hostname})
-    client.enable_repo(REPOS['rhst7']['id'])
-    client.install_katello_host_tools()
-    yield {'client': client, 'host_info': host_info}
+    assert rhel7_contenthost.subscribed
+    host_info = Host.info({'name': rhel7_contenthost.hostname})
+    rhel7_contenthost.enable_repo(REPOS['rhst7']['id'])
+    rhel7_contenthost.install_katello_host_tools()
+    yield {'client': rhel7_contenthost, 'host_info': host_info}
 
 
 @pytest.mark.katello_host_tools
@@ -1849,6 +1849,10 @@ class HostSubscription:
                 in Host.info({'name': self.client.hostname})['subscription-information'][
                     'registered-at'
                 ]
+                or (datetime.utcnow() - timedelta(days=1)).strftime("%Y-%m-%d")
+                in Host.info({'name': self.client.hostname})['subscription-information'][
+                    'registered-at'
+                ]
             ):
                 Host.subscription_unregister({'host': self.client.hostname})
 
@@ -1941,7 +1945,7 @@ def test_positive_attach(request, module_host_subscription, host_subscription_cl
     try:
         module_host_subscription.client.install_katello_agent()
     except ContentHostError:
-        pytest.fail("ContentHostError raised unexpectedly!")
+        pytest.fail('ContentHostError raised unexpectedly!')
 
 
 @pytest.mark.host_subscription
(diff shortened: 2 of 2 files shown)


tstrych:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py

tstrych:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py

JacobCallahan:
I'm not a fan of the current request model. Let's discuss this more tomorrow. We may have a better option available soon.

tstrych:
```
FAILED tests/foreman/cli/test_host.py::test_negative_without_attach - Asserti...
= 1 failed, 46 passed, 1 skipped, 7 deselected, 259 warnings in 7939.57s (2:12:19) =
```
2hours :dagger:  compared to 1h in #8184 

I will look at failed test

tstrych:
Test results of `tests/foreman/cli/test_host.py`
```
ERROR tests/foreman/cli/test_host.py::test_positive_create_and_update_with_content_source
ERROR tests/foreman/cli/test_host.py::test_positive_update_parameters_by_name
ERROR tests/foreman/cli/test_host.py::test_negative_update_arch - requests.ex...
ERROR tests/foreman/cli/test_host.py::test_negative_update_os - requests.exce...
= 43 passed, 1 skipped, 7 deselected, 216 warnings, 4 errors in 7864.67s (2:11:04) =
```

Those failures 3 of them 502 bad gateway  error on basic fixture, and `test_positive_create_and_update_with_content_source ` requires additional setup or we need to disable selinux.

Rerun results:
```
pytest -s -vv tests/foreman/cli/test_host.py::test_positive_create_and_update_with_content_source tests/foreman/cli/test_host.py::test_positive_update_parameters_by_name tests/foreman/cli/test_host.py::test_negative_update_arch tests/foreman/cli/test_host.py::test_negative_update_os

==================== 4 passed, 49 warnings in 265.61s (0:04:25) =====================
 ```

**Review by JacobCallahan — CHANGES_REQUESTED** [review]
Still holding this until the more optimal pattern is available

**Comment by tstrych** [comment]
Yes, this PR will be updated once the new pattern will be available.

**Comment by JacobCallahan** [comment]
@tstrych after the latest broker update, you can now define a `teardown` method on the `ContentHost` class that unregisters your hosts. This will be executed automatically in fixtures and context managers. The `teardown` method must take no arguments.

**Comment by tstrych** [comment]
maybe some rebase will be required, let's see

pasting results after requests was replaced by ContentHost tierdown method.
Screenshot from 2021-05-20 09-26-35
And rerun for: `test_negative_register_twice` and `test_positive_unregister`
Screenshot from 2021-05-20 09-31-10
Screenshot from 2021-05-20 09-30-32

rerun for test_positive_create_and_update_with_content_source is not required, there is no change in the code, and the problem is with proxy settings, it will pass in automation as it always does :+1: 

commits were squashed

**Review by mshriver — APPROVED** [review]

File: tests/foreman/cli/test_host.py [hunk 1]
```diff
@@ -1484,42 +1474,48 @@ def test_positive_provision_baremetal_with_uefi_secureboot():
...
     }
 
 
 @pytest.mark.skip_if_not_set('clients')
 @pytest.fixture(scope="function")
-def katello_host_tools_client(katello_host_tools_repos):
-    client = VirtualMachine(distro=DISTRO_RHEL7)
-    client.create()
+def katello_host_tools_client(katello_host_tools_repos, rhel7_contenthost):
+    client = rhel7_contenthost
```
JacobCallahan, line 1518: why alias this?

tests/foreman/cli/test_host.py
```diff
@@ -1847,8 +1838,20 @@ def _make_activation_key(self, add_subscription=False):
             )
         return activation_key
 
-    def _host_subscription_register(self):
+    def _host_subscription_register(self, request):
```
JacobCallahan, line 1841: Couldn't the new teardown method do this and not require tests to add a `request` argument?

tests/foreman/cli/test_host.py
```diff
@@ -1934,12 +1936,12 @@ def test_positive_attach(module_host_subscription, host_subscription_client):
...
     result = module_host_subscription._client_enable_repo()
-    assert result.return_code == 0
+    assert result.status == 0
     # ensure that katello agent can be installed
     try:
         module_host_subscription.client.install_katello_agent()
-    except VirtualMachineError:
-        pytest.fail("VirtualMachineError raised unexpectedly!")
+    except ContentHostError:
+        pytest.fail("ContentHostError raised unexpectedly!")
```
JacobCallahan, line 1944: Please switch these to single-quote strings.

tests/foreman/cli/test_host.py
```diff
@@ -1967,17 +1969,17 @@ def test_positive_attach_with_lce(module_host_subscription, host_subscription_cl
...
```
JacobCallahan, line 1977: Please switch these to single-quote strings.

tests/foreman/cli/test_host.py
```diff
@@ -2172,17 +2145,17 @@ def test_positive_auto_attach(module_host_subscription, host_subscription_client
...
```
JacobCallahan, line 2158: Please switch these to single-quote strings.

JacobCallahan (changes requested):

robottelo/hosts.py
```diff
@@ -198,7 +205,7 @@ def register_contenthost(
         self,
         org='Default_Organization',
         activation_key=None,
-        lce='Library',
+        lce=None,
```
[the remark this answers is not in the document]
tstrych: I didn't get why you have changed it when you had copied methods from `robottelo/vm.py` to `robottelo/hosts.py`. 
https://github.com/SatelliteQE/robottelo/blob/master/robottelo/vm.py#L489
Please explain if it has a reason. 

But None was there on purpose. If Default Library is used by default you have dangling else in the code. (line 248-260)
You can never access register by consumer id, and error will never be thrown. 

And once the Library is there by default test `test_negative_without_attach` will fail https://github.com/SatelliteQE/robottelo/blob/d142bbcc9b44c5518800873a384517ca95389e22/tests/foreman/cli/test_host.py#L1980

tests/foreman/cli/test_host.py
tstrych: I was just lazy, and this was easier, but agree I should rename it.

tests/foreman/cli/test_host.py
tstrych: sure, as always I forget. thanks

tests/foreman/cli/test_host.py
tstrych: This do not work when the host is registered only from hammer. The record will stay in satellite. If new machine with the same name arrives there will be unwanted behaviour, example: organization mismatch so it will not be registered and test will fail. 

```
Host.subscription_register ->  hammer host subscription register <params>
```

So it has to be cleaned from host and from satellite. Maybe we can introduce smth similar we did for host. 
I am open to ideas.

tstrych: I have to say I didn't test it. If you want me to I will.

robottelo/hosts.py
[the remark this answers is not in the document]
JacobCallahan: Defaulting to Library is a sane choice for the vast majority of tests that will be using this function. For tests that don't want to use Library, they can pass in an alternate value or `None`. Tests relying on the `None` behavior should explicitly pass in `lce=None`.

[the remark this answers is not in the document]
tstrych: understand,
agree, most used will be the library, so it can be default. 
though seems weird to pass lifecycle-environment as `None`.

tstrych:
I have double-checked the only test using `customerid` is this one. 
And it passed `================= 2 passed, 53 deselected in 994.97s (0:16:34) =================` 
(there are two tests because I used `pytest -k test_negative_without_attach`, which fits also for `test_negative_without_attach_with_lce`

robottelo/hosts.py
```diff
@@ -71,6 +71,13 @@ def ip_addr(self):
         ipv4, ipv6 = self.execute('hostname -I').stdout.split()
         return ipv4
 
+    def setup(self):
+        self.remove_katello_ca()
+        self.execute('subscription-manager clean')
+
+    def teardown(self):
+        self.unregister()
+
```
JacobCallahan, line 80: rebasing should bring in the current implementations of these
https://github.com/SatelliteQE/robottelo/blob/master/robottelo/hosts.py#L78-L88

mshriver (approved):

PR #8537, opened 2021-04-21
Handle server.hostnames being empty in xdist

mshriver:
Ultimately the validator should be enforcing that its a list, but because the validation isn't strictly enforced at this point its easy for this field to be empty.

### DIFF
--- a/pytest_fixtures/xdist.py
+++ b/pytest_fixtures/xdist.py
@@ -27,6 +27,8 @@ def align_to_satellite(worker_id, satellite_factory):
     if settings.server.inventory_filter:
         settings.server.hostnames  # need to prime the cache
         hosts = VMBroker().from_inventory(filter=settings.server.inventory_filter)
+        # update the cache_proxy for server.hostnames in case its empty
+        cache_proxy['server.hostnames'] = cache_proxy['server.hostnames'] or []
         cache_proxy['server.hostnames'].extend([host.hostname for host in hosts])
 
     # attempt to align a worker to a satellite


mshriver:
trigger: test-robottelo
pytest: --importance Critical -m tier1 tests/foreman/cli

JacobCallahan (approved):

mshriver:
trigger: test-robottelo
pytest: --importance Critical -m tier1 tests/foreman/cli

latran (approved):
I don't think I've seen multiple operators like that before.  

so `cache_proxy['server.hostnames'] = cache_proxy['server.hostnames'] or []` acts like `cache_proxy['server.hostnames'] = (cache_proxy['server.hostnames'] or [])` where it can be set to either or?

I'm ACKING this, but also wanna get an understanding.

mshriver:
@latran thanks for the review and asking questions!

Its actually a boolean behavior in python, with logical or evaluating the first part of the statement as a boolean.

Its behaving like a ternary statement:

```
var = var if bool(var) else 'default'
```

I'm simplifying this behavior by just using `or`.

This means that when the value in `cache_proxy['server.hostnames']` is 'falsy', the 2nd value after `or` is used.

By 'falsy' I mean `bool() == False` for any given value.

The primary use case for this type of syntax/behavior would be handling a `None` value, where the yaml had the key set but nothing in it.

But!  An empty list is also 'falsy', so when `cache_proxy['server.hostnames']` is defined in the yaml as an empty list, this line will replace it with another empty list.

I don't care to split that nuance here, and explicitly only handle the `None` case by using a ternary like the following roughly equivalent statement.

```
cache_proxy['server.hostnames'] = cache_proxy['server.hostnames'] if cache_proxy['server.hostnames'] is not None else []
```

latran:
Given that it's a ternary statement makes a lot more sense in reading that now.  Thank you!

PR #8539, opened 2021-04-22
Delete NotABug Test Scenario

latran:
This test scenario is no longer valid as the bz is filed as `NotABug`.


JacobCallahan (changes requested):
ACK great catch!

pondrejk (approved):

JacobCallahan (approved):

PR #8543, opened 2021-04-23
Move xdist worker alignment to a hook

mshriver:
This needs to happen earlier in the session than as a fixture

Change up the satellite_factory fixture so that its more appropriately
named and the core factory function is separated.


pytest_plugins/xdist.py
```diff
@@ -4,24 +4,27 @@
...
+    cache_proxy = robottelo.config.settings_proxy._cache  # noqa
     # clear any hostname that may have been previously set
     cache_proxy['server.hostname'] = on_demand_sat = None
 
-    if worker_id in ['master', 'local']:
-        worker_pos = 0
-    else:
-        worker_pos = int(worker_id.replace('gw', ''))
+    worker_id = get_xdist_worker_id(session)
+    worker_pos = 0 if worker_id == 'master' else int(worker_id.replace('gw', ''))
```
JacobCallahan, line 27:
```suggestion
    worker_pos = 0 if worker_id in ['master', 'local'] else int(worker_id.replace('gw', ''))
```

pytest_plugins/xdist.py
```diff
@@ -33,22 +36,27 @@ def align_to_satellite(worker_id, satellite_factory):
...
     logger.info(f"xdist worker {worker_id} " f"was assigned hostname {settings.server.hostname}")
     settings.configure_airgun()
     settings.configure_nailgun()
-    yield
+
+
+@pytest.hookimpl(trylast=True)
+def pytest_sessionfinish(session, exitstatus):
+    # TODO store the deployed sat in the session so we can check it in here.
     if on_demand_sat and settings.server.auto_checkin:
```
JacobCallahan, line 61: how is on_demand_sat in scope here? some pytest magic?

JacobCallahan (changes requested):

pytest_plugins/xdist.py
mshriver: PR in draft state, see TODO above.

pytest_plugins/xdist.py
mshriver: Why check for local? `xdist.get_xdist_worker_id` doesn't return `'local'` AFAIK.

https://github.com/pytest-dev/pytest-xdist/blob/master/src/xdist/plugin.py#L248

JacobCallahan: unfortunately, there are cases where local does pop up. this is something I encountered while testing this fixture.
https://pytest-services.readthedocs.io/en/2.2.1/#infrastructure-fixtures

pytest_plugins/xdist.py
JacobCallahan: hahaha ok, just making sure I wasn't missing out on some weird pytest scope injection or something.

pytest_fixtures/broker.py
```diff
@@ -9,36 +9,34 @@
...
+    )
+    timeout = (1200 + delay) * retry_limit
+    sat = wait_for(
+        vmb.checkout, func_kwargs=broker_args, timeout=timeout, delay=delay, fail_condition=[]
+    )
+    return sat.out
+
+
+def capsule_factory(retry_limit=3, delay=300, **broker_args):
+    vmb = VMBroker(host_classes={'host': Satellite}, workflow='deploy-sat-capsule', **broker_args)
```
JacobCallahan, line 24: This was addressed in another open PR, but pointing out my oversight here as well
```suggestion
    vmb = VMBroker(host_classes={'host': Capsule}, workflow='deploy-sat-capsule', **broker_args)
```

pytest_plugins/xdist.py
mshriver: That's only when using the `worker_id` fixture provided by pytest-services. Using the methods provided by xdist itself, this will only be 'master'.

pytest_fixtures/broker.py
mshriver: Can you fix this in a specific PR, along with `deploy-sat-capsule` being read from config instead of hardcoded?

pytest_plugins/xdist.py
JacobCallahan: I'm curious as to if that is true or not since the fixture version has one line of code `return get_xdist_worker_id(request)`

pytest_fixtures/broker.py
JacobCallahan: This is now being resolved by #8514

JacobCallahan:
@mshriver is this something you want to keep in part now that the target_sat implementation is in?

mshriver:
Nope!

PR #8558, opened 2021-04-30
`test` prepended to azure hosts for auto cleanup

jyejare:
Prepending `test` word to the host names provisioned in Azure so that they will be removed using cloud cleanup script!

### DIFF
--- a/tests/foreman/api/test_computeresource_azurerm.py
+++ b/tests/foreman/api/test_computeresource_azurerm.py
@@ -164,7 +164,7 @@ class TestAzureRMHostProvisioningTestCase:
         request.cls.premium_os_disk = AZURERM_PREMIUM_OS_Disk
         request.cls.platform = AZURERM_PLATFORM_DEFAULT
         request.cls.vm_size = AZURERM_VM_SIZE_DEFAULT
-        request.cls.hostname = gen_string('alpha')
+        request.cls.hostname = f'test_{gen_string("alpha")}'
         request.cls.fullhostname = f'{self.hostname}.{module_domain.name}'.lower()
 
         request.cls.compute_attrs = {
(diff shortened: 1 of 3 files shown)


tests/foreman/api/test_computeresource_azurerm.py
```diff
@@ -164,7 +164,7 @@ def class_setup(self, request, module_domain, module_azurerm_cr, module_azurerm_
         request.cls.premium_os_disk = AZURERM_PREMIUM_OS_Disk
         request.cls.platform = AZURERM_PLATFORM_DEFAULT
         request.cls.vm_size = AZURERM_VM_SIZE_DEFAULT
-        request.cls.hostname = gen_string('alpha')
+        request.cls.hostname = f"test_{gen_string('alpha')}"
```
JacobCallahan, line 167: Please switch these to single-quote strings here and through the other changed lines

JacobCallahan (changes requested):

tests/foreman/api/test_computeresource_azurerm.py
jyejare:
```suggestion
        request.cls.hostname = f'test_{gen_string('alpha')}'
```

This wont work as there is already single quote inside !

jyejare: There are 4 quotes on one line !

JacobCallahan: you can flip them, so it looks like 
```suggestion
        request.cls.hostname = f'test_{gen_string("alpha")}'
```

jyejare: Not sure whats the advantage it has but as per python standards, we should wrap single quotes with double and not vice-versa.

Still made the changes are are ready for your ACK !

jyejare:
The checks failed are not related to this PR.

JacobCallahan (approved):
ACK, thanks!

mshriver (approved):

PR #8559, opened 2021-04-30
VM->Broker conversion for api/test_errata.py

JacobCallahan:
Converted api/test_errata.py
Also included some new fixtures and enhancements to the ContentHost
class
RHEL6 and RHEL8 tests can't be run due to needed template enhancements.

All non-rhel6/8 tests pass locally at the same percentage as before.

### DIFF
--- a/pytest_fixtures/broker.py
+++ b/pytest_fixtures/broker.py
@@ -78,6 +78,13 @@ def rhel8_contenthost():
         yield host
 
 
+@pytest.fixture
+def rhel6_contenthost():
+    """A function-level fixture that provides a content host object based on the rhel6 nick"""
+    with VMBroker(nick='rhel6', host_classes={'host': ContentHost}) as host:
+        yield host
+
+
 @pytest.fixture(scope="module")
 def rhel77_host_module():
     """A module-level fixture that provides a RHEL7.7 host object"""
--- a/pytest_fixtures/xdist.py
+++ b/pytest_fixtures/xdist.py
@@ -3,7 +3,7 @@ import logging
 import random
 
 import pytest
-from broker.broker import VMBroker
+from broker import VMBroker
 
 import robottelo
 from robottelo.config import settings
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -5,6 +5,7 @@ from urllib.parse import urljoin
 from urllib.parse import urlunsplit
 
 from broker.hosts import Host
+from nailgun import entities
 from wait_for import TimedOutError
 from wait_for import wait_for
 
@@ -52,6 +53,23 @@ class ContentHostError(Exception):
 
 class ContentHost(Host):
     run = Host.execute
+    subscribed = False
+
+    @property
+    def nailgun_host(self):
+        """If this host is subscribed, provide access ot its nailgun object"""
+        if self.subscribed:
+            return entities.Host().search(query={'search': self.hostname})[0]
+
+    @property
+    def subscribed(self):
+        """Boolean representation of a content host's subscription status"""
+        return 'Status: Unknown' not in self.execute('subscription-manager status').stdout
+
+    @property
+    def ip_addr(self):
+        ipv4, ipv6 = self.execute('hostname -I').stdout.split()
+        return ipv4
 
     def download_install_rpm(self, repo_url, package_name):
         """Downloads and installs custom rpm on the broker virtual machine.
@@ -235,10 +253,7 @@ class ContentHost(Host):
             cmd += f' --release {releasever}'
         if force:
             cmd += ' --force'
-        result = self.execute(cmd)
-        if 'The system has been registered with ID' in ''.join(result.stdout):
-            self.subscribed = True
-        return result
+        return self.execute(cmd)
 
     def remove_katello_ca(self, capsule=None):
         """Removes katello-ca rpm from the broker virtual machine.
@@ -413,11 +428,6 @@ class ContentHost(Host):
             raise ContentHostError('No distro package available to retrieve release version')
         return self.execute(f"echo '{rh_product_os_releasever}' > /etc/yum/vars/releasever")
 
-    @property
-    def ip_addr(self):
-        ipv4, ipv6 = self.execute("hostname -I").stdout.split()
-        return ipv4
-
 
 class Capsule(ContentHost):
     def restart_services(self):
@@ -456,7 +466,6 @@ class Satellite(Capsule):
         """Import all nailgun entities and wrap them under self.api"""
         from nailgun.config import ServerConfig
         from nailgun.entity_mixins import Entity
-        from nailgun import entities
 
         def inject_config(cls, server_config):
             """inject a nailgun server config into the init of nailgun entity classes"""
--- a/tests/foreman/api/test_errata.py
+++ b/tests/foreman/api/test_errata.py
@@ -20,49 +20,29 @@
 from time import sleep
 
 import pytest
+from broker import VMBroker
 from nailgun import entities
 
+from robottelo import constants
 from robottelo.api.utils import enable_rhrepo_and_fetchid
 from robottelo.api.utils import promote
 from robottelo.cli.factory import setup_org_for_a_custom_repo
 from robottelo.cli.factory import setup_org_for_a_rh_repo
 from robottelo.config import settings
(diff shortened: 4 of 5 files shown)


pytest_fixtures/xdist.py
```diff
@@ -3,7 +3,7 @@
 import random
 
 import pytest
-from broker.broker import VMBroker
+from broker import VMBroker
```
jameerpathan111, line 7: Is this an intended change?

tests/foreman/api/test_errata.py
```diff
@@ -199,25 +179,23 @@ def test_positive_bulk_install_package(module_org, activation_key, custom_repo,
...
-            host_ids=[host_id],
-            package_name=FAKE_1_CUSTOM_PACKAGE_NAME,
-            via_ssh=False,
-            rpm_package_name=FAKE_2_CUSTOM_PACKAGE,
-        )
+    rhel7_contenthost.install_katello_ca()
+    rhel7_contenthost.register_contenthost(module_org.label, activation_key.name)
+    assert rhel7_contenthost.subscribed
+    rhel7_contenthost.enable_repo(constants.REPOS['rhst7']['id'])
+    rhel7_contenthost.install_katello_agent()
```
Gauravtalreja1, line 186: Can you implement a module-level fixture just to optimize these repeated steps? rather than using plain `rhel7_contenthost`, also for rhel6/8 clients if necessary.

tests/foreman/api/test_errata.py
```diff
@@ -199,25 +179,23 @@ def test_positive_bulk_install_package(module_org, activation_key, custom_repo,
...
+    host_id = entities.Host().search(query={'search': f'name={rhel7_contenthost.hostname}'})[0].id
```
jameerpathan111, line 187: @JacobCallahan We can use `nailgun_host` here?

robottelo/hosts.py
```diff
@@ -52,6 +53,23 @@ class ContentHostError(Exception):
 
 class ContentHost(Host):
     run = Host.execute
+    subscribed = False
+
+    @cached_property
+    def nailgun_host(self):
+        """If this host is subscribed, provide access the it's nailgun object"""
```
swadeley, line 60:
```suggestion
        """If this host is subscribed, provide access to it's nailgun object"""
```

swadeley (approved):
ACK

tests/foreman/api/test_errata.py
swadeley: I see `from nailgun import entities`, so I think `Host()` is from Nailgun.  I tried to grep for `nailgun_host` but could not find it.

robottelo/hosts.py
```diff
@@ -52,6 +53,23 @@ class ContentHostError(Exception):
...
+        if self.subscribed:
+            return entities.Host().search(query={'search': self.hostname})[0]
+
+    @property
+    def subscribed(self):
+        """Boolean representation of a content host's subscription status"""
+        return 'Status: Unknown' not in self.execute('subscription-manager status').stdout
```
omkarkhatavkar, line 67: Any specific reason for adding indirect approach that host is registered, you can also able to verify the status as `Status: Current` or `subscription-manager list | egrep -i 'status: '`

pytest_fixtures/broker.py
```diff
@@ -67,10 +67,19 @@ def rhel7_contenthost_module():
 @pytest.fixture
 def rhel8_contenthost():
     """A fixture that provides a content host object based on the rhel8 nick"""
+    pytest.skip('Skipped until rhel8 templates have wget added')
```
tpapaioa, line 70: Why is `wget` necessary to use the fixture? You can provision and register a system without it.

pytest_fixtures/broker.py
```diff
@@ -67,10 +67,19 @@ def rhel7_contenthost_module():
...
     with VMBroker(nick='rhel8', host_classes={'host': ContentHost}) as host:
         yield host
 
 
+@pytest.fixture
+def rhel6_contenthost():
+    """A function-level fixture that provides a content host object based on the rhel6 nick"""
+    pytest.skip('Waiting on RHEL 6 SatLab templates')
+    with VMBroker(nick='rhel6', host_classes={'host': ContentHost}) as host:
```
tpapaioa, line 79: Use `DISTRO_RHEL6`, `DISTRO_RHEL7`, and `DISTRO_RHEL8` from `robottelo.constants`.

robottelo/hosts.py
tpapaioa: There are multiple statuses possible, 

- fully subscribed: `Overall Status: Current` (return code 0)
- partially or fully unsubscribed: `Overall Status: Invalid` (return code 1)
- unregistered: `Overall Status: Unknown` (return code 1)

tests/foreman/api/test_errata.py
```diff
@@ -20,49 +20,28 @@
...
-from robottelo.constants import REAL_0_ERRATA_ID
-from robottelo.constants import REAL_0_RH_PACKAGE
-from robottelo.constants import REAL_1_ERRATA_ID
-from robottelo.constants import REAL_2_ERRATA_ID
-from robottelo.constants import REPOS
-from robottelo.constants import REPOSET
-from robottelo.constants.repos import CUSTOM_SWID_TAG_REPO
-from robottelo.constants.repos import FAKE_3_YUM_REPO
-from robottelo.constants.repos import FAKE_9_YUM_REPO
+from robottelo.constants import repos
```
tpapaioa, line 32: What's the reason for this change?

tpapaioa (changes requested):

pytest_fixtures/broker.py
swadeley: @JacobCallahan is this because of `wget` in `robottelo/hosts.py `? If so, see:
https://github.com/SatelliteQE/robottelo/pull/8556#discussion_r625625501

tests/foreman/api/test_errata.py
JacobCallahan: I'd prefer the tests to be more explicit in their behavior since it is more clear exactly what's going on.

tests/foreman/api/test_errata.py
JacobCallahan: sure can! i'll update that in these places since that was a later change. good catch!

robottelo/hosts.py
JacobCallahan: ha! you're absolutely correct

robottelo/hosts.py
JacobCallahan: Tasos nailed it. The second status is very common since we also have tests where we only have custom subscriptions.

pytest_fixtures/broker.py
JacobCallahan: @swadeley iirc there may be other places where it is needed, but after that change is introduced I can rebase and test again.

tests/foreman/api/test_errata.py
JacobCallahan: IMO it looks better to have one import instead of several from the same location. it is also more explicit where those constants come from when used in code (more clear when something is a local/vs imported constant).

pytest_fixtures/broker.py
JacobCallahan: yep. just confirmed that wget is used by helper methods elsewhere. i'm fine with us waiting on a rhel8 template that meets our requirements.

JacobCallahan:
@jameerpathan111 @Gauravtalreja1 @tpapaioa @omkarkhatavkar I've updated the pr to address your comments

jameerpathan111 (approved):
Ack

robottelo/hosts.py
```diff
@@ -52,6 +53,23 @@ class ContentHostError(Exception):
 
 class ContentHost(Host):
     run = Host.execute
+    subscribed = False
+
+    @cached_property
+    def nailgun_host(self):
+        """If this host is subscribed, provide access ot its nailgun object"""
+        if self.subscribed:
```
mshriver, line 61: Since this depends on whether or not its subscribed, I wouldn't cache the property.

If its referenced before the host is subscribed, then it'll get cached to `None` and its on the caller to delete it before calling it again - making the caller have to check `host.subscribed` anyway.

mshriver (approved):

PR #8566, opened 2021-05-04
Fix yaml templates

mshriver:


omkarkhatavkar (approved):
ACK, looks like we found some more :-)

conf/repos.yaml.template
```diff
@@ -7,29 +7,31 @@ REPOS:
...
-  # RHSCL_REPO: replace_with_rhscl_repo
+  RHSCL_REPO: replace-with-rhscl-repo
   # Ansible 2.8 Repo
-  # ANSIBLE_REPO: replace_with_ansible_repo
+  ANSIBLE_REPO: replace-with-ansible-repo
   # Added swid_tools_repo for installing swid-tools and dnf-plugin-swidtags packages
   # which are essentially required for generating swid tags in RHEL8 content host.
-  # SWID_TOOLS_REPO: replace_with_swid_tools_repo
+  SWID_TOOLS_REPO: replace-with-swid-tools-repo
+  FAKE_REPO_ZOO3: replace-with-zoo-fedorapeople-repo
```
devendra104, line 37: parameter missing
* PUPPET4_REPO

conf/upgrade.yaml.template
```diff
@@ -21,3 +21,4 @@ UPGRADE:
   DOCKER_VM:
   # The upgrade VLAN vm_domain
   VM_DOMAIN:
+  DISTRIBUTION:
```
devendra104, line 24: for upgrade config need to sync with Git Upgrade MR.

JacobCallahan:
@devendra104 are your comments blocking?

conf/repos.yaml.template
mshriver: ack - this one at least isn't hit during test collection.

devendra104:
> @devendra104 are your comments blocking?

no, they are not blocking.

conf/upgrade.yaml.template
devendra104: Since the changes are not blocking so we can include them in another PR.

conf/repos.yaml.template
devendra104: yes, no impact, will cover it in another PR.

PR #8572, opened 2021-05-04
Refactor skip_if_not_set decorator to a pytest mark

mshriver:
Call pytest.skip in test setup with a skip_if mark specific to our use case.

Original implementation was great! But with consolidation to pytest, there's no need for the custom decorator, and we can register this as a mark.

I think its worth maintaining the simplicity of passing args of settings section names, over moving to using the built-in skip method in conjunction with the `setting_is_set` method like `pytest.skip_if(setting_is_set('blah'), reason='blah')`. Using the built-in method would require us to repeat the same reason over and over, and would require importing `setting_is_set`.

Ideally I'm including unit test rewrite, leveraging the work @rplevka did in the reporting unit test to wrap calling pytest.main to validate hooks.

Want to get some buy-in on the new pattern before writing those unit tests though.


pytest_plugins/settings_skip.py
```diff
@@ -0,0 +1,39 @@
...
+        "markers",
+        'skip_if_not_set: List settings sections that must be set for the test to run. '
+        'If settings are missing, the test is skipped in setup.',
+    )
+
+
+def pytest_runtest_setup(item):
+    """Skip in setup if settings mark isn't met
+
+    settings validate method is used, so required fields are checked
```
JacobCallahan, line 19: We'll likely have to make every setting required for this to behave as expected. Otherwise non-required validation won't fail.

tests/foreman/api/test_activationkey.py
```diff
@@ -439,7 +438,6 @@ def test_positive_remove_user():
 
 @pytest.mark.upgrade
 @pytest.mark.run_in_one_thread
-@skip_if_not_set('fake_manifest')
```
JacobCallahan, line 442: Where is this check happening now?

tests/foreman/api/test_user.py
```diff
@@ -631,10 +630,10 @@ def test_positive_ssh_key_in_host_enc(self):
 
 
 @pytest.mark.run_in_one_thread
+@pytest.mark.skip_if_not_set('ldap')
 class TestActiveDirectoryUser:
     """Implements the LDAP auth User Tests with Active Directory"""
 
-    @skip_if_not_set('ldap')
```
JacobCallahan, line 637: Context might not be correct in my view, but I don't see a mark replacement for this and others in this file

tests/robottelo/conftest.py
```diff
@@ -0,0 +1,45 @@
+import glob
+import os
+from pathlib import Path
+from tempfile import NamedTemporaryFile
+
+import pytest
+from fauxfactory import gen_string
+
+
+@pytest.fixture(scope='function')
```
JacobCallahan, line 10: You can drop the explicit scope of you want

pytest_plugins/settings_skip.py
mshriver: These are changes that will have to come with the removal of the legacy settings.  I'm retaining the behavior of the current plugin, and this is where lazy validation of dynaconf sections could take place.

robottelo/config/__init__.py
```diff
@@ -51,3 +51,15 @@
...
+
+
+def setting_is_set(option):
+    """Return either ``True`` or ``False`` if a Robottelo section setting is
+    set or not respectively.
+    """
+    if not settings.configured:
+        settings.configure()
+    # Example: `settings.clients`
+    if getattr(settings, option).validate():
```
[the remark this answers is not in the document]
mshriver: As well as it does in its current implementation in `master` branch. Its only intended to handle the first level attribute of settings, the skip behavior was written to validate the 'block' of settings, not any particular one field.

tests/foreman/api/test_activationkey.py
mshriver: It doesn't look like the test actually uses fake_manifest, but I can add this back in to make sure we're keeping it consistent.

tests/foreman/api/test_user.py
mshriver: Its marked on the class now.

tests/robottelo/conftest.py
mshriver: But I put it there just for you! And because the other fixture already had it set explicitly.

swadeley: Consistency is a wonderful thing

pytest_plugins/settings_skip.py
JacobCallahan: Sounds good! Dynamic validation should allow us to make everything required, since we're only validating small sections at once.

tests/foreman/api/test_activationkey.py
```diff
@@ -491,7 +490,6 @@ def test_positive_fetch_product_content(module_org):
 
 
 @pytest.mark.upgrade
-@skip_if_not_set('fake_manifest')
```
JacobCallahan, line 494: for consistency, do you want to include this as well?

mshriver: No, because its stubbed, so there's no functional need for this setting to be set or not.

JacobCallahan: fair enough!

JacobCallahan (approved):

docs/features/decorators.rst
```diff
@@ -13,27 +13,6 @@ Robottelo decorators are located under
 control if a test must be skipped or executed accordingly with specific
 configurations.
 
-stubbed
```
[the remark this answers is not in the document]
mshriver: Its no longer a decorator, its provided by pytest marks, which are self documenting in `pytest --help`

JacobCallahan:
I'm satisfied with the answers to the remaining questions. With that, the promised ACK is counted and it's time to merge!

PR #8583, opened 2021-05-06
Mark as closed loop.

lhellebr:
BZ 1777992 is being tested by this
since the custom instance type is default

### DIFF
--- a/tests/foreman/longrun/test_provisioning_computeresource.py
+++ b/tests/foreman/longrun/test_provisioning_computeresource.py
@@ -155,6 +155,10 @@ def test_positive_provision_rhev_with_host_group(rhev, provisioning, tear_down):
 
     :expectedresults: The host should be provisioned with host group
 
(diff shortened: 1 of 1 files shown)


JacobCallahan (approved):

mshriver (approved):

tests/foreman/longrun/test_provisioning_computeresource.py
```diff
@@ -155,6 +155,8 @@ def test_positive_provision_rhev_with_host_group(rhev, provisioning, tear_down):
 
     :expectedresults: The host should be provisioned with host group
 
+    :BZ: 1777992
```
mshriver, line 158: This BZ also has a customer case associated with it, so you can set `:customer_scenario:` token as well

lhellebr: Done, thanks

PR #8608, opened 2021-05-13
ansible rex closed loop

pondrejk:


pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k test_positive_run_concurrent_jobs

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -583,6 +584,58 @@ def test_positive_run_reccuring_job(self, fixture_vmsetup, module_org):
...
+        :CaseLevel: System
+
+        :BZ: 1817320
+
+        :parametrized: yes
+        """
+        param_name = 'foreman_tasks_proxy_batch_trigger'
+        GlobalParameter().set({'name': param_name, 'value': False})
+        self.org = module_org
+        self.client = fixture_vmsetup
```
JacobCallahan, line 613: why bother storing these as class attributes?

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -583,6 +584,58 @@ def test_positive_run_reccuring_job(self, fixture_vmsetup, module_org):
...
+        client2 = fixture_vmsetup
```
JacobCallahan, line 614: would this not be the same host?

JacobCallahan (changes requested):

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k test_positive_run_concurrent_jobs

tests/foreman/cli/test_remoteexecution.py
swadeley: That line occurs multiple times in each class in the file, and `fixture_vmsetup` is function class, so the VM is just used per test.  Perhaps another PR to git rid of all use of `self` and change them all to just be client1?

swadeley: and at the same time check bugs cited in this file have UUIDs:
https://bugzilla.redhat.com/show_bug.cgi?id=1817320
https://bugzilla.redhat.com/show_bug.cgi?id=1451675
https://bugzilla.redhat.com/show_bug.cgi?id=1818076

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -583,6 +584,68 @@ def test_positive_run_reccuring_job(self, fixture_vmsetup, module_org):
...
+
+        :CaseLevel: System
+
+        :BZ: 1817320
+
+        :parametrized: yes
+        """
+        param_name = 'foreman_tasks_proxy_batch_trigger'
+        GlobalParameter().set({'name': param_name, 'value': 'false'})
+        self.client = fixture_vmsetup
```
JacobCallahan, line 612: What is the reason for aliasing this?

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -583,6 +584,68 @@ def test_positive_run_reccuring_job(self, fixture_vmsetup, module_org):
...
+        output_msgs = []
+        with VMBroker(nick=fixture_vmsetup, host_classes={'host': ContentHost}) as client2:
```
mshriver, line 614: Since `fixture_vmsetup` is a method that is indirectly parametrized, I believe it will be the client instance here, not  just the indirectly parametrized string.

tests/foreman/cli/test_remoteexecution.py
pondrejk: It's true these are no longer needed, removed the ones in this test, will adjust all others in separate PR. I added comments in the affected bzs

tests/foreman/cli/test_remoteexecution.py
pondrejk: indeed, calling vmbroker instead

tests/foreman/cli/test_remoteexecution.py
pondrejk: I need this to collect output messages later (around line 640)

tests/foreman/cli/test_remoteexecution.py
pondrejk: fixture_vmsetup.param is what I had in mind, cheers

JacobCallahan (approved):

mshriver (approved):

PR #8611, opened 2021-05-14
Fix HC list-by-org test

swadeley:
Hello

test fails if run in isolation. This fixes that in a cost effective way.


mshriver (approved):

swadeley:
Thank you

PR #8612, opened 2021-05-14
Underscore under-estimated

jyejare:
Azure tests fixed failing due to `underscore` in name, fixed it with `hypen`.


mshriver:
That's one of the dumbest vm naming restrictions I've seen!

PR #8616, opened 2021-05-18
migrated the custom certs jenkins job inside robottelo

omkarkhatavkar:
#### PR Description
This is pr is for migrating all custom certs related testing from Jenkins jobs to move in robottelo 

#### Test Result 
will be updating soon ...

### DIFF
--- a/pytest_fixtures/broker.py
+++ b/pytest_fixtures/broker.py
@@ -44,6 +44,15 @@ def capsule_factory():
 
 
 @pytest.fixture
+def param_host(request):
+    """A function-level fixture that provides a host object based broker args"""
+    if isinstance(request.param, dict):
+        nick = request.param.pop('nick')
+        with VMBroker(nick=nick, **request.param) as host:
+            yield host
+
+
+@pytest.fixture
 def rhel7_host():
     """A function-level fixture that provides a host object based on the rhel7 nick"""
     with VMBroker(nick='rhel7') as host:
--- a/pytest_fixtures/certificate.py
+++ b/pytest_fixtures/certificate.py
@@ -1,0 +1,160 @@
+from pathlib import Path
+
+import pytest
+from fauxfactory import gen_string
+
+from robottelo import ssh
+from robottelo.config import settings
+from robottelo.helpers import get_data_file
+from robottelo.ssh import get_connection
+from robottelo.ssh import upload_file
+
+
+@pytest.fixture(scope="module")
+def cert_data():
+    """Get host name, scripts, and create working directory."""
+
+    def _cert_data(hostname=None):
+        sat6_hostname = hostname or settings.server.hostname
+        capsule_hostname = 'capsule.example.com'
+        key_file_name = '{0}/{0}.key'.format(sat6_hostname)
+        cert_file_name = '{0}/{0}.crt'.format(sat6_hostname)
+        ca_bundle_file_name = 'cacert.crt'
+        success_message = "Validation succeeded"
+        cert_data = {
+            'sat6_hostname': sat6_hostname,
+            'capsule_hostname': capsule_hostname,
+            'key_file_name': key_file_name,
+            'cert_file_name': cert_file_name,
+            'ca_bundle_file_name': ca_bundle_file_name,
+            'success_message': success_message,
+        }
+        return cert_data
+
+    return _cert_data
+
+
+@pytest.fixture()
+def cert_setup(cert_data):
+    """copy all configuration files to satellite host for generating custom certs"""
+
+    def _cert_setup(hostname=None):
+        # Need a subdirectory under ssl-build with same name as Capsule name
+        hostname = hostname or settings.server.hostname
+        _cert_data = cert_data(hostname)
+        with get_connection(timeout=100, hostname=hostname) as connection:
+            connection.run('mkdir ssl-build/{}'.format(_cert_data['capsule_hostname']))
+            # Ignore creation error, but assert directory exists
+            assert connection.run('test -e ssl-build/{}'.format(_cert_data['capsule_hostname']))
+        upload_file(
+            local_file=get_data_file('generate-ca.sh'),
+            remote_file="generate-ca.sh",
+            hostname=hostname,
+        )
+        upload_file(
+            local_file=get_data_file('generate-crt.sh'),
+            remote_file="generate-crt.sh",
+            hostname=hostname,
+        )
+        upload_file(
+            local_file=get_data_file('openssl.cnf'), remote_file="openssl.cnf", hostname=hostname
+        )
+        # create the CA cert.
+        with get_connection(timeout=300, hostname=hostname) as connection:
+            result = connection.run('echo 100001 > serial')
+            result = connection.run("bash generate-ca.sh")
+            assert result.return_code == 0
+        # create the Satellite's cert
+        with get_connection(timeout=300, hostname=hostname) as connection:
+            result = connection.run(
+                "yes | bash {} {}".format('generate-crt.sh', _cert_data['sat6_hostname'])
+            )
+            assert result.return_code == 0
+
+    return _cert_setup
+
+
+@pytest.fixture()
+def certs_cleanup():
+    """cleanup all cert configuration files"""
+    yield
+    files = [
+        'cacert.crt',
+        'certindex*',
+        'generate-*.sh',
+        'capsule_cert' 'openssl.cnf',
+        'private',
+        'serial*',
+        'certs/*',
+        settings.server.hostname,
+    ]
+    with get_connection(timeout=300) as connection:
+        files = " ".join(files)
+        result = connection.run(f"rm -rf {files}")
+        assert result.return_code == 0
+
+
+@pytest.fixture()
+def generate_certs():
+    """generate custom certs in satellite host"""
+    upload_file(
+        local_file=get_data_file('certs.sh'),
+        remote_file="certs.sh",
+    )
+    upload_file(
+        local_file=get_data_file('extensions.txt'),
+        remote_file="extensions.txt",
+    )
+    with get_connection(timeout=300) as connection:
+        result = connection.run("bash certs.sh")
+        assert result.return_code == 0
+
+
+@pytest.fixture()
+def update_system_date():
+    """update the satellite date to verify cert expiration"""
+    result = ssh.command("date -s 'next year'")
+    assert result.return_code == 0
+    yield
+    result = ssh.command("date -s 'last year'")
+    assert result.return_code == 0, 'Failed to revert the date setting'
+
+
+@pytest.fixture(scope="module")
+def file_setup():
+    """Create working directory and file."""
+    capsule_hostname = 'capsule.example.com'
+    tmp_dir = '/var/tmp/{}'.format(gen_string('alpha', 6))
+    caps_cert_file = f'{tmp_dir}/ssl-build/capsule.example.com/cert-data'
+    # Use same path locally as on remote for storing files
+    Path(f'{tmp_dir}/ssl-build/capsule.example.com/').mkdir(parents=True, exist_ok=True)
+    with get_connection(timeout=200) as connection:
+        result = ssh.command(f'mkdir {tmp_dir}')
+        assert result.return_code == 0, 'Create working directory failed.'
+        # Generate a Capsule cert for capsule.example.com
+        result = connection.run(
+            'capsule-certs-generate '
+            '--foreman-proxy-fqdn capsule.example.com '
+            '--certs-tar {}/capsule_certs.tar '.format(tmp_dir),
+            timeout=100,
+        )
+    return {
+        'tmp_dir': tmp_dir,
+        'caps_cert_file': caps_cert_file,
+        'capsule_hostname': capsule_hostname,
+    }
+
+
+@pytest.fixture()
+def custom_certs_factory(cert_data, cert_setup):
+    """generate the custom certs based on-demand hostname"""
+
+    def _custom_certs_factory(hostname=None):
+        if hostname:
+            cert_setup(hostname)
+            return cert_data(hostname)
+        else:
+            cert_setup()
+            return cert_data()
+
+    return _custom_certs_factory
--- a/pytest_fixtures/certificate.py
+++ b/pytest_fixtures/certificate.py
@@ -20,7 +20,7 @@ def cert_data():
         key_file_name = '{0}/{0}.key'.format(sat6_hostname)
         cert_file_name = '{0}/{0}.crt'.format(sat6_hostname)
         ca_bundle_file_name = 'cacert.crt'
-        success_message = "Validation succeeded"
+        success_message = 'Validation succeeded'
         cert_data = {
             'sat6_hostname': sat6_hostname,
             'capsule_hostname': capsule_hostname,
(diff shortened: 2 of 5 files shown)


JacobCallahan (changes requested):
This would likely be easier if you used the Capsule and Satellite host objects from robottelo.hosts. This is especially true now that you can copy files directly from one system to another. It will also significantly speed up execution times. We can talk more about it if you need more information.
Also, please use single quotes and f-strings wherever possible.

omkarkhatavkar:
@JacobCallahan I agree with you, it is increases the scope for this PR, this PR intention to migrate the certs jobs to Robottelo as the old CI will be absolute. We could end up with not running the test cases on new CI. I agree with you suggestion to move it to use the ssh2-python but if we can hold that in another PR it will be great. 

Please let me know your thoughts ? I can create another ticket/tracker to move it to Satellite Host And Capsule Host.

JacobCallahan:
@omkarkhatavkar I'm fine with the Capsule/Satellite object conversions being done in another PR. For this one, just clean up that string formatting.

omkarkhatavkar:
> @omkarkhatavkar I'm fine with the Capsule/Satellite object conversions being done in another PR. For this one, just clean up that string formatting.

https://github.com/SatelliteQE/robottelo/issues/8636 (**Created**)

omkarkhatavkar:
trigger: test-robottelo
pytest: tests/foreman/sys/test_custom_cert_installation.py

omkarkhatavkar:
trigger: test-robottelo
pytest: tests/foreman/sys/test_custom_cert_installation.py

omkarkhatavkar:
trigger: test-robottelo
pytest: tests/foreman/sys/test_katello_certs_check.py

omkarkhatavkar:
trigger: test-robottelo
pytest: tests/foreman/sys/test_katello_certs_check.py

pytest_fixtures/certificate.py
```diff
@@ -0,0 +1,150 @@
...
+    return {
+        'tmp_dir': tmp_dir,
+        'caps_cert_file': caps_cert_file,
+        'capsule_hostname': capsule_hostname,
+    }
+
+
+@pytest.fixture()
+def custom_certs_factory(cert_data, cert_setup):
+    """generate the custom certs based on-demand hostname"""
```
swadeley, line 140:
```suggestion
    """generate the custom certs, use host name if available"""
```
I could not understand this, so just a guess

tests/foreman/sys/test_custom_cert_installation.py
```diff
@@ -0,0 +1,120 @@
...
+        2. Install satellite with custom certs
+        3. Assert output does not report SSL certificate error
+        4. Assert all services are running
+
+    :expectedresults: Satellite should be updated with custom certs
+    """
+    rhel_host = param_host.hostname
+    custom_certs_factory(hostname=rhel_host)
+    with get_connection(hostname=rhel_host) as connection:
+        # subscribe satellite dog_food server
```
swadeley, line 90:
```suggestion
        # subscribe satellite to dog_food server
```

swadeley (approved):
ACK, with some typos

pytest_fixtures/broker.py
```diff
@@ -43,6 +43,15 @@ def factory(retry_limit=3, delay=300, **broker_args):
     return factory
 
 
+@pytest.fixture
+def param_host(request):
+    """A function-level fixture that provides a host object based broker args"""
+    if isinstance(request.param, dict):
+        nick = request.param.pop('nick')
```
mshriver, line 50: I don't think there's any need to separate nick from the rest of the keys in the dictionary, and just pass `**request.param` to VMBroker.

This is a good start to a more abstract broker fixture, but puts the requirement on the test to parametrize everything without providing any defaults.

`param_host` doesn't give much context within the scope of the tests that are using it, and should provide some context to differentiate the types of content hosts that we'll be providing through broker.  I'd suggest something like `rhel_vm`, which at least shows that its a rhel virtual machine. We'll be parametrizing attributes of that rhel vm, like its version, compute resources, boot config (FIPS), and count. We'll also be building `rhel_containers`, and who knows what other OS or host types.

The options and exact parameter values for these parametrization matrixes for these fixtures should be decoupled from the test framework through the settings, so that they're not hardcoded in any way. Test functions, classes, and modules need to be able to limit some of these parameters - but only to what's within scope of the available configuration.

pytest_fixtures/certificate.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+@pytest.fixture(scope="module")
+def cert_data():
+    """Get host name, scripts, and create working directory."""
+
+    def _cert_data(hostname=None):
+        sat6_hostname = hostname or settings.server.hostname
+        capsule_hostname = 'capsule.example.com'
+        key_file_name = '{0}/{0}.key'.format(sat6_hostname)
```
mshriver, line 20: May I suggest:  `'/'.join([sat6_hostname]*2)`

pytest_fixtures/certificate.py
```diff
@@ -0,0 +1,150 @@
...
+        cert_file_name = '{0}/{0}.crt'.format(sat6_hostname)
+        ca_bundle_file_name = 'cacert.crt'
```
mshriver, line 22: Why set variables for these hardcoded strings?

pytest_fixtures/certificate.py
```diff
@@ -0,0 +1,150 @@
...
+from fauxfactory import gen_string
+
+from robottelo import ssh
+from robottelo.config import settings
+from robottelo.helpers import get_data_file
+from robottelo.ssh import get_connection
+from robottelo.ssh import upload_file
+
+
+@pytest.fixture(scope="module")
```
mshriver, line 13: This isn't really a fixture, its just a function that you'll call within your other fixture.

pytest_fixtures/certificate.py
```diff
@@ -0,0 +1,150 @@
...
+    capsule_hostname = 'capsule.example.com'
+    tmp_dir = '/var/tmp/{}'.format(gen_string('alpha', 6))
+    caps_cert_file = f'{tmp_dir}/ssl-build/capsule.example.com/cert-data'
+    # Use same path locally as on remote for storing files
+    Path(f'{tmp_dir}/ssl-build/capsule.example.com/').mkdir(parents=True, exist_ok=True)
+    with get_connection(timeout=200) as connection:
+        result = ssh.command(f'mkdir {tmp_dir}')
+        assert result.return_code == 0, 'Create working directory failed.'
+        # Generate a Capsule cert for capsule.example.com
+        result = connection.run(
```
mshriver, line 125: Is there any reason you can't use (maybe with slight updates so certs_tar can be passed) `Satellite.capsule_certs_generate` for this fixture?

pytest_fixtures/certificate.py
```diff
@@ -0,0 +1,150 @@
...
+        local_file=get_data_file('extensions.txt'),
+        remote_file='extensions.txt',
+    )
+    with get_connection(timeout=300) as connection:
+        result = connection.run('bash certs.sh')
+        assert result.return_code == 0
+
+
+@pytest.fixture(scope='module')
+def file_setup():
```
mshriver, line 114: These fixtures should have more descriptive names, since they're defined in a framework wide fixture module. It would be very easy for a fixture with this name to be defined in other modules in the `pytest_fixtures` directory.  Something like `capsule_cert_generate`

pytest_fixtures/certificate.py
```diff
@@ -0,0 +1,150 @@
...
+    """copy all configuration files to satellite host for generating custom certs"""
+
+    def _cert_setup(hostname=None):
+        # Need a subdirectory under ssl-build with same name as Capsule name
+        hostname = hostname or settings.server.hostname
+        _cert_data = cert_data(hostname)
+        with get_connection(timeout=100, hostname=hostname) as connection:
+            connection.run('mkdir ssl-build/{}'.format(_cert_data['capsule_hostname']))
+            # Ignore creation error, but assert directory exists
+            assert connection.run('test -e ssl-build/{}'.format(_cert_data['capsule_hostname']))
```
mshriver, line 48: f-strings for these please

pytest_fixtures/certificate.py
```diff
@@ -0,0 +1,150 @@
...
+    return _cert_data
+
+
+@pytest.fixture()
+def cert_setup(cert_data):
+    """copy all configuration files to satellite host for generating custom certs"""
+
+    def _cert_setup(hostname=None):
+        # Need a subdirectory under ssl-build with same name as Capsule name
+        hostname = hostname or settings.server.hostname
```
mshriver, line 43: I really think you should be using `robottelo.Satellite` class instances here, and its `execute` method to send these ssh commands, instead of importing `get_connection`. Changing that pattern will move all of this ssh setup to ssh2-python through broker, which is far faster. It also gets rid of the need to use all of these context managers

pytest_fixtures/certificate.py
```diff
@@ -0,0 +1,150 @@
...
+        )
+        # create the CA cert.
+        with get_connection(timeout=300, hostname=hostname) as connection:
+            result = connection.run('echo 100001 > serial')
+            result = connection.run('bash generate-ca.sh')
+            assert result.return_code == 0
+        # create the Satellite's cert
+        with get_connection(timeout=300, hostname=hostname) as connection:
+            result = connection.run(
+                'yes | bash {} {}'.format('generate-crt.sh', _cert_data['sat6_hostname'])
```
mshriver, line 70: f-strings for these - one of the arguments is a bare string, which doesn't need to be formatted in and can be part of the string directly.

pytest_fixtures/certificate.py
```diff
@@ -0,0 +1,150 @@
...
+            )
+            assert result.return_code == 0
+
+    return _cert_setup
+
+
+@pytest.fixture()
+def certs_cleanup():
+    """cleanup all cert configuration files"""
+    yield
```
mshriver, line 80: Is this restoring previous cert configuration on the satellite host?

pytest_fixtures/certificate.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+@pytest.fixture(scope='module')
+def file_setup():
+    """Create working directory and file."""
+    capsule_hostname = 'capsule.example.com'
+    tmp_dir = '/var/tmp/{}'.format(gen_string('alpha', 6))
+    caps_cert_file = f'{tmp_dir}/ssl-build/capsule.example.com/cert-data'
+    # Use same path locally as on remote for storing files
+    Path(f'{tmp_dir}/ssl-build/capsule.example.com/').mkdir(parents=True, exist_ok=True)
```
mshriver, line 120: Please do not use `/var/tmp/` on the machine running the pytest session. There is a tmp_dir defined in the robottelo config,  `settings.robottelo.tmp_dir`.

https://github.com/SatelliteQE/robottelo/blob/master/conf/robottelo.yaml.template#L11

pytest_fixtures/certificate.py
```diff
@@ -0,0 +1,150 @@
...
+    with get_connection(timeout=300) as connection:
+        result = connection.run('bash certs.sh')
+        assert result.return_code == 0
+
+
+@pytest.fixture(scope='module')
+def file_setup():
+    """Create working directory and file."""
+    capsule_hostname = 'capsule.example.com'
+    tmp_dir = '/var/tmp/{}'.format(gen_string('alpha', 6))
```
mshriver, line 117: It doesn't look like what's created in this directory on the local system is ever cleaned up?

you should use `tempfile.TemporaryDirectory` with the `dir` parameter set to `settings.robottelo.tmp_dir` to create a temporary directory in a context manager, so that it will get cleaned up when the context exists.

pytest_fixtures/certificate.py
```diff
@@ -0,0 +1,150 @@
...
+    # Use same path locally as on remote for storing files
+    Path(f'{tmp_dir}/ssl-build/capsule.example.com/').mkdir(parents=True, exist_ok=True)
+    with get_connection(timeout=200) as connection:
+        result = ssh.command(f'mkdir {tmp_dir}')
+        assert result.return_code == 0, 'Create working directory failed.'
+        # Generate a Capsule cert for capsule.example.com
+        result = connection.run(
+            'capsule-certs-generate '
+            '--foreman-proxy-fqdn capsule.example.com '
+            '--certs-tar {}/capsule_certs.tar '.format(tmp_dir),
```
mshriver, line 128: f-strings

pytest_fixtures/certificate.py
```diff
@@ -0,0 +1,150 @@
...
+        'caps_cert_file': caps_cert_file,
+        'capsule_hostname': capsule_hostname,
+    }
+
+
+@pytest.fixture()
+def custom_certs_factory(cert_data, cert_setup):
+    """generate the custom certs based on-demand hostname"""
+
+    def _custom_certs_factory(hostname=None):
```
mshriver, line 142: I wouldn't really call this a fixture if you've written it as a factory like this. Any actual certificate setup defined within this factory function stack ends up happening in the pytest runtest loop's call phase, not the setup phase.

tests/foreman/sys/test_custom_cert_installation.py
```diff
@@ -0,0 +1,120 @@
...
+        1. Generate the custom certs
+        2. Update satellite with custom certs
+        3. Assert output does not report SSL certificate error
+        4. Assert all services are running
+
+    :expectedresults: Satellite should be updated with custom certs
+    """
+    satellite = satellite_latest.hostname
+    cert_data = custom_certs_factory(hostname=satellite)
+    with get_connection(hostname=satellite) as connection:
```
mshriver, line 42: Please use `satellite_latest.execute()` to send ssh commands to the host. This is taking a brand new broker host instance and then only reading its hostname and using the older ssh methods, which are slower :)

tests/foreman/sys/test_custom_cert_installation.py
```diff
@@ -0,0 +1,120 @@
...
+                cert_data['cert_file_name'],
+                cert_data['key_file_name'],
+                cert_data['ca_bundle_file_name'],
+            ),
+            output_format='plain',
+        )
+        assert result.return_code == 0, 'Failed to generate the custom cert check command'
+
+        # update the custom certs in Satellite
+        command = f'satellite-installer --scenario satellite \
```
mshriver, line 54: Please use the new InstallerCommand class to build this installer command:

https://github.com/SatelliteQE/robottelo/blob/4430782b17b33fceda28e52af84b69fdecbdaa85/robottelo/helpers.py#L718

pytest_fixtures/certificate.py
```diff
@@ -0,0 +1,150 @@
+from pathlib import Path
```
mshriver, line 1: This comment applies to the whole module.

Given how these fixture functions are defined (factory), and that the content of these functions is intended to be executed as test steps, not as setup, I think almost none of these should be fixtures.

They should largely be functions that are defined either on `broker.hosts.Host`, `robottelo.hosts.ContentHost`, `robottelo.hosts.Capsule`, or `robottelo.hosts.Satellite`.

Then you can limit certificate fixtures to functions that actually _call_ those instance methods on `broker.hosts.Host` based instances that the fixtures get from VmBroker contexts (or existing fixtures).  They would actually execute within the setup phase of the pytest runtest loop.

Your test functions would then call the relevant Host/ContentHost/Capsule/Satellite instance methods to execute the test steps of setting up certificates and installing Satellite.

This makes those functions available to the entire framework on an instance of the necessary class, to be used by any fixture or test function that needs to deal with certificates.

tests/foreman/sys/test_custom_cert_installation.py
```diff
@@ -0,0 +1,120 @@
...
+
+    :steps:
+        1. Generate the custom certs
+        2. Install satellite with custom certs
+        3. Assert output does not report SSL certificate error
+        4. Assert all services are running
+
+    :expectedresults: Satellite should be updated with custom certs
+    """
+    rhel_host = param_host.hostname
```
mshriver, line 87: Same as above - use the `robottelo.hosts.ContentHost` instance you have here, and call `execute` on it to send ssh commands.

tests/foreman/sys/test_custom_cert_installation.py
```diff
@@ -0,0 +1,120 @@
...
+    with get_connection(hostname=rhel_host) as connection:
+        # subscribe satellite dog_food server
+        dogfood_url = (
+            f'{settings.repos.dogfood_repo_host}/pub/katello-ca-consumer-latest.noarch.rpm'
+        )
+        activation_key = f'satellite-{settings.robottelo.satellite_version}-qa-rhel7'
+        result = connection.run(f'yum -y localinstall {dogfood_url}', output_format='plain')
+        assert result.return_code == 0, 'Failed to install the katello-ca-consumer certs'
+        result = connection.run(
+            f'subscription-manager register --org Sat6-CI --activationkey {activation_key}',
```
mshriver, line 98: This org should never be hardcoded.

tests/foreman/sys/test_custom_cert_installation.py
```diff
@@ -0,0 +1,120 @@
...
+    :expectedresults: Satellite should be updated with custom certs
+    """
+    rhel_host = param_host.hostname
+    custom_certs_factory(hostname=rhel_host)
+    with get_connection(hostname=rhel_host) as connection:
+        # subscribe satellite dog_food server
+        dogfood_url = (
+            f'{settings.repos.dogfood_repo_host}/pub/katello-ca-consumer-latest.noarch.rpm'
+        )
+        activation_key = f'satellite-{settings.robottelo.satellite_version}-qa-rhel7'
```
mshriver, line 94: The activation key `rhel7` component is going to be dependent on the content host that `param_host` fixture provided. You need to resolve it dynamically.

mshriver (changes requested):
Lots of things to discuss, hope I wasn't too rude in my review.

pytest_fixtures/certificate.py
mshriver: You can use that for `base_name`, and then pass `f'{base_name}.key` and `f'{base_name}.crt'` below

omkarkhatavkar:
This is already done with PR https://github.com/SatelliteQE/robottelo/pull/8884 hence closing this

PR #8624, opened 2021-05-19
Add broker_directory environment variable in config/__init__.py

JacobCallahan:
With the removal of robottelo.properties pending, this will need to be set here.

### DIFF
--- a/robottelo/config/__init__.py
+++ b/robottelo/config/__init__.py
@@ -43,6 +43,7 @@ try:
 except ValidationError:
     logger.warning("Dynaconf validation failed, continuing for the sake of unit tests")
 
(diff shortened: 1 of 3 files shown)


robottelo/config/__init__.py
```diff
@@ -43,6 +43,7 @@
 except ValidationError:
     logger.warning("Dynaconf validation failed, continuing for the sake of unit tests")
 
+os.environ["BROKER_DIRECTORY"] = settings.broker.broker_directory
```
JacobCallahan, line 46:
```suggestion
os.environ["BROKER_DIRECTORY"] = dynaconf_settings.broker.broker_directory    
```

JacobCallahan:
Closing as this has become too messy and the docblock failures are completely unrelated.

PR #8626, opened 2021-05-20
Add test delete product with synced docker repo

swadeley:
Closed loop: BZ#1867287

Adding test for:

[Bug 1867287 - Error Row was updated or deleted by another transaction when deleting docker repository](https://bugzilla.redhat.com/show_bug.cgi?id=1867287)

### DIFF
--- a/tests/foreman/api/test_repository.py
+++ b/tests/foreman/api/test_repository.py
@@ -82,6 +82,15 @@ def repo_options(request, module_org, module_product):
 
 
 @pytest.fixture
+def repo_options_custom_product(request, module_org):
+    """Return the options that were passed as indirect parameters."""
+    options = getattr(request, 'param', {}).copy()
+    options['organization'] = module_org
+    options['product'] = entities.Product(organization=module_org).create()
+    return options
+
+
+@pytest.fixture
 def env(module_org):
     """Create a new puppet environment."""
     return entities.Environment(organization=[module_org]).create()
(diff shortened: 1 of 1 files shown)


swadeley:
trigger: test-robottelo
pytest: tests/foreman/api/test_repository.py -k test_positive_delete_product_with_synced_repo

JacobCallahan (approved):

tests/foreman/api/test_repository.py
```diff
@@ -1659,6 +1668,38 @@ def test_positive_synchronize(self, repo):
...
+
+        :expectedresults: A product with a synced Docker repository can be deleted.
+
+        :CaseImportance: High
+
+        :BZ: 1867287
+        """
+        repo.sync()
+        assert repo.read().content_counts['docker_manifest'] >= 1
+        assert repo.product.delete()
```
mshriver, line 1701: Going to merge for the BZ coverage, but I think you should take a look at this module's test content with an eye on how efficient it is. This exact repository is created and synced for `test_positive_synchronize`, and factors like whether or not there is a custom product on the repo, what type of repo it is, etc make for great parameters.

Parametrization itself could be simplified, so that we're parametrizing key lists from dictionaries that hold core repository options. This provides better parameter contexts in the node ID (instead of the integers that the current parametrization style results in), and fewer test functions to maintain. A single test function can provide functional coverage for several actions without becoming overly complex or long.

While we'll repeat lots of API POST calls with slightly different data, we won't be doing it with slightly different test cases that have large chunks of repeated docblock and parametrization marks.

mshriver (approved):

tests/foreman/api/test_repository.py
swadeley: @mshriver Thank you for the review and feedback. As the module is not mine (I was just doing this PR as the Jira card was still in my backlog) I just wanted to make a non-disruptive addition. As I am writing news test for 6.10 I can learn to make better use of parametrisation.

swadeley:
Thank you

tests/foreman/api/test_repository.py
mshriver: ack, thanks @swadeley for adding the coverage!  One of those things that stuck out to me, and I figured it was better to call it when I see it :)

PR #8638, opened 2021-05-25
cli rex fixes

pondrejk:
applied some feedback collected in https://github.com/SatelliteQE/robottelo/pull/8608

### DIFF
--- a/tests/foreman/cli/test_remoteexecution.py
+++ b/tests/foreman/cli/test_remoteexecution.py
@@ -41,32 +41,44 @@ from robottelo.hosts import ContentHost
 
 
 @pytest.fixture()
-def fixture_vmsetup(request, module_org):
-    """Create Org, Lifecycle Environment, Content View, Activation key,
-    VM, install katello-ca, register it, add remote execution key
-    """
-    # Create VM and register content host
-    with VMBroker(nick=request.param, host_classes={'host': ContentHost}) as client:
-        client.install_katello_ca()
-        client.register_contenthost(org=module_org.label, lce='Library')
-        assert client.subscribed
-        add_remote_execution_ssh_key(client.ip_addr)
-        Host.set_parameter(
-            {
-                'host': client.hostname,
-                'name': 'remote_execution_connect_by_ip',
-                'value': 'True',
-            }
-        )
-        yield client
+def fixture_vmsetup(request):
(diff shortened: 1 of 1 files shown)


omkarkhatavkar (approved):
ACK

mshriver:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py

mshriver (changes requested):
After some discussion with @pondrejk this test needs some work for its secondary VM fixture.

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -41,32 +41,44 @@
...
-        Host.set_parameter(
-            {
-                'host': client.hostname,
-                'name': 'remote_execution_connect_by_ip',
-                'value': 'True',
-            }
-        )
-        yield client
+def fixture_vmsetup(request):
+    """ Create VM and register content host """
```
mshriver, line 45: Given the limited scope of this fixture (this module), these are non-blocking comments. 

Some thoughts for fixture implementation that are important when fixtures are at a global/framework level:

1. Pass through **request.param to the VMBroker constructor, so tests can parametrize any additional kwargs. This allows the fixture to scale without having to duplicate all of the possible kwargs.
2. Return a consistent data structure from the fixture, regardless of whether `count` is indirectly parametrized. This way the tests using the fixture have a consistent structure to deal with, regardless of whether they parametrized or not.
3. Consider how the ContentHost (or classes which inherit it and are more specific) has setup/teardown methods which can be defined and will be executed automatically within the VMBroker context manager.  The statements in the `_setup_host` function here could be in a `RexHost` class which inherits ContentHost, and implements `RexHost.setup` to do that specific setup.  Maybe the setup for this test module or functions are too specific to derive value from abstracting the setup actions into the host class where it would be easily re-usable in other test modules.

mshriver (approved):

PR #8646, opened 2021-05-26
The right place of helpers

jyejare:
The helpers in Configs are now moved to helpers.

- Removed helper functions / methods from settings objects from both Legacy and DynaConf Settings.
- Now all settings helpers are receding in New `settingsUtils` class in `helpers.py` as Static Methods.
    - Since we are moving out of Legacy settings, all the utils are now focusing on/using the DynaConf way of reading the settings.
- Number of test files are affected by this change and all of them are updated for the above changes.

Happy Helpers !

### DIFF
--- a/robottelo/helpers.py
+++ b/robottelo/helpers.py
@@ -5,6 +5,7 @@ import random
 import re
 from tempfile import mkstemp
 from urllib.parse import urljoin  # noqa
+from urllib.parse import urlunsplit
 
 import requests
 from nailgun.config import ServerConfig
@@ -18,6 +19,7 @@ from robottelo.constants import RHEL_6_MAJOR_VERSION
 from robottelo.constants import RHEL_7_MAJOR_VERSION
 from robottelo.errors import GCECertNotFoundError
 from robottelo.logging import logger
+from robottelo.host_info import get_sat_version
 
 
 class DataFileError(Exception):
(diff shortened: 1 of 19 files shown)


robottelo/helpers.py
```diff
@@ -797,3 +799,78 @@ def from_cmd_str(cls, command='satellite-installer', cmd_str=None):
     def __repr__(self):
         """Custom repr will give the constructed command output"""
         return self.get_command()
+
+
+class settingsUtils:
```
mshriver, line 804: I think these methods should largely be part of the `robottelo.hosts.ContentHost`, `Capsule`, or `Satellite` instances, instead of helpers - or just replaced entirely like the case of `capsule_hostname`.

jyejare: @mshriver Ohh ! Misunderstood the Jira ticket or the ticket was very old that didnt have clarity. We should have updated it. Never mind !

I think instead of reworking, I should late @JacobCallahan handle it in his PR of LegacySettings removal. Thoughts @JacobCallahan  ?

JacobCallahan: I'll revisit this on enim past my current phase of the refactor

jyejare: @JacobCallahan Ok. Closing this one in favor of your TODO work. Let me know if you need my help.

PR #8654, opened 2021-05-29
Subscription UI vm broker

akhil-jha:
The first two tests are failing due to bz1948758. Third test is failing at Line 510: `session.organization.select(org_name=org.name)`. I checked with pdb, the host gets registered with the satellite successfully.

### DIFF
--- a/tests/foreman/ui/test_subscription.py
+++ b/tests/foreman/ui/test_subscription.py
@@ -508,16 +508,18 @@ def test_positive_candlepin_events_processed_by_STOMP(session, rhel7_contenthost
     rhel7_contenthost.register_contenthost(org.name, ak.name)
     with session:
         session.organization.select(org_name=org.name)
-        host = session.contenthost.read(rhel7_contenthost.hostname, widget_names='details')['details']
+        host = session.contenthost.read(rhel7_contenthost.hostname, widget_names='details')[
+            'details'
+        ]
         sub_status = host['subscription_status']
         assert "Unentitled" in sub_status
         with manifests.clone() as manifest:
             upload_manifest(org.id, manifest.content)
         session.contenthost.add_subscription(rhel7_contenthost.hostname, DEFAULT_SUBSCRIPTION_NAME)
         session.browser.refresh()
-        updated_sub_status = session.contenthost.read(rhel7_contenthost.hostname, widget_names='details')[
-            'details'
-        ]['subscription_status']
+        updated_sub_status = session.contenthost.read(
+            rhel7_contenthost.hostname, widget_names='details'
+        )['details']['subscription_status']
         assert "Fully entitled" in updated_sub_status
         response = entities.Ping().search_json()["services"]["candlepin_events"]
         assert response["status"] == "ok"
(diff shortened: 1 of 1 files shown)


JacobCallahan (changes requested):
Make sure to use single quote string wherever possible and I'll do a more technical check when I can.

tests/foreman/ui/test_subscription.py
```diff
@@ -290,30 +289,28 @@ def test_positive_view_vdc_subscription_products(session):
...
+        activation_key=repos_collection.setup_content_data['activation_key']['name'],
+        install_katello_agent=False,
+    )
+    with session:
+        session.organization.select(org.name)
+        session.contenthost.add_subscription(rhel7_contenthost.hostname, VDC_SUBSCRIPTION_NAME)
+        provided_products = session.subscription.provided_products(VDC_SUBSCRIPTION_NAME)
+        # ensure that subscription provided products list is not empty and that the product is
+        # in the provided products.
+        assert provided_products and product_name in provided_products
```
JacobCallahan, line 304:
```suggestion
        assert product_name in provided_products
```
As long as provided_products is always a list (or other iterable), then the first check isn't needed.
This type of check is used throughout the other tests in this PR.

I know this was part of the original tests, but could you clean these up as well? :)

tests/foreman/ui/test_subscription.py
```diff
@@ -510,22 +504,23 @@ def test_positive_candlepin_events_processed_by_STOMP(session):
...
-            assert "0 Failed" in response["message"]
+    rhel7_contenthost.install_katello_ca()
+    rhel7_contenthost.register_contenthost(org.name, ak.name)
+    with session:
+        session.organization.select(org_name=org.name)
+        host = session.contenthost.read(rhel7_contenthost.hostname, widget_names='details')[
+            'details'
+        ]
+        sub_status = host['subscription_status']
+        assert 'Unentitled' in sub_status
```
JacobCallahan, line 515:
```suggestion
        assert 'Unentitled' in host['subscription_status']
```

JacobCallahan (changes requested):

tests/foreman/ui/test_subscription.py
akhil-jha: Done

tests/foreman/ui/test_subscription.py
akhil-jha: Done. :)

JacobCallahan (approved):
ACK, nice!

tests/foreman/ui/test_subscription.py
```diff
@@ -250,7 +249,7 @@ def test_positive_access_manifest_as_another_admin_user(test_name):
 
 @pytest.mark.libvirt_content_host
```
jameerpathan111, line 250:
```suggestion
```

JacobCallahan: good catch!

jameerpathan111 (approved):

PR #8656, opened 2021-05-31
addition in check to validate the katello-cert-check installer command

omkarkhatavkar:
More Details https://bugzilla.redhat.com/show_bug.cgi?id=1858819

### DIFF
--- a/tests/foreman/sys/test_katello_certs_check.py
+++ b/tests/foreman/sys/test_katello_certs_check.py
@@ -174,6 +174,13 @@ class TestKatelloCertsCheck:
         }
         assert result.return_code == 0
         assert cert_data['success_message'] in result.stdout
+
+        # validate correct installer used
+        pattern, count = 'satellite-installer --scenario satellite', 2
+        assert len(re.findall(pattern, result.stdout)) == count
+        pattern, count = 'foreman-installer --scenario katello', 0
+        assert len(re.findall(pattern, result.stdout)) == count
+
         # validate all checks passed
         assert not any(flag for flag in re.findall(r"\[([A-Z]+)\]", result.stdout) if flag != 'OK')
         # validate options in output


omkarkhatavkar:
trigger: test-robottelo
pytest: tests/foreman/sys/test_katello_certs_check.py

omkarkhatavkar:
one test failed `assert result.return_code == 0, 'Not all services are running` not related to this PR change

tests/foreman/sys/test_katello_certs_check.py
```diff
@@ -174,6 +174,13 @@ def validate_output(self, result, cert_data):
         }
         assert result.return_code == 0
         assert cert_data['success_message'] in result.stdout
+
+        # validate correct installer used
+        pattern, count = 'satellite-installer --scenario satellite', 2
```
jyejare, line 179: Why should we have it twice or why we are hardcoding this value ?

jyejare (commented):
ACK pending question.

tests/foreman/sys/test_katello_certs_check.py
```diff
@@ -174,6 +174,13 @@ def validate_output(self, result, cert_data):
...
+        assert len(re.findall(pattern, result.stdout)) == count
+        pattern, count = 'foreman-installer --scenario katello', 0
```
Gauravtalreja1, line 181:
```suggestion
        pattern, count = 'satellite-installer --scenario katello', 0
```

tests/foreman/sys/test_katello_certs_check.py
JacobCallahan: There are two alternate forms of the installer command in the output, based on the bug description. So there should be two matches in the valid output.

tests/foreman/sys/test_katello_certs_check.py
JacobCallahan: This check is to make sure foreman-installer isn't being recommended.

JacobCallahan (approved):

tests/foreman/sys/test_katello_certs_check.py
Gauravtalreja1: Oh sry, I missed BZ despcription.

tests/foreman/sys/test_katello_certs_check.py
omkarkhatavkar: @jyejare katello-cert-check command returns the two installer commands, hence the match count is 2, we are not recommended to give 'foreman-installer' as a command. About `hardcoding` can you give me more suggestions I'm not able to understand?      

`To install the Red Hat Satellite Server with the custom certificates, run:

    satellite-installer --scenario satellite \
                      --certs-server-cert "/root/ownca/xyz.com/xyz.com.crt" \
                      --certs-server-key "/root/ownca/xyz.com/xyz.com.key" \
                      --certs-server-ca-cert "/root/ownca/xyz.com/cacert.crt"

To update the certificates on a currently running Red Hat Satellite installation, run:

    satellite-installer --scenario satellite \
                      --certs-server-cert "/root/ownca/xyz.com/xyz.com.crt" \
                      --certs-server-key "/root/ownca/xyz.com/xyz.com.key" \
                      --certs-server-ca-cert "/root/ownca/xyz.com/cacert.crt" \
                      --certs-update-server --certs-update-server-ca

To use them inside a NEW $CAPSULE, rerun this command with -t capsule`

jyejare: ACK! Thanks for the detailed description.

@omkarkhatavkar Yeah there is no such way to calculate / how many commands the STDOUT will produce. The only way I think is to initiate a constant variable in `constants/__init__.py` with value `2`. This will help the next author to understand it better.

But this is not a blocking comment though.

jyejare (approved):

PR #8657, opened 2021-05-31
VMBroker conversion rex in upgrades

pondrejk:
Needs review from upgrade specialists :) Also I wasn't sure how to do the vm cleanup postupgrade

### DIFF
--- a/tests/upgrades/test_remoteexecution.py
+++ b/tests/upgrades/test_remoteexecution.py
@@ -23,11 +23,8 @@ from upgrade_tests import pre_upgrade
 from upgrade_tests.helpers.scenarios import create_dict
 from upgrade_tests.helpers.scenarios import get_entity_data
 
-from robottelo.cleanup import cleanup_of_provisioned_server
 from robottelo.config import settings
-from robottelo.constants import DISTRO_RHEL7
 from robottelo.helpers import add_remote_execution_ssh_key
-from robottelo.vm import VirtualMachine
 
 
 @pytest.fixture(scope='class')
@@ -70,7 +67,9 @@ class TestScenarioREXCapsule:
     """
 
     @pre_upgrade
-    def test_pre_scenario_remoteexecution_external_capsule(self, request, default_location):
+    def test_pre_scenario_remoteexecution_external_capsule(
+        self, request, default_location, rhel7_contenthost
+    ):
         """Run REX job on client registered with external capsule
 
         :id: preupgrade-261dd2aa-be01-4c34-b877-54b8ee346561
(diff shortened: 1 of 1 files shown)


tests/upgrades/test_remoteexecution.py
```diff
@@ -96,17 +95,7 @@ def test_pre_scenario_remoteexecution_external_capsule(self, request, default_lo
...
-        )
-        client.create()
-        request.addfinalizer(
-            lambda: cleanup_of_provisioned_server(
-                hostname=client.hostname,
-                provisioning_server=self.libvirt_vm,
-                distro=DISTRO_RHEL7,
-            )
-        )
+        client = VMBroker(nick='rhel7', host_classes={'host': ContentHost})
```
jyejare, line 98: Instead, you should use the existing function level fixture :
https://github.com/SatelliteQE/robottelo/blob/master/pytest_fixtures/broker.py#L47

tests/upgrades/test_remoteexecution.py
```diff
@@ -207,17 +193,8 @@ def test_pre_scenario_remoteexecution_satellite(
...
-        client.create()
-        request.addfinalizer(
-            lambda: cleanup_of_provisioned_server(
-                hostname=client.hostname,
-                provisioning_server=self.libvirt_vm,
-                distro=DISTRO_RHEL7,
-            )
-        )
+
+        client = VMBroker(nick='rhel7', host_classes={'host': ContentHost})
```
jyejare, line 197: Instead, you should use the existing function level fixture :
https://github.com/SatelliteQE/robottelo/blob/master/pytest_fixtures/broker.py#L47

jyejare (changes requested):
ACK pending comments.

tests/upgrades/test_remoteexecution.py
```diff
@@ -96,17 +95,7 @@ def test_pre_scenario_remoteexecution_external_capsule(self, request, default_lo
...
-        )
-        client.create()
-        request.addfinalizer(
-            lambda: cleanup_of_provisioned_server(
-                hostname=client.hostname,
-                provisioning_server=self.libvirt_vm,
-                distro=DISTRO_RHEL7,
-            )
-        )
+        client = rhel7_contenthost
```
jyejare, line 98: You can directly use it, you need not initialise the `client` variable for it.

Below line will change to:
```
rhel7_contenthost.install_capsule_katello_ca(capsule=self.proxy_name)
```
Same others.

jyejare:
trigger: test-robottelo
pytest: -m pre_upgrade tests/upgrades/test_remoteexecution.py

mshriver (approved):
The test failures in PRT are well outside the scope of what you're doing here.

Your conversion to using VMBroker content hosts is clean.

PR #8659, opened 2021-05-31
Contentmanagement api vm broker

akhil-jha:
Capsule installation is failing due to a package.

PR #8660, opened 2021-05-31
adding tmp directory

akhil-jha:
I encountered this when my capsule installation in the PR #8659  failed in PRT 127 with error `failed on setup with "FileNotFoundError: [Errno 2] No such file or directory:<virtualenv>robottelo/tmp/capsule-<ip>.log` in sat-results. The missing directory is `robottelo/tmp`. When the capsule installation failed, according to [this](https://github.com/SatelliteQE/robottelo/blob/3dd2a277701ac46467b9e97279c43db1081d5489/robottelo/hosts.py#L504) , the capsule.log should be copied over to the base testing machine.

Or am I missing something?

### DIFF
--- a/tmp/example-capsule-ip.log
+++ b/tmp/example-capsule-ip.log
@@ -1,0 +1,1 @@
+sample error file


JacobCallahan (changes requested):
Let's hold on this and see if there is a better way.

mshriver:
@JacobCallahan Shouldn't this be resolved by `broker.session.sftp_read` calling `Path.mkdir` before trying to write?

mshriver (changes requested):
The framework should handle writing this file whether or not the file or the target directory exists.

JacobCallahan:
@mshriver that is indeed where this should be handled. I can update the existing broker PR to account for this.
If that solution is agreeable to everyone else, then this PR can be closed.

akhil-jha:
> @mshriver that is indeed where this should be handled. I can update the existing broker PR to account for this.
> If that solution is agreeable to everyone else, then this PR can be closed.

I think that should be okay. 
Closing this PR.

JacobCallahan:
This is solved by the addition of these lines in Broker's session class.
https://github.com/SatelliteQE/broker/pull/109/files#diff-6e88e617cde4e6aa876445eaaa4845b26cee35f289eb83c21cbbde2b103b44b0R74

PR #8661, opened 2021-06-01
Convert UI hostcollection to VMBroker

swadeley:
Hello

Convert UI host collection to VMBroker

### DIFF
--- a/tests/foreman/ui/test_hostcollection.py
+++ b/tests/foreman/ui/test_hostcollection.py
@@ -101,7 +101,7 @@ def vm_content_hosts(request, module_loc, module_repos_collection):
         for client in clients:
             module_repos_collection.setup_virtual_machine(client)
             update_vm_host_location(client, module_loc.id)
-    return clients
+        yield clients
 
 
 @pytest.fixture
--- a/tests/foreman/ui/test_hostcollection.py
+++ b/tests/foreman/ui/test_hostcollection.py
@@ -19,6 +19,7 @@
 import time
 
 import pytest
+from broker import VMBroker
 from nailgun import entities
 
 from robottelo.api.utils import promote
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_hostcollection.py
```diff
@@ -95,9 +96,10 @@ def module_repos_collection_module_stream(module_org, module_lce):
 
 @pytest.fixture
 def vm_content_hosts(request, module_loc, module_repos_collection):
+    distro = module_repos_collection.distro
     clients = []
     for _ in range(2):
-        client = VirtualMachine(distro=module_repos_collection.distro)
+        client = VMBroker(nick=distro, host_classes={'host': ContentHost})
```
mshriver, line 102: So, this only gets you a `VMBroker` instance - not the host instance. `VMBroker` does not have `create` or `destroy` methods defined, so I don't believe this will work at all.

This needs to be converted to use the VMBroker context manager, which will call `checkout` for all the hosts defined by the VMBroker instance ( the count can be included, like the use below, instead of iterating).

The call to `request.addfinalizer` can be removed, as well as the call to `client.create`, as the context manager takes care of both of these things.

mshriver (changes requested):
Small but significant change to using the context manager in the `vm_content_hosts` fixture.

tests/foreman/ui/test_hostcollection.py
swadeley: OK, changes made, but still to be tested

swadeley:
Hello

Running pytest on the file, I get two unrelated fails due to airgun issues:

def test_positive_remove_package_group
fails with:
ElementNotInteractableException: Message: element not interactable


and

def test_positive_install_errata
fails with
WidgetOperationFailed: Failed to set the checkbox to requested value

swadeley:
trigger: test-robottelo
pytest: tests/foreman/ui/test_hostcollection.py

swadeley:
this now passes locally:
tests/foreman/ui/test_hostcollection.py::test_positive_remove_package_group

swadeley:
this test now passes locally
tests/foreman/ui/test_hostcollection.py::test_positive_install_errata

I think I was hitting my VMBroker limit which causes tests to fail when it cannot set-up enough content hosts

mshriver (approved):

ColeHiggins2 (approved):
ACK

JacobCallahan (approved):

swadeley:
Thank you

PR #8664, opened 2021-06-02
Fix crud infrastructure tests for 6.10

jameerpathan111:
This PR fixes few issues related to new PatternFly 4 changes for Infrastructure tab in Satellite 6.10. 
I can't provide test result for http proxy test because of [bz#1966942](https://bugzilla.redhat.com/show_bug.cgi?id=1966942)
and I was not able to run computeresource test. With that said, the code change in this PR is not likely to cause any regression.

```
$ pytest tests/foreman/ui/test_domain.py -k test_positive_end_to_end 
platform linux -- Python 3.8.6, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
rootdir: /home/jpathan/projects/robottelo, configfile: pyproject.toml
plugins: ibutsu-1.13, services-2.2.1, cov-2.11.1, mock-3.5.1, reportportal-5.0.7, xdist-2.2.1, forked-1.3.0

tests/foreman/ui/test_domain.py .                                        [100%]

=========== 1 passed, 5 deselected, 10 warnings in 176.66s (0:02:56) ===========
```

Note: Depends on https://github.com/SatelliteQE/airgun/pull/575

### DIFF
--- a/tests/foreman/ui/test_computeresource.py
+++ b/tests/foreman/ui/test_computeresource.py
@@ -118,7 +118,7 @@ def test_positive_end_to_end(session, rhev_data, module_org, module_loc, module_
         assert not session.computeresource.search(name)
         assert session.computeresource.search(new_name)[0]['Name'] == new_name
         session.computeresource.delete(new_name)
-        assert not session.computeresource.search(new_name)
+        assert not entities.AbstractComputeResource().search(query={'search': f'name={new_name}'})
 
 
 @pytest.mark.on_premises_provisioning
--- a/tests/foreman/ui/test_domain.py
+++ b/tests/foreman/ui/test_domain.py
@@ -174,4 +174,4 @@ def test_positive_end_to_end(session, module_org, module_loc, valid_domain_name)
         assert session.domain.search(new_name)[0]['Description'] == new_name
         # Delete domain
         session.domain.delete(new_name)
-        assert not session.domain.search(new_name)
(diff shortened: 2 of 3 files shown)


tests/foreman/ui/test_computeresource.py
```diff
@@ -118,7 +118,7 @@ def test_positive_end_to_end(session, rhev_data, module_org, module_loc, module_
         assert not session.computeresource.search(name)
         assert session.computeresource.search(new_name)[0]['Name'] == new_name
         session.computeresource.delete(new_name)
-        assert not session.computeresource.search(new_name)
+        assert not entities.AbstractComputeResource().search(query={'search': f'name={new_name}'})
```
ColeHiggins2, line 121: Given that this is a UI test, are there any other ways to assert this query ? Perhaps another use of "session" in the assertion? The same goes for the ones below.

jameerpathan111: The issue is once we delete the entity(compute resource, http proxy etc.) and if there is no other entity of same kind available, we see the main landing page. The `session.computeresource.search(new_name)` kind of assertion won't work because there will be no serchbox. 
There are few other UI test which use same assertion style for this kind of scenario. That's why I changed it this way.
I don't know if there's any other way we can do assertion for this. Any suggestions?

ColeHiggins2: That makes sense, should be fine as is.

JacobCallahan (approved):

PR #8680, opened 2021-06-10
Parametrize rhel content host fixtures

mshriver:
Very early draft.

We may need the complexity offered by even far more powerful parametrization driven by marks.

The ManageIQ/integration_tests project had an abstract mark parametrization, used for compute resource providers, this may turn into that instead.

https://github.com/ManageIQ/integration_tests/blob/5c9fea2b256a283700e664b31f586b2681d139b3/cfme/markers/env.py

https://github.com/ManageIQ/integration_tests/blob/5c9fea2b256a283700e664b31f586b2681d139b3/cfme/markers/env_markers/provider.py

https://github.com/ManageIQ/integration_tests/blob/5c9fea2b256a283700e664b31f586b2681d139b3/cfme/utils/testgen.py

### DIFF
--- a/tests/foreman/api/test_subscription.py
+++ b/tests/foreman/api/test_subscription.py
@@ -234,7 +234,7 @@ def test_positive_delete_manifest_as_another_user(function_org):
 
 
 @pytest.mark.tier2
-def test_positive_subscription_status_disabled(module_ak, rhel7_contenthost, module_org):
+def test_positive_subscription_status_disabled(module_ak, rhel_contenthost, module_org):
     """Verify that Content host Subscription status is set to 'Disabled'
      for a golden ticket manifest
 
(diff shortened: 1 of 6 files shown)


rplevka:
@mshriver I like this concept - are you planning to work on this and eventually un-draft it?

jaryn (approved):
Easy to understand code.

tests/foreman/api/test_subscription.py
```diff
@@ -232,7 +232,7 @@ def test_positive_delete_manifest_as_another_user(function_org):
 
 
 @pytest.mark.tier2
-def test_positive_subscription_status_disabled(module_ak, rhel7_contenthost, module_org):
+def test_positive_subscription_status_disabled(module_ak, rhel_contenthost, module_org):
```
mshriver, line 235: @ColeHiggins2 do you think this is a good test to parametrize rhel versions for this test case?

JacobCallahan (approved):

PR #8685, opened 2021-06-14
Add stubbed test cases for pulp3 feature area

latran:
Stubbed test cases for pulp3.  I'm putting these in ContentView component for now because they're stubbed.  It feels needless to create a separate file for just pulp stubs.  If automation can be written for this in the future, then I'll transition them to a separate file specifically for pulp.

### DIFF
--- a/tests/foreman/api/test_contentview.py
+++ b/tests/foreman/api/test_contentview.py
@@ -1722,3 +1722,33 @@ class TestContentViewRedHatOstreeContent:
         content_view.publish()
         promote(content_view.read().version[0], module_lce.id)
         assert len(content_view.read().version[0].read().environment) == 2
+
+    @pytest.mark.stubbed
+    @pytest.mark.tier1
+    def test_positive_check_pulp3_is_installed(self):
+        """Ensure that pulp3 is installed
+
+        :id: 5d5690bf-56c7-4dba-bb09-de0f5f4861b1
+
+        :expectedresults: Pulp3 is active
+
+        :CaseLevel: System
+
+        :CaseImportance: Critical
+        """
+        pass
+
+    @pytest.mark.stubbed
+    @pytest.mark.tier1
+    def test_positive_check_pulp2_is_removed(self):
+        """Ensure that pulp2 is removed from system
+
+        :id: 4b0ebd09-e417-4c42-a437-d11db724e671
+
+        :expectedresults: Cannot log into pulp2 db
+
+        :CaseLevel: System
+
+        :CaseImportance: Critical
+        """
+        pass
--- a/tests/foreman/ui/test_contentview.py
+++ b/tests/foreman/ui/test_contentview.py
@@ -4111,3 +4111,33 @@ def test_positive_depsolve_with_module_errata(session, module_org):
             assert items['Name'] in rpm_pack
         assert result['module_streams']['table'][0]['Name'] == mod_stream
         assert result['errata']['table'][0]['Errata ID'] == FAKE_4_ERRATA_ID
+
+    @pytest.mark.stubbed
+    @pytest.mark.tier1
+    def test_positive_check_pulp3_is_installed():
+        """Ensure that pulp3 is installed
+
+        :id: 5cee7333-67de-4916-923c-f238b989f587
+
+        :expectedresults: Pulp3 is active
+
+        :CaseLevel: System
+
+        :CaseImportance: Critical
+        """
+        pass
+
+    @pytest.mark.stubbed
+    @pytest.mark.tier1
+    def test_positive_check_pulp2_is_removed():
+        """Ensure that pulp2 is removed from system
+
+        :id: e89673be-f43a-4daf-8dfa-fa86f02952be
+
+        :expectedresults: Cannot log into pulp2 db
+
+        :CaseLevel: System
+
+        :CaseImportance: Critical
+        """
+        pass
--- a/tests/foreman/cli/test_contentview.py
+++ b/tests/foreman/cli/test_contentview.py
@@ -4916,3 +4916,33 @@ class TestContentViewFileRepo:
         """
         result = ssh.command('sudo -u postgres psql -d foreman -c "\\d katello_repository_rpms"')
         assert 'id|bigint' in result.stdout[3].replace(' ', '')
+
+    @pytest.mark.stubbed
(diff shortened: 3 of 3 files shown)


mshriver:
Converted to draft as its failing code quality checks.

tests/foreman/api/test_contentview.py
```diff
@@ -1722,3 +1722,33 @@ def test_positive_publish_promote_with_RH_ostree_and_other(
         content_view.publish()
         promote(content_view.read().version[0], module_lce.id)
         assert len(content_view.read().version[0].read().environment) == 2
+
+    @pytest.mark.stubbed
+    @pytest.mark.tier1
+    def test_positive_check_pulp3_is_installed(self):
```
mshriver, line 1728: Do we need separate manual test cases for this? Is there no easy automation path for checking pulp3 processes or RPMs?

There are no test steps, so I'm not sure what your intent is for validating that pulp3 is active and pulp2 is not.

For a manual case, this just doubles the number of test results you need to enter for what is essentially two steps of one test case.

latran: The test cases can be condensed and I can add test steps as necessary for manual processes.  For now, these are just stubs that I can revisit to add automation later or can leave in draft state until automation is added.  One verification is via UI that pulp3 is installed.  Another is ensuring that you cannot connect to the pulp2 db.

tests/foreman/api/test_contentview.py
```diff
@@ -1722,3 +1722,33 @@ def test_positive_publish_promote_with_RH_ostree_and_other(
...
+
+        :id: 4b0ebd09-e417-4c42-a437-d11db724e671
+
+        :expectedresults: Cannot log into pulp2 db
+
+        :CaseLevel: System
+
+        :CaseImportance: Critical
+        """
+        pass
```
devendra104, line 1754: pass is not required.

tests/foreman/ui/test_contentview.py
```diff
@@ -4111,3 +4111,33 @@ def test_positive_depsolve_with_module_errata(session, module_org):
             assert items['Name'] in rpm_pack
         assert result['module_streams']['table'][0]['Name'] == mod_stream
         assert result['errata']['table'][0]['Errata ID'] == FAKE_4_ERRATA_ID
+
+    @pytest.mark.stubbed
+    @pytest.mark.tier1
+    def test_positive_check_pulp3_is_installed():
+        """Ensure that pulp3 is installed
+
```
devendra104, line 4119: Do we have anything in UI for pulp package installation verification?

tests/foreman/ui/test_contentview.py
```diff
@@ -4111,3 +4111,33 @@ def test_positive_depsolve_with_module_errata(session, module_org):
...
+        :CaseLevel: System
+
+        :CaseImportance: Critical
+        """
+        pass
+
+    @pytest.mark.stubbed
+    @pytest.mark.tier1
+    def test_positive_check_pulp2_is_removed():
+        """Ensure that pulp2 is removed from system
```
devendra104, line 4133: same as above

tests/foreman/api/test_contentview.py
```diff
@@ -1722,3 +1722,33 @@ def test_positive_publish_promote_with_RH_ostree_and_other(
...
+
+    @pytest.mark.stubbed
+    @pytest.mark.tier1
+    def test_positive_check_pulp2_is_removed(self):
```
devendra104, line 1743: Do we have any API for pulp2 removal package verification?

tests/foreman/ui/test_contentview.py
latran: Services under Capsules provide Pulpcore version and supported content types.  That's the only means I know of to see that it's pulp3.

tests/foreman/ui/test_contentview.py
latran: Same as above.

tests/foreman/api/test_contentview.py
latran: I haven't checked, but if the API hits pulp3, then we know pulp2 is not present.  There's no other real good way to verify this other than try and connect to pulp2 db.

tests/foreman/ui/test_contentview.py
devendra104: I think we are already covering this in the CLI part so nothing extra needs to verify on the UI side right?

tests/foreman/api/test_contentview.py
devendra104: yes, we need to check it, if the API is there then we are good with this just need to update the test steps otherwise we can cover it in the CLI part.

tests/foreman/api/test_contentview.py
swadeley: Hi, I added a check for presence of Pulp3 file system, which would not exists if Pulp3 had not been installed: https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/sys/test_pulp3_filesystem.py

tests/foreman/api/test_contentview.py
```diff
@@ -1722,3 +1722,33 @@ def test_positive_publish_promote_with_RH_ostree_and_other(
...
+    @pytest.mark.tier1
+    def test_positive_check_pulp3_is_installed(self):
+        """Ensure that pulp3 is installed
+
+        :id: 5d5690bf-56c7-4dba-bb09-de0f5f4861b1
+
+        :expectedresults: Pulp3 is active
+
+        :CaseLevel: System
+
```
ColeHiggins2, line 1736: Should we add a CaseAutomation tag since these are stubbed test cases? The same goes for other stubbed tests.
```:CaseAutomation: NotAutomated```

tests/foreman/api/test_contentview.py
mshriver: @swadeley Great! I think that's the perfect place for such a test. 

@latran With that test in place, is there anything actually specific to the content view API that you intend to test here to ensure pulp3 is present? Presence of specific fields in GET type requests, specific search/list parameters, POST parameters?

devendra104:
@latran Can't we move this PR into the draft state?

JacobCallahan:
@latran can you rebase this and see if checks pass?

PR #8691, opened 2021-06-15
Update host power control

mshriver:
Consolidate to a single workflow for power control
Update ContentHost.power_control function behavior

The test case using this function is failing before the call to `ContentHost.power_control` is made. I've otherwise tested the function with `VmState.STOPPED`, `VmState.RUNNING`, and `reboot`. 

`tests/foreman/cli/test_contentview.py::TestContentView::test_positive_remove_cv_version_from_multi_env_capsule_scenario`

### DIFF
--- a/tests/foreman/ui/test_computeresource_gce.py
+++ b/tests/foreman/ui/test_computeresource_gce.py
@@ -32,9 +32,7 @@ from robottelo.helpers import download_gce_cert
 @pytest.mark.tier2
 @pytest.mark.upgrade
 @pytest.mark.skip_if_not_set('http_proxy', 'gce')
-def test_positive_default_end_to_end_with_custom_profile(
-    session, module_org, module_loc, module_gce_settings
-):
+def test_positive_default_end_to_end_with_custom_profile(session, module_org, module_loc):
     """Create GCE compute resource with default properties and apply it's basic functionality.
 
     :id: 59ffd83e-a984-4c22-b91b-cad055b4fbd7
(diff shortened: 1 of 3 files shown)


JacobCallahan (approved):

tstrych:
This is awesome, I knew this functionality is missing in host, now other ever passing tests can be written :stuck_out_tongue:

tests/foreman/ui/test_computeresource_gce.py
```diff
@@ -32,9 +32,7 @@
 @pytest.mark.tier2
 @pytest.mark.upgrade
 @pytest.mark.skip_if_not_set('http_proxy', 'gce')
-def test_positive_default_end_to_end_with_custom_profile(
-    session, module_org, module_loc, module_gce_settings
```
tstrych, line 36: is there a reason why `module_gce_settings` is removed ?

tstrych (commented):
ACK, pending comment.

latran (approved):
ACK, but I need some clarification here.  Is the fix to ensure that we're checking if a host is running and then reboot if it doesn't?

tests/foreman/ui/test_computeresource_gce.py
mshriver: The fixture doesn't exist

mshriver:
@latran right now the workflow on reboot is sensitive to whether or not the VM is running. I could add a step here to ensure that, but I think the workflow should be improved instead. It's been discussed in the MR's for the power control workflow, and @ogajduse  has started to work on those improvements.

The 'fix' is that vm-power-cycle no longer exists, and the power control workflow now takes a `vm_operation` extra arg.

tstrych (approved):

PR #8695, opened 2021-06-16
Stub test case to verify pulp3 logs are captured

shweta83:
Added stub test case to verify Pulp3 Logs are being captured using correlation id.

### DIFF
--- a/tests/foreman/cli/test_logging.py
+++ b/tests/foreman/cli/test_logging.py
@@ -230,3 +230,28 @@ def test_positive_logging_from_dynflow(module_org):
                 assert match, "Request ID not found"
                 logger.info("Request ID found for logging from dynflow ")
     assert POST_line_found, "The POST command to enable a repo was not found in logs."
+
+
+@pytest.mark.tier4
+@pytest.mark.stubbed
+def test_positive_logging_from_pulp3():
+    """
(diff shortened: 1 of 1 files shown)


omkarkhatavkar:

tests/foreman/cli/test_logging.py
```diff
@@ -230,3 +230,29 @@ def test_positive_logging_from_dynflow(module_org):
                 assert match, "Request ID not found"
                 logger.info("Request ID found for logging from dynflow ")
     assert POST_line_found, "The POST command to enable a repo was not found in logs."
+
+
+@pytest.mark.tier4
+@pytest.mark.stubbed
+@pytest.mark.skip_if_open("BZ:1732522")
```
swadeley, line 237: That bug is VERIFIED so we do not want to spend processor time checking if its open.

tests/foreman/cli/test_logging.py
```diff
@@ -230,3 +230,29 @@ def test_positive_logging_from_dynflow(module_org):
...
+def test_positive_logging_verify_pulp3_logs():
```
swadeley, line 238:
```suggestion
def test_positive_logging_from_pulp3():
```
I think to many words with similar meanings

swadeley (approved):
ACK with suggestions

JacobCallahan (approved):

PR #8699, opened 2021-06-17
Fix modified gpgkey init

JacobCallahan:
Elements of this modification should have remained instances of self

### DIFF
--- a/robottelo/config/__init__.py
+++ b/robottelo/config/__init__.py
@@ -133,8 +133,8 @@ def configure_nailgun():
 
     def patched_gpgkey_init(self, server_config=None, **kwargs):
         """Set a default value on the ``content`` field."""
-        gpgkey_init(settings, server_config, **kwargs)
-        settings._fields['content'].default = str(
+        gpgkey_init(self, server_config, **kwargs)
+        self._fields['content'].default = str(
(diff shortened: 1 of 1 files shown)


robottelo/config/__init__.py
```diff
@@ -133,8 +133,8 @@ def configure_nailgun():
 
     def patched_gpgkey_init(self, server_config=None, **kwargs):
         """Set a default value on the ``content`` field."""
-        gpgkey_init(settings, server_config, **kwargs)
-        settings._fields['content'].default = str(
+        gpgkey_init(self, server_config, **kwargs)
+        self._fields['content'].default = str(
```
mshriver, line 137: Shouldn't this be setting `_fields` on the GPGKey entity instance returned by calling `entities.GPGKey.__init__`? The `gpgkey_init()` return needs to be captured and referenced here?

mshriver: Taking a second look, `self` is just _VERY_ misleading in this context.

JacobCallahan: discussed in chat

mshriver: The ugliness of this baby won't be fully addressed here.

mshriver (approved):

PR #8700, opened 2021-06-17
Robottelo_Upgrades upstream image build Action

jyejare:
- Support to build the upgrades_robottelo upstream image to be used by pipelines:
    - Upgrade scenario
    - Upgrade All Tiers
    - FM

### DIFF
--- a/.github/workflows/merge_to_master.yml
+++ b/.github/workflows/merge_to_master.yml
@@ -5,6 +5,8 @@ on:
   push:
     branches:
       - master
+  pull_request:
+    types: [ "opened", "synchronize", "reopened" ]
 
 env:
     PYCURL_SSL_LIBRARY: openssl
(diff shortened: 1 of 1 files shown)


jyejare:
This PR is no more required, we will create two triggers for upgrades_robottelo BuildConfig / ImageStream in OCP to rebuild the downstream image whenever there is change in either robottelo or satellite6-upgrade image in quay.

.github/workflows/merge_to_master.yml
```diff
@@ -107,3 +109,44 @@ jobs:
...
+        with:
+          registry: ${{ secrets.QUAY_SERVER }}
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+
+      - name: Build and push image to Quay
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./Dockerfile
```
mshriver, line 150: This container is already built in this workflow.

The secrets for access to quay are not available in workflows executed against PR's from forks.

jyejare: I am closing this PR as its not required ! I need your ACK on Upgrade build container!

PR #8701, opened 2021-06-18
Remove manual connection for broker-based hosts

JacobCallahan:
Broker 0.1.19 now auto-connects when it needs to.
This means that our host classes no longer need to handle this
connection.

Do not merge until broker 0.1.19 is released!

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -68,12 +68,11 @@ class SatelliteHostError(Exception):
 class ContentHost(Host):
     run = Host.execute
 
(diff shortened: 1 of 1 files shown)


mshriver (approved):

mshriver:
Want to include broker 0.1.20 in this PR?

JacobCallahan:
@mshriver do you mean add 0.1.20 in the requirements? If so, we can just leave that up to the normal process

devendra104 (approved):
ACK

PR #8705, opened 2021-06-21
Small fix for GHA for on_push to 6.9.z

Gauravtalreja1:
Currently, on push to 6.9.z changes are being pushed to `robottelo:latest`, so that's why there is no tag for 6.9.z in quay yet.

### DIFF
--- a/.github/workflows/merge_to_master.yml
+++ b/.github/workflows/merge_to_master.yml
@@ -4,7 +4,7 @@ name: update_robottelo_image
 on:
   push:
     branches:
-      - master
+      - 6.9.z
 
 env:
     PYCURL_SSL_LIBRARY: openssl
@@ -16,6 +16,8 @@ jobs:
     strategy:
       matrix:
         python-version: [3.8, 3.9]
+    env:
+      SATELLITE_VERSION: 6.9
     steps:
       - name: Checkout Robottelo
         uses: actions/checkout@v2


.github/workflows/merge_to_master.yml
```diff
@@ -1,10 +1,7 @@
 # CI stages to execute against master branch on PR merge
 name: update_robottelo_image
 
-on:
-  push:
-    branches:
-      - master
```
Gauravtalreja1, line 7: I guess specifying branch for on push isn't required for master as well, as it should run for all branches.

Please correct me if not or let me know if the same changes should be backported for master?

mshriver: I think we should explicitly list branches that we want to push for - because the pyup.io bot creates branches on the main fork, so I think this would end up running against those branches.

JacobCallahan (changes requested):
Will discuss changes needed for this in chat

.github/workflows/merge_to_master.yml
lpramuk: we need to have `- 6.9.z` instead of `- master`

lpramuk (changes requested):

lpramuk (approved):

mshriver (approved):

PR #8706, opened 2021-06-21
Fix datetime serialization

mshriver:
We're seeing some pipeline failures with datetime instance serialization failures within the xdist workers.

I was unable to recreate the failure locally, it may also be dependent on the RP logging integration being active.

Though I couldn't recreate out of the pipeline, it seemed an appropriate change regardless of whether it is the core pattern responsible for the serialization error.

Double checked for any other tests using datetime for parametrization, didn't find any, but did find some datetime formatting strings that were getting reused multiple times, and created some string constants.

### DIFF
--- a/tests/foreman/api/test_syncplan.py
+++ b/tests/foreman/api/test_syncplan.py
@@ -50,27 +50,24 @@ from robottelo.logging import logger
 from robottelo.utils.issue_handlers import is_open
 
 
-@filtered_datapoint
-def valid_sync_dates():
-    """Returns a list of valid sync dates."""
-    return [
-        # Today
-        datetime.now(),
-        # 5 minutes from now
-        datetime.now() + timedelta(seconds=300),
-        # 5 days from now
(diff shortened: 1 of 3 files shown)


tests/foreman/api/test_syncplan.py
```diff
@@ -50,27 +50,24 @@
...
-        # 5 minutes from now
-        datetime.now() + timedelta(seconds=300),
-        # 5 days from now
-        datetime.now() + timedelta(days=5),
-        # Yesterday
-        datetime.now() - timedelta(days=1),
-        # 5 minutes ago
-        datetime.now() - timedelta(seconds=300),
-    ]
+sync_date_deltas = {
```
mshriver, line 53: This is the core thing that I'm addressing in the PR - moving parametrization from datetime instances to a signed int used for timedelta kwargs within the test cases.

JacobCallahan (approved):

PR #8708, opened 2021-06-22
Automated stubbed test case for logging of pulp3

shweta83:
Added automated test case to verify logs from pulpcore. Pulp3 logs can be verified using correlation id present in production log.

### DIFF
--- a/tests/foreman/cli/test_logging.py
+++ b/tests/foreman/cli/test_logging.py
@@ -24,7 +24,12 @@ from nailgun import entities
 
 from robottelo import manifests
 from robottelo import ssh
+from robottelo.cli.factory import make_product
+from robottelo.cli.factory import make_repository
+from robottelo.cli.product import Product
+from robottelo.cli.repository import Repository
 from robottelo.cli.subscription import Subscription
+from robottelo.constants.repos import FAKE_0_YUM_REPO
 from robottelo.logging import logger
 from robottelo.ssh import upload_file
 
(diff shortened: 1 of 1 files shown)


mshriver (approved):

tests/foreman/cli/test_logging.py
```diff
@@ -233,25 +238,44 @@ def test_positive_logging_from_dynflow(module_org):
...
+    # Get the id of repository sync from task
+    task_out = ssh.command(
+        "hammer task list | grep -F \'Synchronize repository {\"text\"=>\"repository\'"
+    ).stdout[0][:8]
+    prod_log_out = ssh.command(f'grep  {task_out} {source_log}').stdout[0]
+    # Get correlation id of pulp from production logs
+    pulp_correlation_id = re.search(r'\[I\|bac\|\w{8}\]', prod_log_out).group()[7:15]
+    # verify pulp correlation id in message
+    message_log = ssh.command(f'cat {test_logfile} | grep {pulp_correlation_id}').stdout
+    assert message_log
```
Gauravtalreja1, line 281: I guess here `messge_log` should contain the return_code of ssh_command to assert.

latran (approved):
Acking, but pending @Gauravtalreja1 's comment.

tests/foreman/cli/test_logging.py
shweta83: Since I am returning list from ssh.command using stdout, we do not need return code here. It can be done in either ways.

Gauravtalreja1: yes, returning stdout is fine, grepping will either return stdout or not, so that's working for you, but I guess the more appropriate way to test would be to assert with return_code.

mshriver:
@shweta83 Please don't include merge commits in your history, use rebase instead to bring changes to master into your development branches.

tests/foreman/cli/test_logging.py
shweta83: Added the suggested change.

shweta83:
> @shweta83 Please don't include merge commits in your history, use rebase instead to bring changes to master into your development branches.

Sorry for that. Will be cautious in future.

mshriver:
No need to apologize! It'll end up blocking merge though

mshriver (approved):

shweta83:
> No need to apologize! It'll end up blocking merge though

Ack

PR #8713, opened 2021-06-22
added 6.10 stubbed test cases

ColeHiggins2:
Added 6.10 Feature stubbed test case and SCA test case.  

Both are manual test cases and will not be automated.

### DIFF
--- a/tests/foreman/api/test_computeresource_azurerm.py
+++ b/tests/foreman/api/test_computeresource_azurerm.py
@@ -157,7 +157,7 @@ class TestAzureRMComputeResourceTestCase:
 
         :CaseLevel: Acceptance
 
-        :CaseAutomation: NotAutomated
+        :CaseAutomation: ManualOnly
 
         :steps:
             1. Create Azure Compute Resource
@@ -175,6 +175,8 @@ class TestAzureRMComputeResourceTestCase:
             6. US Gov Virginia
 
         :BZ: 1829107
+
+        :customer_scenario: true
         """
 
 
--- a/tests/foreman/api/test_computeresource_azurerm.py
+++ b/tests/foreman/api/test_computeresource_azurerm.py
@@ -146,6 +146,37 @@ class TestAzureRMComputeResourceTestCase:
         portal_nws = azurermclient.list_network()
         assert len(portal_nws) == len(cr_nws['results'])
 
+    @pytest.mark.stubbed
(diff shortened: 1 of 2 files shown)


tests/foreman/api/test_computeresource_azurerm.py
```diff
@@ -146,6 +146,37 @@ def test_positive_check_available_networks(self, azurermclient, module_azurerm_c
...
+    def test_gov_cloud_regions_from_azure_compute_resources(self):
+        """Check Azure Gov Cloud for US Cloud region options when creating a compute resource
+
+        :id: 65e03f7e-a2c7-4541-b92b-38d80ab48175
+
+        :CaseImportance: Medium
+
+        :CaseLevel: Acceptance
+
+        :CaseAutomation: NotAutomated
```
mshriver, line 160: Since there is no plan to automate this coverage, you can use ManualOnly here

tests/foreman/api/test_computeresource_azurerm.py
```diff
@@ -146,6 +146,37 @@ def test_positive_check_available_networks(self, azurermclient, module_azurerm_c
...
+
+        :expectedresults: All regions below should be visible
+            1. US DoD Central
+            2. US DoD East
+            3. US Gov Arizona
+            4. US Gov Iowa
+            5. US Gov Texas
+            6. US Gov Virginia
+
+        :BZ: 1829107
```
mshriver, line 177: This metadata should include `customer_scenario: true`

tests/foreman/api/test_computeresource_azurerm.py
```diff
@@ -146,6 +146,39 @@ def test_positive_check_available_networks(self, azurermclient, module_azurerm_c
...
+
+        :customer_scenario: true
```
devendra104, line 179:
```suggestion
        :customerscenario: true    
```

devendra104 (approved):

tests/foreman/api/test_computeresource_azurerm.py
mshriver: as @devendra104 pointed out the testimony token doesn't actually have an underscore, always check against testimony.yaml or run pre-commit.

latran (approved):
ACK.  In the future, make sure you mark "resolved" for any change request that you've completed.

PR #8723, opened 2021-06-24
Fixed flaw in a UI VMware CR test

ppochop:
The assertion in 'else' branch would always fail.
Can't provide positive test results as this test requires a fix in airgun as well (currently working on that).


omkarkhatavkar:

mshriver:
Looks good, but since you mentioned needing some airgun updates for the test too, I'm going to mark draft for now. We can use the pr-tester job once you've got an airgun PR up.

lhellebr (approved):
Based solely on diff, LGTM

PR #8727, opened 2021-06-25
Rp rerun failed refactor

rplevka:
Reworked plugin for [de]selecting tests based on the appropriate Report Portal launch.

### DIFF
--- a/conf/report_portal.yaml.template
+++ b/conf/report_portal.yaml.template
@@ -7,3 +7,5 @@ REPORT_PORTAL:
   # To skip the rerun, if the failed tests in last run more than fail_threshold
   # if its not set, 20% by default will be considered
   FAIL_THRESHOLD: 0
+  # name of the launch for reporting results to
+  LAUNCH_NAME: launch-name
--- a/pytest_plugins/rerun_rp/rerun_rp.py
+++ b/pytest_plugins/rerun_rp/rerun_rp.py
@@ -1,6 +1,7 @@
 import pytest
 
 from robottelo.config import settings
+from robottelo.hosts import get_sat_version
 from robottelo.logging import logger
 from robottelo.report_portal.portal import ReportPortal
 
@@ -11,63 +12,18 @@ class LaunchError(Exception):
     pass
 
 
-def _transform_rp_tests_to_pytest(tests):
-    """Formating test names toc check 'member of pytest test items' operation
-
-    :param list tests: The list of tests those name to be changed
-    :return list: The transformed list of tests
-    """
-    tests = [str(test).replace('::', '.') for test in tests]
-    return tests
-
-
-def _get_tests(launch, **kwargs):
-    """The helper function that actually connects to Report Portal and fetch tests of a launch
-
-    :param rp.launch launch: The launch object from where the tests will be read
-    :param dict kwargs: The params to `launch.tests` function
-        i.e status, user, defect_types
-    :return list: The list of test names retrieved from RP launch
-    """
-    tests = []
-    if 'status' in kwargs:
-        all_status = kwargs.pop('status')
-        for status in all_status:
-            if status == 'failed' and 'defect_types' in kwargs:
-                defect_types = kwargs.pop('defect_types')
-                for dtype in defect_types:
-                    tests.extend(launch.tests(**kwargs, status=status, defect_type=dtype).keys())
-            else:
-                tests.extend(launch.tests(**kwargs, status=status).keys())
-    else:
-        tests.extend(launch.tests(**kwargs).keys())
-    transformed_tests = _transform_rp_tests_to_pytest(tests)
-    return transformed_tests
-
-
-def _get_test_collection(selectable_tests, items):
-    """Returns the selected and deselected items"""
-    # Select test item if its in failed tests else deselect
-    logger.debug('Selecting/Deselecting tests based on latest launch test results.')
-    selected = []
-    deselected = []
-    for item in items:
-        test_item = f"{item.location[0]}.{item.location[2]}"
-        if test_item in selectable_tests:
-            selected.append(item)
-        else:
-            deselected.append(item)
-    return selected, deselected
-
-
-def _validate_launch(launch, sat_version):
-    """Stop session based on RP Launch statistics and info"""
-    fail_percent = round(int(launch.statistics['failed']) / int(launch.statistics['total']) * 100)
+def _validate_launch(launch):
+    """Stops the pytest session based on RP Launch statistics and info"""
+    fail_percent = round(
+        int(launch['statistics']['executions'].get('failed', 0))
+        / int(launch['statistics']['executions'].get('total'))
+        * 100
+    )
     fail_threshold = settings.report_portal.fail_threshold
     if fail_percent > fail_threshold:
         raise LaunchError(
-            f'The latest launch of Satellite verson {sat_version} has {fail_percent}% tests '
-            f'failed. Which is higher than the threshold of {fail_threshold}%. '
+            f'The reference launch {launch["name"]} (id: {launch["id"]}) has {fail_percent}% '
+            f'of failed tests. It is higher than the threshold of {fail_threshold}%. '
             'Examine the failures thoroughly and check for any major issue.'
         )
 
@@ -75,7 +31,7 @@ def _validate_launch(launch, sat_version):
 def pytest_addoption(parser):
     """Add options for pytest to collect only failed/skipped and user tests"""
     help_text = f'''
-        Collects tests in Report Portal to run only failed status tests.
+        Collects tests from Report Portal to run only failed status tests.
 
         Usage: --only-failed [options]
 
@@ -100,43 +56,75 @@ def pytest_addoption(parser):
     '''
     parser.addoption("--user", nargs='?', help=help_text)
     help_text = '''
-        Rerun Upgrade launch tests instead of Regular Tier launch from Report Portal,
+        Uses given launch as a reference for parsing RP tests to be used with
+        --only-failed, --only-skipped or --user
 
-        Usage: --upgrades-rerun
+        Usage: --rp-reference-launch-uuid [value]
+
+        Value: report_portal_launch_uuid
     '''
-    parser.addoption("--upgrades-rerun", action='store_true', default=False, help=help_text)
+    parser.addoption("--rp-reference-launch-uuid", nargs='?', help=help_text)
 
 
 @pytest.hookimpl(tryfirst=True)
 def pytest_collection_modifyitems(items, config):
     """
-    Collects and modifies tests collection based on pytest options to select tests marked as
-    failed/skipped and user specific tests in Report Portal
+    Collects and modifies test collection based on the pytest options to select the tests marked as
+    failed/skipped and user-specific tests in Report Portal
     """
+    rp_url = settings.report_portal.portal_url or config.getini('rp_endpoint')
+    rp_uuid = config.getini('rp_uuid') or settings.report_portal.api_key
+    # prefer dynaconf setting before ini config as pytest-reportportal plugin uses default value
+    # for `rp_launch` if none is set there
+    rp_launch_name = settings.report_portal.launch_name or config.getini('rp_launch')
+    rp_project = config.getini('rp_project') or settings.report_portal.project
     fail_args = config.getoption('only_failed', False)
     skip_arg = config.getoption('only_skipped', False)
     user_arg = config.getoption('user', False)
-    upgrades_rerun = config.getoption('upgrades_rerun', False)
-    if not any([fail_args, skip_arg, user_arg, upgrades_rerun]):
-        return
-    rp = ReportPortal()
-    version = settings.server.version
-    sat_version = f'{version.base_version}.{version.epoch}'
-    logger.info(f'Fetching Report Portal launches for target Satellite version: {sat_version}')
-    launch = next(
-        iter(
-            rp.launches(
-                sat_version=sat_version, launch_type='upgrades' if upgrades_rerun else 'satellite6'
-            ).values()
-        )
+    ref_launch_uuid = config.getoption('rp_reference_launch_uuid', None) or config.getoption(
+        'rp_rerun_of', None
     )
-    _validate_launch(launch, sat_version)
+    tests = []
+    if not any([fail_args, skip_arg, user_arg]):
+        return
+    rp = ReportPortal(rp_url=rp_url, rp_api_key=rp_uuid, rp_project=rp_project)
+
+    if ref_launch_uuid:
+        logger.info(f'Fetching A reference Report Portal launch {ref_launch_uuid}')
+        ref_launches = rp.get_launches(uuid=ref_launch_uuid)
+        if not ref_launches:
+            raise LaunchError(
+                f'Provided reference launch {ref_launch_uuid} was not found or is not finished'
+            )
+    else:
+        sat_release = get_sat_version().base_version
+        sat_snap = settings.server.version.get('snap', '')
+        if not all([sat_release, sat_snap, (len(sat_release.split('.')) == 3)]):
+            raise pytest.UsageError(
+                '--failed|skipped-only requires a reference launch id or'
+                ' a full satellite version (x.y.z-a.b) to be provided.'
+                f' sat_release: {sat_release}, sat_snap: {sat_snap} were provided instead'
+            )
+        sat_version = f'{sat_release}-{sat_snap}'
+        logger.info(
+            f'Fetching A reference Report Portal launch by Satellite version: {sat_version}'
+        )
+
+        ref_launches = rp.get_launches(name=rp_launch_name, sat_version=sat_version)
+        if not ref_launches:
+            raise LaunchError(
+                f'No suitable Report portal launches for name: {rp_launch_name}'
+                f' and version: {sat_version} found'
+            )
+
     test_args = {}
     test_args.setdefault('status', list())
+    if skip_arg:
+        test_args['status'].append('SKIPPED')
     if fail_args:
-        test_args['status'].append('failed')
+        test_args['status'].append('FAILED')
         if not fail_args == 'all':
-            defect_types = fail_args.split(',') if ',' in fail_args else [fail_args]
+            defect_types = fail_args.split(',')
             allowed_args = [*rp.defect_types.keys()]
             if not set(defect_types).issubset(set(allowed_args)):
                 raise pytest.UsageError(
@@ -144,14 +132,22 @@ def pytest_collection_modifyitems(items, config):
                     f'\'{fail_args}\'. It should be none/one/mix of {allowed_args}'
                 )
             test_args['defect_types'] = defect_types
-    if skip_arg:
-        test_args['status'].append('skipped')
     if user_arg:
         test_args['user'] = user_arg
-    rp_tests = _get_tests(launch, **test_args)
-    selected, deselected = _get_test_collection(rp_tests, items)
+    test_args['paths'] = config.args
+    for ref_launch in ref_launches:
+        _validate_launch(ref_launch)
+        tests.extend(rp.get_tests(launch=ref_launch, **test_args))
+    # remove inapplicable tests from the current test collection
+    deselected = [
+        i
+        for i in items
+        if f'{i.location[0]}.{i.location[2]}'.replace('::', '.')
+        not in [t['name'].replace('::', '.') for t in tests]
+    ]
+    selected = list(set(items) - set(deselected))
     logger.debug(
-        f'Selected {len(selected)} and deselected {len(deselected)} tests based on latest '
+        f'Selected {len(selected)} and deselected {len(deselected)} tests based on latest/given-/ '
         'launch test results.'
     )
     config.hook.pytest_deselected(items=deselected)
--- a/robottelo/report_portal/portal.py
+++ b/robottelo/report_portal/portal.py
@@ -1,5 +1,3 @@
-import re
-
 import requests
 from tenacity import retry
 from tenacity import stop_after_attempt
@@ -9,9 +7,6 @@ from robottelo.config import settings
 from robottelo.logging import logger
 
 
-launch_types = ['satellite6', 'upgrades']
-
-
 class ReportPortal:
     """Represents ReportPortal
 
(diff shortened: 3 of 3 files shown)


robottelo/report_portal/portal.py
```diff
@@ -226,45 +168,71 @@ def _test_params(self, status, defect_type, user):
...
+                headers=self.headers,
+                params=params,
+                verify=False,
+            )
+            resp.raise_for_status()
+            resp_tests.extend(resp.json()['content'])
+            total_pages = resp.json()['page']['totalPages']
+            page += 1
+
+        # Only select tests matching the supplied paths
```
rplevka, line 220: This section should not be really necessary (the final de-select process should take care of unfiltered results too), but I wanted to keep it here as a placeholder for a case, the RP search bug gets resolved, so we can turn this into the API filter parameters and move the heavy lifting to the server side.

robottelo/report_portal/portal.py
```diff
@@ -74,197 +80,119 @@ def _format_launches(self, launches):
...
-        if snap_version:
-            if snap_version not in launches_.keys():
-                raise ValueError(
-                    f'The given snap_version \'{snap_version}\' is not available '
-                    f'in satellite version \'{sat_version}\''
-                )
+        # fixme: we might want to implement pagination eventually
+        if importances is None:
+            importances = self.importance_levels
+        params = {'page.page': 1, 'page.size': 500, 'page.sort': 'startTime,desc'}
```
rplevka, line 107: it's not really likely that we'll get >500 applicable launches, so I haven't implemented the pagination handling

robottelo/report_portal/portal.py
```diff
@@ -9,37 +7,43 @@
...
+        'system_issue': 'si001',
         'zalenium_issue': 'si_rhufdawx03e8',
-        'to_investigate': 'TI001',
-        'no_defect': 'ND001',
+        'to_investigate': 'ti001',
+        'no_defect': 'nd001',
     }
-    statuses = ['failed', 'passed', 'skipped', 'interrupted', 'in_progress']
+    statuses = ['FAILED', 'PASSED', 'SKIPPED', 'INTERRUPTED', 'IN_PROGRESS']
+    importance_levels = ['Low', 'Medium', 'High', 'Critical']
```
pondrejk, line 30: meanwhile `Fips` was added as an importance level

pondrejk (approved):
minor comments

jyejare:
@rplevka Would you wait for https://github.com/reportportal/client-Python/pull/159 to get merged OR shall I review this one and remove this code once upstream(client-Python/159) is merged cause we will use upstream client code directly.

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -100,58 +58,115 @@ def pytest_addoption(parser):
...
     user_arg = config.getoption('user', False)
-    upgrades_rerun = config.getoption('upgrades_rerun', False)
-    if not any([fail_args, skip_arg, user_arg, upgrades_rerun]):
+    sat_version = config.getoption('sat_version', None)
+    ref_launch_uuid = config.getoption('rp-reference-launch-uuid', None) or config.getoption(
+        'rp-rerun-of', None
+    )
+    tests = []
+    # rplevka - fixme
+    # upgrades_rerun = config.getoption('upgrades_rerun', False)
```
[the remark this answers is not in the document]
rplevka: thanks for catching this - removed

robottelo/report_portal/portal.py
```diff
@@ -55,7 +59,7 @@ def headers(self):
         """
         return {'Authorization': f'Bearer {self.rp_api_key}'}
 
-    def _format_launches(self, launches):
+    '''def _format_launches(self, launches):
```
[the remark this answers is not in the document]
rplevka: yeah, we can remove it. thanks

robottelo/report_portal/portal.py
rplevka: Thanks - added.

rplevka:
@jyejare,
I'd like to have this merged until your changes propagate from upstream.

rplevka:
I've slightly updated the order we source the configuration parameters - I now prefer dynaconf settings and then configini - the reason is that upstream plugin uses a default value with `rp_launch_name` and sets it to `Pytest Launch` - making the ini option always set.

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -100,58 +58,108 @@ def pytest_addoption(parser):
...
+        logger.info(f'Fetching A reference Report Portal launch {ref_launch_uuid}')
+        try:
+            ref_launches = rp.get_launches(uuid=ref_launch_uuid)
+        except IndexError:
+            raise pytest.UsageError(
+                f'Provided reference launch {ref_launch_uuid} was not found or is not finished'
+            )
+    else:
+        if not sat_version:
+            sat_release = getenv('SAT_VERSION') or settings.server.version.get('release', '')
```
mshriver, line 115: Why prefer these two (seemingly unique) environment variables, when `settings.server.version.*` values can be provided via dynaconf envvars?

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -100,58 +58,108 @@ def pytest_addoption(parser):
...
     test_args = {}
     test_args.setdefault('status', list())
+    if skip_arg:
+        test_args['status'].append('SKIPPED')
     if fail_args:
-        test_args['status'].append('failed')
+        test_args['status'].append('FAILED')
         if not fail_args == 'all':
-            defect_types = fail_args.split(',') if ',' in fail_args else [fail_args]
+            defect_types = fail_args.split(',')
```
mshriver, line 141: This appears to remove the ability to pass a single option for fail_args?

mshriver (approved):
Couple questions, but this looks like a great change @rplevka! More efficient API use by filtering launches and tests server side.

pytest_plugins/rerun_rp/rerun_rp.py
rplevka: nah, I checked that `split` method handles the absence of the delimiter in the string nicely:

```
In [1]: "single_arg".split(",")                                                                                        
Out[1]: ['single_arg']

```

pytest_plugins/rerun_rp/rerun_rp.py
rplevka: I didn't realize that the dynaconf env vars are a thing, sorry.

mshriver: I'd remove it if you didn't have a specific plan for it. We really want to drive the pattern of using dynaconf settings + envvars to control things at runtime - and to consolidate on reading satellite version from settings. 

@JacobCallahan actually wrote a helper method to enforce a priority of resolving the version of the currently targeted satellite instance:
https://github.com/SatelliteQE/robottelo/blob/fb818ec0ed8f5bef823d107c78b8aed7f18e7140/robottelo/host_info.py#L156

It uses the Satellite class constructor which will resolve settings.server.hostname and get its version property, and if that fails will read `settings.server.version.release`, and then `settings.robottelo.satellite_version`, and then finally looks at `robottelo.constants.SATELLITE_VERSION` if there isn't a target satellite instance and `server.yaml` or `robottelo.yaml` don't have fields set.

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -100,58 +58,108 @@ def pytest_addoption(parser):
...
+    )
+    rp_project = config.getini('rp_project') or settings.report_portal.project
     fail_args = config.getoption('only_failed', False)
     skip_arg = config.getoption('only_skipped', False)
     user_arg = config.getoption('user', False)
-    upgrades_rerun = config.getoption('upgrades_rerun', False)
-    if not any([fail_args, skip_arg, user_arg, upgrades_rerun]):
+    sat_version = config.getoption('sat_version', None)
+    ref_launch_uuid = config.getoption('rp_reference_launch_uuid', None) or config.getoption(
+        'rp_rerun_of', None
```
jyejare, line 98: Where does this option coming of ?

I dont see the option defined anywhere.

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -100,58 +58,108 @@ def pytest_addoption(parser):
...
+    rp_uuid = getenv('RP_UUID') or config.getini('rp_uuid') or settings.report_portal.api_key
+    rp_launch_name = (
+        getenv('RP_LAUNCH') or settings.report_portal.launch_name or config.getini('rp_launch')
+    )
+    rp_project = config.getini('rp_project') or settings.report_portal.project
     fail_args = config.getoption('only_failed', False)
     skip_arg = config.getoption('only_skipped', False)
     user_arg = config.getoption('user', False)
-    upgrades_rerun = config.getoption('upgrades_rerun', False)
-    if not any([fail_args, skip_arg, user_arg, upgrades_rerun]):
```
jyejare, line 120: How are we handling the upgrades rerun ? 

A help message can be updated on how to rerun the upgrade launch tests though.

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -100,58 +58,108 @@ def pytest_addoption(parser):
...
-    logger.info(f'Fetching Report Portal launches for target Satellite version: {sat_version}')
-    launch = next(
-        iter(
-            rp.launches(
-                sat_version=sat_version, launch_type='upgrades' if upgrades_rerun else 'satellite6'
-            ).values()
+    rp = ReportPortal(rp_url=rp_url, rp_uuid=rp_uuid, rp_project=rp_project)
+
+    if ref_launch_uuid:
+        logger.info(f'Fetching A reference Report Portal launch {ref_launch_uuid}')
```
jyejare, line 106:
```suggestion
        logger.info(f'Fetching A reference Report Portal launch from {ref_launch_uuid}')
```

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -100,58 +58,108 @@ def pytest_addoption(parser):
...
+        try:
+            ref_launches = rp.get_launches(uuid=ref_launch_uuid)
+        except IndexError:
+            raise pytest.UsageError(
+                f'Provided reference launch {ref_launch_uuid} was not found or is not finished'
```
jyejare, line 111:
```suggestion
                f'Provided reference launch uuid {ref_launch_uuid} was not found or is not finished'
```

pytest_plugins/rerun_rp/rerun_rp.py
jyejare: Agree with @mshriver here, we really need not to explicitly have our own Env var for satellite version when DynaConf provides that ability.

If other Env vars in this PR are for such a reason let's remove them.

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -100,58 +58,108 @@ def pytest_addoption(parser):
...
+            sat_release = getenv('SAT_VERSION') or settings.server.version.get('release', '')
+            sat_snap = getenv('SAT_SNAP') or settings.server.version.get('snap', '')
+            if not all([sat_release, sat_snap]):
+                raise pytest.UsageError(
+                    '--failed|skipped-only requires a reference launch id or'
+                    ' satellite version to be provided'
+                )
+            sat_version = f'{sat_release}-{sat_snap}'
+        logger.info(
+            f'Fetching A reference Report Portal launch by Satellite' f' version: {sat_version}'
```
jyejare, line 124: I think this should be moved above all along with `ref_launch_uuid` and this whole code can be refactored to fail fast, like:

```suggestion
            if not (ref_launch_uuid or all([sat_release, sat_snap])):
                raise pytest.UsageError(
                    '--failed|skipped-only requires a reference launch id or'
                    ' satellite version to be provided'
                )
            elif ref_launch_uuid:
                # same code that u have now if ref_launch_uuid is provided
            else:
                # the code just if sat_release and sat_snap provided.
                
```

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -100,58 +58,108 @@ def pytest_addoption(parser):
...
-        iter(
-            rp.launches(
-                sat_version=sat_version, launch_type='upgrades' if upgrades_rerun else 'satellite6'
-            ).values()
+    rp = ReportPortal(rp_url=rp_url, rp_uuid=rp_uuid, rp_project=rp_project)
+
+    if ref_launch_uuid:
+        logger.info(f'Fetching A reference Report Portal launch {ref_launch_uuid}')
+        try:
+            ref_launches = rp.get_launches(uuid=ref_launch_uuid)
```
jyejare, line 108: why launches, it should be a single launch ?

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -100,58 +58,108 @@ def pytest_addoption(parser):
...
+        logger.info(
+            f'Fetching A reference Report Portal launch by Satellite' f' version: {sat_version}'
         )
-    )
-    _validate_launch(launch, sat_version)
+
+        ref_launches = rp.get_launches(name=rp_launch_name, sat_version=sat_version)
+    if not ref_launches:
+        raise LaunchError(
+            f'No suitable Report portal launches for name: {rp_launch_name}'
```
jyejare, line 130: Include ref_launch_uuid as well in this message cause we are also calculating the ref_launches there. And this is a combined message for all of them.

robottelo/report_portal/portal.py
```diff
@@ -9,37 +7,43 @@
...
-        'no_defect': 'ND001',
+        'to_investigate': 'ti001',
+        'no_defect': 'nd001',
     }
-    statuses = ['failed', 'passed', 'skipped', 'interrupted', 'in_progress']
+    statuses = ['FAILED', 'PASSED', 'SKIPPED', 'INTERRUPTED', 'IN_PROGRESS']
+    importance_levels = ['Low', 'Medium', 'High', 'Critical', 'Fips']
 
-    def __init__(self):
+    def __init__(self, rp_url, rp_uuid, rp_project):
```
jyejare, line 32:
```suggestion
    def __init__(self, rp_url, rp_api_key, rp_project):
```

This sounds clear!

robottelo/report_portal/portal.py
```diff
@@ -55,216 +59,119 @@ def headers(self):
...
+        just an overview test-item data of every launch in Report Portal
 
         :param str sat_version: The satellite version
             If its not specified, then latest count of launches of satellite versions returned
-        :param str launch_type: Either satellite6 or upgrades,
-            default returns only non-upgrade launches
+        :param bool include_unfinished: Includes launches with status IN_PROGRESS or INTERRUPTED
+        :param list importances: A list of importance levels we want to fetch launches for
+        :param str name: Name of the launch to be filtered
+        :param str uuid: UUID of the launch to be fetched - overrides the other parameters
```
jyejare, line 76:
```suggestion
        :param str uuid: Optional, UUID of the launch to be fetched - overrides the other parameters
```

robottelo/report_portal/portal.py
jyejare: But this might take some toll on performance to fetch all of them in one page / one API request.

Though its not the case for us, cause we are having enough filters to fetch only one. So this count can be reduced here to 50 or even 10.

robottelo/report_portal/portal.py
```diff
@@ -55,216 +59,119 @@ def headers(self):
...
+            'page.size': 50,
+            'page.sort': 'name',
+            'filter.eq.launchId': launch["id"],
+            'filter.ne.type': "SUITE",
+        }
+
+        # parse the test filter parameters and turn them into API filter parameters
+        if test_args is None:
+            test_args = {}
+        if not isinstance(test_args, dict):
```
jyejare, line 139: Instead make test_args a keyword argument.

```
def get_tests(self, launch=None, *test_args):
```

jyejare (commented):
Thanks for the PR @rplevka , this is really a nicer improvement. Also, thanks to the RP5 improvements to their APIs, they have made sorting and fetching the launch and tests with improved and workable parameters(which were not working in RP4) :) 

Requested some changes which might not sound blockers thoug.

robottelo/report_portal/portal.py
```diff
@@ -55,216 +59,119 @@ def headers(self):
...
-        tests_ = {test['name']: test for test in data}
-        return tests_
+
+        # Only select tests matching the supplied paths. This is a workaround for RP API limitation
+        # - unable to combine multiple filters of a same type
+        if test_args.get('paths'):
+            resp_tests = [
+                test
+                for test in resp_tests
+                if any([path for path in test_args['paths'] if path in test['name']])
```
jaryn, line 175: You could have used a generator instead of list comprehention, which would be more lean.

conf/report_portal.yaml.template
```diff
@@ -7,3 +7,5 @@ REPORT_PORTAL:
   # To skip the rerun, if the failed tests in last run more than fail_threshold
   # if its not set, 20% by default will be considered
   FAIL_THRESHOLD: 0
+  # name of the launch for reporting results to
+  LAUNCH_NAME: launch-name
```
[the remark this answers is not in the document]
rplevka: what format do you have in mind? There's no pre-defined or required format for the launch name.

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -11,71 +13,27 @@ class LaunchError(Exception):
...
+        int(launch['statistics']['executions']['failed'])
+        / int(launch['statistics']['executions']['total'])
+        * 100
+    )
     fail_threshold = settings.report_portal.fail_threshold
     if fail_percent > fail_threshold:
         raise LaunchError(
-            f'The latest launch of Satellite verson {sat_version} has {fail_percent}% tests '
-            f'failed. Which is higher than the threshold of {fail_threshold}%. '
+            f'The reference launch has {fail_percent}% '
```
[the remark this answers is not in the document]
rplevka: right, maybe let's go with `uuid` or `id` which references to the specific launch.

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -100,58 +58,108 @@ def pytest_addoption(parser):
...
 @pytest.hookimpl(tryfirst=True)
 def pytest_collection_modifyitems(items, config):
     """
-    Collects and modifies tests collection based on pytest options to select tests marked as
-    failed/skipped and user specific tests in Report Portal
+    Collects and modifies test collection based on the pytest options to select the tests marked as
+    failed/skipped and user-specific tests in Report Portal
     """
+    rp_url = settings.report_portal.portal_url or config.getini('rp_endpoint')
+    rp_uuid = getenv('RP_UUID') or config.getini('rp_uuid') or settings.report_portal.api_key
```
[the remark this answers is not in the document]
rplevka: yes, as discussed with @mshriver , i'm about to drop this and go with dynaconf settings

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -1,3 +1,5 @@
+from os import getenv
```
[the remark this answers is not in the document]
rplevka: This is going away since we're not going to source env vars.

JacobCallahan:
@rplevka @jyejare can you revisit this?

jyejare:
@rplevka Did you make/reject changes as suggested and is it ready for 2nd round review ?

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -100,58 +58,108 @@ def pytest_addoption(parser):
...
     """
-    Collects and modifies tests collection based on pytest options to select tests marked as
-    failed/skipped and user specific tests in Report Portal
+    Collects and modifies test collection based on the pytest options to select the tests marked as
+    failed/skipped and user-specific tests in Report Portal
     """
+    rp_url = settings.report_portal.portal_url or config.getini('rp_endpoint')
+    rp_uuid = getenv('RP_UUID') or config.getini('rp_uuid') or settings.report_portal.api_key
+    rp_launch_name = (
+        getenv('RP_LAUNCH') or settings.report_portal.launch_name or config.getini('rp_launch')
```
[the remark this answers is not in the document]
rplevka: This was already discussed.
The reason for using this order is that the[ upstream lib uses a default value](https://github.com/reportportal/agent-python-pytest/blob/master/pytest_reportportal/plugin.py#L254) for `rp_launch` attribute if this hasn't been set.
This would never allow dynaconf setting to be used.

rplevka:
@jyejare I finally got back to this to address your comments, please review

robottelo/report_portal/portal.py
rplevka: used 20 as a reasonable compromise

pytest_plugins/rerun_rp/rerun_rp.py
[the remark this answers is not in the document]
mshriver: I think its worth commenting this in-line

pytest_plugins/rerun_rp/rerun_rp.py
mshriver: @rplevka looks like this still needs an update. I think you should use `robottelo.host.get_sat_version` (unstable branch location, its in host_info.py in master)

For snap number, we need to make sure we're setting `ROBOTTELO_SERVER__RELEASE__SNAP` in CI pipelines.

I also don't think we need an additional option (`sat_version`) for it on the CLI - just always use get_sat_version and `settings.server.release.snap`.

pytest_plugins/rerun_rp/rerun_rp.py
mshriver: Its an option of Report Portal's pytest agent.

https://github.com/reportportal/agent-python-pytest/blob/40e314c08b6714dcdef1884a113b65865514da17/pytest_reportportal/config.pyi#L19

pytest_plugins/rerun_rp/rerun_rp.py
mshriver: How would upgrade be different here?

pytest_plugins/rerun_rp/rerun_rp.py
```diff
@@ -100,58 +57,109 @@ def pytest_addoption(parser):
...
+        if not sat_version:
+            sat_release = getenv('SAT_VERSION') or settings.server.version.get('release', '')
+            sat_snap = getenv('SAT_SNAP') or settings.server.version.get('snap', '')
+            if not all([sat_release, sat_snap]):
+                raise pytest.UsageError(
+                    '--failed|skipped-only requires a reference launch id or'
+                    ' satellite version to be provided'
+                )
+            sat_version = f'{sat_release}-{sat_snap}'
+        logger.info(
```
mshriver, line 121: Double check your string formatting, looks like this was on multiple lines and then `Black` formatted it. Its two strings instead of one, and there's an `f` prefix for the string that doesn't have any substitutions.

rplevka: Thanks for catching this @mshriver , indeed, this was `Black`

pytest_plugins/rerun_rp/rerun_rp.py
mshriver: @rplevka everything else in this PR looks good, and I was hoping to merge, but I don't want to include use of this env var and include a pattern that we don't want to perpetuate. Bumping to draft state so I know when to re-review.

mshriver (changes requested):
Use of `SAT_VERSION` and `SNAP_VERSION` env vars should be removed.

pytest_plugins/rerun_rp/rerun_rp.py
rplevka: apparently we've got some float to string issue:
```
(Pdb++) from robottelo.host_info import get_sat_version
(Pdb++) get_sat_version
<function get_sat_version at 0x7efd3390ec10>
(Pdb++) get_sat_version()
<Version('6.1')>
```

rplevka: @mshriver testing the code with `robottelo.hosts.get_sat_version` revealed, that since this (plugin) code executes way before `align_to_satellite` fixture does - and `get_sat_version` relies on that while determining the full version from the host inventory.
This fact makes it to fall back sourcing the version from the settings, and if the setting is not set, it falls back to the `SATELLITE_VERSION` constant as a last resort - this last resort is a no-go for us, as it never contains a `.z` bit of the version (for apparent reasons). and it always defaults to `0`.

This would not work for our `z-branch` rerun pipelines.

I suggest strictly passing the version in a form of dynaconf env var in our pipelines, same as we do with the snap version.

rplevka: Updated the code so it works with

```
$ ROBOTTELO_SERVER__VERSION__RELEASE=6.10.0 ROBOTTELO_SERVER__VERSION__SNAP=18.0 pytest  tests/foreman/api/test_operatingsystem.py  --collect-only --only-failed
```
yet, fails with `ROBOTTELO_SERVER__VERSION__RELEASE=6.10`.

mshriver: ack, having `settings.server.version.release set to the full x.y.z string in CI is something we can and should do. We already calculate it based on what we get back from the satellite instance checkout.

In the case of this hook - if it doesn't have a full version, just skip the collection modification. If you want to select only failed cases, you have to be specific about the reference version.

PR #8728, opened 2021-06-25
parse help message, new test search based on the fields

tstrych:
I think with a bit of work test like this can be added to all entities. 
I would like to move parse method somewhere else but not sure where :neutral_face: 

Also firstly I made the test the different way. I asserted all the fields with all hosts but, that can be potentially really time consuming if sat has a lot of hosts. So I decided to add steps 1 and 4 and use only one host.

The `parse_two_columns` method is not necessary, but I though maybe it can be useful in future if we want to check also options from help messages. Let me know what you think about the way I parse this. 

Test result:
```
tests/foreman/cli/test_host.py::test_positive_search_all_field_sets
=========== 1 passed, 55 deselected, 4 warnings in 76.35s (0:01:16) ============
```

### DIFF
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -152,6 +152,161 @@ def function_user(function_host):
     user.delete()
 
 
+# -------------------------- HELP SCENARIOS --------------------------
+def parse_two_columns(content, options_start_with_dash=False):
+    """
+    Parse part of the help message which consist from two columns,
+    for example Options or Search / Order fields
+
+    Example:
+    >>> content = ['activation_key                  string']
+    >>> parse_two_columns(content)
+    >>> {'activation_key': 'string'}
+
+    or:
+    >>> content =  [
+    >>> ' --thin THIN                                   Only list thin and name of hosts',
+    >>> '                                               One of true/false, yes/no, 1/0.'
+    >>> ]
+    >>> parse_two_columns(content, True)
+    >>> {'--thin THIN': ' Only list thin and name of hosts One of true/false, yes/no, 1/0.'}
+
+
+    :param content: list of strings, item is equal to line in help message
+    :param options_start_with_dash: when line is starting with dash and line has only one column
+    and previous_line is not None, then line value is appended to the previous line
+    :return: options dictionary
+    """
+    options = {}
+    previous_line = None
+    for line in content:
+        # strip whitespaces and split for 2 or more whitespaces
+        line_lst = list(filter(None, line.strip().split('  ')))
+        if not line_lst:  # skip empty lines
+            continue
+        if len(line_lst) > 2:
+            raise ValueError(
+                'line is expected to have maximum of 2 fields, parameter and explanation, '
+                f'but has {len(line_lst), line_lst}'
+            )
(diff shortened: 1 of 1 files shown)


JacobCallahan:
I think we should have a larger team conversation about whether or not we want to test this kind of behavior in robottelo.

tstrych:
Two possibilities, merge this as it is already written or close the pull request.

JacobCallahan:
@tstrych did you get a response from dev about their coverage of this?

tstrych:
upstream PR with the test, where you just check if the category name starts with capital letter
https://github.com/theforeman/hammer-cli/pull/342/files#diff-ee9ab7035ec10719ed2f749114a402e46d52f3c1f5cff9f5b10f3358b71abcfe

help test: if the help message looks same 
https://github.com/theforeman/hammer-cli/blob/master/test/functional/help_test.rb

if hammer host list shows column
https://github.com/theforeman/hammer-cli-foreman/blob/master/test/unit/host_test.rb#L26

there is no test which tests the functionality like this one, search all fields dynamically, 
hammer help do not use API call, so it can run without internet connection, so the unit tests have no data for 

I am thinking if this test is bringing more value than mock. 
And how much added value it have, compared to work with some changes. 
Right now:  feeling yes, let's try that. Give it a chance. 
But later on: please remove it please just because I have other things than just fixing the help test.

pondrejk (approved):

jyejare:
@tstrych I agree with @JacobCallahan that it needs a larger discussion on having the `help` strings tests in robottelo. TBH, Robottelo is a pure functional testing framework, we don't worry much about the usability testing here in this framework just as this PR and testing the `help` text.

There is also a localization and internationalization team, might be testing these help texts. But altogether I wont invest in testing the help text at all at least in robottelo.

tstrych:
Understand. Definitely I think, it's not good idea to spread this approach across other entities. But it's already been implemented, why not to give it a chance. And to experiment a bit with this approach. 

But I understand if you don't want to. The reasoning you mentioned makes sense to me. I have no problem if this PR will be closed. Somebody should determine that. I can ask on next joint meeting.

tests/foreman/cli/test_host.py
```diff
@@ -152,6 +152,161 @@ def function_user(function_host):
...
+        try:
+            if options_start_with_dash and line_lst[0].startswith('-'):
+                options[line_lst[0]] = line_lst[1]
+                previous_line = line_lst[0]
+            else:
+                options[line_lst[0]] = line_lst[1]
+        except IndexError:
+            if options_start_with_dash and previous_line is not None:
+                # append message to previous line value
+                options[previous_line] = options[previous_line] + ' ' + line_lst[0]
```
mshriver, line 202: I think f-string would work well here instead of concatenating.

mshriver (approved):
I'm leaving this one up to @JacobCallahan to merge or not, but I'm up for letting @tstrych utilize this test in 6.10 and determine whether its providing value.

JacobCallahan:
We'll merge this one and see how it goes. For the moment this is neither an encouraged or discouraged practice.

JacobCallahan (approved):

PR #8737, opened 2021-06-28
Removing test cases that relies on /var/lib/pulp/

latran:
The `/var/lib/pulp/published/` has been removed due to pulp3.  These test cases that relies on /var/lib/pulp` file path is no longer valid.  So far, there's no "easy" way to get this information anymore.


jameerpathan111 (approved):
Ack

mshriver (approved):

PR #8742, opened 2021-06-29
Initial destructive test refactors

JacobCallahan:
This commit introduces three refactors and some new patterns for
destructive tests.
These are part of the first phase of destructive test refactoring.

### DIFF
--- a/pytest_fixtures/broker.py
+++ b/pytest_fixtures/broker.py
@@ -168,3 +168,10 @@ def registered_hosts(organization_ak_setup, content_hosts):
         vm.register_contenthost(org.label, ak.name)
         assert vm.subscribed
     return content_hosts
+
+
+@pytest.fixture
+def destructive_sat(satellite_latest):
+    """Destructive tests require changing settings.server.hostname for now"""
+    with satellite_latest as sat:
+        yield sat
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -15,6 +15,8 @@ from wrapanapi.entities.vm import VmState
 from robottelo import constants
 from robottelo import ssh
 from robottelo.cli.factory import CLIFactoryError
+from robottelo.config import configure_airgun
+from robottelo.config import configure_nailgun
 from robottelo.config import settings
 from robottelo.helpers import install_katello_ca
 from robottelo.helpers import InstallerCommand
@@ -853,12 +855,19 @@ class Satellite(Capsule):
     def __init__(self, hostname=None, **kwargs):
         hostname = hostname or settings.server.hostname
         super().__init__(hostname=hostname, **kwargs)
-        self._init_nailgun()
-        self._init_cli()
-        self._init_airgun()
+        # create dummy classes for later population
+        self._api = type('api', (), {'_configured': False})
+        self._cli = type('cli', (), {'_configured': False})
+        self._ui_session = None
 
-    def _init_nailgun(self):
+    @property
+    def api(self):
         """Import all nailgun entities and wrap them under self.api"""
+        if not self._api:
+            self._api = type('api', (), {'_configured': False})
+        if self._api._configured:
+            return self._api
+
         from nailgun.config import ServerConfig
         from nailgun.entity_mixins import Entity
 
@@ -878,21 +887,26 @@ class Satellite(Capsule):
             verify=False,
         )
         # add each nailgun entity to self.api, injecting our server config
-        self.api = lambda: None
         for name, obj in entities.__dict__.items():
             try:
                 if Entity in obj.mro():
-                    setattr(self.api, name, inject_config(obj, self.nailgun_cfg))
+                    setattr(self._api, name, inject_config(obj, self.nailgun_cfg))
             except AttributeError:
                 # not everything has an mro method, we don't care about them
                 pass
+        return self._api
 
-    def _init_cli(self):
+    @property
+    def cli(self):
         """Import all robottelo cli entities and wrap them under self.cli"""
+        if not self._cli:
+            self._cli = type('cli', (), {'_configured': False})
+        if self._cli._configured:
+            return self._cli
+
         import importlib
         from robottelo.cli.base import Base
 
-        self.cli = lambda: None
         for file in Path('robottelo/cli/').iterdir():
             if file.suffix == '.py' and not file.name.startswith('_'):
                 cli_module = importlib.import_module(f'robottelo.cli.{file.stem}')
@@ -901,13 +915,18 @@ class Satellite(Capsule):
                         if Base in obj.mro():
                             # set our hostname as a class attribute
                             obj.hostname = self.hostname
-                            setattr(self.cli, name, obj)
+                            setattr(self._cli, name, obj)
                     except AttributeError:
                         # not everything has an mro method, we don't care about them
                         pass
+        return self._cli
 
-    def _init_airgun(self):
+    @property
+    def ui_session(self):
         """Initialize an airgun Session object and store it as self.ui_session"""
+        if self._ui_session:
+            return self._ui_session
+
         from airgun.session import Session
 
         def get_caller():
@@ -917,12 +936,13 @@ class Satellite(Capsule):
                 if frame.function.startswith('test_'):
                     return frame.function
 
-        self.ui_session = Session(
+        self._ui_session = Session(
             session_name=get_caller(),
             user=settings.server.admin_username,
             password=settings.server.admin_password,
             hostname=self.hostname,
         )
+        return self._ui_session
 
     @cached_property
     def version(self):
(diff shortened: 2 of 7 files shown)


JacobCallahan:
trigger: test-robottelo
pytest: tests/foreman/cli/test_ldapauthsource.py -k test_rhsso_two_factor_login_using_hammer

tests/foreman/cli/test_ldapauthsource.py
```diff
@@ -482,7 +486,7 @@ def test_rhsso_login_using_hammer(
 
         :CaseImportance: High
         """
-        result = AuthLogin.oauth(
+        result = destructive_sat.cli.AuthLogin.oauth(
             {
```
jameerpathan111, line 490: @JacobCallahan I get that this will be executed on destructive satellite created using `destructive_sat` but what about the code written in fixtures that this test is using(ex. rhsso_setting_setup)? Won't we need to change that as well, like you have done here.

pytest_fixtures/broker.py
```diff
@@ -168,3 +168,10 @@ def registered_hosts(organization_ak_setup, content_hosts):
         vm.register_contenthost(org.label, ak.name)
         assert vm.subscribed
     return content_hosts
+
+
+@pytest.fixture
+def destructive_sat(satellite_latest):
+    """Destructive tests require changing settings.server.hostname for now"""
+    with satellite_latest as sat:
+        yield sat
```
jameerpathan111, line 177: @JacobCallahan If we use this for each destructive test then that will increase overall test execution time significantly. Are we ok with that?

tests/foreman/cli/test_ldapauthsource.py
JacobCallahan: the destructive_sat fixture uses the Satellite object's context manager to force robottelo to point to the destructive satellite while the test is running.

pytest_fixtures/broker.py
JacobCallahan: we are indeed. there will be different scopes throughout destructive tests. some will be function-scope like this, while others will be module or session scope.

robottelo/hosts.py
```diff
@@ -940,3 +960,18 @@ def capsule_certs_generate(self, capsule, **extra_kwargs):
         install_cmd = InstallerCommand.from_cmd_str(cmd_str=result.stdout)
         install_cmd.opts['certs-tar-file'] = f'/root/{capsule.hostname}-certs.tar'
         return install_cmd
+
+    def __enter__(self):
+        """Satellite objects can be used as a context manager to temporarily force everything
+        to use the Satellite object's hostname.
+        """
+        self.__old_hostname = settings.server.hostname
```
omkarkhatavkar, line 968: What is this __old_hostname ? is that will be created by default based on the properties?  If we are planning to run destructive tests as separate threads do we required to store the existing hostname value? what is the worth

tests/foreman/ui/test_contenthost.py
```diff
@@ -1399,10 +1399,29 @@ def test_content_access_after_stopped_foreman(session, vm, foreman_service_teard
...
     """
-    result = vm.run(f'yum -y install {FAKE_1_CUSTOM_PACKAGE}')
+    sat = foreman_service_teardown
+    org = sat.api.Organization().create()
+    # adding remote_execution_connect_by_ip=Yes at org level
+    sat.api.Parameter(
+        name='remote_execution_connect_by_ip', value='Yes', organization=org.id
+    ).create()
+    lce = sat.api.LifecycleEnvironment(organization=org).create()
+    with sat:  # ensure the context section only uses this satellite
```
omkarkhatavkar, line 1409: if you are using content manager on the sat here, will it modify the hostname in settings to old satellite hostname? will we be not missing the foreman teardown step ? or I'm missing anything here.

robottelo/hosts.py
```diff
@@ -853,12 +855,19 @@ class Satellite(Capsule):
     def __init__(self, hostname=None, **kwargs):
         hostname = hostname or settings.server.hostname
         super().__init__(hostname=hostname, **kwargs)
-        self._init_nailgun()
-        self._init_cli()
-        self._init_airgun()
+        # create dummy classes for later population
+        self._api = type('api', (), {'_configured': False})
+        self._cli = type('cli', (), {'_configured': False})
+        self._ui_session = None
```
omkarkhatavkar, line 861: Nice !

robottelo/hosts.py
JacobCallahan: This is for the context manager to store the current hostname so it can be reset upon exit

tests/foreman/ui/test_contenthost.py
JacobCallahan: the teardown fixture uses the same satellite as a fixture requirement now

robottelo/hosts.py
omkarkhatavkar: Yeah you mentioned in the call that we are not going to run this as separate pytest execution in which it make sense, you might need that in running other tests

JacobCallahan:
trigger: test-robottelo
pytest: tests/foreman/cli/test_ldapauthsource.py -k test_rhsso_two_factor_login_using_hammer

pytest_fixtures/broker.py
mshriver: Why create a wrapper here, simply for the name? `satellite_latest` is yielding already.

tests/foreman/ui/test_contenthost.py
mshriver: I'm thinking the teardown fixture should contain the context manager then? Or are you intentionally trying to show a split between direct `sat.api` chains and helper functions?

mshriver (approved):
Couple questions/comments, looks like some great options to get us rolling on object oriented satellite targets.

robottelo/hosts.py
```diff
@@ -853,12 +855,19 @@ class Satellite(Capsule):
...
 
-    def _init_nailgun(self):
+    @property
+    def api(self):
         """Import all nailgun entities and wrap them under self.api"""
+        if not self._api:
+            self._api = type('api', (), {'_configured': False})
```
jaryn, line 867: You are constructing a type here. It is common to name the types with camel case. Would you mind to make it Api or API? Same goes for other use of `type()` in this PR

robottelo/hosts.py
```diff
@@ -853,12 +855,19 @@ class Satellite(Capsule):
     def __init__(self, hostname=None, **kwargs):
         hostname = hostname or settings.server.hostname
         super().__init__(hostname=hostname, **kwargs)
-        self._init_nailgun()
-        self._init_cli()
-        self._init_airgun()
+        # create dummy classes for later population
+        self._api = type('api', (), {'_configured': False})
```
jaryn, line 859: This is bit wild for my taste. Is this some kind of Singletone or [Borg](https://www.oreilly.com/library/view/python-cookbook/0596001673/ch05s23.html) design pattern? Can you explain, perhaps in the comment in the code, why are there both -- lines 859 and 867? Is this because you want to have the `Satellite._api` available as class attribute as well as an instance attribute `self._api`?

Type is creating a class. Are you willing To make the name CamelCased? `type('API', ...` or `type('Api', ...` please?

jaryn (commented):
Please help others understand the code.

robottelo/hosts.py
JacobCallahan: While it is technically a class, we're not using it as such. Just an object to assign attributes to. With that in mind, I do not want to make it apparent that it is anything other than another attribute.

robottelo/hosts.py
JacobCallahan: these are just handy objects that play slightly nicer, than some other alternatives, with what we need to do for pickle.

both are instance attributes only. The addition of 867 etc is extra redundancy post-pickle since __init__ is not called when an object is brought back from a pickled state. also see above regarding casing.

robottelo/hosts.py
```diff
@@ -901,13 +915,18 @@ def _init_cli(self):
...
                     except AttributeError:
                         # not everything has an mro method, we don't care about them
                         pass
+        return self._cli
 
-    def _init_airgun(self):
+    @property
+    def ui_session(self):
         """Initialize an airgun Session object and store it as self.ui_session"""
+        if self._ui_session:
```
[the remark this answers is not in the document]
JacobCallahan: It certainly could be. I typically try to keep things relatively less nested, but if it is bugging you, I can adjust.

robottelo/hosts.py
jaryn: why won't the classic
```
class ApiBase:
    _configured = False

class Satellite:
    _api = ApiBase()

    @property
    def api():
        ....
```

suffice?

jaryn (changes requested):
@JacobCallahan You had quite important comments about the api and pickle in here. Since the code is much more often read than written, can you please put the information you wrote here to the code, so everyone reading it would understand better what is the idea about it? Or rather make it more understandable? i.e. do it the standard way?

pytest_fixtures/broker.py
JacobCallahan: satellite_latest doesn't yield from the satellite object's context manager.

tests/foreman/ui/test_contenthost.py
JacobCallahan: this pattern shows how you can just wrap a smaller subsection of a test with a Satellite context manager. This could also be useful when using multiple Satellites in the same test, switching between each's context.

robottelo/hosts.py
JacobCallahan: I don't want to have new helper classes at the module level, this could be confusing for people who might try to import and use them directly. They exist solely to be helper objects that allow attribute injection.

JacobCallahan:
@jaryn can you revisit this? If your comments have been addressed, then this is ready to merge. Thanks!

jameerpathan111 (approved):

PR #8743, opened 2021-06-29
Deleting puppet class and puppet module

latran:
There's a few puppet files that needs to be deleted because 6.10 does not support puppet anymore.

Please do not merge this without approval from @vsedmik because the files I deleted falls under his component ownership.


tstrych (approved):
ACK, pure removal.

vsedmik (approved):
ACK, these can be removed right now. Thank you!

tests/foreman/cli/test_puppetmodule.py
```diff
@@ -1,117 +0,0 @@
...
-    :CaseImportance: Critical
-    """
-    return_value = PuppetModule.list({'organization-id': module_setup['org'].id})
-    for i in range(len(return_value)):
-        result = PuppetModule.info({'id': return_value[i]['id']}, output_format='json')
-        assert result['id'] == return_value[i]['id']
-
-
-@pytest.mark.tier1
-@pytest.mark.upgrade
```
devendra104, line 89: I think puppet would be removed for newly installed satellite but if the customer has upgraded satellite then puppet will work as it is, so I would suggest to keep the upgrade marker test for verifying the behavior of puppet after the upgrade.

jameerpathan111: @devendra104 makes sense. Probably we should also go through upgrade marker tests again to have better coverage.

latran: @devendra104 @jameerpathan111 I'll run through them and see what kind of results I get.  If some of the upgrade tests still uses puppet module for publishing, then we're have to readjust as necessary.

devendra104:
@latran Can't we move this PR to draft state, as we depend on the puppet upgrade test results?

latran:
@devendra104 done

tests/foreman/cli/test_puppetmodule.py
tstrych: I still think this can be removed. But I guess one of us needs to prove the correctness of his idea by manually upgrading the satellite

tstrych: @devendra104 ^^

Firstly 2 stubbed test cases can be removed. They are not going to be implemented.
And `api/test_puppetmodule` and `cli/test_puppetmodule` uses `Repository` entity, and repository does not support content_type `puppet` in sat 6.10.  And puppet modules api endpoints are no longer callable. 
So even the content is saved on the drive it doesn't matter, because we have no satellite way to test it.

Upgrade scenarious should be consired in other tests where we test puppet environments and puppet classes.

devendra104: yes, you are right @tstrych after the upgrade no puppet content would exist on the setup so we are good to remove the content-related test cases.

jyejare:
@latran @vsedmik Sorry for less known about the 6.10 puppet removal here, but I think the puppet has made optional in 6.10 and not completely removed! I might be wrong though. But if I am true, can we stop removing these tests even though it's an optional feature but it should be part of a functional test framework like robottelo.

devendra104 (approved):
ACK

vsedmik:
@jyejare puppet is not removed from 6.10, but the content related part is removed because pulp3 does not support it (puppet-modules in repos, CVs). We target to remove that part only and keep the rest (update it with some workaround). I'm not aware of any option that would allow us to use the puppet-modules in 6.10 anymore.

tests/foreman/api/test_puppetmodule.py
```diff
@@ -1,125 +0,0 @@
...
-:TestType: Functional
-
-:CaseImportance: High
-
-:Upstream: No
-"""
-import pytest
-from nailgun import entities
-
-from robottelo.constants import PUPPET_MODULE_NTP_PUPPETLABS
```
pondrejk, line 22: Should the constant be removed as well? Searching for it, it appears also in:

tests/foreman/api/test_hostgroup.py
tests/foreman/api/test_contentviewversion.py
tests/foreman/api/test_contentview.py

... in relation with puppet repositories, do we want to remove also the tests for puppet repos?

tests/foreman/cli/test_puppetmodule.py
```diff
@@ -1,117 +0,0 @@
...
-
-:CaseImportance: High
-
-:Upstream: No
-"""
-import pytest
-
-from robottelo.cli.factory import make_product
-from robottelo.cli.factory import make_repository
-from robottelo.cli.puppetmodule import PuppetModule
```
pondrejk, line 23: Do we want to remove this class? But again, if you search for it, it appears at a number of places, all need to be reviewed

tests/foreman/cli/test_puppetmodule.py
```diff
@@ -1,117 +0,0 @@
...
-from robottelo.cli.repository import Repository
-from robottelo.config import settings
-from robottelo.constants.repos import FAKE_0_PUPPET_REPO
-from robottelo.constants.repos import FAKE_1_PUPPET_REPO
```
pondrejk, line 27: One would imagine we won't need these constants anymore but we need to check other occurrences

pondrejk (changes requested):
It seems like there is more to remove. I'll understand if you want to make it phase 2, just make sure it is not forgotten

tests/foreman/api/test_puppetmodule.py
tstrych: this is just first PR, in the whole EPIC, 
@latran @vsedmik and I started to work on it. 
I see maybe it's time to send out an email about it.
We already identified I hope all 6.10 puppet failing tests which needs to be removed or fixed. 

Definitely nice catch up, maybe the constants would not be needed afterwards.

tests/foreman/cli/test_puppetmodule.py
tstrych: I think I already answered this above.

pondrejk (approved):

PR #8744, opened 2021-06-30
workaround for puppet module for sat 6.10

tstrych:
multiple tests are failing,  this is the first PR and other will follow. Other PR's will use `import_puppet_module` as well. With additional changes to `import_puppet_module` will be required to test also the previous tests. 

api and cli tests were fix
test results:
```
tests/foreman/api/test_host.py::test_positive_end_to_end_with_puppet_class PASSED [ 14%]
tests/foreman/api/test_host.py::test_positive_read_enc_information PASSED     [ 28%]
tests/foreman/api/test_host.py::test_positive_create_inherit_lce_cv PASSED    [ 42%]
tests/foreman/api/test_host.py::test_positive_create_and_update_with_content_view PASSED [ 57%]
tests/foreman/cli/test_host.py::test_positive_update_host_owner_and_verify_puppet_class_name PASSED [ 71%]
tests/foreman/cli/test_host.py::test_positive_create_with_puppet_class_name PASSED [ 85%]
tests/foreman/cli/test_host.py::test_positive_list_scparams PASSED            [100%]
==================== 7 passed, 58 warnings in 313.21s (0:05:13) =====================
```

From what I understand there is no reason to change any of the test mentioned in this PR  to upgrade scenarios.
- Default content view fixture is fixed.
- New method in utils `import_puppet_module` is introduced, there is a lot of hard coded text, so please do not hesitate to comment it, we can improve it ;)  This method should be used only as temporal fix in sat 6.10.

### DIFF
--- a/pytest_fixtures/api_fixtures.py
+++ b/pytest_fixtures/api_fixtures.py
@@ -554,7 +554,8 @@ def module_env():
 
 @pytest.fixture(scope='module')
 def module_env_search(module_org, module_location, module_import_puppet_module):
-    """Search for puppet environment created inside module_import_puppet_module.
+    """Search for puppet environment created from module_import_puppet_module fixture.
+
     Returns the puppet environment with updated organization and location.
     """
     env = (
@@ -562,9 +563,9 @@ def module_env_search(module_org, module_location, module_import_puppet_module):
         .search(query={'search': f'name={module_import_puppet_module}'})[0]
         .read()
     )
-    env.organization.append(module_org)
+    env.organization = [module_org]
     env.update(['organization'])
-    env.location.append(module_location)
+    env.location = [module_location]
     env.update(['location'])
     return env
 
@@ -586,10 +587,7 @@ def module_puppet_classes(module_env_search):
     module_import_puppet_module.
     """
     return entities.PuppetClass().search(
-        query={
-            'search': f'name ~ {"api_test_classparameters"} '
-            f'and environment = {module_env_search.name}'
-        }
+        query={'search': f'name ~ {"generic_1"} ' f'and environment = {module_env_search.name}'}
     )
 
 
--- a/pytest_fixtures/api_fixtures.py
+++ b/pytest_fixtures/api_fixtures.py
@@ -9,8 +9,8 @@ from wrapanapi import GoogleCloudSystem
 
 from robottelo import manifests
 from robottelo.api.utils import enable_rhrepo_and_fetchid
+from robottelo.api.utils import import_puppet_module
 from robottelo.api.utils import promote
-from robottelo.api.utils import publish_puppet_module
 from robottelo.api.utils import upload_manifest
 from robottelo.config import settings
 from robottelo.constants import AZURERM_RHEL7_FT_BYOS_IMG_URN
@@ -31,7 +31,6 @@ from robottelo.constants import REPOS
 from robottelo.constants import REPOSET
 from robottelo.constants import RHEL_6_MAJOR_VERSION
 from robottelo.constants import RHEL_7_MAJOR_VERSION
-from robottelo.constants.repos import CUSTOM_PUPPET_REPO
 from robottelo.helpers import download_gce_cert
 from robottelo.logging import logger
 
@@ -519,7 +518,7 @@ def module_promoted_cv(module_lce, module_published_cv):
 
 @pytest.fixture(scope='module')
 def default_contentview(module_org):
-    return entities.ContentView(organization=module_org, name=DEFAULT_CV).search()
+    return entities.ContentView(organization=module_org, name=DEFAULT_CV).search()[0]
 
 
 @pytest.fixture(scope='module')
@@ -537,15 +536,9 @@ def module_ak_cv_lce(module_org, module_lce, module_published_cv):
 
 @pytest.mark.skipif((not settings.robottelo.repos_hosting_url), reason='Missing repos_hosting_url')
 @pytest.fixture(scope='module')
-def module_cv_with_puppet_module(module_org):
-    """Returns content view entity created by publish_puppet_module with chosen
-    name and author of puppet module, custom puppet repository and organization.
-    """
-    return publish_puppet_module(
-        [{'author': 'robottelo', 'name': 'generic_1'}],
-        CUSTOM_PUPPET_REPO,
-        organization_id=module_org.id,
-    )
+def module_import_puppet_module(default_sat):
+    """Import puppet module"""
+    return import_puppet_module(sat=default_sat)
 
 
 @pytest.fixture(scope='session')
@@ -560,22 +553,17 @@ def module_env():
 
 
 @pytest.fixture(scope='module')
-def module_env_search(module_org, module_location, module_cv_with_puppet_module):
-    """Search for puppet environment according to the following criteria:
-    Content view from module_cv_with_puppet_module and chosen organization.
-
-    Returns the puppet environment with updated location.
+def module_env_search(module_org, module_location, module_import_puppet_module):
+    """Search for puppet environment created inside module_import_puppet_module.
+    Returns the puppet environment with updated organization and location.
     """
     env = (
         entities.Environment()
-        .search(
-            query={
-                'search': f'content_view={module_cv_with_puppet_module.name} '
-                f'and organization_id={module_org.id}'
-            }
-        )[0]
+        .search(query={'search': f'name={module_import_puppet_module}'})[0]
         .read()
     )
+    env.organization.append(module_org)
+    env.update(['organization'])
     env.location.append(module_location)
     env.update(['location'])
     return env
@@ -595,10 +583,13 @@ def module_lce_library(module_org):
 def module_puppet_classes(module_env_search):
     """Returns puppet class based on following criteria:
     Puppet environment from module_env_search and puppet class name. The name was set inside
-    module_cv_with_puppet_module.
+    module_import_puppet_module.
     """
     return entities.PuppetClass().search(
-        query={'search': f'name ~ {"generic_1"} and environment = {module_env_search.name}'}
+        query={
+            'search': f'name ~ {"api_test_classparameters"} '
+            f'and environment = {module_env_search.name}'
+        }
     )
 
 
--- a/pytest_fixtures/api_fixtures.py
+++ b/pytest_fixtures/api_fixtures.py
@@ -9,7 +9,6 @@ from wrapanapi import GoogleCloudSystem
 
 from robottelo import manifests
 from robottelo.api.utils import enable_rhrepo_and_fetchid
-from robottelo.api.utils import import_puppet_module
 from robottelo.api.utils import promote
 from robottelo.api.utils import upload_manifest
 from robottelo.config import settings
@@ -537,8 +536,13 @@ def module_ak_cv_lce(module_org, module_lce, module_published_cv):
 @pytest.mark.skipif((not settings.robottelo.repos_hosting_url), reason='Missing repos_hosting_url')
 @pytest.fixture(scope='module')
 def module_import_puppet_module(default_sat):
-    """Import puppet module"""
-    return import_puppet_module(sat=default_sat)
+    """Returns custom puppet environment name that contains imported puppet module
+    and puppet class name."""
+    puppet_class = 'generic_1'
+    return {
+        'puppet_class': puppet_class,
+        'env': default_sat.create_custom_environment(repo=puppet_class),
+    }
 
 
 @pytest.fixture(scope='session')
@@ -560,13 +564,12 @@ def module_env_search(module_org, module_location, module_import_puppet_module):
     """
     env = (
         entities.Environment()
-        .search(query={'search': f'name={module_import_puppet_module}'})[0]
+        .search(query={'search': f'name={module_import_puppet_module["env"]}'})[0]
         .read()
     )
-    env.organization = [module_org]
-    env.update(['organization'])
     env.location = [module_location]
-    env.update(['location'])
+    env.organization = [module_org]
+    env.update(['location', 'organization'])
     return env
 
 
@@ -581,13 +584,15 @@ def module_lce_library(module_org):
 
 
 @pytest.fixture(scope='module')
-def module_puppet_classes(module_env_search):
+def module_puppet_classes(module_env_search, module_import_puppet_module):
     """Returns puppet class based on following criteria:
-    Puppet environment from module_env_search and puppet class name. The name was set inside
-    module_import_puppet_module.
+    Puppet environment from module_env_search and puppet class name.
     """
     return entities.PuppetClass().search(
-        query={'search': f'name ~ {"generic_1"} ' f'and environment = {module_env_search.name}'}
+        query={
+            'search': f'name ~ {module_import_puppet_module["puppet_class"]} '
+            f'and environment = {module_env_search.name}'
+        }
     )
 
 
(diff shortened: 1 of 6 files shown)


tstrych:
Code quality fails is not connected.
```
ERROR tests/upgrades/test_capsule.py - dynaconf.vendor.box.exceptions.BoxKeyE...
ERROR tests/upgrades/test_classparameter.py - dynaconf.vendor.box.exceptions....
ERROR tests/upgrades/test_client.py - dynaconf.vendor.box.exceptions.BoxKeyEr...
ERROR tests/upgrades/test_errata.py - dynaconf.vendor.box.exceptions.BoxKeyEr...
ERROR tests/upgrades/test_host.py - dynaconf.vendor.box.exceptions.BoxKeyErro...
ERROR tests/upgrades/test_remoteexecution.py - dynaconf.vendor.box.exceptions...
ERROR tests/upgrades/test_repository.py - dynaconf.vendor.box.exceptions.BoxK...
ERROR tests/upgrades/test_subscription.py - dynaconf.vendor.box.exceptions.Bo...
ERROR tests/upgrades/test_virtwho.py - dynaconf.vendor.box.exceptions.BoxKeyE...
ERROR tests/upgrades/test_yum_plugins.py - dynaconf.vendor.box.exceptions.Box...
```

pytest_fixtures/api_fixtures.py
```diff
@@ -537,15 +536,9 @@ def module_ak_cv_lce(module_org, module_lce, module_published_cv):
...
-    """Returns content view entity created by publish_puppet_module with chosen
-    name and author of puppet module, custom puppet repository and organization.
-    """
-    return publish_puppet_module(
-        [{'author': 'robottelo', 'name': 'generic_1'}],
-        CUSTOM_PUPPET_REPO,
-        organization_id=module_org.id,
-    )
+def module_import_puppet_module(default_sat):
+    """Returns the environment name of imported puppet module"""
```
latran, line 540: Return the puppet environment name or is this different?

pytest_fixtures/api_fixtures.py
```diff
@@ -560,22 +553,19 @@ def module_env():
 
 
 @pytest.fixture(scope='module')
-def module_env_search(module_org, module_location, module_cv_with_puppet_module):
+def module_env_search(module_org, module_location, module_import_puppet_module):
     """Search for puppet environment according to the following criteria:
     Content view from module_cv_with_puppet_module and chosen organization.
```
latran, line 558:
```suggestion
    Content view from module_import_puppet_module and chosen organization.
```

tests/foreman/api/test_host.py
```diff
@@ -1068,7 +1066,7 @@ def test_positive_read_enc_information(
     host_enc_parameters = host_enc_info['data']['parameters']
     assert host_enc_parameters['organization'] == module_org.name
     assert host_enc_parameters['location'] == module_location.name
-    assert host_enc_parameters['content_view'] == module_cv_with_puppet_module.name
+    assert host_enc_parameters['content_view'] == default_contentview.name.replace(' ', '_')
```
latran, line 1069: Can you elaborate the need for `.replace(' ', '_')`?  Just need clarification, that's all.

latran (changes requested):

tests/foreman/api/test_host.py
tstrych: enc parameters are saved this way, they use `_` underscore when there is multiple words

It was not problem before because contentview was created  inside `publish_puppet_module` with one word - 'alpha' of length from 6 to 12, but now it's Default Content view  which needts to be transformed to 'Default_Content_view'

robottelo/api/utils.py
```diff
@@ -130,6 +130,31 @@ def publish_puppet_module(puppet_modules, repo_url, organization_id=None):
...
+def import_puppet_module(sat):
+    """Download install and import puppet class. This is workaround for 6.10, there is no longer
+    content view
+
+    :param sat: satellite object
+    :return: puppet environment name,
+        environment name is likely to be searched in next steps of the test
+    """
+    env_name = 'test_environment'
+    class_parameters_package_name = 'robottelo-api_test_classparameters-0.2.0.tar.gz'
```
vsedmik, line 142: Would it possible to pass the module package name (or perhaps list of names) as a parameter?
In different tests we are using different modules (or combination). Searching for `'author': 'robottelo'` I've found we use these:
- generic_1
- generic_2
- api_test_classparameters
- cli_test_classparameters
- ui_test_classparameters

robottelo/api/utils.py
```diff
@@ -130,6 +130,31 @@ def publish_puppet_module(puppet_modules, repo_url, organization_id=None):
     return cv.read()
 
 
+def import_puppet_module(sat):
+    """Download install and import puppet class. This is workaround for 6.10, there is no longer
```
vsedmik, line 134: "Download, install and import puppet module."

vsedmik (changes requested):
Added few comments but looks good!

pytest_fixtures/api_fixtures.py
```diff
@@ -560,22 +553,17 @@ def module_env():
...
 
 @pytest.fixture(scope='module')
-def module_env_search(module_org, module_location, module_cv_with_puppet_module):
-    """Search for puppet environment according to the following criteria:
-    Content view from module_cv_with_puppet_module and chosen organization.
-
-    Returns the puppet environment with updated location.
+def module_env_search(module_org, module_location, module_import_puppet_module):
+    """Search for puppet environment created inside module_import_puppet_module.
+    Returns the puppet environment with updated organization and location.
```
jyejare, line 558:
```suggestion
    """Search for puppet environment created from module_import_puppet_module fixture.
    
    Returns the puppet environment with updated organization and location.
```

Some wording and line between description and return statements!

robottelo/api/utils.py
```diff
@@ -130,6 +130,31 @@ def publish_puppet_module(puppet_modules, repo_url, organization_id=None):
...
+    content view
+
+    :param sat: satellite object
+    :return: puppet environment name,
+        environment name is likely to be searched in next steps of the test
+    """
+    env_name = 'test_environment'
```
jyejare, line 141: Keeping the constant name shouldn't fail this on running on the same satellite. Cant we just use `gen_string` from fauxfactory to have a randomly generated name every time ?

tests/foreman/api/test_host.py
jyejare: Why won't you use the `default_contentview.label` instead?

jyejare (commented):
ACK pending minor comments / questions / suggestions.

robottelo/api/utils.py
tstrych: no need for that, this is not failing, because command `puppet module install` will finish with module already installed
and on the other hand: let's say we installed the `api_class_parameters` first and then we want to install second module `cli_class_parameters`, that also works, cli is appended inside the test_environment, and api is still there.

tstrych: or better said puppet environment can contain multiple different puppet modules and same ones are not installed again, so it doesn't cause any issues

jyejare: @tstrych ACK !

tstrych: I just find out when it should be used. And why we should generate it to reduce test flakiness. 
Example.
You have one test with `api_class_parameters`
Second one with `cli_class_parameters`
Third one again with `api_class_parameters` and you forget to use the proper value when specifying the custom puppet module value, test will still pass because you already have them in `test_environment`. 
This will not be a big issue but we should test what we want to test, reduce the accidental test passes.
So thanks @jyejare . I will use generator.

tstrych:
problem when using different puppet environments, started to chat with devels

robottelo/api/utils.py
```diff
@@ -130,6 +133,38 @@ def publish_puppet_module(puppet_modules, repo_url, organization_id=None):
...
+        f'{custom_puppet_module_repo}',
+    )
+    sat.execute(
+        f'puppet module install {custom_puppet_module_repo} '
+        f'--target-dir /etc/puppetlabs/code/environments/{env_name}/modules/'
+    )
+    smart_proxy = (
+        entities.SmartProxy().search(query={'search': f'name={settings.server.hostname}'})[0].read()
+    )
+    smart_proxy.import_puppetclasses()
```
ekohl, line 164: https://apidocs.theforeman.org/foreman/latest/apidoc/v2/environments/import_puppetclasses.html suggests you can run it only for a single environment, but maybe then the environment must already exist and it's the DB ID. Overall it would be cleaner to only import the environment you touched, to prevent accidentally breaking other tests.

robottelo/api/utils.py
```diff
@@ -130,6 +133,38 @@ def publish_puppet_module(puppet_modules, repo_url, organization_id=None):
...
+    :return: Puppet environment name,
+        environment name is likely to be searched in next steps of the test
+    """
+    if puppet_module_repo not in list(CUSTOM_PUPPET_MODULE_REPOS.values()):
+        raise ValueError(
+            f'Custom puppet module mismatch, actual custom puppet module repo: '
+            f'"{puppet_module_repo}", does not match any available puppet module: '
+            f'"{list(CUSTOM_PUPPET_MODULE_REPOS.values())}"'
+        )
+    env_name = gen_string('alpha')
```
ekohl, line 151: This implies you generate a new environment every time you run this method. Shouldn't it be named `create_puppet_environment`?

tstrych: yes I agree partly, name should be connected to Environment :) 
but we use namings like `entities.Environment().create()`, so in this case better naming can be  `puppet_environment_contains_custom_puppet_repo` or at least `create_custom_environment`.

JacobCallahan:
@tstrych any update on this?

tstrych:
@JacobCallahan yes, yesterday afternoon Tomer respond on the BZ. So ball is now in my hands ;)  Tomorrow I will be talking with @vsedmik about that, you can join us if you have time for that.

tstrych:
```
tests/foreman/api/test_host.py::test_positive_end_to_end_with_puppet_class PASSED      [ 14%]
tests/foreman/api/test_host.py::test_positive_read_enc_information PASSED              [ 28%]
tests/foreman/api/test_host.py::test_positive_create_inherit_lce_cv PASSED             [ 42%]
tests/foreman/api/test_host.py::test_positive_create_and_update_with_content_view PASSED [ 57%]
tests/foreman/cli/test_host.py::test_positive_update_host_owner_and_verify_puppet_class_name PASSED [ 71%]
tests/foreman/cli/test_host.py::test_positive_create_with_puppet_class_name PASSED     [ 85%]
tests/foreman/cli/test_host.py::test_positive_list_scparams PASSED                     [100%]
========================= 7 passed, 66 warnings in 405.68s (0:06:45) =========================
```

tstrych:
problems were resolved by adding `puppet-proxy` and `puppet-ca-proxy` and by changing in `puppet-classes` into `puppet-class-ids`

tstrych:
TODO: address ekohl naming suggestion

robottelo/api/utils.py
```diff
@@ -130,6 +133,38 @@ def publish_puppet_module(puppet_modules, repo_url, organization_id=None):
     return cv.read()
 
 
+def import_puppet_module(sat, puppet_module_repo=CUSTOM_PUPPET_MODULE_REPOS['generic_1']):
```
mshriver, line 136: Every function that takes a satellite, capsule, or content host as its argument, I'll be asking:

Why isn't this a class method on the Satellite class?

pytest_fixtures/api_fixtures.py
```diff
@@ -560,23 +553,19 @@ def module_env():
...
-            query={
-                'search': f'content_view={module_cv_with_puppet_module.name} '
-                f'and organization_id={module_org.id}'
-            }
-        )[0]
+        .search(query={'search': f'name={module_import_puppet_module}'})[0]
         .read()
     )
-    env.location.append(module_location)
+    env.organization = [module_org]
```
mshriver, line 566: Why is this done as separate calls instead of updating both fields at once?

pytest_fixtures/api_fixtures.py
```diff
@@ -595,10 +584,10 @@ def module_lce_library(module_org):
 def module_puppet_classes(module_env_search):
     """Returns puppet class based on following criteria:
     Puppet environment from module_env_search and puppet class name. The name was set inside
-    module_cv_with_puppet_module.
+    module_import_puppet_module.
     """
     return entities.PuppetClass().search(
-        query={'search': f'name ~ {"generic_1"} and environment = {module_env_search.name}'}
+        query={'search': f'name ~ {"generic_1"} ' f'and environment = {module_env_search.name}'}
```
mshriver, line 590: This might have gotten adjusted by one of the pre-commit tools, there's an extra `f` string in here.

Also, `{"generic_1"}` doesn't really make sense being in `{}` within the f-string, as its just a string, there's no need for interpolation.

robottelo/api/utils.py
JacobCallahan: It certainly could be

tests/foreman/api/test_host.py
```diff
@@ -1068,7 +1081,7 @@ def test_positive_read_enc_information(
     host_enc_parameters = host_enc_info['data']['parameters']
     assert host_enc_parameters['organization'] == module_org.name
     assert host_enc_parameters['location'] == module_location.name
-    assert host_enc_parameters['content_view'] == module_cv_with_puppet_module.name
+    assert host_enc_parameters['content_view'] == default_contentview.label
```
latran, line 1084: Is there a `default_contentview.name`?  Keeping up with consistencies, that's all.

tstrych: i change it on purpose here :) 
`host_enc_parameters['content_view'] = Default_Content_View`
`default_contentview.label = Default_Content_View`
`default_contentview.name = Default Content View` 
so just easier to use `label` as e.g `replace(' ', '_')`

robottelo/api/utils.py
tstrych: will move thanks

pytest_fixtures/api_fixtures.py
tstrych: this is forgotten change from testing, and meanwhile formatted by pre-commit, good eye. thanks :) 
sure, no need for f'string before variable module_env_search, brackets needs to stay -> query formatting

pytest_fixtures/api_fixtures.py
tstrych: I don't know :smile: :smile: So I will change it too.

pytest_fixtures/api_fixtures.py
mshriver: If you want brackets in an f-string, you should double them so that they're interpreted as such.

https://docs.python.org/3/library/string.html#format-string-syntax

`If you need to include a brace character in the literal text, it can be escaped by doubling: {{ and }}.`

robottelo/api/utils.py
```diff
@@ -130,6 +133,38 @@ def publish_puppet_module(puppet_modules, repo_url, organization_id=None):
...
+    """Download, install and import puppet module. This is workaround for 6.10, there is no longer
+    content view
```
vsedmik, line 138: "...there is no longer way to import puppet modules into a content view." 
(just nitpicking)

tstrych: now it can be used this way, where puppet_class is for example `generic_1`
`default_sat.create_custom_environment(repo=puppet_class)`

tstrych:
comments were addressed
test_results: 
`========================= 7 passed, 64 warnings in 320.44s (0:05:20) =========================`

jyejare (approved):

latran (approved):
Issues were addressed, acking.

PR #8759, opened 2021-07-09
Remove references to module variables

spusateri:
This PR is meant to resolve unwanted behaviour with dynaconf loading rhsso settings for all tests that import the module, even those that do not explicitly use them. When the settings are stored as variables, they are loaded each time the module is imported, whether they are used in the test or not. To fix this unwanted behaviour, I replaced the reference to the variable with the value it was defined as in the original code.

### DIFF
--- a/robottelo/rhsso_utils.py
+++ b/robottelo/rhsso_utils.py
@@ -168,7 +168,7 @@ def delete_rhsso_group(group_name):
     group_details = get_rhsso_groups_details(group_name)
     run_command(
         cmd=f"{KEY_CLOAK_CLI} delete -r {settings.rhsso.realm} groups/{group_details['id']}",
-        hostnamesettings.rhsso.host_name,
+        hostname=settings.rhsso.host_name,
     )
 
 
@@ -179,7 +179,7 @@ def update_client_configuration(json_content):
     update_cmd = (
         f"{KEY_CLOAK_CLI} update clients/{client_id} -f update_client_info -s enabled=true --merge"
     )
-    run_command(cmd=update_cmd, hostname=settings.rhsso.host_name))
+    run_command(cmd=update_cmd, hostname=settings.rhsso.host_name)
 
 
 def get_oidc_token_endpoint():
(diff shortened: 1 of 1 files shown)


robottelo/rhsso_utils.py
```diff
@@ -50,16 +45,16 @@ def get_rhsso_client_id():
...
             settings.rhsso.host_url.replace('https://', 'http://'),
-            realm,
-            rhsso_user,
-            rhsso_password,
+            settings.rhsso.realm,
+            settings.rhsso.rhsso_user,
+            settings.rhsso.rhsso_password,
         ),
-        hostname=rhsso_host,
+        hostname=str(settings.rhsso.host_name),
```
mshriver, line 52: These `str()` wrappers can be removed, its an artifact of the SettingsNodeWrapper behavior.

robottelo/rhsso_utils.py
```diff
@@ -114,44 +109,44 @@ def create_new_rhsso_user(client_id, username=None):
...
     )
     return RHSSO_NEW_USER
 
 
 def update_rhsso_user(username, group_name=None):
     user_details = get_rhsso_user_details(username)
-    RHSSO_USER_UPDATE["realm"] = f"{realm}"
+    RHSSO_USER_UPDATE["settings.rhsso.realm"] = f"{settings.rhsso.realm}"
```
mshriver, line 131: The string use of `"realm"` for the dict key shouldn't be changed.

Also, single quotes.

mshriver (changes requested):
One required change to dict key reference

robottelo/rhsso_utils.py
spusateri: Added in latest commit

spusateri:
Added requested changes in latest commit

mshriver:
@spusateri check the GHA checks, you've got some pre-commit failures.

ColeHiggins2 (approved):
ACK, pending pre-commit fixes

robottelo/rhsso_utils.py
```diff
@@ -179,7 +179,7 @@ def update_client_configuration(json_content):
     update_cmd = (
         f"{KEY_CLOAK_CLI} update clients/{client_id} -f update_client_info -s enabled=true --merge"
     )
-    run_command(cmd=update_cmd, hostname=settings.rhsso.host_name))
```
jaryn: How comes this was not caught by some linter?

mshriver (approved):

PR #8760, opened 2021-07-09
Remove random interval picking, restore parameter IDs

mshriver:
I think this is the change that should have been made in #8751 instead of regressing the parametrization. 

I didn't catch this behavior when testing the initial parametrization change in #8706 because of the random behavior in the test case. This removes that random behavior, so a single test run now shows what the consistent test result should be.

```
❯ pytest -vx tests/foreman/api/test_syncplan.py::test_positive_update_interval
============================================================================================================ test session starts ============================================================================================================
platform linux -- Python 3.8.11, pytest-6.2.4, py-1.10.0, pluggy-0.13.1 -- /home/mshriver/repos/satqe/robottelo/.robottelo/bin/python3.8
cachedir: .pytest_cache
shared_function enabled - OFF - scope:  - storage: file
rootdir: /home/mshriver/repos/satqe/robottelo, configfile: pyproject.toml
plugins: forked-1.3.0, xdist-2.3.0, services-2.2.1, reportportal-5.0.8, mock-3.6.1, ibutsu-1.16
collected 4 items                                                                                                                                                                                                                           

tests/foreman/api/test_syncplan.py::test_positive_update_interval[hourly] PASSED                                                                                                                                                      [ 25%]
tests/foreman/api/test_syncplan.py::test_positive_update_interval[daily] PASSED                                                                                                                                                       [ 50%]
tests/foreman/api/test_syncplan.py::test_positive_update_interval[weekly] PASSED                                                                                                                                                      [ 75%]
tests/foreman/api/test_syncplan.py::test_positive_update_interval[custom_cron] PASSED
```

PR #8757, opened 2021-07-08
[cherry-pick] Fix valid_sync_interval, no underscore required

swadeley:
Hello

a cherry-pick from https://github.com/SatelliteQE/robottelo/pull/8751

PR #8761, opened 2021-07-09
Datetime parametrization 69z

mshriver:
I think this should go in instead of #8757

Same commit as #8760 

6.9.z branch hasn't included the change to parametrization yet, so the diff is smaller that 8760

### DIFF
--- a/tests/foreman/api/test_syncplan.py
+++ b/tests/foreman/api/test_syncplan.py
@@ -374,10 +374,8 @@ def test_positive_update_interval(module_org, interval):
     if interval == SYNC_INTERVAL['custom']:
         sync_plan.cron_expression = gen_choice(valid_cron_expressions())
     sync_plan = sync_plan.create()
-    # get another random interval and workaround issue #7231
-    new_interval = gen_choice(valid_sync_interval())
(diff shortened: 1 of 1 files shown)


swadeley (approved):
ACK, thank you

tests/foreman/api/test_syncplan.py
```diff
@@ -374,10 +374,8 @@ def test_positive_update_interval(module_org, interval):
     if interval == SYNC_INTERVAL['custom']:
         sync_plan.cron_expression = gen_choice(valid_cron_expressions())
     sync_plan = sync_plan.create()
-    # get another random interval and workaround issue #7231
-    new_interval = gen_choice(valid_sync_interval())
-    while new_interval == interval:
-        new_interval = gen_choice(valid_sync_interval())
+    # get another random interval
```
swadeley, line 377:
```suggestion
    # ensure "new interval" not equal to "interval"
```

swadeley (approved):
ACK with text change

mshriver:
This one is up to @swadeley if we want this PR or #8757

swadeley:
lets use this one, thank you

PR #8762, opened 2021-07-09
Fix valid_sync_interval again

swadeley:
but keep parameter names

### DIFF
--- a/tests/foreman/api/test_syncplan.py
+++ b/tests/foreman/api/test_syncplan.py
@@ -65,7 +65,7 @@ sync_date_deltas = {
 @filtered_datapoint
 def valid_sync_interval():
     """Returns a list of valid sync intervals."""
-    return ['hourly', 'daily', 'weekly', 'custom cron']
+    return {i.replace(' ', '_'): i for i in ['hourly', 'daily', 'weekly', 'custom cron']}
 
 
 def validate_task_status(repo_id, max_tries=6):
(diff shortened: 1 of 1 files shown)


tests/foreman/api/test_syncplan.py
```diff
@@ -367,10 +367,8 @@ def test_positive_update_interval(module_org, interval):
     if interval == SYNC_INTERVAL['custom']:
         sync_plan.cron_expression = gen_choice(valid_cron_expressions())
     sync_plan = sync_plan.create()
-    # get another random interval
-    new_interval = gen_choice(valid_sync_interval())
-    while new_interval == interval:
-        new_interval = gen_choice(valid_sync_interval())
+    # ensure "new interval" not equal to "interval"
+    new_interval = 'hourly' if interval != 'hourly' else 'daily'
```
jyejare, line 371: Do you mean something else here :
```suggestion
    new_interval = 'daily' if interval != 'hourly' else 'hourly'
```

Because when I read it, it sounds weird as `hourly, if the interval is not equal to hourly else daily` ! Thats total opposite to the condition :)

mshriver: no, we want to set a different interval than the one that was parametrized. So set it to 'hourly' if the parametrized `interval` is not hourly. Otherwise set it to 'daily'`.

mshriver (approved):

swadeley:
sorry, I was too late, this was first :

https://github.com/SatelliteQE/robottelo/pull/8760

mshriver:
@swadeley don't close this one! I'd rather you own it :)

tests/foreman/api/test_syncplan.py
swadeley: @jyejare we are just trying to use a simple method to ensure `new_interval` is not the same as `interval` so as not to have to call the `valid_sync_interval()` function which Mike has changed so as to display user friendly values. See https://github.com/SatelliteQE/robottelo/pull/8760#issue-686984814

swadeley:
Thank you

PR #8767, opened 2021-07-12
Add a test for insights remediation feature

jameerpathan111:
This pr depends on: https://github.com/SatelliteQE/airgun/pull/576 
I also need to add `rh_cloud.yaml` file in our gitlab.
Test result:
```
$ pytest -Wignore tests/foreman/ui/test_rhcloud_insights.py 
collected 1 item

tests/foreman/ui/test_rhcloud_insights.py .E                             [100%]

==================================== ERRORS ====================================
________________ ERROR at teardown of test_rhcloud_insights_e2e ________________

    @pytest.fixture
    def unset_set_cloud_token():
        """A function-level fixture to unset and reset rh cloud token value."""
        setting_update('rh_cloud_token', '')
        yield
>       setting_update('rh_cloud_token', settings.rh_cloud.token)

pytest_fixtures/rh_cloud.py:22: 
robottelo/rh_cloud_utils.py:94: in setting_update
    setting.update({'value'})
../nailgun/nailgun/entity_mixins.py:1042: in update
    return self.read(attrs=self.update_json(fields))
../nailgun/nailgun/entity_mixins.py:1018: in update_json
    response = self.update_raw(fields)
../nailgun/nailgun/entity_mixins.py:998: in update_raw
    return client.put(
../nailgun/nailgun/client.py:172: in put
    data = dumps(data)
/usr/local/lib/python3.8/json/__init__.py:231: in dumps
    return _default_encoder.encode(obj)
/usr/local/lib/python3.8/json/encoder.py:199: in encode
    chunks = self.iterencode(o, _one_shot=True)
/usr/local/lib/python3.8/json/encoder.py:257: in iterencode
    return _iterencode(o, 0)

self = <json.encoder.JSONEncoder object at 0x7f28443ad8e0>
o = <SettingsNodeWrapper for "rh_cloud.token" wrapping value of type str: ***trimmed**output**token***value>

    def default(self, o):
        """Implement this method in a subclass such that it returns
        a serializable object for ``o``, or calls the base implementation
        (to raise a ``TypeError``).
    
        For example, to support arbitrary iterators, you could
        implement default like this::
    
            def default(self, o):
                try:
                    iterable = iter(o)
                except TypeError:
                    pass
                else:
                    return list(iterable)
                # Let the base class default method raise the TypeError
                return JSONEncoder.default(self, o)
    
        """
>       raise TypeError(f'Object of type {o.__class__.__name__} '
                        f'is not JSON serializable')
E       TypeError: Object of type str is not JSON serializable

/usr/local/lib/python3.8/json/encoder.py:179: TypeError
ERROR tests/foreman/ui/test_rhcloud_insights.py::test_rhcloud_insights_e2e - ...
==================== 1 passed, 1 error in 752.80s (0:12:32) ====================

```
The test is currently failing at teardown, will fix this issue soon.

### DIFF
--- a/conf/rh_cloud.yaml.template
+++ b/conf/rh_cloud.yaml.template
@@ -1,0 +1,2 @@
+RH_CLOUD:
+  #Token:
--- a/conf/rh_cloud.yaml.template
+++ b/conf/rh_cloud.yaml.template
@@ -1,2 +1,2 @@
 RH_CLOUD:
-  #Token:
+  Token: rh_cloud_token_value
--- a/pytest_fixtures/rh_cloud.py
+++ b/pytest_fixtures/rh_cloud.py
@@ -23,7 +23,7 @@ def unset_set_cloud_token():
     rh_cloud_token_setting = entities.Setting().search(query={'search': 'name="rh_cloud_token"'})[
         0
     ]
-    rh_cloud_token_setting.value = 'invalid_token_value'
+    rh_cloud_token_setting.value = ''
     rh_cloud_token_setting.update({'value'})
     yield
     rh_cloud_token_setting = entities.Setting().search(query={'search': 'name="rh_cloud_token"'})[
@@ -65,5 +65,5 @@ def rhel8_insights_vm(organization_ak_setup, rhel8_contenthost_module):
 def fixable_rhel8_vm(rhel8_insights_vm):
     """A function-level fixture to create dnf related insights recommendation for rhel8 host."""
     rhel8_insights_vm.run('dnf update -y dnf')
-    rhel8_insights_vm.run("sed -i -e '/^best/d' /etc/dnf/dnf.conf")
-    rhel8_insights_vm.run('insights-client --register')
+    rhel8_insights_vm.run('sed -i -e "/^best/d" /etc/dnf/dnf.conf')
+    rhel8_insights_vm.run('insights-client')
--- a/pytest_fixtures/rh_cloud.py
+++ b/pytest_fixtures/rh_cloud.py
@@ -17,6 +17,14 @@ def set_rh_cloud_token():
 
 
 @pytest.fixture(scope='module')
+def enable_lab_features():
+    """A module-level fixture to enable lab features."""
+    setting_update('lab_features', True)
+    yield
+    setting_update('lab_features', False)
+
+
+@pytest.fixture(scope='module')
 def organization_ak_setup(module_manifest_org):
     """A module-level fixture to create an Activation key in module_org"""
     ak = entities.ActivationKey(
--- a/pytest_fixtures/rh_cloud.py
+++ b/pytest_fixtures/rh_cloud.py
@@ -4,7 +4,6 @@ from nailgun import entities
 from robottelo.config import settings
 from robottelo.constants import DEFAULT_SUBSCRIPTION_NAME
 from robottelo.constants import DISTRO_RHEL8
-from robottelo.helpers import add_remote_execution_ssh_key
 
 
 @pytest.fixture
--- a/pytest_fixtures/rh_cloud.py
+++ b/pytest_fixtures/rh_cloud.py
@@ -43,10 +43,10 @@ def organization_ak_setup(module_manifest_org):
 def rhel8_insights_vm(default_sat, organization_ak_setup, rhel8_contenthost_module):
     """A module-level fixture to create rhel8 content host registered with insights."""
     org, ak = organization_ak_setup
+    rhel8_contenthost_module.configure_rex(satellite=default_sat, org=org, register=False)
     rhel8_contenthost_module.configure_rhai_client(
         satellite=default_sat, activation_key=ak.name, org=org.label, rhel_distro=DISTRO_RHEL8
     )
-    add_remote_execution_ssh_key(rhel8_contenthost_module.ip_addr)
     yield rhel8_contenthost_module
 
 
--- a/pytest_fixtures/rh_cloud.py
+++ b/pytest_fixtures/rh_cloud.py
@@ -5,30 +5,21 @@ from robottelo.config import settings
 from robottelo.constants import DEFAULT_SUBSCRIPTION_NAME
 from robottelo.constants import DISTRO_RHEL8
 from robottelo.helpers import add_remote_execution_ssh_key
-from robottelo.rh_cloud_utils import setting_update
 
 
 @pytest.fixture
-def set_rh_cloud_token():
+def set_rh_cloud_token(default_sat):
     """A function-level fixture to set rh cloud token value."""
-    setting_update('rh_cloud_token', settings.rh_cloud.token)
+    default_sat.update_setting('rh_cloud_token', settings.rh_cloud.token)
     yield
-    setting_update('rh_cloud_token', '')
+    default_sat.update_setting('rh_cloud_token', '')
 
 
 @pytest.fixture
-def unset_rh_cloud_token():
+def unset_rh_cloud_token(default_sat):
     """A function-level fixture to unset rh cloud token value."""
     yield
-    setting_update('rh_cloud_token', '')
-
-
-@pytest.fixture(scope='module')
-def enable_lab_features():
-    """A module-level fixture to enable lab features."""
-    setting_update('lab_features', True)
-    yield
-    setting_update('lab_features', False)
+    default_sat.update_setting('rh_cloud_token', '')
 
 
 @pytest.fixture(scope='module')
@@ -49,11 +40,11 @@ def organization_ak_setup(module_manifest_org):
 
 
 @pytest.fixture(scope='module')
-def rhel8_insights_vm(organization_ak_setup, rhel8_contenthost_module):
+def rhel8_insights_vm(default_sat, organization_ak_setup, rhel8_contenthost_module):
     """A module-level fixture to create rhel8 content host registered with insights."""
     org, ak = organization_ak_setup
     rhel8_contenthost_module.configure_rhai_client(
-        activation_key=ak.name, org=org.label, rhel_distro=DISTRO_RHEL8
+        satellite=default_sat, activation_key=ak.name, org=org.label, rhel_distro=DISTRO_RHEL8
     )
     add_remote_execution_ssh_key(rhel8_contenthost_module.ip_addr)
     yield rhel8_contenthost_module
--- a/pytest_fixtures/rh_cloud.py
+++ b/pytest_fixtures/rh_cloud.py
@@ -8,18 +8,12 @@ from robottelo.helpers import add_remote_execution_ssh_key
 from robottelo.rh_cloud_utils import setting_update
 
 
-@pytest.fixture(scope='module')
-def set_rh_cloud_token():
-    """A module-level fixture to set rh cloud token value."""
-    setting_update('rh_cloud_token', settings.rh_cloud.token)
-
-
 @pytest.fixture
-def unset_set_cloud_token():
-    """A function-level fixture to unset and reset rh cloud token value."""
-    setting_update('rh_cloud_token', '')
-    yield
+def set_rh_cloud_token():
+    """A function-level fixture to set rh cloud token value."""
     setting_update('rh_cloud_token', settings.rh_cloud.token)
+    yield
+    setting_update('rh_cloud_token', '')
 
 
 @pytest.fixture(scope='module')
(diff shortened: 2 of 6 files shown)


robottelo/rh_cloud_utils.py
```diff
@@ -100,3 +102,10 @@ def get_report_data(report_path):
             if file_name != 'metadata.json':
                 json_data = json.load(tarobj.extractfile(file_))
     return json_data
+
+
+def setting_update(name, value):
```
mshriver, line 107: I don't think this is a great pattern to have this 'helper' function here. This to me looks like a core `Capsule` instance method, that could reference `self.api.Setting()`

pytest_fixtures/rh_cloud.py
```diff
@@ -1,7 +1,25 @@
...
 from robottelo.constants import DEFAULT_SUBSCRIPTION_NAME
+from robottelo.constants import DISTRO_RHEL8
+from robottelo.helpers import add_remote_execution_ssh_key
+from robottelo.rh_cloud_utils import setting_update
+
+
+@pytest.fixture(scope='module')
+def set_rh_cloud_token():
+    """A module-level fixture to set rh cloud token value."""
+    setting_update('rh_cloud_token', str(settings.rh_cloud.token))
```
mshriver, line 14: You shouldn't have to wrap the token with `str` here, is this a leftover pattern from NodeSettingsWrapper use?

pytest_fixtures/rh_cloud.py
```diff
@@ -1,7 +1,25 @@
...
+
+
+@pytest.fixture
+def unset_set_cloud_token():
```
mshriver, line 18: This is a pretty strange pattern, combined with the module level fixture.  Why not have a single function scoped fixture that sets the value, yields, and then 'unsets' the value. You achieve the same configuration pattern across test cases, without trying to do setup for the next testcase in teardown of the previous.

pytest_fixtures/rh_cloud.py
JacobCallahan: @jameerpathan111 ahhh was this not rebased to a newer version of robottelo? you shouldn't have to deal with the SettingsNodeWrapper

robottelo/rh_cloud_utils.py
jameerpathan111: @mshriver I didn't get it. Can you tell me what exactly I need to change?

pytest_fixtures/rh_cloud.py
jameerpathan111: updated the pr.

pytest_fixtures/rh_cloud.py
jameerpathan111: updated the pr.

jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/ui/test_rhcloud_insights.py
airgun: 576

jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/ui/test_rhcloud_insights.py
airgun: 576

pytest_fixtures/broker.py
```diff
@@ -83,6 +83,13 @@ def rhel8_contenthost():
         yield host
 
 
+@pytest.fixture(scope="module")
+def rhel8_contenthost_module():
+    """A fixture that provides a content host object based on the rhel8 nick"""
+    with VMBroker(nick='rhel8', host_classes={'host': ContentHost}) as host:
```
latran, line 89: Is there a reason to use RHEL8 as oppose to RHEL7? Asking because there's no rhel8 template to use as far as I'm aware.

latran (changes requested):

pytest_fixtures/broker.py
latran: I take that back

jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/ui/test_rhcloud_insights.py
airgun: 576

jameerpathan111:
PRT is not working for me, looks like jenkins is down.

pytest_fixtures/broker.py
jameerpathan111: I choose rhel8 because:
To test the remediation feature properly I need a misconfigured host. Meaning I need host which have insights recommendation which can be remediated from Satellite. There's a way for me to break rhel8 host and get 
recommendation for it. By removing `best` option from `/etc/dnf/dnf.conf` file I will get insights recommendation to fix it(i.e `The dnf installs lower versions of packages when the "best" option is not present in the /etc/dnf/dnf.conf`). And because I know what the recommendation is then I can use that to verify if satellite remediated the issue or not.

tests/foreman/ui/test_rhcloud_insights.py
```diff
@@ -0,0 +1,98 @@
...
+        f'Remote action: Insights remediations for selected issues on {rhel8_insights_vm.hostname}'
+    )
+    with session:
+        session.organization.select(org_name=org.name)
+        session.cloudinsights.save_token_sync_hits(settings.rh_cloud.token)
+        # This is a workaround for BZ#1983575. Remove following two code lines once BZ is fixed.
+        session.browser.refresh()
+        session.cloudinsights.sync_hits()
+
+        timestamp = (datetime.utcnow() - timedelta(minutes=2)).strftime('%Y-%m-%d %H:%M')
```
jaryn, line 70: Is it really necessary to wait 2 minutes to check the job has started 2 minutes ago? I would prefer to check for some job state. Was this discussed with the developers?

tests/foreman/ui/test_rhcloud_insights.py
```diff
@@ -0,0 +1,98 @@
...
+            search_rate=15,
+            max_tries=10,
+        )
+        job_invocation = wait_for_tasks(
+            search_query=f'{job_query} and started_at >= "{timestamp}"',
+            search_rate=15,
+            max_tries=10,
+        )
+        assert job_invocation[0].result == 'success'
+        timestamp = (datetime.utcnow() - timedelta(minutes=2)).strftime('%Y-%m-%d %H:%M')
```
jaryn, line 96: I see same pattern here as above. I think not only the QE would benefit from having some way to check the actual task state. It could save 4 minutes of job run.

tests/foreman/ui/test_rhcloud_insights.py
```diff
@@ -0,0 +1,98 @@
...
+            max_tries=10,
+        )
+        result = session.cloudinsights.search(query)[0]
+        assert result['Hostname'] == rhel8_insights_vm.hostname
+        assert (
+            result['Recommendation'] == 'The dnf installs lower versions of packages when the '
+            '"best" option is not present in the /etc/dnf/dnf.conf'
+        )
+        session.cloudinsights.remediate(query)
+        wait_for_tasks(
```
jaryn, line 85: I see the code first tries to check some task started in time t-2 minutes, then it checks whether some other task started in time t. My question now is: Is this reliable pattern? Sorry but I have doubts. 

I would recommend recommend to work on not checking the task being started at specific time.

jaryn (changes requested):

tests/foreman/ui/test_rhcloud_insights.py
jameerpathan111: @jaryn nice catch. It's not required I had added it by mistake. Updated the pr.

tests/foreman/ui/test_rhcloud_insights.py
jameerpathan111: @jaryn timestamp is for searching task started after that time. I had added t-2 minutes by mistake.
It's reliable because along with timestamp, I am using job_query, which is combination of name of task and hostname of the client for which it's been executed.

jameerpathan111: The second task is just for assertion purpose. First one will wait for task to finish, second one is to get/read status of the task i.e if the task finished successfully because first one doesn't provide the right data. Maybe I should find another way to get that info.

jaryn: The test now checks the same thing twice (from line 84 until line 93). That is unnecessary. Please fix

Then on the line 94 it checks the result is success. I don't see any waiting for successful result. Only for job start.
Same goes on line 104. I see there only single refresh is done and then there is an assert for some state. Shouldn't there be some waiting for reaching the state being done?

jaryn (changes requested):

JacobCallahan:
@jameerpathan111 can you revisit this?

jameerpathan111:
@JacobCallahan I will try to work on it this week, moving pr to draft state for now.

JacobCallahan:
> @JacobCallahan I will try to work on it this week, moving pr to draft state for now.

@jameerpathan111 any luck?

jameerpathan111:
@JacobCallahan I haven't look into it yet, will try to fix issues in this sprint.

conf/rh_cloud.yaml.template
```diff
@@ -0,0 +1,2 @@
+RH_CLOUD:
```
mshriver, line 1: Please add a basic validator for this file. I'm thinking TOKEN itself shouldn't be commented out, but should just have an empty value.

https://github.com/SatelliteQE/robottelo/blob/master/robottelo/config/validators.py

pytest_fixtures/broker.py
mshriver: @latran we do have rhel8 templates and nick available internally.

robottelo/rh_cloud_utils.py
mshriver: So, you have a function that's using `entities`, which is assuming the nailgun ServerConfig based on reading settings.server.hostname - implicit behavior that we're trying to remove from robottelo.

Updating a setting is something that is likely to be a common action in the test session, and this entity is used about a dozen times across the repo.

Instead of creating this helper method with very small scope and functionality where no one else is likely to find it, for use within your test module for rh_cloud, invites this function to be re-defined multiple places as people want to make settings updates.

Instead, you could make this a method on the `robottelo.hosts.Satellite` class, so that it is visible, discoverable, and re-usable across the framework.  Its implementation will also be more explicit, as it should be able to reference `self.api.Setting()` to get the settings instance, instead of using the implicit ServerConfig when entities classes are imported and referenced.

Then within tests, anyone can reference `default_sat.update_setting(name=<key>, value=<something>)` without having to import anything, or make implicit assumptions about what satellite is getting its settings updated.

jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/ui/test_rhcloud_insights.py

jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/ui/test_rhcloud_insights.py

jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/ui/test_rhcloud_insights.py

jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/ui/test_rhcloud_insights.py

jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/ui/test_rhcloud_insights.py

jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/ui/test_rhcloud_insights.py

jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/ui/test_rhcloud_insights.py

tests/foreman/ui/test_rhcloud_insights.py
jameerpathan111: updated the pr.

jameerpathan111:
Test result:
```
00:52:49  + py.test tests/foreman/ui/test_rhcloud_insights.py --junit-xml=sat-results.xml -o junit_suite_name=sat-result
00:52:57  ============================= test session starts ==============================
00:52:57  collected 1 item
00:52:57  
01:06:58  tests/foreman/ui/test_rhcloud_insights.py .                              [100%]
01:06:58  ================== 1 passed, 4 warnings in 828.63s (0:13:48) ===================
```
See PRT job 267 for more info.

conf/rh_cloud.yaml.template
jameerpathan111: updated the pr.

tests/foreman/ui/test_rhcloud_insights.py
```diff
@@ -0,0 +1,100 @@
...
+        session.cloudinsights.remediate(query)
+        result = wait_for_tasks(
+            search_query=f'{job_query} and started_at >= "{timestamp}"',
+            search_rate=15,
+            max_tries=10,
+        )
+        task_output = entities.ForemanTask().search(query={'search': result[0].id})
+        try:
+            assert task_output[0].result == 'success'
+        except AssertionError:
```
mshriver, line 89: This pattern is unnecessary, you can pass a string as a second argument to the `assert` statement to define the message that is shown when the assert fails.

```
assert task_output[0].result == 'success', f'result: {result}\n task_output: {task_output}'
```

jameerpathan111: updated the pr.

mshriver (approved):

jameerpathan111:
Thank you :)

PR #8774, opened 2021-07-15
Added ansible version parameter in upgrade template conf file

devendra104:


conf/upgrade.yaml.template
```diff
@@ -22,3 +22,5 @@ UPGRADE:
   # The upgrade VLAN vm_domain
   VM_DOMAIN:
   DISTRIBUTION:
+  # Ansible repo version
+  ANSIBLE_REPO_VERSION: "2.9"
```
jyejare, line 26: We also need to have this update in Gitlab `openshift/confs/upgrade.yaml` since we are copying gitlab confs in Dockerfile to build the robottelo container and not this one. The change here is purely for manually testing the robottelo tests.

I dont see that MR yet on Gitlab/satelliteqe-jenkins ?

jyejare (changes requested):
ACK pending Gitlab MR.

conf/upgrade.yaml.template
mshriver: It's already there, it was added to satellite6-upgrade repository, but because of dynaconf's pre_load path is relative, the template file here is read in during the GHA pipeline.

mshriver (approved):

conf/upgrade.yaml.template
devendra104: It is already there.

mshriver (approved):

JacobCallahan (approved):

PR #8780, opened 2021-07-19
more post-rename checks

### LINKED ISSUE #8578: additional post-rename checks
Find and add mores assertions for checking the hostname was correctly changed by satellite-change-hostname (e.g. grepping configuration files, ...) Some inspiration may be found in the discussion of cust. case 02848267

pondrejk:
fixes #8578

### DIFF
--- a/tests/foreman/sys/test_rename.py
+++ b/tests/foreman/sys/test_rename.py
@@ -137,6 +137,10 @@ class TestRenameHost:
                 new_hostname in result.stdout['published-at']
             ), 'repository published path not updated correctly'
 
+            # check for any other occurences of old hostname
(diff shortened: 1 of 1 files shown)


swadeley (approved):
Hi, I was a bit surprised to see '-r' at the end of the line rather than just after 'grep', but if it works, then OK. (I did check it).

jameerpathan111 (approved):

jyejare (approved):

jyejare:
@pondrejk That's a sane way to check the remaining pieces of the last hostname! But are you sure we dont store the history of changed hostnames somewhere? Else this test will always fail.

Not a blocker comment though! Merging now!

PR #8787, opened 2021-07-23
Proxy port fixture

shweta83:


pytest_fixtures/sys_fixtures.py
```diff
@@ -21,3 +23,12 @@ def relax_bfa(default_sat):
     """Relax BFA protection against failed login attempts"""
     if default_sat:
         default_sat.cli.Settings.set({'name': 'failed_login_attempts_limit', 'value': '0'})
+
+
+@pytest.fixture(autouse=True, scope='session')
+def proxy_port_range(default_sat):
+    """Assigns port range for fake_capsules"""
+    port_pool_range = settings.fake_capsules.port_range
+    port_range = port_pool_range.replace(", ", "-")
```
omkarkhatavkar, line 32: sing quote would be nicer here

pytest_fixtures/sys_fixtures.py
```diff
@@ -21,3 +23,12 @@ def relax_bfa(default_sat):
...
+    if default_sat.execute(f'semanage port -l | grep {str(port_range)}').status != 0:
```
omkarkhatavkar, line 33: I think no need to convert to `str` here ?

omkarkhatavkar (approved):
ACK, pending query

pytest_fixtures/sys_fixtures.py
shweta83: yes, no need for str here. It was left by mistake. I will update it

pytest_fixtures/sys_fixtures.py
shweta83: yes. done

omkarkhatavkar (approved):
ACK

pytest_fixtures/sys_fixtures.py
```diff
@@ -21,3 +23,12 @@ def relax_bfa(default_sat):
     """Relax BFA protection against failed login attempts"""
     if default_sat:
         default_sat.cli.Settings.set({'name': 'failed_login_attempts_limit', 'value': '0'})
+
+
+@pytest.fixture(autouse=True, scope='session')
+def proxy_port_range(default_sat):
+    """Assigns port range for fake_capsules"""
+    port_pool_range = settings.fake_capsules.port_range
+    port_range = port_pool_range.replace(', ', '-')
```
devendra104, line 32: Can anyone let me why we keep the port range like `9091, 14999` instead of `9091-14999`?

rplevka: @devendra104 
I believe there were historical reasons - mainly the way the original `robottelo.properties` ini config parser worked - a config value in CSV format was automatically parsed into a list by the INI reader, while the "-"-separated value was parsed as a string.
This is no longer used so we can use whatever format we find fit.

shweta83: @devendra104  @rplevka  So can we update the ports range format in "fake_capsules.yaml"?

mshriver: @shweta83 absolutely - please provide an update to the template here:
https://github.com/SatelliteQE/robottelo/blob/master/conf/fake_capsules.yaml.template#L6

And then make the update to our internal repo to match.

mshriver: Feel free to consider options to how you define this - if you want to separately define `start` and `end` values, or supply them as a string with a dash.

devendra104: Thanks, @rplevka for the detailed information.
@shweta83 I think the port range format would be much appropriate here but it's up to you, I am ok with both.

mshriver:
@shweta83 please squash the merge commit, I'd encourage you to use `fetch` + `rebase` instead of `pull` in your workflow.

Here's an alias I use to make it easy to update my development branches with commits from the upstream master branch.

```
[alias]
    update = !git fetch upstream && git rebase upstream/master 
```

shweta83:
> @shweta83 please squash the merge commit, I'd encourage you to use `fetch` + `rebase` instead of `pull` in your workflow.
> 
> Here's an alias I use to make it easy to update my development branches with commits from the upstream master branch.
> 
> ```
> [alias]
>     update = !git fetch upstream && git rebase upstream/master 
> ```
Thanks Mike :). I use `fetch` + `rebase` only.  I squashed commits though.

mshriver (approved):
@shweta83 please cherry-pick this commit into a branch for 6.9.z, it needs this test fix as well.

pytest_fixtures/sys_fixtures.py
```diff
@@ -21,3 +23,11 @@ def relax_bfa(default_sat):
     """Relax BFA protection against failed login attempts"""
     if default_sat:
         default_sat.cli.Settings.set({'name': 'failed_login_attempts_limit', 'value': '0'})
+
+
+@pytest.fixture(autouse=True, scope='session')
+def proxy_port_range(default_sat):
+    """Assigns port range for fake_capsules"""
+    port_pool_range = settings.fake_capsules.port_range
```
mshriver, line 31: The problem that you're running into with unit tests is that `default_sat` is not actually provisioned, there's no target to run on.

That's why the other session scoped autouse fixtures first check `if default_sat` before executing anything.

shweta83: Thanks Mike

mshriver (approved):

PR #8788, opened 2021-07-26
ansible collection repo test

pondrejk:
```
pytest tests/foreman/ui/test_repository.py -k test_positive_sync_ansible_collection_gallaxy_repo
================================================ test session starts ================================================
platform linux -- Python 3.9.6, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
plugins: forked-1.3.0, services-2.2.1, reportportal-5.0.8, mock-3.6.1, cov-2.12.1, xdist-2.3.0, ibutsu-1.16
collected 29 items / 28 deselected / 1 selected                                                                     

tests/foreman/ui/test_repository.py .                                                                         [100%]

============================= 1 passed, 28 deselected, 10 warnings in 80.99s (0:01:20) ==============================
```

### DIFF
--- a/robottelo/constants/__init__.py
+++ b/robottelo/constants/__init__.py
@@ -310,7 +310,13 @@ NOT_IMPLEMENTED = 'This is a Manual test!'
 
 SYNC_INTERVAL = {'hour': "hourly", 'day': "daily", 'week': "weekly", 'custom': "custom cron"}
 
-REPO_TYPE = {'yum': "yum", 'puppet': "puppet", 'docker': "docker", 'ostree': "ostree"}
+REPO_TYPE = {
+    'yum': "yum",
+    'puppet': "puppet",
+    'docker': "docker",
+    'ostree': "ostree",
+    "ansible_collection": "ansible collection",
+}
 
 DOWNLOAD_POLICIES = {
     'on_demand': "On Demand",
--- a/robottelo/constants/repos.py
+++ b/robottelo/constants/repos.py
@@ -52,3 +52,6 @@ FAKE_0_YUM_REPO_STRING_BASED_VERSIONS = (
     'https://fixtures.pulpproject.org/rpm-string-version-updateinfo/'
 )
 EPEL_REPO = f'{REPOS_URL}/epel_repo'
+
+ANSIBLE_GALAXY = 'https://galaxy.ansible.com/'
+ANSIBLE_HUB = 'https://cloud.redhat.com/api/automation-hub/'
--- a/tests/foreman/ui/test_repository.py
+++ b/tests/foreman/ui/test_repository.py
@@ -36,6 +36,7 @@ from robottelo.constants import REPO_TYPE
 from robottelo.constants import REPOSET
 from robottelo.constants import VALID_GPG_KEY_BETA_FILE
 from robottelo.constants import VALID_GPG_KEY_FILE
+from robottelo.constants.repos import ANSIBLE_GALAXY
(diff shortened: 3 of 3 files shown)


tests/foreman/ui/test_repository.py
```diff
@@ -740,6 +742,41 @@ def test_positive_end_to_end_custom_ostree_crud(session, module_prod):
...
+
+    :expectedresults: All content synced successfully
+
+    :CaseLevel: Integration
+
+    :CaseImportance: High
+
+    """
+    repo_name = f'gallaxy-{gen_string("alpha")}'
+    requirements = '''
```
jaryn, line 758: Note: You can use [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) to remove the leading whitespaces

pondrejk: thanks, but I don't think fauxfactory outputs spaces in alpha strings, it is just random ascii characters https://github.com/omaciel/fauxfactory/blob/c5d139ccb36d933f064576a1d0ea8832317a6f75/fauxfactory/factories/strings.py#L107

robottelo/constants/__init__.py
```diff
@@ -310,7 +310,13 @@
 
 SYNC_INTERVAL = {'hour': "hourly", 'day': "daily", 'week': "weekly", 'custom': "custom cron"}
 
-REPO_TYPE = {'yum': "yum", 'puppet': "puppet", 'docker': "docker", 'ostree': "ostree"}
+REPO_TYPE = {
+    'yum': "yum",
+    'puppet': "puppet",
```
tstrych, line 315: Note: puppet repo type is no longer supported in 6.10, but I guess this will be part of other PR

tests/foreman/ui/test_repository.py
```diff
@@ -740,6 +741,41 @@ def test_positive_end_to_end_custom_ostree_crud(session, module_prod):
...
+    ---
+    collections:
```
tstrych, line 759: do we want to have such constant in  [robottelo/constants/__init__.py](https://github.com/SatelliteQE/robottelo/blob/master/robottelo/constants/__init__.py) ?

tstrych (approved):
ACK

robottelo/constants/repos.py
```diff
@@ -52,3 +52,6 @@
     'https://fixtures.pulpproject.org/rpm-string-version-updateinfo/'
 )
 EPEL_REPO = f'{REPOS_URL}/epel_repo'
+
+ANSIBLE_GALAXY = 'https://galaxy.ansible.com/'
```
mshriver, line 56: I think it would be better to use these as Dynaconf validator defaults instead of constants. That way its actually configurable, but doesn't have to be set in the yaml.

tests/foreman/ui/test_repository.py
mshriver: If anything I'd recommend storing this as a file in a `data` directory and read it into a string in the test module. Considering this non-blocking.

mshriver (approved):

robottelo/constants/repos.py
JacobCallahan: That's an interesting case. How often likely is it that we would be able to deviate from this value?

mshriver: Well, we will have our own galaxy mirror via pulp before too long. We could have more control over what content is available to test against.

Thoughts/Notes for future updates, not something that needs to be addressed now.

PR #8791, opened 2021-07-26
Remove tox dependency and configuration

mshriver:
Its no longer used in GHA and is not necessary.


JacobCallahan (approved):

requirements-optional.txt
```diff
@@ -3,7 +3,6 @@ codecov==2.1.11
 flake8==3.9.2
 pytest-cov==2.12.1
 redis==3.5.3
-tox==3.23.1
```
Gauravtalreja1, line 6: Pls, remove one remaining `tox` occurrence which is in `.gitignore`.

And I believe this should be backported/removed to the 6.9.z branch.

JacobCallahan: Removing tox references from .gitignore could cause local config options to leak into the repo. It should be fine to keep the ignore sections in place for now.

PR #8769, opened 2021-07-13
Fixes due to changes in hosts.py

swadeley:
Hello

some fixes due to changes in hosts.py [1]

[1] https://github.com/SatelliteQE/robottelo/pull/8665

PR #8669, opened 2021-06-03
Convert ui.test_activationkey to VMBroker

mshriver:
Mostly using existing features, used VMBroker directly for multi-count
test coverage

Also adjusted the imports

Test results are consistent with master branch - there are test cases failing because of non-interactable elements in search.

PR #8665, opened 2021-06-02
Removing VirtualMachine class and methods

mshriver:
Remove the `VirtualMachine` and `CapsuleVirtualMachine` classes/modules, moving any important functions into `robottelo.host` classes.

Move robottelo.cli.factory.setup_virtual_machine into ContentHost class

Create global capsule_configured fixture

Add ContentHost.power_control function, for use in tests that need to
power off/on/cycle hosts. Single test updated to use the new methods,
but the power_off/on workflows DNE yet, so the test will fail.

Add robottelo.hosts.Capsule.nailgun_capsule property, acts like
Host.nailgun_host. Update relevant uses.

Flatten imports in test_contentview and api/test_contentmanagement

Move robottelo.cli.factory.setup_virtual_machine into ContentHost class

PR #8662, opened 2021-06-01
Remove everything relating to old ini-based config

JacobCallahan:
This is an early draft. Still plenty of work to do and the final commit message will actually be useful. But for now, here is the progress so far.

PR #8807, opened 2021-07-28
Backport dynaconf and broker consolidation

mshriver:
Backporting commits from:

#8662 
#8769 
#8669 
#8665 

There are only a few small changes pulled from most of these, 8769 had not yet been backported at all.  Due to some pipeline behavior I wanted to double check the content from the above PRs.


swadeley (approved):
ACK

JacobCallahan (approved):

PR #8812, opened 2021-07-30
install cockpit on default_sat in robottelo

tstrych:
just pushing draft, so after vacation I can continue with written TODO's

### DIFF
--- a/pytest_fixtures/sys_fixtures.py
+++ b/pytest_fixtures/sys_fixtures.py
@@ -1,6 +1,7 @@
 import pytest
 
 from robottelo.config import settings
+from robottelo.hosts import SatelliteHostError
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


pytest_fixtures/sys_fixtures.py
```diff
@@ -32,3 +33,39 @@ def proxy_port_range(default_sat):
...
+def install_cockpit_plugin(default_sat, register_to_dogfood):
+    cmd_result = default_sat.execute(
+        'foreman-maintain packages install -y tfm-rubygem-foreman_remote_execution-cockpit'
+    )
+    if cmd_result.status != 0:
+        raise SatelliteHostError(
+            f'Error during cockpit installation, installation output: {cmd_result.stdout}'
+        )
+    cmd_result = default_sat.execute(
+        f'sshpass -p "{settings.server.ssh_password}" ssh-copy-id \
```
jaryn, line 67: I know it is not a concern now, but note that passing password on command-line is not considered safe as any user/process can read it (`ps aux`).

jaryn (approved):

pytest_fixtures/sys_fixtures.py
tstrych: Thanks for your comment. Sure you are right. 
But I think we do not care. Because no evil doer should be able to log into our machine.
We already use the same principle in` tests/robottelo/test_ssh.py `
`tests/foreman/uitest_ldap_authentication.py`
`robottelo.ssh.py`
`virtwtho_utils.py`

tstrych:
One more thing I would like to add, can be in next PR, to not install cockpit on every run. 
We should be able to ask if it's already enabled. And install it only when it's not.

tstrych:
```
Launching pytest with arguments -s -vv -k test_positive_cockpit tests/foreman/ui/test_host.py
========== 1 passed, 35 deselected, 3 warnings in 1090.43s (0:18:10) ===========
```

synkd (approved):

JacobCallahan (approved):
@tstrych very nicely done! 
One note for the future (and you can make an issue for this if you'd like) is that these methods could be integrated into the ContentHost and Satellite classes respectively. This is non-blocking for your current PR

pytest_fixtures/sys_fixtures.py
ogajduse: I got you both @jaryn and @tstrych. However, I simply do not like the implementation that is using plaintext passwords. Looking at the `sshpass` man page, there is a `-f` option that allows us to pass the path to a file containing the password instead of passing password directly.
Every SatLab VM has `id_rsa_hudson` key installed in the `authorized_keys`, that means we can make use of it and use pubkey auth instead of the current solution.
Is that doable for all the modules that you mentioned @tstrych?

PR #8816, opened 2021-08-02
add fixture to record test markers as junit testCase properties

rplevka:
This fixture should automatically record all testimony tags (mainly importance, component, assignee) as a junit xml properties.


mshriver (approved):
I think this is a fine start to injecting the marker metadata into the XML properties.

It will end up including `parametrize` properties that aren't unique per test, since it will read the marker for all of the parameters. This isn't ideal, but isn't much of an issue.

If we want to limit scope more, and tie this together with testimony metadata more directly than just adding all marks, we can also modify `user_properties` of the items at collection time, in `pytest_plugins/testimony_markers.py`. Here is where the docblock tokens are getting turned into the markers that this fixture relies on.

JacobCallahan (approved):

PR #8819, opened 2021-08-03
Adding skip_if_open for failed tests

akhil-jha:
Tests failing due `prune_known_hosts!`. 
The test fails on the first run, but if ran again, it passes.
Adding `:BZ: 1983043`

### DIFF
--- a/tests/foreman/api/test_errata.py
+++ b/tests/foreman/api/test_errata.py
@@ -168,7 +168,6 @@ def _fetch_available_errata(module_org, host, expected_amount, timeout=120):
 
 @pytest.mark.upgrade
 @pytest.mark.tier3
-@pytest.mark.skip_if_open("BZ:1983043")
 def test_positive_install_in_hc(module_org, activation_key, custom_repo, rh_repo):
     """Install errata in a host-collection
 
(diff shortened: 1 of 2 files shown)


swadeley:
@akhil-jha I think you should not add this now, the bug[1] is already in `Status: | MODIFIED` , it should be in the next snap.


[1]
[Bug 1983043 - undefined method `prune_known_hosts!` when executing a remote job](https://bugzilla.redhat.com/show_bug.cgi?id=1983043)

mshriver:
Good point @swadeley, @akhil-jha Let's just add this BZ to the testimony `:BZ:` token to indicate coverage where relevant.

Is this fix going to be backported to z-streams? Consider opening a PR against `6.9.z` branch

akhil-jha:
I haven't seen this bug in 6.9.z yet. Adding in z-streams would be pointless. ***I think***
Sure, I'll add `:BZ:` in the test.

mshriver (approved):

mshriver:
@swadeley assigning you, go for the squash+merge if you agree with the update.

swadeley (approved):
ACK

PR #8821, opened 2021-08-03
Removing export from hammer repository command

latran:
`hammer repository export` doesn't exist anymore so removing that command

### DIFF
--- a/robottelo/cli/repository.py
+++ b/robottelo/cli/repository.py
@@ -12,7 +12,6 @@ Subcommands::
 
     create                        Create a repository
     delete                        Destroy a repository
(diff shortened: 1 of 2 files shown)


pondrejk:
the robottelo test also needs updating
https://github.com/SatelliteQE/robottelo/blob/ea1987e335fe41b566e3b898b47e0a640e1bfeea/tests/robottelo/test_cli_method_calls.py#L62

pondrejk (commented):
ack pending test update

swadeley (approved):
ACK

mshriver (approved):

PR #8832, opened 2021-08-09
commented out OStree tests

ColeHiggins2:
Ostrees were removed in satellite 6.10, Functionality will return in 7.0. Tests are commented out and should not be run during automation until then.


mshriver (approved):
Just don't tell anybody this was my idea.

JacobCallahan (approved):
It is all @mshriver's fault!!

PR #8840, opened 2021-08-11
fixed the ldap tests related ui

omkarkhatavkar:


tests/foreman/ui/test_ldap_authentication.py
```diff
@@ -1537,10 +1537,9 @@ def test_deleted_idm_user_should_not_be_able_to_login(auth_source_ipa, ldap_tear
 
     :steps:
         1. Create a new auth source with onthefly enabled
-        2. Create a new user in IDM and assigning a group to it.
-        3. Login to satellite to create the user
-        4. Delete the user from IDM
-        5. Try login to the satellite from the user
+        2. Login to satellite to create the user
```
omkarkhatavkar, line 1540: No need to keep that user in separate group, as it was just related to deleting the user

omkarkhatavkar:
trigger: test-robottelo
pytest: tests/foreman/ui/test_ldap_authentication.py -k 'test_email_of_the_user_should_be_copied or test_deleted_idm_user_should_not_be_able_to_login'

tests/foreman/ui/test_ldap_authentication.py
swadeley: but how can you delete the user in IDM if you do not create it?

swadeley (approved):
ACK, with question about creating user before deletion and "onthefly"

tests/foreman/ui/test_ldap_authentication.py
omkarkhatavkar: Yes once the user is created in IDM and having `onthefly` setting set in ldap server, On first login user get created in Satellite.

PR #8848, opened 2021-08-16
Temporary fix for boxkey errors in ssh.py

JacobCallahan:
This is a temporary fix resulting from issues we're seeing around import
order causing boxkey errors with settings.server.hostname.
By delaying the import of he settings object, we should only import
after that setting has been set.


mshriver (approved):

swadeley (approved):
ACK

PR #8849, opened 2021-08-16
[6.9.z herry-pick] Temporary fix for boxkey errors in ssh.py

JacobCallahan:
This is a temporary fix resulting from issues we're seeing around import
order causing boxkey errors with settings.server.hostname.
By delaying the import of he settings object, we should only import
after that setting has been set.

### DIFF
--- a/robottelo/ssh.py
+++ b/robottelo/ssh.py
@@ -10,7 +10,6 @@ from io import StringIO
 import paramiko
 
 from robottelo.cli import hammer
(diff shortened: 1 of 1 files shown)


mshriver:
lightning round

mshriver:
Failing some unit tests, looks like an import was missed. my fault for rushing the merge - unit tests weren't run against #8848  when it failed pre-commit.

PR #8850, opened 2021-08-16
[6.9.z] Fix failing ssh unittests

JacobCallahan:
Modify mock path to adjust for the ssh module no longer importing
settings at the module level.


mshriver (commented):
posthumously approved

PR #8860, opened 2021-08-18
Treat settings.server.hostname as a property

JacobCallahan:
This change overrides the behavior of the DynaBox class to allow for
proprty-like behavior only for settings.server.hostname

### DIFF
--- a/robottelo/config/__init__.py
+++ b/robottelo/config/__init__.py
@@ -4,7 +4,7 @@ from pathlib import Path
 from urllib.parse import urljoin
 from urllib.parse import urlunsplit
 
-from dynaconf import LazySettings
+from dynaconf import base
 from dynaconf.validator import ValidationError
 
 from robottelo.config.validators import VALIDATORS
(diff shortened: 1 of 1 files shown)


JacobCallahan:
trigger: test-robottelo
pytest: tests/foreman/ui/test_repository.py -n 2

JacobCallahan:
trigger: test-robottelo
pytest: tests/foreman/ui/test_repository.py -n 2

mshriver (commented):
If #8861 is successful as well, I think I'd prefer to use the `.set` method instead of creating this wrapper

robottelo/config/__init__.py
```diff
@@ -16,7 +16,27 @@
     os.environ['ROBOTTELO_DIR'] = str(robottelo_root_dir)
 
 
-settings = LazySettings(
+class DelayedBox(base.DynaBox):
+    """This class allows settings.server.hostname to have like a property"""
```
swadeley, line 20:
```suggestion
    """This class allows settings.server.hostname to behave like a property"""
```

robottelo/config/__init__.py
```diff
@@ -16,7 +16,27 @@
...
+
+    def __getattribute__(self, name):
```
jaryn, line 22: Can you please explain why the `__getattribute__` is used? The intended behavior described in the docstring could be done with properties.

```
class DelayedBox(base.DynaBox):
    @property
    def hostname(self):
        return self._hostname

    @hostname.setter
    def set_hostname(self, value):
        self._hostname = value

JacobCallahan: I didn't use property because I want more fine control over the get/set behavior since modified hostname must only apply under `settings.server` and not other theoretical hostname paths.
For example: `settings.dogfood.hostname`


Ultimately, this PR will almost certainly not be the correct solution for our problem. Because of this, I'll likely close this PR in the near future.

JacobCallahan:
Closing for now as this likely won't solve the underlying issue we're seeing.

PR #8861, opened 2021-08-18
Attempt to use dynaconf's explcit set to change settings.server.hostname

JacobCallahan:
This may resolve issues we're seeing with unresolved hostnames

This is an alternative to #8860


JacobCallahan:
trigger: test-robottelo
pytest: tests/foreman/ui/test_repository.py -n 2

mshriver (approved):

JacobCallahan:
The results of the PRT run are largely consistent with previous runs of this test module. The main difference is that the tests that failed due to not having a hostname now make it past that stage.

PR #8872, opened 2021-08-23
Move helper methods onto ContentHost/Capsule/Satellite

mshriver:
- Added `ContentHost.install_katello_ca`, now takes a satellite instance as a required argument
- Added ContentHost.remove_katello_ca, now does not require a satellite instance/hostname
- New cached property added to hosts.Satellite for the katello CA rpm URL
- Updated all callers of install_katello_ca to pass a satellite instance via default_sat fixture
- Added `ContentHost.configure_rex`, to replace functions in two test modules
- Updated use of ContentHost helper methods, passing Satellite instance where necessary
- Remove endtoend utils module
  - Move the steps executed against the content host into the tests directly
  - Use existing content host fixtures instead of VMBroker in the tests directly
- Update cli/test_remoteexecution, remove ssh module dep and use configure_rex
- Update ContentHost methods to pass satellite, including products.RepositoryCollection
- Update uses of ContentHost.configure_rhai_client
- Remove ssh module dependency in robottelo.hosts


robottelo/hosts.py
```diff
@@ -249,21 +248,53 @@ def install_katello_host_tools(self):
...
+        self.execute('yum erase -y $(rpm -qa |grep katello-ca-consumer)')
+        # Checking the return_code here to verify katello-ca rpm is actually
+        # not present in the system
+        result = self.execute('rpm -qa |grep katello-ca-consumer')
+        if result.return_code == 0:
+            raise ContentHostError(f'katello-ca rpm(s) are still installed: {result.stdout}')
+        # Resetting rhsm.conf to point to cdn
+        # TODO don't hardcode these endpoints,
+        # read them from settings.repos or robottelo.constants.repos
+        rhsm_updates = [
```
JacobCallahan, line 288: We should be able to just use `subscription-manager clean` to reset

JacobCallahan (changes requested):
Will do a full review tomorrow

robottelo/hosts.py
mshriver: updating, can replace this whole block with `self.execute`

robottelo/hosts.py
```diff
@@ -249,21 +272,41 @@ def install_katello_host_tools(self):
...
+        result = self.execute(f'rpm -q katello-ca-consumer-{satellite.hostname}')
+        # Checking the return_code here to verify katello-ca rpm is actually
+        # present in the system
+        if result.status != 0:
+            ContentHostError('Failed to download and install the katello-ca rpm')
+
+    def remove_katello_ca(self):
+        """Removes katello-ca rpm from the broker virtual machine.
+
+        :param: str capsule: (optional) Capsule hostname
```
mshriver, line 297: remove this line

robottelo/hosts.py
```diff
@@ -107,6 +107,29 @@ def ip_addr(self):
...
+            r'(?P<distro>.+) release (?P<major>\d+)(.(?P<minor>\d+))?', result.stdout[0]
+        )
+        if match is None:
+            raise ContentHostError(f'Not able to parse release string "{result.stdout[0]}"')
+        groups = match.groupdict()
+        return (
+            groups['distro'],
+            int(groups['major']),
+            groups['minor'] if groups['minor'] is None else int(groups['minor']),
+        )
```
JacobCallahan, line 131: Would this be easier if you used a `Version` instance instead? Even potentially returning a `Version` instance?

robottelo/hosts.py
```diff
@@ -249,21 +272,40 @@ def install_katello_host_tools(self):
...
         :return: None.
         :raises robottelo.hosts.ContentHostError: If katello-ca wasn't
             installed.
         """
-        try:
-            install_katello_ca(hostname=self.hostname, sat_hostname=sat_hostname)
-        except AssertionError:
-            raise ContentHostError('Failed to download and install the katello-ca rpm')
+        cert_rpm_url = satellite.url_katello_ca_rpm
+        self.execute(f'rpm -Uvh {cert_rpm_url}')
```
JacobCallahan, line 285:
```suggestion
        self.execute(f'rpm -Uvh {satellite.url_katello_ca_rpm}')
```

robottelo/hosts.py
```diff
@@ -457,7 +486,8 @@ def configure_puppet(self, rhel_repo=None, proxy_hostname=None):
         # sat6 under the capsule --> certifcates or on capsule via cli "puppetserver
         # ca list", so that we sign it.
         self.execute('puppet agent -t')
-        ssh.command(cmd='puppetserver ca sign --all', hostname=proxy_hostname)
+        with Host(proxy_hostname) as proxy_host:
```
JacobCallahan, line 489: where is this context manager coming from?

robottelo/hosts.py
```diff
@@ -479,7 +509,27 @@ def execute_foreman_scap_client(self, policy_id=None):
...
+        """Setup a VM host for remote execution.
+
+        :param Satellite satellite: a hosts.Satellite instance
+        :param str org: The organization entity, label attr is used
+        :param int subnet: (Optional) Nailgun subnet entity id, to be used by the vm_client host.
+        :param bool by_ip: Whether remote execution will use ip or host name to access server.
+        """
+        self.install_katello_ca(satellite)
+        self.register_contenthost(org.label, lce='Library')
+        assert self.subscribed
```
JacobCallahan, line 522: This should be optional, allowing users to configure rex on already registered hosts.

robottelo/hosts.py
```diff
@@ -479,7 +509,27 @@ def execute_foreman_scap_client(self, policy_id=None):
...
+        add_remote_execution_ssh_key(self.ip_addr)
+        if subnet_id is not None:
+            Host.update({'name': self.hostname, 'subnet-id': subnet_id})
+        if by_ip:
+            # connect to host by ip
+            Host.set_parameter(
+                {'host': self.hostname, 'name': 'remote_execution_connect_by_ip', 'value': 'True'}
+            )
+
+    def configure_rhai_client(self, satellite, activation_key, org, rhel_distro, register=True):
```
JacobCallahan, line 532: same comment about not forcing a registration per call of this method

robottelo/hosts.py
```diff
@@ -796,6 +849,15 @@ def nailgun_capsule(self):
 
         return NailgunCapsule().search(query={'search': f'name={self.hostname}'})[0]
 
+    @cached_property
+    def is_upstream(self):
```
JacobCallahan, line 853: this would be a good check to add to the Satellite's `version` property

tests/foreman/api/test_activationkey.py
```diff
@@ -424,7 +424,7 @@ def test_positive_remove_user():
     """
     password = gen_string('alpha')
     user = entities.User(password=password, login=gen_string('alpha'), admin=True).create()
-    cfg = get_nailgun_config()
+    cfg = deepcopy(default_sat.nailgun_cfg)
```
JacobCallahan, line 427: that nailgun_cfg object doesn't exist until the Satellite's API property is evaluated.

tests/foreman/api/test_contentview.py
```diff
@@ -1188,7 +1190,7 @@ def test_positive_admin_user_actions(content_view, function_role, module_org, mo
         login=user_login,
         password=user_password,
     ).create()
-    cfg = get_nailgun_config()
+    cfg = deepcopy(default_sat.nailgun_cfg)
```
JacobCallahan, line 1193: same comment about this not existing until api is poked

JacobCallahan (changes requested):

robottelo/hosts.py
mshriver: oooooo yes it would. Want me to include that change in the scope? I'll have to update uses to reference the relevant attributes of the version instance instead of indexing.

robottelo/hosts.py
mshriver: My imagination, Host doesn't have its own context manager, just VMBroker.

robottelo/hosts.py
mshriver: ack, going to default to doing these steps though since all uses of it currently expect it. Pass bool false for a kwarg to disable it on demand.

robottelo/hosts.py
mshriver: same comment about default keeping current behavior.

tests/foreman/api/test_activationkey.py
mshriver: yeah that's what I was worried about. Its basically chance that its been called by the time this is done.

Thoughts on the pattern overall? Leave this commit out of the PR for further discussion? Add a call to `default_sat.api` before referencing the attribute in all these uses? Add a call to the `Satellite.__init__` to `self.api`?

robottelo/hosts.py
JacobCallahan: If the scope isn't too big..

tests/foreman/api/test_activationkey.py
mshriver: Discussed strategies a bit offline, with @JacobCallahan having some good ideas for where to init this attribute effectively.

I think I'm going to pull the commit for this into a separate PR, as its scope will grow and it should be thoroughly reviewed by itself.

tests/foreman/api/test_contentview.py
mshriver: Moved this commit, scope will get too big.

tests/foreman/api/test_activationkey.py
mshriver: Commit has been moved to #8882

robottelo/hosts.py
mshriver: image

jaryn (approved):

robottelo/hosts.py
mshriver: Looks like this method, as copied, might not even work. Splitting the distro name out, since `packaging.version.Version` doesn't support such a field.

mshriver: Done, 009e1f7ef

robottelo/hosts.py
mshriver: done, 04d071b6e

robottelo/hosts.py
```diff
@@ -107,6 +108,33 @@ def ip_addr(self):
         ipv4, ipv6 = self.execute('hostname -I').stdout.split()
         return ipv4
 
+    def _get_redhat_release(self):
```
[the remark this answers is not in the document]
mshriver: Done

JacobCallahan (approved):

PR #8878, opened 2021-08-24
[cherrypick] add fixture to record test markers as junit testCase properties

rplevka:


pytest_fixtures/reporting_fixtures.py
```diff
@@ -56,3 +56,9 @@ def record_testsuite_timestamp_xml(record_testsuite_property):
 def record_test_timestamp_xml(request):
     now = datetime.datetime.utcnow()
     request.node.user_properties.append(('start_time', now.strftime(FMT_XUNIT_TIME)))
+
+
+@pytest.fixture(autouse=True, scope='function')
+def record_test_markers_xml(request, record_property):
+    for marker in request.node.iter_markers():
+        request.node.user_properties.append((marker.name, marker.args[0]))
```
mshriver, line 64: @rplevka This caused a bug in the master branch, see #8873

mshriver (changes requested):
The args index needs to be safe, not all markers have args

mshriver (approved):

PR #8882, opened 2021-08-24
Move nailgun cfg helper

mshriver:
CONTAINS #8872, commit will be removed before review.


JacobCallahan:
@mshriver is this something you can continue or would you rather we involve someone aligned to the helpers refactor effort?

mshriver:
Definitely forgot about this one - and it would be better suited for someone involved in helpers refactor to make this change.

PR #8858, opened 2021-08-18
Use default_sat instead of settings.server.hostname

mshriver:
Fixtures and tests using settings.server.hostname should reference default_sat.hostname instead

In addition to the results below, I ran a full high importance CI pipeline collecting the test modules that have hit BoxKeyError - I did not see any BoxKeyError failures in any of the test results.

### api/test_host.py
```
14:01:09  + py.test tests/foreman/api/test_host.py --junit-xml=sat-results.xml -o junit_suite_name=sat-result
14:01:19  ============================= test session starts ==============================
14:01:19  collected 62 items / 13 deselected / 49 selected
14:01:19  
14:07:53  tests/foreman/api/test_host.py ......................................... [ 83%]
14:09:10  ........                                                                 [100%]
```

### api/test_hostgroup.py

test_inherit_puppetclass failure is being addressed by @tstrych 
test_positive_create_with_properties I'm debugging locally
```
 =========================== short test summary info ============================
14:34:22  FAILED tests/foreman/api/test_hostgroup.py::TestHostGroup::test_inherit_puppetclass
14:34:22  FAILED tests/foreman/api/test_hostgroup.py::TestHostGroup::test_positive_create_with_properties
14:34:22  ===== 2 failed, 49 passed, 2 deselected, 333 warnings in 177.91s (0:02:57) =====
```

### api/test_role.py

Failing test is failing in master
```
20:14:37  =========================== short test summary info ============================
20:14:37  FAILED tests/foreman/api/test_role.py::TestCannedRole::test_negative_create_taxonomies_by_org_admin
20:14:37  = 1 failed, 42 passed, 1 skipped, 6 deselected, 474 warnings in 505.91s (0:08:25) =
```

### api/test_repository.py

Test failures, generally consistent with master branch results
```
= 23 failed, 154 passed, 4 skipped, 7 deselected, 815 warnings, 34 errors in 1642.23s (0:27:22) =
```

### api/test_user.py

```
18:21:31  + py.test tests/foreman/api/test_user.py --junit-xml=sat-results.xml -o junit_suite_name=sat-result
18:21:39  ============================= test session starts ==============================
18:21:40  collected 144 items / 3 deselected / 141 selected
18:21:40  
18:22:11  tests/foreman/api/test_user.py ......................................... [ 29%]
18:22:47  ........................................................................ [ 80%]
18:24:12  ............................    

18:24:12  ========= 141 passed, 3 deselected, 417 warnings in 153.38s (0:02:33) ==========
```

[...]

PR #8883, opened 2021-08-25
Use default_sat instead of settings.server.hostname

mshriver:
Add Satellite.url and Satellite.port

Fixtures and tests using settings.server.hostname should reference
default_sat.hostname instead

Update func_locker._get_default_scope
Local settings import, and flattened logic

Resolve BoxKeyError in ui/test_jobinvocation

Backporting PR #8858


JacobCallahan (approved):

swadeley (approved):
ACK

PR #8888, opened 2021-08-25
Improve Satellite context manager

JacobCallahan:
This change aims to avoid potential issues with nested context managers
of the same satellite hostname.
Fixed typo that resulted in a mangled old_hostname variable.

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -984,16 +984,21 @@ class Satellite(Capsule):
         """Satellite objects can be used as a context manager to temporarily force everything
         to use the Satellite object's hostname.
         """
-        self.__old_hostname = settings.server.hostname
-        settings.server.hostname = self.hostname
-        configure_nailgun()
-        configure_airgun()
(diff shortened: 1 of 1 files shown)


robottelo/hosts.py
```diff
@@ -984,16 +984,21 @@ def __enter__(self):
...
-        self.__old_hostname = settings.server.hostname
-        settings.server.hostname = self.hostname
-        configure_nailgun()
-        configure_airgun()
+        self._revert = False
+        # if the hostname is the same, don't make a change
+        if self.hostname != settings.server.hostname:
+            self._revert = True
+            self._old_hostname = settings.server.hostname
+            settings.server.hostname = self.hostname
```
mshriver, line 992: Use `set` method here?

mshriver (approved):

robottelo/hosts.py
JacobCallahan: that was a red herring, it had the same effect :/

PR #8890, opened 2021-08-25
Refactor ssh module and helper functions

mshriver:
Remove most uses of get_connection outside of the ssh module

Convert tests to use default_sat.execute, to use status instead of
return_code, and to treat stdout as a string instead of a list

Updated several uses of download_file and upload_file as well, in order
to eliminate ssh module use from test files that were being modified.

Moved two small functions from test modules where they were duplicated
into robottelo.helpers

Changed target for this branch to `unstable`, which @JacobCallahan created to collect this MR and some of his outstanding work continuing the ssh module refactoring.


JacobCallahan (commented):
ACK, testing in unstable environment

PR #8893, opened 2021-08-26
Improve reliability of ui custom cron test

swadeley:
Remove cli custom cron stubbed test

Hello robottelers

I have made the UI custom cron test more reliable by preventing the test starting less than 2 mins before the next sync event set by the custom cron expression.

I have removed the stubbed CLI custom cron test as it essentially duplicates the UI test except it called for an unrealistic 1 minute interval. I don't believe there is any valid use case for that test. The UI test with 5 minute interval will catch a failure to sync on the first sync event.

Thank you

UPDATE: I had to change the guard time to 3 mins to get it to pass reliably. It was that or change cron event to multiples of 10 which would have made the test time longer.

### DIFF
--- a/tests/foreman/cli/test_syncplan.py
+++ b/tests/foreman/cli/test_syncplan.py
@@ -386,20 +386,6 @@ def test_negative_synchronize_custom_product_past_sync_date(module_org):
         validate_task_status(repo['id'], module_org.id, max_tries=2)
 
 
-@pytest.mark.stubbed
-@pytest.mark.tier4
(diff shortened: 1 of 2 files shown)


devendra104 (approved):

devendra104:
@swadeley Can you check the causes of failure in the PR tester? locally it works.

mshriver:
robottelo-runner hit a browser session issue, not related to @swadeley changes here.

swadeley:
Thank you

PR #8899, opened 2021-08-27
Remove CLI version of inc_updates test

swadeley:
Hello Robo-testers

There are two incremental update tests in this longrun module, one API and one CLI. As they both have the same test steps its really a duplication from errata enumeration logic point of view. In addition, the tests have to run one after the other, and since we converted to pytest format that became a problem. Hence this PR to remove the CLI version. Plenty test coverage in the usual API and UI test directories.

Thank you


JacobCallahan (approved):

swadeley:
Thank you

PR #8904, opened 2021-08-27
Azure CLI test fixes with 2 new context managers

jyejare:
Azure CLI tests failing with Timeout error are now being fixed with a fix of set_hammer_api_timeout. Also, its converted to context managers.

Now there are 2 new context managers:
1. hammer_api_timeout
2. skip_yum_during_provisioning

### DIFF
--- a/robottelo/api/utils.py
+++ b/robottelo/api/utils.py
@@ -939,44 +939,6 @@ def create_org_admin_user(orgs, locs):
     return user
 
 
-@contextmanager
-def skip_yum_update_during_provisioning(template=None):
-    """Hides the yum update command with echo text
-
-    :param str template: The template name where the yum update will be hidden
-    """
-    old = 'yum -t -y update'
-    new = 'echo "Yum update skipped for faster automation testing"'
-    update_provisioning_template(name=template, old=old, new=new)
-    yield
-    update_provisioning_template(name=template, old=new, new=old)
-
-
-@contextmanager
-def hammer_api_timeout(timeout=-1):
-    """Set hammer API request timeout on Satellite
-
-    :param int timeout: request timeout in seconds
-    """
-    new_timeout = f':request_timeout: {timeout}'
-    cli_conf = "~/.hammer/cli.modules.d/foreman.yml"
-
-    if ssh.command(f"grep -i 'request_timeout' {cli_conf}").return_code != 0:
-        ssh.command(f"echo '  {new_timeout}' >> {cli_conf}")
-        revert_method = 'remove'
-    else:
-        old_timeout = ssh.command(f"sed -ne '/:request_timeout.*/p' {cli_conf}").stdout[0]
-        ssh.command(f"sed -ir 's/{old_timeout.strip()}/{new_timeout}/' {cli_conf}")
-        revert_method = 'replace'
-
-    yield
-
-    if revert_method == 'remove':
-        ssh.command(f"sed -i '/{new_timeout}/d' {cli_conf}")
-    else:
-        ssh.command(f"sed -ie 's/{new_timeout}/{old_timeout.strip()}/' {cli_conf}")
-
-
 def update_rhsso_settings_in_satellite(revert=False):
     """Update or Revert the RH-SSO settings in satellite"""
     rhhso_settings = {
--- a/robottelo/api/utils.py
+++ b/robottelo/api/utils.py
@@ -939,42 +939,42 @@ def create_org_admin_user(orgs, locs):
     return user
 
 
-def skip_yum_update_during_provisioning(template=None, reverse=False):
+@contextmanager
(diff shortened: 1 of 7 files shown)


robottelo/api/utils.py
```diff
@@ -940,42 +940,42 @@ def create_org_admin_user(orgs, locs):
...
     """
-    default_timeout = f':request_timeout: {120}'
     new_timeout = f':request_timeout: {timeout}'
-    if not reverse:
-        return ssh.command(
-            "sed -ie 's/{}/{}/' ~/.hammer/cli.modules.d/foreman.yml".format(
-                default_timeout, new_timeout
-            )
-        )
+    cli_conf = "~/.hammer/cli.modules.d/foreman.yml"
```
omkarkhatavkar, line 963: single quote

tests/foreman/api/test_computeresource_azurerm.py
```diff
@@ -238,29 +238,28 @@ def class_host_ft(
...
+                name=self.hostname,
+                provision_method='image',
+                image=module_azurerm_finishimg,
+                root_pass=gen_string('alphanumeric'),
+                environment=module_puppet_environment,
+                puppet_proxy=default_smart_proxy,
+                puppet_ca_proxy=default_smart_proxy,
+            ).create()
+            yield host
+            with satellite_setting('destroy_vm_on_host_delete=True'):
```
omkarkhatavkar, line 261: if this is getting repeated it can go in the fixture as well ?

robottelo/api/utils.py
```diff
@@ -940,42 +940,42 @@ def create_org_admin_user(orgs, locs):
...
+
+    if ssh.command(f"grep -i 'request_timeout' {cli_conf}").return_code != 0:
```
omkarkhatavkar, line 965: will be good if we access using the satellite class object, for this PR I'm good.

omkarkhatavkar (approved):
ACK, Pending some comments

robottelo/api/utils.py
```diff
@@ -940,42 +940,42 @@ def create_org_admin_user(orgs, locs):
...
+        ssh.command(f"echo '  {new_timeout}' >> {cli_conf}")
```
devendra104, line 966: I think we can use `default_sat.execute` instead of `ssh.command` to run the command on satellite.

robottelo/api/utils.py
omkarkhatavkar: @shweta83 I would rather create a separate PR for ssh module, As I see this Pr intention was to quickly fix the test for CI failures. We are going to have have separate branch for SSH module as well.

robottelo/api/utils.py
mshriver: Agreed, and this was discussed above. This isn't in a pytest fixture or test case, so the satellite instance will have to be passed down.  Its not a strict requirement to the PR, just makes for additional instances that @JacobCallahan or I have to replace in the unstable branch before it can be merged to master.

mshriver: In fact, this whole method should get moved to the `Capsule` class. Then its references to `self.execute`

mshriver: Either that or converted to a fixture. I think either of those options are better than just leaving it as an `api.utils` function that needs to be imported. The context manager pattern is nice to have, and I think indicates that this should be a fixture. Its only uses are in the azure test module.

mshriver (changes requested):
The pattern is a good improvement to the test functionality. Given the scope/use of the methods that are being updated here, I think you should take things a step farther and move the functions to the Host based class(es) or make them pytest fixtures.

robottelo/api/utils.py
jyejare: @mshriver Its not really Azure specific, this is a temporary update to satellite Hammer config that changes its timeout so that satellite hammer command wont fail with `TimeoutError`. This context manager can be used everywhere in a satellite like `Content View publish`, `Repo Sync` and `Provisioing(Azure, EC2, VMWare etc)`. The customers also been asked to adjust this timeout as per their environment.

And hence I am moving this under `Satellite` class as a context manager. Along with this `skip_yum_update_during_provisioning` is also going into satellite class.

tests/foreman/api/test_computeresource_azurerm.py
jyejare: @omkarkhatavkar  the full hostname there is different every time and is being initiated per class. That can be a good improvement but I would like to keep that separate from this PR.

jyejare:
@mshriver @omkarkhatavkar @devendra104 @shweta83 Please revisit the context managers are now moved to Satellite class  as both relates to satellite more.

mshriver (approved):

devendra104 (approved):

PR #8915, opened 2021-08-31
puppet test parts removed in upgrade tests

tstrych:
- test_capsule - puppet file repo was removed
- test_contentview - puppet artifacts were removed, and also other options were considered, read below in thoughts section, if you want ;) 

@vsedmik  and @latran are component owners of these tests.
Tagging also @devendra104  to double check upgrade part.
So please take a look. 

Test results, I tried to run pre-upgrade and post-upgrade on 6.9 and again on 6.10, all passed
```
6.9
tests/upgrades/test_contentview.py::TestContentView::test_cv_preupgrade_scenario PASSED                                                                                                [100%]
tests/upgrades/test_contentview.py::TestContentView::test_cv_postupgrade_scenario PASSED                                                                                               [100%]

6.10
tests/upgrades/test_contentview.py::TestContentView::test_cv_preupgrade_scenario PASSED                                                                                                [100%]
tests/upgrades/test_contentview.py::TestContentView::test_cv_postupgrade_scenario PASSED                                                                                               [100%]
```

Thoughts:
- test_contentview -  in pre upgrade we can keep the puppet repository and modules, the problem is that pre_upgrade also runs on 6.10 (master) branch, even though the machine is sat 6.9., it's not worth to keep the puppet modules code in master branch just for one test, or better said just for puppet repository search in 6.10, (because in postupgrade there is no longer possible to  list puppet modules, remove puppet module repo or  add puppet module repo, so only thing that would stay is `entities.Repository(organization=org.id).search(......`
- i think ideally pre upgrade should run on 6.9 and post on 6.10 branch

From my point of view this is much cleaner solution as a whole.

### DIFF
--- a/tests/upgrades/test_capsule.py
+++ b/tests/upgrades/test_capsule.py
@@ -27,7 +27,6 @@ from upgrade_tests.helpers.scenarios import rpm2
 from robottelo.api.utils import call_entity_method_with_timeout
 from robottelo.api.utils import promote
 from robottelo.config import settings
-from robottelo.constants.repos import CUSTOM_PUPPET_REPO
 from robottelo.datafactory import gen_string
 from robottelo.upgrade_utility import create_repo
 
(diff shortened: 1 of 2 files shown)


devendra104 (approved):

vsedmik (approved):
LGTM

mshriver (approved):

PR #8916, opened 2021-08-31
realm cli tests fixes

pondrejk:
Fixing cli realm tests. This suite expected realm capsule to be enabled beforehand (which was done by the old CI), this is no longer the case. Here I reuse some fixtures from the ldap_authentication module + add configure_realm fixture and config. Merging is pending on the satelliteqe-jenkins pr with the idm config. Tests now depend on the potentially destructive action so mark was added

```
pytest tests/foreman/cli/test_realm.py -k test_positive
================================================== test session starts ===================================================
platform linux -- Python 3.9.6, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
plugins: forked-1.3.0, services-2.2.1, reportportal-5.0.8, mock-3.6.1, cov-2.12.1, xdist-2.3.0, ibutsu-1.16
collected 15 items / 10 deselected / 5 selected                                                                          

tests/foreman/cli/test_realm.py .....                                                                              [100%]

================================ 5 passed, 10 deselected, 6 warnings in 432.66s (0:07:12) ================================
```


pytest_fixtures/satellite_auth.py
```diff
@@ -265,31 +267,53 @@ def enable_external_auth_rhsso(enroll_configure_rhsso_external_auth):
...
         hostname=settings.ipa.hostname,
     )
-    result = run_command(
-        cmd=f'ipa host-add --random {settings.server.hostname}',
-        hostname=settings.ipa.hostname,
-    )
+    output = ssh.command(f'ipa host-find {settings.server.hostname}')
+    if output.return_code != 0:
+        result = run_command(
+            cmd=f'ipa host-add --random {settings.server.hostname}',
```
mshriver, line 277: Please don't reference `settings.server.hostname` in fixtures or tests directly, unless absolutely necessary.

There is a `default_sat` fixture, which will provide you a satellite instance. It has a `hostname` property that you can reference here, ensuring that when this fixture runs its executing ssh commands against the correct target satellite.

pytest_fixtures/subscription_fixtures.py
```diff
@@ -0,0 +1,54 @@
+import pytest
+
+from robottelo.config import settings
+from robottelo.constants import RHEL_7_MAJOR_VERSION
+from robottelo.rhsso_utils import run_command
+
+
+def unsubscribe():
+    """unregisters a machine from cdn"""
+    run_command('subscription-manager unregister')
```
mshriver, line 10: There is a `ContentHost.unregsiter` method that should be used here.

I think it would be good to include an additional call to `clean` in the current function.

pytest_fixtures/subscription_fixtures.py
```diff
@@ -0,0 +1,54 @@
...
+    run_command('subscription-manager clean')
+
+
+@pytest.fixture(scope='session')
+def clean_rhsm():
+    """removes pre-existing candlepin certs and resets RHSM."""
+    # removing the katello-ca-consumer
+    run_command('rpm -qa | grep katello-ca-consumer | xargs -r rpm -e')
```
mshriver, line 18: There is a `ContentHost.remove_katello_ca` method that can be used here, which includes a `subscription-manager clean`.

I think it would be far more reliable for the fixture to store the original rhsm.conf file before the test, than to try and mangle the file in-place.  This has the potential to destroy the rhsm.conf at a session scope, creating difficult to debug test failures.

pytest_fixtures/subscription_fixtures.py
```diff
@@ -0,0 +1,54 @@
...
+
+@pytest.fixture(scope='session')
+def subscribe_satellite(clean_rhsm):
+    """subscribe satellite to cdn"""
+    run_command(
+        'subscription-manager register --force --user={} --password={} {}'.format(
+            settings.subscription.rhn_username,
+            settings.subscription.rhn_password,
+            # set release to "7Server" currently with this scope
+            f'--release="{RHEL_7_MAJOR_VERSION}Server"',
```
mshriver, line 41: I think this needs to be more dynamic, as the version of RHEL that Satellite is running on will change.

There is a property that you can use to actually look this up at run-time, instead of hard coding it.

Include `default_sat` in the fixture parameters, and then reference `default_sat.os_version`.  This will give you a `packaging.Version` instance, which has properties for `major`, `minor`, etc.  `default_sat.os_version.major` will give you the major RHEL version that the target satellite is running on.

pondrejk:
@mshriver, thanks for the review but given the priority of automation fixes I'd like to do the refactoring work separately. There is a lot more in fixtures/satellite_auth.py that needs addressing outside the scope of this PR. As for pytest_fixtures/subscription_fixtures.py, I just elevated it from tests/foreman/ui/test_ldap_authentication.py  where it lived happily for a long time, waiting 2 more weeks for refactor shouldn't do much harm.

pytest_fixtures/satellite_auth.py
mshriver: @pondrejk as discussed, I'm not going to enforce this change.  But note that you very well may encounter BoxKeyErrors on these tests while this fixture is referencing `settings.server.hostname` directly. The resolution for this is in using `default_sat`, which @JacobCallahan or I will apply in the 'unstable' branch.

pytest_fixtures/subscription_fixtures.py
swadeley: @pondrejk why not write an issue to track this request if you do not have time to do it in this sprint?

pytest_fixtures/subscription_fixtures.py
swadeley: @pondrejk why not write an issue to track this request if you do not have time to do it in this sprint?

conf/idm.yaml.template
```diff
@@ -0,0 +1,5 @@
+# IDM server details
+IDM:
+  IDM_PASSWORD:
+  IDM_SERVER_IP:
+  KEYTAB_URL:
```
omkarkhatavkar, line 5: We already have the IDM Details, why are you creating new file ?

omkarkhatavkar (changes requested):
I think other things are okay, only idm template file is not required, you can use ipa file

pondrejk:
https://github.com/SatelliteQE/robottelo/issues/8930 created for the voiced concerns

conf/idm.yaml.template
pondrejk: ah, I didn't see the ip and keytab settings anywhere  so I created the config. If you feel strongly about it I can put those settings to the ipa file, though I'll need to change the satelliteqe-jenkins pr due to this

mshriver: I'd agree with Omkar here, thought I had made a comment about this but can't find it.  Doesn't seem worth creating a new settings namespace for these, when they're so closely related to existing namespaces.

pondrejk:
updated to use ipa config here and in satelliteqe-jenkins

omkarkhatavkar (approved):
ACK

mshriver:
thanks @pondrejk !

pytest_fixtures/subscription_fixtures.py
pondrejk: I figure this didn't mark as outdated because the lines in question were removed after line 18, rhsm config editing wasn't necessary. The method Mike mentions should imho be also elevated into some fixture, I don't want to dig into fixture refactoring in this PR

pytest_fixtures/satellite_auth.py
```diff
@@ -248,48 +250,70 @@ def enroll_configure_rhsso_external_auth():
...
-
     run_command(
         cmd=f'echo {settings.ipa.password} | kinit admin',
         hostname=settings.ipa.hostname,
     )
-    result = run_command(
-        cmd=f'ipa host-add --random {settings.server.hostname}',
-        hostname=settings.ipa.hostname,
-    )
+    output = ssh.command(f'ipa host-find {default_sat}')
```
omkarkhatavkar, line 274:
```suggestion
    output = ssh.command(f'ipa host-find {default_sat.hostname}')
```

mshriver:
trigger: test-robottelo
pytest: tests/foreman/cli/test_realm.py -k test_positive

conf/ipa.yaml.template
```diff
@@ -15,3 +15,6 @@ IPA:
   GROUPS:
     - satadmins
     - satusers
+  IDM_PASSWORD:
```
lpramuk, line 18: Basically you are trying to introduce dubious entry for `IPA:` -> `PASSWORD:`
Stick to `settings.ipa.password`

lpramuk (changes requested):
Fix adding dubious settings

conf/ipa.yaml.template
```diff
@@ -15,3 +15,6 @@ IPA:
...
+  IDM_SERVER_IP:
```
lpramuk, line 19: `IDM_SERVER_IP` = forward resolved `IPA:` -> `HOSTNAME:`

So I suggest you to drop this setting and use this instead:
```
socket.gethostbyname(settings.ipa.hostname)
```

lpramuk (changes requested):
Drop `IDM_SERVER_IP` setting and use `socket.gethostbyname(settings.ipa.hostname)` instead

pytest_fixtures/satellite_auth.py
```diff
@@ -248,48 +250,70 @@ def enroll_configure_rhsso_external_auth():
...
+    run_command(cmd='mv /root/freeipa.keytab /etc/foreman-proxy')
+    run_command(cmd='chown foreman-proxy:foreman-proxy /etc/foreman-proxy/freeipa.keytab')
     run_command(
-        cmd=f"ipa-client-install --password '{password}' "
-        f'--domain {domain} '
-        f'--server {settings.ipa.hostname} '
-        f'--realm {domain.upper()} -U'
+        cmd='satellite-installer --foreman-proxy-realm true '
+        f'--foreman-proxy-realm-principal realm-proxy@{realm} '
+        f'--foreman-proxy-dhcp-nameservers {settings.ipa.idm_server_ip}'
```
lpramuk, line 312: Please change to this:
`f'--foreman-proxy-dhcp-nameservers {socket.gethostbyname(settings.ipa.hostname)}'`

mshriver:
@pondrejk the unstable branch has been tested and merged back into master, some of the functions that you're using here have moved, necessitating making these updates with the new pattern for ssh interaction.

Please reach out if you'd like some assistance making the changes. I'm happy to push a commit here to update things, since you opened this PR before the new pattern was being enforced.

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_realm.py -k test_positive

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_realm.py -k test_positive

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_realm.py -k test_positive

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_realm.py -k test_positive

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_realm.py -k test_positive

pytest_fixtures/satellite_auth.py
```diff
@@ -248,48 +251,70 @@ def enroll_configure_rhsso_external_auth():
...
+            if 'Random password' in line:
+                _, password = line.split(': ', 2)
+                break
+        ssh.command(
+            cmd=f'ipa service-add HTTP/{default_sat.hostname}',
+            hostname=settings.ipa.hostname,
+        )
+        _, domain = settings.ipa.hostname.split('.', 1)
+        result = ssh.command(
+            cmd=f"ipa-client-install --password '{password}' "
```
mshriver, line 291: `password` isn't set, resulting in an exception here.

This suggests the `if 'Random password' in line` check above is never hit, so the variable is never named in this scope.

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_realm.py -k test_positive

pytest_fixtures/satellite_auth.py
pondrejk: hm, the issue was that the iteration went sign by sign, now it should be looking at lines

pondrejk:
@mshriver  rpr now passes, so marking as ready, cheers

pytest_fixtures/satellite_auth.py
```diff
@@ -248,48 +251,70 @@ def enroll_configure_rhsso_external_auth():
...
-
     run_command(
         cmd=f'echo {settings.ipa.password} | kinit admin',
         hostname=settings.ipa.hostname,
     )
-    result = run_command(
-        cmd=f'ipa host-add --random {settings.server.hostname}',
-        hostname=settings.ipa.hostname,
-    )
+    output = ssh.command(f'ipa host-find {default_sat.hostname}')
```
mshriver, line 275: For these cases, since the command is being run on the `default_sat` instance, you should use `default_sat.execute()` here instead of `ssh.command`. `ssh.command` is still relying on trying to read settings.server.hostname.

mshriver: Its a mix here of commands against `settings.ipa.hostname` and default_sat, this comment applies to all the places where you're running commands against default_sat.  Using `ssh.command` for the commands send to the IPA host is acceptable - though not ideal.  Ideal would be creating a host instance for it.

```
from robottelo.hosts import ContentHost

ipa_host = ContentHost(settings.ipa.hostname)

default_sat.execute('yum -y --disableplugin=foreman-protector install ipa-client ipa-admintools')
ipa_host.execute(f'echo {settings.ipa.password} | kinit admin')
output = default_sat.execute(f'ipa host-find {default_sat.hostname}')
if output.status != 0:
    result = ipa_host.execute(f'ipa host-add --random {default_sat.hostname}')
    <etc>
```

tests/foreman/cli/test_realm.py
```diff
@@ -32,7 +32,10 @@
 
 
 @pytest.mark.tier1
-def test_positive_delete_by_name(module_fake_proxy):
+@pytest.mark.destructive
```
mshriver, line 35: With these tests being marked destructive, the `subscribe_satellite` fixture needs to be using `destructive_sat` instead of `default_sat`. 

This mark controls test collection, but doesn't control replacing `default_sat` instance.

The core patterns for this need some updates, it came up in the reviewers call this morning. @JacobCallahan @omkarkhatavkar

pondrejk: Hm, so what to do with that? This applies to `enroll_idm_and_configure_external_auth`, `configure_realm` and `module_fake_proxy` as well, right?

mshriver: This does become complicated and redundant if you've got some destructive tests and some non-destructive tests using the same fixtures.  Hence why the pattern needs some updating.

JacobCallahan:
> @JacobCallahan opinion on the scope of this PR here and impact due to destructive not being collected at the moment?
> 
> It needs to be addressed at the root (default_sat fixture), otherwise we'll have to repeat a bunch of fixture definitions.

Correct, ideally this should be solved at the default_sat level. For better or worse, default_sat is tied into a number of things and conflicts with targets when other satellites want to use helper methods and related fixtures.

pondrejk:
@mshriver thanks for the comments, I believe I have addressed everything

PR #8917, opened 2021-09-01
Remove RHAI tests

jameerpathan111:
RHAI plugin is depricated in satellite 6.9 and removed from Satellite 6.10
We no longer need RHAI tests for Satellite >= 6.10.


devendra104 (approved):

omkarkhatavkar (approved):
ACK

PR #8922, opened 2021-09-01
puppet test removal in api contentmanagemenet

tstrych:
@latran If you think I should change a test, please let me know what would you expect in the new one instead of puppet repository and puppet modules.


devendra104 (approved):

mshriver (approved):
Merging as is, if @latran would like to replace this coverage with a different content type, that can be done in a separate PR.

latran:
It's fine to remove this entire test.  Ensure that syncing content to capsule is actually covered elsewhere with different content type, if not, please create a separate PR for that.

latran:
And @mshriver got to this PR before I can even review it. -_-

PR #8926, opened 2021-09-02
Rewrote satellite sync for new implmentation

latran:
Satellite sync has changed for 6.10.  Rewrote test cases for entire class and removed puppet

```
pytest tests/foreman/cli/test_satellitesync.py 

===================== 1 failed, 15 passed, 1 skipped, 26 deselected, 3 warnings in 1882.53s (0:31:22) ==========
```

The one failure is for `test_positive_export_import_redhat_cv_with_huge_contents` which appears to have encounter a timeout that I can't resolve at the moment.  Will look into that failure later.  Also removed some N/A test cases from file.

### DIFF
--- a/tests/foreman/cli/test_satellitesync.py
+++ b/tests/foreman/cli/test_satellitesync.py
@@ -16,16 +16,16 @@
 
 :Upstream: No
 """
-import json
+import os.path
 from random import randint
 
 import pytest
-from fauxfactory import gen_integer
 from fauxfactory import gen_string
-from nailgun import entities
 
 from robottelo import manifests
 from robottelo.cli.base import CLIReturnCodeError
+from robottelo.cli.content_export import ContentExport
+from robottelo.cli.content_import import ContentImport
 from robottelo.cli.contentview import ContentView
 from robottelo.cli.factory import make_content_view
 from robottelo.cli.factory import make_lifecycle_environment
@@ -33,62 +33,32 @@ from robottelo.cli.factory import make_org
 from robottelo.cli.factory import make_product
 from robottelo.cli.factory import make_repository
 from robottelo.cli.package import Package
-from robottelo.cli.puppetmodule import PuppetModule
 from robottelo.cli.repository import Repository
 from robottelo.cli.repository_set import RepositorySet
 from robottelo.cli.settings import Settings
 from robottelo.cli.subscription import Subscription
-from robottelo.config import robottelo_tmp_dir
-from robottelo.config import settings
+from robottelo.constants import CONTAINER_REGISTRY_HUB
 from robottelo.constants import DEFAULT_CV
-from robottelo.constants import ENVIRONMENT
 from robottelo.constants import PRDS
 from robottelo.constants import REPOS
 from robottelo.constants import REPOSET
 
 
+EXPORT_DIR = '/var/lib/pulp/exports/'
+IMPORT_DIR = '/var/lib/pulp/imports/'
+
+
 @pytest.fixture(scope='class')
-def class_export_directory(default_sat):
-    """Create a directory for export, configure permissions and satellite settings,
-    at the end remove the export directory with all exported repository archives"""
-    export_dir = gen_string('alphanumeric')
-
-    # Create a new 'export' directory on the Satellite system
-    result = default_sat.execute(f'mkdir /mnt/{export_dir}')
-    assert result.status == 0
-
-    result = default_sat.execute(f'chown foreman.foreman /mnt/{export_dir}')
-    assert result.status == 0
-
-    result = default_sat.execute(f'ls -Z /mnt/ | grep {export_dir}')
-    assert result.status == 0
-    assert len(result.stdout) >= 1
-    assert 'unconfined_u:object_r:mnt_t:s0' in result.stdout
-
-    # Fix SELinux policy for new directory
-    result = default_sat.execute(
-        f'semanage fcontext -a -t foreman_var_run_t "/mnt/{export_dir}(/.*)?"'
-    )
-    assert result.status == 0
-
-    result = default_sat.execute(f'restorecon -Rv /mnt/{export_dir}')
-    assert result.status == 0
-
-    # Assert that we have the correct policy
-    result = default_sat.execute(f'ls -Z /mnt/ | grep {export_dir}')
-    assert result.status == 0
-    assert len(result.stdout) >= 1
-    assert 'unconfined_u:object_r:foreman_var_run_t:s0' in result.stdout
-
-    # Update the 'pulp_export_destination' settings to new directory
-    Settings.set({'name': 'pulp_export_destination', 'value': f'/mnt/{export_dir}'})
-    # Create an organization to reuse in tests
-    yield export_dir
-    default_sat.execute(f'rm -rf /mnt/{export_dir}')
-
-
-class ExportDirectoryNotSet(Exception):
-    """Raise when export Directory is not set or found"""
+def config_export_import_settings():
+    """Check settings and set download policy for export.  Reset to original state after import"""
+    download_policy_value = Settings.info({'name': 'default_download_policy'})['value']
+    rh_download_policy_value = Settings.info({'name': 'default_redhat_download_policy'})['value']
+    disconnected_mode_value = Settings.info({'name': 'content_disconnected'})['value']
+    Settings.set({'name': 'default_redhat_download_policy', 'value': 'immediate'})
+    yield
+    Settings.set({'name': 'default_download_policy', 'value': download_policy_value})
+    Settings.set({'name': 'default_redhat_download_policy', 'value': rh_download_policy_value})
+    Settings.set({'name': 'content_disconnected', 'value': disconnected_mode_value})
 
 
 @pytest.mark.run_in_one_thread
@@ -96,17 +66,20 @@ class TestRepositoryExport:
     """Tests for exporting a repository via CLI"""
 
     @pytest.mark.tier3
-    def test_positive_export_custom_product(self, class_export_directory, module_org, default_sat):
-        """Export a repository from the custom product
+    def test_positive_export_complete_version_custom_repo(
+        self, default_sat, export_cleanup, module_org
+    ):
+        """Export a custom repository via complete version
 
         :id: 9c855866-b9b1-4e32-b3eb-7342fdaa7116
 
-        :expectedresults: Repository was successfully exported, rpm files are
+        :expectedresults: Repository was successfully exported, exported files are
             present on satellite machine
 
         :CaseLevel: System
         """
-        # Create custom product and repository
+        # setup custom repo
+        cv_name = gen_string('alpha')
         product = make_product({'organization-id': module_org.id})
         repo = make_repository(
             {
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_satellitesync.py
```diff
@@ -16,95 +16,113 @@
...
+                'organization-id': module_org.id,
+                'repository-id': repo['id'],
+            }
+        )
+        ContentView.publish({'id': cv['id']})
+        cv = ContentView.info({'id': cv['id']})
+        assert len(cv['versions']) == 1
+        cvv = cv['versions'][0]
+        # Verify export directory is empty
+        result = ssh.command(f'find {export_dir} -name "*.json"')
```
mshriver, line 95: Please use `default_sat` fixture provided instance to call `default_sat.execute` instead of continuing to rely on the ssh module.  Given the scope of this change in this PR, it should be a very fast and straightforward change.

tests/foreman/cli/test_satellitesync.py
```diff
@@ -16,95 +16,113 @@
...
+        assert len(result.stdout) == 0
+        result = ssh.command(f'find {export_dir} -name "*.tar.gz"')
+        assert len(result.stdout) == 0
+        # Export content view
+        ContentExport.completeVersion({'id': cvv['id'], 'organization-id': module_org.id})
+        # Verify export directory is not empty
+        result = ssh.command(f'find {export_dir} -name "*.json"')
+        assert result.return_code == 0
+        assert len(result.stdout) > 0
```
mshriver, line 104: `> 0` is redundant when asserting against `len()`

tests/foreman/cli/test_satellitesync.py
```diff
@@ -170,42 +301,114 @@ def test_positive_export_rh_product(self, class_export_directory, module_org):
...
+        ContentView.add_repository(
+            {
+                'id': cv['id'],
+                'organization-id': my_org['id'],
+                'repository-id': repo['id'],
+            }
+        )
+        ContentView.publish({'id': cv['id']})
+        # Verify export directory is empty
+        result = ssh.command(f'find {cv_export_dir} -name "*.json"')
```
mshriver, line 388: These 4/5 lines for asserting against the `find`s are repeated several times. Consider defining a function in the test module that takes the `default_sat` instance and directory as arguments and makes the assertions against the output and status.

tests/foreman/cli/test_satellitesync.py
```diff
@@ -230,20 +433,19 @@ def class_export_entities():
...
     yield export_dir
     # Deletes directory created for CV export Test
-    ssh.command(f'rm -rf {export_dir}')
+    ssh.command(f'rm -rf {export_dir}/{module_org.name}')
 
 
-def get_export_base():
-    return '/var/lib/pulp/katello-export'
+def get_import_base():
+    return '/var/lib/pulp/imports/'
```
mshriver, line 445: This really doesn't need to be a function.

tests/foreman/cli/test_satellitesync.py
```diff
@@ -230,20 +433,19 @@ def class_export_entities():
 
 
 @pytest.fixture(scope='function')
-def function_export_cv_directory():
-    """Create directory for CV export and at the end remove it"""
-    export_dir = f'{get_export_base()}/{gen_string("alpha")}'
-    ssh.command(f'mkdir {export_dir}')
+def export_cv_dir(module_org):
+    """Get directory for CV export and at the end remove it"""
+    export_dir = '/var/lib/pulp/exports/'
```
mshriver, line 438: Why not ensure the directory exists via mkdir, as was done previously?

latran: Previously, you have to create some directory and put the exports in.  New implementation puts export files in `/var/lib/pulp/exports/` by default.

latran:
trigger: robottelo
pytest: tests/foreman/cli/test_satellitesync.py

latran:
trigger: test-robottelo
pytest: tests/foreman/cli/test_satellitesync.py

latran:
There's one failure:

```
=================================== FAILURES ===================================
20:46:56  _ TestContentViewSync.test_positive_export_import_redhat_cv_with_huge_contents _
```

There was an issues with enabling rhel7 repo.  I'll investigate and open a separate PR to fix this, but this does not impact the overall PR.

tests/foreman/cli/test_satellitesync.py
```diff
@@ -230,20 +384,22 @@ def class_export_entities():
...
     yield export_dir
     # Deletes directory created for CV export Test
-    ssh.command(f'rm -rf {export_dir}')
+    default_sat.execute(f'rm -rf {export_dir}/{module_org.name}')
 
 
-def get_export_base():
-    return '/var/lib/pulp/katello-export'
+def validate_filepath(default_sat, import_export_cv_dir):
```
mshriver, line 395: Nice, reduces repetition of the find pattern.

tests/foreman/cli/test_satellitesync.py
```diff
@@ -415,58 +635,92 @@ def test_positive_export_import_default_org_view(
...
         importing_cvv = ContentView.info(
             {'name': importing_cv_name, 'organization-id': importing_org['id']}
         )['versions']
         assert len(importing_cvv) >= 1
         imported_packages = Package.list({'content-view-version-id': importing_cvv[0]['id']})
-        assert len(imported_packages) > 0
-        assert len(exported_packages) == len(imported_packages)
+        assert len(imported_packages)
+        assert len(cv_packages) == len(imported_packages)
+        # reset disconnected mode and download policy
```
mshriver, line 715: This `reset` action needs to be done in a finalizer, set immediately after the setting is modified.

There are several actions/assertions in the test that could cause an exception - meaning the setting is never restored. This causes very difficult to debug test failures, because a failure in this test case impacts the satellite instance for the remainder of the test session.

I would recommend using the decorator implementation of registering a finalizer, since its more than one statement.

```
@request.addfinalizer
def _settings_reset():
    Settings.set({'name': "default_download_policy", 'value': "on_demand"})
    Settings.set({'name': 'content_disconnected', 'value': "No"})
```

Furthermore, it would be best to read the original setting, instead of assuming and hard coding a value that's the default at the time of writing the test.

mshriver (changes requested):
The pattern for resetting modified settings needs to be revisited, where the individual comment was made and in all of the other test functions that used the same pattern.

Consider writing a function scoped fixture to read settings, yield, and then restore the original values for a given set of settings fields.  That way you don't have to repeat the finalizer function in every test case that needs it.

latran:
trigger: test-robottelo
pytest: tests/foreman/cli/test_satellitesync.py

latran:
trigger: test-robottelo
pytest: tests/foreman/cli/test_satellitesync.py

tests/foreman/cli/test_satellitesync.py
```diff
@@ -16,95 +16,121 @@
...
-    assert result.return_code == 0
-    assert len(result.stdout) >= 1
-    assert 'unconfined_u:object_r:foreman_var_run_t:s0' in result.stdout[0]
-
-    # Update the 'pulp_export_destination' settings to new directory
-    Settings.set({'name': 'pulp_export_destination', 'value': f'/mnt/{export_dir}'})
-    # Create an organization to reuse in tests
-    yield export_dir
-    ssh.command(f'rm -rf /mnt/{export_dir}')
+    """Grab export directory at the end remove the export directory with
```
mshriver, line 50: Maybe I'm just reading the diff wrong, but this function has no content other than the docblock.

mshriver: This fixture is still used by multiple test functions, so I think you've got some missing content here.

tests/foreman/cli/test_satellitesync.py
```diff
@@ -16,95 +16,121 @@
...
+@pytest.fixture(scope='class')
+def config_export_import_settings():
+    """Check settings and set download policy for export.  Reset to original state after import"""
+    download_policy_value = Settings.info({'name': 'default_download_policy'})['value']
+    rh_download_policy_value = Settings.info({'name': 'default_redhat_download_policy'})['value']
+    disconnected_mode_value = Settings.info({'name': 'content_disconnected'})['value']
+    assert download_policy_value == 'immediate'
+    assert rh_download_policy_value == 'on_demand'
+    assert disconnected_mode_value == 'false'
+    Settings.set({'name': 'default_download_policy', 'value': 'immediate'})
```
mshriver, line 63: Why do you call `set` on this when it's asserted to be `immediate` already on L60?

tests/foreman/cli/test_satellitesync.py
```diff
@@ -16,95 +16,121 @@
...
+        cv = ContentView.info({'id': cv['id']})
+        assert len(cv['versions']) == 1
+        cvv = cv['versions'][0]
+        # Verify export directory is empty
+        assert validate_filepath(default_sat, export_dir) == ''
+        # Export content view
+        ContentExport.completeVersion({'id': cvv['id'], 'organization-id': module_org.id})
+        # Verify export directory is not empty
+        assert validate_filepath(default_sat, export_dir) != ''
+        # test cleanup: remove org dir in export dir
```
mshriver, line 116: Test cleanup should be done in finalizers, not in the test function itself. Maybe the impact is low here if cleanup doesn't run, but its the same comment that I made before - if any of the statements raise exceptions or assertions fail during the test, this cleanup won't run.

mshriver: If you use the fixtures for org creation instead of creating them in the test function by calling `make_org()` (other test functions, this is already using `module_org`), you can write a fixture that yields and then does this cleanup.

`export_dir` is defined several times in this module, always to the same path. That can be defined as a capsule class attribute, or as a constant in the module - that way the cleanup function doesn't have to hard code it.

tests/foreman/cli/test_satellitesync.py
```diff
@@ -230,20 +400,22 @@ def class_export_entities():
...
-    """Create directory for CV export and at the end remove it"""
-    export_dir = f'{get_export_base()}/{gen_string("alpha")}'
-    ssh.command(f'mkdir {export_dir}')
+def export_cv_dir(default_sat, module_org):
+    """Get directory for CV export and at the end remove it"""
+    export_dir = '/var/lib/pulp/exports/'
     yield export_dir
     # Deletes directory created for CV export Test
-    ssh.command(f'rm -rf {export_dir}')
+    default_sat.execute(f'rm -rf {export_dir}/{module_org.name}')
```
mshriver, line 408: there's an extra `/` in the path here. It shouldn't negatively effect the behavior, but there are more consistent ways to construct these paths using `pathlib.PurePath` instances.

tests/foreman/cli/test_satellitesync.py
latran: Must have missed that, I'll delete since that docblock functionality is in another method.

tests/foreman/cli/test_satellitesync.py
latran: It's suppose to be `rh_download_policy_value` that gets set, not `download_policy_value`.

latran:
trigger: test-robottelo
pytest: tests/foreman/cli/test_satellitesync.py

latran:
trigger: test-robottelo
pytest: tests/foreman/cli/test_satellitesync.py

tests/foreman/cli/test_satellitesync.py
```diff
@@ -232,20 +381,30 @@ def class_export_entities(default_sat):
...
+    """Deletes dir for version export"""
+    yield
     # Deletes directory created for CV export Test
-    default_sat.execute(f'rm -rf {export_dir}')
+    assert default_sat.execute(f'rm -rf {EXPORT_DIR}/{module_org.name}').stdout == ''
 
 
-def get_export_base():
-    return '/var/lib/pulp/katello-export'
+def validate_filepath(default_sat, EXPORT_DIR):
```
mshriver, line 399: I don't think you mean to pass `EXPORT_DIR` here, since its now defined at the module level.

latran:
trigger: test-robottelo
pytest: tests/foreman/cli/test_satellitesync.py

latran:
Only two failures that involved redhat repo sync but I can open a separate pr for those issues.

mshriver (approved):
Definitely needs a review from someone more familiar with the component, but the pytest pattern changes here are looking great.

tests/foreman/cli/test_satellitesync.py
```diff
@@ -16,97 +16,117 @@
 
 :Upstream: No
 """
-import json
+import os.path
```
JacobCallahan, line 19: why not pathlib's Path?

tests/foreman/cli/test_satellitesync.py
```diff
@@ -16,97 +16,117 @@
...
-    """Raise when export Directory is not set or found"""
+@pytest.fixture(scope='class')
+def config_export_import_settings():
+    """Check settings and set download policy for export.  Reset to original state after import"""
+    download_policy_value = Settings.info({'name': 'default_download_policy'})['value']
+    rh_download_policy_value = Settings.info({'name': 'default_redhat_download_policy'})['value']
+    disconnected_mode_value = Settings.info({'name': 'content_disconnected'})['value']
+    assert download_policy_value == 'immediate'
+    assert rh_download_policy_value == 'on_demand'
+    assert disconnected_mode_value == 'false'
```
JacobCallahan, line 59: what is the purpose of these assertions?

tests/foreman/cli/test_satellitesync.py
```diff
@@ -232,20 +382,30 @@ def class_export_entities(default_sat):
...
     # Deletes directory created for CV export Test
-    default_sat.execute(f'rm -rf {export_dir}')
+    assert default_sat.execute(f'rm -rf {EXPORT_DIR}/{module_org.name}').stdout == ''
 
 
-def get_export_base():
-    return '/var/lib/pulp/katello-export'
+def validate_filepath(default_sat):
+    """Checks the existence of certain files in a dir"""
+    result = default_sat.execute(
```
JacobCallahan, line 402: i'd recommend changing the name from "default_sat" to something that doesn't match a fixture name. this will make it less confusing for those that might think this is accepting the default_sat fixture. this would also be something a user could use against another non-default satellite object.

tests/foreman/cli/test_satellitesync.py
```diff
@@ -271,28 +431,32 @@ def _create_cv(cv_name, repo, organization, publish=True):
     return content_view, cvv_id
 
 
-def _enable_rhel_content(organization, repo_name, releasever=None, product=None, sync=True):
+def _enable_rhel_content(
+    default_sat, module_org, repo_name, releasever=None, product=None, sync=True
```
JacobCallahan, line 435: same comment about using fixture names as argument names for non-fixture functions :)

tests/foreman/cli/test_satellitesync.py
```diff
@@ -271,28 +431,32 @@ def _create_cv(cv_name, repo, organization, publish=True):
...
     """Enable and/or Synchronize rhel content
 
     :param organization: The organization directory into which the rhel
         contents will be enabled
     :param bool sync: Syncs contents to repository if true else doesnt
     :return: Repository cli object
     """
-    manifests.upload_manifest_locked(organization['id'], interface=manifests.INTERFACE_CLI)
+    with manifests.clone() as manifest:
+        default_sat.put(manifest, manifest.filename)
```
JacobCallahan, line 445: great use of the new manifest pattern!

tests/foreman/cli/test_satellitesync.py
```diff
@@ -950,141 +1148,75 @@ def test_negative_reimport_cv_with_same_major_minor(
...
-                    'organization-id': imported_entities['importing_org']['id'],
-                }
-            )
+            ContentImport.version({'organization-id': importing_org['id'], 'path': import_path})
         assert (
-            f'''the Content View '{class_export_entities["exporting_cv_name"]}' '''
-            'is greater or equal to the version you are trying to import'
+            f'''Content View Version specified in the metadata - '{export_cv_name} 1.0' '''
+            'already exists. If you wish to replace the existing version, '
+            f'''delete {export_cv_name} 1.0 and try again.'''
```
JacobCallahan, line 1188: This is something leftover from before, and seen a few other times in this file. Mainly an annoyance
```suggestion
            f'Content View Version specified in the metadata - "{export_cv_name} 1.0" '
            'already exists. If you wish to replace the existing version, '
            f'delete {export_cv_name} 1.0 and try again.'
```

JacobCallahan (changes requested):

tests/foreman/cli/test_satellitesync.py
```diff
@@ -1098,141 +1230,34 @@ def test_negative_export_cv_with_on_demand_repo(self, function_export_cv_directo
 
         :expectedresults:
 
-            1. Export fails with error 'All exported repositories must be set to an immediate
-                download policy and re-synced' should be displayed.
-
+            1. Export fails with error '"message": "NOTE: Unable to fully
+                export this version because it contains repositories without the
+                'immediate' download policy. Update the download policy and sync
+                affected repositories. Once synced republish the content view and
```
swadeley, line 1236:
```suggestion
                affected repositories. Once synced, republish the content view and
```

swadeley (approved):
ACK pending typos

tests/foreman/cli/test_satellitesync.py
latran: Was not aware of pathlib so used os.path for path manipulation.  I can try to implement pathlib and see if that fits my needs for some of the path extraction.

tests/foreman/cli/test_satellitesync.py
latran: It's a check for ensuring that those are the default sets for each of those fields.  Tests can either fail ASAP or I can add conditional statements to set it to default if it's not.  What are you thoughts?

tests/foreman/cli/test_satellitesync.py
mshriver: I'd consider this optional for the scope of this PR, but would also strongly recommend moving to using `PurePath` for your use-case in this module.

tests/foreman/cli/test_satellitesync.py
latran: I kept the single quote for `{export_cv_name} 1.0` to align with the error message so this can pass.

tests/foreman/cli/test_satellitesync.py
JacobCallahan: It is optional, but recommended :D

tests/foreman/cli/test_satellitesync.py
JacobCallahan: I personally don't think it matters what the settings were before and, since it likely doesn't matter for your tests, shouldn't be a fail condition for yours

tests/foreman/cli/test_satellitesync.py
JacobCallahan: ok, then can you instead use double quotes in place of the triple single quotes? :)

latran:
@JacobCallahan All changes should be made now except for purepath :)

JacobCallahan (approved):

PR #8938, opened 2021-09-03
Removed puppet modules in contentviewversion

latran:
Puppet modules are being removed and in some cases replaced with docker to use for testing.

```
pytest tests/foreman/api/test_contentviewversion.py 
====================== 16 passed, 3 deselected, 332 warnings in 682.45s (0:11:22) =================
```

@devendra104 can you please review the deleted test cases for upgrade?  I don't think it's needed, but can't hurt to have a second opinion.  Same with you @tstrych

### DIFF
--- a/tests/foreman/api/test_contentviewversion.py
+++ b/tests/foreman/api/test_contentviewversion.py
@@ -16,8 +16,6 @@
 
 :Upstream: No
 """
-import random
-
 import pytest
 from fauxfactory import gen_string
 from nailgun import entities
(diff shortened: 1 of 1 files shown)


jameerpathan111:
@latran should we update the test steps of stubbed tests as well?

tstrych (approved):
ACK

tstrych:
I agree with Jameer, stubbed test comments should be updated as well. This is non-blocking for this PR.

mshriver (approved):

PR #8941, opened 2021-09-06
Automated stubbed test case for Products feature

adarshdubey-star:
**Test Results - 6.10.0**

```
pytest tests/foreman/cli/test_product.py -k test_positive_product_sync_state
================================== test session starts=========================================
platform linux -- Python 3.8.6, pytest-6.2.5, py-1.10.0, pluggy-0.13.1
plugins: services-2.2.1, mock-3.6.1, reportportal-5.0.8, forked-1.3.0, xdist-2.3.0, ibutsu-1.16, cov-2.12.1
collected 17 items / 16 deselected / 1 selected                                                                                                                                                                   

tests/foreman/cli/test_product.py s  

========================1 skipped, 16 deselected, 3 warnings in 2.18s ===========================                          
```

### DIFF
--- a/tests/foreman/cli/test_product.py
+++ b/tests/foreman/cli/test_product.py
@@ -18,7 +18,9 @@
 """
 import pytest
 from fauxfactory import gen_alphanumeric
+from fauxfactory import gen_integer
 from fauxfactory import gen_string
+from fauxfactory import gen_url
 
 from robottelo import ssh
 from robottelo.api.utils import wait_for_tasks
@@ -318,14 +320,14 @@ def test_positive_assign_http_proxy_to_products(module_org):
         assert r['http-proxy']['http-proxy-policy'] == 'none'
 
 
-@pytest.mark.stubbed
 @pytest.mark.tier2
-def test_positive_product_sync_state():
+@pytest.mark.skip_if_open('BZ:1999541')
+def test_positive_product_sync_state(module_org):
     """hammer product info shows correct sync state.
 
     :id: 58af6239-85d7-4b8b-bd2d-ab4cd4f29840
 
-    :BZ: 1803207
+    :BZ: 1803207,1999541
 
     :customerscenario: true
 
@@ -335,6 +337,36 @@ def test_positive_product_sync_state():
         3. Successfully sync another repository under the same product.
         4. Run `hammer product info --product-id <id>` again.
 
+
     :expectedresults: hammer should show 'Sync Incomplete' in both cases.
     """
-    pass
+    product = make_product({'organization-id': module_org.id})
+    repo_a1 = make_repository(
+        {
+            'organization-id': module_org.id,
+            'product-id': product['id'],
+            'name': gen_string('alpha'),
+            'url': f'{gen_url(scheme="https")}:{gen_integer(min_value=10, max_value=9999)}',
+        }
+    )
+
+    with pytest.raises(CLIReturnCodeError):
+        Repository.synchronize({'id': repo_a1['id']})
+
+    product_info = Product.info({'id': product['id'], 'organization-id': module_org.id})
+    product_list = Product.list({'organization-id': module_org.id})
+    assert product_info['sync-state-(last)'] in [p.get('sync-state') for p in product_list]
+
+    repo_a2 = make_repository(
+        {
+            'organization-id': module_org.id,
+            'product-id': product['id'],
+            'name': gen_string('alpha'),
+            'url': settings.repos.yum_0.url,
+        },
+    )
+
+    Repository.synchronize({'id': repo_a2['id']})
+    product_info = Product.info({'id': product['id'], 'organization-id': module_org.id})
+    product_list = Product.list({'organization-id': module_org.id})
+    assert product_info['sync-state-(last)'] in [p.get('sync-state') for p in product_list]


omkarkhatavkar:

tests/foreman/cli/test_product.py
```diff
@@ -318,23 +320,52 @@ def test_positive_assign_http_proxy_to_products(module_org):
...
+
+    repo_a2 = make_repository(
+        {
+            'organization-id': module_org.id,
+            'product-id': product['id'],
+            'name': gen_string('alpha'),
+            'url': settings.repos.yum_0.url,
+        },
+    )
+    Repository.synchronize({'id': repo_a2['id']})
```
shweta83, line 368: Please handle the exception here  as well.

tests/foreman/cli/test_product.py
```diff
@@ -318,23 +320,52 @@ def test_positive_assign_http_proxy_to_products(module_org):
...
 
+    :CaseImportance: Critical
+
     :Steps:
         1. Sync a custom repository that fails.
         2. Run `hammer product info --product-id <id>`.
         3. Successfully sync another repository under the same product.
         4. Run `hammer product info --product-id <id>` again.
 
+
```
shweta83, line 342: Please add "CaseLevel" here too.

tests/foreman/cli/test_product.py
akhil-jha: Yes, adding exception is a good thing, but I wonder if it is really needed here(or at L354). I mean `Repository.synchronize` should already have an error handling within.

tests/foreman/cli/test_product.py
adarshdubey-star: Yes , will add it

tests/foreman/cli/test_product.py
```diff
@@ -318,23 +320,52 @@ def test_positive_assign_http_proxy_to_products(module_org):
...
     """hammer product info shows correct sync state.
 
     :id: 58af6239-85d7-4b8b-bd2d-ab4cd4f29840
 
-    :BZ: 1803207
+    :BZ: 1803207,1999541
 
     :customerscenario: true
 
+    :CaseImportance: Critical
```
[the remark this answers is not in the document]
adarshdubey-star: Okay will update it.

tests/foreman/cli/test_product.py
shweta83: It can be skipped if `Repository.synchronize` is already handling the exception  but it would be good to add before `Product.info`.

tests/foreman/cli/test_product.py
```diff
@@ -318,23 +320,55 @@ def test_positive_assign_http_proxy_to_products(module_org):
...
     """hammer product info shows correct sync state.
 
     :id: 58af6239-85d7-4b8b-bd2d-ab4cd4f29840
 
-    :BZ: 1803207
+    :BZ: 1803207,1999541
 
     :customerscenario: true
 
+    :CaseLevel: Component
```
mshriver, line 334: These aren't strictly necessary, this metadata is specified in the module level docblock (L7 and L15), which will apply to all classes and functions within the module.

adarshdubey-star: Okay will update it.

tests/foreman/cli/test_product.py
```diff
@@ -335,6 +337,34 @@ def test_positive_product_sync_state():
...
+            'product-id': product['id'],
+            'name': gen_string('alpha'),
+            'url': f'{gen_url(scheme="https")}:{gen_integer(min_value=10, max_value=9999)}',
+        }
+    )
+    with pytest.raises(CLIReturnCodeError):
+        Repository.synchronize({'id': repo_a1['id']})
+    product_info = Product.info({'id': product['id'], 'organization-id': module_org.id})
+    product_list = Product.list({'organization-id': module_org.id})
+    assert product_info['sync-state-(last)'] == product_list[0]['sync-state']
```
akhil-jha, line 356: I am doubtful at this assertion.
I **think** you're assuming the value at `[0]` is the product you created in L343. Will that always be the case? If you create multiple product(like L343) and then do a `Product.list()`, will `[0]` be the one you want to assert?

adarshdubey-star: Actually, it's about multiple repositories in a single product. And asserting the information of its state. Hence used two things to check it out and verify from product Information as well as a product list.

akhil-jha: Right. But what I am concerned about is the value in `product_list`. Are you sure the product you want to assert is at `[0]`? Could it be at `[1]` or `[2]`?

adarshdubey-star: Yes, as there is a dictionary in the list its needs an index to be accessed, provided with the respective key. About [1] and [2] it will basically fail because of index error as there is only one element in the list the dictionary itself.

akhil-jha: My bad. I was unclear. I didn't take in account that this is a new organization and there is only one product created.

akhil-jha (approved):

tests/foreman/cli/test_product.py
mshriver: @akhil-jha I think you might have a point here, that this specific index can easily make this test flaky. The organization is coming from a module scoped fixture, and its entirely possible that other test cases in this module have created products under the org.

I think you're right that the assertion should be changed to look at sync-state from all of the products.

```
assert product_info['sync-state-(last)'] in [p.get('sync-state') for p in product_list]
```

adarshdubey-star: Got it!

tests/foreman/cli/test_product.py
```diff
@@ -335,6 +337,34 @@ def test_positive_product_sync_state():
...
+    product = make_product({'organization-id': module_org.id})
+    repo_a1 = make_repository(
+        {
+            'organization-id': module_org.id,
+            'product-id': product['id'],
+            'name': gen_string('alpha'),
+            'url': f'{gen_url(scheme="https")}:{gen_integer(min_value=10, max_value=9999)}',
+        }
+    )
+    with pytest.raises(CLIReturnCodeError):
```
mshriver, line 352: Which of the 4 statements do you expect to raise a `CLIReturnCodeError`?

I suspect its not the last statement, the assertion, which suggests some of these lines aren't executing during the test.

The `with` catches and handles the exception type passed to it, but doesn't change the fact that the lines inside of its context after the exception is raised do not execute. When the expected exception occurs, the context is exited.

The line which raises the exception should always be the last line of the `pytest.raises` context.

https://docs.pytest.org/en/6.2.x/reference.html#pytest-raises

mshriver (changes requested):
Need to be more explicit in `pytest.raises` context use

tests/foreman/cli/test_product.py
adarshdubey-star: On L(353) when the sync occurs as it's a dummy URL the metadata cannot be fetched and surely will improvise the pytest.raises with that only.

mshriver:
trigger: test-robottelo
pytest: tests/foreman/cli/test_product.py -k test_positive_product_sync_state

mshriver (approved):
Looks good on inspection, holding merge on results of Robottelo-runner

PR #8948, opened 2021-09-09
Remove puppet from contentview in cli

latran:
Removal of all puppet related test cases that uses puppet content.

Results:
```
pytest tests/foreman/cli/test_contentview.py

==================== 1 failed, 71 passed, 9 skipped, 6 deselected, 35 warnings, 1 error in 6766.33s (1:52:46) =================== 
```

The one failure has another PR which I will revisit, fix, and close out.

@devendra104 Please review that one upgrade test case that I removed.  I think it's ok to remove since it uses puppet content.  @tstrych Please also review as well.

### DIFF
--- a/tests/foreman/cli/test_contentview.py
+++ b/tests/foreman/cli/test_contentview.py
@@ -40,7 +40,6 @@ from robottelo.cli.location import Location
 from robottelo.cli.module_stream import ModuleStream
 from robottelo.cli.org import Org
 from robottelo.cli.product import Product
-from robottelo.cli.puppetmodule import PuppetModule
 from robottelo.cli.repository import Repository
 from robottelo.cli.repository_set import RepositorySet
 from robottelo.cli.role import Role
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_contentview.py
```diff
@@ -3892,50 +3455,26 @@ def test_positive_remove_cv_version_from_multi_env_capsule_scenario(
...
-        )
-        assert len(puppet_modules) > 0
-        puppet_module = puppet_modules[0]
-        # Setup a docker repo
         docker_product = cli_factory.make_product({'organization-id': module_org.id})
         docker_repository = cli_factory.make_repository(
             {
                 'content-type': 'docker',
-                'docker-upstream-name': 'busybox',
+                'docker-upstream-name': constants.CONTAINER_UPSTREAM_NAME,
```
tstrych, line 3462: please provide information, why docker upstream name is changing here, respectively why `busybox` was used before

tstrych (approved):
ACK

tests/foreman/cli/test_contentview.py
latran: I'm assuming `busybox` was used before because it returns content after syncing.  As I was testing this, I found that that's no longer the case and using `library/busybox` does return content after syncing and switched over to using the new name change to better reflect the update.

mshriver (approved):

PR #8400, opened 2021-03-04
Add SELinux and systemd checks for pulp3 filesystem

swadeley:
Hello

Some low level checks to ensure the new filesystem for Pulp3 is ready for the migration from Pulp2

PR #8949, opened 2021-09-09
Replace symlinks checks in contentmanagement

vsedmik:
Pulp3 does not use the symlinks nor the on-disk directory structure that pulp2 used for content anymore. Instead it uses "artifacts" which hold the content and repo metadata, but they are hidden from the user (using hashes for filenames and dirs).
So in this PR I replaced the symlink checks with checks for the artifact existence. I feel it's less deterministic, I'm not happy about that, but I can't see any better solution right now.

I would appreciate your input @latran @swadeley

Test results:
```
(venv39) [vsedmik@localhost robottelo]$ pytest tests/foreman/api/test_contentmanagement.py::TestCapsuleContentManagement::test_positive_on_demand_sync tests/foreman/api/test_contentmanagement.py::TestCapsuleContentManagement::test_positive_update_with_immediate_sync tests/foreman/api/test_contentmanagement.py::TestCapsuleContentManagement::test_positive_uploaded_content_library_sync
=========================================================================================================== test session starts ===========================================================================================================
platform linux -- Python 3.9.6, pytest-6.2.5, py-1.10.0, pluggy-0.13.1
rootdir: /home/vsedmik/PycharmProjects/robottelo, configfile: pyproject.toml
plugins: forked-1.3.0, services-2.2.1, xdist-2.3.0, reportportal-5.0.8, ibutsu-1.16, mock-3.6.1
collected 3 items                                                                                                                                                                                                                         

tests/foreman/api/test_contentmanagement.py ...                                                                                                                                                                                     [100%]

=============================================================================================== 3 passed, 3 warnings in 2973.05s (0:49:33) ================================================================================================
```

### DIFF
--- a/robottelo/helpers.py
+++ b/robottelo/helpers.py
@@ -366,6 +366,26 @@ def form_repo_path(org=None, lce=None, cv=None, cvv=None, prod=None, repo=None, 
     return os.path.join(PULP_PUBLISHED_YUM_REPOS_PATH, repo_path)
 
 
+def form_repo_url(capsule, org=None, lce=None, cv=None, prod=None, repo=None):
+    """Forms url of a repo or CV published on a Satellite or Capsule.
+
+    :param object capsule: Capsule or Satellite object providing its url
+    :param str org: organization label
+    :param str lce: lifecycle environment label
+    :param str cv: content view label
+    :param str prod: product label
+    :param str repo: repository label
+    :return: url of the specific repo or CV
+    """
+    if not all([capsule, org, prod, repo]):
+        raise ValueError('`capsule`, `org`, `prod` and `repo` arguments are required')
+
+    if lce and cv:
+        return f'{capsule.url}/pulp/content/{org}/{lce}/{cv}/custom/{prod}/{repo}/'
+    else:
+        return f'{capsule.url}/pulp/content/{org}/Library/custom/{prod}/{repo}/'
+
+
 def create_repo(name, repo_fetch_url=None, packages=None, wipe_repodata=False, hostname=None):
     """Creates a repository from given packages and publishes it into pulp's
     directory for web access.
--- a/robottelo/host_info.py
+++ b/robottelo/host_info.py
@@ -2,6 +2,7 @@
 import functools
 import re
 
+import requests
 from packaging.version import Version
 from ssh2.exceptions import AuthenticationError
 
@@ -99,30 +100,51 @@ def get_repo_files(repo_path, extension='rpm', hostname=None):
     return sorted(repo_file for repo_file in result.stdout if repo_file)
 
 
-def get_repomd_revision(repo_path, hostname=None):
-    """Fetches a revision of repository.
+def get_repo_files_by_url(url, extension='rpm'):
+    """Returns a list of repo files (for example rpms) in a specific repository
+    published at some url.
 
-    :param str repo_path: unix path to the repo, e.g. '/var/lib/pulp/fooRepo'
-    :param str optional hostname: hostname or IP address of the remote host. If
-        ``None`` the hostname will be get from ``main.server.hostname`` config.
+    :param url: url where the repo or CV is published
+    :param extension: extension of searched files. Defaults to 'rpm'
+    :return:  list representing rpm package names
+    """
+    if not url.endswith('/'):
+        url += '/'
+
+    result = requests.get(url, verify=False)
+    if result.status_code != 200:
+        raise requests.HTTPError(f'{url} is not accessible')
+
+    links = re.findall(r'(?<=href=").*?(?=">)', result.text)
+
+    if 'Packages/' not in links:
+        return sorted(line for line in links if extension in line)
+
+    files = []
+    subs = get_repo_files_by_url(f'{url}Packages/', extension='/')
+    for sub in subs:
+        files.extend(get_repo_files_by_url(f'{url}Packages/{sub}', extension))
+
+    return sorted(files)
+
+
+def get_repomd_revision(repo_url):
+    """Fetches a revision of a repository.
+
+    :param str repo_url: the 'Published_At' link of a repo
     :return: string containing repository revision
     :rtype: str
     """
     repomd_path = 'repodata/repomd.xml'
-    result = ssh.command(
-        f"grep -oP '(?<=<revision>).*?(?=</revision>)' {repo_path}/{repomd_path}",
-        hostname=hostname,
-    )
-    # strip empty lines
-    stdout = [line for line in result.stdout if line]
-    if result.return_code != 0 or len(stdout) != 1:
-        raise CLIReturnCodeError(
-            result.return_code,
-            result.stderr,
-            f'Unable to fetch revision for {repo_path}. Please double check your '
-            'hostname, path and contents of repomd.xml',
-        )
-    return stdout[0]
+    result = requests.get(f'{repo_url}/{repomd_path}', verify=False)
+    if result.status_code != 200:
+        raise requests.HTTPError(f'{repo_url}/{repomd_path} is not accessible')
+
+    match = re.search('(?<=<revision>).*?(?=</revision>)', result.text)
+    if not match:
+        raise ValueError(f'<revision> not found in {repo_url}/{repomd_path}')
+
+    return match.group(0)
 
 
 class SatVersionDependentValues:
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -983,9 +983,9 @@ class Capsule(ContentHost):
                 f'{settings.robottelo.tmp_dir}/capsule-{self.ip_addr}.log',
             )
             raise CapsuleHostError(f'foreman installer failed at capsule host: {result.stderr}')
-        result = self.execute('systemctl status pulp_celerybeat.service')
+        result = self.execute('satellite-maintain service status')
         if 'inactive (dead)' in '\n'.join(result.stdout):
-            raise CapsuleHostError('pulp_celerybeat service not running')
+            raise CapsuleHostError('a core service is not running at capsule host')
 
 
 class Satellite(Capsule):
--- a/tests/foreman/api/test_contentmanagement.py
+++ b/tests/foreman/api/test_contentmanagement.py
@@ -33,11 +33,12 @@ from robottelo.api.utils import upload_manifest
 from robottelo.config import settings
 from robottelo.helpers import create_repo
 from robottelo.helpers import form_repo_path
+from robottelo.helpers import form_repo_url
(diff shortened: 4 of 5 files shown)


robottelo/hosts.py
```diff
@@ -983,9 +983,9 @@ def capsule_setup(self, sat_host=None, **installer_kwargs):
                 f'{settings.robottelo.tmp_dir}/capsule-{self.ip_addr}.log',
             )
             raise CapsuleHostError(f'foreman installer failed at capsule host: {result.stderr}')
-        result = self.execute('systemctl status pulp_celerybeat.service')
+        result = self.execute('satellite-maintain service status')
         if 'inactive (dead)' in '\n'.join(result.stdout):
-            raise CapsuleHostError('pulp_celerybeat service not running')
+            raise CapsuleHostError('some of the core service is not running')
```
swadeley, line 988:
```suggestion
            raise CapsuleHostError('a core service is not running')
```

swadeley:
Hello @vsedmik 

We have new basic check that Pulp3 file system structure is setup[1] . Beside the many repo and CV tests we have always had, and continue to add to, we  now have new errata enumeration tests that use Default Content View.  While those new errata tests are designed to exercise katello's errata enumeration logic they would not work if the Pulp3 file system was not working to store the packages and their updates. However, those new tests are not on an external Capsule, so your additional checks are not redundant.

Do we have a test that deletes one of the artefacts and then does an advanced sync to ensure the artefact is recreated? That would be a useful addition.

[1] https://github.com/SatelliteQE/robottelo/pull/8400

vsedmik:
> We have new basic check that Pulp3 file system structure is setup[1] . Beside the many repo and CV tests we have always had, and continue to add to, we now have new errata enumeration tests that use Default Content View. While those new errata tests are designed to exercise katello's errata enumeration logic they would not work if the Pulp3 file system was not working to store the packages and their updates. However, those new tests are not on an external Capsule, so your additional checks are not redundant.

Well, I didn't want to say that the Pulp3 file system does not work to store the content, it surely does. I just meant that it can't be checked directly in the Pulp2-ish manner - like check on-disk /var/lib/pulp/published/... directory for the symlinks named just the same like the upstream packages, count them, check if they are broken or not (on_demand vs. immediate). With Pulp3 we have just those anonymous artefacts on disk.

> Do we have a test that deletes one of the artefacts and then does an advanced sync to ensure the artefact is recreated? That would be a useful addition.

I don't think so, not yet.

tests/foreman/api/test_contentmanagement.py
```diff
@@ -659,26 +644,19 @@ def test_positive_iso_library_sync(self, capsule_configured):
...
 
-        :expectedresults:
-
-            1. After initial syncing only symlinks are present on both
-               satellite and capsule, no real packages were fetched.
-            2. All the symlinks are pointing to non-existent files.
-            3. Attempt to download package is successful
-            4. Downloaded package checksum matches checksum of the source
-               package
+        :expectedresults: multiple assertions across the code pass
```
latran, line 653: This is too general.  If would be best to include what expectedresults are here so anyone reading this would quickly know.

mshriver: @vsedmik These steps and expected results are to help a reader (think new component owner) understand what the test automation in the function is doing, and checking. It also serves as a guide if for some reason the test needs to be manually stepped through.

We certainly don't expect you to repeat the automation assertions word for word, and expected results should actually align with the `:steps:` token, which was missing here.

Please update the expectedresults that were here to describe what you're now asserting in the test.

mshriver (approved):

vsedmik:
I should be able to parse the published url and get the content and revision from there, which would allow me to get some of the checks back to the game. So turning to Draft until explored and implemented or refused.

vsedmik:
Results:
```
(venv39) [vsedmik@localhost robottelo]$ pytest \
> tests/foreman/api/test_contentmanagement.py::TestCapsuleContentManagement::test_positive_capsule_sync \
> tests/foreman/api/test_contentmanagement.py::TestCapsuleContentManagement::test_positive_on_demand_sync \
> tests/foreman/api/test_contentmanagement.py::TestCapsuleContentManagement::test_positive_update_with_immediate_sync \
> tests/foreman/api/test_contentmanagement.py::TestCapsuleContentManagement::test_positive_uploaded_content_library_sync
=========================================================================================================== test session starts ===========================================================================================================
platform linux -- Python 3.9.6, pytest-6.2.5, py-1.10.0, pluggy-0.13.1
rootdir: /home/vsedmik/PycharmProjects/robottelo, configfile: pyproject.toml
plugins: forked-1.3.0, services-2.2.1, xdist-2.3.0, reportportal-5.0.8, ibutsu-1.16, mock-3.6.1
collected 4 items                                                                                                                                                                                                                         

tests/foreman/api/test_contentmanagement.py ....                                                                                                                                                                                    [100%]

=============================================================================================== 4 passed, 3 warnings in 3883.34s (1:04:43) ================================================================================================
```

latran (approved):
My issue is addressed: ACK

robottelo/helpers.py
```diff
@@ -366,6 +366,18 @@ def form_repo_path(org=None, lce=None, cv=None, cvv=None, prod=None, repo=None,
     return os.path.join(PULP_PUBLISHED_YUM_REPOS_PATH, repo_path)
 
 
+def form_repo_url(hostname=None, org=None, lce=None, cv=None, prod=None, repo=None):
```
mshriver, line 369: Please avoid referencing settings.server.hostname as a default, writing a new helper with this old pattern.

Make an instance of `satellite` a required kwarg, and use `satellite.url` property instead of `f'https://{hostname}'`

robottelo/host_info.py
```diff
@@ -99,28 +99,52 @@ def get_repo_files(repo_path, extension='rpm', hostname=None):
...
-    :param str repo_path: unix path to the repo, e.g. '/var/lib/pulp/fooRepo'
-    :param str optional hostname: hostname or IP address of the remote host. If
-        ``None`` the hostname will be get from ``main.server.hostname`` config.
+    :param url: url where the repo or CV is published
+    :param extension: extension of searched files. Defaults to 'rpm'
+    :return:  list representing rpm package names
+    """
+    if not url.endswith('/'):
+        url += '/'
+    result = ssh.command(
```
mshriver, line 112: Why run a curl command on the satellite server, instead of using the `requests` library to make the GET request from the python process this is running under?

robottelo/host_info.py
```diff
@@ -99,28 +99,52 @@ def get_repo_files(repo_path, extension='rpm', hostname=None):
...
+
+    :param str repo_url: the 'Published_At' link of a repo
     :return: string containing repository revision
     :rtype: str
     """
     repomd_path = 'repodata/repomd.xml'
     result = ssh.command(
-        f"grep -oP '(?<=<revision>).*?(?=</revision>)' {repo_path}/{repomd_path}",
-        hostname=hostname,
+        f"curl -k {repo_url}/{repomd_path} | grep -oP '(?<=<revision>).*?(?=</revision>)'",
```
mshriver, line 138: This would be a good opportunity to make this request directly, instead of using ssh to run curl from the target satellite.

tests/foreman/api/test_contentmanagement.py
```diff
@@ -659,26 +627,26 @@ def test_positive_iso_library_sync(self, capsule_configured):
...
 
         :expectedresults:
 
-            1. After initial syncing only symlinks are present on both
-               satellite and capsule, no real packages were fetched.
-            2. All the symlinks are pointing to non-existent files.
-            3. Attempt to download package is successful
-            4. Downloaded package checksum matches checksum of the source
-               package
+            1. A custom yum repository is successfully synced and ContentView published
```
mshriver, line 638: Thanks for cleaning up these statements and making them accurate for the updated test.

robottelo/host_info.py
vsedmik: yes, of course, palmface at myself

robottelo/host_info.py
vsedmik: same palmface here

robottelo/helpers.py
vsedmik: We use this for both, Satellite and Capsule, so I'll pass the Capsule instead. Sat inherits from Caps anyway.

mshriver: ack - yep! The important part is using a Cap/Sat instance and its hostname property.

vsedmik:
Tests rerun after latest changes:
`=== 4 passed, 4 warnings in 4107.61s (1:08:27) ===`

robottelo/helpers.py
```diff
@@ -366,6 +366,26 @@ def form_repo_path(org=None, lce=None, cv=None, cvv=None, prod=None, repo=None,
...
+
+    :param object capsule: Capsule or Satellite object providing its url
+    :param str org: organization label
+    :param str lce: lifecycle environment label
+    :param str cv: content view label
+    :param str prod: product label
+    :param str repo: repository label
+    :return: url of the specific repo or CV
+    """
+    if not all([capsule, org, prod, repo]):
```
mshriver, line 380: This type of validation is in a way built into the language itself.

These arguments are `optional` because they have a default assignment in the function signature - `org=None`.

By omitting the default assignment (like the `capsule` kwarg), calling the function without one of its required args/kwargs will produce a `TypeError` describing the keyword or positional argument that's missing. 

You'll want to resort to checks like this only when specific combinations of arguments are required - for example `repo` is required only when `prod` is not empty.

I won't consider this a blocker as its not harmful - its just there's a more pythonic way of expressing the requirement to pass these arguments for the function to work.

robottelo/host_info.py
```diff
@@ -99,30 +100,51 @@ def get_repo_files(repo_path, extension='rpm', hostname=None):
...
+    :return:  list representing rpm package names
+    """
+    if not url.endswith('/'):
+        url += '/'
+
+    result = requests.get(url, verify=False)
+    if result.status_code != 200:
+        raise requests.HTTPError(f'{url} is not accessible')
+
+    links = re.findall(r'(?<=href=").*?(?=">)', result.text)
```
mshriver, line 118: Nice work, was interested to see how you dealt with processing this without `grep` being an option.

mshriver (approved):

robottelo/helpers.py
vsedmik: Thank you for the explanation! So in this case I could use just `def form_repo_url(capsule, org, prod, repo, lce=None, cv=None):` and remove the `if not all` condition, right?

Ok, I'll update it in the next PR.

PR #8961, opened 2021-09-13
Add ibutsu custom metadata

mshriver:
component field in metadata is treated specially in ibutsu

### DIFF
--- a/pytest_plugins/metadata_markers.py
+++ b/pytest_plugins/metadata_markers.py
@@ -57,7 +57,18 @@ assignee_regex = re.compile(
 
 @pytest.hookimpl(tryfirst=True)
 def pytest_collection_modifyitems(session, items, config):
-    """Add markers for testimony tokens"""
+    """Add markers and user_properties for testimony token metadata
+
+    user_properties is used by the junit plugin, and thus by many test report systems
+    Handle test function/class/module/session scope metadata coming from test docblocks
+    Apply user_properties, ibutsu metadata, and pytest markers
+
+    Markers for metadata use the testimony token name as the mark name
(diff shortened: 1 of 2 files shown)


pytest_plugins/metadata_markers.py
```diff
@@ -100,6 +111,16 @@ def pytest_collection_modifyitems(session, items, config):
             ("start_time", datetime.datetime.utcnow().strftime(FMT_XUNIT_TIME))
         )
 
+        # add custom ibutsu metadata fields for test case grouping and heatmaps
+        if hasattr(item, "_ibutsu"):
+            item._ibutsu["data"]["metadata"].update(
+                {
+                    # TODO Work with ibutsu team, better mechanism for defining 'special' data
+                    # TODO Add sat version to this item data at test execution time
+                    "component": item.get_closest_marker('component').args[0]
```
omkarkhatavkar, line 120: will this help in search the components in ibutsu ? Along with this can we also include the username (as we have that in testimony token as well )

mshriver: user_properties are already stored on the test results in ibutsu under metadata, allowing for filtering by assignee (or any of the other values that are added to user_properties through this hook already.

Ibutsu treats `component` and `env` as special metadata values, and requires them to be set for Jenkins analysis and heatmap widgets to render.  This just sets the field explicitly so these widgets will start to work for us.

I'm working with the ibutsu team to also re-work the widget dependency on these fields, and to add more filtering capability to the UI.

jameerpathan111 (approved):
LGTM

JacobCallahan (approved):

PR #8966, opened 2021-09-14
ui host gce fix

tstrych:
I would like to say thanks to @peterdragun  who take care of gce tests.
This is just a small follow up. 

Fixture `remove_vm_on_delete` is added.
It changes the default setting from false to yes.
So we clean up the gce vm too once we delete the host.
But on the other hand I have to add run in one thread so other tests can run at the same time if the vm should stay. Not sure if test like that exists. I went with safe approach here. If you want me to remove the run in one thread marker let me know.

fixture is function based, so in any other time is something running in parallel the `remove_vm_on_delete` is set to false.

The test `test_positive_delete_libvirt` have not been tested, that's another chapter, but I just saw same principles so I fixed it also there. 

I can't remember last time when these tests where green :) 
`pytest tests/foreman/ui/test_host.py -k test_positive_gce_`
Screenshot from 2021-09-14 17-24-26

### DIFF
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -49,6 +49,7 @@ from robottelo.cli.job_invocation import JobInvocation
 from robottelo.cli.proxy import Proxy
 from robottelo.cli.scap_policy import Scappolicy
 from robottelo.cli.scapcontent import Scapcontent
+from robottelo.cli.settings import Settings
 from robottelo.config import settings
 from robottelo.constants import ANY_CONTEXT
 from robottelo.constants import DEFAULT_ARCHITECTURE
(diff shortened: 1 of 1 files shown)


mshriver (approved):
I see your concern about the fixture modifying a setting that applies beyond the scope of the test.

run_in_one_thread is the appropriate marker, though the marker hook implementation really needs some updating to account for the xdist worker alignment strategy. We no longer use the marker to restrict test execution in CI, as our xdist strategy effectively implements run-in-one-thread behavior by having unique satellite targets for each worker.

jameerpathan111 (approved):

peterdragun (approved):

tstrych:
3 approvals should be enough :stuck_out_tongue:  merging this :smile:

PR #8972, opened 2021-09-15
Raise DependantTestFailed error in post_upgrade if pre_upgrade test failed

jyejare:
Raise the `DependantTestFailed` error for the post_upgrade test if its dependant pre_upgrade test fails. So that the test will be marked as `blocked` in polarion instead of skipped.

### DIFF
--- a/tests/upgrades/conftest.py
+++ b/tests/upgrades/conftest.py
@@ -11,7 +11,7 @@ This module is intended to be used for upgrade tests, that have two run stages,
     Second stage: post_upgrade::
 
         The test marked as post upgrade will be run, any saved data in dependent pre_upgrade stage
-        can be restored, if a dependent test fail, the post upgrade test will be skipped
+        can be restored, if a dependent test fail, the post upgrade test will be errored
 
         pytest -m "post_upgrade" tests/foreman/upgrade
 
@@ -231,6 +231,17 @@ def pre_upgrade_data(request):
     return data
 
 
+def pytest_configure(config):
+    """Register custom markers to avoid warnings."""
+    markers = [
+        "pre_upgrade: Mark tests to run before upgrade.",
+        "post_upgrade(depend_on=None): Mark tests to run after upgrade.",
+        "fail: Mark test to fail if dependant test fails.",
+    ]
+    for marker in markers:
+        config.addinivalue_line("markers", marker)
+
+
 def pytest_sessionstart(session):
     """Do some setup for automation-tools and satellite6-upgrade"""
     # Fabric Config setup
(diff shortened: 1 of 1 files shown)


tests/upgrades/conftest.py
```diff
@@ -11,7 +11,7 @@
     Second stage: post_upgrade::
 
         The test marked as post upgrade will be run, any saved data in dependent pre_upgrade stage
-        can be restored, if a dependent test fail, the post upgrade test will be skipped
+        can be restored, if a dependent test fail, the post upgrade test will be errored
```
mshriver, line 14: failed

tests/upgrades/conftest.py
```diff
@@ -342,11 +341,20 @@ def pytest_collection_modifyitems(items, config):
...
+    pass
+
+
+@pytest.fixture(autouse=True)
+def fail_xfail_test(request):
+    """Called after fail mark added on post_upgrade tests using above collection modification"""
+    fail = request.node.get_closest_marker('fail')
+    if fail:
+        reason = fail.kwargs.get('reason')
+        raise DependentTestFailed(reason)
```
mshriver, line 360: I think it would be better to call `pytest.fail` here, not end the test in an ERROR during setup.

tests/upgrades/conftest.py
```diff
@@ -342,11 +341,20 @@ def pytest_collection_modifyitems(items, config):
...
+
+class DependentTestFailed(Exception):
+    """Raise when the dependent test fails"""
+    pass
+
+
+@pytest.fixture(autouse=True)
+def fail_xfail_test(request):
+    """Called after fail mark added on post_upgrade tests using above collection modification"""
+    fail = request.node.get_closest_marker('fail')
```
mshriver, line 357: marks need to be registered for pytest to not throw warnings about them.

tests/upgrades/conftest.py
jyejare: No, pytest doesn't allow me to fail the test from the fixture.

tests/upgrades/conftest.py
jyejare: As said, pytest doesn't allow me to fail the test from the test fixture. Even if I try to call `pytest.fail()`, pytest shows the test as error :
`E      Failed: The dependant test tests/testing/test_ups.py::TestSomethingWorks::test_postthing failed!`

tests/upgrades/conftest.py
jyejare: Done!

tests/upgrades/conftest.py
```diff
@@ -342,11 +352,20 @@ def pytest_collection_modifyitems(items, config):
...
+                                reason=f'The dependant test {item.nodeid} failed!'))
+
+
+class DependentTestFailed(Exception):
+    """Raise when the dependent test fails"""
+    pass
+
+
+@pytest.fixture(autouse=True)
+def fail_xfail_test(request):
```
mshriver, line 366: Let's give this a more descriptive name, since `xfail` isn't actually in use here or what this fixture is doing. Consider `post_upgrade_dependant_fail`

mshriver:
Looks like there's an extra file, maybe from testing?

tests/upgrades/conftest.py
jyejare: But raising an explicit exception provides a better granularity than just a plain `failed` error, and it would also help in ibutsu to filter tests with `DependentTestFailed` error!

So I would better stick to the explicit error!

jyejare:
@mshriver Removed the extra file, please revisit !

mshriver (approved):

jyejare:
@SatelliteQE/airgun-tier-1-reviewers Please review!

PR #8976, opened 2021-09-15
Use broker's new interactive shell to replace old paramiko functionality

JacobCallahan:
This change switches out the remaining paramiko-related code for our new
method of using host objects. Specifically it uses a host object's
session to create new interactive shell instances.


mshriver (approved):

JacobCallahan:
Requires #8977

rplevka (approved):

PR #8991, opened 2021-09-23
[Deep Sleep] Upgrade Scenarios docstrings moved to class

jyejare:
Moved all Upgrade Scenario docstrings to a class level so that we can push one scenario as a one test case to polarion using https://github.com/jyejare/betelgeuse/pull/1 .

### DIFF
--- a/tests/upgrades/conftest.py
+++ b/tests/upgrades/conftest.py
@@ -21,7 +21,7 @@ This module is intended to be used for upgrade tests, that have two run stages,
 
         from upgrade_tests import pre_upgrade, post_upgrade
 
-        @pre_upgrade
+        @pytest.mark.pre_upgrade
         def test_capsule_pre_upgrade(save_test_data):
             '''A pre upgrade test that should pass and save data'''
             assert 1 == 1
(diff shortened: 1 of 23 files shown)


mshriver (changes requested):
Per our discussions, I thought we were not going down this path of collecting and applying test metadata of the class separately for some paths.

jyejare:
@mshriver I raised this PR after discussion with @vijay8451 , he understands and agrees with the strategy to push the upgrade scenarios from class docstrings. 

This decision is purely made on POC basis since Polarion is used purely as a test management tool and hence we are good with this strategy. @vijay8451 was one of the contributors in upgrade scenarios so he understands the architecture of upgrade scenarios. If management later sees any issues with class strategy, we will go back to the per def basis strategy. But I dont see that will ever happen.

latran:
@JacobCallahan can you provide feedback for this based on Mike's comment?

JacobCallahan:
@latran @mshriver this is something @jyejare and I talked about a while ago and will be revisited at a later time.

jyejare:
@latran yeah, for time being we would like to keep this in draft until we decide that #9226 isn't doing the job well :)

JacobCallahan:
Closing this for now, but this can be revisited if needed in the future.

PR #8997, opened 2021-09-27
Remove rhevm cases in virt-who configure plugin

hkx303:
**Description**
[Bug 1930740](https://bugzilla.redhat.com/show_bug.cgi?id=1930740) was verified:
rhevm provider for the virt-who was deprecated and is no longer supported, for RHV4 hypervisors it's recommended to use the libvirt mode and deploy virt-who locally to the hypervisor. We should drop the rhevm option in the configuration plugin.
So, we needn't these rhevm cases


omkarkhatavkar (approved):
ACK

mshriver (approved):

PR #8993, opened 2021-09-23
Consolidate module_loc and module_org fixtures

mshriver:
rename to module_location

create new shared smart_proxy_location module

remove all the duplicate fixture defintions, mostly in the UI test
modules

PR #8999, opened 2021-09-28
Missed module_loc changes

mshriver:
These were lost in the rebase of #8993 from unstable to master branches, with some changes having gone into the auth fixtures in the meanwhile.


swadeley (approved):
ACK

PR #9000, opened 2021-09-28
Update `ncat` use, address hanging pytest

mshriver:
In CI pipelines I observed several xdist workers hung on various test cases - all tracing back to `default_url_on_new_port` usage.

Through debugging, I found `ncat` use was causing the channel to hang on close, which was mitigated by backgrounding the process.  Even with backgrounding the channel running ncat will hang if read, so I added a bit of logic to identify the specific pid. I then ensure this pid is killed.

Local testing easily replicated the hang, and I was able to resolve the hanging behavior.  This PR will not address test failures in the impacted module that aren't related to the ncat tunnel directly. There are some issues with tests impacted by this hang that are unrelated.

I created a shared fixture wrapping `default_url_on_new_port`, and applied it to the CLI realm test module. This fixture can be applied to two other test modules using default_url_on_new_port directly. Again not including that in the scope of this PR which aims to resolve the hung xdist workers and unblock CI pipelines.

### DIFF
--- a/tests/foreman/cli/test_realm.py
+++ b/tests/foreman/cli/test_realm.py
@@ -21,271 +21,268 @@ import random
 import pytest
 from fauxfactory import gen_string
 
-from robottelo.cleanup import capsule_cleanup
 from robottelo.cleanup import realm_cleanup
 from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.factory import CLIFactoryError
-from robottelo.cli.factory import make_proxy
 from robottelo.cli.factory import make_realm
 from robottelo.cli.realm import Realm
 
 
-@pytest.fixture(scope='module')
-def _make_proxy(request):
-    """Create a Proxy and register the cleanup function"""
-    proxy = make_proxy()
-    request.addfinalizer(lambda: capsule_cleanup(proxy['id']))
-    return proxy
+pytestmark = [pytest.mark.run_in_one_thread]
 
 
-@pytest.mark.run_in_one_thread
-class TestRealm:
-    """Tests for Realms via Hammer CLI, must be run on QE Satellite Server.
-    Requires enroll_idm() and configure_realm() to configure the test environment."""
+@pytest.mark.tier1
+def test_positive_delete_by_name(module_fake_proxy):
+    """Realm deletion by realm name
 
-    @pytest.mark.tier1
-    def test_positive_delete_by_name(self, _make_proxy):
-        """Realm deletion by realm name
+    :id: ef3967e6-d53d-4dec-b74f-c20448d5fc6d
 
-        :id: ef3967e6-d53d-4dec-b74f-c20448d5fc6d
+    :expectedresults: Realm is deleted
+    """
+    realm = make_realm({'realm-proxy-id': module_fake_proxy.id, 'realm-type': 'Active Directory'})
+    Realm.delete({'name': realm['name']})
+    with pytest.raises(CLIReturnCodeError):
+        Realm.info({'id': realm['id']})
 
-        :expectedresults: Realm is deleted
-        """
-        realm = make_realm({'realm-proxy-id': _make_proxy['id'], 'realm-type': 'Active Directory'})
-        Realm.delete({'name': realm['name']})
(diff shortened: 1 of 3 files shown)


tests/foreman/cli/test_realm.py
```diff
@@ -25,17 +25,32 @@
...
     """Create a Proxy and register the cleanup function"""
-    proxy = make_proxy()
-    request.addfinalizer(lambda: capsule_cleanup(proxy['id']))
-    return proxy
+    args = {'name': gen_string(str_type='alpha')}
+    newport = get_available_capsule_port()
+    try:
+        with default_url_on_new_port(9090, newport) as url:
+            args['url'] = url
+            logger.debug('creating proxy object')
```
mshriver, line 47: at the moment this is the last line I see in the debug, and then it gets stuck at a very odd place and I never see the log message from inside `create_object`. 

The traceback makes it look like it went straight over these calls and is trying to do cleanup.

```Traceback (most recent call first):
  File "/home/mshriver/repos/satqe/robottelo/tests/foreman/cli/test_realm.py", line 51, in _make_proxy
    capsule_cleanup(proxy['id'])
  <built-in method next of module object at remote 0x7fe919160090>
  File "/home/mshriver/repos/satqe/robottelo/.robottelo/lib64/python3.8/site-packages/_pytest/fixtures.py", line 1181, in call_fixture_func
```

mshriver:
trigger: test-robottelo
pytest: tests/foreman/cli/test_realm.py tests/foreman/cli/test_capsule.py::test_positive_refresh_features_by_id tests/foreman/cli/test_organization.py tests/foreman/cli/test_location.py tests/foreman/api/test_location.py

tests/foreman/cli/test_realm.py
```diff
@@ -21,70 +21,62 @@
...
         """Realm deletion by realm ID
 
         :id: 7c1aca0e-9724-40de-b38f-9189bdae0514
 
         :expectedresults: Realm is deleted
         """
-        proxy = _make_proxy
-        realm = make_realm({'realm-proxy-id': proxy['id'], 'realm-type': 'Active Directory'})
+        realm = make_realm(
+            {'realm-proxy-id': module_fake_proxy['id'], 'realm-type': 'Active Directory'}
```
JacobCallahan, line 60: Why item access here when you used attribute access above?

JacobCallahan (changes requested):
Very great investigative work! Just one question for you.

tests/foreman/cli/test_realm.py
mshriver: ah! didn't complete my copy-paste before marking this RFR, thanks!

mshriver: Updated

mshriver:
trigger: test-robottelo
pytest: tests/foreman/cli/test_realm.py tests/foreman/cli/test_capsule.py::test_positive_refresh_features_by_id tests/foreman/cli/test_organization.py tests/foreman/cli/test_location.py tests/foreman/api/test_location.py

JacobCallahan (approved):
Nice!

mshriver:
trigger: test-robottelo
pytest: tests/foreman/cli/test_realm.py tests/foreman/cli/test_capsule.py::test_positive_refresh_features_by_id tests/foreman/cli/test_organization.py tests/foreman/cli/test_location.py tests/foreman/api/test_location.py

mshriver:
The 9 failures in build 318 for robottelo-runner aren't related to this commit.

tests/foreman/cli/test_realm.py
```diff
@@ -21,271 +21,268 @@
...
+
+    :expectedresults: Realm not found
+    """
+    with pytest.raises(CLIReturnCodeError):
+        Realm.info({'name': gen_string('alpha', random.randint(1, 30))})
+
+
+@pytest.mark.tier2
+def test_negative_info_nonexistent_realm_id():
+    """Get info for a realm with an ID that does not exists
```
swadeley, line 281:
```suggestion
    """Get info for a realm with an ID that does not exist
```

latran (approved):
ACK

PR #9005, opened 2021-09-29
Fix katello agent tests to run on different satellite

Gauravtalreja1:
Katello-agent is disabled for 6.10 and can be enabled by satellite-installer flags, so moving these tests to run on different module-level satellite

Moving tests to respective modules which are not related to katello-agent, @swadeley @tstrych request to review.

Also fixes - https://github.com/SatelliteQE/robottelo/issues/8852

Signed-off-by: Gaurav Talreja <gtalreja@redhat.com>

PR #9006, opened 2021-09-29
katello agent is no longer enabled by default

tstrych:
that leads to:
 - `wait_for_errata_applicability_task` is not longer working because there are no longer shown `Actions::Katello::Host::<name_of_the_action>`  actions shown in task manager
 - when tested locally passed 7/7 times, but it is possible that errata will no be applied before the next commad/assert will ran
 -  the test was failing with `This action uses katello-agent, which is currently disabled. Use remote execution instead`, so I listened and I used remote execution instead :smile: 

I looked at other occurrences in the code, afterwards spoke with @Gauravtalreja1 and he helped me to identify the issue and he also knows that the tests in `test_katello_agent.py` should be reviewed. EDIT: are already in review #9005 is the proof of it :+1: 

```
collected 56 items / 55 deselected / 1 selected
tests/foreman/cli/test_host.py::test_positive_install_package_via_rex[rhel7_contenthost0]
========== 1 passed, 55 deselected, 5 warnings in 1487.45s (0:24:47) ===========
```

### DIFF
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -16,7 +16,6 @@
 
 :Upstream: No
 """
-import time
 from datetime import datetime
 from datetime import timedelta
 from random import choice
@@ -31,7 +30,6 @@ from fauxfactory import gen_string
 from nailgun import entities
 
 from robottelo.api.utils import promote
-from robottelo.api.utils import wait_for_errata_applicability_task
 from robottelo.cli.activationkey import ActivationKey
 from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.factory import add_role_permissions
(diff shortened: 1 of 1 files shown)


tstrych:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py::test_positive_install_package_via_rex tests/foreman/cli/test_host.py::test_positive_erratum_applicability

mshriver (approved):

tests/foreman/cli/test_host.py
```diff
@@ -1880,21 +1875,36 @@ def test_positive_erratum_applicability(katello_host_tools_client):
 
 @pytest.mark.katello_host_tools
 @pytest.mark.tier3
-def test_negative_install_package(katello_host_tools_client):
-    """Attempt to install a package to a host remotely
+def test_positive_install_package_via_rex(
```
JacobCallahan, line 1878: is this not a test we already have?

tstrych: changed the description, 
and we have [one test](https://github.com/SatelliteQE/robottelo/blob/6a596f61b60e546efa55f488177ea41722e7c9b9/tests/foreman/cli/test_remoteexecution.py#L351) that installs multiple packages the same way I did, but no test that is using `hammer host package list` 

To differentiate the test even more I update the job invocation command, instead of `job-template` I used `remote-execution-feature`

JacobCallahan (approved):

PR #9008, opened 2021-09-29
Improve test cleanup

spusateri:
This PR is meant to improve test cleanup. Currently, the test fails when running against the same Satellite multiple times, this is because the repo, product and orgs created by the fixtures are not deleted after the test finishes. 

In addition, there is an open BZ [#009069](https://bugzilla.redhat.com/show_bug.cgi?id=2009069) causing this test to fail, so I have added the 'skip_if_open' pytest mark until it is resolved.

### DIFF
--- a/tests/foreman/ui/test_containerimagetag.py
+++ b/tests/foreman/ui/test_containerimagetag.py
@@ -47,6 +47,7 @@ def module_repository(module_product):
     return repo
 
 
+@pytest.mark.skip_if_open("BZ:2009069")
 @pytest.mark.tier2
 def test_positive_search(session, module_org, module_product, module_repository):
     """Search for a docker image tag and reads details of it
@@ -61,8 +62,8 @@ def test_positive_search(session, module_org, module_product, module_repository)
     with session:
         session.organization.select(org_name=module_org.name)
         search = session.containerimagetag.search('latest')
-        assert module_product.name == search[0]['Product Name']
-        assert module_repository.name == search[0]['Repository Name']
+        assert module_product.name in [i['Product Name'] for i in search]
+        assert module_repository.name in [i['Repository Name'] for i in search]
         values = session.containerimagetag.read('latest')
         assert module_product.name == values['details']['product']
         assert module_repository.name == values['details']['repository']


tests/foreman/ui/test_containerimagetag.py
```diff
@@ -27,12 +27,16 @@
 
 @pytest.fixture(scope="module")
 def module_org():
-    return entities.Organization().create()
+    org = entities.Organization().create()
+    yield org
+    org.delete()
```
swadeley, line 32: Deleting orgs can cause race conditions, this is not safe. Deleting anything just waits time given we do not keep the Satellite after testing.

tests/foreman/ui/test_containerimagetag.py
```diff
@@ -27,12 +27,16 @@
...
 
 
 @pytest.fixture(scope="module")
 def module_product(module_org):
-    return entities.Product(organization=module_org).create()
+    product = entities.Product(organization=module_org).create()
+    yield product
+    product.delete()
```
swadeley, line 39: Deleting anything just waits time given we do not keep the Satellite after testing.

tests/foreman/ui/test_containerimagetag.py
```diff
@@ -44,9 +48,11 @@ def module_repository(module_product):
         url=CONTAINER_REGISTRY_HUB,
     ).create()
     repo.sync()
-    return repo
+    yield repo
+    repo.delete()
```
swadeley, line 52: Deleting anything just waits time given we do not keep the Satellite after testing.

swadeley (changes requested):
Deleting anything just waits time given we do not keep the Satellite after testing.

tests/foreman/ui/test_containerimagetag.py
mshriver: When the org is module scoped, what is the race condition this test module would hit during module teardown?

tests/foreman/ui/test_containerimagetag.py
mshriver: What would you suggest @spusateri do instead of cleaning up entities if he's seeing test instability if this test module is executed more than once?

swadeley: yes random names?

spusateri: Hi, I have confirmed that we are using random names here by running multiple times and viewing within the Satellite. Since we are using the entities `EntitiesCreateMixin`, and name is a require value, a random name will be assigned when create_missing() is called even though I am not passing a specific random name as with `url` or `content_type`.

tests/foreman/ui/test_containerimagetag.py
spusateri: I think there is a happy medium here, as the test will only fail when run multiple times for either product or repo, not organization, so we can remove the yield for organization and use the same org over consecutive runs. 

However, I think the point regarding race conditions should be discussed further by the team so that we have a clear standard going forward. In theory, each test should be agnostic to one another, meaning they are responsible for their own setup and teardown. For example, a test should not assume the org it is using is not being used by another test, rather, that test should specifically create or select an org that is specific that that test or module. I understand there is some nuance and tradeoffs that must be accounted for in the way we distribute tests in CI, so I think this is an important point that should be worked out by the team so we have a more clear standard going forward

tests/foreman/ui/test_containerimagetag.py
mshriver: I think this indicates a bit more failure analysis is necessary to identify what's flaky and actually failing. If there's a conflict in satellite with this repository type, this cleanup should absolutely be included. Otherwise, there may be an entirely different path and pattern we need to address.

mshriver:
@spusateri Please rebase and squash out the merge commit and then it looks like this can be marked 'Ready for review'

spusateri:
trigger: test-robottelo pytest: tests/foreman/ui/test_containerimagetag.py::test_positive_search

spusateri:
trigger: test-robottelo 
pytest: tests/foreman/ui/test_containerimagetag.py::test_positive_search

mshriver (approved):

latran (approved):
ACK

PR #9017, opened 2021-09-30
Remove no longer valid cache_clear call

JacobCallahan:
The get_client function is no longer cached, so the method is no longer
attached.
Also moved broker's file logging to debug

### DIFF
--- a/robottelo/ssh.py
+++ b/robottelo/ssh.py
@@ -1,6 +1,4 @@
 """Utility module to handle the shared ssh connection."""
-from ssh2 import exceptions
-
 from robottelo.cli import hammer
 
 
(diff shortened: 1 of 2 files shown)


mshriver (approved):

robottelo/ssh.py
```diff
@@ -50,18 +50,7 @@ def command(
...
-        # If the client is in a bad state, invalidate the cache and get a new one
-        get_client.cache_clear()
-        client = get_client(
-            hostname=hostname,
-            username=username,
-            password=password,
-            port=port,
-        )
-        result = client.execute(cmd, timeout=timeout)
+    result = client.execute(cmd, timeout=timeout)
```
mshriver, line 53: Approved, but gotta remove the unused exceptions import

JacobCallahan: ahhh true!

PR #9023, opened 2021-10-01
Add UI test for package group filtering

swadeley:
Hello

I could not add hammer test due to a bug[1] I found while attempting to do that, so its just the UI test for now.

Test relates to 6.10 user stores.

[1] [Bug 1982288 - Accept package group ID or name when creating CV filter rule](https://bugzilla.redhat.com/show_bug.cgi?id=1982288)


swadeley:
trigger: test-robottelo
pytest: tests/foreman/ui/test_contentview.py -k test_positive_filter_by_pkg_group_name

tests/foreman/ui/test_contentview.py
```diff
@@ -4123,3 +4123,44 @@ def test_positive_depsolve_with_module_errata(session, module_org):
...
+
+    :CaseImportance: Medium
+    """
+
+    filter_name = gen_string('alpha')
+    repo_name = gen_string('alpha')
+    package_group = 'birds'
+    expected_packages = (('cockateel'), ('duck'), ('penguin'), ('stork'))
+    create_sync_custom_repo(module_org.id, repo_name=repo_name)
+    repo = entities.Repository(name=repo_name).search(query={'organization_id': module_org.id})[0]
```
mshriver, line 4145: Consider using `default_sat.api` to access these nailgun entity classes.

tests/foreman/ui/test_contentview.py
```diff
@@ -4123,3 +4123,44 @@ def test_positive_depsolve_with_module_errata(session, module_org):
...
+        session.contentviewfilter.add_package_group(cv.name, filter_name, package_group)
+        cvf = session.contentviewfilter.read(cv.name, filter_name)
+        assert cvf['content_tabs']['assigned'][0]['Name'] == package_group
+        # Publish CV and check filter worked
+        session.contentview.publish(cv.name)
+        result = session.contentview.read_version(cv.name, VERSION)
+        # Assert we do not have more than the expected number of packages
+        assert len(result['rpm_packages']['table']) == 4
+        # Assert the package name are the expected ones
+        for items in result['rpm_packages']['table']:
```
mshriver, line 4165: Consider making a more direct comparison here that provides a bit more context on failure.

Removing the for loop, this also makes checking against a hardcoded length number redundant, since the list equality will ensure that there are no extraneous or missing packages.
```
assert expected_packages == [pkg['Name'] for pkg in result['rpm_packages']['table']]
```

mshriver (approved):
Two non-blocking suggestions for improvement, otherwise looks good. Any BZ's to link to this test case?

tests/foreman/ui/test_contentview.py
swadeley: OK, doing it now

tests/foreman/ui/test_contentview.py
swadeley: HI, `expected_packages` is a tuple (same as some other tests), so I was thinking `expected_packages = tuple([pkg['Name'] for pkg in result['rpm_packages']['table']])`, but then it seems neater just to change `expected_packages` to a list. Doing that now.

swadeley:
> Two non-blocking suggestions for improvement, otherwise looks good. Any BZ's to link to this test case?
no, was in a list of user stories but no BZ.

thanks for the review

swadeley:
trigger: test-robottelo
pytest: tests/foreman/ui/test_contentview.py -k test_positive_filter_by_pkg_group_name

tests/foreman/ui/test_contentview.py
```diff
@@ -4123,3 +4123,43 @@ def test_positive_depsolve_with_module_errata(session, module_org):
...
+            },
+        )
+        session.contentviewfilter.add_package_group(cv.name, filter_name, package_group)
+        cvf = session.contentviewfilter.read(cv.name, filter_name)
+        assert cvf['content_tabs']['assigned'][0]['Name'] == package_group
+        # Publish CV and check filter worked
+        session.contentview.publish(cv.name)
+        result = session.contentview.read_version(cv.name, VERSION)
+        # Assert only the expected packages are present
+        assert expected_packages == [pkg['Name'] for pkg in result['rpm_packages']['table']]
```
omkarkhatavkar, line 4165: Is the packages are alphabetically always ? and is that important as well ? you can convert it them to set make it brittle 
`
assert set(expected_packages) == set([pkg['Name'] for pkg in result['rpm_packages']['table']])`

PR #9028, opened 2021-10-05
test is no longer blocked by BZ

tstrych:
`================== 1 passed, 4 warnings in 284.89s (0:04:44) ===================`

Airgun [PR](https://github.com/SatelliteQE/airgun/pull/625)


tstrych:
trigger: test-robottelo
pytest: tests/foreman/ui/test_hardwaremodel.py
airgun: 625

tests/foreman/ui/test_hardwaremodel.py
```diff
@@ -67,10 +66,8 @@ def test_positive_end_to_end(session, module_org, module_location):
         host_values = session.host.read(host_name, 'additional_information')
         assert host_values['additional_information']['hardware_model'] == new_name
         # Make an attempt to delete hardware model that associated with host
-        with pytest.raises(AssertionError) as context:
-            session.hardwaremodel.delete(new_name)
-        assert f"error: '{new_name} is used by {host_name}'" in str(context.value)
+        session.hardwaremodel.delete(new_name, cancel_dialog=True)
```
swadeley, line 69: I wonder why you no longer check there is an error message

mshriver: See the airgun PR, it introduced the `cancel_dialog` behavior that consumes the flash message that previously caused an AssertionError.

mshriver: I'm not convinced this change is one that we want to make though.

The BZ was closed as WONTFIX, so this is behavior that the test should now expect.  I don't think silently consuming the flash message within the delete method is the right approach, just so that the test doesn't have a raises clause.

I think its important that the test continues to assert against the content of the flash message that is shown when the hardware model is deleted when attached to a host.

The airgun PR does make an important change though - clicking `cancel` when the delete call fails, so that the page isn't 'stuck' after this.

tstrych:
trigger: test-robottelo
pytest: tests/foreman/ui/test_hardwaremodel.py
airgun: 625

tstrych:
trigger: test-robottelo
pytest: tests/foreman/ui/test_hardwaremodel.py
airgun: 625

mshriver (approved):

PR #9039, opened 2021-10-07
small fixes in cli test_host

tstrych:
for `test_positive_dump_enc_yaml` haven't found why this has changed, but hammer commands are parsed differently in robottelo, what has been list is now just a string

response.stdout of hammer command also looks different, that's why the yaml parsing has changed too

I would be happier if I know what has introduced this change because right now it feels I want to merge it kind of blindly.


tstrych:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py::test_positive_set_multi_line_and_with_spaces_parameter_value tests/foreman/cli/test_host.py::test_positive_dump_enc_yaml

PR #9043, opened 2021-10-08
Read rhel_compose_id parameter from the content_host conf

ogajduse:
> This PR is a piece needed to have PIT automation done

## Problem statement

For RHEL candidate deployments we use a different AT workflow than it is used for content hosts in the standard pipelines.
Currently, there is no way to set the `rhel_compose_id` parameter that is needed for the *deploy_rhel_cmp_template*.

## Solution
The patch included :slightly_smiling_face:


ogajduse:
trigger: test-robottelo
pytest: pytest tests/foreman/api/test_subscription.py::test_positive_subscription_status_disabled[rhel_8]

PR #9044, opened 2021-10-08
assign satellite_fixture marker to tests using sat fixtures

rplevka:
similar to `-m "content_host"`, this commit should implement an auto marker, that is assigned on test items, using any of the satellite fixtures (the ones that spawn satellites) - `-m satellite_fixture`.


jyejare:
@rplevka Is this still relevant   or should we just close ??

PR #9070, opened 2021-10-15
customerscenario check

pondrejk:


pondrejk:
@JacobCallahan @mshriver 
Hi, still no luck regarding the bz key secret, I access it like this in the gh workflow:

```  
env:
  ROBOTTELO_BUGZILLA__API_KEY: '${{ secrets.BUGZILLA_KEY }}'
```
Which doesn't propagate to the script, I tried to use both the dynaconf setting and sourcing it directly from env:

```  
bz_key = settings.bugzilla.api_key
click.echo(len(bz_key))
click.echo(len(os.environ["ROBOTTELO_BUGZILLA__API_KEY"]))
```
and in the wf output I get 0 length in both cases:

```
Querying bugzilla
0
0
```

I'd be grateful for any suggestions

scripts/customer_scenarios.py
```diff
@@ -0,0 +1,106 @@
...
+                    bug["customer_cases_count"] = len(customer_cases)
+                if bug["customer_cases_count"] > 0:
+                    output.append(f'{path} {test}')
+    return set(output)
+
+
+@main.command()
+@click.option('--paths', nargs=1)
+def run(paths):
+    settings.configure()
```
mshriver, line 89: There is no need to call `settings.configure`, this is an old pattern from the INIReader implementation. The settings object that you import is ready to read against.

.github/workflows/pull_request.yml
```diff
@@ -71,3 +71,14 @@ jobs:
...
+      - name: Get changed files
+        run: |
+          URL="https://api.github.com/repos/SatelliteQE/robottelo/pulls/${{ github.event.pull_request.number }}/files"
+          echo "PATHS=$(curl -s -X GET -G $URL | jq -r '.[] | .filename' | tr '\r\n' ',')" >> $GITHUB_ENV
+
+      - name: Customer scenario check
+        run: |
+          make customer-scenario-check paths=${{env.PATHS}}
+        env:
+          ROBOTTELO_BUGZILLA__API_KEY: '${{ secrets.BUGZILLA_KEY }}'
```
mshriver, line 84: Taking a close look at the secrets use to try to help debug the behavior you're seeing. Quotes here are unnecessary.

mshriver:
Target branch changed, merging. This should allow the action to read the secret necessary to run the script.

Issue #6868, opened 2019-04-21
ui_airgun.bookmark test cases prone to race conditions


rplevka:
We should generally avoid the assertions based on blindly looking at 1st element in the results table, as this approach doesn't take into account the fact that other tests running in parallel may create their own instances and thus their instances may appear in the results.

e.g. `test_positive_update_bookmark_public` collides with `test_positive_create_bookmark_public`

but such cases are all over the place.

mshriver:
A larger discussion about the expected behavior during `balance` or `run-on-one` xdist server modes are in use: [server.yaml](https://github.com/SatelliteQE/robottelo/blob/064176eba3e059faf690953a9986fb692e3e3e8f/conf/server.yaml.template#L18)

@JacobCallahan I think some new behavior is in order to enforce the `run_in_one_thread` mark that currently exists in coordination with the `server.yaml` setting, uncollecting test cases when necessary.

I think that the `run_in_one_thread` mark could be updated to not be thread specific, but to indicate a test whose target satellite should be isolated from other xdist workers.  This condition could be met in any of the `xdist_behavior` modes, depending on the number of workers, available satellites in `inventory.yaml` or `settings.server.hostnames`, and xdist balancing.


Separate from that specific consideration, this issue is also valid for the case where a test assumes there won't be any other entities on the system of a given type. Tests should be using unique (random) entity names/descriptions, and anywhere that its listing entities to validate presence of a specific entity, it should not assume the desired entity is the first in the list.

PR #9091, opened 2021-10-21
Update bookmark tests to be more parallel safe

sambible:
Based on #6868, changing the tests to search through the list for matches rather than blindly looking at only the first element when checking for a bookmark to exist.


omkarkhatavkar:

PR #9103, opened 2021-10-28
Add RH Cloud manual tests based on BZ scenarios 

jameerpathan111:


omkarkhatavkar (approved):

tests/foreman/cli/test_rhcloud_inventory.py
```diff
@@ -139,3 +132,79 @@ def test_positive_sync_inventory_status():
 
     :CaseLevel: System
     """
+
+
+@pytest.mark.stubbed
+def test_rhcloud_inventory_bz_1962694():
```
mshriver, line 138: I don't think naming the test case for the BZ is a good practice. The BZ context is available in the metadata, and provides zero context when viewing test results concerning what the test function is for. Can't tell a simple field validation test from a complex integration test.

Please name the test cases according to their relevant coverage. 

Is there no way to automate any of these three test cases, as indicated by `ManualOnly` metadata?

tests/foreman/cli/test_rhcloud_inventory.py
```diff
@@ -139,3 +132,79 @@ def test_positive_sync_inventory_status():
...
+            ForemanInventoryUpload::Async::GenerateAllReportsJob.perform_now
+
+    :expectedresults: If organization had more hosts than specified by max_org_size variable
+        then report won't be uploaded.
+
+    :CaseImportance: Low
+
+    :BZ: 1962694
+
+    :CaseAutomation: ManualOnly
```
mshriver, line 158: Is there no way to automate any of these test cases, as indicated by `ManualOnly`?

mshriver (changes requested):
Please update test names

tests/foreman/cli/test_rhcloud_inventory.py
jameerpathan111: They are hard to automate, some are destructive in nature and not worth automating.

jameerpathan111:
> Please update test names

@mshriver I have updated the test names.

tests/foreman/cli/test_rhcloud_inventory.py
jameerpathan111: @mshriver updated the pr.

tests/foreman/cli/test_rhcloud_inventory.py
mshriver: Ack, double checking this is what you wanted, instead of `NotAutomated`

mshriver (approved):

PR #9104, opened 2021-10-29
Add tests for disabled syncplan logic

swadeley:
Hello

adding a test to ensure a cancelled recurring logic does not prevent an upgrade or disable the previously associated sync plan.

 Closed loop: BZ#1887511

### DIFF
--- a/tests/upgrades/test_syncplan.py
+++ b/tests/upgrades/test_syncplan.py
@@ -60,7 +60,7 @@ class TestSyncPlan:
     def test_pre_disabled_sync_plan_logic(self, request):
         """Pre-upgrade scenario that creates a sync plan with both disabled and enabled recurring logic.
 
-        :id:
+        :id: c75bd43d-d868-461a-9fc3-a1fc7dccc77a
 
         :steps:
             1. Create Product
@@ -144,7 +144,7 @@ class TestSyncPlan:
         """Upgrade proceedes without RecurringLogicCancelledExceptionerror.
         After upgrade, Sync interval should still be enabled.
 
-        :id:
+        :id: b7139f47-805b-47c9-84aa-bd8c6fb00b40
 
         :steps:
             1. Verify sync plan exists and works.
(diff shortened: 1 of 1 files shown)


tests/upgrades/test_syncplan.py
```diff
@@ -56,6 +56,49 @@ def test_pre_sync_plan_migration(self, request):
...
+            organization=org, name=f'{request.node.name}_syncplan', interval="weekly"
+        ).create()
+        product = entities.Product(organization=org, name=f'{request.node.name}_prod').create()
+        entities.Repository(product=product, name=f'{request.node.name}_repos').create()
+        sync_plan.add_products(data={'product_ids': [product.id]})
+        product.sync()
+        product = product.read()
+        assert product.sync_plan.id == sync_plan.id
+        # Note the recurring logic ID for later assert a new one was created
+        count = sync_plan.foreman_tasks_recurring_logic.id
```
pondrejk, line 90: not sure why we assign `count = 0` on line 78, did you plan to do some assert here?

tests/upgrades/test_syncplan.py
```diff
@@ -56,6 +56,49 @@ def test_pre_sync_plan_migration(self, request):
...
+        # Cancel the recurring logic
+        entities.RecurringLogic(id=count).read()
+        entities.RecurringLogic(id=count).cancel()
+        # Re-enable the sync plan (it will get a new recurring logic)
+        sync_plan.enabled = True
+        sync_plan.update(['enabled'])
+        sync_plan = sync_plan.read()
+        assert sync_plan.enabled
+        # Assert a new recurring logic was assigned
+        assert sync_plan.foreman_tasks_recurring_logic.id > count
```
pondrejk, line 100: why rely on the fact that IDs represent count, doing just new_id != old_id would be more change-proof

pondrejk (approved):

pondrejk:
minor comments

tests/upgrades/test_syncplan.py
mshriver: I'd agree here, calling it `count` is misleading.

swadeley: I guess I chose "count" as the sync plan ID is incremented, will change it as suggested

swadeley: As this was already merged, I created new PR https://github.com/SatelliteQE/robottelo/pull/9125

PR #9108, opened 2021-11-01
Gha more changes

pondrejk:


pondrejk:
@mshriver @JacobCallahan sadly, the problem still persists

scripts/customer_scenarios.py
```diff
@@ -40,67 +41,65 @@ def get_bz_data(paths):
...
+        'id',
+        'summary',
+        'status',
+        'resolution',
+        'cf_last_closed',
+        'flags',
+        'external_bugs',
     ]
-    bz_url = "https://bugzilla.redhat.com"
+    bz_url = 'https://bugzilla.redhat.com'
```
Gauravtalreja1, line 52:
```suggestion
    bz_url = settings.bugzilla.url
```

JacobCallahan:
@pondrejk can you add a check in your script to validate whether or not the bz key is being populated in dynaconf? Don't print it out, just check to see if it isn't None and has a len > 0

ColeHiggins2:
Did we get this to work after adding the checks?

pondrejk:
Hello, I'm working on a rewrite, the current implementation won't work as github doesn't allow to access secrets from forks

JacobCallahan:
> Hello, I'm working on a rewrite, the current implementation won't work as github doesn't allow to access secrets from forks 

Lets discuss this more during the JFT call

pondrejk:
Some update on progress, I changed the workflow in various ways:
- it now checks bzs for all tests, not just for files changed in the latest PR
- as such it is not tied to merge or pr action but it is scheduled to run weekly
- if it finds missing tags, it opens an issue with a list of tests to address (assigned to me for the time being)

This all has been tested in my fork. The reason I'm keeping this in the draft state is that I'm getting some false positives when querying bugzilla for external bugs, I'm investigating the problem, and will share updates here

pondrejk:
The issue mentioned above is now fixed, pr is ready for review. @JacobCallahan @mshriver @Gauravtalreja1 @ColeHiggins2

JacobCallahan (approved):

scripts/customer_scenarios.py
```diff
@@ -0,0 +1,103 @@
...
+    bz_fields = [
+        'id',
+        'summary',
+        'status',
+        'resolution',
+        'cf_last_closed',
+        'external_bugs',
+    ]
+    bz_url = settings.bugzilla.url
+    bz_key = settings.bugzilla.api_key
```
jyejare, line 56: Need not to initialize the vars as you have used it only once, you can direct use it :)

Non Blocking comment!

jyejare (approved):

latran (approved):
ack

PR #9110, opened 2021-11-02
DRAFT Dynamically replace default_sat from marks

mshriver:
Adjust align_to_xdist fixture, splitting a session scoped default worker
alignment from the function scoped satellite hostname selection based on
test context.

First relevant mark is destructive.

Yet untested, pushing the branch early though.


jameerpathan111:
@mshriver can we re-investigate this, this will be helpful for some of my tests.

PR #9113, opened 2021-11-03
test are using rhel host instead of rhel7

tstrych:
This PR introduced the usage of all `rhel_contenthost` versions right now (rhel 6, rhel7 and rhel8), specified in conf/content_host.yaml
The product, repo and repository-set is set based on test name which have the rhel version in brackets.
example: `tests/foreman/cli/test_host.py::test_positive_erratum_applicability[rhel6]`

- New change introduced that all tests with fixtures that were at module level are now at function level. The consequence: not only 3 hosts are created for each version, but also the repo setup has to be done fore each of them. 

- `repo_setup` and `rhel_version` fixture should be moved elsewhere to some general place, what do you think the right place is ?  

- I was thinking that rhel_contenthost should run only on the default one if not specified otherwise. So it would not take forever.

- if pit marker specified all rhel versions will run, the problem with this approach is, if somebody wants to run multiple versions and his test is not valid for pit testing, is another marker/marker's the solution ? that would mean we have to parse the markers that are running and change the settings value during the run, this was the best idea I was able to come with. So I would appreciate some ideas from the team, especially @JacobCallahan as he is the creator of the rhel_contenthost. Maybe smth like this ?
```
@pytest.fixture
def select_rhel_version(request, pytestconfig):
    markers_arg = pytestconfig.getoption('-m')
    request.cls.marker_name = markers_arg
    if 'pit' not in markers_arg:
        settings.content_host.rhel_versions = [settings.content_host.default_version]
 ```

- no test results, impossible to test locally, just take so much time, manifest upload takes 30 minutes per each test with contenthost per each version used 

- will test it here with PRT, so I can do other work, instead of waiting, please wait for test_results

### DIFF
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -16,6 +16,7 @@
 
 :Upstream: No
 """
+import os
 from datetime import datetime
 from datetime import timedelta
 from random import choice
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/foreman/cli/test_host.py [hunk 1]
```diff
@@ -66,6 +67,38 @@
 from robottelo.hosts import ContentHostError
 
 
+@pytest.fixture(scope="function")
+def rhel_version():
+    current_test_name = os.environ.get('PYTEST_CURRENT_TEST')
+    start = current_test_name.find('[') + 1
+    end = current_test_name.find(']')
```
tstrych, line 74: start and end used just because of black formatting and flake.
black wants:
`[current_test_name.find('[') + 1 : current_test_name.find(']')]`
flake didn't like the space after `1`

tstrych:
trigger: test-robottelo
pytest: -m 'host_subscription or katello_host_tools or host_create' tests/foreman/cli/test_host.py

JacobCallahan:
@tstrych instead of doing repo setup in a fixture, reading names to determine version, this could be done with helper functions. This is something I've been talking about for some time now and will be part of the helpers refactor. We can talk about more tomorrow.

tstrych:
after talk with jcallahan marking this as draft, helper refactor topic will be introduced and will provide the easier way how to implement this, till then, this can sit

JacobCallahan:
Removed 6.10 label since the helpers refactor will be a post-6.10 effort

tstrych:
This effort was implemented by rplevka later on.

PR #9117, opened 2021-11-04
sftp is slow, using scp for now

tstrych:
tested by `test_xxx` 
with average around `6` seconds

```
def test_xxx():
    from robottelo import manifests
    from robottelo import ssh
    from time import time
    sum = 0
    loops = 10
    for x in range(loops):
        st = time()
        client = ssh.get_client()
        ct = time()
        with manifests.clone() as manifest:
            client.put(manifest, manifest.filename)
        ft = time()
        print(f"get_client takes: {ct-st} seconds\nmanifest_clone takes: {ft-ct} seconds")
        sum += (ft-ct)
    print(f'average is {sum / loops}')
```

what, just looked at the copied files  and they have different sizes :O ?
```
 2.3M Nov  4 11:16 manifest-1636039012.zip
 2.5M Nov  4 11:16 manifest-1636039008.zip
 2.6M Nov  4 11:16 manifest-1636039004.zip
 2.9M Nov  4 11:16 manifest-1636038999.zip
 2.5M Nov  4 11:16 manifest-1636038995.zip
 2.6M Nov  4 11:16 manifest-1636038991.zip
 2.5M Nov  4 11:16 manifest-1636038987.zip
 2.5M Nov  4 11:16 manifest-1636038983.zip
 3.6M Nov  4 11:16 manifest-1636038965.zip
```

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -437,9 +437,9 @@ class ContentHost(Host):
             with NamedTemporaryFile(dir=robottelo_tmp_dir) as content_file:
                 content_file.write(local_path.content.read())
                 content_file.seek(0)
-                self.session.sftp_write(source=content_file.name, destination=remote_path)
+                self.session.scp_write(source=content_file.name, destination=remote_path)
         else:
-            self.session.sftp_write(source=local_path, destination=remote_path)
+            self.session.scp_write(source=local_path, destination=remote_path)
 
     def put_ssh_key(self, source_key_path, destination_key_name):
         """Copy ssh key to virtual machine ssh path and ensure proper permission is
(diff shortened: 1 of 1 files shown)


tstrych:
tried to unzip those files, and can't, seems that there are not transferred correctly

ogajduse:
As a robottelo reviewer, I would be concerned about the security of `scp`.
See https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/4CYOOLAWIIQTYAZXF2XXVALBS5OX4WM6/

JacobCallahan:
@ogajduse ACK and this was one of the reasons why sftp was used initially. However, the performance hit right now greatly outweighs what little risk we might incur. Some local testing of the difference between the two below

```python
In [16]: %timeit box.session.scp_write("manifest_robottelo-automation_20211103T174310Z.zip", "/root/manifest.zip")
3.96 s ± 191 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [17]: %timeit box.session.sftp_write("manifest_robottelo-automation_20211103T174310Z.zip", "/root/manifest.zip")
13min 24s ± 23.2 s per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

JacobCallahan (approved):
ACK and great investigation! This would also be an easy win for a 6.9.z cherry-pick once merged

JacobCallahan:
Hold on this for now, more discussions happening around this.

PR #9119, opened 2021-11-04
Remove positive_create_with_gpg_key_by_name 

sambible:
Create no longer supports gpg_key_name, so this test always fails. Since it only tests functionality that has been removed, removing the test seems like a reasonable option - maybe there is a better way? Leaving it draft, looking for feedback.


ColeHiggins2 (approved):
ACK

sambible:
Going to comment out this for now, pending https://bugzilla.redhat.com/show_bug.cgi?id=2008656 being resolved.

PR #9121, opened 2021-11-05
Timeout is accepted in miliseconds by the ssh2 module, see line 3074 of session.c

lhellebr:


lhellebr:
Also, line 167 of session.pyx

JacobCallahan (approved):
This could also be expanded on that "smart timeout" values are also accepted

pondrejk (approved):

PR #9131, opened 2021-11-09
fix in MD5 repo test

pondrejk:
Checking the published path is no longer viable with pulp3, hence different assertions

```
pytest tests/foreman/cli/test_repository.py -k test_positive_sync_publish_promote_cv
==================================================== test session starts =====================================================
                                                                     

tests/foreman/cli/test_repository.py s.s                                                                               [100%]


============================ 1 passed, 2 skipped, 231 deselected, 6 warnings in 95.49s (0:01:35) =============================

### DIFF
--- a/tests/foreman/cli/test_repository.py
+++ b/tests/foreman/cli/test_repository.py
@@ -2577,19 +2577,18 @@ class TestMD5Repository:
         """
         lce = make_lifecycle_environment({'organization-id': module_org.id})
         Repository.synchronize({'id': repo['id']})
+        synced_repo = Repository.info({'id': repo['id']})
+        assert synced_repo['sync']['status'].lower() == 'success'
+        assert synced_repo['content-counts']['packages'] == '35'
         cv = make_content_view({'organization-id': module_org.id})
         ContentView.add_repository({'id': cv['id'], 'repository-id': repo['id']})
         ContentView.publish({'id': cv['id']})
         content_view = ContentView.info({'id': cv['id']})
         cvv = content_view['versions'][0]
         ContentView.version_promote({'id': cvv['id'], 'to-lifecycle-environment-id': lce['id']})
-        result = default_sat.execute(
-            f"ls /var/lib/pulp/published/yum/https/repos/{module_org.label}/{lce['label']}/"
-            f"{cv['label']}/custom/{module_product.label}/{repo['label']}/Packages/b"
-            " | grep bear-4.1-1.noarch.rpm"
-        )
-        assert result.status == 0
-        assert result.stdout
+        cv = ContentView.info({'id': cv['id']})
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: tests/foreman/cli/test_repository.py [hunk 1]
```diff
@@ -2577,19 +2577,17 @@ def test_positive_sync_publish_promote_cv(self, repo, module_org, module_product
         """
         lce = make_lifecycle_environment({'organization-id': module_org.id})
         Repository.synchronize({'id': repo['id']})
+        synced_repo = Repository.info({'id': repo['id']})
+        assert synced_repo['sync']['status'] == 'Success'
```
jyejare, line 2581:
```suggestion
        assert synced_repo['sync']['status'].lower() == 'success'
```
This would save you from future silly failure in the `success` word.

tests/foreman/cli/test_repository.py
```diff
@@ -2577,19 +2577,17 @@ def test_positive_sync_publish_promote_cv(self, repo, module_org, module_product
...
         ContentView.version_promote({'id': cvv['id'], 'to-lifecycle-environment-id': lce['id']})
-        result = default_sat.execute(
-            f"ls /var/lib/pulp/published/yum/https/repos/{module_org.label}/{lce['label']}/"
-            f"{cv['label']}/custom/{module_product.label}/{repo['label']}/Packages/b"
-            " | grep bear-4.1-1.noarch.rpm"
-        )
-        assert result.status == 0
-        assert result.stdout
+        cv = ContentView.info({'id': cv['id']})
+        assert synced_repo['id'] in [repo['id'] for repos in cv['yum-repositories']]
```
jyejare, line 2590: This does not guarantees that the rpms of `synced_repo` are promoted to `lce['id']`. This is only checking if the yum repository is added to the CV which is not the best assertion after the above line of `version_promote`. 

See if you could find out the synced_repo is promoted to lce['id'] via CLI / API (Nailgun), else keep the existing assertion (modify the assertion if the fs changed here, in case).

jyejare (changes requested):

latran (approved):
ACK

tests/foreman/cli/test_repository.py
jyejare: @pondrejk This is a blocking comment, which I would you to consider before merging the PR. The implementation is a completely changing assertion that has no gain.

pondrejk: Hello @jyejare did you see the added assertion on line 2591?  I check lces via the cv api, the other way around seems not viable right now

jyejare: Now I see it !

jyejare (approved):

PR #9140, opened 2021-11-10
changed registered host to rhel7 only

ColeHiggins2:
Fixed tests failure, Removed parametrized testing for rhel6, rhel7, and rhel8. 

Open for discussion on whether we need to write 2 more tests for rhel6 and rhel8

### DIFF
--- a/tests/foreman/api/test_subscription.py
+++ b/tests/foreman/api/test_subscription.py
@@ -260,7 +260,9 @@ def test_positive_subscription_status_disabled(
 @pytest.mark.tier2
 @pytest.mark.pit_client
 @pytest.mark.pit_server
-def test_sca_end_to_end(module_ak, rhel_contenthost, module_org, rh_repo, custom_repo, default_sat):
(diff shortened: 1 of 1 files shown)


JacobCallahan (approved):
ACK. Merge pending test results. You can hold on the other rhel versions for now, but keep this test in mind when we get around to the major parameterization push.

synkd (approved):
ACK.

I do think it would be a good idea to write tests for RHEL 6 and 8. RHEL 6 currently requires an Extended Lifecyle Support (ELS) subscription to access content; we could potentially catch any issues arising from SCA and ELS subscriptions with that test. And RHEL 8 uses (and, I assume, RHEL 9 will use) a different product certificate from earlier RHEL versions; we may be able to catch version-specific SCA issues there as well.

JacobCallahan:
@ColeHiggins2 can you rebase this pr, should help you checks pass.

swadeley (approved):
ACK

PR #9154, opened 2021-11-12
remove unnecessary slash

sambible:
a long investigation into a simple problem.


sambible:
trigger: test-robottelo
pytest: tests/foreman/cli/test_repository.py::test_positive_upload_file_to_file_repo

JacobCallahan:
Can you elaborate on the problem and fix?
Also, is the PRT failure related?

latran (changes requested):
There's a few more test cases that uses that file location.  Run those test cases as well to ensure that simple change doesn't negatively impact the others.

PR #9169, opened 2021-11-17
Remove use of pre-made repofiles in favor of repofile construction

JacobCallahan:
This change gets rid of repofiles previously hosted on "rhel_repo_host"
in favor of direct references to repo baseurls already established on
"rhel_os_repo_host"

We will now construct repofiles on the host instead of copying them from one location to another. This is more in line with what we were already doing in a number of other instances.

### DIFF
--- a/robottelo/config/validators.py
+++ b/robottelo/config/validators.py
@@ -234,13 +234,6 @@ VALIDATORS = dict(
     ],
     repos=[
         Validator(
-            'repos.rhel6_repo',
-            'repos.rhel7_repo',
-            'repos.rhel8_repo',
-            must_exist=True,
-            is_type_of=str,
-        ),
-        Validator(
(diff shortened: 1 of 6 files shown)


JacobCallahan:
Lots of testing to do before this can be merged, likely will uncover more areas that need to be modified.

JacobCallahan:
Moving out of draft since initial tests look good. We can handle the rest in production

swadeley:
@JacobCallahan Please check merge conflicts.

Please remember to justify a change in the intro comment when its not a simple case of fixing something that is broken (features must have benefits).

JacobCallahan:
@swadeley thanks :) updated

pondrejk (approved):

swadeley (approved):
ACK

PR #9170, opened 2021-11-18
Remove newline join.

lhellebr:
Result of execute is just a string, newline join on that string results
in single characters on a line. Remove that.


ogajduse (approved):

lhellebr:
```
$ pytest tests/foreman/cli/test_reporttemplates.py::test_positive_report_help | grep passed
======================== 1 passed, 3 warnings in 18.83s ========================
```

JacobCallahan (approved):

PR #9171, opened 2021-11-18
Take whole string and strip whitespace.

lhellebr:
A single string is returned. Do not take only its first character. Split
newline off of id.


pondrejk (approved):

swadeley (approved):
ACK

PR #9185, opened 2021-11-24
Fix receptor test.

lhellebr:
Results now contain one string instead of list of lines.
Also, it's necessary to output debugging information such as repolist
and invocation output to determine why this fails in the automation.

### DIFF
--- a/tests/foreman/cli/test_remoteexecution.py
+++ b/tests/foreman/cli/test_remoteexecution.py
@@ -40,6 +40,7 @@ from robottelo.constants import PRDS
 from robottelo.constants import REPOS
 from robottelo.constants import REPOSET
 from robottelo.hosts import ContentHost
+from robottelo.logging import logger
 from robottelo.utils.issue_handlers import is_open
 
 
@@ -531,18 +532,24 @@ class TestRemoteExecution:
             in ["succeeded", "failed"],
             timeout="1500s",
         )
+
+        result = JobInvocation.get_output({'id': invocation_id, 'host': default_sat.hostname})
+        logger.debug(f"Invocation output>>\n{result}\n<<End of invocation output")
+        # if installation fails, it's often due to missing rhscl repo -> print enabled repos
+        repolist = default_sat.execute('yum repolist')
+        logger.debug(f"Repolist>>\n{repolist}\n<<End of repolist")
+
         assert entities.JobInvocation(id=invocation_id).read().status == 0
-        result = ' '.join(
-            JobInvocation.get_output({'id': invocation_id, 'host': default_sat.hostname})
-        )
         assert 'project-receptor.satellite_receptor_installer' in result
         assert 'Exit status: 0' in result
         # check that there is one receptor conf file and it's only readable
         # by the receptor user and root
         result = default_sat.execute('stat /etc/receptor/*/receptor.conf --format "%a:%U"')
-        assert result.stdout == '400:foreman-proxy'
+        assert all(
+            filestats == '400:foreman-proxy' for filestats in result.stdout.strip().split('\n')
+        )
         result = default_sat.execute('ls -l /etc/receptor/*/receptor.conf | wc -l')
-        assert result.stdout == '1'
+        assert int(result.stdout.strip()) >= 1
 
 
 class TestAnsibleREX:


lhellebr:
The test passes locally.

omkarkhatavkar:
trigger: test-robottelo
pytest: -k test_positive_run_receptor_installer tests/foreman/cli/test_remoteexecution.py

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -531,18 +532,22 @@ def test_positive_run_receptor_installer(self, default_sat):
...
-        result = ' '.join(
-            JobInvocation.get_output({'id': invocation_id, 'host': default_sat.hostname})
-        )
         assert 'project-receptor.satellite_receptor_installer' in result
         assert 'Exit status: 0' in result
         # check that there is one receptor conf file and it's only readable
         # by the receptor user and root
         result = default_sat.execute('stat /etc/receptor/*/receptor.conf --format "%a:%U"')
-        assert result.stdout == '400:foreman-proxy'
+        assert result.stdout.strip() == '400:foreman-proxy'
```
devendra104, line 548: Can we do this like `"400:foreman-proxy" in result.stdout`?

lhellebr: That's what I initially wanted to do but that would also match a lot of unexpected stuff like `69400:foreman-proxy-nonexistent-user LASAGNA WITH SPAM` so I opted to do it this way.

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -531,18 +532,22 @@ def test_positive_run_receptor_installer(self, default_sat):
             in ["succeeded", "failed"],
             timeout="1500s",
         )
+
+        result = JobInvocation.get_output({'id': invocation_id, 'host': default_sat.hostname})
+        logger.debug(f"Invocation output>>\n{result}\n<<End of invocation output")
+        # if installation fails, it's often due to missing rhscl repo -> print enabled repos
+        repolist = default_sat.execute('yum repolist')
+        logger.debug(f"Repolist>>\n{repolist}\n<<End of repolist")
```
jyejare, line 540: Is this kept for debugging purposes or intentional?

If it's intentional, then I think this is not needed and its also adds to test execution time!

tests/foreman/cli/test_remoteexecution.py
jyejare: @lhellebr if there are more strings in `result.stdout` then this will never pass with `==` operation. Maybe a Regex will help here!

jyejare (changes requested):

tests/foreman/cli/test_remoteexecution.py
lhellebr: From the OP:
```
Also, it's necessary to output debugging information such as repolist
and invocation output to determine why this fails in the automation.
```

jyejare: @lhellebr If you are willing to know that for CI then that's fine but you can always get that output if u run it locally.

lhellebr: That way, I can get a different output since the automation environment can be different. My use case: I need to verify that in automation, there is a certain repository missing.

Anyway, it's good to see the result of a failure in the log when investigating automation failures - that's what the logs are for. The job log isn't available currently.

jyejare: Sure! I understand the concern!

jyejare (approved):

jyejare:
@lhellebr Does the 1 test fail in PRT is expected?

lhellebr:
trigger: test-robottelo
pytest: -k test_positive_run_receptor_installer tests/foreman/cli/test_remoteexecution.py

lhellebr:
It still passes locally. Trying to rerun.

lhellebr:
trigger: test-robottelo
pytest: -k test_positive_run_receptor_installer tests/foreman/cli/test_remoteexecution.py

lhellebr:
trigger: test-robottelo
pytest: -k test_positive_run_receptor_installer tests/foreman/cli/test_remoteexecution.py

lhellebr:
trigger: test-robottelo
pytest: -k test_positive_run_receptor_installer tests/foreman/cli/test_remoteexecution.py

lhellebr:
Still passes locally.
In multiple PRT testing runs, there were two possible outcomes:

- `KeyError: 'template_invocations'` which is not expected but I also highly doubt it's related, seems like some race condition
- `AssertionError: assert 1 == 0` which is expected, I'm trying to solve it by getting the debug

lhellebr:
However, I see I'm not getting the needed debug in PRT so I think I won't get it in automation either. Maybe I need higher log level than `debug`? I'll try.

lhellebr:
trigger: test-robottelo
pytest: -k test_positive_run_receptor_installer tests/foreman/cli/test_remoteexecution.py

devendra104:
@lhellebr The PRT is getting failed because of the no `rhel-7-server-satellite-6.10-rpms` repository enabled for that test case same was happening with the high importance and upgrade all tier pipeline job.

```
tests/foreman/cli/test_remoteexecution.py:534: in test_positive_run_receptor_installer
    assert entities.JobInvocation(id=invocation_id).read().status == 0
E   assert 1 == 0
E     +1
E     -0
```

Could you enable the required repository which has `satellite_receptor_installer` plugin to avoid failure at this point?

Rest of the changes are looking good to me.

jyejare:
@devendra104 ++, thanks for deep diving in the test failure !

lhellebr:
I'm well aware of that failure, it's the one I mentioned in OP. The same happens in automation. Fix for this is a subject to discussion and will be part of a different PR.

mshriver (approved):

PR #9186, opened 2021-11-24
remove pit_client from 1 non-compliant test

rplevka:
The test  does not use the proper VM fixtures so it ignores the PIT deploy workflow. Thus it provides VM running incorrect rhel version.


ogajduse (approved):
ACK
We need to board the scenarios ASAP. Rework of the REX tests is a bigger effort that we can not fit into the scenario onboarding timeframe.

vsedmik (approved):
ACK

PR #9187, opened 2021-11-25
6.9 backport changes in rex services test

pondrejk:
upon request from pit team

### DIFF
--- a/tests/foreman/cli/test_remoteexecution.py
+++ b/tests/foreman/cli/test_remoteexecution.py
@@ -35,6 +35,7 @@ from robottelo.cli.recurring_logic import RecurringLogic
 from robottelo.cli.task import Task
 from robottelo.config import settings
 from robottelo.hosts import ContentHost
+from robottelo.utils.issue_handlers import is_open
 
 
 @pytest.fixture()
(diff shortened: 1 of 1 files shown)


pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k test_positive_run_packages_and_services_job

devendra104 (approved):
ACK

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -629,14 +630,18 @@ def test_positive_run_concurrent_jobs(self, fixture_vmsetup, module_org):
         GlobalParameter().delete({'name': param_name})
         assert len(GlobalParameter().list({'search': param_name})) == 0
 
+    nick_params = [{'nick': 'rhel7'}, {'nick': 'rhel7_fips'}, {'nick': 'rhel8'}]
+    if not is_open('BZ:1811166'):
+        nick_params.append({'nick': 'rhel8_fips'})
```
ogajduse, line 635: I somehow do not like this section. I know that this is just a backport, but I would like to ask @peterdragun to have look at this as he played around with fixture and test parametrization lately.

peterdragun: I don't think we have any better option here unfortunately. Other option would be skip test with this parameter directly in test function but this would be even worse IMO.

peterdragun (approved):

ogajduse (approved):

tests/foreman/cli/test_remoteexecution.py
ogajduse: Thanks!

PR #9188, opened 2021-11-25
fix export_import RH content tests

rmynar:
This change should fix _test_positive_export_import_redhat_cv_with_huge_contents_.
It was necessary to use separate manifest for each tests that enables and syncs RH repo, so _function_org_ fixture is used instead of _module_org_. Otherwise we get following error:
_Import is the same as existing data
Owner has already imported from another subscription management application. The following conflicts were found: [ MANIFEST_SAME, DISTRIBUTOR_CONFLICT ]_
Ability to extend timeout was added to export/import commands that may take long to run.
Tests _test_positive_export_import_redhat_cv_ and _test_positive_export_import_redhat_cv_with_huge_contents_ are nearly the same, but the second one uses huge repo, needs to have timeout extended and takes (very) long to run.

### DIFF
--- a/robottelo/cli/content_export.py
+++ b/robottelo/cli/content_export.py
@@ -39,36 +39,36 @@ class ContentExport(Base):
         return cls.execute(cls._construct_command(options), output_format='json')
 
     @classmethod
-    def completeLibrary(cls, options):
+    def completeLibrary(cls, options, timeout=None):
         """
         Make full library export
         """
         cls.command_sub = 'complete library'
-        return cls.execute(cls._construct_command(options), output_format='json')
+        return cls.execute(cls._construct_command(options), output_format='json', timeout=timeout)
 
     @classmethod
-    def completeVersion(cls, options):
+    def completeVersion(cls, options, timeout=None):
         """
         Make full CV version export
         """
         cls.command_sub = 'complete version'
-        return cls.execute(cls._construct_command(options), output_format='json')
+        return cls.execute(cls._construct_command(options), output_format='json', timeout=timeout)
 
     @classmethod
-    def incrementalLibrary(cls, options):
+    def incrementalLibrary(cls, options, timeout=None):
         """
         Make make incremental library export
         """
         cls.command_sub = 'incremental library'
-        return cls.execute(cls._construct_command(options), output_format='json')
+        return cls.execute(cls._construct_command(options), output_format='json', timeout=timeout)
 
     @classmethod
-    def incrementalVersion(cls, options):
+    def incrementalVersion(cls, options, timeout=None):
         """
         Make make incremental CV version export
         """
         cls.command_sub = 'incremental version'
-        return cls.execute(cls._construct_command(options), output_format='json')
+        return cls.execute(cls._construct_command(options), output_format='json', timeout=timeout)
 
     @classmethod
     def generateMetadata(cls, options):
--- a/tests/foreman/cli/test_satellitesync.py
+++ b/tests/foreman/cli/test_satellitesync.py
@@ -1099,8 +1099,7 @@ class TestContentViewSync:
         cvv = cv['versions'][0]
         # Export cv
         path = ContentExport.completeVersion(
-            {'id': cvv['id'], 'organization-id': function_org.id},
-            timeout=7200000
+            {'id': cvv['id'], 'organization-id': function_org.id}, timeout=7200000
         )
         # grab export dir and check all exported files are there
         export_dir = os.path.dirname(path['message'])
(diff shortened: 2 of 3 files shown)


robottelo/cli/content_export.py
```diff
@@ -39,36 +39,36 @@ def list(cls, options=None):
         return cls.execute(cls._construct_command(options), output_format='json')
 
     @classmethod
-    def completeLibrary(cls, options):
+    def completeLibrary(cls, options, timeout=None):
```
lhellebr, line 42: I'd prefer setting some long timeout to setting no timeout at all. I'd also think about taking this value from config by default - there are some timeout related settings. Applies to all instances of this.

tests/foreman/cli/test_satellitesync.py
```diff
@@ -905,7 +905,7 @@ def test_positive_export_import_promoted_cv(
     @pytest.mark.tier3
     @pytest.mark.upgrade
     def test_positive_export_import_redhat_cv(
-        self, export_cleanup, config_export_import_settings, module_org, default_sat
+        self, export_cleanup, config_export_import_settings, function_org, default_sat
```
lhellebr, line 908: Can't the org be reused in the whole module?

tests/foreman/cli/test_satellitesync.py
```diff
@@ -1061,28 +1060,54 @@ def test_positive_export_import_redhat_cv_with_huge_contents(
...
-        rhel_repo_name = 'rhscl7'
-        product = 'rhscl'
-        releasever = '7Server'
-        rhel_repo = _enable_rhel_content(
-            default_sat, class_export_entities['exporting_org'], rhel_repo_name, releasever, product
+        cv_name = gen_string('alpha')
+
+        with manifests.clone() as manifest:
+            default_sat.put(manifest, manifest.filename)
+        Subscription.upload({'file': manifest.filename, 'organization-id': function_org.id})
```
lhellebr, line 1067: Can this be a fixture instead?

tests/foreman/cli/test_satellitesync.py
```diff
@@ -1061,28 +1060,54 @@ def test_positive_export_import_redhat_cv_with_huge_contents(
...
+        RepositorySet.enable(
```
lhellebr, line 1068: Can these also be fixtures, especially since they only sync RH repos?

tests/foreman/cli/test_satellitesync.py
```diff
@@ -1061,28 +1060,54 @@ def test_positive_export_import_redhat_cv_with_huge_contents(
...
+            {
+                'basearch': 'x86_64',
+                'name': REPOSET['rhscl7'],
```
lhellebr, line 1071: Is there some specific reason to use rhscl or is it just a random RH repo?

robottelo/cli/content_export.py
ogajduse: I would set the timeout for each and every function here. I would rather see it set in in the `execute` method.
If there is no timeout specified in the `execute` function call in `robottelo/cli/content_export.py`, it would use the default timeout value from the config.

lhellebr: Isn't that the current behavior?

ogajduse: Disregard what I wrote above. 
We already have it implemented here:
https://github.com/SatelliteQE/robottelo/blob/99b70756a965db1af0dfb770814033d71c8cb774/robottelo/hosts.py#L102-L104

So every command that is executed on the host has assigned the timeout value from the robottelo config.

If @rmynar wants to use the timeout from the robottelo config, he has to use `None`. If he wants to have no timeout for the command he should use `0`. And of course there might be timeout explicitly defined. We can specify it in time units. See the broker helper:

https://github.com/SatelliteQE/broker/blob/da7c35f24f0d44d262268e8b6988e7f250ba012e/broker/helpers.py#L372-L391

The question is whether we want to set the timeout here in the `completeLibrary` function or leave it up to the test that is using the function.

rmynar: The problematics of setting timeout was described by @ogajduse.
I would like to use timeout as non-mandatory (by default use value from settings) and increase it only in specific tests.

tests/foreman/cli/test_satellitesync.py
rmynar: Unfortunately, we need to use specific content for each test, therefore we need "clean" organization.

lhellebr (approved):

tests/foreman/cli/test_satellitesync.py
rmynar: Some reasonably large repo needs to be used. RHSCL was chosen by original author @latran. This is only a fix of failed test.

tests/foreman/cli/test_satellitesync.py
rmynar: Well, might be. But I don't think it's necessary while we use it only twice. But it would deserve some refactoring if we add more similar tests.

tests/foreman/cli/test_satellitesync.py
rmynar: Same as above ^

ogajduse (approved):
ACK from my side.
@rmynar please fix the CI failures.
I would also like to see @latran's input here.

swadeley:
> ACK from my side. @rmynar please fix the CI failures. I would also like to see @latran's input here.

@latran is AFK , but as its a test fix I think its OK to proceed.

swadeley (approved):
ACK

pondrejk (approved):

PR #9192, opened 2021-11-26
host traces test introduced

tstrych:
this is low priority but nice to have :smile: 
this was 6.8 feature if I am not mistaken 
I should probably add some asserts around service status after the downgrade, and after `hammer host traces resolve`. To see if the test really restarted the service.  -> Marking as TODO. 

I wanted to use `tracer -a` or `tracer --all` from client, to assert the results, but code snippets below, shows that channel read method will never stops from unknown reason, which is very low level issue, and I didn't find enough courage to dig deeper.

```
from broker.hosts import Host
my_host = Host("my.fqdn.com")
res = my_host.execute("tracer -a")
```
or 
```
from broker.session import Session
s = Session(hostname='<fqdn>', password='<passwd>')
channel = s.session.open_session()
channel.execute('tracer -a')
channel.read() # << the problematic command
```
If somebody wants to debug this issue, this is as far as I get, it would require to understand the channel read method, all paths lead to C-language :smile:  :smiley: 
https://github.com/ParallelSSH/ssh2-python/blob/master/ssh2/channel.pyx#L105-L145

### DIFF
--- a/pytest_fixtures/core/contenthosts.py
+++ b/pytest_fixtures/core/contenthosts.py
@@ -9,6 +9,7 @@ from broker import VMBroker
 
 from robottelo import constants
 from robottelo.config import settings
+from robottelo.constants import REPOS
 from robottelo.hosts import ContentHost
 
 
@@ -89,13 +90,38 @@ def content_hosts(request):
 
 @pytest.fixture(scope='module')
 def registered_hosts(organization_ak_setup, content_hosts, default_sat):
-    """Fixture that registers content hosts to Satellite."""
+    """Fixture that registers content hosts to Satellite, based on rh_cloud setup"""
     org, ak = organization_ak_setup
     for vm in content_hosts:
         vm.install_katello_ca(default_sat)
         vm.register_contenthost(org.label, ak.name)
         assert vm.subscribed
     return content_hosts
+
+
+@pytest.fixture(scope="function")
+def katello_host_tools_host(setup_rhst_repo, rhel7_contenthost, default_sat):
+    """Register content host to Satellite and install katello-host-tools on the host,
+    based on hosts setup"""
+    rhel7_contenthost.install_katello_ca(default_sat)
+    rhel7_contenthost.register_contenthost(
+        setup_rhst_repo['org'].label,
+        setup_rhst_repo['ak'].name,
+    )
+    assert rhel7_contenthost.subscribed
+    rhel7_contenthost.enable_repo(REPOS[setup_rhst_repo['repo_name']]['id'])
+    rhel7_contenthost.install_katello_host_tools()
+    yield rhel7_contenthost
+
+
+@pytest.fixture(scope="function")
+def katello_host_tools_tracer_host(katello_host_tools_host, default_sat):
+    """Install katello-host-tools-tracer, add REx key and create custom
+    repositories on the host"""
+    katello_host_tools_host.install_tracer()
+    katello_host_tools_host.add_rex_key(satellite=default_sat)
+    katello_host_tools_host.create_custom_rhel_repo_file_to_downgrade_packages()
+    yield katello_host_tools_host
 
 
 @pytest.fixture
--- a/robottelo/cli/host.py
+++ b/robottelo/cli/host.py
@@ -460,3 +460,58 @@ class HostInterface(Base):
         """Create new network interface for host"""
         cls.command_sub = 'create'
         cls.execute(cls._construct_command(options), output_format='csv')
+
+
+class HostTraces(Base):
+    """Manages traces on your hosts
+
+    Usage::
+        hammer host traces [OPTIONS] SUBCOMMAND [ARG] ...
+
+    Subcommands::
+        list                          List services that need restarting on the host
+        resolve                       Resolve Traces
+    """
+
+    command_base = 'host traces'
+
+    @classmethod
+    def list(cls, options=None):
+        """List services that need restarting on the host.
+
+        Usage::
+
+            hammer host traces list [OPTIONS]
+
+        Options::
+
+            --fields FIELDS               Show specified fields or predefined field sets only.
+                                        (See below) Comma separated list of values.
+                                        Values containing comma should be quoted or
+                                        escaped with backslash.
+                                       JSON is acceptable and preferred way for complex parameters
+            --host[-id]                   Name/id of the host
+        """
+        cls.command_sub = 'list'
+        return cls.execute(cls._construct_command(options), output_format='csv')
+
+    @classmethod
+    def resolve(cls, options=None):
+        """Resolve Traces
+
+        Usage::
+
+            hammer host traces resolve [OPTIONS]
+
+        Options::
+
+            --async                       Do not wait for the task
+            --host[-id]                   Name/id of the host
+            --trace-ids TRACE_IDS         Array of Trace ids
+                                       Comma separated list of values.
+                                       Values containing comma should be quoted or
+                                       escaped with backslash.
+                                       JSON is acceptable and preferred way for complex parameters
+        """
+        cls.command_sub = 'resolve'
+        cls.execute(cls._construct_command(options))
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -29,6 +29,7 @@ from fauxfactory import gen_ipaddr
 from fauxfactory import gen_mac
 from fauxfactory import gen_string
 from nailgun import entities
+from packaging import version
 
 from robottelo.api.utils import promote
 from robottelo.cli.activationkey import ActivationKey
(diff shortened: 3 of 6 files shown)


tests/foreman/cli/test_host.py
```diff
@@ -2558,3 +2559,65 @@ def test_positive_dump_enc_yaml(default_sat):
     assert f'fqdn: {default_sat.hostname}' in enc_dump
     assert f'ip: {default_sat.ip_addr}' in enc_dump
     assert 'ssh-rsa' in enc_dump
+
+
+# -------------------------- HOST TRACE SUBCOMMAND SCENARIOS -------------------------
+def generate_custom_rhel_repo_string(major, minor):
+    domain = 'http://download.eng.bos.redhat.com'
```
jyejare, line 2566: Remove this immediately as its pointing to internal / private repository and make this part of dynaConf from where u ll access this URL from confs.

Let me know if u need any help to understand this !

robottelo/cli/host.py
```diff
@@ -460,3 +460,27 @@ def create(cls, options=None):
...
+    Subcommands::
+        list                          List services that need restarting on the host
+        resolve                       Resolve Traces
+    """
+
+    command_base = 'host traces'
+
+    @classmethod
+    def list(cls, options=None):
+        cls.command_sub = 'list'
```
ogajduse, line 480: Please add documentation for each command.
```suggestion
    def list(cls, options=None):
        """List services that need restarting on the host
        Usage:
            hammer host traces list [OPTIONS]
        Options:
         --host[-id]                   Set the current host context for the request.
                                       Name/Id can be used
         -h, --help                    Print help
        """
        cls.command_sub = 'list'
```

robottelo/cli/host.py
```diff
@@ -460,3 +460,27 @@ def create(cls, options=None):
...
+        return cls.execute(cls._construct_command(options), output_format='csv')
+
+    @classmethod
+    def resolve(cls, options=None):
+        cls.command_sub = 'resolve'
```
ogajduse, line 485:
```suggestion
    def resolve(cls, options=None):
        """Resolve Traces
        Usage:
            hammer host traces resolve [OPTIONS]
        Options:
         --async                       Do not wait for the task
         --host[-id]                   Set the current host context for the request.
                                       Name/Id can be used
         --trace-ids TRACE_IDS         Array of Trace Ids
                                       Comma separated list of values. Values containing comma
                                       should be quoted or escaped with backslash. JSON is
                                       acceptable and preferred way for complex parameters
         -h, --help                    Print help
        """
        cls.command_sub = 'resolve'
```

ogajduse (changes requested):
Great to see the coverage for tracer! :tada: 
I left some comments.
@tstrych Can you please include also an API test containing the same logic, but using nailgun instead? Nailgun API wrapper functions are in place already.

tests/foreman/cli/test_host.py
```diff
@@ -2558,3 +2559,65 @@ def test_positive_dump_enc_yaml(default_sat):
...
+    [custom-rhel]
+    name=custom-rhel
+    baseurl={baseurl}
+    enabled=1
+    gpgcheck=0
+    """
+
+
+@pytest.mark.tier3
+def test_positive_traces(katello_host_tools_client, default_sat, katello_host_tools_repos):
```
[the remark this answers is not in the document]
tstrych: this is blocked by introduction of helper methods mentioned in #9113
and just created issue https://github.com/SatelliteQE/robottelo/issues/9202

you will install the tracer anyway, only question is where, and agree with you it can be in some fixture,  it's better  code immersion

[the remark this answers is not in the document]
ogajduse: I do not want to block this MR on this. Feel free to resolve this thread.

tests/foreman/cli/test_host.py
```diff
@@ -2558,3 +2559,65 @@ def test_positive_dump_enc_yaml(default_sat):
...
+
+# -------------------------- HOST TRACE SUBCOMMAND SCENARIOS -------------------------
+def generate_custom_rhel_repo_string(major, minor):
+    domain = settings.repos.rhel_os_repo_host
+    if major == 8:
+        baseurl = (
+            f'{domain}/rhel-{major}/rel-eng/RHEL-{major}/'
+            f'latest-RHEL-{major}.{minor}/compose/AppStream/x86_64/os/'
+        )
+    elif major == 7:
```
Gauravtalreja1, line 2572: Just a thought, can't we add this logic in `ContentHost.enable_repo` to enable this repos? 
It seems we already we have this repos in `conf/repos.yml`, so can leverage settings like `settings.repos.rhel7_os` and `settings.repos.rhel8_os.appstream`

tstrych: I will think about this

tstrych: I put it in api utils, but definitely makes sense to move it to content host, therefore you do not need to pass the arguments. So good idea thanks.

But usage of configuration constants you propose is another approach

approach I propose, you create the path, so you need to fix it everytime it will change your approach is better, because `conf/repos.constants` will be up to date, but still the path needs to be parsed because in this case `minor - 1` version is required, so if the path will change and you parse let say based on `/` you can still find yourself in position to fix this method, but you don't need to fix the text

tests/foreman/cli/test_host.py
```diff
@@ -2558,3 +2559,65 @@ def test_positive_dump_enc_yaml(default_sat):
...
+
+    client.add_rex_key(satellite=default_sat)
+    # generate file with older minor version
+    file_name = '/etc/yum.repos.d/rhel.repo'
+    file_content = generate_custom_rhel_repo_string(
+        major=client.os_version.major,
+        minor=client.os_version.minor - 1,
+    )
+    for line in file_content.split():
+        client.execute(f'echo {line} >> {file_name}')
```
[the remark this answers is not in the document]
tstrych: I found create_custom_repos in robottelo/hosts.py

tests/foreman/cli/test_host.py
tstrych: @Gauravtalreja1 
at the end i decided to not use `rhel8_os.appstream` because it's not garuanteed that `conf/repos` is same version as `conf/content_host` , right now i have rhel 8.4 and repo appstream for rhel 8.2

I would rather use the 8.3 here instead of  8.1 or 8.2

tstrych:
trigger: test-robottelo
pytest: tests/foreman/api/test_host.py::TestHostTraces::test_positive_tracer_list_and_resolve tests/foreman/cli/test_host.py::test_positive_tracer_list_and_resolve

ogajduse (changes requested):
Requesting changes in fixtures. Please implement the fixture that would yield host with tracer installed. It might be reused for future tracer test cases.
I would say that yielding the `rhel7_contenthost` object is better than yielding a `dict`.

pytest_fixtures/core/contenthosts.py
```diff
@@ -98,6 +99,21 @@ def registered_hosts(organization_ak_setup, content_hosts, default_sat):
...
+    based on hosts setup"""
+    rhel7_contenthost.install_katello_ca(default_sat)
+    rhel7_contenthost.register_contenthost(
+        setup_rhst_repo['org'].label,
+        setup_rhst_repo['ak'].name,
+    )
+    assert rhel7_contenthost.subscribed
+    rhel7_contenthost.enable_repo(REPOS[setup_rhst_repo['repo_name']]['id'])
+    rhel7_contenthost.install_katello_host_tools()
+    yield {'client': rhel7_contenthost, 'hostname': rhel7_contenthost.hostname}
```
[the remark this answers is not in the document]
ogajduse: Reopening. If the patch works, please refactor it also in other places where the `katello_host_tools_host` fixture is used.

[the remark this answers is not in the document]
ogajduse: However, I am not sure if it should be done in the scope of this PR.

tstrych:
trigger: test-robottelo
pytest: tests/foreman/api/test_host.py::TestHostTraces::test_positive_tracer_list_and_resolve tests/foreman/cli/test_host.py::test_positive_tracer_list_and_resolve

tests/foreman/cli/test_host.py
Gauravtalreja1: @tstrych Yes, we can keep the same versions in `conf/repos` and `conf/content_host`, updating in `satelliteqe-jenkins` MR#457, we missed this earlier when updating rhel8 versions, thanks!

Also, I just check we've `settings.repos.rhel8_repo` constant pointing to `sesame/pub/rhel8.repo`, now as sesame is off, not sure where we keep that. 

Sorry, I didn't get your point of using the `minor-1` version for repos, but if it is a requirement for the test then its fine, I wouldn't bother you :)

pytest_fixtures/core/contenthosts.py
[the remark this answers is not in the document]
tstrych: you are correct, probably not part of this issue, I will create a jira card issue, this is not causing any problems in automation as it's just bad code design and only in one module, which i take care of.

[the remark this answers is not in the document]
tstrych: DONE :)

tests/foreman/cli/test_host.py
tstrych: thanks to your comments I reworked it to use method that already exists in contenthost class so thanks :)

JacobCallahan:
@ogajduse can you revisit?

ogajduse (approved):
Wonderful! Thanks, @tstrych!

pondrejk (approved):

PR #9194, opened 2021-11-26
Ignore SSL errors on manifest download

rplevka:
Otherwise the method fails on downloading from servers with incorrectly configured SSL (e.g. self-signed certs)

### DIFF
--- a/robottelo/manifests.py
+++ b/robottelo/manifests.py
@@ -32,9 +32,9 @@ class ManifestCloner:
         """Download and cache the manifest information."""
         if self.template is None:
             self.template = {}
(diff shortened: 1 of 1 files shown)


jameerpathan111 (approved):

vsedmik (approved):

ogajduse:
@mshriver 
Do we want to install RH IT CA cert to the robottelo image and use
```python
requests.get("https://satqe-jenkins-csb-satellite-qe.ourinfra.com", verify="/etc/pki/ca-trust/source/anchors/2015-RH-IT-Root-CA.pem")
```
instead?

PR #9203, opened 2021-11-30
Fix and update api/cli endtoend tests to use rhel7 client

Gauravtalreja1:
**Test Results:**
```
$ pytest -q -p no:warnings tests/foreman/endtoend/test_api_endtoend.py::TestEndToEnd::test_positive_end_to_end
.                                                                                                                                 [100%]
1 passed in 721.67s (0:12:01)
```
```
$ pytest -q -p no:warnings tests/foreman/endtoend/test_cli_endtoend.py -k test_positive_cli_end_to_end
.                                                                                                                                 [100%]
1 passed, 3 deselected in 968.85s (0:16:08)
```

### DIFF
--- a/tests/foreman/endtoend/test_api_endtoend.py
+++ b/tests/foreman/endtoend/test_api_endtoend.py
@@ -27,6 +27,7 @@ from fauxfactory import gen_string
 from nailgun import client
 from nailgun import entities
 
+from robottelo import constants
 from robottelo import manifests
 from robottelo.api.utils import enable_rhrepo_and_fetchid
 from robottelo.api.utils import promote
@@ -35,18 +36,9 @@ from robottelo.config import get_credentials
 from robottelo.config import get_url
 from robottelo.config import setting_is_set
 from robottelo.config import settings
-from robottelo.constants import DEFAULT_LOC
-from robottelo.constants import DEFAULT_ORG
-from robottelo.constants import DEFAULT_SUBSCRIPTION_NAME
-from robottelo.constants import PRDS
-from robottelo.constants import REPOS
-from robottelo.constants import REPOSET
 from robottelo.constants.repos import CUSTOM_RPM_REPO
 from robottelo.helpers import get_nailgun_config
 from robottelo.utils.issue_handlers import is_open
-
-
-AK_CONTENT_LABEL = 'rhel-6-server-rhev-agent-rpms'
 
 
 API_PATHS = {
@@ -1029,9 +1021,11 @@ class TestEndToEnd:
 
         :expectedresults: 'Default Organization' is found
         """
-        results = entities.Organization().search(query={'search': f'name="{DEFAULT_ORG}"'})
+        results = entities.Organization().search(
+            query={'search': f'name="{constants.DEFAULT_ORG}"'}
+        )
         assert len(results) == 1
-        assert results[0].name == DEFAULT_ORG
+        assert results[0].name == constants.DEFAULT_ORG
 
     @pytest.mark.build_sanity
     def test_positive_find_default_loc(self):
@@ -1041,9 +1035,9 @@ class TestEndToEnd:
 
         :expectedresults: 'Default Location' is found
         """
-        results = entities.Location().search(query={'search': f'name="{DEFAULT_LOC}"'})
+        results = entities.Location().search(query={'search': f'name="{constants.DEFAULT_LOC}"'})
         assert len(results) == 1
-        assert results[0].name == DEFAULT_LOC
+        assert results[0].name == constants.DEFAULT_LOC
 
     @pytest.mark.build_sanity
     def test_positive_find_admin_user(self):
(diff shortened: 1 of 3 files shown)


tests/foreman/endtoend/test_api_endtoend.py
```diff
@@ -1140,34 +1136,33 @@ def test_positive_end_to_end(self, fake_manifest_is_set, default_sat, rhel6_cont
         repositories = []
 
         # step 2.5: Create custom YUM repository
-        repo1 = entities.Repository(
+        yum_repo = entities.Repository(
```
jyejare, line 1139:
```suggestion
        custom_repo = entities.Repository(
```

Optional, non-blocking comment!

tests/foreman/endtoend/test_api_endtoend.py
```diff
@@ -1140,34 +1136,33 @@ def test_positive_end_to_end(self, fake_manifest_is_set, default_sat, rhel6_cont
...
                     basearch='x86_64',
                     org_id=org.id,
-                    product=PRDS['rhel'],
-                    repo=REPOS['rhva6']['name'],
-                    reposet=REPOSET['rhva6'],
-                    releasever='6Server',
+                    product=constants.PRDS['rhel'],
+                    repo=constants.REPOS['rhst7']['name'],
+                    reposet=constants.REPOSET['rhst7'],
+                    releasever=None,
```
jyejare, line 1153: Instead update `releasever` in `enable_rhrepo_and_fetchid` to set default value to None!

tests/foreman/endtoend/test_api_endtoend.py
```diff
@@ -1228,21 +1227,21 @@ def test_positive_end_to_end(self, fake_manifest_is_set, default_sat, rhel6_cont
...
-        rhel6_contenthost.register_contenthost(org.label, activation_key_name)
-        assert rhel6_contenthost.subscribed
-        # Install rpm on client
-        package_name = 'python-kitchen'
-        result = rhel6_contenthost.execute(f'yum install -y {package_name}')
-        assert result.status == 0
-        # Verify that the package is installed by querying it
-        result = rhel6_contenthost.run(f'rpm -q {package_name}')
-        assert result.status == 0
+        with VMBroker(nick='rhel7', host_classes={'host': ContentHost}) as client:
```
jyejare, line 1236: What's your thought on running this for multiple RHEL versions? If yes, use https://github.com/SatelliteQE/robottelo/blob/master/pytest_fixtures/core/contenthosts.py#L32

jyejare (commented):
ACK pending question.

omkarkhatavkar (approved):

tests/foreman/endtoend/test_api_endtoend.py
Gauravtalreja1: @jyejare Yes, we can consider this in the future, maybe in 7.0 with single client tools repos.
currently, we've separate tools repos for each rhel, which will make this test more complicated.

tests/foreman/endtoend/test_api_endtoend.py
Gauravtalreja1: Updated.

tests/foreman/endtoend/test_api_endtoend.py
Gauravtalreja1: Updated

tests/foreman/endtoend/test_api_endtoend.py
ogajduse: Please use `rhel7_contenthost` fixture for this purpose. Do not call broker directly from the test unless you have a strong reason for it.

tests/foreman/endtoend/test_cli_endtoend.py
```diff
@@ -335,17 +329,18 @@ def test_positive_cli_end_to_end(fake_manifest_is_set, rhel6_contenthost, defaul
...
-    rhel6_contenthost.register_contenthost(org['label'], activation_key['name'])
-    assert rhel6_contenthost.subscribed
-    # Install rpm on client
-    package_name = 'python-kitchen'
-    result = rhel6_contenthost.execute(f'yum install -y {package_name}')
-    assert result.status == 0
-    # Verify that the package is installed by querying it
-    result = rhel6_contenthost.run(f'rpm -q {package_name}')
-    assert result.status == 0
+    with VMBroker(nick='rhel7', host_classes={'host': ContentHost}) as client:
```
ogajduse, line 332: Same here, use `rhel7_contenthost` fixture instead of directly calling broker.

ogajduse (changes requested):
The current design suggests using content host fixtures instead of directly calling broker from the test function. Using any content host fixture helps with test filtering. Tests can be filtered using the pytest's `-k` argument, which is exactly what we do in the PIT pipeline.

Please use the `rhel7_contenthost` fixture unless you have strong to keep the current implementation.

tests/foreman/endtoend/test_api_endtoend.py
jyejare: @Gauravtalreja1 No it should not be that challenging with ContentHost class supporting - https://github.com/SatelliteQE/robottelo/blob/master/robottelo/hosts.py#L228.

tests/foreman/endtoend/test_cli_endtoend.py
Gauravtalreja1: @ogajduse Updated 👍

tests/foreman/endtoend/test_api_endtoend.py
jyejare: @Gauravtalreja1 Since the test focuses on fixing the test, I wont force here to run it for all RHEL versions. @JacobCallahan @mshriver may confirm that we should soon have plans to run the client test on all RHEL versions across robottelo !

PR #9204, opened 2021-11-30
Test: positive create with new location and organization

sambible:
Writing a test to cover https://bugzilla.redhat.com/show_bug.cgi?id=1992967 

First test written, give me all the feedback :smiling_face_with_tear:

### DIFF
--- a/tests/foreman/cli/test_repository.py
+++ b/tests/foreman/cli/test_repository.py
@@ -36,6 +36,7 @@ from robottelo.cli.factory import make_content_credential
 from robottelo.cli.factory import make_content_view
 from robottelo.cli.factory import make_filter
 from robottelo.cli.factory import make_lifecycle_environment
+from robottelo.cli.factory import make_location
 from robottelo.cli.factory import make_org
 from robottelo.cli.factory import make_product
 from robottelo.cli.factory import make_repository
@@ -44,6 +45,7 @@ from robottelo.cli.factory import make_user
 from robottelo.cli.file import File
 from robottelo.cli.filter import Filter
 from robottelo.cli.module_stream import ModuleStream
+from robottelo.cli.org import Org
 from robottelo.cli.package import Package
 from robottelo.cli.product import Product
 from robottelo.cli.repository import Repository
@@ -559,6 +561,34 @@ class TestRepository:
         """
         for key in 'url', 'content-type', 'name':
             assert repo.get(key) == repo_options[key]
+
+    @pytest.mark.tier1
+    def test_positive_create_with_new_organization_and_location(self):
+        """Check if repo can be created with a new Organization and Location.
+
+        :parametrized: no
+
+        :expectedresults: Repository is successrully created.
+
+        :BZ: 1992967
+
+        "CaseImportance: Critical
+        """
+        new_org = make_org()
+        new_location = make_location()
+        new_product = make_product(
+            {'organization-id': new_org['id'], 'description': 'test_product'}
+        )
+        Org.add_location({'location-id': new_location['id'], 'name': new_org['name']})
+        assert Org.info({'id': new_org['id']})['locations'][0] == new_location['name']
+        new_repo = make_repository(
+            {
+                'location-id': new_location['id'],
+                'organization-id': new_org['id'],
+                'product-id': new_product['id'],
+            }
+        )
+        assert new_repo['organization'] == new_org['name']
 
     @pytest.mark.tier1
     @pytest.mark.parametrize(
--- a/tests/foreman/cli/test_repository.py
+++ b/tests/foreman/cli/test_repository.py
@@ -596,7 +596,6 @@ class TestRepository:
             "grep \"undefined method `id' for nil:NilClass (NoMethodError)\""
         )
         assert result.status == 1
-        assert result.stdout == ""
 
     @pytest.mark.tier1
     @pytest.mark.parametrize(
--- a/tests/foreman/cli/test_repository.py
+++ b/tests/foreman/cli/test_repository.py
@@ -593,9 +593,10 @@ class TestRepository:
 
         result = default_sat.execute(
             "cat /var/log/foreman/production.log | "
-            "grep \"NoMethodError: undefined method `id' for nil:NilClass\""
+            "grep \"undefined method `id' for nil:NilClass (NoMethodError)\""
         )
-        assert result == 0
+        assert result.status == 1
+        assert result.stdout == ""
 
     @pytest.mark.tier1
     @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_repository.py
```diff
@@ -560,6 +562,34 @@ def test_positive_create_docker_repo_with_name(self, repo_options, repo):
...
+    def test_positive_create_with_new_organization_and_location(self):
+        """Check if repo can be created with a new Organization and Location.
+
+        :parametrized: no
+
+        :expectedresults: Repository is successrully created.
+
+        :BZ: 1992967
+
+        "CaseImportance: Critical
```
synkd, line 575: This should be `:CaseImportance:`

synkd (changes requested):
In looking at the BZ, the original issue was not that the repo was not created, it was that an error was displayed despite the fact that the repo was created successfully. So it looks to me like this test would pass regardless of whether or not the issue reported in the BZ was present. I would consider refactoring this test to verify that the error is not present rather than verifying that the repo was created successfully.

However, it's not immediately clear from the BZ whether this issue was present only via CLI, only via UI, or in both cases. It might be worth checking with Jameer or spinning up a 6.10.0 snap 12 to see where this error occurs.

sambible:
Must have misread that then. I can easily check for that. I can also spin up the old snap and see behavior, to make sure it's covered in all cases.

tests/foreman/cli/test_repository.py
```diff
@@ -560,6 +562,34 @@ def test_positive_create_docker_repo_with_name(self, repo_options, repo):
...
+        """
+        new_org = make_org()
+        new_location = make_location()
+        new_product = make_product(
+            {'organization-id': new_org['id'], 'description': 'test_product'}
+        )
+        Org.add_location({'location-id': new_location['id'], 'name': new_org['name']})
```
ColeHiggins2, line 582: Should we assign this to a variable and then assert on it below instead of a direct call ?

ColeHiggins2: Actually, should be fine

tests/foreman/cli/test_repository.py
```diff
@@ -560,6 +562,34 @@ def test_positive_create_docker_repo_with_name(self, repo_options, repo):
         for key in 'url', 'content-type', 'name':
             assert repo.get(key) == repo_options[key]
 
+    @pytest.mark.tier1
+    def test_positive_create_with_new_organization_and_location(self):
+        """Check if repo can be created with a new Organization and Location.
+
+        :parametrized: no
```
jameerpathan111, line 569: @sambible you will also have to add `:id:`. You can fix it by running `pre-commit run -a` or 
```
`In [1]: import uuid

In [2]: print(uuid.uuid4())
04af6170-2eed-4d0d-be8d-b5b86c7ce678
```
Non blocking comment: You can add test steps(`:Steps:`) as well.

sambible:
Changed the test to check for the presence of the error inside the foreman production log, as that is where it was referenced inside the original BZ. Might be a better way to do it, but this is the first solution I came up with.

synkd:
Looks good to me. Let's trigger robottelo-runner to make sure the test passes in CI.

synkd:
trigger: test-robottelo
pytest: tests/foreman/cli/test_repository.py::TestRepository::test_positive_create_with_new_organization_and_location

sambible:
trigger: test-robottelo
pytest: tests/foreman/cli/test_repository.py::TestRepository::test_positive_create_with_new_organization_and_location

synkd (approved):

synkd:
ACK

tests/foreman/cli/test_repository.py
```diff
@@ -560,6 +562,34 @@ def test_positive_create_docker_repo_with_name(self, repo_options, repo):
         for key in 'url', 'content-type', 'name':
             assert repo.get(key) == repo_options[key]
 
+    @pytest.mark.tier1
+    def test_positive_create_with_new_organization_and_location(self):
+        """Check if repo can be created with a new Organization and Location.
+
```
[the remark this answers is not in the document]
sambible: yeah, learned that's how those get generated today

tests/foreman/cli/test_repository.py
```diff
@@ -560,6 +562,42 @@ def test_positive_create_docker_repo_with_name(self, repo_options, repo):
...
+                'product-id': new_product['id'],
+            }
+        )
+
+        result = default_sat.execute(
+            "cat /var/log/foreman/production.log | "
+            "grep \"undefined method `id' for nil:NilClass (NoMethodError)\""
+        )
+        assert result.status == 1
+        assert result.stdout == ""
```
akhil-jha, line 599: Any reason why we are asserting stdout for empty string? If the status is 1, shouldn't that be enough here?

sambible:
trigger: test-robottelo
pytest: tests/foreman/cli/test_repository.py::TestRepository::test_positive_create_with_new_organization_and_location

tests/foreman/cli/test_repository.py
```diff
@@ -560,6 +562,41 @@ def test_positive_create_docker_repo_with_name(self, repo_options, repo):
...
+
+        :id: 9ea4f2a9-f339-4215-b301-cd39c6b5c474
+
+        :parametrized: no
+
+        :expectedresults: No error is present when Repository is created
+
+        :BZ: 1992967
+
+        :CaseImportance: Critical
```
swadeley, line 577:
```suggestion
        :CaseImportance: High
```

because bug comment 0 says " I can see that the repository is created despite of the error shown."

swadeley (approved):
ACK.
Don't forget to change the  "qe_test_coverage" flag in the BZ to "+"

tests/foreman/cli/test_repository.py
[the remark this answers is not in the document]
swadeley: `uuidgen` also works

swadeley:
> However, it's not immediately clear from the BZ whether this issue was present only via CLI, only via UI, or in both cases. It might be worth checking with Jameer 

@jameerpathan111 it would be nice to know that^

ColeHiggins2 (approved):
ACK

tests/foreman/cli/test_repository.py
```diff
@@ -560,6 +562,41 @@ def test_positive_create_docker_repo_with_name(self, repo_options, repo):
         for key in 'url', 'content-type', 'name':
             assert repo.get(key) == repo_options[key]
 
+    @pytest.mark.tier1
+    def test_positive_create_with_new_organization_and_location(self, default_sat):
```
latran, line 566: Rename to `test_positive_create_repo_with_new_organization_and_location`

latran (changes requested):

tests/foreman/cli/test_repository.py
swadeley: and because there are already lots of Critical tests that create repos

latran (approved):
ACK

swadeley (approved):
ACK

PR #9212, opened 2021-12-02
Metadata for RHEL version

shweta83:
RHEL VERSION is now configurable in robottelo.yaml

### DIFF
--- a/conf/robottelo.yaml.template
+++ b/conf/robottelo.yaml.template
@@ -15,4 +15,6 @@ ROBOTTELO:
   RUN_ONE_DATAPOINT: false
   # Satellite version supported by this branch
   # UNDR version is used for some URL composition
-  SATELLITE_VERSION: "6.9"
+  SATELLITE_VERSION: "7.0"
+  # The Base OS RHEL Version(x.y) where the satellite would be installed
+  RHEL_VERSION: "7.9"
--- a/conf/server.yaml.template
+++ b/conf/server.yaml.template
@@ -8,6 +8,8 @@ SERVER:
     # RELEASE: populate with satellite version
     # The snap version currently testing (if applicable)
     # SNAP:
+    # The RHEL Base OS Version(x.y) where the Satellite is installed
+    # RHEL_VERSION:
     # The source of Satellite packages. Can be one of:
     # internal, ga, beta
     SOURCE: "internal"
--- a/pytest_plugins/metadata_markers.py
+++ b/pytest_plugins/metadata_markers.py
@@ -4,6 +4,7 @@ import re
 
 import pytest
 
+from robottelo.hosts import get_sat_rhel_version
 from robottelo.logging import collection_logger as logger
 
 FMT_XUNIT_TIME = '%Y-%m-%dT%H:%M:%S'
@@ -69,6 +70,9 @@ def pytest_collection_modifyitems(session, items, config):
     Control test collection for custom options related to testimony metadata
 
     """
+    # get RHEL version of the satellite
+    rhel_version = get_sat_rhel_version().base_version
+
     # split the option string and handle no option, single option, multiple
     # config.getoption(default) doesn't work like you think it does, hence or ''
     importance = [i for i in (config.getoption('importance') or '').split(',') if i != '']
@@ -107,6 +111,7 @@ def pytest_collection_modifyitems(session, items, config):
         # add markers as user_properties so they are recorded in XML properties of the report
         for marker in item.iter_markers():
             item.user_properties.append((marker.name, next(iter(marker.args), None)))
+        item.user_properties.append(("BaseOS", rhel_version, None))
         item.user_properties.append(
             ("start_time", datetime.datetime.utcnow().strftime(FMT_XUNIT_TIME))
         )
@@ -154,6 +159,7 @@ def pytest_collection_modifyitems(session, items, config):
                 )
                 deselected.append(item)
                 continue
+
             selected.append(item)
 
     # selected will be empty if no filter option was passed, defaulting to full items list
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -36,7 +36,6 @@ from robottelo.helpers import get_data_file
 from robottelo.helpers import InstallerCommand
 from robottelo.helpers import validate_ssh_pub_key
 
-
 POWER_OPERATIONS = {
     VmState.RUNNING: 'running',
     VmState.STOPPED: 'stopped',
@@ -59,6 +58,20 @@ def get_sat_version():
         else:
             sat_version = SATELLITE_VERSION
     return Version('9999' if 'nightly' in sat_version else sat_version)
+
+
+def get_sat_rhel_version():
+    """Try to read rhel_version from Satellite host
+    if not available fallback to robottelo configuration."""
+
+    try:
+        rhel_version = Satellite().os_version
+    except (AuthenticationError, ContentHostError, BoxKeyError):
+        if hasattr(settings.server.version, 'rhel_release'):
+            rhel_version = str(settings.server.version.rhel_release)
+        elif hasattr(settings.robottelo, 'rhel_version'):
+            rhel_version = settings.robottelo.rhel_version
+    return Version(rhel_version)
 
 
 def setup_capsule(satellite, capsule, registration_args=None, installation_args=None):


conf/server.yaml.template
```diff
@@ -8,6 +8,8 @@ SERVER:
     # RELEASE: populate with satellite version
     # The snap version currently testing (if applicable)
     # SNAP:
+    # The base os rhel version where the satellite installed
+    # RHEL_RELEASE: 7.9
```
ogajduse, line 12: I would keep consistent naming here.
```suggestion
    # RHEL_VERSION: 7.9
```

conf/server.yaml.template
```diff
@@ -8,6 +8,8 @@ SERVER:
     # RELEASE: populate with satellite version
     # The snap version currently testing (if applicable)
     # SNAP:
+    # The base os rhel version where the satellite installed
```
ogajduse, line 11:
```suggestion
    # The RHEL Base OS Version(x.y) where the Satellite is installed
```

robottelo/hosts.py
```diff
@@ -61,6 +61,20 @@ def get_sat_version():
     return Version('9999' if 'nightly' in sat_version else sat_version)
 
 
+def get_sat_rhel_version():
+    """Try to read rhel_version from Satellite host
+    if not available fallback to DynaConfs."""
```
ogajduse, line 66: We might decide to change the configuration manager in the future. So I would keep the names general.
```suggestion
    if not available fallback to robottelo configuration."""
```

ogajduse (changes requested):
Suggesting minor changes in comments and variable names.

@shweta83 Did you encounter any specific case where you were missing the functionality that you have added?

pondrejk (approved):
ack pending comments

jyejare:
Ping @shweta83 , please resolve @ogajduse comments !

robottelo/hosts.py
shweta83: Updated as suggested.

conf/server.yaml.template
shweta83: Updated as suggested.

conf/server.yaml.template
shweta83: Updated as suggested.

shweta83:
> ack pending comments

Updated as suggested.

conf/server.yaml.template
ogajduse: @shweta83 Please enclose the value for `RHEL_VERSION` with quotes the same way you did it in `conf/robottelo.yaml.template`. String type should protect us from `6.10` being `6.1`. I know it is just a template, but people frequently copy the templates.

shweta83: Updated.

latran (approved):
ack

jyejare (changes requested):
@shweta83 it seems you missed the gist of this PR :)

You need to use the `get_sat_rhel_version` def and attach the return value(the rhel version) to the test-results XML as metadata!

shweta83:
> @shweta83 it seems you missed the gist of this PR :)
> 
> You need to use the `get_sat_rhel_version` def and attach the return value(the rhel version) to the test-results XML as metadata!

True @jyejare. I have updated the PR with the changes.

conf/robottelo.yaml.template
```diff
@@ -16,3 +16,5 @@ ROBOTTELO:
   # Satellite version supported by this branch
   # UNDR version is used for some URL composition
   SATELLITE_VERSION: "6.9"
```
jyejare, line 18: This should be  updated now to `7.0`!

conf/server.yaml.template
```diff
@@ -8,6 +8,8 @@ SERVER:
     # RELEASE: populate with satellite version
     # The snap version currently testing (if applicable)
     # SNAP:
+    # The RHEL Base OS Version(x.y) where the Satellite is installed
+    # RHEL_VERSION: "7.9"
```
jyejare, line 12: Leave the default value as blank similar to satellite release and snap versions!

pytest_plugins/metadata_markers.py
```diff
@@ -24,6 +25,10 @@ def pytest_addoption(parser):
         '--assignee',
         help='Comma separated list of assignees to include in test collection',
     )
+    parser.addoption(
```
jyejare, line 28: We dont need the BaseOS to filter the test collection because we dont have tests/components that work on satellite installed on Base RHEL8 but not on Base RHEL7. It's like, all tests should work on RHEL7 and RHEL8 so there is no choice to choose the tests based on BaseOS. 

So just remove this pytest option!

pytest_plugins/metadata_markers.py
```diff
@@ -32,6 +37,7 @@ def pytest_configure(config):
         'importance: CaseImportance testimony token, use --importance to filter',
         'component: Component testimony token, use --component to filter',
         'assignee: Assignee testimony token, use --assignee to filter',
+        'BaseOS: BaseOS testimony token, use --BaseOS to filter',
```
jyejare, line 40: Remove the option as per above comment !

pytest_plugins/metadata_markers.py
```diff
@@ -74,6 +86,7 @@ def pytest_collection_modifyitems(session, items, config):
     importance = [i for i in (config.getoption('importance') or '').split(',') if i != '']
     component = [c for c in (config.getoption('component') or '').split(',') if c != '']
     assignee = [a for a in (config.getoption('assignee') or '').split(',') if a != '']
+    BaseOS = [z for z in (config.getoption('BaseOS') or '').split(',') if z != '']
```
jyejare, line 89: Remove this as per above comment !

pytest_plugins/metadata_markers.py
```diff
@@ -103,6 +116,12 @@ def pytest_collection_modifyitems(session, items, config):
             doc_assignee = assignee_regex.findall(docstring)
             if doc_assignee and 'assignee' not in item_mark_names:
                 item.add_marker(pytest.mark.assignee(doc_assignee[0]))
+            doc_baseos = baseos_regex.findall(docstring)
+            if doc_baseos and 'BaseOS' not in item_mark_names:
+                item.add_marker(pytest.mark.BaseOS(doc_baseos[0]))
+            else:
+                rhel_version = get_sat_rhel_version().base_version
+                item.add_marker(pytest.mark.BaseOS(rhel_version))
```
jyejare, line 124: Remove this as Satellite BaseOS marker wont help with any purpose for now or in recent future.

pytest_plugins/metadata_markers.py
```diff
@@ -122,7 +141,7 @@ def pytest_collection_modifyitems(session, items, config):
             )
 
         # exit early if no filters were passed
-        if importance or component or assignee:
+        if importance or component or assignee or BaseOS:
```
jyejare, line 144: Remove !

pytest_plugins/metadata_markers.py
```diff
@@ -154,6 +173,14 @@ def pytest_collection_modifyitems(session, items, config):
...
                 deselected.append(item)
                 continue
+            BaseOS_marker = item.get_closest_marker('BaseOS').args[0]
+            if BaseOS and BaseOS_marker not in BaseOS:
+                logger.debug(
+                    f'Deselected test {item.nodeid} due to "--BaseOS {BaseOS}",'
+                    f'test has BaseOS mark: {BaseOS_marker}'
+                )
+                deselected.append(item)
+                continue
```
jyejare, line 183: Not required since we are not filtering tests based on satellite BaseOS as per the above comment!

Please remove!

jyejare (changes requested):
Overall it seems we are trying to implement the markers for Satellite BaseOS and filter the tests based on that BaseOS marker. 

We are not looking for that from this PR but we are looking for each test item to have a BaseOS metadata (User Property) that would help up to filter the RHEL7 and RHEL8 tests in ibutsu like tool.

Please refer the good example of Metadata addition to test item is https://github.com/SatelliteQE/robottelo/blob/3cd6050840836ff2d402b2b98c803bc7584732e5/pytest_plugins/metadata_markers.py#L110 

Also, Use `SatBaseOS` key for user_property instead of `BaseOS` else this will confuse between Client Version and Sat BaseOS version for client tests !

pytest_plugins/metadata_markers.py
shweta83: @jyejare I added this as an option to filter out the the tests based on RHEL version if there will be different tests for RHEL7 and RHEL8 in coming satellite versions but since we don't have that use case in near future then I can remove the marker and just keeping the user_properties.

robottelo/hosts.py
```diff
@@ -61,6 +61,20 @@ def get_sat_version():
     return Version('9999' if 'nightly' in sat_version else sat_version)
 
 
+def get_sat_rhel_version():
```
jyejare, line 64: I would like to see this cached while seeding this to the user_properties, be it call outside the loop  and use or cachec this function is fine !

robottelo/hosts.py
```diff
@@ -45,6 +45,7 @@
 }
 
 
+@cached_property
```
JacobCallahan, line 48: you would actually use the `@cached` decorator here. `cached_property` is for use in classes. However, I am going to recommend against caching.

JacobCallahan (changes requested):

robottelo/hosts.py
shweta83: > We can actually ignore caching here since there could be instances when `settings.server.hostname` is changed and this method is used. If a cached version was used, then an incorrect version could be returned.

@JacobCallahan @jyejare Also, caching didn't make much difference in execution time so I removed the caching for now.

JacobCallahan (approved):
ACK, nice job!

robottelo/hosts.py
jyejare: @shweta83 Right, it won't because you have stored the RHEL version before you started looping over items.

@JacobCallahan Even though `settings.server.hostname` changes, there is 0% chance that test will ask to deploy RHEL8 satellite when tests running on RHEL7 pipeline.

jyejare (approved):

PR #9216, opened 2021-12-02
test: Unable to select organization for Sync management

sambible:
Creating a manual test case for https://bugzilla.redhat.com/show_bug.cgi?id=1860957

There has been discussion around automating this, and it was determined to be more work than value. Not sure if this is the correct place to stick this, would like feedback.

### DIFF
--- a/tests/foreman/ui/test_repository.py
+++ b/tests/foreman/ui/test_repository.py
@@ -1062,3 +1062,27 @@ def test_positive_rpm_metadata_display():
         correctly.
     """
     pass
+
+
+@pytest.mark.stubbed
+@pytest.mark.tier2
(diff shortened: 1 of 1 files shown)


akhil-jha (approved):
Not a blocker, but you could add caseimportance as well. I test_repository.py has a caseimportance of high, but the test doesn't seem high to me. More like medium.

jyejare (approved):
ACK, this is the most common workflow customers might access the sync status page, I mean the chances are high that customers will route to the sync status page this way. 

So let's keep it as high importance only.

ColeHiggins2 (approved):
ACK

sambible:
Added case importance, will merge after all checks run

PR #9218, opened 2021-12-06
Update the organization to module_manifest_org for virtwho plugin cases 

hkx303:
**Description**
Update the organization from `default_org` to `module_manifest_org` for virtwho plugin cases to fix failed cases. Those errors were imported from https://github.com/SatelliteQE/robottelo/pull/8993

**Test Result**
All test cases work fine locally.

### DIFF
--- a/robottelo/virtwho_utils.py
+++ b/robottelo/virtwho_utils.py
@@ -274,16 +274,19 @@ def deploy_configure_by_command(command, hypervisor_type, debug=False, org='Defa
         return deploy_validation(hypervisor_type)
 
 
-def deploy_configure_by_script(script_content, hypervisor_type, debug=False):
+def deploy_configure_by_script(
+    script_content, hypervisor_type, debug=False, org='Default_Organization'
+):
     """Deploy and run virt-who service by the shell script.
     :param str script_content: get the script by UI or API.
     :param str hypervisor_type: esx, libvirt, rhevm, xen, libvirt, kubevirt
     :param bool debug: if VIRTWHO_DEBUG=1, this option should be True.
+    :param str org: Organization Label
     """
     script_filename = "/tmp/deploy_script.sh"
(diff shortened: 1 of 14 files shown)


robottelo/virtwho_utils.py
```diff
@@ -407,22 +410,23 @@ def virtwho_package_locked():
     assert "Packages are locked" in result[1]
 
 
-def create_http_proxy(name=None, url=None, type='https'):
+def create_http_proxy(name=None, url=None, http_type='https', org_name=DEFAULT_ORG):
```
mshriver, line 413: Only thing I would recommend here is to pass the `org` instance, and reference the field you need within this function (name, label, etc)

mshriver (approved):

robottelo/virtwho_utils.py
```diff
@@ -407,22 +410,23 @@ def virtwho_package_locked():
...
     :param url: URL of the proxy including schema (https://proxy.example.com:8080)
-    :param type: https or http
+    :param http_type: https or http
+    :param org_name: name of the organization
     :return:
     """
-    default_org = entities.Organization().search(query={'search': f'name="{DEFAULT_ORG}"'})[0]
+    org = entities.Organization().search(query={'search': f'name="{org_name}"'})[0]
     http_proxy_name = name or gen_string('alpha', 15)
     http_proxy_url = url or '{}:{}'.format(
```
swadeley, line 424: Can you try use f string here? I notice there are quite a few ".format" in the file, a job for another PR to fix them.

robottelo/virtwho_utils.py
hkx303: Updated, thanks

robottelo/virtwho_utils.py
hkx303: Good catch! Copied that from another file before, updated.

JacobCallahan:
@swadeley can you revisit?

swadeley (approved):
ACK

PR #9220, opened 2021-12-06
#9215 package_install with ReX

### LINKED ISSUE #9215: Host.package_install fails: This action uses katello-agent, which is currently disabled
```
=================================================== FAILURES ====================================================
_____________________________________ test_positive_generate_hostpkgcompare _____________________________________

local_org = {'compute-resources': {}, 'created-at': '2021/12/01 16:14:08', 'description': {}, 'domains': {}, ...}
local_ak = {'auto-attach': 'true', 'content-view': 'cBnrWEvauS', 'description': {}, 'host-collections': {}, ...}
local_content_view = {'activation-keys': {}, 'components': {}, 'composite': 'false', 'container-image-repositories': {}, ...}
local_environment = {'description': {}, 'id': '33', 'label': 'I28J9H', 'library': 'false', ...}
default_sat = <robottelo.hosts.Satellite object at 0x7f908f670760>

    @pytest.mark.tier3
    def test_positive_generate_hostpkgcompare(
        local_org, local_ak, local_content_view, local_environment, default_sat
    ):
        """Generate 'Host - compare content hosts packages' report
    
        :id: 572fb387-86f2-40e2-b2df-e8ec26433610
    
    
        :setup: Installed Satellite with Organization, Activation key,
                Content View, Content Host, Subscriptions, and synced fake repo.
    
        :steps:
            1. hammer report-template generate --name 'Host - compare content hosts package' ...
    
        :expectedresults: report is scheduled and generated containing all the expected information
                          regarding host packages
    
        :CaseImportance: Medium
    
        :BZ: 1860430
        """
        # Add subscription to Satellite Tools repo to activation key
        setup_org_for_a_rh_repo(
            {
                'product': PRDS['rhel'],
                'repository-set': REPOSET['rhst7'],
                'repository': REPOS['rhst7']['name'],
                'organization-id': local_org['id'],
                'content-view-id': local_content_view['id'],
                'lifecycle-environment-id': local_environment['id'],
                'activationkey-id': local_ak['id'],
            }
        )
        setup_org_for_a_custom_repo(
            {
                'url': settings.repos.yum_6.url,
                'organization-id': local_org['id'],
                'content-view-id': local_content_view['id'],
                'lifecycle-environment-id': local_environment['id'],
                'activationkey-id': local_ak['id'],
            }
        )
    
        hosts_info = []
        with VMBroker(nick='rhel7', host_classes={'host': ContentHost}, _count=2) as hosts:
            for client in hosts:
                # Create RHEL hosts via broker and register content host
                client.install_katello_ca(default_sat)
                # Register content host, install katello-agent
                client.register_contenthost(local_org['label'], local_ak['name'])
                assert client.subscribed
                hosts_info.append(Host.info({'name': client.hostname}))
                client.enable_repo(REPOS['rhst7']['id'])
                client.install_katello_agent()
            hosts_info.sort(key=lambda host: host['name'])
    
            host1, host2 = hosts_info
>           Host.package_install({'host-id': host1['id'], 'packages': FAKE_0_CUSTOM_PACKAGE_NAME})

tests/foreman/cli/test_reporttemplates.py:916: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
robottelo/cli/host.py:132: in package_install
    return cls.execute(cls._construct_command(options), output_format='csv')
robottelo/cli/base.py:253: in execute
    return cls._handle_response(response, ignore_stderr=ignore_stderr)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <class 'robottelo.cli.host.Host'>, response = , ignore_stderr = None

    @classmethod
    def _handle_response(cls, response, ignore_stderr=None):
        """Verify ``status`` of the CLI command.
    
        Check for a non-zero return code or any stderr contents.
    
        :param response: a result object, returned by :mod:`robottelo.ssh.command`.
        :param ignore_stderr: indicates whether to throw a warning in logs if
            ``stderr`` is not empty.
        :returns: contents of ``stdout``.
        :raises robottelo.cli.base.CLIReturnCodeError: If return code is
            different from zero.
        """
        if isinstance(response.stderr, tuple):
            # the current contents of response.stderr is a tuple with the following properties
            # (<len(message)>,<message>).
            # This behavior could (and maybe should) change in the near future.
            # In the meantime, we don't need it here so we just use the message itself
            response.stderr = response.stderr[1]
        if isinstance(response.stderr, bytes):
            response.stderr = response.stderr.decode()
        if response.status != 0:
            full_msg = (
                f'Command "{cls.command_base} {cls.command_sub}" '
                f'finished with status {response.status}\n'
                f'stderr contains:\n{response.stderr}'
            )
            error_data = (response.status, response.stderr, full_msg)
            if cls._db_error_regex.search(full_msg):
                raise CLIDataBaseError(*error_data)
>           raise CLIReturnCodeError(*error_data)
E           robottelo.cli.base.CLIReturnCodeError: CLIReturnCodeError(status=128, stderr='This command uses katello agent and will be removed in favor of remote execution in a future release.\nThe remote execution equivalent is `hammer job-invocation create --feature katello_package_install`.\nCould not install packages:\n  This action uses katello-agent, which is currently disabled. Use remote execution instead.\n', msg='Command "host package install" finished with status 128\nstderr contains:\nThis command uses katello agent and will be removed in favor of remote execution in a future release.\nThe remote execution equivalent is `hammer job-invocation create --feature katello_package_install`.\nCould not install packages:\n  This action uses katello-agent, which is currently disabled. Use remote execution instead.\n'

robottelo/cli/base.py:104: CLIReturnCodeError
```

lhellebr:
This needs discussion before getting merged.
This is intended to fix #9215 where a method that is being deprecated is being used for package installation, resulting to a failure. I decided to fix this by the way recommended in the error message, that is, by using remote execution. I decided to not stick to backwards compatibility since it would make the function usage less straightforward for the future and it shouldn't be hard to fix all usages of this function in this repo.
The fix is in `host.py` and an example of the required changes per usage are in `test_reporttemplates.py`.
Things to consider:
- Is it ok to make this backwards incompatible?
- Should all the invocations of the function be used as part of this PR, or as a separate PR?
- Is it a good decision to use ReX, as recommended?
- Is it ok that `add_rex_key` must be called beforehand or should it be called as part of `package_install` function? If the latter, it would require the function to obtain a host object.

### DIFF
--- a/robottelo/cli/host.py
+++ b/robottelo/cli/host.py
@@ -126,10 +126,13 @@ class Host(Base):
         return facts
 
     @classmethod
-    def package_install(cls, options):
+    def package_install(cls, hosts='', package=''):
         """Install packages remotely."""
-        cls.command_sub = 'package install'
-        return cls.execute(cls._construct_command(options), output_format='csv')
+
+        return cls.execute(
+            f'job-invocation create --feature katello_package_install'
+            f' --search-query "{hosts}" --inputs "package={package}"'
+        )
 
     @classmethod
     def package_list(cls, options):
--- a/tests/foreman/cli/test_reporttemplates.py
+++ b/tests/foreman/cli/test_reporttemplates.py
@@ -908,12 +908,17 @@ def test_positive_generate_hostpkgcompare(
             hosts_info.append(Host.info({'name': client.hostname}))
             client.enable_repo(REPOS['rhst7']['id'])
             client.install_katello_agent()
+            # Add ReX key
+            client.add_rex_key(satellite=default_sat)
         hosts_info.sort(key=lambda host: host['name'])
 
         host1, host2 = hosts_info
-        Host.package_install({'host-id': host1['id'], 'packages': FAKE_0_CUSTOM_PACKAGE_NAME})
-        Host.package_install({'host-id': host1['id'], 'packages': FAKE_1_CUSTOM_PACKAGE})
-        Host.package_install({'host-id': host2['id'], 'packages': FAKE_2_CUSTOM_PACKAGE})
+        Host.package_install(hosts=f"id={host1['id']}", package=FAKE_0_CUSTOM_PACKAGE_NAME)
+        Host.package_install(hosts=f"id={host1['id']}", package=FAKE_1_CUSTOM_PACKAGE)
+        Host.package_install(hosts=f"id={host2['id']}", package=FAKE_2_CUSTOM_PACKAGE)
+        # Host.package_install({'host-id': host1['id'], 'packages': FAKE_0_CUSTOM_PACKAGE_NAME})
+        # Host.package_install({'host-id': host1['id'], 'packages': FAKE_1_CUSTOM_PACKAGE})
+        # Host.package_install({'host-id': host2['id'], 'packages': FAKE_2_CUSTOM_PACKAGE})
 
         result = ReportTemplate.generate(
             {
--- a/tests/foreman/cli/test_reporttemplates.py
+++ b/tests/foreman/cli/test_reporttemplates.py
@@ -916,9 +916,6 @@ def test_positive_generate_hostpkgcompare(
         Host.package_install(hosts=f"id={host1['id']}", package=FAKE_0_CUSTOM_PACKAGE_NAME)
         Host.package_install(hosts=f"id={host1['id']}", package=FAKE_1_CUSTOM_PACKAGE)
         Host.package_install(hosts=f"id={host2['id']}", package=FAKE_2_CUSTOM_PACKAGE)
-        # Host.package_install({'host-id': host1['id'], 'packages': FAKE_0_CUSTOM_PACKAGE_NAME})
-        # Host.package_install({'host-id': host1['id'], 'packages': FAKE_1_CUSTOM_PACKAGE})
-        # Host.package_install({'host-id': host2['id'], 'packages': FAKE_2_CUSTOM_PACKAGE})
 
         result = ReportTemplate.generate(
             {


robottelo/cli/host.py
```diff
@@ -126,10 +126,13 @@ def facts(cls, options=None):
...
-    def package_install(cls, options):
+    def package_install(cls, hosts='', package=''):
         """Install packages remotely."""
-        cls.command_sub = 'package install'
-        return cls.execute(cls._construct_command(options), output_format='csv')
+
+        return cls.execute(
+            f'job-invocation create --feature katello_package_install'
+            f' --search-query "{hosts}" --inputs "package={package}"'
+        )
```
Gauravtalreja1, line 135: @lhellebr Issue linked is expected after 6.10, but `Host` CLI's still uses katello-agent, and it just exists to keep warning users to use rex instead of katello-agent, so it doesn't make sense to update Host CLI's for achieving this.
Please consider updating the test instead to use `JobInvocation` CLI options to package_install, check this test for reference  https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/cli/test_host.py#L1912

Gauravtalreja1 (changes requested):

robottelo/cli/host.py
swadeley: That file implements hammer host commands, so you would not change that unless the command was removed  or changed completely. You could update the description if you like.

`~]# hammer host package install --host-id 1 --packages vim
This command uses katello agent and will be removed in favor of remote execution in a future release.
The remote execution equivalent is `hammer job-invocation create --feature katello_package_install`.`

tstrych (changes requested):
I guess your test is failing with 
```
The remote execution equivalent is hammer job-invocation create --feature katello_package_install
Could not install packages:\n  This action uses katello-agent, which is currently disabled. Use remote execution instead.
```

AFAIK the katello agent usage will be deprecated in 7.1 
KA is disabled by default, but can be enabled in sat installer 

I would prefer one of these options:
1. use REX and therefore  jobinvocation,  here is another example which is installing multiple packages from `cli/test_remoteexecution.py`
https://github.com/SatelliteQE/robottelo/blob/ebde94b8755651af341ef760b1590bb4024391a9/tests/foreman/cli/test_remoteexecution.py#L378-L398
2. use old approach after the ka is enabled
`satellite-installer --foreman-proxy-content-enable-katello-agent`

lhellebr:
I understand your concerns. My reason to do this is to encapsulate the package installation, it's not nice to do all this stuff manually wherever I need to install a package on a host.

Tomas's example 1 surely works but it only makes sense to do in a test that specifically tests ReX, I don't think we should do something similar in every test that just needs to install some random package, for example the test in this PR.

Enabling katello-agent is surely possible, but:
- Since it would be awful for a test to run `satellite-installer`, this would need to be specified during templatization
- It seems natural to me to use the same approach that we recommend to the users.

lhellebr:
Neither my solution or your solutions seem ideal to me, can we come up with some other ideas?

JacobCallahan:
@lhellebr it sounds like there are two distinct answer here.

1. When installing packages on a host via Satellite, that action should be done via REX
2. When installing packages on a host, while not testing satellite, that action should be done on the host via ssh connection.

robottelo/cli/host.py
jyejare: Ping @lhellebr ?

lhellebr:
> @lhellebr it sounds like there are two distinct answer here.
> 
>     1. When installing packages on a host via Satellite, that action should be done via REX
> 
>     2. When installing packages on a host, while not testing satellite, that action should be done on the host via ssh connection.

Yes, those cases can be handled differently. But for this PR, we see `package_install` needs to be fixed since we shouldn't use `katello-agent` anymore. The suggested solution is to use ReX. Is there any better way you can think about? Would you rather implement it by SSHing into the machine and running yum directly? Would you drop the function completely?

robottelo/cli/host.py
lhellebr: @jyejare Pong, I continued the discussion in followup comments, is there something specific in this thread you need a reply to?

swadeley: @lhellebr Please reverses the change to the `def package_install` function. That function implements a hammer command that still exists (someone else might need it).  There are other ways of installing packages depending on where you just need the package to be there, e.g. a setup step (search for run and execute examples which use SSH), or if you want to use Satellite to install it as part of the test (see https://github.com/SatelliteQE/robottelo/pull/9220/files#r763746302)

tests/foreman/cli/test_reporttemplates.py
```diff
@@ -908,12 +908,14 @@ def test_positive_generate_hostpkgcompare(
...
+            client.add_rex_key(satellite=default_sat)
         hosts_info.sort(key=lambda host: host['name'])
 
         host1, host2 = hosts_info
-        Host.package_install({'host-id': host1['id'], 'packages': FAKE_0_CUSTOM_PACKAGE_NAME})
-        Host.package_install({'host-id': host1['id'], 'packages': FAKE_1_CUSTOM_PACKAGE})
-        Host.package_install({'host-id': host2['id'], 'packages': FAKE_2_CUSTOM_PACKAGE})
+        Host.package_install(hosts=f"id={host1['id']}", package=FAKE_0_CUSTOM_PACKAGE_NAME)
+        Host.package_install(hosts=f"id={host1['id']}", package=FAKE_1_CUSTOM_PACKAGE)
+        Host.package_install(hosts=f"id={host2['id']}", package=FAKE_2_CUSTOM_PACKAGE)
```
JacobCallahan, line 918: you could avoid Host.package_install altogether 
```suggestion
        res = host1.execute(f'yum -y install {FAKE_0_CUSTOM_PACKAGE_NAME} {FAKE_1_CUSTOM_PACKAGE}')
        assert not res.status
        res = host2.execute(f'yum -y install {FAKE_2_CUSTOM_PACKAGE}')
        assert not res.status
```

robottelo/cli/host.py
jyejare: @lhellebr Ahh I thought the conversation in unresolved, so i just pinged you here!

tests/foreman/cli/test_reporttemplates.py
lhellebr: Of course I can but for the sake of encapsulating this to a simple call, I think it's better to have a function for this than executing a call directly.

robottelo/cli/host.py
lhellebr: @swadeley `someone else might need it` - the function fails by default though! However, I understand it's supposed to implement a certain hammer command, that's why I'm still not sure what is the cleanest solution here.

lhellebr:
I think I'll leave this PR open for discussion and meanwhile fix the failing test by executing the commands directly. However, the issue of `package_install` failing by default will not be fixed by that and other tests will probably fail as a result.

tests/foreman/cli/test_reporttemplates.py
JacobCallahan: Do you need to test that Satellite can install the packages when called through hammer? Or do you just need the packages installed?
If it is the latter, then directly installing is much simpler than all the layers that come with the old method.

JacobCallahan:
@lhellebr in your opinion, does this need to be in before 6.10 branching?

JacobCallahan:
@lhellebr if this is being blocked by the manifest issues you're having, can you work with @rplevka to resolve those issues?

lhellebr:
Yes, this is blocked by that issue. I'm trying to solve it.

I don't think this needs to be in 6.10 before branching and I'll probably close this PR once I create the one that doesn't change `package_install` - at which point my test will be fixed but the roots issue (the fact that `package_install` doesn't work by default anymore) will still be there.

lhellebr:
Workaround for my specific test which does *not* fix the root cause: https://github.com/SatelliteQE/robottelo/pull/9250
I am closing this PR since it's not a good solution. I don't have any better though.

PR #9227, opened 2021-12-09
test: Publishing a new Content View version brings in old metadata files from multiple previous versions; regenerating CV metadata then fixes it

sambible:
A manual test case for https://bugzilla.redhat.com/show_bug.cgi?id=1903367.

This one has a ton of customer cases attached to it, and would be very difficult to automate given the need for a seperate web server to set as an upstream url. Manual case seems good here. Going off @synkd 's manual verification steps.

### DIFF
--- a/tests/foreman/ui/test_repository.py
+++ b/tests/foreman/ui/test_repository.py
@@ -1088,3 +1088,39 @@ def test_positive_select_org_in_any_context():
     :CaseImportance: High
     """
     pass
+
+
+@pytest.mark.stubbed
+@pytest.mark.tier2
+def test_negative_duplicate_metadata_when_publishing_content_view():
+    """When publishing a Content View synced to a remote upstream source that has a new Repo created
+    since the last sync, there should not be muliple copies of metadata (primary.xml.gz) in the view
+
+    :id: 9cee2706-fa51-47a3-ba1d-a35ca66e7cb6
+
+    :customerscenario: true
+
+    :BZ:1903367
+
+    :Steps:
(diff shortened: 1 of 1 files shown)


synkd (commented):
A manual test is definitely an option here. Thinking back to when I verified this BZ, though, I'm not entirely sure why I used a separate Satellite as a web server for hosting the repo. I wonder if this could work if we created the repo in `/var/www/html/pub` on the Satellite being tested (let's call it satellite.example.com) and entered satellite.example.com/pub/repo as the URL for the custom repo. 

I have not tested this setup, but I would be interested to hear what others think about this idea.

sambible:
I'll give testing it a shot, would make the manual case much easier to run

tests/foreman/ui/test_repository.py
```diff
@@ -1088,3 +1088,39 @@ def test_positive_select_org_in_any_context():
...
+@pytest.mark.stubbed
+@pytest.mark.tier2
+def test_negative_duplicate_metadata_when_publishing_content_view():
+    """When publishing a Content View synced to a remote upstream source that has a new Repo created
+    since the last sync, there should not be muliple copies of metadata (primary.xml.gz) in the view
+
+    :id: 9cee2706-fa51-47a3-ba1d-a35ca66e7cb6
+
+    :customerscenario: true
+
```
ColeHiggins2, line 1102: Lets add a :CaseAutomation: NotAutomatted tag to the docstring

sambible:
after some investigation, and chat with @synkd this looks like a pulp2 bug, so not worth making a case for. Closing PR

Issue #8852, opened 2021-08-18
methods should be integrated to ContentHost, Satellite class


tstrych:
- register_to_dogfood
- install_cockpit_plugin

based on https://github.com/SatelliteQE/robottelo/pull/8812#pullrequestreview-731022072

PR #9234, opened 2021-12-15
encapsulate the cockpit sat installer under the Sat class

tstrych:
addressing #8852 
where the install_cockpit_plugin should be moved under sat class. 

the register to dogfood was already moved.

I will rather have fixture in the test than two setup lines in this case:
```
deafult_sat.register_to_dogfood()
default_sat.install_cockpit_plugin()
```

but `satellite-installer` command in `pytest/sys` still feels weird but I don't know if there is a better place let me know if you think so :smile:

### DIFF
--- a/pytest_fixtures/core/sys.py
+++ b/pytest_fixtures/core/sys.py
@@ -47,19 +47,7 @@ def proxy_port_range(default_sat):
 @pytest.fixture
 def install_cockpit_plugin(default_sat):
     default_sat.register_to_dogfood()
-    cmd_result = default_sat.execute(
-        'foreman-installer --enable-foreman-plugin-remote-execution-cockpit', timeout='30m'
-    )
-    if cmd_result.status != 0:
-        raise SatelliteHostError(
-            f'Error during cockpit installation, installation output: {cmd_result.stdout}'
-        )
-    cmd_result = default_sat.execute(
-        f'sshpass -p "{settings.server.ssh_password}" ssh-copy-id \
-        -i ~foreman-proxy/.ssh/id_rsa_foreman_proxy -o StrictHostKeyChecking=no localhost'
-    )
-    if cmd_result.status != 0:
-        raise SatelliteHostError(f'Error during ssh-copy-id, command output: {cmd_result.stdout}')
+    default_sat.install_cockpit()
 
 
 @pytest.fixture(scope='session')
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -1374,3 +1374,13 @@ class Satellite(Capsule):
             raise SatelliteHostError(
                 f'Error during registration, command output: {cmd_result.stdout}'
             )
+
+    def install_cockpit(self):
+        cmd_result = self.execute(
+            'satellite-installer --enable-foreman-plugin-remote-execution-cockpit', timeout='30m'
+        )
+        if cmd_result.status != 0:
+            raise SatelliteHostError(
+                f'Error during cockpit installation, installation output: {cmd_result.stdout}'
+            )
+        self.add_rex_key(self)


tstrych:
```========== 1 passed, 35 deselected, 3 warnings in 1069.43s (0:17:49) ===========```

**Review by JacobCallahan — APPROVED** [review]
Very nice!

File: pytest_fixtures/core/sys.py [hunk 1]
```diff
@@ -47,19 +47,7 @@ def proxy_port_range(default_sat):
...
-        raise SatelliteHostError(
-            f'Error during cockpit installation, installation output: {cmd_result.stdout}'
-        )
-    cmd_result = default_sat.execute(
-        f'sshpass -p "{settings.server.ssh_password}" ssh-copy-id \
-        -i ~foreman-proxy/.ssh/id_rsa_foreman_proxy -o StrictHostKeyChecking=no localhost'
-    )
-    if cmd_result.status != 0:
-        raise SatelliteHostError(f'Error during ssh-copy-id, command output: {cmd_result.stdout}')
+    default_sat.install_cockpit()
```
Gauravtalreja1, line 50: `test_positive_cockpit ` should be marked as a destructive test or can be run on separate sat, I don't think it's wise to run the installer on default_sat

robottelo/hosts.py
```diff
@@ -1374,3 +1374,20 @@ def register_to_dogfood(self):
             raise SatelliteHostError(
                 f'Error during registration, command output: {cmd_result.stdout}'
             )
+
+    def install_cockpit(self):
+        cmd_result = self.execute(
+            'foreman-installer --enable-foreman-plugin-remote-execution-cockpit', timeout='30m'
```
Gauravtalreja1, line 1380:
```suggestion
            'satellite-installer --enable-foreman-plugin-remote-execution-cockpit', timeout='30m'
```

tstrych: it doesn't matter it is just a symlink

pytest_fixtures/core/sys.py
tstrych: the question here, is if we are marking the fixtures or we are marking the tests
agree with you in this case all tests with using mentioned fixture should be destructive, so marking fixture seems correct to me

right now, no pytest fixture has destructive marker on it
right now, there is only one test using install cockpit, this test is marked as destructive

robottelo/hosts.py
swadeley: In this context it does not matter much as you are not testing the installer itself.

robottelo]$ grep -r satellite-installer tests/ | wc -l
36
robottelo]$ grep -r foreman-installer tests/ | wc -l
12

robottelo/hosts.py
```diff
@@ -1374,3 +1374,20 @@ def register_to_dogfood(self):
...
+        )
+        if cmd_result.status != 0:
+            raise SatelliteHostError(
+                f'Error during cockpit installation, installation output: {cmd_result.stdout}'
+            )
+        cmd_result = self.execute(
+            f'sshpass -p "{settings.server.ssh_password}" ssh-copy-id \
+            -i ~foreman-proxy/.ssh/id_rsa_foreman_proxy -o StrictHostKeyChecking=no localhost'
```
swadeley, line 1388: IIRC this is key used for REX? I wonder if you could use:
`robottelo]$ grep -r add_rex_key robottelo/
robottelo/hosts.py:    def add_rex_key(self, satellite, key_path=None)`

pytest_fixtures/core/sys.py
JacobCallahan: Interesting, I haven't considered that. Is installing cockpit really a destructive action?

tstrych: no it isn't, is just to make sure, that nothing else run while the installer run, while installer run nothing else should. 
I know we used the marker called run in one thread for that, but not sure how that works now with dynaconf.

robottelo/hosts.py
tstrych: will check this one

robottelo/hosts.py
tstrych: we can rename all of them, but i think it's not important, IMO it will only slow us in future development, because there will one more thing to check during the pr's, but sure we can create an issue and rename all of them later

jyejare: Regardless of everything `satellite-installer` is more accurate here and as a downstream testing. The same is suggested in documentation https://access.redhat.com/documentation/en-us/red_hat_satellite/6.10/html/managing_hosts/host_management_and_monitoring_using_red_hat_web_console#integrating_satellite_with_red_hat_web_console .

So use `satellite-installer`

pytest_fixtures/core/sys.py
jyejare: @tstrych Since the introduction of satellites as xDist workers, we will have one satellite running one test at a time so we dont need that isolation marker at all! 

Though enabling the cockpit plugin on satellite won't affect other tests/components functionality, but IMHO, the tests that alter the default_sat behavior should be treated special cases, in this case, `destructive_sat` similar to katello-agent tests or `destructive` marker to isolate the test is required !

jyejare (approved):
ACK pending comments !

robottelo/hosts.py
jyejare: @swadeley @tstrych This could be a non-blocking comment though as he just moved the existing code to new place and later he might raise another PR as improvement !

pytest_fixtures/core/sys.py
tstrych: @jyejare but altered behavior should not cause any changes, and if it does it is bug, so we should test with altered behavior, understand that such a bug would be hard to find, and it will impact part or whole automation run, but still think it's worth, we are not seeking ever green numbers, we want it red, so we can hunt the problems down and get back to green and again to red colors :)

JacobCallahan: I am not currently under the belief that cockpit installation would be destructive. Altering a Satellite installation in a way that doesn't negatively impact other components shouldn't be considered destructive.

JacobCallahan:
@tstrych can you update this, please? I'd like to include this before branching off 6.10

pytest_fixtures/core/sys.py
jyejare: @tstrych Yeah So in my last comment I am talking abt that `tests altering the behavior on satellite` should be isolated the way we are doing in `katello-agent` tests or we should find all such tests and make them use the `destructive_sat` fixture so it wont affect anything in any case.

tstrych: and what I am telling you is that your approach, to minimize test result error is great for automation results but it is not for product testing.
so as Jake said while installer actions that doesn't negatively impact satellite should be treated as non destructive

in my  perspective our aim is to get bulletproof product not evergreen automation :) 
maybe I missed your point, and I would be happy to discuss this with you

tstrych:
thanks for waiting on this :) 
```
tests/foreman/ui/test_host.py::test_positive_cockpit
========================================================================= 1 passed, 3 warnings in 292.94s (0:04:52) ==========================================================================```

JacobCallahan:
Merging to make it in to the 6.10 branch. Any remaining issues can be fixed after.

PR #9235, opened 2021-12-15
Replace attrdict with box

sambible:
A simple replacement, ran the tests where I could, everything seemed fine.


JacobCallahan (approved):

ColeHiggins2 (approved):
ACK

PR #9239, opened 2021-12-16
Disabling check for errata_applicability task

swadeley:
Hi

Adding if not is_open('BZ:2017533') section to skip
     call to check for task completion as its ID is not returned.

Due to:

[BZ#2017533 – API call to generate applicability does not return task info](https://bugzilla.redhat.com/show_bug.cgi?id=2017533)

Also added module_org to a search.

Please do not squash commits.

### DIFF
--- a/tests/foreman/longrun/test_inc_updates.py
+++ b/tests/foreman/longrun/test_inc_updates.py
@@ -30,6 +30,7 @@ from robottelo.constants import FAKE_4_CUSTOM_PACKAGE
 from robottelo.constants import PRDS
 from robottelo.constants import REPOS
 from robottelo.constants import REPOSET
+from robottelo.utils.issue_handlers import is_open
 
 pytestmark = [pytest.mark.run_in_one_thread]
 
@@ -149,8 +150,13 @@ def host(
     rhel7_contenthost_module.execute(f'yum install -y {FAKE_4_CUSTOM_PACKAGE}')
     rhel7_contenthost_module.execute('katello-package-upload')
     host = entities.Host().search(query={'search': f'name={rhel7_contenthost_module.hostname}'})
-    # Force host to generate or refresh errata applicability
-    call_entity_method_with_timeout(host[0].errata_applicability, timeout=600)
+    # Disabling call to timout due to BZ#2017533, but if we get no failures, we can remove it.
+    if not is_open('BZ:2017533'):
+        # Force host to generate or refresh errata applicability
+        call_entity_method_with_timeout(host[0].errata_applicability, timeout=120)
+    else:
+        host[0].errata_applicability(synchronous=False)
+
     # Add filter of type include but do not include anything.
     # this will hide all RPMs from selected erratum before publishing.
     entities.RPMContentViewFilter(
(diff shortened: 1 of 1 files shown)


swadeley:
trigger: test-robottelo
pytest: tests/foreman/longrun/test_inc_updates.py::test_positive_noapply_api

swadeley:
trigger: test-robottelo
pytest: tests/foreman/longrun/test_inc_updates.py::test_positive_noapply_api

tests/foreman/longrun/test_inc_updates.py
```diff
@@ -147,8 +150,13 @@ def host(
...
     rhel7_contenthost_module.execute('katello-package-upload')
     host = entities.Host().search(query={'search': f'name={rhel7_contenthost_module.hostname}'})
-    # Force host to generate or refresh errata applicability
-    call_entity_method_with_timeout(host[0].errata_applicability, timeout=600)
+    # Disabling call to timout due to BZ#2017533, but if we get no failures, we can remove it.
+    if not is_open('BZ:2017533'):
+        # Force host to generate or refresh errata applicability
+        call_entity_method_with_timeout(host[0].errata_applicability, timeout=120)
+    else:
+        host[0].errata_applicability(synchronous=False)
```
jyejare, line 158: Won't the async applicability effect in the later stage of this test?

JacobCallahan: can you elaborate on this?

JacobCallahan: @jyejare please? :D

jyejare: We are trying to async the `errata applicability` step which might run in the BG while we move to the next steps and if it didn't complete before the next step/steps where the next step might need errata applicability to complete then we will end up with failing step.

This could affect next step / next test as well if its not sync, I mean we should wait to finish the errata applicability in any case !

JacobCallahan: that's a very good point. we could also set synchronous to True.

JacobCallahan:
This missed the deadline for 6.10 branching. If the change still needs to be brought back, we can have that discussion.

tests/foreman/longrun/test_inc_updates.py
swadeley: Hi, so it seems to me we need to implement a "wait for tasks" type procedure within Nailgun itself. i.e. use API to search for repo publish task, using the name of repo being published not ID of task which we no longer have.

swadeley: Hi,  it seems to me we need to implement a "wait for tasks" type procedure within Nailgun itself. i.e. use API to search for repo publish task, using the name of repo being published not ID of task which we no longer have. 

But please note there is no more `Actions::Katello::Host::GenerateApplicability` task to search for, so for this `errata applicability` case I don't see a way to check for completion of that errata enumeration task, and we don't want to use sleep. I will ask Devs for more ideas, but  I believe the errata applicability is now faster as its not a foreman task.

pondrejk (approved):

swadeley:
Hello

Closing in favour of:

 Change applicability task check #9323 


Thank you

PR #9243, opened 2021-12-16
Add :CaseComponent: to stubbed Ansible tests

synkd:
While reviewing my Tier 1 and 2 component test coverage, I realized that my stubbed Ansible tests did not have a `:CaseComponent:` assigned at the test level and were thus inheriting the `RemoteExecution` component from the test module. This PR adds the `Ansible` component to these stubbed tests.

### DIFF
--- a/tests/foreman/ui/test_remoteexecution.py
+++ b/tests/foreman/ui/test_remoteexecution.py
@@ -284,6 +284,8 @@ def test_positive_ansible_job_check_mode(session):
 
     :CaseAutomation: NotAutomated
 
+    :CaseComponent: Ansible
+
     :assignee: dsynk
     """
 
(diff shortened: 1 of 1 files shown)


sambible:
Looks good to me

jyejare (approved):
ACK with improvement suggestion!

Suggestion:
- Move all the Ansible Tests in a separate class in this same REX test module.
- Set `CaseComponent` and `assignee` or any other common token at the class level and those tokens then will be applied to all the test cases in the class.

This will help in replicating the same tokens over all the tests !

swadeley:
> ACK with improvement suggestion!
> 
> Suggestion:
> 
>     * Move all the Ansible Tests in a separate class in this same REX test module.
> 
I thought we were supposed to stop using classes when we changed our tests from unittest to pytest?

swadeley (approved):
ACK

mshriver:
@swadeley We're discouraging the use of classes when they don't actually provide anything - which in this case with @jyejare suggestion the class would be providing common metadata.  We had a bunch of classes that (with pytest fixture conversions) no longer provided any useful grouping of tests, which we discouraged and largely removed.

mshriver (approved):

PR #9244, opened 2021-12-17
Pulp Task Clean-up stubs

Griffin-Sullivan:
Adding test stubs for the Pulp Task Clean-up feature.

### DIFF
--- a/tests/foreman/api/test_repository.py
+++ b/tests/foreman/api/test_repository.py
@@ -1387,6 +1387,46 @@ class TestRepositorySync:
         """
         pass
 
+    @pytest.mark.stubbed
+    @pytest.mark.tier1
+    def test_positive_bulk_cancel_sync(self):
+        """Bulk cancel 10+ repository syncs
+
+        :id: f9bb1c95-d60f-4c93-b32e-09d58ebce80e
+
+        :steps:
+            1. Add 10+ repos and sync all of them
+            2. Cancel all of the syncs
+            3. Check /var/log/foreman/production.log and /var/log/messages
+
+        :expectedresults: All the syncs stop successfully.
+
+        :CaseImportance: High
+
+        :CaseAutomation: NotAutomated
+        """
+
+    @pytest.mark.stubbed
+    @pytest.mark.tier4
+    @pytest.mark.destructive
+    @pytest.mark.run_in_one_thread
+    def test_positive_reboot_recover_sync(self):
+        """Reboot during repo sync and resume the sync when the Satellite is online
+
+        :id: 4f746e28-444c-4688-b92b-778a6e58d614
+
+        :steps:
+            1. Add a repo and start sync
+            2. Reboot the Satellite while sync is running
+            3. Check Foreman Tasks
+
+        :expectedresults: Repo sync resumes / can resume and finishes successfully
+
+        :CaseImportance: High
+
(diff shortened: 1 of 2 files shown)


tests/foreman/api/test_repository.py
```diff
@@ -1387,6 +1387,46 @@ def test_positive_sync_rh_app_stream(self):
...
+
+        :CaseImportance: High
+
+        :CaseAutomation: NotAutomated
+        """
+        pass
+
+    @pytest.mark.stubbed
+    @pytest.mark.tier1
+    def test_positive_reboot_recover_sync(self):
```
jameerpathan111, line 1412: These tests doesn't seem like `tier1`, maybe they should be `tier3` or `tier4`.
Also, they seem like destructive tests so we should also add 
```
@pytest.mark.destructive
```

tests/foreman/api/test_repository.py
```diff
@@ -1387,6 +1387,46 @@ def test_positive_sync_rh_app_stream(self):
...
+            2. Reboot the Satellite while sync is running
+            3. Check Foreman Tasks
+
+        :expectedresults: Repo sync resumes / can resume and finishes successfully
+
+        :CaseImportance: High
+
+        :CaseAutomation: NotAutomated
+        """
+        pass
```
jameerpathan111, line 1428:
```suggestion
```
Is this required?

JacobCallahan: it isn't. there are also a number of other instances of this in the pr

tests/foreman/api/test_repository.py
JacobCallahan: Good suggestion. I'd recommend you reconsider the tiering proposal for most of these

tests/foreman/api/test_repository.py
Griffin-Sullivan: Oh ok. I saw that for some other stubs and thought it was required. Removing now.

tests/foreman/api/test_repository.py
Griffin-Sullivan: I changed most of them to `tier3` except for the one that just cancels repos syncing. However, I felt like `test_positive_reboot_recover_sync` was possibly `tier4` as a reboot and sync could take a long time. If this doesn't fit the criteria for `tier4` I can update it to `tier3`.

tests/foreman/api/test_repository.py
```diff
@@ -1387,6 +1387,48 @@ def test_positive_sync_rh_app_stream(self):
         """
         pass
 
+    @pytest.mark.stubbed
+    @pytest.mark.tier1
+    @pytest.mark.destructive
+    @pytest.mark.run_in_one_thread
```
JacobCallahan, line 1393:
```suggestion
```
You don't need to make this destructive since it isn't rebooting the system. The same is true for the other tests in this PR that don't involve reboots/manipulating services.

JacobCallahan (changes requested):

swadeley:
Hi @Griffin-Sullivan 

I don't think you need to do these test using CLI and UI, just using API should be fine. Reason is, you are testing how Satellite internal responds to unusual events, you are not testing CLI commands or UI elements.

JacobCallahan (approved):

PR #9250, opened 2021-12-21
package_install doesn't work by default anymore, using a direct execution approach

lhellebr:


JacobCallahan (approved):

tests/foreman/cli/test_reporttemplates.py
```diff
@@ -898,23 +898,27 @@ def test_positive_generate_hostpkgcompare(
...
             client.install_katello_ca(default_sat)
             # Register content host, install katello-agent
             client.register_contenthost(local_org['label'], local_ak['name'])
             assert client.subscribed
-            hosts_info.append(Host.info({'name': client.hostname}))
+            clients.append(client)
             client.enable_repo(REPOS['rhst7']['id'])
             client.install_katello_agent()
-        hosts_info.sort(key=lambda host: host['name'])
+        clients.sort(key=lambda client: client.hostname)
```
akhil-jha, line 912: Why the sort? I don't see where it affects the test..

akhil-jha (approved):
Non blocking comment. ACK

tests/foreman/cli/test_reporttemplates.py
```diff
@@ -898,23 +898,27 @@ def test_positive_generate_hostpkgcompare(
...
+        hosts_info = [Host.info({'name': client.hostname}) for client in clients]
 
         host1, host2 = hosts_info
-        Host.package_install({'host-id': host1['id'], 'packages': FAKE_0_CUSTOM_PACKAGE_NAME})
-        Host.package_install({'host-id': host1['id'], 'packages': FAKE_1_CUSTOM_PACKAGE})
-        Host.package_install({'host-id': host2['id'], 'packages': FAKE_2_CUSTOM_PACKAGE})
+        res = clients[0].execute(
+            f'yum -y install {FAKE_0_CUSTOM_PACKAGE_NAME} {FAKE_1_CUSTOM_PACKAGE}'
+        )
```
jyejare, line 918: How if you implement `package_install` helper in Host class instead of direct execution ? The package install will take list of packages to install!

This can be a non-blocking comment though !

jyejare (approved):
ACK with doable improvement suggestion !

tests/foreman/cli/test_reporttemplates.py
lhellebr: I tried to do that and got quite a lot of disagreement: https://github.com/SatelliteQE/robottelo/pull/9220

tests/foreman/cli/test_reporttemplates.py
lhellebr: So we get the same order as in the report. Otherwise, this can fail e.g. on line 934.

lhellebr:
Responded to comments, I think this can be merged.

PR #9252, opened 2021-12-22
Updated the satellite repository name from satellite6_repo to satellite_repo

devendra104:
Satellite repository name was very version specific and it was ambiguous for the satellite-7 repo so we have changed it from `satellite6_repo` to `satellite_repo`


devendra104:
There is some issue with this PR so closing it and will open the new one.

PR #9253, opened 2021-12-22
Register to dogfood before running the installation

lhellebr:
After further investigation, I still insist that the system MUST be s-m registered to prod. Registering it to Dogfood (or otherwise manually adding the repos) allows for receptor installation but the receptor can't be set up: part of receptor setup is running `insights-client --register` and that only passes on a s-m registered host.

Issue demonstrated by changes in this PR - receptor installation passes but the setup still fails.

Any suggestions other than actually registering the Satellite, as proper per documentation?


JacobCallahan:
@lhellebr can you detail the further investigation that led you to this conclusion?
Also, this change registers to dogfood, not "prod".

lhellebr:
Exactly, this is why this change doesn't actually fix the issue. This is what led me to the conclusion:
```
# insights-client --register
Unable to fetch egg url. Defaulting to /release
Please ensure that the system is registered with RHSM for CERT auth, or that correct credentials are set in /etc/insights-client/insights-client.conf for BASIC auth.
Could not connect to the Insights API. Run insights-client --test-connection for more information.
```
It got fixed once I properly s-m registered the Satellite to prod.

JacobCallahan:
@lhellebr correct. For these tests, the Satellite will need to be registered to RH.
However, the case remains that there are no strict requirements I know of for the rest of our systems to be registered to RH.

lhellebr:
In the second commit, I made the changes necessary to not just enable a repo but actually s-m register to prod, have the repos enabled and insights-client to successfully register. I still think the Satellite should be registered the whole time during automation run as per documentation and that it's not nice for a test to run s-m register/unregister but since the fix is probably not happening anytime soon, I propose this to be merged as it fixes the test.

swadeley:
Hi

Can you not just use existing fixture here https://github.com/SatelliteQE/robottelo/blob/22ab55c1381cbc6369982c4c94fee02a2548b1b6/pytest_fixtures/component/subscription.py#L21

Some ideas there on naming conventions for fixtures.

lhellebr:
I can and I should :)
I just wasn't able to find this fixture as I was grepping for `register`. I knew something like this must exist!

Uploaded a commit using that fixture. Testing is blocked, however, because the login credentials for s-m registration stopped working. Let's wait until that is resolved.

lhellebr:
Fixed to enable additional repos after registration (the fixture doesn't auto-attach). Surprisingly, there was no constant for the satellite repo so I've added it. Test results:
```
============================================== warnings summary ===============================================
tests/foreman/cli/test_remoteexecution.py: 132 warnings
  /home/lhellebr/broker/venv/lib/python3.8/site-packages/urllib3/connectionpool.py:981: InsecureRequestWarning: Unverified HTTPS request is being made to host 'dhcp-3-1.vms.sat.rdu2.redhat.com'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/warnings.html
================================= 1 passed, 132 warnings in 425.96s (0:07:05) =================================
```

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -488,7 +497,9 @@ def test_positive_run_scheduled_job_template_by_ip(self, fixture_vmsetup, defaul
 
     @pytest.mark.tier3
     @pytest.mark.upgrade
-    def test_positive_run_receptor_installer(self, default_sat):
+    def test_positive_run_receptor_installer(
+        self, default_sat, subscribe_satellite, fixture_enable_receptor_repos
```
jyejare, line 501: subscribe_satellite is a `session-scoped` fixture and once it's used in the test it won't tear down for that test, instead, it will wait for the whole pytest session to complete. Do we ever consider the consequences of keeping it registered for the entire session that's initiated somewhere in the middle of a session from a test (like repos for the latest in-dev sat won't be available that dogfood does provide. There could be more which I cant calculate right now)? @SatelliteQE/robottelo-tier-1-reviewers  and @SatelliteQE/robottelo-tier-2-reviewers could jump in !

I would either turn that fixture into function scoped or create a new function scoped fixture to be used here! 

Thoughts @JacobCallahan @mshriver?

jyejare (commented):
ACK pending question!

tests/foreman/cli/test_remoteexecution.py
lhellebr: I've discussed this for quite some time in chat because I don't like this approach either. It seems like using this fixture to do this is a commonly acceptable ground (after all, that's why it is there).
I still think, however, that the Satellite SHOULD be registered the whole time the automation runs against it.

If others think this fixture should be remade to be function-scoped and a teardown of `s-m unregister` should be added, I'm not against it but I'm not sure it's within scope of this PR - it could affect other tests (keeping it as it is has that potential, too, though).

JacobCallahan: #9281 intends to solve the fixture issue

lhellebr:
Are there any unresolved concerns? I think this can be merged.

JacobCallahan:
@lhellebr can you retest (now that the fixture has changed) to make sure everything is still good?

lhellebr:
@JacobCallahan There seems to be an issue with that change: https://github.com/SatelliteQE/robottelo/pull/9281#issuecomment-1014437662

JacobCallahan:
> @JacobCallahan There seems to be an issue with that change: [#9281 (comment)](https://github.com/SatelliteQE/robottelo/pull/9281#issuecomment-1014437662)

@lhellebr thanks for pointing that out! Mind trying again with #9310?

lhellebr:
Can we finally merge this?

lhellebr:
Test output is here, teardown failure is not related to this PR: https://github.com/SatelliteQE/robottelo/pull/9310#issuecomment-1017392704

jyejare (approved):

mshriver (approved):

PR #9257, opened 2021-12-22
Remove codecoverage analysis

JacobCallahan:
There is very little value for these reports in a project like robottelo
They detract from the overall effect of a PR's other checks
With that in mind, I'm removing code coverage checks with this PR until
further notice.


Griffin-Sullivan:
Well it looks like it's working!

jyejare (commented):
@JacobCallahan That's great step towards refinement :) Thanks for raising the PR. 

We need to cherrypick this for 6.10 and probably onboard everyone for this removal as we discussed in the meeting.

jyejare (approved):

pondrejk (approved):

PR #9263, opened 2021-12-27
Removing osp hostname from osp.yaml.template

akhil-jha:
$SUBJECT


jyejare (approved):

Gauravtalreja1:
@akhil-jha Thanks for removing this, we shouldn't keep such hostnames, can you remove the same from 6.9.z/6.10.z branches?

PR #9270, opened 2022-01-06
Log host name if subscription failed

swadeley:
Hello

In 6.10 hosts are not deleted when unregistered.

When you have multiple hosts registered by test automation, any attempt at duplicate registration cause a failure but the reason is not given. If we could log the host name we could check for duplicates.

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -35,6 +35,7 @@ from robottelo.constants import SATELLITE_VERSION
 from robottelo.helpers import get_data_file
 from robottelo.helpers import InstallerCommand
 from robottelo.helpers import validate_ssh_pub_key
+from robottelo.logging import logger
 
 
 POWER_OPERATIONS = {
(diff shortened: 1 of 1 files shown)


robottelo/hosts.py
```diff
@@ -780,7 +781,9 @@ def contenthost_setup(
         repo_labels = repo_labels or []
         self.install_katello_ca(satellite)
         self.register_contenthost(org_label, activation_key=activation_key, lce=lce)
+        result = self.register_contenthost(org_label, activation_key=activation_key, lce=lce)
```
pondrejk, line 784: I suppose this line should replace the one above...

pondrejk (commented):
ack, pending one comment

robottelo/hosts.py
swadeley: Yes!

swadeley: fixed, thank you

pondrejk (approved):

rplevka (approved):

swadeley:
Thank you.

PR #9285, opened 2022-01-11
Close loop bz

shweta83:
Automated BZ https://bugzilla.redhat.com/show_bug.cgi?id=1937009

### DIFF
--- a/tests/foreman/cli/test_location.py
+++ b/tests/foreman/cli/test_location.py
@@ -407,3 +407,27 @@ class TestLocation:
             Location.update({'id': location['id'], 'parent-id': location['id']})
         location = Location.info({'id': location['id']})
         assert location['parent'] == parent_location['name']
+
+    @pytest.mark.skip_if_open("BZ:1937009")
+    @pytest.mark.tier1
+    def test_positive_nested_location(self, request, module_org):
+        """View nested location in an organization
+
+        :id: cce4a33d-7743-47ef-b437-2539fdc73c44
+
+        :customerscenario: true
+
+        :BZ: 1937009
+
+        :expectedresults: Nested location can be viewed
+
+        :CaseImportance: Medium
+        """
+        parent_location = _location(request, {'organization-id': module_org.id})
+        location = _location(
+            request, {'parent-id': parent_location['id'], 'organization-id': module_org.id}
+        )
+        loc_list = Location.list(
+            {'search': f'title={location["title"]}', 'organization-id': module_org.id}
+        )
+        assert loc_list[0]['title'] == location['title']


tests/foreman/cli/test_location.py
```diff
@@ -407,3 +408,29 @@ def test_negative_update_parent_with_child(self, request):
...
+        :CaseImportance: Medium
+        """
+        parent_location = _location(request, {'organization-id': module_org.id})
+        location = _location(
+            request, {'parent-id': parent_location['id'], 'organization-id': module_org.id}
+        )
+        Org.add_location({'location-id': parent_location['id'], 'name': module_org.name})
+        Org.add_location({'location-id': location['id'], 'name': module_org.name})
+        loc_list = Location.list(
+            {'search': 'title=\\"%s\\"' % location['title'], 'organization-id': module_org.id}
```
Griffin-Sullivan, line 434:
```suggestion
            {'search': f'title={location["title"]}', 'organization-id': module_org.id}
```
All the logic in the test looks good. I just think the f string is easier to read.

shweta83: Yes

> All the logic in the test looks good. I just think the f string is easier to read.

Yes. It would be easier to read in f string.  Same is updated.

tests/foreman/cli/test_location.py
```diff
@@ -407,3 +408,29 @@ def test_negative_update_parent_with_child(self, request):
...
+        :expectedresults: Nested location can be viewed
+
+        :CaseImportance: Medium
+        """
+        parent_location = _location(request, {'organization-id': module_org.id})
+        location = _location(
+            request, {'parent-id': parent_location['id'], 'organization-id': module_org.id}
+        )
+        Org.add_location({'location-id': parent_location['id'], 'name': module_org.name})
+        Org.add_location({'location-id': location['id'], 'name': module_org.name})
```
pondrejk, line 432: If I'm reading the associated bugzilla right, then the child location is expected to inherit the org association form the parent. If that is true, the two lines above shouldn't be necessary.

pondrejk (commented):
nice job, ack pending one comment and a test result

shweta83:
trigger: test-robottelo
pytest: tests/foreman/cli/test_location.py -k test_positive_nested_location

tests/foreman/cli/test_location.py
shweta83: @pondrejk Unless we add the locations to the organization, it doesn't show in the organization which is associated to it.

pondrejk: Yes, but isn't that a result of the bug in 1937009? If we want to test the desired behavior mentioned in the bug, I think we shouldn't add the location to the org. The test shouldn't pass now, because the bz is in new state. That is all right because you use skip_if_open. Once the bug is fixed, the test will run and should pass...

shweta83: Yes, I overlooked it. Updated the same.

tests/foreman/cli/test_location.py
```diff
@@ -407,3 +407,27 @@ def test_negative_update_parent_with_child(self, request):
             Location.update({'id': location['id'], 'parent-id': location['id']})
         location = Location.info({'id': location['id']})
         assert location['parent'] == parent_location['name']
+
+    @pytest.mark.skip_if_open("BZ:1937009")
+    @pytest.mark.tier1
+    def test_positive_nested_location(self, request, module_org):
```
jyejare, line 413: Why would you need a request parameter here ? Remove it !

tests/foreman/cli/test_location.py
```diff
@@ -407,3 +407,27 @@ def test_negative_update_parent_with_child(self, request):
...
+        :CaseImportance: Medium
+        """
+        parent_location = _location(request, {'organization-id': module_org.id})
+        location = _location(
+            request, {'parent-id': parent_location['id'], 'organization-id': module_org.id}
+        )
+        loc_list = Location.list(
+            {'search': f'title={location["title"]}', 'organization-id': module_org.id}
+        )
+        assert loc_list[0]['title'] == location['title']
```
jyejare, line 433: Assuming `location` will be placed in 1st position in the location list could fail this test case if the list return preferences changed(now by creation time/id probably) so just safely check if the id/title of location in location list.

```suggestion
        assert location['title'] in [ loc.get('title') for loc in loc_list ]
```

jyejare (approved):
ACK with improvement comment!

jyejare:
trigger: test-robottelo
pytest: tests/foreman/cli/test_location.py -k test_positive_nested_location

tests/foreman/cli/test_location.py
shweta83: @jyejare This will always work because in the previous line it searches location with the id and returns that location only so loc_list will always have one element.

jyejare: Ahh I missed that line and confused with `loc_list` variable ! Anyways you have my ACK already !

Gauravtalreja1 (approved):

latran:
trigger: test-robottelo
pytest: tests/foreman/cli/test_location.py -k test_positive_nested_location

latran (approved):
ack

PR #9288, opened 2022-01-11
fixed test_positive_copy_subscription

ColeHiggins2:
tests.foreman.cli.test_activationkey.test_positive_copy_subscription was failing due indexing a large string value. Assertion now just checks copied activation key has the same subscription as "parent" key.


sambible (approved):
:+1:

latran (approved):
ACK

swadeley (approved):
ACK

PR #9290, opened 2022-01-11
Fix sat-sync test case failure

latran:
Previous test was using the incorrect id.  Fixing it to use the correct
cvv-id.

Test result:
```
pytest tests/foreman/cli/test_satellitesync.py::TestContentViewSync::test_positive_export_import_promoted_cv

================ 1 passed in 170.23s (0:02:50) =============
```


jyejare (approved):

swadeley (commented):
ACK

PR #9296, opened 2022-01-13
Update metadata_markers.py

Griffin-Sullivan:
The `None` parameter is unnecessary. It's being used above with the `next()` function.


jyejare (approved):

swadeley (approved):
ACK

PR #9306, opened 2022-01-17
Satellite Endpoints user properties

jyejare:
The endpoint user property of tests depends on the `item.location` to filter the tests in ibutsu/RP based on api/cli/ui/endtoend/installer/longrun/sys endpoints for a specific compoenent/user(other metadatas) etc.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -13,6 +13,7 @@ pytest_plugins = [
     'pytest_plugins.metadata_markers',
     'pytest_plugins.settings_skip',
     'pytest_plugins.rerun_rp.rerun_rp',
+    'pytest_plugins.fspath_plugins',
     # Fixtures
     'pytest_fixtures.core.broker',
     'pytest_fixtures.core.contenthosts',
--- a/pytest_plugins/fspath_plugins.py
+++ b/pytest_plugins/fspath_plugins.py
@@ -1,0 +1,16 @@
+# File System related Collection Modification/Addition to test cases
+import re
+
+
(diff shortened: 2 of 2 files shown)


JacobCallahan:
Can you elaborate more on what this change does?

jyejare:
@JacobCallahan This adds the user_property of endpoint(api/cli/ui/installer etc) to the test cases in `tests/foreman` directory that can be used to filter the tests of those endpoints in Ibutsu which is not that straight-fwd or probably possible in ibutsu !

mshriver:
I think I would apply a mark (automatically based on path like you have here) so that the existing hook in `pytest_plugins/metadata_markers` adds the user_property from the mark.

jyejare:
@mshriver, Yeah that is also doable but I thought that unnecessary marker addition won't help us in pytest collection via marks cause we already have tests scattered in folders(api/cli/ui etc). Also, the `metadata_markers` plugin adds markers based on testimony tokens, but endpoint property/marker is being added based on fspath hence `metadata_markers` wont be a right place to automate this.

Thoughts ?

jyejare:
Ping @mshriver !

jyejare:
Ping @JacobCallahan @SatelliteQE/robottelo-tier-2-reviewers !

jyejare:
Ping @SatelliteQE/robottelo-tier-1-reviewers !!

swadeley (approved):
ACK

PR #9281, opened 2022-01-07
Minor refactor of subscriptions fixtures

JacobCallahan:
This change simplifies and removes the reliance on run_command

PR #9310, opened 2022-01-17
Fix incorrect check in subscribe_satellite

JacobCallahan:
This previously checked the container object instead of stdout itself.

### DIFF
--- a/pytest_fixtures/component/subscription.py
+++ b/pytest_fixtures/component/subscription.py
@@ -13,6 +13,8 @@ def subscribe_satellite(clean_rhsm, default_sat):
     from robottelo.config import settings
 
     default_sat.register_contenthost(
+        org=None,
+        lce=None,
         username=settings.subscription.rhn_username,
         password=settings.subscription.rhn_password,
         releasever=f'{default_sat.os_version.major}Server',
@@ -25,5 +27,5 @@ def subscribe_satellite(clean_rhsm, default_sat):
         yield
     else:
         pytest.fail('Failed to attach system to pool. Aborting Test!.')
-    default_sat.unsubscribe()
+    default_sat.unregister()
     default_sat.remove_katello_ca()
--- a/pytest_fixtures/component/subscription.py
+++ b/pytest_fixtures/component/subscription.py
@@ -17,9 +17,8 @@ def subscribe_satellite(clean_rhsm, default_sat):
         password=settings.subscription.rhn_password,
         releasever=f'{default_sat.os_version.major}Server',
     )
-    has_success_msg = 'Successfully attached a subscription'
     result = default_sat.subscription_manager_attach_pool([settings.subscription.rhn_poolid])[0]
-    if has_success_msg in result:
+    if 'Successfully attached a subscription' in result.stdout:
         default_sat.enable_repo(
             f'rhel-{default_sat.os_version.major}-server-extras-rpms', force=True
         )
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -415,8 +415,8 @@ class ContentHost(Host):
         consumerid=None,
         force=True,
         releasever=None,
-        username=None,
-        password=None,
+        username=settings.server.admin_username,
+        password=settings.server.admin_password,
         auto_attach=False,
     ):
         """Registers content host on foreman server either by specifying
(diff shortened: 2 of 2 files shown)


lhellebr (changes requested):
Read the docstring of `register_contenthost` function, it's intended for registration to Satellite, not to CDN. It doesn't work even though `clean_rhsm` has been called before it. It tries to run the following:
```
# subscription-manager register --org Default_Organization --environment Library --username USERNAME --password PASSWORD --release 7Server --force
Registering to: subscription.rhsm.redhat.com:443/subscription
Error: Server does not support environments.
```
You see that's a command you would use for registration to the Satellite, not to CDN. The result is the following failure when running the test. Previously the same error happened, as shown in https://github.com/SatelliteQE/robottelo/pull/9281#issuecomment-1014443835 , it was just handled incorrectly but that's not the main flaw:
```
tests/foreman/cli/test_remoteexecution.py::TestRemoteExecution::test_positive_run_receptor_installer ERROR [100%]

================================================= ERRORS ==================================================
_______________ ERROR at setup of TestRemoteExecution.test_positive_run_receptor_installer ________________

clean_rhsm = None, default_sat = <robottelo.hosts.Satellite object at 0x7f0788122760>

    def subscribe_satellite(clean_rhsm, default_sat):
        """subscribe satellite to cdn"""
    
        default_sat.register_contenthost(
            username=settings.subscription.rhn_username,
            password=settings.subscription.rhn_password,
            releasever=f'{default_sat.os_version.major}Server',
        )
        result = default_sat.subscription_manager_attach_pool([settings.subscription.rhn_poolid])[0]
        if 'Successfully attached a subscription' in result.stdout:
            default_sat.enable_repo(
                f'rhel-{default_sat.os_version.major}-server-extras-rpms', force=True
            )
            yield
        else:
>           pytest.fail('Failed to attach system to pool. Aborting Test!.')
E           Failed: Failed to attach system to pool. Aborting Test!.

pytest_fixtures/component/subscription.py:27: Failed
```

mshriver:
@lhellebr Not discounting the accurracy of your statement about the function and its use in test cases, but I think its beyond the scope - what @JacobCallahan fixed here is an obvious bug where the stdout field of the ssh command result isn't be checked.

mshriver (approved):

JacobCallahan:
@lhellebr you're correct. that registration method is geared towards satellite registration. while cdn registration is similar, there are times when you can specify less information.
If you tell me what your ideal registration command would look like, I can adjust the existing implementation in this PR.
Admittedly this is a bit of scope creep, but I'd like you to have something useful!

lhellebr:
@JacobCallahan This worked before https://github.com/SatelliteQE/robottelo/pull/9281 so I think that PR would be a good inspiration!

JacobCallahan:
@lhellebr ok, i've updated and if this version is to your liking, then i think we've made progress

synkd (approved):
ACK

lhellebr (approved):
I can confirm that my test works with this fix, it successfully fails later at line 559 :)
I can imagine this change might not be backwards compatible in some cases where people relied on this function to use org, lce etc. even if not explicitly specified, in one command. After this is considered, ACK.

lhellebr:
Actually, after fixing test setup, it passed but I hit this error in teardown:
```
tests/foreman/cli/test_remoteexecution.py::TestRemoteExecution::test_positive_run_receptor_installer PASSED                              [100%]
tests/foreman/cli/test_remoteexecution.py::TestRemoteExecution::test_positive_run_receptor_installer ERROR                               [100%]

==================================================================== ERRORS ====================================================================
________________________________ ERROR at teardown of TestRemoteExecution.test_positive_run_receptor_installer _________________________________

clean_rhsm = None, default_sat = <robottelo.hosts.Satellite object at 0x7fea900c3c40>

    def subscribe_satellite(clean_rhsm, default_sat):
        """subscribe satellite to cdn"""
    
        default_sat.register_contenthost(
            org=None,
            lce=None,
            username=settings.subscription.rhn_username,
            password=settings.subscription.rhn_password,
            releasever=f'{default_sat.os_version.major}Server',
        )
        result = default_sat.subscription_manager_attach_pool([settings.subscription.rhn_poolid])[0]
        if 'Successfully attached a subscription' in result.stdout:
            default_sat.enable_repo(
                f'rhel-{default_sat.os_version.major}-server-extras-rpms', force=True
            )
            yield
        else:
            pytest.fail('Failed to attach system to pool. Aborting Test!.')
>       default_sat.unsubscribe()
E       AttributeError: 'Satellite' object has no attribute 'unsubscribe'

pytest_fixtures/component/subscription.py:30: AttributeError
```
Looking at https://github.com/SatelliteQE/robottelo/pull/9281/files , I see the method has been removed there but it's still used.

JacobCallahan:
@lhellebr another great catch on the unsubscribe vs unregister. The other tests using this shouldn't be affected since the only functional change for existing uses is that it no longer raises the error if you don't pass what was previously required.

jyejare:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k test_positive_run_receptor_installer

jyejare:
@JacobCallahan Ping, the checks are failing in PRT !!

JacobCallahan:
@jyejare that failure doesn't look related to my change.

jyejare:
@JacobCallahan Ok then I ll let this get merged!

PR #9324, opened 2022-01-21
This commit introduces some of the patterns that will be primarily used

JacobCallahan:
as we move forward with a larger helpers refactor.
Included in this change are the introduction of mixins for our host
objects and a replacement for the cli factory methods.
The goal of this is to consolidate helper functionality under the host
objects, to reduce conflicts between test systems.

### DIFF
--- a/robottelo/host_helpers/cli_factory.py
+++ b/robottelo/host_helpers/cli_factory.py
@@ -1,0 +1,1358 @@
+"""
+This module is a more object oriented replacement for robottelo.cli.factory
+It is not meant to be used directly, but as part of a robottelo.hosts.Satellite instance
+example: my_satellite.cli_factory.make_org()
+"""
+import datetime
+import inspect
+import os
+import pprint
+import random
+from functools import lru_cache
+from functools import partial
+from os import chmod
+from tempfile import mkstemp
+from time import sleep
+
+from box import Box
+from fauxfactory import gen_alpha
+from fauxfactory import gen_alphanumeric
+from fauxfactory import gen_choice
+from fauxfactory import gen_integer
+from fauxfactory import gen_ipaddr
+from fauxfactory import gen_mac
+from fauxfactory import gen_netmask
+from fauxfactory import gen_url
+
+from robottelo import constants
+from robottelo import manifests
+from robottelo.cli.base import CLIReturnCodeError
+from robottelo.config import settings
+
+
+class CLIFactoryError(Exception):
+    """Indicates an error occurred while creating an entity using hammer"""
+
+
+def create_object(cli_object, options, values=None, credentials=None):
+    """
+    Creates <object> with dictionary of arguments.
+
+    :param cli_object: A valid CLI object.
+    :param dict options: The default options accepted by the cli_object
+        create
+    :param dict values: Custom values to override default ones.
+    :param list|tuple credentials: Username and password for non-default user.
+    :raise robottelo.host_helpers.cli_factory.CLIFactoryError: Raise an exception if object
+        cannot be created.
+    :rtype: dict
+    :return: A dictionary representing the newly created resource.
+
+    """
+    options.update(values or {})
+    if credentials:
+        cli_object = cli_object.with_user(*credentials)
+    try:
+        result = cli_object.create(options)
+    except CLIReturnCodeError as err:
+        # If the object is not created, raise exception, stop the show.
+        raise CLIFactoryError(
+            'Failed to create {} with data:\n{}\n{}'.format(
+                cli_object.__name__, pprint.pformat(options, indent=2), err.msg
+            )
+        )
+    # Sometimes we get a list with a dictionary and not a dictionary.
+    if type(result) is list and len(result) > 0:
+        result = result[0]
+    return Box(result)
+
+
+"""
+The following dictionary is used to define the simple make methods in this factory.
+Each key corresponds to the name of the entity (e.g. make_<entity_name>)
+The value of each entity key is a dictionary containing the following:
+    <option_name>: <value population function> Functions will be evaluated at runtime
+    _setup: <setup function>  Use this when a little more complexity is needed (see below)
+    _setup_args: [list of arguments for setup function]
+    _setup_kwargs: {dict of keyword arguments for setup function}
+    _redirect: 'name of make_<entity> function that should actually be ran'
+    _entity_cls: 'name of the EntityClass that should be used instead of key name'
+"""
+ENTITY_FIELDS = {
+    'activation_key': {
+        'name': gen_alphanumeric,
+    },
+    'architecture': {
+        'name': gen_alphanumeric,
+    },
+    'content_view': {'_redirect': 'content_view_with_credentials'},
+    'content_view_with_credentials': {
+        '_entity_cls': 'ContentView',
+        'name': gen_alpha,
+    },
+    'content_view_filter': {
+        'name': gen_alpha,
+    },
+    'content_view_filter_rule': {},
+    'discoveryrule': {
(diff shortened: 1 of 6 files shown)


robottelo/host_helpers/cli_factory.py
```diff
@@ -0,0 +1,285 @@
...
+from fauxfactory import gen_ipaddr
+from fauxfactory import gen_mac
+from fauxfactory import gen_netmask
+
+from robottelo import constants
+from robottelo.cli.base import CLIReturnCodeError
+from robottelo.config import settings
+from robottelo.helpers import update_dictionary
+
+# import os
```
mshriver, line 20: placeholder comment to cleanup

robottelo/host_helpers/cli_factory.py
```diff
@@ -0,0 +1,285 @@
...
+    },
+}
+
+
+class CLIFactory:
+    def __init__(self, satellite):
+        self._satellite = satellite
+
+    def __getattribute__(self, name):
+        """Handle cases when a make_ method is called (e.g. make_activation_key)"""
```
mshriver, line 244: Please add a lot more in-line documentation of the steps you're taking here and the intent of the functions in this class.

robottelo/host_helpers/cli_factory.py
JacobCallahan: placeholder, i had to comment out some imports that aren't used since i removed a large chunk of methods i'm still working on. the vast majority of these imports will be used in the next update

robottelo/host_helpers/cli_factory.py
JacobCallahan: If you have any specific areas of confusion, that would help me figure out all of what needs to be said. I don't want to explain what might be obvious.

robottelo/host_helpers/cli_factory.py
```diff
@@ -0,0 +1,285 @@
...
+        'network': partial(gen_ipaddr, True),
+    },
+    'sync_plan': {
+        'description': gen_alpha,
+        'enabled': 'true',
+        'interval': lambda: random.choice(list(constants.SYNC_INTERVAL.values())),
+        'name': gen_alpha,
+        'sync-date': lambda: datetime.datetime.now().strftime('%Y-%m-%d %H:%M:%S'),
+    },
+    'host': {
```
tstrych, line 138: make_host requires just ip, mac, name, root-password, btw that is not enough
for make_fake_host which is somewhat host organization, location, domain, architecture, os, partition table and medium is required, how can I create make_fake_host ?

tstrych:
I can test your PR with mine, only thing that kind of blocks me is `make_fake_host`.

robottelo/host_helpers/cli_factory.py
JacobCallahan: See if the additional comments help more. I don't want to come anywhere close to the overkill that is nailgun's documentation or teach what `__getattribute__` does via comments. However, the existing hints should give enough information to explain what it going on.

mshriver: The updated docblock (and I think some in-line comments are new?) help explain the purpose. Certainly don't need to document when/why/how `__getattribute__` itself is triggered, just why/how we're overriding it and what actions we're taking on the args or what's being called.

I think some more in-line documentation surrounding the field function execution would help. Placed above the dict definition, so when reading its not a guess why this giant dict is defined with some function objects being set as values but not evaluated.

rplevka:
This is a great addition. Would it be possible to include the `cdn_small` or `rh_small` set of repos as we discussed earlier?
These repos should hold reference to a reasonably small CDN-hosted repo suitable for a given rhel version.

robottelo/host_helpers/cli_factory.py
```diff
@@ -0,0 +1,1358 @@
+"""
+This module is a more object oriented replacement for robottelo.cli.factory
+It is not meant to be used directly, but as part of a robottelo.hosts.Satellite instance
+example: my_satellite.cli_factory.make_org()
```
jyejare, line 4: Why not ?
```suggestion
example: my_satellite.cli.make_org()
```
Remove the word factory altogether ?

robottelo/host_helpers/cli_factory.py
```diff
@@ -0,0 +1,1358 @@
...
+def create_object(cli_object, options, values=None, credentials=None):
+    """
+    Creates <object> with dictionary of arguments.
+
+    :param cli_object: A valid CLI object.
+    :param dict options: The default options accepted by the cli_object
+        create
+    :param dict values: Custom values to override default ones.
+    :param list|tuple credentials: Username and password for non-default user.
+    :raise robottelo.cli.factory.CLIFactoryError: Raise an exception if object
```
jyejare, line 46:
```suggestion
    :raise robottelo.host_helpers.cli_factory.CLIFactoryError: Raise an exception if object
```

jyejare (commented):
ACK pending questions / comments !

robottelo/host_helpers/cli_factory.py
JacobCallahan: cli is already an attribute that exists on the Satellite objects, it maps to the cli classes themselves.

We also can't adopt the nail gun pattern because the cli classes use class methods for their commands, so we don't actually instantiate them.

robottelo/host_helpers/cli_factory.py
```diff
@@ -0,0 +1,1358 @@
...
+        'hypervisor-id': 'hostname',
+        'interval': '60',
+        'name': gen_alphanumeric,
+        'satellite-url': lambda self: self._satellite.hostname,
+    },
+}
+
+
+class CLIFactory:
+    """This class is part of a mixin and not to be used directly. See robottelo.hosts.Satellite"""
```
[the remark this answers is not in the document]
JacobCallahan: The mixin pattern is common enough that we shouldn't be explaining the concept in our codebase.
Also, __new__ would raise more issues than it's worth. We won't stop anyone from instantiating this class, bit we ask them not to directly.

robottelo/host_helpers/cli_factory.py
jyejare: @JacobCallahan Yeah, cli attr is in Satellite object and we should append all the make methods to that cli attr / classes. This will simplify the process of using CLI as `sat.cli.make_component` in accordance with API `sat.api.Component`.

JacobCallahan: @jyejare it's not as simple as that. The cli attr is dynamically generated when needed, which is needed for pickle resilience.  Trying to inect these class methods into a namespace populated by classes would be messy.

mshriver: @jyejare what you're suggesting is something we could tackle by separating the hammer modeling library, but isn't reasonable with the current implementation and integration.  It is a better pattern, and one that the taretto stack suggests.

JacobCallahan:
@jyejare updated and ready for your final ACK/merge

mshriver (approved):

mshriver:
trigger: test-robottelo
pytest: tests/foreman/cli/
deploy_rhel_version: 8

jyejare (approved):

jyejare:
@JacobCallahan ACKed! Not sure if you want to look into PRT failure.

JacobCallahan:
@jyejare those failues look unrelated to my change, but I'm going to try another run

JacobCallahan:
trigger: test-robottelo
pytest: tests/foreman/cli/test_errata.py

JacobCallahan:
All failures appear unrelated to these changes.

PR #9286, opened 2022-01-11
Puppet plugin fixture and tests

vsedmik:
This PR adds:
- fixture for puppet-enabled Satellite
- fixture for `destructive_caps` (to be registered with `destructive_sat`)
- test for puppet enable/disable logic on both, Satellite and Capsule (waiting for BZ fix)
- test fixes in `cli/test_puppetclass`, `api/test_environment`, `cli/test_environment`

TODO:
- more puppet-related test fixes will follow in next PRs

PR #9328, opened 2022-01-26
Api testhost run puppet tests on destructive satellite with enabled puppet

tstrych:
rebased on #9286 
using nailgun #807

test results:
Screenshot from 2022-01-26 12-52-46

### DIFF
--- a/tests/foreman/api/test_host.py
+++ b/tests/foreman/api/test_host.py
@@ -37,6 +37,9 @@ from robottelo.api.utils import promote
 from robottelo.config import get_credentials
 from robottelo.config import settings
 
+from robottelo.constants import DEFAULT_CV
+from robottelo.constants import ENVIRONMENT
+
 
 @pytest.fixture(scope='function')
 def tracer_host(katello_host_tools_tracer_host):
@@ -333,7 +336,7 @@ def test_positive_create_with_inherited_params(module_org, module_location):
 
 
 @pytest.mark.tier1
-def test_positive_create_and_update_with_puppet_proxy(default_sat):
+def test_positive_create_and_update_with_puppet_proxy(module_puppet_enabled_sat):
     """Create a host with puppet proxy specified and then create new host without specified
     puppet proxy and update the new host with the same puppet proxy
 
@@ -344,19 +347,19 @@ def test_positive_create_and_update_with_puppet_proxy(default_sat):
     :CaseImportance: Critical
     """
     # TODO Define the default capsule/SP port + URL on hosts.Capsule
-    proxy = default_sat.api.SmartProxy().search(query={'search': f'url = {default_sat.url}:9090'})[
-        0
-    ]
-    host = default_sat.api.Host(puppet_proxy=proxy).create()
+    proxy = module_puppet_enabled_sat.api.SmartProxy().search(
+        query={'search': f'url = {module_puppet_enabled_sat.url}:9090'}
+    )[0]
+    host = module_puppet_enabled_sat.api.Host(puppet_proxy=proxy).create()
     assert host.puppet_proxy.read().name == proxy.name
-    new_host = default_sat.api.Host().create()
+    new_host = module_puppet_enabled_sat.api.Host().create()
     new_host.puppet_proxy = proxy
     new_host = new_host.update(['puppet_proxy'])
     assert new_host.puppet_proxy.read().name == proxy.name
 
 
 @pytest.mark.tier1
-def test_positive_create_with_puppet_ca_proxy(default_sat):
+def test_positive_create_with_puppet_ca_proxy(module_puppet_enabled_sat):
     """Create a host with puppet CA proxy specified and then create new host without specified
      puppet CA proxy and update the new host with the same puppet CA proxy
 
@@ -366,18 +369,26 @@ def test_positive_create_with_puppet_ca_proxy(default_sat):
 
     :CaseImportance: Critical
     """
-    proxy = entities.SmartProxy().search(query={'search': f'url = {default_sat.url}:9090'})[0]
-    host = entities.Host(puppet_ca_proxy=proxy).create()
+    proxy = module_puppet_enabled_sat.api.SmartProxy().search(
+        query={'search': f'url = {module_puppet_enabled_sat.url}:9090'}
+    )[0]
+    host = module_puppet_enabled_sat.api.Host(puppet_ca_proxy=proxy).create()
     assert host.puppet_ca_proxy.read().name == proxy.name
-    new_host = entities.Host().create()
+    new_host = module_puppet_enabled_sat.api.Host().create()
     new_host.puppet_ca_proxy = proxy
     new_host = new_host.update(['puppet_ca_proxy'])
     assert new_host.puppet_ca_proxy.read().name == proxy.name
 
 
 @pytest.mark.tier2
+@pytest.mark.skip_if_open("BZ:2046573")
 def test_positive_end_to_end_with_puppet_class(
-    module_org, module_location, module_env_search, module_puppet_classes, default_smart_proxy
+    module_puppet_org,
+    module_puppet_loc,
+    module_env_search,
+    module_puppet_classes,
+    module_default_puppet_proxy,
+    module_puppet_enabled_sat,
 ):
     """Create a host with associated puppet classes then remove it and update the host
     with same associated puppet classes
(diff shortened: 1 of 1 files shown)


vsedmik:
#9286 was merged into master a while ago
How do the other tests in test_host.py do? Can they pass?

tstrych:
top right corner of the picture says 48 passed,

vsedmik:
Hmm, looks I've missed the PuppetCA option in the installer command... :(
I will raise a fix PR and investigate the rest.

vsedmik:
It seems the issue is not in the installation, but in use of the destructive_sat.

For example, when I add it to the subnet test so that it looks like this:
`def test_positive_create_and_update_with_subnet(module_destructive_sat, module_location, module_org, module_default_subnet):`

all entities (including the `module_location`, `module_org` and `module_default_subnet` fixtures!!) are created at the destructive sat, instead of the default one. I don't understand why it happens (I would expect that `entities.*` lead to the default_sat) but I suspect it could be the cause.

pytest_fixtures/component/puppet.py
```diff
@@ -114,3 +114,14 @@ def module_puppet_classes(
...
+
+@pytest.fixture(scope="module")
+def module_default_puppet_proxy(module_puppet_enabled_sat):
+    """Use the default installation puppet smart proxy"""
+    proxy = (
+        module_puppet_enabled_sat.api.SmartProxy()
+        .search(query={'search': f'url = {module_puppet_enabled_sat.url}:9090'})[0]
+        .read()
+    )
+    return module_puppet_enabled_sat.api.SmartProxy(id=proxy.id).read()
```
jyejare, line 127: You dont need double read here you have done reading in previous step!

tests/foreman/api/test_host.py
```diff
@@ -325,19 +327,19 @@ def test_positive_create_and_update_with_puppet_proxy(default_sat):
     :CaseImportance: Critical
     """
     # TODO Define the default capsule/SP port + URL on hosts.Capsule
-    proxy = default_sat.api.SmartProxy().search(query={'search': f'url = {default_sat.url}:9090'})[
-        0
-    ]
-    host = default_sat.api.Host(puppet_proxy=proxy).create()
+    proxy = module_puppet_enabled_sat.api.SmartProxy().search(
+        query={'search': f'url = {module_puppet_enabled_sat.url}:9090'}
+    )[0]
```
jyejare, line 332: This should be `module_default_puppet_proxy` that you have created in this same PR!

tests/foreman/api/test_host.py
```diff
@@ -347,18 +349,26 @@ def test_positive_create_with_puppet_ca_proxy(default_sat):
 
     :CaseImportance: Critical
     """
-    proxy = entities.SmartProxy().search(query={'search': f'url = {default_sat.url}:9090'})[0]
-    host = entities.Host(puppet_ca_proxy=proxy).create()
+    proxy = module_puppet_enabled_sat.api.SmartProxy().search(
+        query={'search': f'url = {module_puppet_enabled_sat.url}:9090'}
+    )[0]
```
jyejare, line 354: Same here

tests/foreman/api/test_host.py
```diff
@@ -347,18 +349,26 @@ def test_positive_create_with_puppet_ca_proxy(default_sat):
...
     assert host.puppet_ca_proxy.read().name == proxy.name
-    new_host = entities.Host().create()
+    new_host = module_puppet_enabled_sat.api.Host().create()
     new_host.puppet_ca_proxy = proxy
     new_host = new_host.update(['puppet_ca_proxy'])
     assert new_host.puppet_ca_proxy.read().name == proxy.name
 
 
 @pytest.mark.tier2
+@pytest.mark.skip_if_open("BZ:2046573")
```
jyejare, line 364: Add this BZ in the docstring of this test!

tests/foreman/api/test_host.py
```diff
@@ -347,18 +349,26 @@ def test_positive_create_with_puppet_ca_proxy(default_sat):
...
 def test_positive_end_to_end_with_puppet_class(
-    module_org, module_location, module_env_search, module_puppet_classes, default_smart_proxy
+    module_puppet_org,
+    module_puppet_loc,
+    module_env_search,
+    module_puppet_classes,
+    module_default_puppet_proxy,
```
jyejare, line 370: I think we should rename this fixture as to:
```
module_puppet_enabled_proxy
```
or
```
module_puppet_enabled_default_proxy
```

The default word there points to default_sat so puppet enabled will be helpful!

tests/foreman/api/test_host.py
```diff
@@ -1040,25 +1052,35 @@ def test_positive_read_enc_information(
...
     :CaseLevel: Integration
     """
-    update_smart_proxy(module_location, default_smart_proxy)
+    lce = (
+        module_puppet_enabled_sat.api.LifecycleEnvironment()
+        .search(query={'search': f'name={ENVIRONMENT} and organization_id={module_puppet_org.id}'})[
+            0
+        ]
+        .read()
+    )
```
jyejare, line 1061: Looks like black has done some weird formatting here, please fix it !

tests/foreman/api/test_host.py
```diff
@@ -1330,14 +1352,16 @@ def test_positive_read_puppet_proxy_name(default_sat):
 
     :CaseImportance: Critical
     """
-    proxy = entities.SmartProxy().search(query={'search': f'url = {default_sat.url}:9090'})[0]
-    host = entities.Host(puppet_proxy=proxy).create().read_json()
+    proxy = module_puppet_enabled_sat.api.SmartProxy().search(
+        query={'search': f'url = {module_puppet_enabled_sat.url}:9090'}
+    )[0]
```
jyejare, line 1357: same here !

tests/foreman/api/test_host.py
```diff
@@ -1349,8 +1373,10 @@ def test_positive_read_puppet_ca_proxy_name(default_sat):
 
     :CaseImportance: Critical
     """
-    proxy = entities.SmartProxy().search(query={'search': f'url = {default_sat.url}:9090'})[0]
-    host = entities.Host(puppet_ca_proxy=proxy).create().read_json()
+    proxy = module_puppet_enabled_sat.api.SmartProxy().search(
+        query={'search': f'url = {module_puppet_enabled_sat.url}:9090'}
+    )[0]
```
jyejare, line 1378: Here too!

jyejare (commented):
ACK pending comments !

vsedmik:
Needs rebase on #9382

jyejare:
@tstrych You also have conflicts, please resolve them !

tests/foreman/api/test_host.py
JacobCallahan: if black is formatting it this way, there isn't much more that can be done. perhaps splitting that formatted string, but that's about it.

mshriver: Yeah this is one of those situations where we accepted Black, so that's what we take here. We can shorten the line by creating more variables that end up used once, but it is readable, just a bit ugly.

jyejare:
```suggestion
        .search(
            query={
                'search': f'name={ENVIRONMENT} and organization_id={module_puppet_org.id}'
            })[0].read())
```
Cant we do somthing like this by ourself? Would black still format this ?

tstrych:
```==== 50 passed, 13 deselected, 806 warnings, 2 errors in 4430.17s (1:13:50) ====```

Failed tests: (not connected)
```
test_positive_tracer_list_and_resolve[rhel9]
test_positive_tracer_list_and_resolve[rhel8]
```

File: tests/foreman/api/test_host.py (hunk 6 above)
Comment (mshriver) [reply tests/foreman/api/test_host.py:1081]: dunno, you'd have to stick it in a branch and run Black to see. Likely yes, its going to apply its rules. The only good way to effect it here is to make the value in the dict shorter so it doesn't have to line break.

File: tests/foreman/api/test_host.py (hunk 5 above)
Comment (tstrych) [reply tests/foreman/api/test_host.py]: it's called `session_puppet_enabled_proxy` now

File: pytest_fixtures/component/puppet.py (hunk 1 above)
Comment (tstrych) [reply pytest_fixtures/component/puppet.py]: already fixed in actual fixture

**Review by Gauravtalreja1 — APPROVED** [review]

File: tests/foreman/api/test_host.py (hunk 4 above)
Comment (jyejare) [reply tests/foreman/api/test_host.py]: @tstrych I dont see its added in the docstring of the bug btw!

**Review by jyejare — APPROVED** [review]
ACK with non blocking comments!
- Add BZ to the docstring
- Weird Black formatting

File: tests/foreman/api/test_host.py [hunk 9]
```diff
@@ -366,35 +369,43 @@ def test_positive_create_with_puppet_ca_proxy(default_sat):
...
+    module_puppet_classes,
+    session_puppet_enabled_proxy,
+    session_puppet_enabled_sat,
 ):
     """Create a host with associated puppet classes then remove it and update the host
     with same associated puppet classes
 
     :id: 2690d6b0-441b-44c5-b7d2-4093616e037e
 
+    :BZ: 2046573
```
tstrych, line 396: @jyejare BZ is already added,

jyejare: Ohh, not sure how I missed it !

JacobCallahan (approved):

PR #9329, opened 2022-01-27
Add manual tests for BZ#1615015 and BZ#1886040

shubhamsg199:
Added two manual test cases

1. https://bugzilla.redhat.com/show_bug.cgi?id=1615015
2. https://bugzilla.redhat.com/show_bug.cgi?id=1886040

### DIFF
--- a/tests/foreman/cli/test_foremantask.py
+++ b/tests/foreman/cli/test_foremantask.py
@@ -1,0 +1,52 @@
+"""Tests for the ``foreman_tasks/api/v2/tasks`` path.
+
+:Requirement: Foremantask
+
+:CaseAutomation: Automated
+
+:CaseLevel: Component
+
+:CaseComponent: TasksPlugin
+
+:Assignee: pondrejk
+
+:TestType: Functional
+
+:CaseImportance: High
+
+:Upstream: No
+"""
+import pytest
+
+
+@pytest.mark.stubbed
+@pytest.mark.tier3
+def test_positive_tasks_backup():
+    """Check no tasks are backup by default when cleaning up tasks from the database and
+    no /var/lib/foreman/tasks-backup created for backup tasks.
+
+    :id: 3575492c-7802-4b00-aefc-d53141ad7446
+
+    :BZ: 1615015
+
+    :customerscenario: true
+
+    :steps:
+        1. To run tasks cleanup every 5mins,
+           `satellite-installer --foreman-plugin-tasks-cron-line '5 * * * *'`
+        2. satellite-installer --foreman-plugin-tasks-backup true (which is default:false)
+        3. Verify TASK_BACKUP=true in /etc/cron.d/foreman-tasks (which is default:false)
+        4. Create any product or sync any repo to have a valid task.
+        5. Include this line "TASK_SEARCH='label = Actions::Katello::Product::Create'"
+           to /etc/cron.d/foreman-tasks
+        6. Watch /var/log/foreman/cron.log for task cleanup of above labelled tasks.
+        7. Verify /var/lib/foreman/tasks-backup created and has details of cleaned up tasks.
+
+    :expectedresults: No tasks are backup by default when cleaning up tasks from the database,
+                      and no /var/lib/foreman/tasks-backup created for backup tasks.
+
+    :CaseImportance: High
+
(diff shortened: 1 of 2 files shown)


tests/foreman/cli/test_foremantask.py
```diff
@@ -0,0 +1,47 @@
+"""Tests for the ``foreman_tasks/api/v2/tasks`` path.
+
+:Requirement: Foremantask
+
+:CaseAutomation: Automated
```
Griffin-Sullivan, line 5: I don't know much about docstrings for new files, but since there's no automation in here yet shouldn't this be `NotAutomated`?

tests/foreman/cli/test_foremantask.py
```diff
@@ -0,0 +1,47 @@
...
+
+:CaseImportance: High
+
+:Upstream: No
+"""
+import pytest
+
+
+@pytest.mark.stubbed
+@pytest.mark.tier1
```
Griffin-Sullivan, line 23: It seems like for both tests we are looking at how satellite installer interacts with the system. Therefore I would say both are `tier3`. More detail here or could ask around for what would fit better, but it doesn't seem like either of these tests would be CRUD: https://github.com/SatelliteQE/robottelo/blob/master/pytest_plugins/markers.py#L11

tests/foreman/cli/test_foremantask.py
shubhamsg199: right, I completely forgot about that!

shubhamsg199: right, I totally forgot about that!

tests/foreman/cli/test_foremantask.py
```diff
@@ -0,0 +1,47 @@
...
+@pytest.mark.stubbed
+@pytest.mark.tier3
+def test_positive_tasks_backup():
+    """Check no tasks are backup by default when cleaning up tasks from the database and
+    no /var/lib/foreman/tasks-backup created for backup tasks.
+
+    :id: 3575492c-7802-4b00-aefc-d53141ad7446
+
+    :BZ: 1615015
+
```
jameerpathan111, line 31:
```suggestion

:customerscenario: true
```

jameerpathan111 (changes requested):

tests/foreman/cli/test_http_proxy.py
```diff
@@ -175,3 +175,24 @@ def test_positive_set_content_default_http_proxy(block_fake_repo_access):
...
+
+@pytest.mark.stubbed
+@pytest.mark.tier3
+def test_positive_environment_variable_unset_set():
+    """Verify that satellite installer unsets and then sets back the environment variables
+
+    :id: 596d753b-660b-49cb-b663-ff3cec439564
+
+    :BZ: 1886040
+
```
jameerpathan111, line 188:
```suggestion

:customerscenario: true
```
Both bugzilla's have customer cases attached to them.

jameerpathan111 (changes requested):

tests/foreman/cli/test_foremantask.py
Gauravtalreja1: @shubhamsg199 I think we should keep `Automated` here and shouldn't keep `NotAutomated` at a module level
 instead, we can add `:CaseAutomation: NotAutomated` at the test level

tests/foreman/cli/test_foremantask.py
Gauravtalreja1: Also add necessary `:CaseImportance:` here

tests/foreman/cli/test_http_proxy.py
Gauravtalreja1: Also add necessary `:CaseImportance: ` here

tests/foreman/cli/test_foremantask.py
shubhamsg199: Okay, keeping  `:CaseAutomation: Automated` at the modue level and adding `:CaseAutomation:` at test level

Gauravtalreja1 (approved):
Looks good, Thanks for the changes!

jameerpathan111 (approved):

pondrejk (approved):

PR #9330, opened 2022-01-27
Test Coverage for insights inventory sync

adarshdubey-star:
**Test results**

```
 pytest tests/foreman/cli/test_rhcloud_inventory.py -k test_positive_inventory_recommendation_sync
platform linux -- Python 3.8.6, pytest-6.2.5, py-1.10.0, pluggy-1.0.0
plugins: services-2.2.1, mock-3.6.1, reportportal-5.0.8, forked-1.3.0, ibutsu-2.0.2, cov-3.0.0, xdist-2.5.0

tests/foreman/cli/test_rhcloud_inventory.py .                                                                                                                                                               [100%]

=================================================================================== 1 passed, 5 deselected in 898.61s (0:14:58) ==================================================================================
```

### DIFF
--- a/tests/foreman/cli/test_rhcloud_inventory.py
+++ b/tests/foreman/cli/test_rhcloud_inventory.py
@@ -90,8 +90,13 @@ def test_positive_inventory_generate_upload_cli(
         assert hosts_count == local_file_data['slices_counts'][slice_name]
 
 
-@pytest.mark.stubbed
-def test_positive_inventory_recommendation_sync():
+@pytest.mark.tier3
+def test_positive_inventory_recommendation_sync(
+    set_rh_cloud_token,
+    organization_ak_setup,
+    rhcloud_registered_hosts,
+    rhcloud_sat_host,
+):
     """Tests Insights recommendation sync via foreman-rake commands:
     https://github.com/theforeman/foreman_rh_cloud/blob/master/README.md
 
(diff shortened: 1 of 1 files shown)


jameerpathan111 (approved):

tests/foreman/cli/test_rhcloud_inventory.py
```diff
@@ -107,10 +112,26 @@ def test_positive_inventory_recommendation_sync():
...
-    :CaseAutomation: NotAutomated
+    :CaseAutomation: Automated
 
     :CaseLevel: System
     """
+    org, ak = organization_ak_setup
+    cmd = f'organization_id={org.id} foreman-rake rh_cloud_insights:sync'
+    timestamp = datetime.utcnow().strftime('%Y-%m-%d %H:%M')
+    result = rhcloud_sat_host.execute(cmd)
+    wait_for(
```
Griffin-Sullivan, line 123: You could consider using `wait_for_tasks` here, although I think that might require you to get the task id rather than searching for `Insights full sync...`. So this implementation might be cleaner as is.

adarshdubey-star: Yes, that could be the case, but instead, I chose the timestamp to segregate the required one just to be more clear the Insights full sync gets done ...and the hosts hit data.

Griffin-Sullivan: Awesome just wanted to make sure you were aware of the `wait_for_tasks` function

tests/foreman/cli/test_rhcloud_inventory.py
```diff
@@ -107,10 +112,26 @@ def test_positive_inventory_recommendation_sync():
...
+        .search(query={'search': f'Insights full sync and started_at >= "{timestamp}"'})[0]
+        .result
+        == 'success',
+        timeout=400,
+        delay=15,
+        silent_failure=True,
+        handle_exception=True,
+    )
+    assert result.status == 0
+    assert result.stdout == 'Synchronized Insights hosts hits data\n'
```
jyejare, line 134: You are referring to the result before inventory sync is completed or just started!  You might have status code `0` but it would be for start syncing but not for completion. I recommend you to get this status for sync completion else you will have false pass results.

Also, stdout is correct here ? and is it for start and not for complete? if not for complete, check it for complete to assert the sync works !

jyejare (changes requested):
ACK pending comments !

Also, please reconsider the `Importance` of the test case if its really `High` or should be `Medium`.

tests/foreman/cli/test_rhcloud_inventory.py
adarshdubey-star: In the `wait_for` function there is a small assertion kind of comparison 'result=='sucess'' that takes care of whether the task resulted in success as the cmd got executed previously, it's just checking the status. About the stdout you are right it gets reflected after the completion of the task when the hosts hit insights data.

swadeley (approved):
ACK

PR #9327, opened 2022-01-26
expand content_host.yml to support custom deploy workflows per nick

rplevka:
Partially Addresses: SAT-7681

Please, merge together with the backports:
- 6.10.z: https://github.com/SatelliteQE/robottelo/pull/9336
- 6.9.z:  https://github.com/SatelliteQE/robottelo/pull/9335

This allows us to define custom deploy workflow per individual OS version entry - this is needed as Compose versions are using different naming and thus different deploy workflow.
A sample usage can be seen in the updated `content_host.yml.sample`

PR #9334, opened 2022-01-27
add rhel_contenthost-specific marks to filter os version parametrization

rplevka:
Adding 2 new marks to make RHEL version (or nick) filtering possible:

- `pytest.mark.rhel_nick_match('<regex pattern>')`
- `pytest.mark.rhel_nick_list([applicable, nick, list])`
(feel free to suggest more suitable names)

With nicks for various rhel versions configured in `content_host.yml` file, e.g.:

```yaml
content_host:
  deploy_workflow: deploy-base-rhel
  default_rhel_version: 7
  rhel_versions:
    6: {}
    7: {}
    8: {}
    9_compose:
      deploy_workflow: deploy-rhel-cmp-template
```

__Default behavior__
all tests using `rhel_contenthost` fixture (directly or indirectly) are being parametrized for each nick:

```bash
$ pytest -k "test_positive_subscription_status_disabled"  tests/foreman/api/test_subscription.py --collect-only
==== test session starts ====
...
<Package api>
  <Module test_subscription.py>
    <Function test_positive_subscription_status_disabled[rhel6]>
    <Function test_positive_subscription_status_disabled[rhel7]>
    <Function test_positive_subscription_status_disabled[rhel8]>
    <Function test_positive_subscription_status_disabled[rhel9_compose]>

==== 5/15 tests collected (10 deselected) in 0.04s ====

```

__Altered behavior__
Having the test decorated with one (or possibly more) of the introduced markers, e.g.:
```python
@pytest.mark.tier2
@pytest.mark.rhel_nick_list([8,'9_compose'])
def test_positive_subscription_status_disabled(
    module_ak, rhel_contenthost, module_org, default_sat
):
...
```
will only pick matching nicks and result in the following collection:
```bash
$ pytest -k "test_positive_subscription_status_disabled"  tests/foreman/api/test_subscription.py --collect-only
==== test session starts ====
...
<Package api>
  <Module test_subscription.py>
    <Function test_positive_subscription_status_disabled[rhel8]>
    <Function test_positive_subscription_status_disabled[rhel9_compose]>

=============================== 2/12 tests collected (10 deselected) in 0.03s ===============================
```
Using the regexp matching gives you more freedom in selecting just what you want (uses `re.fullmatch`):
```python
@pytest.mark.tier2
@pytest.mark.rhel_nick_match('[6,9].*') #for whatever crazy reason you might have for this
def test_positive_subscription_status_disabled(
    module_ak, rhel_contenthost, module_org, default_sat
):
...
```
results in:
```bash
$ pytest -k "test_positive_subscription_status_disabled"  tests/foreman/api/test_subscription.py --collect-only
==== test session starts ====
...
<Package api>
  <Module test_subscription.py>
    <Function test_positive_subscription_status_disabled[rhel6]>
    <Function test_positive_subscription_status_disabled[rhel9_compose]>

==== 2/12 tests collected (10 deselected) in 0.03s ====
```

The same approach can be used to define even more matcher marks (e.g. blacklist, ...)

merge preferably after:
https://github.com/SatelliteQE/robottelo/pull/9327

### DIFF
--- a/pytest_plugins/fixture_markers.py
+++ b/pytest_plugins/fixture_markers.py
@@ -1,3 +1,4 @@
+import re
 from inspect import getmembers
 from inspect import isfunction
 
@@ -6,19 +7,43 @@ from robottelo.config import settings
 
 def pytest_generate_tests(metafunc):
     if 'rhel_contenthost' in metafunc.fixturenames:
-        rhel_parameters = []
-        for ver in settings.content_host.rhel_versions.keys():
-            # prefer nick-specific deploy workflow before using the default one
-            workflow = settings.content_host.rhel_versions[ver].get(
-                'deploy_workflow', settings.content_host.deploy_workflow
+        function_marks = getattr(metafunc.function, 'pytestmark', [])
+        # process eventual rhel_version_list markers
+        matchers = [i.args for i in function_marks if i.name == 'rhel_ver_list']
+        list_params = []
+        for matcher in matchers:
+            list_params.extend(
+                [
+                    setting_rhel_ver
+                    for setting_rhel_ver in settings.content_host.rhel_versions
+                    if str(setting_rhel_ver) in str(matcher)
+                ]
             )
-            rhel_parameters.append(dict(workflow=workflow, rhel_version=ver))
-        metafunc.parametrize(
-            'rhel_contenthost',
-            rhel_parameters,
-            ids=[f'rhel{r["rhel_version"]}' for r in rhel_parameters],
-            indirect=True,
-        )
+        # process eventual rhel_version_match markers
+        matchers = [i.args for i in function_marks if i.name == 'rhel_ver_match']
+        match_params = []
+        for matcher in matchers:
+            match_params.extend(
+                [
+                    setting_rhel_ver
+                    for setting_rhel_ver in settings.content_host.rhel_versions
+                    if re.fullmatch(str(matcher[0]), str(setting_rhel_ver))
+                ]
+            )
+        rhel_params = []
+        filtered_versions = set(list_params + match_params)
+        for ver, data in settings.content_host.rhel_versions.items():
+            if ver in (filtered_versions or settings.content_host.rhel_versions):
+                # prefer version-specific deploy workflow before using the default one
+                workflow = data.get('deploy_workflow', settings.content_host.deploy_workflow)
+                rhel_params.append(dict(workflow=workflow, rhel_version=ver))
+        if rhel_params:
+            metafunc.parametrize(
+                'rhel_contenthost',
+                rhel_params,
+                ids=[f'rhel{r["rhel_version"]}' for r in rhel_params],
+                indirect=True,
+            )
 
 
 def pytest_configure(config):
--- a/pytest_plugins/markers.py
+++ b/pytest_plugins/markers.py
@@ -16,6 +16,8 @@ def pytest_configure(config):
         "pit_client: PIT client scenario tests",
         "run_in_one_thread: Sequential tests",
         "build_sanity: Fast, basic tests that confirm build is ready for full test suite",
+        "rhel_ver_list: Filter rhel_contenthost versions by list",
+        "rhel_ver_match: Filter rhel_contenthost versions by regexp",
     ]
     markers.extend(module_markers())
     for marker in markers:


pytest_plugins/fixture_markers.py
```diff
@@ -6,16 +7,40 @@
...
-            for ver in settings.content_host.rhel_versions
-        ]
-        metafunc.parametrize(
-            'rhel_contenthost',
-            rhel_parameters,
-            ids=[f'rhel{r["rhel_version"]}' for r in rhel_parameters],
-            indirect=True,
-        )
+        rhel_parameters = []
+        for ver in settings.content_host.rhel_versions.keys():
```
JacobCallahan, line 11: Could iterate over items here to save the dict lookup later

JacobCallahan (changes requested):
Could you do the evaluations for what are the used rhel versions first, then use that list as a filter for pulling arguments from the settings? This should make it a bit more efficient.
Also, I'd recommend changing "nick" to "ver"? I think that would reduce confusion between this and Broker's concept of nicks.

All in all though, this will be a great change!

pytest_plugins/fixture_markers.py
```diff
@@ -6,16 +7,40 @@
...
+            # prefer nick-specific deploy workflow before using the default one
+            workflow = settings.content_host.rhel_versions[ver].get(
```
jyejare, line 13: Do a corresponding changes to `content_host.yaml.template` to show how rhel_versions now accept `deploy_workflow`!

rplevka: @jyejare when you check the PR descirption, you can see:
>merge preferably after:
>#9327

^ that PR introduces the changes you're requesting

pytest_plugins/fixture_markers.py
rplevka: Yeah, I can certainly do that. I'm just wondering how to do this so it won't make the code unreadable.
See, I first have to extract various kinds of selectors outside of the loop (new selector kinds can be added by time).
Then, I get to iterate over possible multiple instances of the selectors of the same kind inside the loop and then do the same for other kinds...

Maybe the benefits of keeping the selector-kind-specific in the distinct code blocks might outweight the efficiency benefit of the suggested change?

Griffin-Sullivan:
> Also, I'd recommend changing "nick" to "ver"? I think that would reduce confusion between this and Broker's concept of nicks.

I agree. I think `ver` would make more sense in this context.

pondrejk (approved):

rplevka:
> > Also, I'd recommend changing "nick" to "ver"? I think that would reduce confusion between this and Broker's concept of nicks.
> 
> I agree. I think `ver` would make more sense in this context.

Thanks. I already renamed them to `ver`

pytest_plugins/fixture_markers.py
JacobCallahan: maybe something like this? completely untested and written minutes before i go to bed, so definitely don't take this as a "this will work" suggestion. but what i'm thinking about right now
```python
        matchers = [i.args for i in metafunc.function.pytestmark if i.name == 'rhel_version_list']
        for m in matchers:
            if m:
                list_params = [
                    param
                    for param in settings.content_host.rhel_versions 
                    if str(param['rhel_version']) in str(m[0])
                ]
        # process eventual rhel_nick_match markers
        matchers = [i.args for i in metafunc.function.pytestmark if i.name == 'rhel_version_match']
        for m in matchers:
            if m:
                match_params = [
                    param
                    for param in settings.content_host.rhel_versions
                    if re.fullmatch(m[0], str(param['rhel_version']))
                ]
        rhel_params = []
        for ver in settings.content_host.rhel_versions.keys():
            if ver in set(list_params + match_params):
                # prefer version-specific deploy workflow before using the default one
                workflow = settings.content_host.rhel_versions[ver].get(
                    'deploy_workflow', settings.content_host.deploy_workflow
                )
            rhel_params.append(dict(workflow=workflow, rhel_version=ver))
        if rhel_params:
            metafunc.parametrize(
                'rhel_contenthost',
                rhel_params,
                ids=[f'rhel{r["rhel_version"]}' for r in rhel_params],
                indirect=True,
            )
```

JacobCallahan: the `if m:` checks probably aren't necessary and could save on a bit more visual noise.

rplevka: This looks nice.
However, it changes the behavior of how multiple selector markers are treated.
- if multiple instances of `rhel_version_list` mark are used, only the last processed one will take effect.
- if both kinds of marks are used (`rhel_version_list` together with `rhel_version_match`), they will be combined using `OR` logic, instead of the original `AND`

Am I right?
I don't think any of this is wrong tho, we can go with it.

JacobCallahan: I may be wrong, but I read the original flow as:
```
Get all the versions and data from settings
Reduce that list down to what was specified in rhel_version_list
Reduce that list down to what was specified in rhel_version_match
```
Whereas the altered version is
```
Construct a list of desired versions from rhel_version_list
Construct a list of desired versions from rhel_version_match
Get all the versions and data from settings if they are in either of the above lists
```

JacobCallahan: Now that I look at it more, wouldn't this work better than my previous example?
```python
        matchers = [i.args for i in metafunc.function.pytestmark if i.name == 'rhel_version_list']
        list_params = [
            param
            for param in settings.content_host.rhel_versions 
            if str(param['rhel_version']) in str(matchers)
        ] if matchers else []
        # process eventual rhel_version_match markers
        matchers = [i.args for i in metafunc.function.pytestmark if i.name == 'rhel_version_match']
        match_params = [
            param
            for param in settings.content_host.rhel_versions
            if re.fullmatch(str(matchers), str(param['rhel_version']))
        ] if matchers else []
        rhel_params = []
        for ver, data in settings.content_host.rhel_versions.items():
            if ver in set(list_params + match_params):
                # prefer version-specific deploy workflow before using the default one
                workflow = data.get(
                    'deploy_workflow', settings.content_host.deploy_workflow
                )
            rhel_params.append(dict(workflow=workflow, rhel_version=ver))
        if rhel_params:
            metafunc.parametrize(
                'rhel_contenthost',
                rhel_params,
                ids=[f'rhel{r["rhel_version"]}' for r in rhel_params],
                indirect=True,
            )
```

rplevka: well, I guess, but this does not cover the scenario when no selector marks are passed and thus all versions are supposed to be picked.
I'm increasingly more leaning towards my original solution, tbh.

tstrych (approved):
ACK, happy to see this additions, thanks

JacobCallahan (approved):
ACK. Merge pending CI readiness

pytest_plugins/fixture_markers.py
```diff
@@ -6,19 +7,42 @@
...
+        function_marks = getattr(metafunc.function, 'pytestmark', [])
+        # process eventual rhel_version_list markers
+        matchers = [i.args for i in function_marks if i.name == 'rhel_ver_list']
+        list_params = []
+        for matcher in matchers:
+            list_params.extend(
+                [
+                    param
+                    for param in settings.content_host.rhel_versions.keys()
+                    if str(param) in str(matcher)
```
jyejare, line 19: This will ease the code undesranding a bit:
```suggestion
                    if str(setting_rhel_ver) in str(fn_rhel_ver)
```

jyejare (commented):
ACK with minor nitpick and improvement comments!

Nitpicks are mostly related to variable renaming for someone new to understand quickly / easily !

pytest_plugins/fixture_markers.py
```diff
@@ -6,19 +7,42 @@
...
-                'deploy_workflow', settings.content_host.deploy_workflow
+        function_marks = getattr(metafunc.function, 'pytestmark', [])
+        # process eventual rhel_version_list markers
+        matchers = [i.args for i in function_marks if i.name == 'rhel_ver_list']
+        list_params = []
+        for matcher in matchers:
+            list_params.extend(
+                [
+                    param
+                    for param in settings.content_host.rhel_versions.keys()
```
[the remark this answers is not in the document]
rplevka: that is on purpose - i'm targeting only the key names, not the values:
```
                [
                    param
                    for param in settings.content_host.rhel_versions.keys()
                    if str(param) in str(matcher)
                ]
```

pytest_plugins/fixture_markers.py
```diff
@@ -6,19 +7,42 @@
...
+            match_params.extend(
+                [
+                    param
+                    for param in settings.content_host.rhel_versions.keys()
+                    if re.fullmatch(str(matcher[0]), str(param))
+                ]
+            )
+        rhel_params = []
+        for ver, data in settings.content_host.rhel_versions.items():
+            if ver in (set(list_params + match_params) or settings.content_host.rhel_versions):
```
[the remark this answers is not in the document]
rplevka: i mean - sure, but is this a blocker? We don't use this anywhere else so having it saved in a var could be an overkill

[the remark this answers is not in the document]
jyejare: No its not blocker, but not something I would like to see you in looping! Just an improvement comment !

pytest_plugins/fixture_markers.py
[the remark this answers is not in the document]
jyejare: Yup ofcourse! But I mean to say you need not to explicitly say that via `.keys()`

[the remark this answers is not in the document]
rplevka: I implemented your suggestion

pytest_plugins/fixture_markers.py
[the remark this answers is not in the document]
rplevka: @jyejare  gotcha, comment addressed

pytest_plugins/fixture_markers.py
rplevka: suggestion implemented

tests/foreman/api/test_subscription.py
```diff
@@ -234,6 +234,7 @@ def test_positive_delete_manifest_as_another_user(function_org, default_sat):
 
 
 @pytest.mark.tier2
+@pytest.mark.rhel_ver_match('.*compose')
 def test_positive_subscription_status_disabled(
```
jyejare, line 238: Is this intntional? that you wanted to run this test only for `composes` ?

rplevka: sorry, this was already a part of my testing. not a part of this PR. removed

jyejare (approved):

PR #9347, opened 2022-02-02
Removing a puppet test case that was failing in 6.10

latran:
This test case is no longer applicable after discussing with vlad.  Removing for so it gives a better automation result.


sambible (approved):

vsedmik (approved):
ACK

swadeley (approved):
ACK

PR #9349, opened 2022-02-03
foreman-maintain replaced to satellite-maintain

tstrych:
we should use downstream name for foreman-maintain and that is satellite maintain

I did rename files from `/robottello/cli/fm*` to `/robottelo/cli/sm*` and I didn't found any usages of these files in robotello, but it didn't feel quite right so maybe I just didn't found the usages or I need explanation why we use them. 

other usages of FM are in: `upgrade.yaml`  and `subscription.yaml`

### DIFF
--- a/tests/foreman/sys/test_katello_certs_check.py
+++ b/tests/foreman/sys/test_katello_certs_check.py
@@ -272,7 +272,7 @@ class TestKatelloCertsCheck:
             assert 'SSL certificate verification failed' not in result.stdout
             assert result.stdout.count('ok') == 7
             # assert all services are running
-            result = satellite.execute('foreman-maintain health check --label services-up -y')
+            result = satellite.execute('satellite-maintain health check --label services-up -y')
             assert result.status == 0, 'Not all services are running'
         finally:
             # revert to original certs
@@ -283,7 +283,7 @@ class TestKatelloCertsCheck:
             result = satellite.execute('hammer ping')
             assert result.status == 0, f'Hammer Ping failed:\n{result.stderr}'
             # assert all services are running
-            result = satellite.execute('foreman-maintain health check --label services-up -y')
+            result = satellite.execute('satellite-maintain health check --label services-up -y')
             assert result.status == 0, 'Not all services are running'
 
     @pytest.mark.destructive
@@ -335,7 +335,7 @@ class TestKatelloCertsCheck:
         assert 'SSL certificate verification failed' not in result.stdout
         assert result.stdout.count('ok') == 7
         # assert all services are running
-        result = rhel_vm.execute('foreman-maintain health check --label services-up -y')
+        result = rhel_vm.execute('satellite-maintain health check --label services-up -y')
         assert result.status == 0, 'Not all services are running'
 
     @pytest.mark.destructive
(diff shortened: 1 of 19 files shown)


swadeley (approved):
ACK

tests/foreman/sys/test_katello_certs_check.py
```diff
@@ -131,7 +131,7 @@ def validate_output(self, result, cert_data):
         # validate correct installer used
         pattern, count = 'satellite-installer --scenario satellite', 2
         assert len(re.findall(pattern, result.stdout)) == count
-        pattern, count = 'foreman-installer --scenario katello', 0
```
Gauravtalreja1, line 134: I believe this should not be changed, as this step/test validates if the correct installer is used, @omkarkhatavkar can confirm?

Gauravtalreja1 (changes requested):

tests/foreman/sys/test_katello_certs_check.py
tstrych: it tests that there is 0 `katello` options, in installer, it doesn't matter which installer you are looking at

omkarkhatavkar: @Gauravtalreja1 is correct, earlier we used to get a lot of issues with the cert generation command we wanted to ensure that foreman-installer options should not be seen. Please revert this.

Gauravtalreja1:
@tstrych any status update?

tests/foreman/sys/test_katello_certs_check.py
```diff
@@ -129,7 +129,7 @@ def validate_output(self, result, cert_data):
         assert cert_data['success_message'] in result.stdout
 
         # validate correct installer used
-        pattern, count = 'satellite-installer --scenario satellite', 2
+        pattern, count = 'foreman-installer --scenario satellite', 2
```
Gauravtalreja1, line 132: I guess, this should not be changed for this test, as it checks first if the right installer is used, and the upstream installer is not used, @omkarkhatavkar can confirm?

omkarkhatavkar: It should be `satellite-installer`

omkarkhatavkar:
@tstrych  Can you please run the PRT and see if there are any failures related to this change?

tests/foreman/sys/test_katello_certs_check.py
tstrych: good catch thanks :)

tstrych:
@omkarkhatavkar sure, I can run it, but I don't see how it will provide much valid information. This should not change the logic, so if the test fails, it should fail even before.

tstrych:
trigger: test-robottelo
pytest: tests/foreman/api/test_remoteexecution.py tests/foreman/cli/test_puppetplugin.py tests/foreman/cli/test_rhcloud_inventory.py tests/foreman/installer/test_installer.py tests/foreman/longrun/test_remote_DB.py tests/foreman/sys/test_foreman_service.py tests/foreman/sys/test_katello_certs_check.py tests/foreman/ui/test_contenthost.py tests/foreman/ui/test_ldap_authentication.py

mshriver (approved):

PR #9353, opened 2022-02-08
Isolate rhel support, params, and deploy

mshriver:
The content host yaml was trying to do far too much with the
increasingly abstracted test generation for rhel versions.

This creates a new supportability.yaml, which is not a template but a
direct definition of the supported versions of RHEL for the current
master branch of robottelo. This will change when backported to 6.10,
and in future master commits.

Remove handling of the workflow selection during test generation. The
fixture being indirectly parametrized already has a function that deals
with processing host arguments used for deployment.

### Testing

Below is from a CI pipeline build to test this change. The collection of `tests/foreman/api` shows test cases using the rhel version specific content host fixtures and test_subscription.py shows an example of the abstracted `rhel_contenthost` fixture parametrized by the supported rhel versions.

```
 10:00:59  + py.test -rEfs --tb=short --durations=20 --durations-min=600.0 -n 1 --include-stubbed --junit-xml=sat-Critical-results.xml -o junit_suite_name=sat-Critical tests/foreman/api/test_subscription.py::test_positive_subscription_status_disabled
10:01:03  ============================= test session starts ==============================
10:01:03  platform linux -- Python 3.8.8, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
10:01:03  shared_function enabled - OFF - scope:  - storage: file
10:01:03  rootdir: /opt/app-root/src/robottelo, configfile: pyproject.toml
10:01:03  plugins: forked-1.4.0, ibutsu-2.0.2, mock-3.6.1, reportportal-5.0.11, services-2.2.1, xdist-2.5.0
10:01:03  gw0 I
10:01:07  gw0 [4]
10:01:07  
10:58:55  F..F                                                                     [100%]
10:58:55  =================================== FAILURES ===================================
10:58:55  ______________ test_positive_subscription_status_disabled[rhel6] _______________
10:58:55  [gw0] linux -- Python 3.8.8 /opt/app-root/bin/python3.8
10:58:55  tests/foreman/api/test_subscription.py:255: in test_positive_subscription_status_disabled
10:58:55      assert rhel_contenthost.subscribed
10:58:55  E   assert False
10:58:55  E    +  where False = <robottelo.hosts.ContentHost object at 0x7f18a6354b80>.subscribed
10:58:55  ______________ test_positive_subscription_status_disabled[rhel9] _______________
10:58:55  [gw0] linux -- Python 3.8.8 /opt/app-root/bin/python3.8
10:58:55  tests/foreman/api/test_subscription.py:255: in test_positive_subscription_status_disabled
10:58:55      assert rhel_contenthost.subscribed
10:58:55  E   assert False
10:58:55  E    +  where False = <robottelo.hosts.ContentHost object at 0x7f18b21350d0>.subscribed
10:58:55  --- generated xml file: /opt/app-root/src/robottelo/sat-Critical-results.xml ---
10:58:55  ============================= slowest 20 durations =============================
10:58:55  1742.77s setup    tests/foreman/api/test_subscription.py::test_positive_subscription_status_disabled[rhel6]
10:58:55  
10:58:55  (11 durations < 600s hidden.  Use -vv to show these durations.)
10:58:55  =========================== short test summary info ============================
10:58:55  FAILED tests/foreman/api/test_subscription.py::test_positive_subscription_status_disabled[rhel6]
10:58:55  FAILED tests/foreman/api/test_subscription.py::test_positive_subscription_status_disabled[rhel9]
10:58:55  =================== 2 failed, 2 passed in 3470.00s (0:57:49) ===================
```

### DIFF
--- a/pytest_plugins/fixture_markers.py
+++ b/pytest_plugins/fixture_markers.py
@@ -15,7 +15,7 @@ def pytest_generate_tests(metafunc):
             list_params.extend(
                 [
                     setting_rhel_ver
-                    for setting_rhel_ver in settings.content_host.rhel_versions
+                    for setting_rhel_ver in settings.supportability.content_hosts.rhel.versions
                     if str(setting_rhel_ver) in str(matcher)
                 ]
             )
@@ -26,17 +26,15 @@ def pytest_generate_tests(metafunc):
             match_params.extend(
                 [
                     setting_rhel_ver
-                    for setting_rhel_ver in settings.content_host.rhel_versions
+                    for setting_rhel_ver in settings.supportability.content_hosts.rhel.versions
                     if re.fullmatch(str(matcher[0]), str(setting_rhel_ver))
                 ]
             )
         rhel_params = []
         filtered_versions = set(list_params + match_params)
-        for ver, data in settings.content_host.rhel_versions.items():
-            if ver in (filtered_versions or settings.content_host.rhel_versions):
-                # prefer version-specific deploy workflow before using the default one
-                workflow = data.get('deploy_workflow', settings.content_host.deploy_workflow)
-                rhel_params.append(dict(workflow=workflow, rhel_version=ver))
+        # default to all supported versions if no filters were found
+        for ver in filtered_versions or settings.supportability.content_hosts.rhel.versions:
+            rhel_params.append(dict(rhel_version=ver))
         if rhel_params:
             metafunc.parametrize(
                 'rhel_contenthost',
--- a/robottelo/config/validators.py
+++ b/robottelo/config/validators.py
@@ -5,6 +5,9 @@ from robottelo.constants import VALID_GCE_ZONES
 
 
 VALIDATORS = dict(
+    supportability=[
+        Validator('supportability.content_hosts.rhel.versions', must_exist=True, is_type_of=list)
+    ],
     server=[
         Validator('server.hostname', default=''),
         Validator('server.hostnames', must_exist=True, is_type_of=list),
@@ -29,9 +32,8 @@ VALIDATORS = dict(
         Validator('server.ssh_password', default=None),
     ],
     content_host=[
-        Validator('content_host.rhel_versions', must_exist=True),
         Validator('content_host.default_rhel_version', must_exist=True),
-        Validator('content_host.deploy_workflow', must_exist=True),
+        Validator('content_host.deploy_workflow.default', must_exist=True),
     ],
     subscription=[
         Validator('subscription.rhn_username', must_exist=True),
--- a/.gitignore
+++ b/.gitignore
@@ -68,6 +68,7 @@ tests/foreman/pytest.ini
 /settings.local.yaml
 /settings*.local.yaml
 /conf/*.yaml
+!/conf/supportability.yaml
 
 # I don't know where those 2 files come from
 # but they are always there.
--- a/conf/content_host.yaml.template
+++ b/conf/content_host.yaml.template
@@ -1,13 +1,8 @@
 content_host:
-    deploy_workflow: workflow-name
+    deploy_workflow:
+        default:
+        rhel9: # example of workflow override for version
     default_rhel_version: 7
-    rhel_versions:
-      6: {}
-      7: {}
-      8: {}
-      9_compose:
-        deploy_workflow: deploy-rhel-cmp-template
-
     hardware:
         RHEL6:
             RELEASE: 6.10
@@ -21,7 +16,7 @@ content_host:
             RELEASE: 8.3
             MEMORY: 1536 MiB
             CORES: 1
-        RHEL9_compose:
+        RHEL9:
             RELEASE: 9.0
             COMPOSE: RHEL-9.0.0-20220118.1
             MEMORY: 1536 MiB
--- a/conf/supportability.yaml
+++ b/conf/supportability.yaml
@@ -1,0 +1,4 @@
+supportability:
+  content_hosts:
+    rhel:
+      versions: [6, 7, 8, 9]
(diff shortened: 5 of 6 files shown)


conf/supportability.yml
```diff
@@ -0,0 +1,4 @@
+supportability:
```
jyejare, line 1: Why we are having `supportability` YAML instead of being specific for RHEL Versions or just versions (versions.yaml)? What else you are planning to add to this supportability.yaml ?

If not, even this could have been a part of existing content_host.yaml with just one line under content_host header/context:

```
versions: [6, 7, 8, 9]
```

pytest_plugins/fixture_markers.py
```diff
@@ -26,17 +26,15 @@ def pytest_generate_tests(metafunc):
...
             )
         rhel_params = []
         filtered_versions = set(list_params + match_params)
-        for ver, data in settings.content_host.rhel_versions.items():
+        for ver in settings.supportability.content_hosts.rhel.versions:
             if ver in (filtered_versions or settings.content_host.rhel_versions):
-                # prefer version-specific deploy workflow before using the default one
-                workflow = data.get('deploy_workflow', settings.content_host.deploy_workflow)
-                rhel_params.append(dict(workflow=workflow, rhel_version=ver))
+                rhel_params.append(dict(rhel_version=ver))
```
jyejare, line 37: Where does `rhel_version` here is coming from, I dont see any Var declared with it.

OR I am missing something :)

robottelo/config/validators.py
```diff
@@ -29,9 +32,8 @@
         Validator('server.ssh_password', default=None),
     ],
     content_host=[
-        Validator('content_host.rhel_versions', must_exist=True),
         Validator('content_host.default_rhel_version', must_exist=True),
-        Validator('content_host.deploy_workflow', must_exist=True),
+        Validator('content_host.deploy_workflow.default', must_exist=True),
```
jyejare, line 36: If its must exist the we should remove the dummy test from `workflow-name` from the YAML otherwise we will have failed test cases with wrong workflow at least locally as we have a default value for it in CI yaml for content_host.

jyejare (commented):
Overall, the change seems to add cosmetic changes to the YAMLs which I believe we need not to do it unless we have some strong reasons why we need `supportability.yaml` other than rhel versions.

conf/supportability.yml
rplevka: I think one of the goals here is to have the supportability isolated from the rest of the configs, since the way we handle the configs now is that we share them across all robottelo branches (satellite) versions.
However, you can't have rhel9 support in older sat version support matrices.
Also, we agreed that handling this from within a pipeline (by injecting dynaconf vars based on some conditionals) is not really well maintainable, nor readable).

pytest_plugins/fixture_markers.py
mshriver: rhel_version isn't a variable label, its the kwarg to `dict` instantiation.

conf/supportability.yml
mshriver: Supportability can also include things like what versions of hypervisors Satellite supports, but is separately defining what Satellite supports. Not necessarily what will be tested.

This allows for clear logical separation during test generation and collection between what Sat supports and what we want to generate for.

If I just state it as a list of versions in content_host, its more difficult to grow and scale as we do things like support hosts other than RHEL.  It also doesn't provide the logical separation.

robottelo/config/validators.py
mshriver: removed field from the template. ultimately doesn't matter, user needs to make updates to the templates after copying them. If they don't the test fails exactly as expected when the tower workflow call fails. (unless they happened to name their workflow the same as the example).

mshriver:
`Overall, the change seems to add cosmetic changes to the YAMLs which I believe we need not to do it unless we have some strong reasons why we need supportability.yaml other than rhel versions.`

The change is beyond cosmetic.

Within CI it was attempted to inject a data structure into `settings.content_hosts.rhel_version` in order to have rhel 9 present for Sat 7 and missing for Sat 6. This is ultimately not maintainable - and even forces an override of the data structures already existing in the yaml. 

```env.ROBOTTELO_content_host__rhel_versions = "@jinja {{ {6: {}, 7: {}, 8: {}, '9_compose': {'deploy_workflow': 'deploy-rhel-cmp-template'}} if this.robottelo.satellite_version==7.0 else {6: {},7: {},8: {} }}"```

This is forcing too much logic into the configuration layer.

If the yaml has anything other than an empty dict for one of these rhel versions, we have to try and read it before updating it trying to add/remove a rhel version key - otherwise we overwrite what's in the yaml and force assumptions that only rhel 9 needs a custom workflow.

This structure makes it possible to directly inject the necessary customizations to change deploy workflows for specific versions of rhel. It also logically separates the supported rhel versions from data that goes into the deploy arguments for that given rhel version (this is being discussed in a comment chain on this PR).

JacobCallahan (approved):
ACK. Merge pending CI readiness

pytest_plugins/fixture_markers.py
jyejare: Ysah but then Isnt it should be string, or again I m missing something.

conf/supportability.yml
jyejare: @mshriver Satellite really doesn't provisions/differs functionality based on hypervisor versions! Also, Satellite doesn't supports hosts other than RHEL(as you stated above). I commented here because I dont recall anything apart from RHEL versions that need to sit under the `supportability` matrix! Even in CI under `PipelineVars.Groovy` we have a RHEL_Versions as the only supportability matrix there! The max I can think of is Satellite versions apart from RHEL Versions !

But agree that this provides a clear separation between test generation and collection !

jyejare (approved):

rplevka (approved):

mshriver:
Thanks @JacobCallahan @jyejare @rplevka for prompt review. CI changes merged, so I kicked this in

PR #9358, opened 2022-02-08
Extend the puppet disable test case

vsedmik:
Based on this FailedQA https://bugzilla.redhat.com/show_bug.cgi?id=2034552#c3
I'm extending the puppet disable test case (otherwise it would pass).

### DIFF
--- a/tests/foreman/cli/test_puppetplugin.py
+++ b/tests/foreman/cli/test_puppetplugin.py
@@ -88,13 +88,13 @@ def test_positive_enable_disable_logic(destructive_sat, destructive_caps):
     assert_puppet_status(destructive_caps, expected=False)
 
     # Try to enable puppet on Capsule and check it failed.
-    result = destructive_caps.execute(enable_capsule_cmd.get_command(), timeout='15m')
+    result = destructive_caps.execute(enable_capsule_cmd.get_command(), timeout='20m')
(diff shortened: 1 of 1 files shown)


pondrejk (approved):

tests/foreman/cli/test_puppetplugin.py
```diff
@@ -117,11 +117,15 @@ def test_positive_enable_disable_logic(destructive_sat, destructive_caps):
     )
 
     # Disable puppet on Capsule and check it succeeded.
-    result = destructive_caps.execute('foreman-maintain plugin purge-puppet')
+    result = destructive_caps.execute(
+        'foreman-maintain plugin purge-puppet --remove-all-data', timeout='20m'
```
Gauravtalreja1, line 121: Optional: ForemanMaintain robottelo CLI can be implemented for this subcommand and can be used here, with default timeout set to 20m

vsedmik: Yes it can, but it's out of scope of this PR. In this quick fix I just wanted to harden the disable case (plus added timeout as the default one was not enough). We can add the CLI support for `plugin` command elsewhere and update this test after that.

swadeley (approved):
ACK

PR #9365, opened 2022-02-14
update broker to 0.1.34

JacobCallahan:
this change improves the reliability of ssh commands through broker hosts.


swadeley (approved):
ACK

ogajduse (approved):
CI still did not finish, it seems to be stuck. I am not able to merge now.

PR #9371, opened 2022-02-15
[DNM] Downgrade pip to 22.0.2 until dependcy resolved for 22.0.3

jyejare:
DNM


jyejare:
Closing this PR as we have seen the same issue with 22.0.2. I was suspecting the resolver logic change in 22.0.3 could be the cause of it but that's not the reason as per my debug session with @omkarkhatavkar !

The changelog of 22.0.3 :
```
Only calculate topological installation order, for packages that are going to be installed/upgraded.

This fixes an AssertionError that occured when determining installation order, for a very specific combination of upgrading-already-installed-package + change of dependencies + fetching some packages from a package index. This combination was especially common in Read the Docs’ builds.
```

PR #9376, opened 2022-02-16
Mark another BZ tested by this testcase

lhellebr:


tests/foreman/ui/test_computeresource_ec2.py
```diff
@@ -65,6 +65,7 @@ def test_positive_default_end_to_end_with_custom_profile(
     :CaseLevel: Integration
 
     :BZ: 1451626
+    :BZ: 2032530
```
pondrejk, line 68: `:BZ: 1451626, 2032530` is how its done elsewhere, not sure if testimony wouldn't override the first occurrence with the second

pondrejk (changes requested):

tests/foreman/ui/test_computeresource_ec2.py
lhellebr: Good point, thanks!

jyejare (approved):
ACK. Make sure you add `qe-test-coverage +` on that bug.

PR #8391, opened 2021-03-01
Markers to filter tests in new pipeline

jyejare:
This PR includes:

#### 1. Markers to filter following tests in robottelo:

- [x] API
- [x] CLI
- [x] UI
- [x] End To End
- [x] Longrun
- [x] RHAI

#### 2. Pytest Plugin that modifies the collection of these tests based on new following options to select the tests:

- [x] **--include-onprem-provisioning** => Selects on-premises provisioning tests 
- [x] **--include-libvirt** => Selects libvirt based content host tests 
- [x] **--include-external-auth** => Selects External Auth based tests 
- [x] **--include-vlan-networking** => Selects VLAN networking tests 

**Most IMP:** The tests of which inclusion is not selected using above option(s), then those tests will be deselected by default. That also means if the option(s) is not provided to pytest it will be considered as to deselect those tests by default.

PR #9378, opened 2022-02-17
Enabling tests. They are not provisioning. Should never have been disabled.

lhellebr:


ogajduse:
Linking PR that introduced those markers: https://github.com/SatelliteQE/robottelo/pull/8391

lhellebr:
The thing is, these are NOT provisioning tests. They can run on any CR, not necesarrily L2 networked with automation. This can easily run on the old CRs if they are configured (and they seem to be, e.g. [0]).
What I mean by "these are not provisioning tests"? They are tests that e.g. create a CR in Satellite, power-cycle its VM etc. but never create any new VM or a Satellite host. So they are CR tests, not provisioning tests.
Tl;dr, I don't see any reason why these shouldn't run against the CRs they used to run against before. I also ran them locally and most of them pass.
[0]

jyejare:
@lhellebr The CR tests(not provisioning tests) needs CR to connect and unless we deploy those CRs(except Libvirt deployable on sat itself) in real the CR tests are going to fail. You can try testing CR creation manually with dummy CRs details.

Also, for connecting / provisioing with on-prem CRs in old infra, I think we concluded that time (while implementing uncollect plugin) that the sat deployed on new infra wont be able to talk to old infra due to separate networks. 

We were also planned to decommission those CRs at that time and I am not sure what happened to those CRs now. @devendra104 @lpramuk @bherrin3  might know about it !

lhellebr:
As I said, I successfully ran those tests locally against those CRs.
Anyway, marking them as provisioning seems to be too careful since they don't have as high requirements as other provisioning tests. As the owner of these tests, I don't think they should ever have been disabled. Unless they are actually unreachable from automation but I doubt that since they were reachable when I tried locally.

mshriver (approved):

mshriver:
@lhellebr I agree with your assessment here that none of the tests marked by the removed markers are doing any host provisioning. They just depend on an available CR endpoint, and in some cases on existing images and VMs.

PR #9379, opened 2022-02-17
Enabling tests. They are not provisioning. Should never have been disabled.

lhellebr:


lhellebr:
Cherrypick of https://github.com/SatelliteQE/robottelo/pull/9378

PR #9387, opened 2022-02-22
Configure Cloud Connector through UI + assure that CC is not configured before

lhellebr:
This PR does two things:

- It adds a new UI test that configures Cloud Connector through WebUI. Check part is based on the CLI test that does the same.
- Ensures that neither CLI or UI tests are run if there is already Cloud Connector installed. In practice, that shouldn't happen in automation since each testcase has a different importance.

Depends on https://github.com/SatelliteQE/airgun/pull/671

### DIFF
--- a/tests/foreman/ui/test_remoteexecution.py
+++ b/tests/foreman/ui/test_remoteexecution.py
@@ -588,6 +588,8 @@ def test_positive_configure_cloud_connector(
 
     :CaseComponent: RHCloud-CloudConnector
 
+    :CaseImportance: Medium
+
     :assignee: lhellebr
 
     :BZ: 1818076


tests/foreman/ui/test_remoteexecution.py
```diff
@@ -553,3 +566,76 @@ def test_positive_matcher_field_highlight(session):
...
+    # Set Host parameter source_display_name to something random.
+    # To avoid 'name has already been taken' error when run multiple times
+    # on a machine with the same hostname.
+    host_id = Host.info({'name': default_sat.hostname})['id']
+    Host.set_parameter(
+        {'host-id': host_id, 'name': 'source_display_name', 'value': gen_string('alpha')}
+    )
+
+    with session:
+        if session.cloudinventory.is_cloud_connector_configured():
```
omkarkhatavkar, line 609: is there any way you can know this using API or is it UI is the only way to know this ?

tests/foreman/ui/test_remoteexecution.py
```diff
@@ -553,3 +566,76 @@ def test_positive_matcher_field_highlight(session):
...
+            pytest.skip(
+                'Cloud Connector has already been configured on this system. '
+                'It is possible to reconfigure it but then the test would not really '
+                'check if everything is correctly configured from scratch. Skipping.'
+            )
+        session.cloudinventory.configure_cloud_connector()
```
omkarkhatavkar, line 615: if you wanted to configure this from scratch why not adding in fixture (a session/module scope) ensure that will be ran only once in the any worker satellite.

tests/foreman/ui/test_remoteexecution.py
lhellebr: Perhaps not in API but it is possible to execute commands directly on Satellite, e.g. `systemct status ...` to determine whether there is Cloud Connector configured. However, I chose to use the UI approach here since that's the typical workflow, the session needs to be opened anyway and thinking about it, it should also be added in the asserts part - I'll do that.

tests/foreman/ui/test_remoteexecution.py
lhellebr: Because that's what I'm testing, I can't use fixture for the very thing the test should be testing.

omkarkhatavkar (approved):

pondrejk (approved):

PR #9400, opened 2022-03-01
Use host's os_version instead of conf

vsedmik:
@JacobCallahan @mshriver As per offline conversation, replacing OS version from settings by `self.os_version`.


tstrych (approved):
ACK

JacobCallahan (approved):

PR #9406, opened 2022-03-02
Remove unittest2 dependency

JacobCallahan:
This is an old library that is no longer maintained.
Simply importing unittest2 will crash modern python due to changes in
the collections module.


Gauravtalreja1 (approved):

swadeley (approved):
ACK

PR #9409, opened 2022-03-03
Sort generated ids for marker-based content host parametrization

JacobCallahan:
This should help us resolve issues we're having with xdist collection. In the below examples, you can see that rhel8 and rhel7 swap places in test ordering. After the change, 7 comes before 8

before
Screenshot from 2022-03-03 11-11-19

after
Screenshot from 2022-03-03 11-11-07

### DIFF
--- a/pytest_plugins/fixture_markers.py
+++ b/pytest_plugins/fixture_markers.py
@@ -36,6 +36,7 @@ def pytest_generate_tests(metafunc):
         for ver in filtered_versions or settings.supportability.content_hosts.rhel.versions:
             rhel_params.append(dict(rhel_version=ver))
         if rhel_params:
+            rhel_params.sort(key=lambda r: str(r['rhel_version']))
             metafunc.parametrize(
                 'rhel_contenthost',
                 rhel_params,


pytest_plugins/fixture_markers.py
```diff
@@ -39,7 +39,7 @@ def pytest_generate_tests(metafunc):
             metafunc.parametrize(
                 'rhel_contenthost',
                 rhel_params,
-                ids=[f'rhel{r["rhel_version"]}' for r in rhel_params],
+                ids=sorted(f'rhel{r["rhel_version"]}' for r in rhel_params),
```
mshriver, line 42: I think this leaves us open for rhel_params and the ids list to end up out of order with each other.

The params list should be sorted, and then the ids picked from the sorted list, so that when the parameters and their IDs are zipped together they don't get out of order.

```
In [31]: rhel_params = [{'rhel_version': '7', 'rhel': '7.6'}, {'rhel_version': '8', 'rhel': '8.5'}, {'rhel_version': '7_fips', 'rhel': '7 fips'}, {'rhel_version': '8_fips', 'rhel': '8 fips'}]

In [32]: for i in zip(rhel_params, sorted([f"rhel{r['rhel_version']}" for r in rhel_params])):
    ...:     print(i)
    ...: 
({'rhel_version': '7', 'rhel': '7.6'}, 'rhel7')
({'rhel_version': '8', 'rhel': '8.5'}, 'rhel7_fips')
({'rhel_version': '7_fips', 'rhel': '7 fips'}, 'rhel8')
({'rhel_version': '8_fips', 'rhel': '8 fips'}, 'rhel8_fips')
```

Versus if we sort the param list first:

```
In [31]: rhel_params = [{'rhel_version': '7', 'rhel': '7.6'}, {'rhel_version': '8', 'rhel': '8.5'}, {'rhel_version': '7_fips', 'rhel': '7 fips'}, {'rhel_version': '8_fips', 'rhel': '8 fips'}]

In [32]: for i in zip(rhel_params, sorted([f"rhel{r['rhel_version']}" for r in rhel_params])):
    ...:     print(i)
    ...: 
({'rhel_version': '7', 'rhel': '7.6'}, 'rhel7')
({'rhel_version': '8', 'rhel': '8.5'}, 'rhel7_fips')
({'rhel_version': '7_fips', 'rhel': '7 fips'}, 'rhel8')
({'rhel_version': '8_fips', 'rhel': '8 fips'}, 'rhel8_fips')

```

Maybe its still possible for xdist workers to collect in different orders, in which case we could be extra sure and sort the ids list *again* - but in that case it will always be the same order as the parameter values are parametrized in.

mshriver (approved):

PR #9412, opened 2022-03-04
Cockpit rhel host parametrization

peterdragun:
~~Merge after: SatelliteQE/airgun/pull/676~~ :heavy_check_mark: 

Adding parametrization for cockpit content host testing.

### DIFF
--- a/pytest_fixtures/core/contenthosts.py
+++ b/pytest_fixtures/core/contenthosts.py
@@ -107,21 +107,45 @@ def registered_hosts(organization_ak_setup, content_hosts, default_sat):
 @pytest.fixture(scope="function")
 def katello_host_tools_host(default_sat, module_org, rhel_contenthost):
     """Register content host to Satellite and install katello-host-tools on the host."""
+    repo = settings.repos['SATCLIENT_REPO'][f'RHEL{rhel_contenthost.os_version.major}']
+    register_host_custom_repo(default_sat, module_org, rhel_contenthost, [repo])
+    rhel_contenthost.install_katello_host_tools()
+    yield rhel_contenthost
+
+
+@pytest.fixture(scope="function")
+def cockpit_host(default_sat, module_org, rhel_contenthost):
+    """Register content host to Satellite and install cockpit on the host."""
+    rhelver = rhel_contenthost.os_version.major
+    if rhelver > 7:
+        repo = [settings.repos[f'rhel{rhelver}_os']['baseos']]
+    else:
+        repo = [settings.repos['rhel7_os'], settings.repos['rhel7_extras']]
+    register_host_custom_repo(default_sat, module_org, rhel_contenthost, repo)
+    rhel_contenthost.execute(f"hostnamectl set-hostname {rhel_contenthost.hostname} --static")
+    rhel_contenthost.install_cockpit()
+    rhel_contenthost.add_rex_key(satellite=default_sat)
+    yield rhel_contenthost
+
+
+def register_host_custom_repo(default_sat, module_org, rhel_contenthost, repo_urls):
+    """Register content host to Satellite and sync repos"""
     # prepare Product and appropriate Satellite Client repo on satellite
     rhelver = rhel_contenthost.os_version.major
     prod = entities.Product(
         organization=module_org, name=f'rhel{rhelver}_{gen_string("alpha")}'
     ).create()
-    sattools_repo = entities.Repository(
-        organization=module_org,
-        product=prod,
-        content_type='yum',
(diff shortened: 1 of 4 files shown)


peterdragun:
trigger: test-robottelo
pytest: tests/foreman/ui/test_host.py -k "test_positive_cockpit[rhel7] or test_positive_cockpit[rhel8]"
airgun: 676

pytest_fixtures/core/contenthosts.py
```diff
@@ -107,21 +107,45 @@ def registered_hosts(organization_ak_setup, content_hosts, default_sat):
...
         organization=module_org, name=f'rhel{rhelver}_{gen_string("alpha")}'
     ).create()
-    sattools_repo = entities.Repository(
-        organization=module_org,
-        product=prod,
-        content_type='yum',
-        url=settings.repos['SATCLIENT_REPO'][f'RHEL{rhelver}'],
-    ).create()
-    sattools_repo.sync()
+    for url in repo_urls:
```
tstrych, line 138: why do you need  repo_urls here ?
isn't the one repo per version enough ?

In the previous version you enable one repo per version, 
you don't need more of them right ? for one rhel version one custom satclient repo

robottelo/hosts.py
```diff
@@ -413,6 +413,16 @@ def install_capsule_katello_ca(self, capsule=None):
         if result.status != 0:
             raise ContentHostError('Failed to install the katello-ca rpm')
 
+    def install_cockpit(self):
+        """Installs cockpit on the broker virtual machine.
```
tstrych, line 417: Install cockpit on the Content Host.

(no need for broker virtual machine)

tstrych (approved):
ACK, pending comments.

please also provide test Result for RHEL9.

pytest_fixtures/core/contenthosts.py
peterdragun: The cockpit on RHEL 7 requires additional extras, but those need base os RPMs to install correctly.

robottelo/hosts.py
peterdragun: Just wanted to use a similar docstring to other functions in this module

pytest_fixtures/core/contenthosts.py
```diff
@@ -107,21 +107,45 @@ def registered_hosts(organization_ak_setup, content_hosts, default_sat):
...
+@pytest.fixture(scope="function")
+def cockpit_host(default_sat, module_org, rhel_contenthost):
+    """Register content host to Satellite and install cockpit on the host."""
+    rhelver = rhel_contenthost.os_version.major
+    if rhelver > 7:
+        repo = [settings.repos[f'rhel{rhelver}_os']['baseos']]
+    else:
+        repo = [settings.repos['rhel7_os'], settings.repos['rhel7_extras']]
+    register_host_custom_repo(default_sat, module_org, rhel_contenthost, repo)
+    rhel_contenthost.execute(f"hostnamectl set-hostname {rhel_contenthost.hostname} --static")
```
pondrejk, line 125: just curious, why this is needed? Isn't this pointing to some problem at satlab side?

tests/foreman/ui/test_host.py
```diff
@@ -2318,9 +2318,10 @@ def test_positive_gce_cloudinit_provision_end_to_end(
 
 @pytest.mark.destructive
 @pytest.mark.upgrade
+@pytest.mark.rhel_ver_match('[^6].*')
```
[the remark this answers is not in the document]
pondrejk: In this case I think it makes sense to include fips hosts to such integration tests. If we checked only for the webconsole button then no, but the actual communication between host and the UI could possibly be influenced by stricter fips rules, so I think it's good to have that coverage

pondrejk (approved):

pytest_fixtures/core/sys.py
```diff
@@ -1,6 +1,7 @@
 from urllib.parse import urlparse
 
 import pytest
+from nailgun import entities
```
Gauravtalreja1, line 4: Instead of importing nailgun here, you could call like `default_sat.cli.Setting().###`

mshriver: @Gauravtalreja1 close! To keep things on the api, use `default_sat.api.Setting` to reference the same class from `nailgun.entities.Setting`.

pytest_fixtures/core/contenthosts.py
peterdragun: The cockpit is using static hostname in UI, so without this command, it is reporting localhost. Not sure if we want to set the static hostname on all hosts from satlab. Without this command transient hostname is still set to the correct value.

PR #9425, opened 2022-03-08
Fix several Repo Tests

sambible:
1) This request isn't throwing an SSL error, so the test fails. Checking for a 403 error, which is returned, seems to capture the same behavior, and works.

2) Changing some regex messages to match what is actually being returned, while keeping the intent of the test.

### DIFF
--- a/tests/foreman/api/test_repository.py
+++ b/tests/foreman/api/test_repository.py
@@ -30,7 +30,6 @@ from nailgun import client
 from nailgun import entities
 from nailgun.entity_mixins import TaskFailedError
 from requests.exceptions import HTTPError
-from requests.exceptions import SSLError
 
 from robottelo import constants
 from robottelo import datafactory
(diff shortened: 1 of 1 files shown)


JacobCallahan (approved):

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/api/test_repository.py -k test_positive_access_protected_repository

jameerpathan111:
> 1. This request isn't throwing an SSL error, so the test fails. Checking for a 403 error, which is returned, seems to capture the same behavior, and works.

@sambible Did you check with devs if it's expected? If it was throwing a proper error message before and not now, then I think we need to report this issue.

jameerpathan111 (approved):

mshriver (approved):
Thanks for the investigation and fix @sambible. @jameerpathan111 good question(s) about the context of the updates. Before it was actually relying on nailgun to raise an exception, so its possible nailgun changed here in addition to Satellite itself.

PR #9435, opened 2022-03-16
[Merge with CI PR]  Secrets from vault via DynaConf

jyejare:
**The PR integrates Hashicorp Vault as a secret store via DynaConf.**

The PR is dependent on following 2 DynaConf Upstream PRs which are now merged:

- [x] https://github.com/rochacbruno/dynaconf/pull/725 
- [x] https://github.com/rochacbruno/dynaconf/pull/729 

And CI MR (Need to merge along with this PR):
- [x] CI changes (satelliteqe-jenkins !570)

Tests Results:
----------------

Check Critical Tests : https://{stage_url}/job/jyejare-6.11-rhel7-critical/2
Check High Tests: https://{stage_url}/job/jyejare-6.11-rhel7-high/1

### DIFF
--- a/.env
+++ b/.env
@@ -1,1 +1,0 @@
-# This file is here for dynaconf to set this as root level directory
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@
 broker==0.1.35
 cryptography==36.0.2
 deepdiff==5.8.0
-dynaconf==3.1.7
+dynaconf[vault]==3.1.8
 fauxfactory==3.1.0
 Inflector==3.0.1
 jinja2==3.1.1
--- a/robottelo/config/__init__.py
+++ b/robottelo/config/__init__.py
@@ -28,6 +28,7 @@ def get_settings():
         includes=["settings.local.yaml", ".secrets.yaml", ".secrets_*.yaml"],
         envless_mode=True,
         lowercase_read=True,
+        load_dotenv=True,
     )
     settings.validators.register(**VALIDATORS)
 
--- a/robottelo/config/validators.py
+++ b/robottelo/config/validators.py
@@ -88,12 +88,20 @@ VALIDATORS = dict(
     ],
     container_repo=[
         Validator(
-            'container_repo.registry_url',
-            'container_repo.registry_username',
-            'container_repo.registry_password',
-            'container_repo.repos_to_sync',
+            'container_repo.registries.redhat.url',
+            'container_repo.registries.redhat.username',
+            'container_repo.registries.redhat.password',
+            'container_repo.registries.redhat.repos_to_sync',
+            'container_repo.registries.redhat.long_pass',
             must_exist=True,
-        )
+        ),
+        Validator(
+            'container_repo.registries.quay.url',
+            'container_repo.registries.quay.username',
+            'container_repo.registries.quay.password',
+            'container_repo.registries.quay.repos_to_sync',
+            must_exist=True,
+        ),
     ],
     discovery=[Validator('discovery.discovery_iso', must_exist=True)],
     distro=[
@@ -114,15 +122,10 @@ VALIDATORS = dict(
             'docker.private_registry_username',
             'docker.private_registry_password',
             must_exist=True,
-        )
+        ),
     ],
     ec2=[
-        Validator(
-            'ec2.access_key',
-            'ec2.secret_key',
-            'ec2.region',
-            must_exist=True,
-        ),
+        Validator('ec2.access_key', 'ec2.secret_key', 'ec2.region', must_exist=True),
         Validator('ec2.managed_ip', is_in=('Private', 'Public'), default='Private'),
         Validator('ec2.region', default='us-west-2'),
         Validator('ec2.security_group', default=['default']),
@@ -149,6 +152,7 @@ VALIDATORS = dict(
         Validator(
             'git.username',
             'git.password',
+            'git.github_token',
             'git.ssh_port',
             'git.http_port',
             'git.hostname',
@@ -162,23 +166,26 @@ VALIDATORS = dict(
             'http_proxy.username',
             'http_proxy.password',
             must_exist=True,
-        )
+        ),
     ],
     ipa=[
         Validator(
-            'ipa.basedn_ipa',
-            'ipa.grpbasedn_ipa',
-            'ipa.hostname_ipa',
-            'ipa.password_ipa',
-            'ipa.username_ipa',
-            'ipa.user_ipa',
+            'ipa.hostname',
+            'ipa.username',
+            'ipa.password',
+            'ipa.idm_password',
+            'ipa.basedn',
+            'ipa.grpbasedn',
+            'ipa.user',
             'ipa.otp_user',
-            'ipa.time_based_secret',
             'ipa.disabled_ipa_user',
             'ipa.group_users',
             'ipa.groups',
+            'ipa.idm_server_ip',
+            'ipa.keytab_url',
+            'ipa.time_based_secret',
             must_exist=True,
-        )
+        ),
     ],
     ldap=[
         Validator(
@@ -186,23 +193,23 @@ VALIDATORS = dict(
             'ldap.grpbasedn',
             'ldap.hostname',
             'ldap.nameserver',
-            'ldap.password',
             'ldap.realm',
             'ldap.username',
+            'ldap.password',
             'ldap.workgroup',
             must_exist=True,
-        )
+        ),
     ],
     open_ldap=[
         Validator(
             'open_ldap.base_dn',
             'open_ldap.group_base_dn',
             'open_ldap.hostname',
-            'open_ldap.password',
             'open_ldap.username',
+            'open_ldap.password',
             'open_ldap.open_ldap_user',
             must_exist=True,
-        )
+        ),
     ],
     oscap=[
         Validator(
@@ -216,15 +223,11 @@ VALIDATORS = dict(
             'osp.username',
             'osp.password',
             'osp.tenant',
-            'osp.project_domain_id',
             'osp.security_group',
             'osp.vm_name',
-            'osp.image_os',
-            'osp.image_arch',
-            'osp.image_username',
-            'osp.image_name',
+            'osp.project_domain_id',
             must_exist=True,
-        )
+        ),
     ],
     performance=[Validator('performance.time_hammer', default=False)],
     report_portal=[
@@ -236,6 +239,7 @@ VALIDATORS = dict(
         ),
         Validator('report_portal.fail_threshold', default=20),
     ],
+    rh_cloud=[Validator('rh_cloud.token', required=True)],
     repos=[
         Validator(
             'repos.rhel6_os',
@@ -251,7 +255,7 @@ VALIDATORS = dict(
             'repos.rhscl_repo',
             'repos.ansible_repo',
             'repos.swid_tools_repo',
-            must_exist=False,
+            must_exist=True,
             is_type_of=str,
         ),
     ],
@@ -267,26 +271,28 @@ VALIDATORS = dict(
             'rhev.image_arch',
             'rhev.image_username',
             'rhev.image_password',
+            'rhev.ca_cert',
             'rhev.image_name',
             must_exist=True,
-        )
+        ),
     ],
     rhsso=[
         Validator(
             'rhsso.host_name',
             'rhsso.host_url',
             'rhsso.rhsso_user',
-            'rhsso.user_password',
+            'rhsso.rhsso_password',
             'rhsso.realm',
+            'rhsso.totp_secret',
             must_exist=True,
-        )
+        ),
     ],
     remotedb=[
         Validator(
             'remotedb.server',
             'remotedb.db_server',
-            'remotedb.common_db_password',
-            must_exist=False,
+            'remotedb.db_password',
+            must_exist=True,
         ),
         Validator('remotedb.foreman.username', default='foreman'),
         Validator('remotedb.foreman.db_name', default='foreman'),
@@ -357,6 +363,6 @@ VALIDATORS = dict(
             'vmware.image_password',
             'vmware.image_name',
             must_exist=True,
-        )
+        ),
     ],
 )
--- a/scripts/vault_login.py
+++ b/scripts/vault_login.py
@@ -1,0 +1,109 @@
+#!/usr/bin/env python
+# This Enables and Disables individuals OIDC token to access secrets from vault
+import json
+import os
+import re
+import subprocess
+import sys
+import time
+
+from robottelo.errors import InvalidVaultURLForOIDC
+
+HELP_TEXT = (
+    "Vault CLI in not installed in your system, "
+    "refer link https://learn.hashicorp.com/tutorials/vault/getting-started-install to "
+    "install vault CLI as per your system spec!"
+)
+
+# Color Codes
+YELLOW = '\033[1;33m'
+REDL = '\033[3;31m'
+REDD = '\033[1;31m'
+GREEN = '\033[1;32m'
+WHITEL = '\033[1;30m'
+
+
+def _load_env():
+    print("\nFetching vault data from .env file and Exporting vault URL .....\n")
+    ROOT_PATH = os.path.realpath(
+        os.path.join(os.path.dirname(os.path.abspath(__file__)), os.path.pardir)
+    )
+    with open(ROOT_PATH + '/.env') as envfile:
+        envdata = envfile.read()
+    return ROOT_PATH, envdata
+
+
+def _export_vault_url(envdata):
+    vaulturl = re.findall('VAULT_URL_FOR_DYNACONF=(.*)', envdata)[0]
+    if 'localhost:8200' in vaulturl:
+        raise InvalidVaultURLForOIDC(
+            f"{REDL}{vaulturl} doesnt supports OIDC login,"
+            "please change url to corp vault in env file!"
+        )
+    os.environ['VAULT_ADDR'] = vaulturl
+
+
+def _vault_command(command: str):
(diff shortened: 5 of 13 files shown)


jyejare:
Both the DynaConf PRs are now merged :)

jyejare:
@SatelliteQE/robottelo-tier-1-reviewers and @SatelliteQE/robottelo-tier-2-reviewers The PR is ready for review but hold merging for CI PR.

robottelo/config/validators.py
```diff
@@ -1,10 +1,75 @@
 from dynaconf import Validator
 
+from robottelo.config.vault import secrets
 from robottelo.constants import AZURERM_VALID_REGIONS
 from robottelo.constants import VALID_GCE_ZONES
 
 
 VALIDATORS = dict(
+    # Loading secrets from vault at the top,
+    # so that bulk validation fail in some setting wont skip further secrets loading
```
lhellebr, line 10: How would this happen? Wouldn't it only fail for that certain section?

lhellebr (approved):
This needs to be tested but LGTM

robottelo/config/validators.py
JacobCallahan: validation currently happens on the entire settings object and once a validation error is encountered, it stops

robottelo/config/validators.py
```diff
@@ -1,10 +1,75 @@
...
+            'virtwho.rhevm.hypervisor_password', default=secrets.get('rhevm_hypervisor_password')
+        ),
+        Validator(
+            'virtwho.libvirt.hypervisor_password',
+            default=secrets.get('libvirt_hypervisor_password'),
+        ),
+        Validator(
+            'virtwho.kubevirt.hypervisor_password',
+            default=secrets.get('kubevirt_hypervisor_password'),
+        ),
```
jyejare, line 68: @JacobCallahan @mshriver I think we should remove these now ?

mshriver: There are still references to these secrets within the test framework, can't remove them until we actually remove those tests, which won't be happening anytime soon.

.env
```diff
@@ -1 +1,23 @@
...
+# VAULT_SECRET_ID_FOR_DYNACONF=
+# Authenticate with AWS IAM https://www.vaultproject.io/docs/auth/aws
+# The IAM Credentials can be retrieved from the standard providers: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html
+# VAULT_AUTH_WITH_IAM_FOR_DYNACONF=False
+# VAULT_AUTH_ROLE_FOR_DYNACONF=vault-role
+# Authenticate with root token
+# VAULT_ROOT_TOKEN_FOR_DYNACONF=root-token
+# Path of secrets
+VAULT_MOUNT_POINT_FOR_DYNACONF=apps
+VAULT_PATH_FOR_DYNACONF=mna-satqe-hashicorp/robottelo
```
mshriver, line 23: I think this should be far more generic for this field.

requirements.txt
```diff
@@ -4,7 +4,9 @@
 broker==0.1.35
 cryptography==36.0.2
 deepdiff==5.7.0
-dynaconf==3.1.7
+# dynaconf==3.1.7 <- The latest dynaconf for vault changes is not released, using git based
+git+https://github.com/rochacbruno/dynaconf.git@master#egg=dynaconf
```
mshriver, line 8: Have you requested a release with those updates? I'm hesitant to go with unreleased master branch as our dependency.

robottelo/config/validators.py
```diff
@@ -88,12 +147,18 @@
     ],
     container_repo=[
         Validator(
-            'container_repo.registry_url',
-            'container_repo.registry_username',
-            'container_repo.registry_password',
-            'container_repo.repos_to_sync',
+            'container_repo.registries.redhat.url',
```
mshriver, line 150: Thanks for cleaning this up!

robottelo/config/validators.py
```diff
@@ -1,10 +1,75 @@
...
+            'virtwho.libvirt.hypervisor_password',
+            default=secrets.get('libvirt_hypervisor_password'),
+        ),
+        Validator(
+            'virtwho.kubevirt.hypervisor_password',
+            default=secrets.get('kubevirt_hypervisor_password'),
+        ),
+        Validator('vmware.username', default=secrets.get('vmware_username')),
+        Validator('vmware.password', default=secrets.get('vmware_password')),
+        Validator('vmware.image_password', default=secrets.get('satqe_shared_password')),
```
mshriver, line 71: This is hardcoding all of the vault keys into source here, this is a huge downside to the approach of using validator defaults.

.env
jyejare: Didnt get the comment :) Can you elaborate on this ?

robottelo/config/validators.py
jyejare: Agree with hardcoding but I dont see how that could be a downside? 
The hardcoding in validators is the same as the hardcoded names of setting in validators!

Everyone in the team has access to secrets so shouldn't be an issue to access, and update the secrets.

jyejare:
@lhellebr The test results are available in Stage Jenkins:

Check Critical Tests : https://{stage_url}/job/jyejare-6.11-rhel7-critical/2
Check High Tests: https://{stage_url}/job/jyejare-6.11-rhel7-high/1

.env
mshriver: This file in our upstream open-source repo should not contain the vault path that we use internally.

robottelo/config/validators.py
mshriver: There is a big difference between the two, and its not about access/updating.

The names of the settings map directly to the yaml files and data structure within them. `vmware.image_password` as an argument for the validator is dictating both the structure of the configuration yaml and the acceptable values for a given field.

This structure that you've proposed now hardcodes the paths within vault, in our logic code, instead of within the config files themselves. Users of robottelo shouldn't have to modify this module to not use `satqe_shared_password` as their vault key, and they shouldn't be forced to use a key like `satqe_shared_password`.

Assuming I want to change the credential for this `vmware.image_password` field. It is now hardcoded in source as `satqe_shared_password`, and I can't update vault without impacting every other field mapped to this vault key, and/or updating source code.

This is violating the 12 factor app principles, which is a big reason for why we adopted dynaconf. Config should be cleanly and clearly separated from code, with the only 'overlap' being these validator paths and conf file location, because the config can't be completely arbitrary. 

@JacobCallahan and I discussed the option of trying to use dynaconf dynamic variable formatting for this, using `ssh_password: '@format {this.secrets.ssh_password}'` within the yaml file, and letting the dynaconf instance for `settings` object resolve from its own secrets loading, instead of creating two dynaconf instances. 

This isn't my decision alone to make, and I've thoroughly voiced my issues with this approach and have suggested the approach of using a yamlLoader to proxy these lookups like IQE has done.  If you and @JacobCallahan have heard these warnings and would like to proceed with this implementation, I won't stop you. But I think its setting us up for problems as-is in this PR.

.env
```diff
@@ -1 +1,23 @@
 # This file is here for dynaconf to set this as root level directory
+
```
mshriver, line 2: There are no docs included guiding users on what to do for local configuration of vault.

This file should be a template, or this block dumped by vault_login script if not present.

The user can't modify this file without source control seeing it. The contents here should be in an example file, so that the user can have a `.env` file locally.

robottelo/config/__init__.py
```diff
@@ -15,7 +15,7 @@
     os.environ['ROBOTTELO_DIR'] = str(robottelo_root_dir)
 
 
-settings = LazySettings(
+settings = Dynaconf(
```
mshriver, line 18: I think we should continue to explicitly use LazySettings

requirements.txt
mshriver: I'd say your fork is a better option than master, neither are as good as a release.

These requirements alone don't seem to be giving me enough to run `vault_login` script, as the executable for vault isn't included. 

Are you expecting users to have added the hashicorp repo and installed via DNF? 
https://learn.hashicorp.com/tutorials/vault/getting-started-install

.env
jyejare: Cool! This can be moved to example file, there is no harm though !

requirements.txt
jyejare: Yes! I am expecting people to install vault themselves, depending on their OS. The installation differs for different OSes.

robottelo/config/__init__.py
jyejare: Its mentioned here: https://github.com/rochacbruno/dynaconf/blob/master/dynaconf/__init__.py#L19

robottelo/config/validators.py
jyejare: @mshriver RE:

> This structure that you've proposed now hardcodes the paths within vault, in our logic code, instead of within the config files themselves. Users of robottelo shouldn't have to modify this module to not use satqe_shared_password as their vault key, and they shouldn't be forced to use a key like satqe_shared_password.

Users can still use config files to not use secrets from vault, by just updating the setting in conf. infact,  version controlling the vault secret keys with validators will allow us to monitor the secret names changes in vault PnT store and hence we will have better control over modifying secrets from Sat rover group in the vault.

> Assuming I want to change the credential for this vmware.image_password field. It is now hardcoded in source as satqe_shared_password, and I can't update vault without impacting every other field mapped to this vault key, and/or updating source code.

Again, this comes under the version controlling the change in secrets and a better control as mentioned in my previous comment. Updating the vault for specific secret credentials and its impact on other field mapped to the key isnt an issue with the hardcoded approach itself. Its the issue of using same credentials for multiple secrets and thats issue even now with OCP secrets. To resolve it, we can have 1:1 mapping with secret and its credential even though the credentials are the same for multiple secrets.

> This is violating the 12 factor app principles, which is a big reason for why we adopted dynaconf. Config should be cleanly and clearly separated from code, with the only 'overlap' being these validator paths and conf file location, because the config can't be completely arbitrary.

TBH, I am not a fan of principles unless the refactor / approach makes things easier to access, monitor and control the credentials from Vault. 

> @JacobCallahan and I discussed the option of trying to use dynaconf dynamic variable formatting for this, using ssh_password: '@format {this.secrets.ssh_password}' within the yaml file, 

What I dont like the YAML approach is the user has to remember the format and the entire path e.g `this.secrets.ssh_password` in case the user updated the conf locally with some credential string and now wants to revert back so that the user can access secret from the vault. This way we are forcing users to remember the path or every time go and look into CI to pick copy the path to access secrets instead of automation(now Validators) doing it for us. 

> and letting the dynaconf instance for settings object resolve from its own secrets loading, instead of creating two dynaconf instances.

This could be the only reason I would go with the YAML approach but anyways they are two different objects of dynaConf settings and shouldn't be interfering with each other.

jyejare: The changes were made as requested after @JacobCallahan convinced me on some security aspects of it!

requirements.txt
mshriver: 3.1.8 is out (and there's a pyup bot PR available with it) so this change can be dropped.

mshriver: @jyejare thanks for pushing for the dynaconf release!

robottelo/config/__init__.py
mshriver: I don't see there that `LazySettings` is or will be deprecated, just that its the default alias.

That's why we use it explicitly here. If Dynaconf upstream wants to change the default alias, our object would change without our knowledge.

scripts/vault_login.py
```diff
@@ -0,0 +1,100 @@
...
+        print("\033[1;32mSuccess! OIDC token removed from Env file successfully!")
+
+
+def _vault_status():
+    vstatus = _vault_command('vault token lookup')
+    if vstatus.returncode == 0:
+        print(str(vstatus.stdout.decode('UTF-8')))
+
+
+if __name__ == '__main__':
```
mshriver, line 90: I feel like this is written in bash, in python :) 

Click is really good for creating extensible command structures. This file is largely a convenience - the only variable necessary to be able to use vault is the token, and to be able to use `vault` on the command line the URL.

I don't want us mirroring the vault CLI instead of using both `VAULT_ADDR` and `VAULT_URL_FOR_DYNACONF` in the `.env.example` file. 

That file is read by dynaconf, but it also can be sourced in one's local env in order to just use `vault` CLI directly and have the URL set.

Having a tool that makes integrating with the vault CLI is helpful, but I'm not sure this is the right approach. Using python's entrypoints and command groups implemented through click creates a far more extensible interface for doing things like this to 'work' with the framework and setup a testing environment.

scripts/vault_login.py
```diff
@@ -0,0 +1,100 @@
...
+    print("\033[1;32mSuccess! New OIDC token is added to .env file to access secrets from vault!")
+
+
+def _vault_logout(root_path, envdata):
+    # Teardown - Setting dymmy token in env file
+    envdata = re.sub('.*VAULT_TOKEN_FOR_DYNACONF=.*', "# VAULT_TOKEN_FOR_DYNACONF=myroot", envdata)
+    with open(root_path + '/.env', 'w') as envfile:
+        envfile.write(envdata)
+    vlogout = _vault_command('vault token revoke -self')
+    if vlogout.returncode == 0:
```
mshriver, line 80: These checks against returncode are redundant, `_vault_command` will `sys.exit()` if its non-zero.

scripts/vault_login.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+    os.environ['VAULT_ADDR'] = vaulturl
+
+
+def _vault_command(command: str):
+    vcommand = subprocess.run(command, capture_output=True, shell=True)
+    if vcommand.returncode != 0:
+        verror = str(vcommand.stderr)
+        if 'vault: command not found' in verror:
+            print(f"\033[1;31mError! {HELP_TEXT}")
```
mshriver, line 44: These are effectively 'magic' numbers and are really annoying to work with. constants should be defined to describe what they are.

scripts/vault_login.py
```diff
@@ -0,0 +1,100 @@
...
+        elif 'Error revoking token' in verror:
+            print("\033[1;32mToken is alredy revoked!")
+        elif 'Error looking up token' in verror:
+            print("\033[1;33mWarning! Vault not logged in, please run 'make vault-login'!")
+        else:
+            print(f"\033[1;31mError! {verror}")
+        sys.exit(0)
```
mshriver, line 51: 0 exit code is for a successful result, I'd suggest using something else

scripts/vault_login.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def _vault_login(root_path, envdata):
+    print(
+        "\033[1;30mWarning! The browser is about to open for vault OIDC login, "
+        "close the tab once the sign in is done!"
+    )
+    time.sleep(5)
+    if _vault_command(command="vault login -method=oidc").returncode == 0:
+        os.system("vault token renew -i 10h")
```
mshriver, line 62: Why use `os.system` here and `subprocess.run` above?

scripts/vault_login.py
```diff
@@ -0,0 +1,100 @@
...
+        sys.exit(0)
+    return vcommand
+
+
+def _vault_login(root_path, envdata):
+    print(
+        "\033[1;30mWarning! The browser is about to open for vault OIDC login, "
+        "close the tab once the sign in is done!"
+    )
+    time.sleep(5)
```
mshriver, line 60: You really should use `input` here instead of a static 5s sleep. it could easily take the user longer than that to complete OIDC if its not already up.

mshriver (changes requested):
I think the vault login script needs some discussion, but the rest is looking good - lots of cleanup to the validators, thanks!

scripts/vault_login.py
jyejare: Ofcorse there are multiple ways of implementing things, but since this is just a helper module to be used with `Make`, I kept it simple! 

I would not clock the PR on this helper module improvements/re-implementation, we can merge this PR and later one can raise a PR to improve it !

scripts/vault_login.py
jyejare: Removed the check!

scripts/vault_login.py
jyejare: Not a blocker for this PR and defining constants won't be reused outside of this module! Lets keep it simple!

scripts/vault_login.py
jyejare: Replaced with 1, mistakenly kept it !

scripts/vault_login.py
jyejare: Forgot to replace :)

Updated !

scripts/vault_login.py
jyejare: The browser will open post this sleep. The sleep is for a user to not surprise why the browser is opened or just to inform that `vault login -method=oidc` will open the browser ! 

The vault login is not async , so the execution wont proceed to the next line until it finishes OIDC login.

scripts/vault_login.py
jyejare: But still kept updated :)

jyejare:
@mshriver Pushed changes to vault login script !

scripts/vault_login.py
mshriver: I would consider this a blocker. These formatting codes are used many times in this module, and make it kinda hard to read what the messages are, and what formatting is being applied.

jyejare: @mshriver I updated the code already!

mshriver (approved):
I've expressed some concerns with the structure of the bash scripts for vault login integration, but am not going to block merge on that.

Thanks @jyejare for being willing to pivot away from the initial proposed implementation to address my concerns with hardcoding paths into the validators!

mshriver:
@jyejare time for a backport to 6.10.z!

PR #9445, opened 2022-03-23
Remove 'Debian' condition and removing it from list

latran:
The bz 1709683 is in a wontfix status so it's best to remove debian and remove the condition. This aligns with: https://github.com/SatelliteQE/nailgun/pull/822


JacobCallahan (approved):

latran:
Closing this now.  The skip function works after #9449

PR #9463, opened 2022-03-29
Improve validation script to add more context to Validation and Attribute errors

spusateri:
This PR clarifies the output of the validation script, by determining the type of error and printing the `.name` of the validator that failed in each file.


jyejare:
@JacobCallahan By removing from reviewer pool, you mean you are working on dynaconf PR to have an individual setting validation? And if yes, we dont need this PR ?

JacobCallahan:
@jyejare no, I mean I was there when this was written and tested. Others should review this since I basically already have.

scripts/validate_config.py
```diff
@@ -1,17 +1,38 @@
...
-from robottelo.config import settings
+try:
+    from robottelo.config import settings
+except Exception as err:
+    print(f"Encountered the following exception, continuing for the sake of validation\n {err}")
+
+    from dynaconf import LazySettings
+    from robottelo.config.validators import VALIDATORS
+
+    settings = LazySettings(
```
mshriver, line 12: Not liking that this is duplicating the code in the config module itself, and then becomes sensitive to any changes that we make there.

I think its worth taking this a step farther, and creating a function in `robottelo/config/__init__.py` to return the raw settings object before validation.

Then `robottelo/config/__init__.py` can still create a module label `settings` and validate it, so existing imports don't change.  This validation script can use the function to get the same LazySettings instance as the rest of the framework, just without being validated before import.

mshriver: We would not. Within `robottelo.config` we would still define `settings`, just by calling the function that's been defined. No imports would change.

jyejare: Ping @spusateri !

spusateri: @mshriver @jyejare I think the most recent commit reflects the requested changes

spusateri: Actually, I'm not sure my solution is correct, does `get_settings` need to return the settings object _before_ they are validated in `robotello/config/__init__.py` or is it okay as-is?

spusateri:
This should be ready for review now

scripts/validate_config.py
jyejare: @spusateri I think so, else you will be in the loop!

robottelo/config/__init__.py
```diff
@@ -15,21 +15,31 @@
...
+        includes=["settings.local.yaml", ".secrets.yaml", ".secrets_*.yaml"],
+        envless_mode=True,
+        lowercase_read=True,
+    )
+    settings.validators.register(**VALIDATORS)
+
+    try:
+        settings.validators.validate()
+    except ValidationError as err:
+        logger.warning(f'Dynaconf validation failed, continuing for the sake of unit tests\n{err}')
```
JacobCallahan, line 37: validation shouldn't happen in the get_settings function. it is a step someone may perform after.

JacobCallahan (changes requested):

robottelo/config/__init__.py
mshriver: Since we're currently ignoring all validation errors anyway, why should this method not return validated settings? I'd expect this function to return a validated, usable settings object, not a settings object that I then have to import validators for, or call methods on to check it before I use it.

spusateri: @JacobCallahan What would be an example of needing to perform settings validation after? I can always move the try/except block outside of get_settings

PR #9466, opened 2022-03-31
Add test Content Source selection

swadeley:
Closed loop: BZ#1834921

[Bug 1834921](https://bugzilla.redhat.com/show_bug.cgi?id=1834921) - while registering the new client content source is not found if the content sources are more than the default per page entry.

Hello

this test is to ensure that the pagination setting does not affect the selection of a Content Source when multiple Capsule are present (more than the per-page setting).

Leaving in draft as its blocked by this bug I found while writing the test:
[Bug 2070487](https://bugzilla.redhat.com/show_bug.cgi?id=2070487) - Setting a Content Source is not persistent


swadeley:
Rebased.

swadeley:
Rebased.

PR #9469, opened 2022-03-31
Fix test_content_access_after_stopped_foreman

swadeley:
Hello

A lot of changes where made to this test during refactoring exercises  and it stopped working. After consulting with @omkarkhatavkar , who wrote the original, I have simplified it.

Thank you

### DIFF
--- a/tests/foreman/ui/test_contenthost.py
+++ b/tests/foreman/ui/test_contenthost.py
@@ -1809,8 +1809,6 @@ def test_content_access_after_stopped_foreman(target_sat, rhel7_contenthost):
 
     :CaseImportance: Medium
 
-    :CaseComponent: Infrastructure
-
     :parametrized: yes
     """
     org = target_sat.api.Organization().create()
--- a/tests/foreman/ui/test_contenthost.py
+++ b/tests/foreman/ui/test_contenthost.py
@@ -1834,8 +1834,8 @@ def test_content_access_after_stopped_foreman(target_sat, rhel7_contenthost):
     )
     result = rhel7_contenthost.execute(f'yum -y install {FAKE_1_CUSTOM_PACKAGE}')
     assert result.status == 0
-    assert target_sat.execute('satellite-maintain service stop --only foreman').status == 0
-    result = target_sat.execute('satellite-maintain service status --only=foreman')
+    assert target_sat.cli.Service.stop(options={'only': 'foreman'}).status == 0
+    assert target_sat.cli.Service.status(options={'only': 'foreman'}).status == 1
     assert result.status == 1
     result = rhel7_contenthost.execute(f'yum -y install {FAKE_0_CUSTOM_PACKAGE}')
     assert result.status == 0
--- a/tests/foreman/ui/test_contenthost.py
+++ b/tests/foreman/ui/test_contenthost.py
@@ -1834,11 +1834,11 @@ def test_content_access_after_stopped_foreman(target_sat, rhel7_contenthost):
     )
     result = rhel7_contenthost.execute(f'yum -y install {FAKE_1_CUSTOM_PACKAGE}')
     assert result.status == 0
-    assert target_sat.execute('systemctl stop foreman').status == 0
+    assert target_sat.execute('satellite-maintain service stop --only foreman').status == 0
     result = target_sat.execute('satellite-maintain service status --only=foreman')
     assert result.status == 1
     result = rhel7_contenthost.execute(f'yum -y install {FAKE_0_CUSTOM_PACKAGE}')
     assert result.status == 0
-    assert target_sat.execute('systemctl start foreman').status == 0
+    assert target_sat.execute('satellite-maintain service start --only foreman').status == 0
     result = target_sat.execute('satellite-maintain service status --only=foreman')
     assert result.status == 0
--- a/tests/foreman/ui/test_contenthost.py
+++ b/tests/foreman/ui/test_contenthost.py
@@ -1839,6 +1839,5 @@ def test_content_access_after_stopped_foreman(target_sat, rhel7_contenthost):
     assert result.status == 1
     result = rhel7_contenthost.execute(f'yum -y install {FAKE_0_CUSTOM_PACKAGE}')
     assert result.status == 0
-    assert target_sat.execute('satellite-maintain service start --only foreman').status == 0
-    result = target_sat.execute('satellite-maintain service status --only=foreman')
-    assert result.status == 0
+    assert target_sat.cli.Service.start(options={'only': 'foreman'}).status == 0
+    assert target_sat.cli.Service.status(options={'only': 'foreman'}).status == 0
--- a/tests/foreman/ui/test_contenthost.py
+++ b/tests/foreman/ui/test_contenthost.py
@@ -1798,7 +1798,7 @@ def test_search_for_virt_who_hypervisors(session, default_location):
 @pytest.mark.destructive
 @pytest.mark.run_in_one_thread
 @pytest.mark.upgrade
-def test_content_access_after_stopped_foreman(foreman_service_teardown, rhel7_contenthost):
+def test_content_access_after_stopped_foreman(target_sat, rhel7_contenthost):
     """Install a package even after foreman service is stopped
 
     :id: 71ae6a56-30bb-11eb-8489-d46d6dd3b5b2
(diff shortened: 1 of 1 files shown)


swadeley:
There are still problems with this test, `repos_collection.setup_content` is picking up the Satellite from `conf/server.yaml` and not from the context manager.

mshriver:
@swadeley this is a general 'problem' with the behavior of those 'helper' classes from the products module. @JacobCallahan and @jyejare should be working on rewriting/moving those functions very soon.  The `align_to_xdist` function *should* be taking care of it but we have issues with trying to use a singular settings instance within pytest.

swadeley:
Hi, I spent some time with command like ` sat.cli.Repository.create` and `sat.cli.Repository.synchronize` but then I would stail have to redot the section on setting up the content host. Then I thought there is no real need for a separate Satellite just for the one test, so I decide to simply still further and use default_sat and just one repo, let see if it works.

swadeley:
trigger: robottelo
pytest: tests/foreman/ui/test_contenthost.py -k test_content_access_after_stopped_foreman

tests/foreman/ui/test_contenthost.py
```diff
@@ -1489,22 +1485,21 @@ def test_content_access_after_stopped_foreman(
...
     result = rhel7_contenthost.execute(f'yum -y install {FAKE_1_CUSTOM_PACKAGE}')
     assert result.status == 0
-    assert default_sat.execute('systemctl stop foreman').status == 0
-    result = default_sat.execute('satellite-maintain service status --only=foreman')
+    assert sat.execute('systemctl stop foreman').status == 0
+    result = sat.execute('satellite-maintain service status --only=foreman')
     assert result.status == 1
     result = rhel7_contenthost.execute(f'yum -y install {FAKE_0_CUSTOM_PACKAGE}')
     assert result.status == 0
+    assert sat.execute('systemctl start foreman').status == 0
```
jyejare, line 1503: I think destructive sat is the perfect satellite for this case even though the test is running in single thread. The only reason is `due to some reason foreman service didnt start or tracebacked, this will block further test cases !`, in short `we cannot provision a perfect teardown for this case`!

jyejare (changes requested):

lhellebr (changes requested):
I agree with Jitendra a destructive_sat should be used here.
However, if we didn't have it or if it is too hard to use here for some reason, I'd be willing to ACK this.

tests/foreman/ui/test_contenthost.py
```diff
@@ -1475,11 +1473,9 @@ def test_content_access_after_stopped_foreman(
 
     :CaseComponent: Infrastructure
 
-    :Assignee: lpramuk
-
     :parametrized: yes
     """
-    sat = foreman_service_teardown
+    sat = default_sat
```
mshriver, line 1478: Can I suggest just using `default_sat` in the test function instead of re-labeling it? With the old fixture name it was cumbersome, but `default_sat` is simple.

tests/foreman/ui/test_contenthost.py
mshriver: Its marked destructive, which will block it in current pipelines. destructive marked tests do need to be moved to destructive_sat, but need to be audited for their fixture use.

@JacobCallahan thoughts on merging this as-is? I'm willing to, and we're getting close to spending time on the destructive sat tests.

tests/foreman/ui/test_contenthost.py
swadeley: Hi, looks like I made a commit for that on the 1st of April

swadeley:
trigger: robottelo
pytest: tests/foreman/ui/test_contenthost.py -k test_content_access_after_stopped_foreman

swadeley:
I have just rebased to bring in @JacobCallahan 's recent "Make "default_sat" fixture usage more cooperative" PR.

tests/foreman/ui/test_contenthost.py
```diff
@@ -1458,9 +1458,7 @@ def test_search_for_virt_who_hypervisors(session, default_location):
 @pytest.mark.destructive
 @pytest.mark.run_in_one_thread
 @pytest.mark.upgrade
-def test_content_access_after_stopped_foreman(
-    foreman_service_teardown, rhel7_contenthost, target_sat
-):
+def test_content_access_after_stopped_foreman(default_sat, rhel7_contenthost):
```
mshriver, line 1461: I think your rebase didn't quite go in the right direction. `target_sat` is the correct fixture name now.

swadeley:
trigger: test-robottelo
pytest: tests/foreman/ui/test_contenthost.py -k test_content_access_after_stopped_foreman

swadeley:
trigger: test-robottelo
pytest: tests/foreman/ui/test_contenthost.py -k test_content_access_after_stopped_foreman

swadeley:
Hi @JacobCallahan , after rebasing, the test is failing because while it provisions a new Satellite
instances, when it gets to `repos_collection` commands it still picks up the default sat.

For example, it still tries to execute `hammer lifecycle-environment info` on the Satellite listed in my conf/server.yaml file, rather than on the new. Is that expected? Have I read too much into the benefits of the https://github.com/SatelliteQE/robottelo/pull/9562 ?

    >       repos_collection.setup_virtual_machine(
                rhel7_contenthost, target_sat, install_katello_agent=False
            )

    tests/foreman/ui/test_contenthost.py:1492:

JacobCallahan:
@swadeley the changes introduced with the target_sat behavior enable fixture reuse. What you're looking for now is involved with the helpers refactor. Essentially, no destructive test should use a helper method like the one you mentioned. Instead, those need to be brought under the Satellite object instances, so they can be used via target_sat.

cc @jyejare

tests/foreman/ui/test_contenthost.py
```diff
@@ -1473,36 +1471,34 @@ def test_content_access_after_stopped_foreman(
...
+    repos_collection = RepositoryCollection(
+        distro=DISTRO_RHEL7,
+        repositories=[
+            YumRepository(url=settings.repos.yum_1.url),
+        ],
+    )
+    repos_collection.setup_content(org.id, lce.id, upload_manifest=False)
+    repos_collection.setup_virtual_machine(
+        rhel7_contenthost, target_sat, install_katello_agent=False
+    )
```
jyejare, line 1494: This will be refactored as https://github.com/SatelliteQE/robottelo/pull/9497/files#diff-9021f84900ea80f6128124119a665669cecf60c29ec23da7bbb49611833e739fR1825-R1834

But wait for https://github.com/SatelliteQE/robottelo/pull/9497 to get merge :)

swadeley: Thanks all

swadeley:
trigger: test-robottelo
pytest: tests/foreman/ui/test_contenthost.py -k test_content_access_after_stopped_foreman

swadeley:
trigger: test-robottelo
pytest: tests/foreman/ui/test_contenthost.py -k test_content_access_after_stopped_foreman

swadeley:
trigger: test-robottelo
pytest: tests/foreman/ui/test_contenthost.py -k test_content_access_after_stopped_foreman

swadeley:
HI, PRT 987 and 988 fails, but test passes locally

swadeley:
trigger: test-robottelo
pytest: tests/foreman/ui/test_contenthost.py -k test_content_access_after_stopped_foreman

JacobCallahan (approved):
ACK, merge pending PRT results (assuming no PRT-specific issues)

tests/foreman/ui/test_contenthost.py
```diff
@@ -1811,34 +1811,34 @@ def test_content_access_after_stopped_foreman(foreman_service_teardown, rhel7_co
...
     result = rhel7_contenthost.execute(f'yum -y install {FAKE_1_CUSTOM_PACKAGE}')
     assert result.status == 0
-    assert foreman_service_teardown.execute('systemctl stop foreman').status == 0
-    result = foreman_service_teardown.execute('satellite-maintain service status --only=foreman')
+    assert target_sat.execute('satellite-maintain service stop --only foreman').status == 0
+    result = target_sat.execute('satellite-maintain service status --only=foreman')
```
Gauravtalreja1, line 1838: Would suggest using maintain CLI's here
```suggestion
    assert target_sat.cli.Service.stop(options={'only': 'foreman'}).status == 0
    result = target_sat.cli.Service.status(options={'only': 'foreman'}
```

tests/foreman/ui/test_contenthost.py
```diff
@@ -1811,34 +1811,34 @@ def test_content_access_after_stopped_foreman(foreman_service_teardown, rhel7_co
...
     assert result.status == 1
     result = rhel7_contenthost.execute(f'yum -y install {FAKE_0_CUSTOM_PACKAGE}')
     assert result.status == 0
+    assert target_sat.execute('satellite-maintain service start --only foreman').status == 0
+    result = target_sat.execute('satellite-maintain service status --only=foreman')
+    assert result.status == 0
```
Gauravtalreja1, line 1844:
```suggestion
    assert target_sat.cli.Service.start(options={'only': 'foreman'}).status == 0
    assert target_sat.cli.Service.status(options={'only': 'foreman'}).status == 0
```

tests/foreman/ui/test_contenthost.py
```diff
@@ -1798,7 +1798,7 @@ def test_search_for_virt_who_hypervisors(session, default_location):
 @pytest.mark.destructive
 @pytest.mark.run_in_one_thread
 @pytest.mark.upgrade
-def test_content_access_after_stopped_foreman(foreman_service_teardown, rhel7_contenthost):
```
Gauravtalreja1, line 1801: In addition to this test, there is another test[1] added that is similar, which might be worth considering for this target_sat simplification. Therefore, you could completely remove this `foreman_service_teardown` 

And `:CaseComponent: Infrastructure` is owned by @lpramuk , so I think `:Assignee:` shouldn't be removed
 
[1] https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/sys/test_foreman_service.py

Gauravtalreja1 (changes requested):

tests/foreman/ui/test_contenthost.py
swadeley: @Gauravtalreja1 I am a bit confused, "foreman_service_teardown" has been removed, or should I say it is removed in this PR.

Re your last point: If @lpramuk wants this test he can raise a PR to move it to his module. I think its unfair on a component owner to have other another component owner's test in their module. I will remove the CaseComponent.

swadeley:
trigger: test-robottelo
pytest: tests/foreman/ui/test_contenthost.py -k test_content_access_after_stopped_foreman

jyejare:
@lhellebr and @Gauravtalreja1 Can we merge this now ? OR you still want to hold on ?

JacobCallahan (approved):

tests/foreman/ui/test_contenthost.py
```diff
@@ -1809,36 +1809,33 @@ def test_content_access_after_stopped_foreman(foreman_service_teardown, rhel7_co
...
+    repos_collection.setup_virtual_machine(
+        rhel7_contenthost, target_sat, install_katello_agent=False
+    )
     result = rhel7_contenthost.execute(f'yum -y install {FAKE_1_CUSTOM_PACKAGE}')
     assert result.status == 0
-    assert foreman_service_teardown.execute('systemctl stop foreman').status == 0
-    result = foreman_service_teardown.execute('satellite-maintain service status --only=foreman')
+    assert target_sat.cli.Service.stop(options={'only': 'foreman'}).status == 0
+    assert target_sat.cli.Service.status(options={'only': 'foreman'}).status == 1
     assert result.status == 1
```
Gauravtalreja1, line 1837: @swadeley you missed removing `assert result.status == 1` here, it could fail here

swadeley:
Thank you!

tests/foreman/ui/test_contenthost.py
swadeley: @Gauravtalreja1 I just applied your changes, anyway, too late now, lets see how it goes.

PR #9482, opened 2022-04-06
Remove Openscap Puppet tests

jameerpathan111:
Removing Puppet related tests of Openscap.
I will be raising separate PR to add test scenarios to run on puppet enabled satellite.

### DIFF
--- a/tests/foreman/longrun/test_oscap.py
+++ b/tests/foreman/longrun/test_oscap.py
@@ -26,15 +26,11 @@ from robottelo.cli.ansible import Ansible
 from robottelo.cli.arfreport import Arfreport
 from robottelo.cli.factory import make_hostgroup
 from robottelo.cli.factory import make_scap_policy
-from robottelo.cli.factory import make_tailoringfile
 from robottelo.cli.factory import setup_org_for_a_custom_repo
 from robottelo.cli.host import Host
 from robottelo.cli.job_invocation import JobInvocation
 from robottelo.cli.proxy import Proxy
-from robottelo.cli.scap_policy import Scappolicy
-from robottelo.cli.scap_tailoring_files import TailoringFiles
 from robottelo.cli.scapcontent import Scapcontent
-from robottelo.cli.scparams import SmartClassParameter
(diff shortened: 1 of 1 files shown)


jameerpathan111:
trigger: robottelo
pytest: tests/foreman/longrun/test_oscap.py

jyejare:
@jameerpathan111 Any reasons to remove these tests now and again add everything later with puppet enabled sat?

jameerpathan111:
> @jameerpathan111 Any reasons to remove these tests now and again add everything later with puppet enabled sat?

Because it'll be lot of refactoring, testing and I'd probably add it in different module that's why I want to do it separately.
Till then I'd like at least ansible tests to work. But if you want I can keep them and just remove puppet fixture which cause ansible tests to fail.

jyejare:
@jameerpathan111 Cool! Lets continue with this PR and later you can raise a follow up puppet PR !

tests/foreman/longrun/test_oscap.py
```diff
@@ -150,290 +136,9 @@ def update_scap_content(module_org):
...
-        # Satellite6.
-        arf_report = Arfreport.list({'search': f'host={vm.hostname.lower()}', 'per-page': 1})
-        assert arf_report is not None
-        Arfreport.delete({'id': arf_report[0].get('id')})
-
-
-@pytest.mark.upgrade
-@pytest.mark.tier4
-@pytest.mark.parametrize('distro', [DISTRO_RHEL7, DISTRO_RHEL8])
+@pytest.mark.parametrize('distro', [DISTRO_RHEL6, DISTRO_RHEL7, DISTRO_RHEL8])
```
jyejare, line 141: Any reasons to add RHEL6 since it was not earlier ? I would instead go for RHEL9. 

Just a thought, not a blocker comment !

jyejare (approved):
ACK with a non blocking comment !

ColeHiggins2 (approved):
If Puppet is coming back in a later version, I would suggest you just comment out the code. Up to you though. ACK

PR #9483, opened 2022-04-06
Fixed and improved Satellite Sync test cases

rmynar:
- changed component owner from @latran to @rmynar
- improved handling export and import archives (most failures were caused by garbage in these directories)
- new function `move_pulp_archive` that replaces recurring code and makes code more readable

```
$ pytest --disable-warnings tests/foreman/cli/test_satellitesync.py
=============================================================================== test session starts ===============================================================================
platform linux -- Python 3.10.4, pytest-7.1.1, pluggy-1.0.0
rootdir: /home/rmynar/rmynar/robottelo, configfile: pyproject.toml
plugins: cov-3.0.0, services-2.2.1, forked-1.4.0, mock-3.7.0, xdist-2.5.0, ibutsu-2.0.2, reportportal-5.0.12
collected 36 items / 16 deselected / 20 selected                                                                                                                                  

tests/foreman/cli/test_satellitesync.py ..............s.....                                                                                                                [100%]

===================================================== 19 passed, 1 skipped, 16 deselected, 22 warnings in 7266.37s (2:01:06) ======================================================
```

### DIFF
--- a/tests/foreman/cli/test_satellitesync.py
+++ b/tests/foreman/cli/test_satellitesync.py
@@ -64,13 +64,58 @@ def config_export_import_settings():
     Settings.set({'name': 'content_disconnected', 'value': disconnected_mode_value})
 
 
+@pytest.fixture(scope='function')
+def export_cleanup_function(default_sat, function_org):
+    """Deletes export/import dirs of function org"""
+    yield
+    # Deletes directories created for export/import test
+    assert default_sat.execute(f'rm -rf {EXPORT_DIR}/{function_org.name}').stdout == ''
+    assert default_sat.execute(f'rm -rf {IMPORT_DIR}/{function_org.name}').stdout == ''
+
+
+@pytest.fixture(scope='function')
+def export_cleanup_module(default_sat, module_org):
+    """Deletes export/import dirs of module_org"""
+    yield
+    # Deletes directories created for export/import test
+    assert default_sat.execute(f'rm -rf {EXPORT_DIR}/{module_org.name}').stdout == ''
+    assert default_sat.execute(f'rm -rf {IMPORT_DIR}/{module_org.name}').stdout == ''
+
+
+def validate_filepath(sat_obj, org):
+    """Checks the existence of certain files in a dir"""
+    result = sat_obj.execute(
+        fr'find {EXPORT_DIR}{org.name} -type f \( -name "*.json" -o -name "*.txt" \)'
+    )
+    return result.stdout
+
+
+def move_pulp_archive(sat_obj, org, export_message):
+    """
+    Moves exported archive(s) and its metadata into import directory,
+    sets ownership, returns import path
+    """
+    sat_obj.execute(f'mv {EXPORT_DIR}/{org.name} {IMPORT_DIR}')
+    sat_obj.execute(f'chown -R pulp:pulp {IMPORT_DIR}')
+
+    # removes averything before export path,
+    # replaces EXPORT_PATH by IMPORT_PATH,
+    # removes metadata filename
+    import_path = os.path.dirname(
+        IMPORT_DIR
+        + export_message[export_message.find(EXPORT_DIR) + len(EXPORT_DIR) :]  # noqa: E203
+    )
+
+    return import_path
+
+
 @pytest.mark.run_in_one_thread
 class TestRepositoryExport:
     """Tests for exporting a repository via CLI"""
 
     @pytest.mark.tier3
     def test_positive_export_complete_version_custom_repo(
-        self, default_sat, export_cleanup, module_org
+        self, default_sat, export_cleanup_module, module_org
     ):
         """Export a custom repository via complete version
 
@@ -106,15 +151,15 @@ class TestRepositoryExport:
         assert len(cv['versions']) == 1
         cvv = cv['versions'][0]
         # Verify export directory is empty
-        assert validate_filepath(default_sat) == ''
+        assert validate_filepath(default_sat, module_org) == ''
         # Export content view
         ContentExport.completeVersion({'id': cvv['id'], 'organization-id': module_org.id})
         # Verify export directory is not empty
-        assert validate_filepath(default_sat) != ''
+        assert validate_filepath(default_sat, module_org) != ''
 
     @pytest.mark.tier3
     def test_positive_export_incremental_version_custom_repo(
-        self, default_sat, export_cleanup, module_org
+        self, default_sat, export_cleanup_module, module_org
     ):
         """Export custom repo via incremental export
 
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_satellitesync.py
```diff
@@ -64,13 +64,58 @@ def config_export_import_settings():
...
+@pytest.fixture(scope='function')
+def export_cleanup_function(default_sat, function_org):
+    """Deletes export/import dirs of function org"""
+    yield
+    # Deletes directories created for export/import test
+    assert default_sat.execute(f'rm -rf {EXPORT_DIR}/{function_org.name}').stdout == ''
+    assert default_sat.execute(f'rm -rf {IMPORT_DIR}/{function_org.name}').stdout == ''
+
+
+@pytest.fixture(scope='function')
```
swadeley, line 76:
```suggestion
@pytest.fixture(scope='function')
```
Is that the right scope?
If it is, could you have just one fixture and pass in the org name using "request" to get it?
This is non-blocking comment.

swadeley (approved):
ACK with question.

tests/foreman/cli/test_satellitesync.py
```diff
@@ -64,13 +64,58 @@ def config_export_import_settings():
     Settings.set({'name': 'content_disconnected', 'value': disconnected_mode_value})
 
 
+@pytest.fixture(scope='function')
+def export_cleanup_function(default_sat, function_org):
+    """Deletes export/import dirs of function org"""
+    yield
```
Griffin-Sullivan, line 70: Does the empty `yield` just ensure the teardown is performed after the test is finished?

tests/foreman/cli/test_satellitesync.py
mshriver: I don't think this can be flattened down to one fixture - but I think you're right that the scope of this fixture should me `module`.

rmynar: That scope is on purpose. It should cleanup the export/import directories after each testcase(function) of same module.
I know it might look like a hack, I'll try to come up with some better implementation for 6.11.

mshriver: Gotcha! Can you add an inline comment so its clear (in the future) that is the intent here. I don't think its too much of a hack, maybe its not elegant but its explicit and not a widespread pattern.

tests/foreman/cli/test_satellitesync.py
```diff
@@ -64,13 +64,58 @@ def config_export_import_settings():
...
+
+def validate_filepath(sat_obj, org):
+    """Checks the existence of certain files in a dir"""
+    result = sat_obj.execute(
+        fr'find {EXPORT_DIR}{org.name} -type f \( -name "*.json" -o -name "*.tar.gz" \)'
+    )
+    return result.stdout
+
+
+def move_pulp_archive(sat_obj, org, export_message):
```
mshriver, line 93: Though it may only be used in this test module, this is a function that I'd suggest exist on the `robottelo.hosts.Capsule` class, since its first argument is a satellite object.

tests/foreman/cli/test_satellitesync.py
```diff
@@ -64,13 +64,58 @@ def config_export_import_settings():
...
+    sets ownership, returns import path
+    """
+    sat_obj.execute(f'mv {EXPORT_DIR}/{org.name} {IMPORT_DIR}')
+    sat_obj.execute(f'chown -R pulp:pulp {IMPORT_DIR}')
+
+    # removes averything before export path,
+    # replaces EXPORT_PATH by IMPORT_PATH,
+    # removes metadata filename
+    import_path = os.path.dirname(
+        IMPORT_DIR
```
mshriver, line 105: For readability and to avoid the E203, I'd create a variable here

Secondly I'd advise against using `+` for string concatenation, especially since we don't type hint, use `f-string` instead. 

Third, did you check whether you can make this replacement reliably with `replace` instead of directly trying to slice and concatenate? Or better yet, with `Pathlib` tools like `relative_to`, `join_path`, or `parts`?

https://docs.python.org/3/library/pathlib.html

rmynar: This is a bit more complicated... We don't receive just path to archive, but need to extract it from an export message, which looks like this:
`Generated /export_dir/org_name/and_some_more_subdirectories/metadata.json`
So in that function we move the exported archive and metadata into import directory while keeping the subdirectory structure, then we return the import path (without metadata filename).
I think it could be more readable with `re`, so I suggest this change:
`import_path = os.path.dirname(re.sub(rf'.*{EXPORT_DIR}', IMPORT_DIR, export_message))`

tests/foreman/cli/test_satellitesync.py
```diff
@@ -64,13 +64,58 @@ def config_export_import_settings():
     Settings.set({'name': 'content_disconnected', 'value': disconnected_mode_value})
 
 
+@pytest.fixture(scope='function')
+def export_cleanup_function(default_sat, function_org):
```
latran, line 68:
```suggestion
def export_import_cleanup_function(default_sat, function_org):
```

tests/foreman/cli/test_satellitesync.py
```diff
@@ -64,13 +64,58 @@ def config_export_import_settings():
...
+    """Deletes export/import dirs of function org"""
+    yield
+    # Deletes directories created for export/import test
+    assert default_sat.execute(f'rm -rf {EXPORT_DIR}/{function_org.name}').stdout == ''
+    assert default_sat.execute(f'rm -rf {IMPORT_DIR}/{function_org.name}').stdout == ''
+
+
+@pytest.fixture(scope='function')
+def export_cleanup_module(default_sat, module_org):
```
latran, line 77:
```suggestion
def export_import_cleanup_module(default_sat, module_org):
```

latran (changes requested):
Just some naming changes to make things clearer.

tests/foreman/cli/test_satellitesync.py
mshriver: @lhellebr I like that re.sub approach as well, thanks for explaining!

tests/foreman/cli/test_satellitesync.py
rmynar: agree, renamed

tests/foreman/cli/test_satellitesync.py
rmynar: agree, renamed

tests/foreman/cli/test_satellitesync.py
rmynar: implemented in latest commit

mshriver (approved):
Great enhancement to your tests @rmynar !

PR #9486, opened 2022-04-08
fix search for infrastructure hosts

pondrejk:
The change hostname script does not clean the previous hostname form the host list -- BZ#1994685  This ghost host then shows up in the infrstructure host search results, causing failed assertions. This PR modifies the asserts to account for the problem until the bz is fixed

### DIFF
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -67,6 +67,7 @@ from robottelo.datafactory import invalid_values_list
 from robottelo.datafactory import valid_data_list
 from robottelo.datafactory import valid_hosts_list
 from robottelo.hosts import ContentHostError
+from robottelo.utils.issue_handlers import is_open
 
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_host.py
```diff
@@ -843,14 +844,20 @@ def test_positive_list_infrastructure_hosts(
     Host.update({'name': default_sat.hostname, 'new-organization-id': module_org.id})
     # list satellite hosts
     hosts = Host.list({'search': 'infrastructure_facet.foreman=true'})
-    assert len(hosts) == 1
+    if is_open('BZ:1994685'):
+        assert len(hosts) == 2
+    else:
+        assert len(hosts) == 1
```
jyejare, line 850: I might try to delete the ghost host with `if BZ open` condition here if it's easy to identify that host for deletion. And then will stick to hosts count assertion to 1.

Not a blocking comment though!

jyejare (approved):
ACK with suggestion, with ready to get ACKed and Nacked.

lhellebr (approved):
LGTM

mshriver (approved):

PR #9493, opened 2022-04-11
changing scope for rex infra user fixtures

pondrejk:
Creating users within these fixtures tends to fail in automation on missing taxonomy. My hypothesis is that the users are being created before the module org. Changing the fixture scope for user creation to battle this.

### DIFF
--- a/tests/foreman/cli/test_remoteexecution.py
+++ b/tests/foreman/cli/test_remoteexecution.py
@@ -936,8 +936,8 @@ class TestAnsibleREX:
 class TestRexUsers:
     """Tests related to remote execution users"""
 
-    @pytest.fixture(scope='module')
-    def module_rexmanager_user(self, module_org):
+    @pytest.fixture(scope='class')
+    def class_rexmanager_user(self, module_org):
         """Creates a user with Remote Execution Manager role"""
         password = gen_string('alpha')
         rexmanager = gen_string('alpha')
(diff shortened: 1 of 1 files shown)


pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k test_positive_rex_against_infra_hosts

ColeHiggins2 (approved):
ACK

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -936,7 +936,7 @@ def test_positive_install_ansible_collection(self, fixture_sca_vmsetup, module_g
 class TestRexUsers:
     """Tests related to remote execution users"""
 
-    @pytest.fixture(scope='module')
+    @pytest.fixture(scope='class')
```
jyejare, line 939: Since we are moving towards ReX and Ansible as defaults in future releases to replace this fixtures would be really helpful to move them to `pytest_fixtures` package under REX fixture module. Probably there you need not to change the scope to class level from existing module! But if you still want to continue here just rename the fixtures to prepend with `class_` not `module_`

Not blocking your PR fo this though!

jyejare (approved):

tests/foreman/cli/test_remoteexecution.py
pondrejk: updated, thanks!

PR #9509, opened 2022-04-20
Add tests for puppet availability on upgraded setup

vsedmik:
Puppet is supposed to remain enabled on the upgraded setup, so I'm adding tests to check it in upgrades.
I've also updated Report to ConfigReport in CLI as the original is deprecated:
```
[root@sat ~]# hammer report --help
report command is deprecated and will be removed in one of the future versions. Please use config-report command instead.
```

Tests results against upgraded setup:
```
(venv39) [vsedmik@localhost robottelo]$ pytest -m post_upgrade tests/upgrades/test_puppet.py
================================================================================= test session starts =================================================================================
platform linux -- Python 3.9.9, pytest-7.1.1, pluggy-0.13.1
rootdir: /home/vsedmik/PycharmProjects/robottelo, configfile: pyproject.toml
plugins: forked-1.3.0, services-2.2.1, mock-3.7.0, lazy-fixture-0.6.3, ibutsu-2.0.2, xdist-2.5.0, reportportal-5.0.12
collected 3 items                                                                                                                                                                     

tests/upgrades/test_puppet.py ...                                                                                                                                               [100%]

============================================================================ 3 passed, 1 warning in 29.50s ============================================================================
```

### DIFF
--- a/tests/upgrades/test_puppet.py
+++ b/tests/upgrades/test_puppet.py
@@ -1,0 +1,89 @@
+"""Test for Remote Execution related Upgrade Scenario's
+
+:Requirement: UpgradedSatellite
+
+:CaseAutomation: Automated
+
+:CaseLevel: Acceptance
+
+:CaseComponent: Puppet
+
+:Assignee: vsedmik
+
+:TestType: Functional
+
+:CaseImportance: Medium
+
+:Upstream: No
+"""
+import pytest
+
+from robottelo.config import settings
+from robottelo.hosts import Capsule
+
+
+@pytest.fixture(scope='module', params=[True, False], ids=["satellite", "capsule"])
+def upgrade_server(request, default_sat):
+    if request.param:
+        return default_sat
+    else:
+        return Capsule(settings.upgrade.capsule_hostname)
+
+
+class TestPuppet:
+    """
+    Test puppet availability on the upgraded setup.
+    """
+
+    @pytest.mark.post_upgrade
+    def test_post_puppet_active(self, upgrade_server):
+        """Test that puppet remains active after the upgrade on both,
+        Satellite and Capsule.
+
+        :id: postupgrade-6360e928-ba0f-41a7-9aaa-bea87cb6342d
+
+        :steps:
+            1. Check for capsule features.
+            2. Check for puppetserver status.
+
+        :expectedresults:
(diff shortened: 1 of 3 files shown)


tests/upgrades/test_puppet.py
```diff
@@ -0,0 +1,89 @@
...
+
+:Upstream: No
+"""
+import pytest
+
+from robottelo.config import settings
+from robottelo.hosts import Capsule
+
+
+@pytest.fixture(scope='module', params=[True, False], ids=["satellite", "capsule"])
```
mshriver, line 25: Thank you for using ids for your parameters!

tests/upgrades/test_puppet.py
```diff
@@ -0,0 +1,89 @@
...
+        assert 'puppetca' in result
+
+        result = upgrade_server.execute('rpm -q puppetserver')
+        assert result.status == 0
+
+        result = upgrade_server.execute('systemctl status puppetserver')
+        assert 'active (running)' in result.stdout
+
+    @pytest.mark.post_upgrade
+    def test_post_puppet_reporting(self, default_sat):
```
mshriver, line 67: Consider combining these into one testcase, I don't know that the content of this case is enough to warrant splitting the behavior. Its all pretty basic validation that puppet is installed, running, and connected to satellite still.  Within one case you still get clear indication of what exactly is broken about puppet following upgrade, and there's not much point in checking that `puppet agent` returns 0 if the puppetserver service isn't running, or package isn't installed.

You could I think even simplify further and iterate the checks over `default_sat` and the `Capsule` instance within the test case instead of parametrizing, so you don't check the agent status on the `default_sat` twice.

Not a blocking comment - but have a thought on what you gain by having separate medium importance test cases here.

mshriver (approved):

tests/upgrades/test_puppet.py
vsedmik: Hehe, yep I had it with iteration through `[default_sat, caps]` before, but decided to parametrize like a 'pro' later. :)
But you are right this can me merged into one case, no special need to have two. :+1:

jyejare (approved):

latran (approved):
ACK

ColeHiggins2 (approved):
Ack

pondrejk (approved):

PR #9510, opened 2022-04-20
Add tests for convert2rhel feature

peterdragun:
~~Depends on nailgun PR: SatelliteQE/nailgun/pull/827~~

Note that this PR requires changes in conf, so please merge alongside PR to our conf repo.

### DIFF
--- a/conf/content_host.yaml.template
+++ b/conf/content_host.yaml.template
@@ -2,6 +2,9 @@ content_host:
     deploy_workflow:
         default:
         rhel9: # example of workflow override for version
+    DEPLOY_SCENARIO:
+        default:
+        centos:
     default_rhel_version: 7
     hardware:
         RHEL6:
@@ -19,5 +22,11 @@ content_host:
         RHEL9:
             RELEASE: 9.0
             COMPOSE: RHEL-9.0.0-20220118.1
+            MEMORY: 1536 MiB
+            CORES: 1
+        CENTOS:
+            MEMORY: 1536 MiB
+            CORES: 1
+        ORACLE_LINUX:
             MEMORY: 1536 MiB
             CORES: 1
--- a/conf/content_host.yaml.template
+++ b/conf/content_host.yaml.template
@@ -1,6 +1,6 @@
 content_host:
     default_rhel_version: 7
-    default_workflow: deploy-base-rhel
+    default_deploy_workflow: deploy-base-rhel
     deploy_kwargs:
         RHEL6:
             RELEASE: 6.10
@@ -19,13 +19,15 @@ content_host:
             COMPOSE: RHEL-9.0.0-20220118.1
             MEMORY: 1536 MiB
             CORES: 1
-        CENTOS:
-            SCENARIO: centos
-            DEPLOY_WORKFLOW: deploy-centos
-            MEMORY: 1536 MiB
+        CENTOS7:
+            deploy_workflow: deploy-centos
+            scenario: centos
+            release: '7'
+            MEMORY: 1GiB
             CORES: 1
-        ORACLE_LINUX:
-            SCENARIO: oracle
-            DEPLOY_WORKFLOW: deploy-oracle-linux
-            MEMORY: 1536 MiB
+        ORACLE7:
+            deploy_workflow: deploy-oracle-linux
+            scenario: oracle
+            release: '7.9'
+            MEMORY: 1GiB
             CORES: 1
--- a/conf/content_host.yaml.template
+++ b/conf/content_host.yaml.template
@@ -1,12 +1,7 @@
 content_host:
-    deploy_workflow:
-        default:
-        rhel9: # example of workflow override for version
-    DEPLOY_SCENARIO:
-        default:
-        centos:
     default_rhel_version: 7
-    hardware:
+    default_workflow: deploy-base-rhel
+    deploy_kwargs:
         RHEL6:
             RELEASE: 6.10
             MEMORY: 1GiB
@@ -25,8 +20,12 @@ content_host:
             MEMORY: 1536 MiB
             CORES: 1
         CENTOS:
+            SCENARIO: centos
+            DEPLOY_WORKFLOW: deploy-centos
             MEMORY: 1536 MiB
             CORES: 1
         ORACLE_LINUX:
+            SCENARIO: oracle
+            DEPLOY_WORKFLOW: deploy-oracle-linux
             MEMORY: 1536 MiB
             CORES: 1
--- a/conf/repos.yaml.template
+++ b/conf/repos.yaml.template
@@ -58,6 +58,12 @@ REPOS:
     RHEL9: "@format {this[repos].MOCK_SERVICE_REPO_BASE}/epel-9-x86_64/"
   MOCK_SERVICE_RPM: "robottelo-mock-service"
 
+  CONVERT2RHEL:
+    CONVERT_TO_RHEL_REPO: http://convert2rhel_repo.example.com
+    SSL_CERT_ORACLE: http://cert.example.com
+    UBI7: http://ubi7_repo.example.com
+    UBI8: http://ubi8_repo.example.com
+
 # Robottelo fixtures repos
   MODULE_STREAM_0:
     URL:
--- a/robottelo/constants/__init__.py
+++ b/robottelo/constants/__init__.py
@@ -2149,9 +2149,3 @@ RH_SAT_ROLES = [
     'settings',
     'sync_plans',
 ]
-
-CONVERT_TO_RHEL_REPO = ""
-SSL_CERT_ORACLE = ""
-
-UBI_ORACLE7 = ""
-UBI_ORACLE8 = ""
--- a/robottelo/constants/__init__.py
+++ b/robottelo/constants/__init__.py
@@ -375,7 +375,7 @@ REPOSET = {
         'rhel8_aps': 'Red Hat Enterprise Linux 8 for x86_64 - AppStream (Kickstart)',
         'rhel9': 'Red Hat Enterprise Linux 9 for x86_64 - BaseOS Beta (Kickstart)',
     },
-    'rhel8_bos': 'Red Hat Enterprise Linux 8 for x86_64 - BaseOS RPMs 8',
+    'rhel8_bos': 'Red Hat Enterprise Linux 8 for x86_64 - BaseOS (RPMs)',
     'rhel8_aps': 'Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)',
     'rhel7_extra': 'Red Hat Enterprise Linux 7 Server - Extras (RPMs)',
     'rhel7_optional': 'Red Hat Enterprise Linux 7 Server - Optional (RPMs)',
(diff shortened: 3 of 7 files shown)


peterdragun:
trigger: test-robottelo
pytest: tests/foreman/api/test_convert2rhel.py
nailgun: 827

peterdragun:
trigger: test-robottelo
pytest: tests/foreman/api/test_convert2rhel.py

peterdragun:
trigger: test-robottelo
pytest: tests/foreman/api/test_convert2rhel.py

peterdragun:
trigger: test-robottelo
pytest: tests/foreman/api/test_convert2rhel.py

tests/foreman/api/test_convert2rhel.py
```diff
@@ -0,0 +1,324 @@
...
+    act_key = entities.ActivationKey(
+        organization=org,
+        content_view=cv,
+        environment=lce,
+    ).create()
+    act_key.add_subscriptions(data={'subscription_id': subscription_id})
+    return act_key
+
+
+def update_cv(cv, lce, repos):
```
akhil-jha, line 68: is it not possible to combine them(this and above 2 function) into a single fixture?

tests/foreman/api/test_convert2rhel.py
```diff
@@ -0,0 +1,324 @@
...
+        # install and set correct kernel, based on convert2rhel docs
+        result = host.execute(
+            'yum install -y kernel && '
+            'grubby --set-default /boot/vmlinuz-'
+            '`rpm -q --qf "%{BUILDTIME}\t%{EVR}.%{ARCH}\n" kernel | sort -nr | head -1 | cut -f2`'
+        )
+        assert result.status == 0
+        host.install_katello_ca(default_sat)
+        try:
+            result = host.execute('reboot', timeout='5s')
```
akhil-jha, line 203: Do you think https://github.com/SatelliteQE/robottelo/blob/67c8165fa670e36584fdd9f06ffb0af5aea51b80/robottelo/hosts.py#L193 could be a replacement here?

tests/foreman/api/test_convert2rhel.py
peterdragun: I can combine creating an activation key with updating the content view, but `create_repo` is still needed for the UBI repo in `oracle_host` fixture. Should I combine those 2?

tests/foreman/api/test_convert2rhel.py
peterdragun: :+1: Didn't know about this option, I will try it

robottelo/constants/__init__.py
```diff
@@ -2148,3 +2149,9 @@
     'settings',
     'sync_plans',
 ]
+
+CONVERT_TO_RHEL_REPO = "https://cdn.redhat.com/content/public/convert2rhel/{}/x86_64/os/"
+SSL_CERT_ORACLE = "https://ftp.redhat.com/redhat/convert2rhel/redhat-uep.pem"
+
+UBI_ORACLE7 = "https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi/server/7/7Server/x86_64/os"
+UBI_ORACLE8 = "https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi8/8/x86_64/baseos/os/"
```
jyejare, line 2157: These internal redhat links shouldnt be public here in Github.

Instead fetch this from `repo/conf.yaml`!

jyejare: @peterdragun Remove ASAP these urls from here!

peterdragun: Those links are actually publicly available on multiple places, docs and blogs about converting to rhel. For example, see the documentation for convert2rhel.

shubhamsg199 (changes requested):

conf/repos.yaml.template
```diff
@@ -58,6 +58,12 @@ REPOS:
     RHEL9: "@format {this[repos].MOCK_SERVICE_REPO_BASE}/epel-9-x86_64/"
   MOCK_SERVICE_RPM: "robottelo-mock-service"
 
+  CONVERT2RHEL:
+    CONVERT_TO_RHEL_REPO = http://convert2rhel_repo.example.com
+    SSL_CERT_ORACLE = http://cert.example.com
+    UBI7 = http://ubi7_repo.example.com
+    UBI8 = http://ubi8_repo.example.com
```
Gauravtalreja1, line 65:
```suggestion
  CONVERT2RHEL:
    CONVERT_TO_RHEL_REPO: http://convert2rhel_repo.example.com
    SSL_CERT_ORACLE: http://cert.example.com
    UBI7: http://ubi7_repo.example.com
    UBI8: http://ubi8_repo.example.com
```

robottelo/constants/__init__.py
```diff
@@ -634,7 +634,8 @@
     },
     'rhel8_bos': {
         'id': 'rhel-8-for-x86_64-baseos-rpms',
-        'name': 'Red Hat Enterprise Linux 8 for x86_64 - BaseOS RPMs 8',
+        'name': 'Red Hat Enterprise Linux 8 for x86_64 - BaseOS RPMs 8.5',
```
Gauravtalreja1, line 637: question, why hardcoding rhel8 repos to 8.5? any reason?

peterdragun: Convert2rhel needs to use specific minor version repos.
As I am the only one that is currently using this constant, I think it is OK hardcode those.

mshriver: @peterdragun that's a dangerous assumption to make.

`rhel8_aps` is being used in `tests/foreman/api/test_contentmanagement.py`, and it doesn't look like rhel8_bos is used directly, but the field here could be composed within tests.

I think a much more robust implementation would set this based on `content_host.os_version` within the test.  The update to rhel 8.6 being available (and the default for rhel8 contenthost checkouts) will happen separately from this PR, and will break this assumption in the future.

tests/foreman/api/test_convert2rhel.py
```diff
@@ -0,0 +1,323 @@
...
+    smart_proxy_location,
+    module_promoted_cv,
+    module_lce,
+    ssl_cert,
+    version,
+):
+    """Deploy and register Oracle host"""
+    rh_repos = request.getfixturevalue(f"enable_rhel{version}_subscriptions")
+    conf = {
+        "workflow": "deploy-oracle-linux",
```
mshriver, line 185: Same as above - `conf/content_host.yaml` structure should be updated to include non-rhel settings for workflow names and kwargs.

robottelo/constants/__init__.py
peterdragun: that test in `tests/foreman/api/test_contentmanagement.py` is actually using the kickstart package not that I was changing. And those kickstart repos are locked to 8.5 or 8.4, even version doesn't match between aps and bos repos. So I am really the only one using this repo.

Actually, I wanted to change the version based on the content host version but, then I couldn't find a way to keep the same parametrization that I have created. Currently, all repos for RHEL are synced just once and then they are used 2 times. I will try to find the solution for this, but I couldn't is it okay to create a new constant with name e.g. `rhel85_bos`?

tests/foreman/api/test_convert2rhel.py
```diff
@@ -0,0 +1,323 @@
...
+    """Create credetial with SSL cert for Oracle Linux"""
+    res = requests.get(settings.repos.convert2rhel.ssl_cert_oracle)
+    res.raise_for_status()
+    return entities.ContentCredential(
+        content=res.text, organization=module_org, content_type='cert'
+    ).create()
+
+
+@pytest.fixture(scope='module')
+def manifest(module_org):
```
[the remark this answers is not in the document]
peterdragun: there already is a fixture in taxonomies named `module_manifest_org`, but it is creating a new org instead of using `module_org` which is used in my other fixtures. This org with manifest is used a lot, so I didn't want to change it to use `module_org` because I am not sure what it could cause.

conf/content_host.yaml.template
```diff
@@ -2,6 +2,9 @@ content_host:
     deploy_workflow:
         default:
         rhel9: # example of workflow override for version
+    DEPLOY_SCENARIO:
```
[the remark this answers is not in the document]
JacobCallahan: @mshriver @peterdragun 
I am also preferring #2 and think that it can be included in the scope of this PR instead of putting it off for later.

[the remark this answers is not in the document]
peterdragun: Scenarios and workflows were moved to `deploy_kwargs` as suggested. I have also created `default_workflow` for base RHEL content hosts to reduce redundancy in config.

peterdragun:
@mshriver (maybe also @JacobCallahan ) Can you please look at my new implementation of content_host config?

tests/foreman/api/test_convert2rhel.py
```diff
@@ -0,0 +1,314 @@
...
+    )
+    result = entities.JobInvocation(id=job['id']).read()
+    assert result.succeeded == 1
+
+    # check facts: correct os and valid subscription status
+    host_content = entities.Host(id=centos.hostname).read_json()
+    assert (
+        host_content['operatingsystem_name'].startswith(f"RHEL Server {version}")
+        or host_content['operatingsystem_name'].startswith(f"RedHat {version}")
+        or host_content['operatingsystem_name'].startswith(f"RHEL {version}")
```
ogajduse, line 312: Is this related to the [BZ2080347](https://bugzilla.redhat.com/show_bug.cgi?id=2080347)?
If so, please put the BZ link to the comment for reference.

ogajduse (commented):
Nice work @peterdragun! I left a couple of comments. A lot of them are non-blocking ones.

tests/foreman/api/test_convert2rhel.py
```diff
@@ -0,0 +1,314 @@
...
+            org_id=module_org.id,
+            product=REPOS[name]['product'],
+            repo=REPOS[name]['name'] + minor,
+            reposet=REPOS[name]['reposet'],
+            releasever=REPOS[name]['releasever'] + minor,
+        )
+        # Sync step because repo is not synced by default
+        rh_repo = entities.Repository(id=rh_repo_id).read()
+        rh_repo.sync(timeout=2500)
+        rh_repos.append(rh_repo)
```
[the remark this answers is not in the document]
peterdragun: I can definitely make the first repo synchronization async, but I would keep the second one synchronous. To make sure all repos are synced before starting the conversion.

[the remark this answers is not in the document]
ogajduse: I see that you have made the changes already. If the first async sync would fail, you would get to know it once you want to consume its content. Isn't it too late? Would implementing some wait_for logic make sense here?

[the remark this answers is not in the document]
peterdragun: :+1: adding wait_for logic and all syncs now can be async with assert on task status on each sync

pondrejk (approved):

rplevka (approved):

PR #9516, opened 2022-04-21
Solved an issue with Satellite objects modifying classes

JacobCallahan:
The Satellite object's api and cli attributes nest entity classes from
their respective places. These classes were intended to provide
isolation between the main testing satellite and Satellite object
instances.
However, the previous implementation modified the entity classes that
the rest of the framework used as well instead of creating copies of the
classes specific to that Satellite object.
This change fixes that by copying the classes (through on-demand
subclassing) to ensure there is true isolation.


latran (approved):
ACK

swadeley (approved):
ACK

PR #9520, opened 2022-04-22
test for accessible content status

ColeHiggins2:
Customer scenario testing that candlepin status is accurate when no certificate has been updated

### DIFF
--- a/tests/foreman/cli/test_repository.py
+++ b/tests/foreman/cli/test_repository.py
@@ -28,14 +28,11 @@ from fauxfactory import gen_url
 from nailgun import entities
 from wait_for import wait_for
 
-from robottelo import manifests
-from robottelo.api.utils import upload_manifest
 from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.content_export import ContentExport
 from robottelo.cli.content_import import ContentImport
 from robottelo.cli.contentview import ContentView
 from robottelo.cli.factory import CLIFactoryError
-from robottelo.cli.factory import make_activation_key
 from robottelo.cli.factory import make_content_credential
 from robottelo.cli.factory import make_content_view
 from robottelo.cli.factory import make_filter
@@ -140,24 +137,6 @@ def repo(repo_options):
 def gpg_key(module_org):
     """Create a new GPG key."""
     return make_content_credential({'organization-id': module_org.id})
-
-
-@pytest.fixture(scope='class')
-def host_setup(module_org):
-    with manifests.clone(name='golden_ticket') as manifest:
-        upload_manifest(module_org.id, manifest.content)
-    new_product = make_product({'organization-id': module_org.id})
-    new_repo = make_repository({'product-id': new_product['id']})
-    Repository.synchronize({'id': new_repo['id']})
-    new_ak = make_activation_key(
-        {
-            'lifecycle-environment': 'Library',
-            'content-view': 'Default Organization View',
-            'organization-id': module_org.id,
-            'auto-attach': False,
-        }
-    )
-    return new_ak
 
 
 class TestRepository:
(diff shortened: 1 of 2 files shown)


tests/foreman/cli/test_repository.py
```diff
@@ -2041,6 +2041,35 @@ def test_negative_update_red_hat_repo(self, module_manifest_org):
...
+            :BZ: 2010138
+
+            :CaseImportance: Critical
+            """
+            rhel7_contenthost_class.install_katello_ca(default_sat)
+            rhel7_contenthost_class.register_contenthost(
+                module_org.label, golden_ticket_host_setup['name']
+            )
+            assert rhel7_contenthost_class.subscribed
+            rhel7_contenthost_class.run('yum repolist')
```
latran, line 2067: Can you add an assertion on what this should return as?  Just so it can go along the lines with the bz verification step.

latran (changes requested):

tests/foreman/cli/test_repository.py
```diff
@@ -2041,6 +2041,35 @@ def test_negative_update_red_hat_repo(self, module_manifest_org):
         repo_info = Repository.info({'organization-id': module_manifest_org.id, 'id': rh_repo_id})
         assert repo_info['url'] in [repo.get('url') for repo in repo_list]
 
+        @pytest.mark.tier1
+        def test_positive_accessible_content_status(
+            module_org, golden_ticket_host_setup, rhel7_contenthost_class, default_sat
+        ):
+            """Verify that the Candlepin respose accesible_content returns a 304 when no
```
JacobCallahan, line 2048: "respose accesible"

JacobCallahan (changes requested):

latran:
@ColeHiggins2 any updates? it's been in draft for a while.

tests/foreman/cli/test_repository.py
ColeHiggins2: The only assertion I could do here would be to see if the custom repo is in there. However, the point of the 'yum repolist' in this case is to send requests to candlepin, not check the repo.

tests/foreman/cli/test_repository.py
```diff
@@ -139,6 +142,24 @@ def gpg_key(module_org):
     return make_content_credential({'organization-id': module_org.id})
 
 
+@pytest.fixture(scope='class')
+def host_setup(module_org):
```
JacobCallahan, line 146: I think could be useful at the global level. the scoping is a bit more containing since most tests don't fall under a class, but it could be more generalized.
```suggestion
def host_setup(module_org, class_target_sat):
```

tests/foreman/cli/test_repository.py
```diff
@@ -139,6 +142,24 @@ def gpg_key(module_org):
...
+    with manifests.clone(name='golden_ticket') as manifest:
+        upload_manifest(module_org.id, manifest.content)
+    new_product = make_product({'organization-id': module_org.id})
+    new_repo = make_repository({'product-id': new_product['id']})
+    Repository.synchronize({'id': new_repo['id']})
+    new_ak = make_activation_key(
```
JacobCallahan, line 152: let's try to use the new helpers pattern
```suggestion
    new_product = class_target_sat.cli_factory.make_product({'organization-id': module_org.id})
    new_repo = class_target_sat.cli_factory.make_repository({'product-id': new_product['id']})
    Repository.synchronize({'id': new_repo['id']})
    new_ak = class_target_sat.cli_factory.make_activation_key(
```

JacobCallahan (changes requested):
Small changes, but looking great!

latran (approved):
ack

JacobCallahan (approved):
ACK, pending checks

PR #9523, opened 2022-04-25
Remove libvirt_discovery tag for whole module

lhellebr:
These tests should never have been marked as libvirt_discovery because they are not discovery tests

### DIFF
--- a/tests/foreman/cli/test_computeresource_libvirt.py
+++ b/tests/foreman/cli/test_computeresource_libvirt.py
@@ -48,8 +48,6 @@ from robottelo.constants import FOREMAN_PROVIDERS
 from robottelo.constants import LIBVIRT_RESOURCE_URL
 from robottelo.datafactory import parametrized
 
-pytestmark = pytest.mark.libvirt_discovery
-
 
 def valid_name_desc_data():
     """Random data for valid name and description"""


mshriver (approved):

jyejare (commented):
Yeah, Interesting , why its marked with libvirt discovery, could be libvirt provisioning instead.

BTW , removing this marker will now run all the tests on CI, are we ready for it ?

Also, If libvirt provisioning is in place, then feel free to remove the marker and from deselection plugin as well.

lhellebr:
These tests are not either discovery OR provisioning. They are just CRUD tests that do things like add CR to the Satellite and then delete it.

About skipping, I see a `@pytest.mark.skip_if_not_set('libvirt')` there which should do the trick?

jyejare:
@lhellebr Agree that it is just CRUD tests to Libvirt but even the CRUD needs Libvirt in place, these tests will fail with no libvirt deployed(or given libvirt URL) to connect to !

Again if we have libvirt in place, we can easily apply skip_if_not_set marker. But if its not, then even though the values are in place for libvirt conf tests will fail with unavailability!

jyejare:
@mshriver @lhellebr My suggestion is not to validate the `libvirt_url is not None` but the actual integration between satellite and libvirt that would allow satellite to provision on libvirt.

lhellebr:
@jyejare This is not a provisioning test. There are different tests for provisioning and they are (and should be) currently disabled since there is no provisioning environment available. The fact that you can add a Libvirt CR doesn't mean you can provision on it, due to network setup.

PR #9524, opened 2022-04-25
Fix list comprehension of org help command

latran:
The org list command has been failing due to list comprehension.

Note that line 97 uses `len(lines) - 1` to account for `description` appearing twice, but in separate section so it's still valid.  Not sure how to handle that or to just completely remove it from the list.  Any thought is appreciated.

Result:
```
pytest tests/foreman/cli/test_organization.py -k test_verify_bugzilla_1078866
================= 1 passed, 36 deselected in 17.89s ============
```

### DIFF
--- a/tests/foreman/cli/test_organization.py
+++ b/tests/foreman/cli/test_organization.py
@@ -86,15 +86,15 @@ def test_verify_bugzilla_1078866():
     # org list --help:
     result = Org.list({'help': True}, output_format=None)
     # get list of lines and check they all are unique
-    lines = [line for line in result if line != '' and '----' not in line]
+    lines = [line.strip() for line in result.splitlines() if line and '----' not in line]
     assert len(set(lines)) == len(lines)
 
     # org info --help:info returns more lines (obviously), ignore exception
     result = Org.info({'help': True}, output_format=None)
 
     # get list of lines and check they all are unique
-    lines = [line for line in result['options']]
-    assert len(set(lines)) == len(lines)
+    lines = [line for line in result['options'] if '----' not in line]
+    assert len(set(lines)) == len(lines) - 1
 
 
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_organization.py
```diff
@@ -86,15 +86,15 @@ def test_verify_bugzilla_1078866():
...
     assert len(set(lines)) == len(lines)
 
     # org info --help:info returns more lines (obviously), ignore exception
     result = Org.info({'help': True}, output_format=None)
 
     # get list of lines and check they all are unique
-    lines = [line for line in result['options']]
-    assert len(set(lines)) == len(lines)
+    lines = [line for line in result['options'] if '----' not in line]
+    assert len(set(lines)) == len(lines) - 1
```
mshriver, line 97: @latran your comment in the PR description should be an in-line comment here, otherwise the next owner of the test is going to wonder why this is being done.

At a certain point I question the value of what we're checking here, even though this is a regression BZ (Satellite 6.0). 

There are some tests that are written against the full-help text for hammer, but it may be that the original BZ would not manifest against the full-help context.

https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/cli/test_hammer.py#L86

This test, and its parser function, I think could be improved to provide coverage against this type of failure (duplicated options) in all hammer commands, instead of just `organization`.  The parser function could raise an exception that the test would end up hitting, or the test itself can validate against the options.  In that test, the options are part of a data structure, and you don't have to worry about the overall length of the help output being sensitive to one line of difference.

mshriver (approved):
Approving the change in its small scope, but I think this low value test function's coverage can be achieved elsewhere.

jyejare (approved):

Issue #9517, opened 2022-04-22
fix tests using module_host_subscription.enable_repo


tstrych:
https://github.com/SatelliteQE/robottelo/blob/00837024b76b5c04a1f2d007c091b65dab44b82c/tests/foreman/cli/test_host.py#L2090

I shoot myself into leg :gun: in the #9319 
should be `module_host_subscription.client.enable_repo`

PR #4599, opened 2017-04-24
[6.3][cherry-pick] qe test coverage 1366437 - host subscription sub commands

ldjebran:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1199515
original PR: https://github.com/SatelliteQE/robottelo/pull/4515

Faced an other bug when tried to run the tests (open a new as the commands works as expected in 6.2.9)
host subscription register is a fundamental functionality for this series of sub commands and tests
https://bugzilla.redhat.com/show_bug.cgi?id=1444886

also facing the known bug on 6.3 not able to install packages on subscribed host

PR #9530, opened 2022-04-26
markers fix and subscription fix

tstrych:
fixing issue #9517, where I shoot myself into the leg,
marker fix was requested by @jyejare 

test `test_negative_without_attach_with_lce` will not pass, and I want to ask you to merge this anyway. 

But I feel sane now, finally.  TLDR I will remove these tests and I will implement them again in separate PR, so it doesn't matter 

this test is pure pain for around 2 years 

here https://github.com/SatelliteQE/robottelo/pull/4599/files the cli subscription tests were firstly introduced, 
and test_negative_without_attach_with_lce failed ever since I remember.  Only time when it was not failing, when I probably found some bug so I was lucky to fix the test hoping that solution is correct, but it probably wasn't.  

ENOUGH... of fixing the mess, It's such a mess, that test marked with cli_host_subscription, all of them should be re-implemented.

starting with NOT using `setup_org_for_a_rh_repo` in setup, which is making things so complicated, because inside of it the subscription is added to activation key,  so straight away you have to workaround your own solution, because you already enabled the repository set and synchronized repo and even added subscription to activation key via `activationkey_add_subscription_to_repo`

And now who should know, what is correct and what is not. When you register the host with lce instead of ak, should the repo still be enabled or not ? IDK the answer for that question. And you better not register with activation key because the subscription command that you want to test is useless if subscription is already there. I mean what the hell.


tstrych:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py -m cli_host_subscription

JacobCallahan (approved):
I'm here for this! Gut it and implement it how you think it should be done!

tests/foreman/cli/test_host.py
```diff
@@ -2309,7 +2301,7 @@ def test_positive_auto_attach(request, module_host_subscription, host_subscripti
     host = Host.info({'name': module_host_subscription.client.hostname})
     module_host_subscription._register_client(activation_key=activation_key)
     Host.subscription_auto_attach({'host-id': host['id']})
-    result = module_host_subscription.enable_repo(module_host_subscription.client.repository_id)
+    result = module_host_subscription.client.enable_repo(module_host_subscription.repository_id)
```
mshriver, line 2304: I don't think this method actually returns anything, unless there is another `enable_repo` defined somewhere?

https://github.com/SatelliteQE/robottelo/blob/67c8165fa670e36584fdd9f06ffb0af5aea51b80/robottelo/hosts.py#L248

It can be easily updated to return the result of `self.execute` so that you can assert against the status.

mshriver (approved):
@tstrych way to address an issue with the value and accuracy of the tests here head on instead of letting it continue to fail.

If you really wanted to take a deep dive on this test module, the class and functions defined within the test are trying to extend the CLI modeling classes for hammer. @jyejare and @JacobCallahan are looking to do some heavy refactoring around this pattern and abstraction of 'helpers' within the CLI tests.  You could start to take a heavier hand with moving that behavior into fixtures, and clearly separating test coverage from test setup/teardown.

You can also start to reference the CLI classes through `Satellite.cli.<EntityClassName>` instead of importing them in the test module, might be a good PR to make that switch with as its copy-paste one-line changes.

devendra104:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py -m cli_host_subscription

tests/foreman/cli/test_host.py
mshriver: Note this assertion was the only failure in Robottelo-Runner checks. 

`AttributeError: 'NoneType' object has no attribute 'status'`

tstrych:
Robotello runner should pass as the assert that was breaking the test was removed. The test_negative_without_attach_with_lce needs rework as mentioned it is planned for the future.

tstrych:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py -m cli_host_subscription

tests/foreman/cli/test_host.py
tstrych: yes I forget to remove it from all occurrences, it's fixed now.

tstrych: I believe it's better to: `subscription manager repos --enabled` before it's added and afterwards.

PR #9538, opened 2022-04-29
Fix cli user/usergroup tests

shubhamsg199:
`settings.configure` was used as a workaround for some configuration issues back then
This change fixes failing cli User and Usergroup tests

### DIFF
--- a/tests/foreman/cli/test_user.py
+++ b/tests/foreman/cli/test_user.py
@@ -58,7 +58,6 @@ class TestUser:
         saved on sat. roles is a dict so keys are role's id respective value is
         the role itself
         """
-        settings.configure()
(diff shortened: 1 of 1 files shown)


mshriver (approved):

mshriver:
@shubhamsg199 we really shouldn't be calling this method since we got rid of the old INI readers and compatibility layer.

@JacobCallahan would you agree he can take this PR one step further and remove the `configured_settings` fixture and all other calls to `settings.configure`?
https://github.com/SatelliteQE/robottelo/search?q=%27settings.configure%27

JacobCallahan:
@mshriver absolutely!!

@shubhamsg199 fantastic find! Please remove other instances of this call as well.

PR #9549, opened 2022-05-03
[6.9 cherry-pick] Skipping manifest refresh tests/setup

JacobCallahan:
Due to a change in the way manifests are handled in the portal, we need
to avoid refreshing manifests.
This change skips a number of tests and disables manifest refresh
behavior in a fixture that had dozens of downstream tests.


swadeley (approved):
ACK

PR #9551, opened 2022-05-04
Adding some branding tests

akhil-jha:
1. Adding some basic test to check correct version and 'Beta' check.

2. Depends on https://github.com/SatelliteQE/airgun/pull/701

### DIFF
--- a/tests/foreman/ui/test_branding.py
+++ b/tests/foreman/ui/test_branding.py
@@ -1,0 +1,41 @@
+"""Tests for branding
+
+:Requirement: Branding
+
+:CaseAutomation: Automated
+
+:CaseLevel: Acceptance
+
+:CaseComponent: Branding
+
+:Assignee: akjha
+
+:TestType: Functional
+
+:CaseImportance: High
+
+:Upstream: No
+"""
+from airgun.session import Session
+
+
+def test_verify_satellite_login_screen_info(target_sat):
+    """Verify Satellite version on the login screen
+
+    :id: f48110ad-29b4-49b1-972a-a70075a05732
+
+    :Steps: Get the info from the login screen
+
+    :expectedresults:
+        1. Correct Satellite version
+        2. If 'Beta' is present, should fail until a GA snap is received
+
+    :CaseLevel: System
+
+    :BZ: 1315849, 1367495, 1372436, 1502098, 1540710, 1582476, 1724738,
+         1959135, 2076979, 1687250, 1686540, 1742872, 1805642
+    """
+    with Session(login=False) as session:
+        version = session.login.read_sat_version()
+    assert target_sat.version in version['version']
+    assert 'Beta' not in version['version'], '"Beta" should not be there'


tests/foreman/ui/test_branding.py
```diff
@@ -0,0 +1,57 @@
...
+    :expectedresults: Correct version depending on the snap
+
+    :CaseLevel: System
+
+    :BZ: 1315849, 1367495, 1372436, 1502098, 1540710, 1582476, 1724738, 1959135, 2076979
+    """
+    result = default_sat.execute(
+        f'curl -k {default_sat.url}/users/login | '
+        f'grep -e "LoginPage" -e "{settings.server.version.release}"'
+    )
```
Gauravtalreja1, line 38: It would be nice if you could add some checks here for satellite/satellite-capsule RPMs as well or are you planning to add additional tests for it?

tests/foreman/ui/test_branding.py
```diff
@@ -0,0 +1,57 @@
...
+    assert 'LoginPage' in result.stdout
+    assert str(settings.server.version.release) in result.stdout
```
Gauravtalreja1, line 40: Considering that you already grepped these values in your command, asserting again isn't necessary here, right?

tests/foreman/ui/test_branding.py
```diff
@@ -0,0 +1,57 @@
...
+
+:CaseImportance: High
+
+:Upstream: No
+"""
+from robottelo.config import settings
+
+
+def test_check_version_login_screen(default_sat):
+    """Check correct version on the login screen
```
Gauravtalreja1, line 23:
```suggestion
    """Verify Satellite version on the login screen
```

tests/foreman/ui/test_branding.py
```diff
@@ -0,0 +1,57 @@
...
+
+    :id: f48110ad-29b4-49b1-972a-a70075a05732
+
+    :Steps: Curl the login screen
```
Gauravtalreja1, line 27:
```suggestion
    :Steps: Check Satellite version on login screen
```

tests/foreman/ui/test_branding.py
```diff
@@ -0,0 +1,57 @@
...
+
+    :expectedresults: Correct version depending on the snap
```
Gauravtalreja1, line 29:
```suggestion
    :expectedresults: Correct Satellite version depending on the snap
```

tests/foreman/ui/test_branding.py
```diff
@@ -0,0 +1,57 @@
...
+    :Steps: Curl the login screen
+
+    :expectedresults: Should fail until a GA snap is received
+
+    :CaseLevel: System
+
+    :BZ: 1687250, 1686540, 1742872, 1805642,
+    """
+    result = default_sat.execute(f'curl -k {default_sat.url}/users/login | grep "Beta"')
+    assert result.status == 1, '"Beta" should not be there'
```
Gauravtalreja1, line 57: I guess, you mean to assert this error when beta exists, and not if doesn't exist, right?

tests/foreman/ui/test_branding.py
```diff
@@ -0,0 +1,57 @@
...
+    assert 'LoginPage' in result.stdout
+    assert str(settings.server.version.release) in result.stdout
+
+
+def test_beta_on_login_screen(default_sat):
+    """Check 'Beta' on the login screen
+
+    :id: f4448b74-8b09-4e00-99a9-47c212b53eb5
+
+    :Steps: Curl the login screen
```
Gauravtalreja1, line 48:
```suggestion
    :Steps: Check Satellite version on login screen
```

tests/foreman/ui/test_branding.py
akhil-jha: Could you help me understand `checks here for satellite/satellite-capsule RPMs`? I wonder what you are saying falls under branding component.

tests/foreman/ui/test_branding.py
akhil-jha: These are just simple assertions. Sure, not necessary. But just being explicit here.

tests/foreman/ui/test_branding.py
akhil-jha: This test should fail until we get a GA snap.
I will have a cherry-pick for 6.10 wherein this test should pass.

tests/foreman/ui/test_branding.py
akhil-jha: same as above.

jyejare:
trigger: test-robottelo
pytest: tests/foreman/ui/test_branding.py

tests/foreman/ui/test_branding.py
swadeley: There is not much point in having one step in the docstring. How about:
Use curl to get the login screen
Use grep to get the version
Assert version in result.

If you think Line 35 is one step:
Use curl to get the login page and grep for the version.
Assert version in result.

tests/foreman/ui/test_branding.py
swadeley: Consistency is a good thing, so I vote to keep the assert. (If forced to choose, I would rather drop `-e "{settings.server.version.release}` and keep the assert line; but thats just my opinion)

swadeley (approved):
ACK with comments

tests/foreman/ui/test_branding.py
mshriver: I think its a lot better to assert this against `default_sat.version` instead of the settings value.

mshriver: Agreed with @swadeley here, this isn't a very strong assertion and is basically validating that `grep` works.

tests/foreman/ui/test_branding.py
mshriver: Why curl this page and not use widgetastic/browser integration to make these assertions?  There should be an element in the view that contains the release, and you should be able to read that widget.

akhil-jha:
trigger: test-robottelo
pytest: tests/foreman/ui/test_branding.py
airgun: 701

tests/foreman/ui/test_branding.py
akhil-jha: Yep, did that

tests/foreman/ui/test_branding.py
akhil-jha: new changes. Resolving this.

tests/foreman/ui/test_branding.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@pytest.fixture(scope='module')
+def get_sat_version():
+    with Session(login=False) as session:
+        version = session.login.read_sat_version()
+    return version
+
+
+def test_verify_satellite_version_login_screen(target_sat, get_sat_version):
```
mshriver, line 30: Personally I'd combine these two tests, I don't think separating them provides much value or test stability.

mshriver (approved):
Some suggestions made, but nothing blocking

akhil-jha:
trigger: test-robottelo
pytest: tests/foreman/ui/test_branding.py

mshriver (approved):

PR #9556, opened 2022-05-06
Add sleep time after get virtwho status to get the complete mapping info in rhsm.log

yanpliu:
Hey,
Will get the mapping info after get virtwho status, because of the network issue, it did not get the complete mapping info after restart virtwho immediately. so add sleep time can get the full mapping info.

Cases pass for hypervisor libvirt, esx, kubevirt, hyperv.

### DIFF
--- a/robottelo/virtwho_utils.py
+++ b/robottelo/virtwho_utils.py
@@ -220,12 +220,12 @@ def _get_hypervisor_mapping(hypervisor_type):
     :raises: VirtWhoError: If hypervisor_name is None.
     :return: hypervisor_name and guest_name
     """
-    logs, _ = wait_for(
-        get_rhsm_log,
-        fail_condition=lambda logs: "Host-to-guest mapping being sent to" not in logs,
+    wait_for(
+        lambda: 'Host-to-guest mapping being sent to' in get_rhsm_log(),
         timeout=10,
         delay=2,
     )
+    logs = get_rhsm_log()
     mapping = list()
     entry = None
     guest_name, guest_uuid = get_guest_info(hypervisor_type)
--- a/robottelo/virtwho_utils.py
+++ b/robottelo/virtwho_utils.py
@@ -140,7 +140,7 @@ def get_virtwho_status():
     """
     _, logs = runcmd('cat /var/log/rhsm/rhsm.log')
     error = len(re.findall(r'\[.*ERROR.*\]', logs))
-    ret, stdout = runcmd('systemctl status virt-who')
+    ret, stdout = runcmd('systemctl status virt-who; sleep 10')
     running_stauts = ['is running', 'Active: active (running)']
     stopped_status = ['is stopped', 'Active: inactive (dead)']
     if ret != 0:
--- a/robottelo/virtwho_utils.py
+++ b/robottelo/virtwho_utils.py
@@ -8,6 +8,7 @@ from fauxfactory import gen_integer
 from fauxfactory import gen_string
 from fauxfactory import gen_url
 from nailgun import entities
+from wait_for import wait_for
 
 from robottelo import ssh
 from robottelo.cli.base import Base
@@ -140,7 +141,7 @@ def get_virtwho_status():
     """
     _, logs = runcmd('cat /var/log/rhsm/rhsm.log')
     error = len(re.findall(r'\[.*ERROR.*\]', logs))
-    ret, stdout = runcmd('systemctl status virt-who; sleep 10')
+    ret, stdout = runcmd('systemctl status virt-who')
     running_stauts = ['is running', 'Active: active (running)']
     stopped_status = ['is stopped', 'Active: inactive (dead)']
     if ret != 0:
@@ -203,7 +204,15 @@ def get_configure_option(option, filename):
         raise VirtWhoError(f"option {option} is not exist or not be enabled in {filename}")
 
 
-def _get_hypervisor_mapping(logs, hypervisor_type):
+def get_rhsm_log():
+    """
+    Return the content of log file /var/log/rhsm/rhsm.log
+    """
+    _, logs = runcmd('cat /var/log/rhsm/rhsm.log')
+    return logs
+
+
+def _get_hypervisor_mapping(hypervisor_type):
     """Analysing rhsm.log and get to know: what is the hypervisor_name
     for the specific guest.
     :param str logs: the output of rhsm.log.
@@ -211,6 +220,12 @@ def _get_hypervisor_mapping(logs, hypervisor_type):
     :raises: VirtWhoError: If hypervisor_name is None.
     :return: hypervisor_name and guest_name
     """
+    logs, _ = wait_for(
+        get_rhsm_log,
+        fail_condition=lambda logs: "Host-to-guest mapping being sent to" not in logs,
+        timeout=10,
+        delay=2,
+    )
     mapping = list()
     entry = None
     guest_name, guest_uuid = get_guest_info(hypervisor_type)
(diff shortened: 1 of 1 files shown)


omkarkhatavkar:

JacobCallahan (approved):
Non-blocking question: would it be better to sleep before the status rather than after?

hkx303 (approved):

hkx303:
Actually, we'd better add a function to check if the mapping info is finished or not , let's consider it in the future

yanpliu:
> Non-blocking question: would it be better to sleep before the status rather than after?

Hello JacobCallahan,
In fact this sleep time should add before this code 
``` 
_, logs = runcmd('cat /var/log/rhsm/rhsm.log')
hypervisor_name, guest_name = _get_hypervisor_mapping(logs, hypervisor_type)
```
 in "def deploy_validation(hypervisor_type):" 
``` 
def deploy_validation(hypervisor_type):
    """Checkout the deploy result
    :param str hypervisor_type: esx, libvirt, rhevm, xen, libvirt, kubevirt
    :raises: VirtWhoError: If failed to start virt-who servcie.
    :ruturn: hypervisor_name and guest_name
    """
    status = get_virtwho_status()
    if status != 'running':
        raise VirtWhoError("Failed to start virt-who service")
    _, logs = runcmd('cat /var/log/rhsm/rhsm.log')
    hypervisor_name, guest_name = _get_hypervisor_mapping(logs, hypervisor_type)
    for host in Host.list({'search': hypervisor_name}):
        Host.delete({'id': host['id']})
    restart_virtwho_service()
    return hypervisor_name, guest_name
``` 

And it should also after run deploy command in "def deploy_configure_by_command(command, hypervisor_type, debug=False, org='Default_Organization'):"
``` 
def deploy_configure_by_command(command, hypervisor_type, debug=False, org='Default_Organization'):
    """Deploy and run virt-who servcie by the hammer command.

    :param str command: get the command by UI/CLI/API, it should be like:
        `hammer virt-who-config deploy --id 1 --organization-id 1`
    :param str hypervisor_type: esx, libvirt, rhevm, xen, libvirt, kubevirt
    :param bool debug: if VIRTWHO_DEBUG=1, this option should be True.
    :param str org: Organization Label
    """
    virtwho_cleanup()
    guest_name, guest_uuid = get_guest_info(hypervisor_type)
    Host.delete({'name': guest_name})
    register_system(get_system(hypervisor_type), org=org)
    ret, stdout = runcmd(command)
    if ret != 0 or 'Finished successfully' not in stdout:
        raise VirtWhoError(f"Failed to deploy configure by {command}")
    if debug:
        return deploy_validation(hypervisor_type)
``` 
In the deploy command it will restart virtwho and generate the mapping info in rhsm.log.
If there is no sleep time, when network is week,  "    _, logs = runcmd('cat /var/log/rhsm/rhsm.log')
    hypervisor_name, guest_name = _get_hypervisor_mapping(logs, hypervisor_type)" in "def deploy_validation(hypervisor_type):" , the rhsm.log did not get the include the mapping info.

JacobCallahan:
@yanpliu is that a change that you would like to make with this PR or another?

yanpliu:
> @yanpliu is that a change that you would like to make with this PR or another?

@JacobCallahan Have realized the changes in this PR. Add message check and time.sleep in _get_hypervisor_mapping method.

robottelo/virtwho_utils.py
```diff
@@ -203,14 +204,18 @@ def get_configure_option(option, filename):
...
     for the specific guest.
     :param str logs: the output of rhsm.log.
     :param str hypervisor_type: esx, libvirt, rhevm, xen, libvirt, kubevirt
     :raises: VirtWhoError: If hypervisor_name is None.
     :return: hypervisor_name and guest_name
     """
+    _, logs = runcmd('cat /var/log/rhsm/rhsm.log')
+    if "Host-to-guest mapping being sent to" not in logs:
+        time.sleep(20)
+        _, logs = runcmd('cat /var/log/rhsm/rhsm.log')
```
JacobCallahan, line 218: This could be a good case to use `wait_for` instead of a single set sleep time. You can find a number of examples in the robottelo codebase.

JacobCallahan (changes requested):
One small suggestion. Looking good overall!

robottelo/virtwho_utils.py
yanpliu: > This could be a good case to use `wait_for` instead of a single set sleep time. You can find a number of examples in the robottelo codebase.

Thank you @JacobCallahan for your suggestion.  Have used the wait_for instead of sleep time.

yanpliu:
@mshriver Could you please help to review and merge this PR? Thank you very much!

robottelo/virtwho_utils.py
```diff
@@ -203,14 +204,28 @@ def get_configure_option(option, filename):
...
+
+def _get_hypervisor_mapping(hypervisor_type):
     """Analysing rhsm.log and get to know: what is the hypervisor_name
     for the specific guest.
     :param str logs: the output of rhsm.log.
     :param str hypervisor_type: esx, libvirt, rhevm, xen, libvirt, kubevirt
     :raises: VirtWhoError: If hypervisor_name is None.
     :return: hypervisor_name and guest_name
     """
+    logs, _ = wait_for(
```
mshriver, line 223: This isn't quite right - `logs` is going to be empty, the label isn't assigned and wait_for doesn't return until the wait loop is complete.   `fail_condition` is not a function to be executed, the output of `func` is compared to `fail_condition` to evaluate whether or not to continue the loop.

Right now this is executing `get_rhsm_log`, which returns a string. `wait_for` is then checking `get_rhsm_log() == <fail_condition lambda>. 

You'll need to do something like this:

```
logs,  = wait_for(
    lambda: 'Host-to-guest mapping being sent to' in get_rhsm_log(),
    timeout=10,
    delay=2
```

That way, every iteration, it will call get_rhsm_log() and get a fresh copy of the log content, check whether the string is in it or not, resulting in a boolean. This boolean is then checked against fail_condition, which by default is `False`.  So if the string wasn't in the log, the func result is False, which is equal to the fail_condition, and the looping continues.

https://github.com/RedHatQE/wait_for/blob/master/wait_for/__init__.py#L146

mshriver (changes requested):
wait_for use needs adjusted.

robottelo/virtwho_utils.py
yanpliu: > This isn't quite right - `logs` is going to be empty, the label isn't assigned and wait_for doesn't return until the wait loop is complete. `fail_condition` is not a function to be executed, the output of `func` is compared to `fail_condition` to evaluate whether or not to continue the loop.
> 
> Right now this is executing `get_rhsm_log`, which returns a string. `wait_for` is then checking `get_rhsm_log() == <fail_condition lambda>.
> 
> You'll need to do something like this:
> 
> ```
> logs,  = wait_for(
>     lambda: 'Host-to-guest mapping being sent to' in get_rhsm_log(),
>     timeout=10,
>     delay=2
> ```
> 
> That way, every iteration, it will call get_rhsm_log() and get a fresh copy of the log content, check whether the string is in it or not, resulting in a boolean. This boolean is then checked against fail_condition, which by default is `False`. So if the string wasn't in the log, the func result is False, which is equal to the fail_condition, and the looping continues.
> 
> https://github.com/RedHatQE/wait_for/blob/master/wait_for/__init__.py#L146

@mshriver  when we use like this, the logs  value is True, in fact, we need to return the content of logs,  and we should analyze the log in the following code. 
In fact, whether the  'Host-to-guest mapping being sent to' in get_rhsm_log() or not , the content of logs from wait_for need to be returned to "logs" parameter.
Then we will analyze the logs in the following code.

yanpliu:
@mshriver I have updated the wait_for , could you please review it? Thank you very much!

mshriver (approved):

robottelo/virtwho_utils.py
mshriver: Thanks for making the update!  This is a consequence of the pattern in this case - you can't always get back the last value of the core function. There are tricky ways around it by having the `get_logs` function return a tuple, but its not really worth the complication. You did the right thing, just adding another call to `get_logs` after wait_for completes.

PR #9571, opened 2022-05-11
katello-ca installation for rhel8,9 fips hosts

pondrejk:
With system-wide crypto policies, fips enabled rhel8 and 9 hosts seem to be less forgiving than rhel7_fips. In this case, rpm -Uvh pulling straight from the url would be aborted with `curl: (60) SSL certificate problem: self signed certificate in certificate chain`. The proposed variant succeeds on these fips hosts.

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -376,7 +376,11 @@ class ContentHost(Host, ContentHostMixins):
         :raises robottelo.hosts.ContentHostError: If katello-ca wasn't
             installed.
         """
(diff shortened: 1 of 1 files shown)


mshriver:
@pondrejk flake is failing, looks like the line is too long. Make sure you have pre-commit setup in your dev env to avoid these failures.

jyejare (commented):
ACK pending checks failing !

JacobCallahan (approved):

mshriver (approved):

PR #9572, opened 2022-05-11
Fixed activation key test for deleting subscriptions 

ColeHiggins2:
Removed out of date assertions for deleting subscriptions on an activation key


ColeHiggins2:
Old PR got merged with new, making new PR

PR #9573, opened 2022-05-11
fixed activation key test for deleting subs

ColeHiggins2:
Removed out of date assertions for deleting subscriptions on an activation key


mshriver (approved):
Much better assertion pattern!

jyejare (approved):

PR #9584, opened 2022-05-13
Fix e2e bulk update test failure

spusateri:
This PR fixes an error that would appear when asserting against the uri, due to the equals sign `=` not being encoded properly. To fix this, I parsed the uri components individually using `urlparse()` and used `urlencode()` to properly format the assertion rather than hard coding it and running into potential similar errors in the future

### DIFF
--- a/tests/foreman/ui/test_contenthost.py
+++ b/tests/foreman/ui/test_contenthost.py
@@ -19,6 +19,7 @@
 import re
 from datetime import datetime
 from datetime import timedelta
+from urllib.parse import urlparse
 
 import pytest
 from airgun.session import Session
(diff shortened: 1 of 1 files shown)


spusateri:
trigger: robottelo
pytest: tests/foreman/ui/test_contenthost.py -k test_positive_end_to_end_bulk_update

spusateri:
For some reason I accidentally applied some cherry picked commits when rebasing, will resolve that shortly

spusateri:
I had to remove `urlencode` for the search query because the response returned by satellite is not fully encoded, which would cause the test to fail. I am open to suggestions on how to handle the partial encoding, but for now I think it would be sufficient just to hard code the `%3D` required to get the test to pass

spusateri:
Although this leaves me wondering what change in Satellite would have caused this singular `=` sign to be encoded, while other `=` in the URI are not.

mshriver (approved):

PR #9588, opened 2022-05-16
Rearrange uuid test case to correct class in file

latran:
@vsedmik suggest to putting this test case in the other class scope so it can be a better fit than in `capsule` space.  And I agree.


mshriver (approved):

vsedmik (approved):
ACK, thank you :+1:

PR #9590, opened 2022-05-16
automation for closed loop bz 2044631 for 6.11

latran:
Close loop bug for https://bugzilla.redhat.com/show_bug.cgi?id=2044631.   The CLI does not have a way to check the presence of the packages in a cv so I cannot automate.

Finally, I discovered a bug when I wrote this.  Waiting for it to be fixed before taking this out of draft.


latran:
added skip marker for https://bugzilla.redhat.com/show_bug.cgi?id=2086957

vsedmik (changes requested):
Just one change requested, otherwise looks good to me.

vsedmik (approved):

tests/foreman/ui/test_contentview.py
```diff
@@ -3652,3 +3652,125 @@ def test_positive_inc_update_should_not_fail(session, module_org):
...
+        'basearch': 'x86_64',
+        'releasever': None,
+    }
+    packages = [
+        'foreman-cli',
+        'katello-agent',
+        'katello-host-tools',
+        'katello-host-tools-fact-plugin',
+    ]
+    with manifests.clone() as manifest:
```
mshriver, line 3713: Why not use any of the existing fixtures for module cloning?

mshriver (approved):
Approved, but I think the manifest cloning could be done in setup via existing fixtures.

PR #9591, opened 2022-05-17
Settings test fix

shweta83:
Refactored Settings Tests and fixed test failures:
1. Renamed "Setting.py" to keep the naming consistency of component.
2. Fixed "test_negative_update_hostname_with_empty_fact" in UI
3. Fixed "test_negative_update_sync_timeout" and "test_positive_update_sync_timeout" in API


tests/foreman/ui/test_settings.py
```diff
@@ -499,8 +499,9 @@ def test_negative_update_hostname_with_empty_fact(session, setting_update):
     new_hostname = ""
     property_name = setting_update.name
     with session:
-        response = session.settings.update(property_name, new_hostname)
-        assert response is not None, "Value can't be blank"
+        with pytest.raises(AssertionError) as context:
+            session.settings.update(property_name, new_hostname)
+            assert is_valid_error_message(str(context.value))
```
tstrych, line 504: this is the only usage in the code of `is_valid_error_message` that I have found, so if not explained or not in future plans, please consider comparing it directly here

And right now the method compares error value with one of the three error values, it doesn't seem as a good concept, to find whether an error is one of such, IMO error should be compared directly

tstrych: three error values make more sense if each parametrized value provides a different error :)

tstrych:
trigger: test-robottelo
pytest: tests/foreman/api/test_settings.py::test_positive_update_sync_timeout tests/foreman/api/test_settings.py::test_negative_update_sync_timeout tests/foreman/api/test_settings.py::test_negative_update_hostname_with_empty_fact

tests/foreman/ui/test_settings.py
shweta83: @tstrych `is_valid_error_message` and `valid_error_messages` only used once so I removed these and compared the failures in test itself.

tests/foreman/ui/test_settings.py
```diff
@@ -499,8 +481,9 @@ def test_negative_update_hostname_with_empty_fact(session, setting_update):
     new_hostname = ""
     property_name = setting_update.name
     with session:
-        response = session.settings.update(property_name, new_hostname)
-        assert response is not None, "Value can't be blank"
+        with pytest.raises(AssertionError) as context:
```
jyejare, line 484: Are you sure about `AssertionError` here? I think it should be an error that `session.settings.update` will throw, could be an HTTP Error !

jyejare (commented):
ACK pending questions!

tstrych (approved):
ACK

tests/foreman/ui/test_settings.py
```diff
@@ -35,24 +35,6 @@ def invalid_settings_values():
     return [' ', '-1', 'text', '0']
 
 
-def valid_error_messages():
```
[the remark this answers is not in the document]
mshriver: checked, yep only called in this test module once.

tests/foreman/ui/test_settings.py
mshriver: I think its appropriate, the `update` method hits `assert_no_errors`, and is specifically expecting that behavior here, not for any exception type to occur.

https://github.com/SatelliteQE/airgun/blob/0ad943f3d4149f85ecf9533bc6965ba0e4e18aca/airgun/widgets.py#L780

If the exception is an HTTPError (not AssertionError), something has definitely failed, and the test will end in ERROR because of the unhandled exception.

In the case of the AssertionError, it should contain the text of the message in the UI, which is then validated against.

mshriver (approved):

PR #9603, opened 2022-05-24
refactor1: replaced the entities with target.sat

omkarkhatavkar:
At the test level, `entities` should be replaced with `target_sat.api `. This PR only covers at test level, further PR will update the common functions where the entities are used in destructive tests.

### DIFF
--- a/tests/foreman/api/test_contentview.py
+++ b/tests/foreman/api/test_contentview.py
@@ -754,7 +754,7 @@ class TestContentViewPublishPromote:
 
         :CaseAutomation: Automated
         """
-        org = entities.Organization().create()
+        org = target_sat.api.Organization().create()
         with manifests.clone() as manifest:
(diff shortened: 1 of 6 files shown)


tests/foreman/api/test_contentview.py
```diff
@@ -754,7 +754,7 @@ def test_positive_reboot_recover_cv_publish(self, target_sat):
 
         :CaseAutomation: Automated
         """
-        org = entities.Organization().create()
```
mshriver, line 757: Any reason why you're only updating some test functions in this module? I'd say go for all of them so that you can remove the import.

mshriver (approved):

tests/foreman/api/test_contentview.py
omkarkhatavkar: currently planned to update only destructive tests.

PR #9612, opened 2022-05-25
Refactoring positive_create_mirror_on_sync

sambible:
Mirror on sync option has changed to "mirroring-policy" - this refactors the test to reflect that, and for good measure, tests all mirroring-policy options as well.

### DIFF
--- a/robottelo/constants/__init__.py
+++ b/robottelo/constants/__init__.py
@@ -324,6 +324,11 @@ DOWNLOAD_POLICIES = {
     'immediate': "Immediate",
 }
 
+MIRRORING_POLICIES = {
+    'additive': "Additive",
+    'mirror_complete': "Complete Mirroring",
+    'mirror_content_only': "Content Only",
+}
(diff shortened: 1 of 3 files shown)


robottelo/constants/__init__.py
```diff
@@ -324,6 +324,11 @@
     'immediate': "Immediate",
 }
 
+MIRRORING_POLICIES = {
+    'additive': "Additive",
+    'mirror_complete': "Complete Mirroring",
+    'mirror_content_only': "Content Only",
```
jyejare, line 330: Use single quotes !

jyejare (approved):

robottelo/constants/__init__.py
Griffin-Sullivan: I think he was probably matching the dictionaries on the lines below?

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_repository.py -k test_positive_mirroring_policy

robottelo/constants/__init__.py
sambible: Yeah, I was definitely style matching, doesn't make it right :)

PR #9620, opened 2022-05-27
Add install_rhc in rh_cloud.yaml.template

jameerpathan111:
Dependency for satelliteqe-jenkins MR 578


mshriver (approved):

mshriver:
We don't have hard rules about the validators yet, but it would be a good idea to add a validator for this field, even if its very basic.

devendra104 (approved):

PR #9625, opened 2022-05-27
Create upgrade test for db seed mismatch failure

spusateri:
This PR is created to add an upgrade scenario for this BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2043705

As of the creation of this PR, the pre upgrade scenario is passing but the post upgrade scenario fails with a `TypeError`:

`>       assert org_id == chost[0].organization.id`
`E       TypeError: 'Host' object is not subscriptable`

However, I wanted to open this as a draft so I could start getting feedback on the rest of it

### DIFF
--- a/tests/upgrades/test_hostcontent.py
+++ b/tests/upgrades/test_hostcontent.py
@@ -19,6 +19,8 @@
 import pytest
 from fabric.api import run
 from nailgun import entities
+from upgrade_tests.helpers.scenarios import create_dict
+from upgrade_tests.helpers.scenarios import get_entity_data
 
 
 class TestScenarioDBseedHostMismatch:
@@ -36,40 +38,43 @@ class TestScenarioDBseedHostMismatch:
         8. Upgrade Satellite
         9. Ensure upgrade succeeds
 
-    BZ: 2043705
+    BZ: 2043705, 2028786, 2019467
     """
 
     @pytest.mark.pre_upgrade
-    def test_pre_db_seed_host_mismatch(self, rhel7_contenthost):
+    def test_pre_db_seed_host_mismatch(self, default_sat, rhel7_contenthost):
         """
 
         :id:
 
         :steps:
-            1. Create an Organizaiton1 and Location1
-            2. Create another Organization2 in another Location2
-            3. Create a content host on Organization1 but not in Location2
+            1. Create an Organizaiton1 and Location
+            2. Create another Organization2
+            3. Create a content host on Organization1
+            4. Use rake console to edit the content host to Org2
+            5. Ensure the taxonomy between the content host and Location is broken
 
         :expectedresults:
-            1. The Content Host exists on Organization but not Location
+            1. The content host should be in Organization2 but not Location
 
-        :BZ: 2043705
+        :BZ: 2043705, 2028786, 2019467
+
+
 
         :customerscenario: true
         """
-        org = entities.Organization.create()
-        loc = entities.Location.create(organization=org)
-
-        org2 = entities.Organization.create()
+        org = default_sat.Organization.create()
+        loc = default_sat.Location.create(organization=org)
+        org2 = default_sat.Organization.create()
 
         rhel7_contenthost.install_capsule_katello_ca(capsule=self.proxy_name)
         rhel7_contenthost.register_contenthost(org=org, lce='Library')
 
-        chost = entities.Host().search(query={'search': f'name="{rhel7_contenthost.hostname}"'})
+        chost = default_sat.Host().search(query={'search': f'name="{rhel7_contenthost.hostname}"'})
 
-        assert chost['organization'] == org.name
-        assert chost['organization'] != org2.name
-        assert chost['location'] == loc.name
+        assert chost[0]['organization'] == org.name
+        assert chost[0]['organization'] != org2.name
+        assert chost[0]['location'] == loc.name
 
         rake_host = f"host = Host.where(:name => '{chost.name}')"
         rake_organization = f'; host.organization_id=${org2.id}'
@@ -81,8 +86,13 @@ class TestScenarioDBseedHostMismatch:
         assert 'true' in result
         assert chost['organization'] == org2.name
 
+        global_dict = {self.__class__.__name__: {'client_name': rhel7_contenthost.hostname}}
+        create_dict(global_dict)
+
         # do the upgrade
 
     @pytest.mark.post_upgrade(depend_on=test_pre_db_seed_host_mismatch)
     def test_post_db_seed_host_mismatch(self, default_sat):
         """"""
+
+        # ensure the upgrade succeeds and is in org 1
--- a/tests/upgrades/test_hostcontent.py
+++ b/tests/upgrades/test_hostcontent.py
@@ -106,4 +106,5 @@ class TestScenarioDBseedHostMismatch:
         assert org_id == chost[0].organization.id
         assert loc_id == chost[0].location.id
 
-        VMBroker().from_inventory(f'hostname<{chostname}').checkin()
+        match = VMBroker().from_inventory(f'hostname<{chostname}')
+        VMBroker(hosts=match).checkin()
--- a/tests/upgrades/test_hostcontent.py
+++ b/tests/upgrades/test_hostcontent.py
@@ -1,4 +1,4 @@
-"""Test Hosts Content related Upgrade Scenario's
+"""Test Hosts Content related Upgrade Scenarios
 
 :Requirement: UpgradedSatellite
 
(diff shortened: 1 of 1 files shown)


spusateri:
I know @mshriver previously recommended removing the usage of `entities.Host` in favor of `target_sat.api`
But this fails due to this error:
```
chost = target_sat.api.Host(query={'search': f'name="{rhel7_contenthost.hostname}"'})
    
>       assert org_id == chost.organization.id
E       AttributeError: 'DecClass' object has no attribute 'organization'
```
I would much prefer to use `target_sat.api` if possible, so any suggestions regarding this issue would be greatly appreciated

spusateri:
Thought I screwed up force pushing the changes so I apologize for the messy commit history :spaghetti: 
But I've got this passing locally and think this PR is now ready to move out of draft and get some feedback. 

One thing I know is missing is the `:id:` metadata field, but I was unsure exactly what this id was actually tied to

mshriver:
@spusateri :id: should get auto-generated by pre-commit scripts.

tests/upgrades/test_hostcontent.py
```diff
@@ -0,0 +1,112 @@
...
+import pytest
+from broker.broker import VMBroker
+from upgrade_tests.helpers.scenarios import create_dict
+from upgrade_tests.helpers.scenarios import get_entity_data
+
+from robottelo.hosts import ContentHost
+
+
+class TestScenarioDBseedHostMismatch:
+    """This test scenenario veryfies that the upgrade succeeds even when inconsistencies exist
```
JacobCallahan, line 28:
```suggestion
    """This test scenario verifies that the upgrade succeeds even when inconsistencies exist
```

tests/upgrades/test_hostcontent.py
```diff
@@ -0,0 +1,112 @@
...
+        loc = target_sat.api.Location().create()
+
+        chost_vm = VMBroker(nick='rhel7', host_classes={'host': ContentHost}).checkout()
+
+        chost_vm.install_katello_ca(target_sat)
+        chost_vm.register_contenthost(org=org.label, lce='Library')
+
+        chost = target_sat.api.Host().search(query={'search': chost_vm.hostname})
+
+        assert chost[0].organization.id == org.id
```
JacobCallahan, line 75: Would this work instead?
```suggestion
        assert chost_vm.nailgun_host.organization.id == org.id
```

tests/upgrades/test_hostcontent.py
```diff
@@ -0,0 +1,112 @@
...
+        rake_host = f"host = ::Host.find({chost[0].id})"
+        rake_organization = f"; host.location_id={loc.id}"
+        rake_host_save = "; host.save!"
+        result = target_sat.run(
+            f"echo '{rake_host}{rake_organization}{rake_host_save}' | foreman-rake console"
+        )
+
+        assert 'true' in result.stdout
+        chost = target_sat.api.Host().search(query={'search': chost_vm.hostname})
+        assert chost[0].location.id == loc.id
```
JacobCallahan, line 87: how about this?
```suggestion
        assert chost_vm.nailgun_host.location.id == loc.id
```

tests/upgrades/test_hostcontent.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        chostname = get_entity_data(self.__class__.__name__)['client_name']
+        org_id = get_entity_data(self.__class__.__name__)['organization_id']
+        loc_id = get_entity_data(self.__class__.__name__)['location_id']
+        chost = target_sat.api.Host().search(query={'search': chostname})
+
+        assert org_id == chost[0].organization.id
+        assert loc_id == chost[0].location.id
+
+        VMBroker(host=chost).checkin()
```
JacobCallahan, line 112: You may be able to try this instead. Otherwise, I don't see how that checkin would work.
```suggestion
        VMBroker.from_inventory(f"hostname<{chostname}").checkin()
```

JacobCallahan (changes requested):
Just a few things

tests/upgrades/test_hostcontent.py
```diff
@@ -0,0 +1,112 @@
...
+    in the database between Organization, Location and Content Host.
+
+    Test Steps:
+
+        1. Before Satellite upgrade
+        2. Create a New Organization and Location
+        3. Create a Content Host in the Organization
+        4. Ensure the Org is not in the Location
+        4. Assign the Content Host to the Location using the rake console
+        7. Ensure the Host is in both but the Org is not in Location, creating a mismatch
```
jyejare, line 38: Make docs should fail here ! The numbering inconsistencies !

tests/upgrades/test_hostcontent.py
```diff
@@ -0,0 +1,112 @@
...
+class TestScenarioDBseedHostMismatch:
+    """This test scenenario veryfies that the upgrade succeeds even when inconsistencies exist
+    in the database between Organization, Location and Content Host.
+
+    Test Steps:
+
+        1. Before Satellite upgrade
+        2. Create a New Organization and Location
+        3. Create a Content Host in the Organization
+        4. Ensure the Org is not in the Location
```
jyejare, line 36:
```suggestion
        4. Ensure the Location is not in the Org
```

tests/upgrades/test_hostcontent.py
jyejare, line 38:
```suggestion
        7. Ensure the Host is in both but the Location is not in Org, creating a mismatch
```

tests/upgrades/test_hostcontent.py
```diff
@@ -0,0 +1,112 @@
...
+        8. Upgrade Satellite
+        9. Ensure upgrade succeeds
+
+    BZ: 2043705, 2028786, 2019467
+    """
+
+    @pytest.mark.pre_upgrade
+    def test_pre_db_seed_host_mismatch(self, target_sat):
+        """
+        :id: 28861b9f-8abd-4efc-bfd5-40b7e825a941
```
jyejare, line 48:
```suggestion
        :id: preupgrade-28861b9f-8abd-4efc-bfd5-40b7e825a941
```

tests/upgrades/test_hostcontent.py
jyejare: Do this change all over the PR!

tests/upgrades/test_hostcontent.py
```diff
@@ -0,0 +1,112 @@
...
+
+        :expectedresults:
+            1. The Content Host is assigned to both Location and Org, but Org is not in Location
+
+        :BZ: 2043705, 2028786, 2019467
+
+        :customerscenario: true
+        """
+        org = target_sat.api.Organization().create()
+        loc = target_sat.api.Location().create()
```
jyejare, line 66: You can use function_org and function_location global fixtures !

tests/upgrades/test_hostcontent.py
```diff
@@ -0,0 +1,112 @@
...
+
+        chost_vm = VMBroker(nick='rhel7', host_classes={'host': ContentHost}).checkout()
```
jyejare, line 68: I think we are no more using `nick` based checkouts since we are directly passing argumets to parameters. Check other tests in robottelo.

jyejare: You can also use `rhel7_contenthost` global content host fixture which is not checking in the 
 content host!

tests/upgrades/test_hostcontent.py
```diff
@@ -0,0 +1,112 @@
...
+        create_dict(global_dict)
+
+    @pytest.mark.post_upgrade(depend_on=test_pre_db_seed_host_mismatch)
+    def test_post_db_seed_host_mismatch(self, target_sat):
+        """Check whether the upgrade succeeds, and ensure Content Host
+        exists on the Satellite and has not had any attributes changed
+        """
+        chostname = get_entity_data(self.__class__.__name__)['client_name']
+        org_id = get_entity_data(self.__class__.__name__)['organization_id']
+        loc_id = get_entity_data(self.__class__.__name__)['location_id']
```
jyejare, line 106: You can call `get_entity_data` once in var and then use data from that var!

jyejare (approved):
ACK with non-blocking improvement comments!

tests/upgrades/test_hostcontent.py
spusateri: Hi, I can't seem to find any examples in robottelo that use VMBroker without `nick`, can you elaborate on how I can address this? 

I think we had a discussion about using `rhel7_contenthost` in the automation channel the other day, but perhaps I misunderstood. As I understood it `rhel7_contenthost` is session scoped and would not be available to the `post` scenario since `pre` and `post` are run in separate sessions. I think you had recommended calling broker explicitly to `checkout()` the vm in pre and then to `checkin()` at the end of post, but if there is a better way to handle this situation please let me know and I can fix it

jyejare: @spusateri Doesn't matters if it's a session-scoped but the good part is it doesn't have teardown after its yield also the upgrade scenario job is not `checking_in` all after pre-upgrade tests execution. You can try executing the test with `rhel7_contenthost` locally as well.

So you should be good to use `rhel7_contenthost`.

Fo what explicit parameters we use now with VMBroker, you can see the conf directory keys that we are returning in this def https://github.com/SatelliteQE/robottelo/blob/master/pytest_fixtures/core/contenthosts.py#L16

tests/upgrades/test_hostcontent.py
```diff
@@ -0,0 +1,109 @@
...
+            }
+        }
+
+        create_dict(global_dict)
+
+    @pytest.mark.post_upgrade(depend_on=test_pre_db_seed_host_mismatch)
+    def test_post_db_seed_host_mismatch(self, target_sat):
+        """Check whether the upgrade succeeds, and ensure Content Host
+        exists on the Satellite and has not had any attributes changed
+        """
```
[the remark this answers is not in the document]
spusateri: Updated in last commit, thanks @latran

tests/upgrades/test_hostcontent.py
spusateri: I think I've addressed all of these changes in the commits below

latran (approved):
ack

JacobCallahan (approved):

PR #9626, opened 2022-05-29
Parameterizing for rhel9 and some test fixes

akhil-jha:
Parameterizing some errata tests for rhel9.
Adding rhel9 in all the tests would require refactoring accorss api/cli/ui.
This should atleast cover **SAT-5916** card.

Fixing some test failures.


akhil-jha:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py

akhil-jha:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py

akhil-jha:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py

akhil-jha:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py

akhil-jha:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py

akhil-jha:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py

tests/foreman/api/test_errata.py
```diff
@@ -183,9 +188,7 @@ def test_positive_install_in_hc(module_org, activation_key, custom_repo, target_
 
     :BZ: 1983043
     """
-    with VMBroker(
-        nick=constants.DISTRO_RHEL7, host_classes={'host': ContentHost}, _count=2
-    ) as clients:
+    with VMBroker(nick=distro, host_classes={'host': ContentHost}, _count=2) as clients:
```
Gauravtalreja1, line 191: Here, instead of calling VMBroker directly, you could try using the [content_hosts](https://github.com/SatelliteQE/robottelo/blob/master/pytest_fixtures/core/contenthosts.py#L93) fixture which also returns 2 hosts and use it with a loop similarly `for client in content_hosts:`

akhil-jha:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py

akhil-jha:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py

akhil-jha:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py

akhil-jha:
Test failing for above PRT are intermittent. Will be doing a test_errata refactor and will be fixed in that.

Gauravtalreja1 (approved):
@akhil-jha Thanks for including `content_hosts` fixture in pytest_generate_tests for parameterization

LGTM, will ask Tier2 reviewers to review again.

pytest_plugins/fixture_markers.py
```diff
@@ -6,7 +6,10 @@
 
 
 def pytest_generate_tests(metafunc):
-    if 'rhel_contenthost' in metafunc.fixturenames:
+    content_host_fixture = ''.join(
```
mshriver, line 9: This pattern is a bit odd, if both fixtures are used this will end up concatenating their names and failing the check.

I think what you really want to do is this, which will be true if either string is in the fixturenames list.
```
if any([True for f in ['rhel_contenthost', 'content_hosts'] if f in metafunc.fixturenames])
```

akhil-jha: Hmm...I took it under the impression that only 1 of the fixture will be called at any given point. 
I forgot it also could be `def test(rhel_contenthost, content_hosts)`.

mshriver: Spoke a bit too soon - you want to pick the function name to use in the parametrization below as well.  I'd still avoid joining the function names. You can use comprehension and then just check that the list is non-empty.

pytest_plugins/fixture_markers.py
```diff
@@ -6,7 +6,10 @@
...
+        [i for i in ['rhel_contenthost', 'content_hosts'] if i in metafunc.fixturenames]
+    )
+    if content_host_fixture in metafunc.fixturenames:
```
mshriver, line 12: @JacobCallahan I'd like your thoughts here, because I think this is an anti-pattern, trying to statically set a list of function names here that should have parametrization applied during test generation.

I'm not sure markers will really work for this, because if we apply the mark within the fixture function it won't be added to the node until after this generation function is complete.

JacobCallahan: @mshriver certainly a valid concern.
@akhil-jha is there another way you could do this?

PR #9631, opened 2022-05-30
Remove old longrun/test_remote_DB.py module

Gauravtalreja1:
**Description:**
1. Removing the old longrun/test_remote_DB.py module since backup/restore tests are already covered in [test_backup_restore.py](https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/maintain/test_backup_restore.py) and installer tests are covered by RemoteDB workflows.
2. Moving the remaining Offload-DB test case to a new module under the `maintain` directory.
3. Correcting `:CaseComponent:` for backup/restore tests


tests/foreman/longrun/test_remote_DB.py
```diff
@@ -1,239 +0,0 @@
...
-
-    :CaseLevel: Integration
-
-    :CaseComponent: Installer
-    """
-
-
-@pytest.mark.stubbed
-@pytest.mark.tier1
-def test_upgrade_satellite_eith_all_external_db_SSL():
```
vsedmik, line 223: Do we have this covered in upgrades? If not, we might want to move this test case to `test_offload_DB.py` as well.

vsedmik (approved):
Just one question, otherwise ACK.

tests/foreman/longrun/test_remote_DB.py
Gauravtalreja1: @vsedmik We have a Jenkins job running for remotedb upgrades, so yes, it's covered as well :)

omkarkhatavkar (approved):

PR #9633, opened 2022-05-31
Fix for broker teardown error

adarshdubey-star:
rh_cloud inventory test cases were failing because of `broker teardown error` the issue was prevailing because of 
`return entities.Host().search(query={'search': self.hostname})[0]` this, as it was pointing towards the satellite in the server.yaml/local-sat, not the satellite got checked out by the fixture. Hence finding out the hosts at the wrong place to check out was resulting in the above error.

Thanks to @JacobCallahan for helping with another investigation!

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -126,6 +126,7 @@ class ContentHost(Host, ContentHostMixins):
 
         :param hostname: The fqdn of a ContentHost target
         :param auth: ('root', 'rootpass') or '/path/to/keyfile.rsa'
+        :param satellite: optional parameter satellite object.
         """
         if not hostname:
             raise ContentHostError('A valid hostname must be provided')
@@ -135,13 +136,20 @@ class ContentHost(Host, ContentHostMixins):
         elif isinstance(auth, str):
             # key file based authentication
             kwargs.update({'key_filename': auth})
+        self._satellite = kwargs.get('satellite')
         super().__init__(hostname=hostname, **kwargs)
+
+    @property
+    def satellite(self):
+        if not self._satellite:
+            self._satellite = Satellite()
+        return self._satellite
 
     @property
     def nailgun_host(self):
         """If this host is subscribed, provide access to its nailgun object"""
         if self.subscribed:
-            return entities.Host().search(query={'search': self.hostname})[0]
+            return self.satellite.api.Host().search(query={'search': self.hostname})[0]
 
     @property
     def subscribed(self):
@@ -376,6 +384,7 @@ class ContentHost(Host, ContentHostMixins):
         :raises robottelo.hosts.ContentHostError: If katello-ca wasn't
             installed.
         """
+        self._satellite = satellite
         self.execute(
             f'curl --insecure --output katello-ca-consumer-latest.noarch.rpm \
                     {satellite.url_katello_ca_rpm}'
@@ -404,6 +413,7 @@ class ContentHost(Host, ContentHostMixins):
         if result.status == 0:
             raise ContentHostError(f'katello-ca rpm(s) are still installed: {result.stdout}')
         self.execute('subscription-manager clean')
+        self._satellite = None
 
     def install_capsule_katello_ca(self, capsule=None):
         """Downloads and installs katello-ca rpm on the broker virtual machine.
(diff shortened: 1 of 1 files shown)


robottelo/hosts.py
```diff
@@ -135,13 +136,20 @@ def __init__(self, hostname, auth=None, **kwargs):
         elif isinstance(auth, str):
             # key file based authentication
             kwargs.update({'key_filename': auth})
+        self._satellite = kwargs.get('satellite')
         super().__init__(hostname=hostname, **kwargs)
 
+    @property
+    def satellite(self):
```
jyejare, line 143: I really dont like this approach where we are actually traversing back from the parent object to its child object! Infact, we should initiate host object and play with its properties in satellite class.

robottelo/hosts.py
```diff
@@ -135,13 +136,20 @@ def __init__(self, hostname, auth=None, **kwargs):
...
+        if not self._satellite:
+            self._satellite = Satellite()
```
jyejare, line 145: if you dont provide the satellite hostname argument here, this will always point to the `default_sat` / `setting.server.hostname` which is not being set by satellite_factory(in case you of destructive sat).

Ofcourse you want it to point to destructive_sat since the host will belong to destructive_sat that time not to default_sat.

jyejare (changes requested):
Change the approach for reading the nailgun host from satellite.

robottelo/hosts.py
JacobCallahan: I disagree. It is quite useful for a registered content host to have a reference to the satellite it is registered to. An example of that use is provided in this PR's change to the nailgun_host property.

robottelo/hosts.py
```diff
@@ -135,13 +136,20 @@ def __init__(self, hostname, auth=None, **kwargs):
...
+        return self._satellite
+
     @property
     def nailgun_host(self):
         """If this host is subscribed, provide access to its nailgun object"""
         if self.subscribed:
-            return entities.Host().search(query={'search': self.hostname})[0]
+            return self.satellite.api.Host().search(query={'search': self.hostname})[0]
```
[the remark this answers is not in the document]
JacobCallahan: I disagree. This is a common pattern in robottelo and quite a useful one. Offloading that to the Satellite class creates a disconnect between attributes.

robottelo/hosts.py
```diff
@@ -376,6 +384,7 @@ def install_katello_ca(self, satellite):
         :raises robottelo.hosts.ContentHostError: If katello-ca wasn't
             installed.
         """
+        self._satellite = satellite
```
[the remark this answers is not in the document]
JacobCallahan: This is used as part of the new satellite property introduced in this PR.

JacobCallahan (approved):
ACK, this is a nice improvement!

jyejare:
@adarshdubey-star Please run PRT with destructive marker test which is using nailgun_host property or test locally.

robottelo/hosts.py
mshriver: I don't think the property should resolve this automatically. If a sat instance wasn't passed on instantiation, this function should return `None`. Having a setter would be appropriate as well, so a sat can be set on the host after instantiation.

Not all host instances will necessarily have a satellite that they're registered against. With this behavior, its hard/impossible to tell.

robottelo/hosts.py
```diff
@@ -135,13 +136,20 @@ def __init__(self, hostname, auth=None, **kwargs):
...
+    @property
+    def satellite(self):
+        if not self._satellite:
+            self._satellite = Satellite()
+        return self._satellite
+
     @property
     def nailgun_host(self):
         """If this host is subscribed, provide access to its nailgun object"""
         if self.subscribed:
```
mshriver, line 151: self.subscribed doesn't necessarily indicate that its subscribed to a satellite instance, or that `self.satellite` has been set.  Taking my comment about whether `self.satellite` property should automatically inject the default satellite or return `None`, and maybe this function should check `self.subscribed and self.satellite is not None`.

As far as maintaining current assumptions, you're on track. I think there's a bit of a problem with the assumptions already in `master` branch.

robottelo/hosts.py
mshriver, line 387: What about unsetting `self._satellite` on `remove_katello_ca`?

mshriver (approved):
After some extensive discussion internally related to the comments on this PR, I'm ready to accept pretty much as-is.

My comment about unsetting `self._satellite` on `remove_katello_ca` is still relevant and an easy addition.

JacobCallahan (approved):

PR #9639, opened 2022-06-01
Fix ui filter test and Update PERMISSIONS_UI

shubhamsg199:
The default page size is 20 while reading the filters from the role.
Added a fix to increase the page size to 25 as the test asserts to check if 21 filters are created or not.

**`Test Results:`**
```
$ pytest -k test_positive_create_with_21_filters tests/foreman/ui/test_role.py                                                                                                                                                                                  
tests/foreman/ui/test_role.py .                                                                                                                                                                                                 [100%]
=================================== 1 passed, 9 deselected in 732.89s (0:12:12) =======================================
```
Dependent PR: https://github.com/SatelliteQE/airgun/pull/706


shubhamsg199:
trigger: test-robottelo
pytest: -k 'test_positive_create_with_21_filters' tests/foreman/ui/test_role.py

pondrejk (approved):
ack, but next time it would be nice to have the two changes in separate commits, just in case we would need to revert one of them

tests/foreman/ui/test_role.py
```diff
@@ -448,6 +448,10 @@ def test_positive_create_with_21_filters(session):
             session.filter.create(
                 role_name, {'resource_type': resource_type, 'permission.assigned': [permission]}
             )
+        # setting the page size to 25 as 20 is default
+        view = session.filter.navigate_to(session.filter, 'All', role_name=role_name)
+        per_page_value = view.pagination.per_page
```
mshriver, line 453: You can make this call directly instead of creating a new variable. `view.pagination.per_page.fill(25)`

mshriver (approved):

mshriver (approved):

PR #9646, opened 2022-06-01
Remove usage of docker_tags_whitelist

sambible:
A bit tardy to the party on swapping out docker_tags_whitelist for include_tags, but this should fix a few more test errors. The disabled test was exercising some behavior of docker_tags_whitelist that is not replicated by include_tags, since it doesn't 1 to 1 match the behavior. 

At some point it would be good to refactor/add tests to check the behavior that's now split into include_tags and exclude_tags, but that can be a separate PR

### DIFF
--- a/tests/foreman/api/test_repository.py
+++ b/tests/foreman/api/test_repository.py
@@ -1808,7 +1808,7 @@ class TestDockerRepository:
             {
                 constants.CONTAINER_UPSTREAM_NAME: {
                     'content_type': 'docker',
-                    'docker_tags_whitelist': ['latest'],
+                    'include_tags': ['latest'],
                     'docker_upstream_name': constants.CONTAINER_UPSTREAM_NAME,
                     'name': gen_string('alphanumeric', 10),
                     'url': constants.CONTAINER_REGISTRY_HUB,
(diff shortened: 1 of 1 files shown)


tests/foreman/api/test_repository.py
```diff
@@ -1817,60 +1817,61 @@ def test_negative_synchronize_private_registry_no_passwd(
...
-
-        tags = ['latest']
-        repo.docker_tags_whitelist = tags
-        repo.update(['docker_tags_whitelist'])
-        repo.sync()
-        repo = repo.read()
-
-        assert repo.docker_tags_whitelist == tags
-        assert repo.content_counts['docker_tag'] >= 2
+    # TODO: Rework test to reflect current include/exclude behavior
```
JacobCallahan, line 1834: you can just mark this test as skipped

JacobCallahan (changes requested):

tests/foreman/api/test_repository.py
```diff
@@ -1817,60 +1817,61 @@ def test_negative_synchronize_private_registry_no_passwd(
         ),
         indirect=True,
     )
-    def test_positive_synchronize_docker_repo_with_tags_whitelist(self, repo_options, repo):
-        """Check if only whitelisted tags are synchronized
+    def test_positive_synchronize_docker_repo_with_included_tags(self, repo_options, repo):
+        """Check if only included tags are synchronized
 
         :id: abd584ef-f616-49d8-ab30-ae32e4e8a685
```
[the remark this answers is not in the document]
mshriver: btw if you just remove the line, pre-commit will generate and add one for you automatically.

tests/foreman/api/test_repository.py
sambible: that does make more sense doesn't it

PR #9660, opened 2022-06-06
katello-ca installation from rhel7 sat to rhel8 fips client

pondrejk:
Turns out https://github.com/SatelliteQE/robottelo/pull/9571 only helps when pulling from Satellite on rhel8, I suppose it has to do with rpm version shipped with rhel7. A cleaner solution would be to stop relying on the katello-ca rpm altogether and use global registration api (works well on all fips hosts), but that would be a bigger effort. Anyway, this PR should work until we get there.

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -389,13 +389,18 @@ class ContentHost(Host, ContentHostMixins):
             f'curl --insecure --output katello-ca-consumer-latest.noarch.rpm \
                     {satellite.url_katello_ca_rpm}'
         )
-        self.execute('rpm -Uvh katello-ca-consumer-latest.noarch.rpm')
+        # check if the host is fips-enabled
+        result = self.execute('sysctl crypto.fips_enabled')
+        if 'crypto.fips_enabled = 1' in result.stdout:
+            self.execute('rpm -Uvh --nodigest --nofiledigest katello-ca-consumer-latest.noarch.rpm')
+        else:
+            self.execute('rpm -Uvh katello-ca-consumer-latest.noarch.rpm')
         # Not checking the status here, as rpm could be installed before
         # and installation may fail
-        result = self.execute(f'rpm -q katello-ca-consumer-{satellite.hostname}')
+        result = self.execute('rpm -q katello-ca-consumer*')
         # Checking the status here to verify katello-ca rpm is actually
         # present in the system
-        if result.status != 0:
+        if satellite.hostname not in result.stdout:
             raise ContentHostError('Failed to download and install the katello-ca rpm')
 
     def remove_katello_ca(self):


robottelo/hosts.py
```diff
@@ -389,7 +389,7 @@ def install_katello_ca(self, satellite):
             f'curl --insecure --output katello-ca-consumer-latest.noarch.rpm \
                     {satellite.url_katello_ca_rpm}'
         )
-        self.execute('rpm -Uvh katello-ca-consumer-latest.noarch.rpm')
+        self.execute('rpm -Uvh --nodigest --nofiledigest katello-ca-consumer-latest.noarch.rpm')
```
mshriver, line 392: Does the host have a property you can check, as it appears this workaround should only be applied when using FIPs.

@JacobCallahan thoughts on the impact if we stop having rpm validate the digest of the katello RPM? To me it seems this is going to far applied this broadly.

pondrejk: fair point, I'll look into it

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k test_positive_run_custom_job_template_by_ip[rhel8_fips]

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k test_positive_run_custom_job_template_by_ip[rhel8_fips]

mshriver (approved):
Would like to see passing test results in PRT for both fips and non-fips hosts.

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k 'test_positive_run_custom_job_template_by_ip[rhel8_fips] or test_positive_run_custom_job_template_by_ip[rhel8]'

robottelo/hosts.py
rplevka: I believe this has been addressed.

mshriver (approved):
The FIPS test failure in PRT is occuring after `install_katello_ca` completes, so it does validate this change.  @pondrejk I'm happy to merge this change as-is (its in draft though), or you can include further changes to completely resolve the test failure.

pondrejk:
The prt failure was due to an issue in the fips-checking condition that is now fixed. Result from VS testrun:
```
platform linux -- Python 3.10.4, pytest-7.1.2, pluggy-1.0.0
plugins: forked-1.4.0, xdist-2.5.0, services-2.2.1, reportportal-5.1.1, mock-3.7.0, ibutsu-2.0.2, cov-3.0.0
collected 1 item

tests/foreman/cli/test_remoteexecution.py 
.                                                          [100%]
================= 1 passed, 16 warnings in 1316.90s (0:21:56) ==================
```
@mshriver @rplevka

PR #9664, opened 2022-06-07
Restructuring rh_cloud test cases based on cert auth

adarshdubey-star:
`Removing token-based fixtures for rh_cloud_inventory/insights for test fixes`

### DIFF
--- a/pytest_fixtures/component/rh_cloud.py
+++ b/pytest_fixtures/component/rh_cloud.py
@@ -25,21 +25,6 @@ def rhcloud_manifest_org(rhcloud_sat_host):
     return org
 
 
-@pytest.fixture
-def set_rh_cloud_token(rhcloud_sat_host):
-    """A function-level fixture to set rh cloud token value."""
-    rhcloud_sat_host.update_setting('rh_cloud_token', settings.rh_cloud.token)
-    yield
-    rhcloud_sat_host.update_setting('rh_cloud_token', '')
-
-
-@pytest.fixture
-def unset_rh_cloud_token(rhcloud_sat_host):
-    """A function-level fixture to unset rh cloud token value."""
-    yield
-    rhcloud_sat_host.update_setting('rh_cloud_token', '')
-
-
 @pytest.fixture(scope='module')
 def organization_ak_setup(rhcloud_sat_host, rhcloud_manifest_org):
     """A module-level fixture to create an Activation key in module_org"""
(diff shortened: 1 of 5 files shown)


adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/api/test_rhcloud_inventory.py

adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/ui/test_rhcloud_insights.py

adarshdubey-star:
2 test cases are failing with different errors not related to the changes done above rest are good!

jyejare (commented):
ACK. Looks good, but I wonder how cert auth is being feeded for the RHCloud integration ?especially since you are removing the token based integration. Also the docstrings/test_steps of test cases are not updated to inform how cert auth based integration is being used here directly or indirectly.

Isnt it at all required.

adarshdubey-star:
> ACK. Looks good, but I wonder how cert auth is being feeded for the RHCloud integration ?especially since you are removing the token based integration. Also the docstrings/test_steps of test cases are not updated to inform how cert auth based integration is being used here directly or indirectly.
> 
> Isnt it at all required.

The certificate-based authentication is directly related to the presence of certificates at the organizational level now.
The token-based architecture is completely eradicated, following that implementation I have removed the statement within the test cases which intends to have the `satellite api key `termed as a `token`. There is no explicit parameter to toggle.

`Adding upstream PR for more reference ` - https://github.com/theforeman/foreman_rh_cloud/commit/5838733800f9fa56adfeb47e00cf87932d165675

adarshdubey-star:
> @adarshdubey-star Agree that you removed the `token` mention from docstrings but you are not updating the docstrings thats saying `presence of certificates at the organizational level`. That docstring mention is very important for newbies / external nonQE teams to understand how the test steps are using Authentication mechanism.

@jyejare I thought to keep it the same way, as there is no step going on in the test cases that can or will justify the presence of the authentication, as it's happening implicitly. Also about the understanding part, the same way it's going to remove from the documentation itself and that might create confusion with it.

jyejare (approved):

jameerpathan111:
@adarshdubey-star I forgot about [rh_cloud.yaml.template](https://github.com/SatelliteQE/robottelo/blob/master/conf/rh_cloud.yaml.template) can you update that as well and also the one present in our ci repo.

PR #9564, opened 2022-05-10
Fix puppet cli

vsedmik:
Rebased on #9560, this PR only fixes 2 tests failing since snap 18 (`test_positive_list` and `test_positive_list_with_non_admin_user`, see just 4d14fa7) rather due to changes in cli factory.

Test results:
```
(venv39) [vsedmik@localhost robottelo]$ pytest tests/foreman/cli/test_classparameters.py
============================================ test session starts ============================================
platform linux -- Python 3.9.9, pytest-7.1.2, pluggy-0.13.1
shared_function enabled - OFF - scope:  - storage: file
rootdir: /home/vsedmik/PycharmProjects/robottelo, configfile: pyproject.toml
plugins: forked-1.3.0, services-2.2.1, mock-3.7.0, lazy-fixture-0.6.3, ibutsu-2.0.2, reportportal-5.1.1, xdis
collected 10 items

tests/foreman/cli/test_classparameters.py ..........                                                  [100%]

================================ 10 passed, 35 warnings in 294.99s (0:04:54) ================================
```

PR #9673, opened 2022-06-09
cli_factory pickling

vsedmik:
**Problem statement:**
It seems that raising `CLIFactoryError` added in #9564 causes issues when broker checks in the Satellite and tries to pickle it:
```
  File "/usr/lib64/python3.8/multiprocessing/queues.py", line 239, in _feed
    obj = _ForkingPickler.dumps(obj)
  File "/usr/lib64/python3.8/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
  File "/opt/app-root/lib64/python3.8/site-packages/broker/hosts.py", line 43, in __getstate__
    self._purify()
  File "/opt/app-root/lib64/python3.8/site-packages/broker/hosts.py", line 50, in _purify
    pickle.dumps(obj)
  File "/opt/app-root/src/robottelo/robottelo/host_helpers/cli_factory.py", line 275, in __getattr__
    raise CLIFactoryError(f'unknown factory method name: {name.replace("make_", "")}')
robottelo.host_helpers.cli_factory.CLIFactoryError: unknown factory method name: __getstate__
```

According to https://docs.python.org/3.8/reference/datamodel.html#customizing-attribute-access the `__getattr__` method
should either return the (computed) attribute value or raise an `AttributeError` exception, which should be handled similar to when the attribute was not found. So I guess this way we say there is no `__getstate__` attribute in the class and the pickler should go on.

I'm unable to reproduce the issue locally for some reason, but the PRT seems to pass.

### DIFF
--- a/robottelo/host_helpers/cli_factory.py
+++ b/robottelo/host_helpers/cli_factory.py
@@ -272,7 +272,7 @@ class CLIFactory:
             fields = self._evaluate_functions(fields)
             return partial(create_object, entity_cls, fields)
         else:
-            raise CLIFactoryError(f'unknown factory method name: {name.replace("make_", "")}')
+            raise AttributeError(f'unknown factory method name: {name}')
 
     def _evaluate_function(self, function):
         """Some functions may require an instance reference"""


vsedmik:
trigger: test-robottelo
pytest: tests/foreman/ui/test_puppetclass.py

omkarkhatavkar (approved):

rplevka (approved):

PR #9674, opened 2022-06-09
Add test upgrade without init template

swadeley:
Closed loop: BZ#2063190

Hi, not sure where to put this code, it really needs to run just before the upgrade.
Will need some testing still to make sure other tests do not change the default init template setting.

[Bug 2063190](https://bugzilla.redhat.com/show_bug.cgi?id=2063190) - Upgrading from Satellite 6.9 to Satellite 6.10.3 fails with error "undefined method operatingsystems' for nil:NilClass" during the db:migrate step

### DIFF
--- a/.github/workflows/merge_to_master.yml
+++ b/.github/workflows/merge_to_master.yml
@@ -8,6 +8,7 @@ on:
 
 env:
     PYCURL_SSL_LIBRARY: openssl
+    ROBOTTELO_BUGZILLA__API_KEY: ${{ secrets.BUGZILLA_KEY }}
 
 jobs:
   codechecks:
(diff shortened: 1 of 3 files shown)


jyejare:
@swadeley The upgrade scenarios is the perfect place to add something before upgrade. Also its not always required to write the post definition in that case !

swadeley:
@jyejare  "The upgrade scenarios"? Many files or modules mention this, did you mean I should move the test?

I can delete the `test_post_senario_no_default_inti_template` as it does seem redundant, the test would never run if the upgrade failed.

swadeley:
Rebased.

swadeley:
Thank you.

PR #9682, opened 2022-06-13
Add test coverage for BZ#1204667

shubhamsg199:
Added test coverage for https://bugzilla.redhat.com/show_bug.cgi?id=1204667


shubhamsg199:
trigger: test-robottelo
pytest: -k 'test_positive_user_create_with_latin_email' tests/foreman/cli/test_user.py

tests/foreman/cli/test_user.py
```diff
@@ -133,6 +133,23 @@ def test_positive_CRUD(self):
...
+        :expectedresults: User should be created with the latin email id
+
+        :CaseImportance: High
+
+        :BZ: 1204667
+        """
+        email = f'{gen_string("latin1")}@example.com'
+        user = make_user({'mail': email})
+        result = User.list({'search': f'login = {user["login"]}'})
+        assert result[0]["email"] == email
```
Gauravtalreja1, line 151:
```suggestion
        assert User.list({'search': f'login = {user["login"]}'})[0]["email"] == email
```

Gauravtalreja1 (approved):
ACK, with one suggestion

tests/foreman/cli/test_user.py
```diff
@@ -133,6 +133,23 @@ def test_positive_CRUD(self):
...
+
+        :id: 262a12c2-fd4d-481c-8485-fe70a34ca7b7
+
+        :expectedresults: User should be created with the latin email id
+
+        :CaseImportance: High
+
+        :BZ: 1204667
+        """
+        email = f'{gen_string("latin1")}@example.com'
```
mshriver, line 148: There is something to consider with the behavior of `test_positive_CRUD` function above.

It's randomly picking from the datafactory, for the email. This is just a list of emails, defined here:
https://github.com/SatelliteQE/robottelo/blob/e9b0b4df781ce6a8ddb2051a221365cd6fc9eb3e/robottelo/datafactory.py#L295

If latin emails are supported, this use case can and should be covered by adding such an email to the list in the datafactory.

Picking a random email address within the test function is problematic, if you ask me, because you get random test coverage - and this BZ could easily could have gone unnoticed even with this coverage, with latin never getting picked in any of the snap testruns.

I would recommend you modify the datafactory function, and include a latin email.  Double check the other testcases where this list is used, and verify they pass as well.

Then, considering the increase in test execution time, actually parametrize `test_positive_CRUD` with the list of email types.  There's an enhancement that comes with this, and that's not actually randomizing the email itself as the value, but parametrizing the type of email - changing the datafactory list to a dictionary.

I don't want to blow the scope up on your PR here, but there's no need for this BZ to have unique test function coverage within the framework.

mshriver (changes requested):
Check out existing test coverage, and include latin type emails in the datafactory. 

Thanks for working on improving test coverage!

tests/foreman/cli/test_user.py
shubhamsg199: Thanks for the suggestion Mike.
I have parametrized the test  with email types and kept the datafactory list as the list itself
`changing the datafactory list to a dictionary` might  impact other tests which are treating it as a list across robottelo

mshriver (approved):

mshriver (changes requested):
Approved too early, metadata updates needed.

Test updates look good, thanks for taking the advice on updating existing coverage.  Improving datafactory by making emails a dict and updating impacted testcases would be a great enhancement across the framework.

shubhamsg199:
Updated.
I'll open up a separate PR for the enhancement of making emails a dict and also update impacted test cases with it

pondrejk (approved):

PR #9693, opened 2022-06-16
Helpers refactor 1

shweta83:
Replace UI  `create_fake_host` to CLI  `make_fake_host`


tests/foreman/ui/test_hardwaremodel.py
```diff
@@ -55,9 +59,24 @@ def test_positive_end_to_end(session, module_org, module_location):
         assert hm_values['vendor_class'] == vendor_class
         assert hm_values['info'] == info
         # Create host with associated hardware model
-        host_name = create_fake_host(
-            session, host_template, extra_values={'additional_information.hardware_model': name}
+        host = make_fake_host(
```
mshriver, line 62: This does feel like a test step for the `end to end` context of this test, and as such something that should be done via the UI.

mshriver: I could be convinced otherwise, since at least the update to the host to set the hardware model is done through the UI. But even then, this is changing the test behavior to update the host with hardware model - instead of specifying it on host creation.  

I'd take one more step back with this test function, and look at what its doing in general.

The `host_template` creation should get moved into a fixture, its not part of the test coverage and is done via API anyway.

JacobCallahan: I think the question is whether or not the fake host creation is considered part of the coverage. Something the component owner needs to decide

shweta83: @mshriver @JacobCallahan Our objective to update the hardware model with host is achieved here. Do we need the host to be created through UI only?

tests/foreman/ui/test_host.py
```diff
@@ -2331,18 +2360,32 @@ def test_positive_create_with_puppet_class(
...
         session.organization.select(org_name=module_puppet_org.name)
         session.location.select(loc_name='Any Location')
-        host_name = create_fake_host(
-            session,
-            host_template,
-            extra_values={
-                'host.puppet_environment': module_env_search.name,
-                'puppet_enc.classes.assigned': [module_import_puppet_module['puppet_class']],
-            },
+        host = make_fake_host(
```
mshriver, line 2367: This needs to continue using the UI for the host creation - its a primary aspect of the test coverage here.

tests/foreman/ui/test_host.py
```diff
@@ -626,8 +632,31 @@ def test_negative_delete_primary_interface(session, module_host_template):
     :CaseLevel: System
     """
     interface_id = gen_string('alpha')
+    host_template = entities.Host(organization=module_org, location=module_location)
+    host_template.create_missing()
```
mshriver, line 636: This host entity that gets created just to get a domain ID and MAC address is never cleaned up.

tests/foreman/ui/test_host.py
```diff
@@ -626,8 +632,31 @@ def test_negative_delete_primary_interface(session, module_host_template):
...
+    interface = (
+        'type=interface,mac={},identifier=eth0,name={},domain_id={},'
+        'ip={},primary=true,provision=true'
+    ).format(host_template.mac, gen_string('alpha'), host_template.domain.id, gen_ipaddr())
+    os_name = f'{host_template.operatingsystem.name} {host_template.operatingsystem.major}'
     with session:
-        host_name = create_fake_host(session, module_host_template, interface_id=interface_id)
+        host = make_fake_host(
```
mshriver, line 643: Given the test coverage is concerned with the delete action, the host create does not have to happen through the UI so this is okay.

It should happen in a fixture though, not in the test function itself - because its test setup, and isn't being done via the UI.

tests/foreman/ui/test_host.py
```diff
@@ -626,8 +632,31 @@ def test_negative_delete_primary_interface(session, module_host_template):
     :CaseLevel: System
     """
     interface_id = gen_string('alpha')
+    host_template = entities.Host(organization=module_org, location=module_location)
+    host_template.create_missing()
+    interface = (
+        'type=interface,mac={},identifier=eth0,name={},domain_id={},'
+        'ip={},primary=true,provision=true'
+    ).format(host_template.mac, gen_string('alpha'), host_template.domain.id, gen_ipaddr())
+    os_name = f'{host_template.operatingsystem.name} {host_template.operatingsystem.major}'
```
mshriver, line 641: I don't think pushing all of this into the test scope is a good pattern for the helpers refactor.  We're trading one helper for another, and needlessly creating entities on the satellite.

There is a `module_host_template` fixture in use by this test, which with the UI helper was used to populate the operating system fields.

@jyejare The direction we should be going with test behaviors like this is to expand our fixture definitions to cover these use cases.  The CLI factory function isn't really necessary to use here. We can indirectly parametrize the existing `pytest_fixtures.component.host.function_host` in order to specify additional arguments, if necessary.

Nailgun's `Host.create_missing()` function handles many/all of the necessary fields for creating a fake managed host, and the test setup in robottelo should leverage that. If `interface` isn't being handled right now, the `nailgun.entities.Host` class should be updated.

There is a third module using the `ui.create_fake_host` method, which does not need to create the host via UI for test coverage and would also benefit from a globally defined fixture.

tests/foreman/ui/test_hardwaremodel.py
```diff
@@ -8,7 +8,7 @@
 
 :CaseComponent: Hosts
 
-:Assignee: pdragun
+:Assignee: tstrych
```
peterdragun, line 11: Please don't change the assignee of these tests

tests/foreman/ui/test_host.py
```diff
@@ -8,7 +8,7 @@
 
 :CaseComponent: Hosts
 
-:Assignee: pdragun
+:Assignee: tstrych
```
peterdragun, line 11: Please don't change the assignee of these tests

peterdragun (changes requested):

tests/foreman/ui/test_hardwaremodel.py
jyejare: Interesting - Why there are no conflicts :)

JacobCallahan:
@shweta83 can you revisit the feedback and make updates as appropriate? I'd like to get this in very soon. Thanks!

tests/foreman/ui/test_hardwaremodel.py
peterdragun: Probably wrong rebase on master :)

tests/foreman/ui/test_host.py
```diff
@@ -1296,7 +1330,7 @@ def test_positive_global_registration_end_to_end(
     result = rhel_contenthost.execute('subscription-manager identity')
     assert result.status == 0
     # Assert that a yum update was made this day ("Update" or "I, U" in history)
-    timezone_offset = rhel_contenthost.execute('date +"%:z"').stdout.strip()
+    timezone_offset = rhel_contenthost.execute('date +"%:z"').stdout
```
peterdragun, line 1333: This should also stay without change

tests/foreman/ui/test_host.py
```diff
@@ -2145,97 +2179,6 @@ def enable_new_host_details_ui(target_sat, setting_update):
...
-        2. Initialize synchronization
-        3. Module-Stream Get
-        4. Update the Module-Stream
-        5. Delete the Module-Stream
-
-    :expectedresults: Yum repository with modules is synced, updated, deleted.
-
-    """
-
-
```
peterdragun, line 2238: Those tests should also stay

tests/foreman/ui/test_host.py
```diff
@@ -1570,7 +1604,7 @@ def test_global_registration_with_gpg_repo_and_default_package(
 @pytest.mark.tier2
 @pytest.mark.rhel_ver_match('[^6].*')
 def test_global_registration_upgrade_subscription_manager(
-    session, module_activation_key, module_os, rhel_contenthost
+    session, module_activation_key, module_os, module_proxy, rhel_contenthost
```
peterdragun, line 1607: Any reason why adding this parameter? I believe it was deleted as it was not used

tests/foreman/ui/test_host.py
```diff
@@ -237,17 +262,25 @@ def module_libvirt_hostgroup(
 
 
 @pytest.fixture(scope='module')
-def module_activation_key(module_org_with_manifest, module_target_sat):
+def manifest_org(module_org):
+    """Upload manifest to organization."""
+    with manifests.clone() as manifest:
+        upload_manifest_locked(module_org.id, manifest.content)
+    return module_org
```
peterdragun, line 269: This was moved a while back to taxonomies and renamed to `module_org_with_manifest`

tests/foreman/ui/test_host.py
```diff
@@ -348,33 +384,25 @@ def test_positive_read_from_details_page(session, module_host_template):
...
     """
-    os_name = '{} {}'.format(
-        module_host_template.operatingsystem.name, module_host_template.operatingsystem.major
+
+    host = module_host_template.create()
+    os_name = (
+        f'{module_host_template.operatingsystem.name} {module_host_template.operatingsystem.major}'
     )
-    interface_id = gen_string('alpha')
+    interface = entities.Interface(host=host, primary=False).create()
```
peterdragun, line 392: any reason to use entities instead of target_sat.api?

tests/foreman/ui/test_hardwaremodel.py
shweta83: @peterdragun yes that was a wrong rebase. I updated it.

tests/foreman/ui/test_host.py
shweta83: updated.

tests/foreman/ui/test_hardwaremodel.py
```diff
@@ -15,10 +15,13 @@
 :Upstream: No
 """
 import pytest
+from fauxfactory import gen_alphanumeric
 from fauxfactory import gen_string
 from nailgun import entities
 
-from robottelo.ui.utils import create_fake_host
+from robottelo.cli.factory import make_fake_host
```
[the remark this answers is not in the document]
shweta83: > @shweta83 can you revisit the feedback and make updates as appropriate? I'd like to get this in very soon. Thanks!

@JacobCallahan Updated the PR

tests/foreman/ui/test_host.py
shweta83: updated

tests/foreman/ui/test_host.py
```diff
@@ -388,27 +406,29 @@ def test_positive_read_from_edit_page(session, module_host_template):
...
     :CaseLevel: System
     """
-    os_name = '{} {}'.format(
-        module_host_template.operatingsystem.name, module_host_template.operatingsystem.major
+    host = module_host_template.create()
+    os_name = (
+        f'{module_host_template.operatingsystem.name} {module_host_template.operatingsystem.major}'
     )
-    interface_id = gen_string('alpha')
+    interface = entities.Interface(host=host, primary=False).create()
```
peterdragun, line 413: please use target sat instead of entities

tests/foreman/ui/test_host.py
```diff
@@ -2151,18 +2175,14 @@ def test_positive_read_details_page_from_new_ui(
     session, module_host_template, enable_new_host_details_ui, setting_update
 ):
     """Create new Host and read all its content through details page
-
     :id: ef0c5942-9049-11ec-8029-98fa9b6ecd5a
-
     :expectedresults: Host is created and has expected content
-
```
peterdragun, line 2158: Those missing lines in docstring are probably causing checks to fail, same for other tests.

peterdragun (approved):

tests/foreman/ui/test_hardwaremodel.py
```diff
@@ -55,9 +54,10 @@ def test_positive_end_to_end(session, module_org, module_location):
         assert hm_values['vendor_class'] == vendor_class
         assert hm_values['info'] == info
         # Create host with associated hardware model
-        host_name = create_fake_host(
-            session, host_template, extra_values={'additional_information.hardware_model': name}
-        )
+        host = session.host.create(values)
+        values.update({'additional_information.hardware_model': name})
+        host_name = f'{name}.{host.domain.name}'
+        session.host.update(host_name, {'additional_information.hardware_model': name})
```
jyejare, line 60:
```suggestion
        values.update({'additional_information.hardware_model': name})
        host_name = f'{name}.{host.domain.name}'
        host = session.host.create(values)
```
??

jyejare: TBH, I really didnt like the approach of removing the create_fake_host method and merge its code in tests directly! I believe we made it complex whereas earlier it was easy to identify that we are creating a fake host by just calling that function !!

Especially we are defining the `populate_ui_host_fields` in one test module and calling in another is an insane change! 

I would have moved create_fake_host in satellite class mixins ! so that I could have access it as `sat.create_fake_host()` or in api_utils to access as `sat.api_utils.create_fake_host()`. This would have been easier outside automated tests as to quick deploy the host from python interactive shell !

tests/foreman/ui/test_host.py
```diff
@@ -47,7 +47,30 @@
 from robottelo.constants import PERMISSIONS
 from robottelo.constants import REPO_TYPE
 from robottelo.datafactory import gen_string
-from robottelo.ui.utils import create_fake_host
+
+
+def populate_ui_host_fields(host):
```
jyejare, line 52: create_fake_host was much better and generic as per my last comment !

jyejare (changes requested):
TBH I am not really sold on desolving create_fake_host in tests ! We made it difficult now , look for inline ocmments!

Also, make sure you delete the ui/utils in this PR so no one would want to use it !

jyejare (approved):

jyejare:
Merging this PR as the recent comments have been addressed and hence previous comments from @JacobCallahan and @mshriver are no longer valid!

Issue #9690, opened 2022-06-15
target_sat.api.Media behaves differently then entities.Media


pondrejk:
Wrestling with automation failure for test_positive_rename_satellite in sys/test_rename.py:
```
media = target_sat.api.Media(organization=[module_org], path_=url).create()
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/pondrejk/.local/lib/python3.10/site-packages/nailgun/entities.py", line 4927, in create
    id=self.create_json(create_missing)['id'],
  File "/home/pondrejk/.local/lib/python3.10/site-packages/nailgun/entity_mixins.py", line 920, in create_json
    response.raise_for_status()
  File "/usr/lib/python3.10/site-packages/requests/models.py", line 960, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 422 Client Error: Unprocessable Entity for url: https://target-sat-hostname.com/api/v2/media
```

In a debug session, trying to create media with the same line fails, but trying it out with entities succeeds:

```
from nailgun import entities
media = entities.Media(organization=[module_org], path_=url).create()
-> nailgun.entities.Media(path_='https://SFgSnDLXEB.gov', name='UlgbmQArXkHE', operatingsystem=[], organization=[nailgun.entities.Organization(id=8)], location=[nailgun.entities.Location(id=2)], os_family=None, id=16)
```

pondrejk:
more info:
the error is due to the fact that module_org is not being created on a target_sat, test def goes like:

```
@pytest.mark.destructive
def test_positive_rename_satellite(self, module_org, module_product, target_sat):
```
Action item -- make sure the module_org fixture uses target_sat object

PR #9697, opened 2022-06-16
refactor: moving destructive tests to destructive folder

omkarkhatavkar:
### PR Description:

This pr changes the pattern of how destructive tests are collected and designed such it would leverage the session-scoped(session/module/function) satellite. There was a bug that caused destructive tests not using the module-level satellite. Now with this pr, we would be able to run the destructive tests w.r.t. component name in pytest without disturbing the non-destructive tests. 

Below are the modified changes with this PR

- [x] The destructive tests are migrated from individual component test modules to the destructive folder `/tests/foreman/destructive`  
- [x] Now module_target_sat fixture can be used across all destructive tests and that do not conflict with non-destructive tests 
- [x] UI session is not moved towards the core pytest fixture so it can be used across the repository and not w.r.t. UI folder earlier 
- [x] Segregated and refactored many fixtures and tests to use the right use of fixture satellite (module/function)
- [x] Modified the cli command calls migrated to use satellite.cli.command, satellite.cli.factory.command pattern

PR #9699, opened 2022-06-16
fix in rex alternate dir test

pondrejk:
Removed module_org fixture that was not used in the test and also caused failure (probably due to the combination of target_sat and destructive mark described in #9690)

test result:
```============================= test session starts ==============================
platform linux -- Python 3.10.4, pytest-7.1.2, pluggy-1.0.0
plugins: forked-1.4.0, xdist-2.5.0, services-2.2.1, reportportal-5.1.1, mock-3.7.0, ibutsu-2.0.2, cov-3.0.0
collected 1 item

tests/foreman/cli/test_remoteexecution.py .                              [100%]

================== 1 passed, 16 warnings in 634.66s (0:10:34) =================
```

### DIFF
--- a/tests/foreman/cli/test_remoteexecution.py
+++ b/tests/foreman/cli/test_remoteexecution.py
@@ -255,7 +255,7 @@ class TestRemoteExecution:
 
     @pytest.mark.destructive
     @pytest.mark.rhel_ver_list([7])
-    def test_positive_use_alternate_directory(self, rex_contenthost, module_org, target_sat):
+    def test_positive_use_alternate_directory(self, rex_contenthost, target_sat):
         """Use alternate working directory on client to execute rex jobs
 
         :id: a0181f18-d3dc-4bd9-a2a6-430c2a49809e


akhil-jha:
trigger: test-robottelo
pytest: -k test_positive_use_alternate_directory tests/foreman/cli/test_remoteexecution.py

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -255,7 +255,7 @@ def test_positive_run_custom_job_template_by_ip(self, rex_contenthost, module_or
 
     @pytest.mark.destructive
     @pytest.mark.rhel_ver_list([7])
```
akhil-jha, line 257: You can do `@pytest.mark.rhel_ver_list([7,8,9])`  as well right ?? Is there any reason why it's not done that way?

pondrejk: I do it where it makes sense but I didn't want to quadruple the whole module for little gain. Here we test mostly a server-side procedure that is unlikely to be influenced by the target host OS.

pondrejk:
hm, passes locally but rpr failed in host setup, trying again

pondrejk:
trigger: test-robottelo
pytest: -k test_positive_use_alternate_directory tests/foreman/cli/test_remoteexecution.py

pondrejk:
trigger: test-robottelo
pytest: -k test_positive_use_alternate_directory tests/foreman/cli/test_remoteexecution.py

akhil-jha:
In all the PRT fails, it's failing because `Received HTTP 404 response: {"message":"Organization with id <id> not found"}`
Not getting this locally?
Running PRT again.

akhil-jha:
trigger: test-robottelo
pytest: -k test_positive_use_alternate_directory tests/foreman/cli/test_remoteexecution.py

akhil-jha:
I was able to reproduce it locally. Just rebased it to the master. 

I think the issue is - When the test runs and does a new sat checkout. But when [this](https://github.com/SatelliteQE/robottelo/blob/53fcf5101e44940237299fc6cb8b6c2bf21c3800/pytest_fixtures/core/contenthosts.py#L139) runs, module_org still points to the org created on the previous sat(from server.yaml) whereas target_sat points to the new sat which doesn't have that org created. Basically a mismatch between module_org and satellite. Thus resulting in 
 `{"message":"Organization with id <id> not found"}`.

mshriver (changes requested):
This test is destructive, and is covered by #9697. I think you can close this, as the behavior will be addressed by moving the test as is done in that PR. You can review the test in its new location there.

PR #9702, opened 2022-06-16
Updating test_discoveredhost to DiscoveryImage

Griffin-Sullivan:
Splitting discovery plugin and discovery image into separate components.


mshriver (approved):

latran (approved):
acking.  A few questions: what's the difference between the two? Where does `DiscoveryPlugin` reside now?

akhil-jha (approved):
Ack. I am curious on Lai's question too! I can merge after that

PR #9681, opened 2022-06-13
Install capsule using CDN repos for RHSCL and AE

vsedmik:
In snap 24 the capsule installer fails at RHEL7 with a dependency error because SatLab template provides postgres in version 12.11, while packages available in DF are 12.9.

```
    #x1B[34m2022-06-10 04:59:20#x1B[0m [#x1B[31mERROR #x1B[0m] [#x1B[36mroot#x1B[0m] Failed to ensure rh-postgresql12-postgresql-server, rh-redis5-redis, pulpcore-selinux are installed
    #x1B[34m2022-06-10 04:59:20#x1B[0m [#x1B[31mERROR #x1B[0m] [#x1B[36mroot#x1B[0m] #x1B[1;31mError: Execution of '/bin/yum -d 0 -e 0 -y install rh-postgresql12-postgresql-server' returned 1: Error: Package: rh-postgresql12-postgresql-server-12.9-1.el7.x86_64 (rhel-server-rhscl-7-rpms)
               Requires: rh-postgresql12-postgresql-libs(x86-64) = 12.9-1.el7
               Installed: rh-postgresql12-postgresql-libs-12.11-1.el7.x86_64 (@rhel-server-rhscl-7-rpms)
                   rh-postgresql12-postgresql-libs(x86-64) = 12.11-1.el7
```

So in this PR I register to CDN instead of DF to get latest content from `rhel`, `rhscl` and `ansible_engine` repos, while Satellite/Capsule repos remain the ones defined in `/etc/yum.repos.d./satellite.repo`

PR #9709, opened 2022-06-17
Fix katello-agent tests to use client repo and register to cdn

Gauravtalreja1:
**Description:**
1. Using DF's client repo for katello-agent package install on the client instead of old tools repo from CDN.
2. register_to_cdn instead of register_to_dogfood.
3. Parametrizing tests to run on rhel6/7/8/9 and rhel7/8-fips content hosts.
4. Moving tests to new tier5 for deprecated components.

**Dependant MR**: satlab-tower-MR#562 (merged)

### DIFF
--- a/pytest_fixtures/component/subscription.py
+++ b/pytest_fixtures/component/subscription.py
@@ -4,18 +4,6 @@ from robottelo.constants import DEFAULT_SUBSCRIPTION_NAME
 
 
 @pytest.fixture(scope='module')
-def module_clean_rhsm(module_target_sat):
-    """removes pre-existing candlepin certs and resets RHSM."""
-    module_target_sat.remove_katello_ca()
-
-
-@pytest.fixture
-def func_clean_rhsm(target_sat):
-    """removes pre-existing candlepin certs and resets RHSM."""
-    target_sat.remove_katello_ca()
-
-
-@pytest.fixture(scope='module')
 def default_subscription(module_target_sat, module_org_with_manifest):
     subscription = module_target_sat.api.Subscription(
         organization=module_org_with_manifest.id
@@ -24,39 +12,25 @@ def default_subscription(module_target_sat, module_org_with_manifest):
     return subscription[0]
 
 
-def subscribe_satellite(sat):
-    from robottelo.config import settings
-
-    if sat.os_version.major < 8:
-        release_version = f'{sat.os_version.major}Server'
-    else:
-        release_version = f'{sat.os_version.major}'
-    sat.register_contenthost(
-        org=None,
-        lce=None,
-        username=settings.subscription.rhn_username,
-        password=settings.subscription.rhn_password,
-        releasever=release_version,
-    )
-    result = sat.subscription_manager_attach_pool([settings.subscription.rhn_poolid])[0]
-    if 'Successfully attached a subscription' in result.stdout:
-        # extras is not in RHEL8: https://access.redhat.com/solutions/5331391
-        if sat.os_version.major < 8:
-            sat.enable_repo(f'rhel-{sat.os_version.major}-server-extras-rpms', force=True)
-        yield
-    else:
-        pytest.fail('Failed to attach system to pool. Aborting Test!.')
-    sat.unregister()
-    sat.remove_katello_ca()
-
-
 @pytest.fixture(scope='module')
-def module_subscribe_satellite(module_clean_rhsm, module_target_sat):
-    """subscribe satellite to cdn"""
-    subscribe_satellite(module_target_sat)
+def module_subscribe_satellite(module_target_sat):
+    """Subscribe Satellite to CDN"""
+    module_target_sat.register_to_cdn()
+    # Enable extras repo if os_version is RHEL7
+    if module_target_sat.os_version.major < 8:
+        module_target_sat.enable_repo(
+            f'rhel-{module_target_sat.os_version.major}-server-extras-rpms', force=True
+        )
+    yield
+    module_target_sat.unregister()
 
 
 @pytest.fixture
-def func_subscribe_satellite(func_clean_rhsm, target_sat):
-    """subscribe satellite to cdn"""
-    subscribe_satellite(target_sat)
+def subscribe_satellite(target_sat):
+    """Subscribe Satellite to CDN"""
+    target_sat.register_to_cdn()
+    # Enable extras repo if os_version is RHEL7
+    if target_sat.os_version.major < 8:
+        target_sat.enable_repo(f'rhel-{target_sat.os_version.major}-server-extras-rpms', force=True)
+    yield
+    target_sat.unregister()
--- a/pytest_plugins/markers.py
+++ b/pytest_plugins/markers.py
@@ -10,6 +10,7 @@ def pytest_configure(config):
         "tier2: Tier 2 tests",  # Association tests
         "tier3: Tier 3 tests",  # Systems integration tests
         "tier4: Tier 4 tests",  # Long running tests
+        "tier5: Tier 5 tests",  # Deprecated component tests
         "destructive: Destructive tests",
         "upgrade: Upgrade tests",
         "pit_server: PIT server scenario tests",
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -384,7 +384,7 @@ class ContentHost(Host, ContentHostMixins):
         if result.status != 0:
             raise ContentHostError(f'Failed to install katello-agent: {result.stdout}')
         try:
-            wait_for(lambda: self.execute('systemctl status goferd').status == 0)
+            wait_for(lambda: self.execute('service goferd status').status == 0)
         except TimedOutError:
             raise ContentHostError('katello-agent is not running')
 
@@ -1118,6 +1118,24 @@ class ContentHost(Host, ContentHostMixins):
             raise ContentHostError('There was an error installing katello-host-tools-tracer')
         self.execute('katello-tracer-upload')
 
+    def register_to_cdn(self):
+        """Subscribe satellite to CDN"""
+        self.remove_katello_ca()
+        major_version = self.os_version.major
+        release_version = f'{major_version}Server' if major_version < 8 else f'{major_version}'
+        cmd_result = self.register_contenthost(
+            org=None,
+            lce=None,
+            username=settings.subscription.rhn_username,
+            password=settings.subscription.rhn_password,
+            releasever=release_version,
+        )
+        if cmd_result.status != 0:
+            raise ContentHostError(
+                f'Error during registration, command output: {cmd_result.stdout}'
+            )
+        self.subscription_manager_attach_pool([settings.subscription.rhn_poolid])[0]
+
 
 class Capsule(ContentHost, CapsuleMixins):
     rex_key_path = '~foreman-proxy/.ssh/id_rsa_foreman_proxy.pub'
@@ -1524,3 +1542,31 @@ class Satellite(Capsule, SatelliteMixins):
                 f'Error during cockpit installation, installation output: {cmd_result.stdout}'
             )
         self.add_rex_key(self)
+
+    def wait_for_tasks(
+        self, search_query, search_rate=1, max_tries=10, poll_rate=None, poll_timeout=None
+    ):
+        """Search for tasks by specified search query and poll them to ensure that
+        task has finished.
+
+        :param search_query: Search query that will be passed to API call.
+        :param search_rate: Delay between searches.
+        :param max_tries: How many times search should be executed.
+        :param poll_rate: Delay between the end of one task check-up and
+            the start of the next check-up. Parameter for ``sat.api.ForemanTask.poll()`` method.
+        :param poll_timeout: Maximum number of seconds to wait until timing out.
+            Parameter for ``sat.api.ForemanTask.poll()`` method.
+        :return: List of ``sat.api.ForemanTasks`` entities.
+        :raises: ``AssertionError``. If not tasks were found until timeout.
+        """
+        for _ in range(max_tries):
+            tasks = self.api.ForemanTask().search(query={'search': search_query})
+            if len(tasks) > 0:
+                for task in tasks:
+                    task.poll(poll_rate=poll_rate, timeout=poll_timeout)
+                break
+            else:
+                time.sleep(search_rate)
+        else:
+            raise AssertionError(f"No task was found using query '{search_query}'")
+        return tasks
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -1134,7 +1134,11 @@ class ContentHost(Host, ContentHostMixins):
             raise ContentHostError(
                 f'Error during registration, command output: {cmd_result.stdout}'
             )
-        self.subscription_manager_attach_pool([settings.subscription.rhn_poolid])[0]
+        cmd_result = self.subscription_manager_attach_pool([settings.subscription.rhn_poolid])[0]
+        if cmd_result.status != 0:
+            raise ContentHostError(
+                f'Error during pool attachment, command output: {cmd_result.stdout}'
+            )
 
 
 class Capsule(ContentHost, CapsuleMixins):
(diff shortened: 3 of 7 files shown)


Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_katello_agent.py

robottelo/hosts.py
```diff
@@ -1187,6 +1187,24 @@ def register_to_dogfood(self, ak_type='satellite'):
                 f'Error during registration, command output: {cmd_result.stdout}'
             )
 
+    def register_to_cdn(self):
+        """Subscribe satellite to CDN"""
+        self.remove_katello_ca()
+        major_version = self.os_version.major
+        release_version = f'{major_version}Server' if major_version < 8 else f'{major_version}'
```
akhil-jha, line 1194: Shouldn't subscription-manager be able to set it itself while registering? Why be explicit ?

Gauravtalreja1: As I referred to the existing subscribe_satellite fixture for this helper func, I'm not sure why it is written to include this explicitly and whether it should be kept or not

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_katello_agent.py

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_katello_agent.py

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_katello_agent.py

robottelo/hosts.py
```diff
@@ -376,7 +376,7 @@ def install_katello_agent(self):
         if result.status != 0:
             raise ContentHostError(f'Failed to install katello-agent: {result.stdout}')
         try:
-            wait_for(lambda: self.execute('systemctl status goferd').status == 0)
+            wait_for(lambda: self.execute('service goferd status').status == 0)
```
jyejare, line 379: Why moving to older service module from systemctl ?

robottelo/hosts.py
```diff
@@ -1102,6 +1102,24 @@ def install_tracer(self):
             raise ContentHostError('There was an error installing katello-host-tools-tracer')
         self.execute('katello-tracer-upload')
 
+    def register_to_cdn(self):
+        """Subscribe satellite to CDN"""
+        self.remove_katello_ca()
+        major_version = self.os_version.major
+        release_version = f'{major_version}Server' if major_version < 8 else f'{major_version}'
```
jyejare, line 1109: This all can go into `register_contenthost` method which is more related to registeration.

robottelo/hosts.py
```diff
@@ -1102,6 +1102,24 @@ def install_tracer(self):
...
+        cmd_result = self.register_contenthost(
+            org=None,
+            lce=None,
+            username=settings.subscription.rhn_username,
+            password=settings.subscription.rhn_password,
+            releasever=release_version,
+        )
+        if cmd_result.status != 0:
+            raise ContentHostError(
+                f'Error during registration, command output: {cmd_result.stdout}'
```
jyejare, line 1119: Even this should go into `register_contenthost`.

tests/foreman/cli/test_katello_agent.py
```diff
@@ -18,24 +18,23 @@
 """
 import pytest
 
+from pytest_fixtures.core.contenthosts import register_host_custom_repo
```
jyejare, line 21: **RedFlag!** Not recommended to import the fixtures. Please use them directly in fixture.

tests/foreman/cli/test_katello_agent.py
```diff
@@ -18,24 +18,23 @@
...
 from robottelo.config import settings
 from robottelo.helpers import InstallerCommand
 
-
 pytestmark = [
     pytest.mark.run_in_one_thread,
-    pytest.mark.skip_if_not_set('clients', 'fake_manifest'),
     pytest.mark.destructive,
+    pytest.mark.tier5,
+    pytest.mark.upgrade,
```
jyejare, line 30: This will end up selecting all upgrade tests in present and future. I would recommend adding the flag on selective only tests.

tests/foreman/cli/test_katello_agent.py
```diff
@@ -49,61 +48,19 @@ def sat_with_katello_agent(module_target_sat):
...
-    rhel7_contenthost.install_katello_ca(sat_with_katello_agent)
-    # Register content host and install katello-agent
-    rhel7_contenthost.register_contenthost(
-        katello_agent_repos['org'].label,
-        katello_agent_repos['ak'].name,
+def katello_agent_client(sat_with_katello_agent, rhel_contenthost):
+    """Register content host to Satellite and install katello-agent on the host."""
+    org = sat_with_katello_agent.api.Organization().create()
+    repo = settings.repos['SATCLIENT_REPO'][f'RHEL{rhel_contenthost.os_version.major}']
+    register_host_custom_repo(
```
jyejare, line 56: Add this in fixture parameters to access it. It's a fixture and not recommended to use it as an importable.

tests/foreman/cli/test_katello_agent.py
```diff
@@ -49,61 +48,19 @@ def sat_with_katello_agent(module_target_sat):
...
+        sat_with_katello_agent, org, rhel_contenthost, [repo, settings.repos.yum_1.url]
     )
-    assert rhel7_contenthost.subscribed
-    host_info = sat_with_katello_agent.cli.Host.info({'name': rhel7_contenthost.hostname})
-    rhel7_contenthost.enable_repo(constants.REPOS['rhst7']['id'])
-    rhel7_contenthost.install_katello_agent()
-    yield {'client': rhel7_contenthost, 'host_info': host_info, 'sat': sat_with_katello_agent}
+    rhel_contenthost.install_katello_agent()
+    host_info = sat_with_katello_agent.cli.Host.info({'name': rhel_contenthost.hostname})
```
jyejare, line 60: Since its a setup part, try using API way and not the CLI way! Could be in future :)

Non blocking comment for this PR.

jyejare (changes requested):

robottelo/hosts.py
Gauravtalreja1: For EL6 systemctl doesn't works, whereas the service command works for all

tests/foreman/cli/test_katello_agent.py
Gauravtalreja1: @jyejare This is a function that is written there in pytest_fixtures, and not a fixture, anyway I also think this should be moved to Satellite class in robottelo.hosts or maybe some helper for content hosts tests.
I think this is no RedFlag for this PR, and this can be handled separately in PR?

tests/foreman/cli/test_katello_agent.py
Gauravtalreja1: yes, I would like to include all these tests of the module for upgrade all-tier run

tests/foreman/cli/test_katello_agent.py
Gauravtalreja1: No, it's not a fixture https://github.com/SatelliteQE/robottelo/blob/master/pytest_fixtures/core/contenthosts.py#L144

pytest_fixtures/core/broker.py
```diff
@@ -31,6 +31,7 @@ def _target_sat_imp(request, _default_sat, satellite_factory):
     if request.node.get_closest_marker(name='destructive'):
         new_sat = satellite_factory()
         yield new_sat
+        new_sat.unregister()
```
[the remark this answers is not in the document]
Gauravtalreja1: Not affecting me, but If a SAT system is registered to CDN so I think it is better to unregister before check-in, check similar implementation for capsule https://github.com/SatelliteQE/robottelo/pull/9681/files#diff-9fedcc170c47962a575c235d189eb051bd6a9ba799f7ba27a5c6c378c4980305R136

tests/foreman/cli/test_katello_agent.py
jyejare: Not sure what could be the consequences/side-effects for importing something from fixture modules which are being initiated from the central conftest.py.

But if you want to move this to ContentHost class where u might get some similar methods written there where you can merge into this.

OR Simply turn that function into function level fixture :)

tests/foreman/cli/test_katello_agent.py
jyejare: See my above comment!

robottelo/hosts.py
Gauravtalreja1: I'm not sure if this should be added to register_contenthost, as register_contenthost simply returns `return self.execute(cmd)`, and including this check there is a good idea, I think this will make this PR more huge, so I prefer this to handle in separate PR

robottelo/hosts.py
Gauravtalreja1: I think not, this option is valid only for CDN registration and not for dogfood, and if we do that we will be hardcoding `releasever` param for each registration, which will be for contenthosts as well

tests/foreman/cli/test_katello_agent.py
Gauravtalreja1: I remember I used this because Host.info wasn't implemented in nailgun, will play with nailgun for this, but as this is non-blocking so will resolve this thread first

tests/foreman/cli/test_katello_agent.py
Gauravtalreja1: @jyejare Updated, moved this func to robotello hosts class

robottelo/hosts.py
```diff
@@ -1515,3 +1533,82 @@ def install_cockpit(self):
...
+                    task.poll(poll_rate=poll_rate, timeout=poll_timeout)
+                break
+            else:
+                time.sleep(search_rate)
+        else:
+            raise AssertionError(f"No task was found using query '{search_query}'")
+        return tasks
+
+    def register_host_custom_repo(self, module_org, rhel_contenthost, repo_urls):
+        """Register content host to Satellite and sync repos"""
```
jyejare, line 1566: You are missing the important docstring for this method ! Please add !

jyejare: ^ I mean for parameters !

jyejare:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py

robottelo/hosts.py
```diff
@@ -1515,3 +1533,82 @@ def install_cockpit(self):
                 f'Error during cockpit installation, installation output: {cmd_result.stdout}'
             )
         self.add_rex_key(self)
+
+    def wait_for_tasks(
```
shweta83, line 1537: @Gauravtalreja1 I think this helper method is more suited in previous location(robottelo.api.utils).  @JacobCallahan @jyejare any thoughts?

Gauravtalreja1: Okay, moved it to capsule_mixin under CapsuleInfo class

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_katello_agent.py

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_katello_agent.py

pytest_plugins/markers.py
```diff
@@ -10,6 +10,7 @@ def pytest_configure(config):
         "tier2: Tier 2 tests",  # Association tests
         "tier3: Tier 3 tests",  # Systems integration tests
         "tier4: Tier 4 tests",  # Long running tests
+        "tier5: Tier 5 tests",  # Deprecated component tests
```
JacobCallahan, line 13: I think I missed a discussion around this new tier designation. What is the reasoning behind this?

JacobCallahan (changes requested):
Some small things. Really excited to get this in!

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_katello_agent.py

pytest_fixtures/core/broker.py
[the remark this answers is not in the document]
Gauravtalreja1: Updated, please re-review

pytest_plugins/markers.py
Gauravtalreja1: yes, this new tier is for deprecated components, which will be low prio

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_katello_agent.py

swadeley (approved):
ACK. I saw a few times in text "contenthost"  when it is "content host" in other places. A trivial detail.

ogajduse (changes requested):
I recommend using a different way to mark the tests in order for the `rhel_contenthost` fixture to pick up the new RHEL release once we add it to `supportability.yaml`.
The current approach constrains it to the currently released versions.

pytest_fixtures/component/subscription.py
```diff
@@ -24,39 +12,21 @@ def default_subscription(module_target_sat, module_org_with_manifest):
...
-def module_subscribe_satellite(module_clean_rhsm, module_target_sat):
+def module_subscribe_satellite(module_target_sat):
     """subscribe satellite to cdn"""
-    subscribe_satellite(module_target_sat)
+    module_target_sat.register_to_cdn()
+    # Enable extras repo if os_version is RHEL7
+    if module_target_sat.os_version.major < 8:
+        module_target_sat.enable_repo(
+            f'rhel-{module_target_sat.os_version.major}-server-extras-rpms', force=True
+        )
```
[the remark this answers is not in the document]
Gauravtalreja1: Yes, it would be teardown anyway, but just to be explicit, I would include a step to unregister, and as this fixture registers to CDN and not dogfood, `remove_katello_ca()` won't be required.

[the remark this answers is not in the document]
Gauravtalreja1: Updated

tests/foreman/destructive/test_katello_agent.py
```diff
@@ -149,40 +106,20 @@ def test_positive_install_package(katello_agent_client):
...
-        {'host-id': host_info['id'], 'packages': constants.FAKE_1_CUSTOM_PACKAGE_NAME}
+        {'host-id': host_info['id'], 'packages': constants.FAKE_0_CUSTOM_PACKAGE_NAME}
     )
-    result = client.run(f'rpm -q {constants.FAKE_1_CUSTOM_PACKAGE_NAME}')
-    assert result.status != 0
+    assert client.run(f'rpm -q {constants.FAKE_0_CUSTOM_PACKAGE_NAME}').status != 0
 
 
-@pytest.mark.tier3
+@pytest.mark.rhel_ver_list([6, 7, 8, 9])
```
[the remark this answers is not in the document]
Gauravtalreja1: @ogajduse NICE, Thanks! updated, please re-review

ogajduse (approved):

jameerpathan111 (approved):

robottelo/hosts.py
```diff
@@ -1118,6 +1118,28 @@ def install_tracer(self):
             raise ContentHostError('There was an error installing katello-host-tools-tracer')
         self.execute('katello-tracer-upload')
 
+    def register_to_cdn(self):
```
mshriver, line 1121: @adarshdubey-star this is relevant to your work for CloudConnector as well - thanks to @Gauravtalreja1 we've got this method that you can build off of in order to register within the test setup or test execution (against stage or production)

tests/foreman/destructive/test_katello_agent.py
```diff
@@ -47,63 +47,22 @@ def sat_with_katello_agent(module_target_sat):
...
-    rhel7_contenthost.install_katello_agent()
-    yield {'client': rhel7_contenthost, 'host_info': host_info, 'sat': sat_with_katello_agent}
+    rhel_contenthost.install_katello_agent()
+    host_info = sat_with_katello_agent.cli.Host.info({'name': rhel_contenthost.hostname})
+    yield {'client': rhel_contenthost, 'host_info': host_info, 'sat': sat_with_katello_agent}
 
 
-@pytest.mark.tier3
-@pytest.mark.upgrade
+@pytest.mark.rhel_ver_match(r'[\d]+')
```
mshriver, line 63: @rplevka Check me here, but I'm pretty sure this is redundant, the default will be to parametrize against available rhel versions, and this just matches everything.

mshriver (approved):
Looking very solid to me @Gauravtalreja1, nice work here!

tests/foreman/destructive/test_katello_agent.py
mshriver: @ogajduse I see your comment above - but you suggested a regex that ends in `$`, excluding anything that doesn't end in a digit. This expression won't exclude `_something` contenthost types/verisons.

Gauravtalreja1: @mshriver When I tried this before pushing, it collected all RHEL versions except `_fips`

omkarkhatavkar:
Looking at the blocking comments mostly are resolved. merging this as multiple acks.

PR #9712, opened 2022-06-20
fix marker resolution logic around docstring tags

rplevka:
We compare docstring components in a for loop with a list of already added markers (`item_mark_names`), that we do not update with each iteration. This causes duplicit markers being assigned e.g. in case, where docstring tokens are defined on multiple levels (module, class, function).


vsedmik (approved):

mshriver (approved):
@rplevka great catch!

PR #9714, opened 2022-06-21
Fixes for test failures

shweta83:
Test Failure fixes: 
1. ui.test_settings.py:test_positive_update_email_delivery_method_sendmail
2. api.test_location.py --> Removed Environment while creating entity

### DIFF
--- a/tests/foreman/api/test_location.py
+++ b/tests/foreman/api/test_location.py
@@ -78,7 +78,6 @@ class TestLocation:
         return dict(
             domain=entities.Domain().create(),
             subnet=entities.Subnet().create(),
-            env=entities.Environment().create(),
             host_group=entities.HostGroup().create(),
             template=entities.ProvisioningTemplate().create(),
             test_cr=entities.LibvirtComputeResource().create(),
--- a/tests/foreman/ui/test_settings.py
+++ b/tests/foreman/ui/test_settings.py
@@ -334,6 +334,7 @@ def test_negative_update_email_delivery_method_smtp():
 
 @pytest.mark.run_in_one_thread
 @pytest.mark.tier3
+@pytest.mark.skip_if_open("BZ:2080324")
 def test_positive_update_email_delivery_method_sendmail(session, target_sat):
     """Updating Sendmail params on Email tab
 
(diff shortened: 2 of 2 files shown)


akhil-jha:
I know the PR is in draft so that's why I am asking.....did the fix you made worked for rhel8 locally? Because sendmail doesn't come in rhel8 by default now. So even if you get a `"Email was sent successfully"` on UI, I don't expect it to see in /var/mail/root. You'd have to configure sendmail. Right? 
**Please correct me if I am wrong.**

jyejare:
@shweta83 Is this suppose to be in draft state ?

shweta83:
> @shweta83 Is this suppose to be in draft state ?

This was in draft state because of rhel8 sendmail issue but this test should be skipped now until the BZ is fixed. I updated the PR.

latran (commented):
I'd rather you put this back in draft and wait until the bz is fixed instead of putting the skip marker on the same bz that you're writing automation for.  It can be confusing.  And you can also add the `blocker` label until then as well.

One question though.  Is this fix also slated for 6.11?  I don't see a flag for it in the bz so the `cherrypick` label may not be applicable in this case.

shweta83:
> I'd rather you put this back in draft and wait until the bz is fixed instead of putting the skip marker on the same bz that you're writing automation for. It can be confusing. And you can also add the `blocker` label until then as well.
> 
> One question though. Is this fix also slated for 6.11? I don't see a flag for it in the bz so the `cherrypick` label may not be applicable in this case.

@latran I started this PR with fixing the failure in 6.11 due to changes in UI but it had issue with RHEL8(which impacts 6.11 as well) so I added skip marker here.

mshriver:
@latran @shweta83 I think its fine to add test cases skipped by BZs - it provides 'instant' validation of the BZ in the automation results when we get a snap containing the fix. The BZ should change states and stop blocking test case execution, and then @shweta83 can validate the bug based off of the test automation results.

This type of pattern is exactly what we need to get 'ahead' of dev in automation coverage, and push towards automated build sign-off.

latran:
Alright, I understand now.

mshriver:
@shweta83 above being said - why is the assert removed?  The assert is valid once the BZ is resolved, so it shouldn't be removed.

shweta83:
> @shweta83 above being said - why is the assert removed? The assert is valid once the BZ is resolved, so it shouldn't be removed.

@mshriver You are right. It was done by mistake. I updated the PR

ogajduse (approved):

mshriver (approved):

PR #9716, opened 2022-06-21
Fixed docker tags whitelist tests and updated helper function

ColeHiggins2:
Fixed failing tests in the CLI due to 'docker-tags-whitelist' being replaced with 'include-tags'. Also added new tests based on the repository mirroring policy. 

if it's additive, tags won't be removed
if it's mirroring, then the tags should be removed

### DIFF
--- a/tests/foreman/cli/test_repository.py
+++ b/tests/foreman/cli/test_repository.py
@@ -990,7 +990,7 @@ class TestRepository:
     )
     def test_positive_synchronize_docker_repo_set_tags_later_content_only(self, repo):
         """Verify that adding tags whitelist and re-syncing after
-        synchronizing full repository doesn't remove content that was
+        synchronizing full repository does remove content that was
         already pulled in when mirroring policy is set to content only
 
         :id: 539dc138-8566-40ad-b0de-1d9ec80aa56f
--- a/tests/foreman/cli/test_repository.py
+++ b/tests/foreman/cli/test_repository.py
@@ -916,7 +916,7 @@ class TestRepository:
                     'content-type': 'docker',
                     'docker-upstream-name': CONTAINER_UPSTREAM_NAME,
                     'url': CONTAINER_REGISTRY_HUB,
-                    'docker-tags-whitelist': 'latest',
+                    'include-tags': 'latest',
                 }
             ]
         ),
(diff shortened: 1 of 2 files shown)


tests/foreman/cli/test_repository.py
```diff
@@ -945,29 +945,30 @@ def test_positive_synchronize_docker_repo_with_tags_whitelist(self, repo_options
...
-        already pulled in
+        already pulled in when mirroring policy is set to additive
 
         :id: 97f2087f-6041-4242-8b7c-be53c68f46ff
 
         :parametrized: yes
 
-        :expectedresults: Non-whitelisted tags are not removed
+        :expectedresults: Non-whitelisted tags are not removed when
+        mirroring policy is set to additive
```
mshriver, line 964: This is causing testimony validation to fail because it doesn't handle the newline without an indent or without a numbered list. 

I don't think its really providing much anyway, as the description of the test already includes this expectation.

I'd probably just remove it.

tests/foreman/cli/test_repository.py
```diff
@@ -982,7 +982,45 @@ def test_positive_synchronize_docker_repo_set_tags_later(self, repo):
...
-                    'docker-tags-whitelist': f"latest,{gen_string('alpha')}",
+                    'mirroring-policy': 'mirror_content_only',
+                }
+            ]
+        ),
+        indirect=True,
+    )
+    def test_positive_synchronize_docker_repo_set_tags_later_content_only(self, repo):
+        """Verify that adding tags whitelist and re-syncing after
+        synchronizing full repository doesn't remove content that was
```
swadeley, line 993: Description is same as previous test, I wonder do you mean:
```suggestion
        synchronizing full repository does remove content that was
```

tests/foreman/cli/test_repository.py
```diff
@@ -982,7 +982,45 @@ def test_positive_synchronize_docker_repo_set_tags_later(self, repo):
...
+        already pulled in when mirroring policy is set to content only
+
+        :id: 539dc138-8566-40ad-b0de-1d9ec80aa56f
+
+
+        :parametrized: yes
+
+        :expectedresults: Non-whitelisted tags are removed
```
swadeley, line 1001: really tags are removed? Not content that was not tagged is removed?

swadeley (commented):
Some questions as to test purpose.

tests/foreman/cli/test_repository.py
ColeHiggins2: yes that was a typo. thanks for catching

mshriver (approved):

latran (commented):
There's also `Complete Mirroring`.  Does that impact tags as well?

swadeley (approved):
ACK

mshriver:
@latran I think your comment is relevant and @ColeHiggins2 should respond - but I think that can be addressed in a separate commit.

PR #9724, opened 2022-06-23
Use async repo synchronization

peterdragun:
This should be faster in case there are multiple repos, such as rhel7 cockpit tests. I also had to slightly extend the timeout as tests were failing due to a timeout error in the last snap.


peterdragun:
trigger: test-robottelo
pytest: -k 'test_positive_cockpit' tests/foreman/ui/test_host.py

peterdragun:
The only failing test is for RHEL9 because of the broker error while checking out.

swadeley (approved):
ACK

PR #9662, opened 2022-06-06
fixed the katello cert tests

omkarkhatavkar:
Fixed katello certs tests.  

```
============================= test session starts ==============================
collecting ... collected 13 items / 12 deselected / 1 selected

tests/foreman/sys/test_katello_certs_check.py::TestKatelloCertsCheck::test_positive_install_sat_with_katello_certs[rhel7] 

=========== 1 passed, 12 deselected, 8 warnings in 554.53s (0:09:14) ===========
```

PR #9362, opened 2022-02-10
Refactor tracer test to use rhel_contenthost fixture

rplevka:
I refactored the test for tracer integration.
It is now using `rhel_contenthost` fixture so it is parametrized for all supported rhel versions.

Also, the logic of the test was fixed as the original version wasn't testing the host-side service state properly.

Thanks @ogajduse for writing the custom, mock service for the test.

tracer is supported since rhel7, so we're excluding rhel6 scenario:
```
$ pytest -k tracer_list  tests/foreman/api/test_host.py --collect-only
==== test session starts ====
platform linux -- Python 3.9.9, pytest-6.2.5, py-1.10.0, pluggy-0.13.1
shared_function enabled - OFF - scope:  - storage: file
rootdir: /home/rplevka/work/rplevka/robottelo, configfile: pyproject.toml
plugins: forked-1.3.0, services-2.2.1, reportportal-5.0.8, mock-3.6.1, xdist-2.5.0, ibutsu-2.0.2
collected 65 items / 62 deselected / 3 selected                                                                       

<Package api>
  <Module test_host.py>
    <Class TestHostTraces>
        <Function test_positive_tracer_list_and_resolve[rhel8]>
        <Function test_positive_tracer_list_and_resolve[rhel9]>
        <Function test_positive_tracer_list_and_resolve[rhel7]>

==== 3/65 tests collected (62 deselected) in 0.19s ====

```

results (rhel9 scenario failure is expected, since we don't have the rhel9 client repo yet):

```
$ pytest -k tracer_list  tests/foreman/api/test_host.py -n3
==== test session starts ====
platform linux -- Python 3.9.9, pytest-6.2.5, py-1.10.0, pluggy-0.13.1
shared_function enabled - OFF - scope:  - storage: file
rootdir: /home/rplevka/work/rplevka/robottelo, configfile: pyproject.toml
plugins: forked-1.3.0, services-2.2.1, reportportal-5.0.8, mock-3.6.1, xdist-2.5.0, ibutsu-2.0.2
gw0 [3] / gw1 [3] / gw2 [3]
..E                                                                                                             [100%]
==== ERRORS ====
...
'errors': ["404, message='Not Found', url=URL('https://dogfood...')
...

==== 2 passed, 1 error in 647.86s (0:10:47) ====
```

PR #9729, opened 2022-06-24
[6.10.z] Fix katello-agent tests to use tools repo and register to cdn

Gauravtalreja1:
Cherrypick of https://github.com/SatelliteQE/robottelo/pull/9709 with a few more cherrypicks from master

**Dependant MR**: satlab-tower-MR#562 (merged)

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -122,6 +122,7 @@ class ContentHost(Host):
 
         :param hostname: The fqdn of a ContentHost target
         :param auth: ('root', 'rootpass') or '/path/to/keyfile.rsa'
+        :param satellite: optional parameter satellite object.
         """
         if not hostname:
             raise ContentHostError('A valid hostname must be provided')
@@ -131,13 +132,21 @@ class ContentHost(Host):
         elif isinstance(auth, str):
             # key file based authentication
             kwargs.update({'key_filename': auth})
+        self._satellite = kwargs.get('satellite')
         super().__init__(hostname=hostname, **kwargs)
+
+    @property
+    def satellite(self):
+        if not self._satellite:
+            self._satellite = Satellite()
+        return self._satellite
 
     @property
     def nailgun_host(self):
         """If this host is subscribed, provide access to its nailgun object"""
         if self.subscribed:
-            return entities.Host().search(query={'search': self.hostname})[0]
+            host_list = self.satellite.api.Host().search(query={'search': self.hostname})
+            return None if not host_list else host_list[0]
 
     @property
     def subscribed(self):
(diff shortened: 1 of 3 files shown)


Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_katello_agent.py

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_katello_agent.py

robottelo/hosts.py
```diff
@@ -122,6 +122,7 @@ def __init__(self, hostname, auth=None, **kwargs):
 
         :param hostname: The fqdn of a ContentHost target
         :param auth: ('root', 'rootpass') or '/path/to/keyfile.rsa'
+        :param satellite: optional parameter satellite object.
```
mshriver, line 125: When backporting changes like this, its good to maintain the original commit via `cherry-pick`, in order to clearly separate it from other unique changes that you've committed for this branch.

mshriver:
@Gauravtalreja1 what other PRs are these changes coming from? without having the commits split its impossible to tell what is 'new' to this PR and what is already in master and has been reviewed before.

robottelo/hosts.py
Gauravtalreja1: @mshriver  To make this PR smaller, I intentionally didn't include original commits via cherry-picks, so I just grabbed changes from `robottelo/hosts.py` to make this tests pass :)
Likewise, if you need related PRs that I referred to - #9633 #9662 #9362 #9310 and the rest are unique changes I've made for this branch.

mshriver: Thanks for the context, this helps. You have to be very careful when not chery-picking the commits, as you might be including function signature updates in `robottelo.hosts` classes without making the requisite updates that were included in the original commit(s).

pondrejk (commented):
@Gauravtalreja1 I trust you make sure that tests from cli/test_katello_agent.py are passing.  
I don't see `install_capsule_katello_ca` being used in that test set but it got modified in this PR, so if you could check that any test that actually uses that passes with your changes I would have more confidence to merge this.

robottelo/hosts.py
```diff
@@ -1402,6 +1411,24 @@ def register_to_dogfood(self):
                 f'Error during registration, command output: {cmd_result.stdout}'
             )
 
+    def register_to_cdn(self):
```
JacobCallahan, line 1414: This should go under the ContentHost class and raise a ContentHostError instead of CapsuleHostError

JacobCallahan (changes requested):

robottelo/hosts.py
Gauravtalreja1: @JacobCallahan I agree, ContentHost can also utilize this, updated, please check again, thanks.

JacobCallahan (approved):

Gauravtalreja1:
@pondrejk Yes, I tested earlier all tests with PRT and they pass, and I don't see any modification happening to `install_capsule_katello_ca`.
Instead, there is an update for `install_katello_ca` and `remove_katello_ca` which is from cherrypick PR #9633 and is used for test_katello_agent, and which works perfectly.

PR #9737, opened 2022-06-27
Reverting Broker to 0.2.0

jyejare:
Reverting to Broker 0.2.0 as its failing CI pipelines by throwing an issue for `Unauthorised`.

Identified the issue is happening only in 0.2.1 as its ignoring the `BROKER_AnsibleTower__password` probably due to https://github.com/SatelliteQE/broker/pull/155/files#diff-d41d99d0938716def67a9fefc265952f8e3745574043e364f518fa9f8a7ecf1cR29

Need to fix it and then we can jump on 0.2.1


mshriver:
This puts us back to a failure mode trying to read default vault values if the envvars aren't explicitly set (not just present in `.env` file).

There is something very screwy going on with trying to use dynaconf in broker and in robottelo at the same time.

PR #9749, opened 2022-06-28
Add ansible roles and vars tests

synkd:
This PR adds initial tests for the Satellite UI Ansible Roles and Ansible Variables pages.

These tests depend on https://github.com/SatelliteQE/airgun/pull/642/, which should be merged in the next day or so.

### DIFF
--- a/tests/foreman/ui/test_ansible.py
+++ b/tests/foreman/ui/test_ansible.py
@@ -44,6 +44,9 @@ def test_positive_import_all_roles(default_sat):
         assert available_roles_count == imported_roles_count
         ansible_variables_count = int(session.ansiblevariables.read_total_variables())
         assert ansible_variables_count > 0
+        delete_role = 'theforeman.foreman_scap_client'
+        session.ansibleroles.delete(delete_role)
+        assert not session.ansibleroles.search(delete_role)
 
 
 def test_positive_create_and_delete_variable(default_sat):
--- a/tests/foreman/ui/test_ansible.py
+++ b/tests/foreman/ui/test_ansible.py
@@ -8,7 +8,7 @@
 
 :CaseComponent: Ansible
 
-:Assignee: dsynk
+:Assignee: sbible
 
 :TestType: Functional
 
@@ -16,7 +16,7 @@
 
 :Upstream: No
 """
-from robottelo.datafactory import gen_string
+from fauxfactory import gen_string
 
 
 def test_positive_import_all_roles(target_sat):
--- a/tests/foreman/ui/test_ansible.py
+++ b/tests/foreman/ui/test_ansible.py
@@ -16,20 +16,9 @@
 
 :Upstream: No
 """
-# import datetime
-# import time
-# import pytest
 from airgun.session import Session
 
 from robottelo.datafactory import gen_string
-
-# from broker import VMBroker
-# from nailgun import entities
-
-# from fauxfactory import gen_string
-# from fauxfactory.factories.strings import gen_alpha
-# from wait_for import wait_for
-# from robottelo.hosts import ContentHost
 
 
 def test_positive_import_all_roles(default_sat):
@@ -64,6 +53,9 @@ def test_positive_create_and_delete_variable(default_sat):
     key = gen_string('alpha')
     role = 'redhat.satellite.activation_keys'
     with Session(hostname=default_sat.hostname) as session:
+        preimport_check = session.ansibleroles.preimport_check()
+        if preimport_check is False:
+            session.ansibleroles.import_all_roles()
         session.ansiblevariables.create(
             {
                 'key': key,
@@ -75,12 +67,14 @@ def test_positive_create_and_delete_variable(default_sat):
         assert not session.ansiblevariables.search(key)
 
 
-def test_positive_create_with_overrides(default_sat):
+def test_positive_create_variable_with_overrides(default_sat):
     key = gen_string('alpha')
     role = 'redhat.satellite.activation_keys'
-    param = {'attribute_type': 'fqdn', 'attribute_value': 'example.com', 'value': 'test value'}
     with Session(hostname=default_sat.hostname) as session:
-        session.ansiblevariables.create(
+        preimport_check = session.ansibleroles.preimport_check()
+        if preimport_check is False:
+            session.ansibleroles.import_all_roles()
+        session.ansiblevariables.create_with_overrides(
             {
                 'key': key,
                 'description': 'this is a description',
@@ -88,9 +82,14 @@ def test_positive_create_with_overrides(default_sat):
                 'parameter_type': 'integer',
                 'default_value': '11',
                 'validator_type': 'list',
-                'validator_rule': 'not sure what this means',
-                # 'attribute_order': 'domain \n fqdn \n hostgroup \n os',
-                'matcher_section.params': [param],
+                'validator_rule': '11, 12, 13',
+                'attribute_order': 'domain \n fqdn \n hostgroup \n os',
+                'matcher_section.params': [
+                    {
+                        'attribute_type': {'matcher_key': 'os', 'matcher_value': 'fedora'},
+                        'value': '13',
+                    }
+                ],
             }
         )
-        session.ansiblevariables.donk()
+        assert session.ansiblevariables.search(key)[0]['Name'] == key
--- a/tests/foreman/ui/test_ansible.py
+++ b/tests/foreman/ui/test_ansible.py
@@ -16,8 +16,6 @@
 
 :Upstream: No
 """
-from airgun.session import Session
-
 from robottelo.datafactory import gen_string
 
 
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_ansible.py
```diff
@@ -0,0 +1,121 @@
...
+def test_positive_import_all_roles(target_sat):
+    """Import all Ansible roles available by default.
+
+    :id: 53fe3857-a08f-493d-93c7-3fed331ed391
+
+    :Steps:
+
+        1. Navigate to the Configure > Roles page.
+        2. Click the `Import from [hostname]` button.
+        3. Get totally number of importable roles from pagination?
```
sambible, line 33: Get total number of importable roles from pagination

tests/foreman/ui/test_ansible.py
```diff
@@ -0,0 +1,121 @@
...
+
+:Requirement: Ansible
+
+:CaseAutomation: Automated
+
+:CaseLevel: Acceptance
+
+:CaseComponent: Ansible
+
+:Assignee: dsynk
```
sambible, line 11: sbible, I would think ( unless you want Ansible back? :rocket: )

tests/foreman/ui/test_ansible.py
```diff
@@ -0,0 +1,121 @@
...
+        2. Click the `Import from [hostname]` button.
+        3. Get totally number of importable roles from pagination?
+        4. Fill the `Select All` checkbox.
+        5. Click the `Submit` button.
+        6. Verify that number of imported roles == number of importable roles from step 3.
+        7. Verify that Ansible variables have been imported along with roles.
+
+    :expectedresults: All roles are imported successfully.
+    """
+    with Session(hostname=target_sat.hostname) as session:
```
Gauravtalreja1, line 41: Here, instead of importing `Session` class, you can use `ui_session` property
```suggestion
    with target_sat.ui_session as session:
```

tests/foreman/ui/test_ansible.py
```diff
@@ -0,0 +1,121 @@
...
+
+:TestType: Functional
+
+:CaseImportance: High
+
+:Upstream: No
+"""
+from airgun.session import Session
+
+from robottelo.datafactory import gen_string
```
Gauravtalreja1, line 21: Please use fauxfactory here
```suggestion
from fauxfactory import gen_string
```

tests/foreman/ui/test_ansible.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+    with Session(hostname=target_sat.hostname) as session:
+        available_roles_count = session.ansibleroles.import_all_roles()
+        imported_roles_count = session.ansibleroles.imported_roles_count
+        assert available_roles_count == imported_roles_count
+        ansible_variables_count = int(session.ansiblevariables.read_total_variables())
+        assert ansible_variables_count > 0
+        delete_role = 'theforeman.foreman_scap_client'
+        session.ansibleroles.delete(delete_role)
+        assert not session.ansibleroles.search(delete_role)
```
Gauravtalreja1, line 49: Optional: you might want to optimize this a bit
```suggestion
        assert session.ansibleroles.import_all_roles() == session.ansibleroles.imported_roles_count
        assert int(session.ansiblevariables.read_total_variables()) > 0
        session.ansibleroles.delete('theforeman.foreman_scap_client')
        assert not session.ansibleroles.search('theforeman.foreman_scap_client')
```

tests/foreman/ui/test_ansible.py
```diff
@@ -0,0 +1,121 @@
...
+        4. Delete the Ansible variable.
+        5. Verify that the Ansible Variable has been deleted.
+
+    :expectedresults: The variable is successfully created and deleted.
+    """
+    key = gen_string('alpha')
+    role = 'redhat.satellite.activation_keys'
+    with Session(hostname=target_sat.hostname) as session:
+        preimport_check = session.ansibleroles.preimport_check()
+        if preimport_check is False:
```
Gauravtalreja1, line 71:
```suggestion
        if session.ansibleroles.preimport_check() is False:
```

tests/foreman/ui/test_ansible.py
```diff
@@ -0,0 +1,121 @@
...
+:CaseImportance: High
+
+:Upstream: No
+"""
+from airgun.session import Session
+
+from robottelo.datafactory import gen_string
+
+
+def test_positive_import_all_roles(target_sat):
```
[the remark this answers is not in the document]
synkd: There's not an easy way to delete all imported roles via the UI, but I think you're right that cleanup would be a good idea. Locally, I've been checking the first and last ID numbers of the imported roles and using them in a bash `for` loop to delete them. For example:
```
# hammer ansible roles list
# for i in {1, 26}; do hammer ansible roles delete --id $i; done
```
I've been trying to come up with an awk/bash one-liner to extract those numbers and pipe them to the loop, but I haven't found a way that's not brittle (e.g. `hammer ansible roles list | awk 'NR==4 || NR==28 { print $1}` extracts the numbers but will break if the number of default roles changes). I'm open to suggestions on implementing cleanup.

synkd:
trigger: test-robottelo
pytest: tests/foreman/ui/test_ansible.py

tests/foreman/ui/test_ansible.py
```diff
@@ -0,0 +1,117 @@
...
+        3. Verify that the Ansible variable has been created.
+        4. Delete the Ansible variable.
+        5. Verify that the Ansible Variable has been deleted.
+
+    :expectedresults: The variable is successfully created and deleted.
+    """
+    key = gen_string('alpha')
+    role = 'redhat.satellite.activation_keys'
+    with target_sat.ui_session as session:
+        if session.ansibleroles.preimport_check() is False:
```
mshriver, line 68:
```suggestion
        if not session.ansibleroles.preimport_check():
```

tests/foreman/ui/test_ansible.py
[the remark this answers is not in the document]
mshriver: Not going to consider this a blocker for this PR, but I think there is room for improvement here on cleanup and using a fixture to do the import (as @sambible suggested in previous comment) is a good strategy to deal with both.

Right now, doesn't look like AnsibleRoles has been defined as a nailgun entity. That API implementation would be ideal for capturing what roles got imported during test setup, and using the API to delete them during teardown.

tests/foreman/ui/test_ansible.py
mshriver: Actually you can't do that because this function is pretty broken:
https://github.com/SatelliteQE/airgun/blob/master/airgun/entities/ansible_role.py#L54

It returns None or False, which is pretty ambiguous.  Maybe I should have caught that in the airgun review, we were focusing on bigger issues though.

Acceptable as is, but I'd like you to follow this up with a small PR to rename `preimport_check` to something that makes it clear what's being checked and what is being returned. Something like `AnsibleRolesEntity.roles_exist`.  Better yet, don't have the function at all, and use the integer returned by `imported_roles_count`. That's falsy when `0`, and you could use it's return in the same way here. It would be clear what's being checked. You'll need to verify `view.total_imported_roles.read()` doesn't blow up when the page is empty.

Difficult to keep track of how these small pieces fit together when defining new entities and views from scratch.

synkd: @mshriver I just tried your suggestion of using the integer returned by `imported_roles_count` for this check. Unfortunately, it does indeed blow up if the page is empty. Prior to importing roles, no table is present on the Ansible Roles page. That's why the `preimport_check` function evaluates against `view.pagination.is_displayed` - if there's no pagination displayed, then no roles have been imported.

With this context in mind, would renaming the `preimport_check` function to something more descriptive address your comment sufficiently?

mshriver: I think the imported_roles_count function needs to be updated to not raise exception, but return 0, when there's no table.

Then here you can just directly use `if not imported_roles_count:` to trigger an import.

synkd: @mshriver Ack, I've just opened an Airgun PR that makes this change: https://github.com/SatelliteQE/airgun/pull/726

mshriver (approved):

mshriver:
trigger: test-robottelo
pytest: tests/foreman/ui/test_ansible.py

sambible (approved):
Looks good @synkd

tests/foreman/ui/test_ansible.py
```diff
@@ -0,0 +1,117 @@
...
+        4. Fill the `Select All` checkbox.
+        5. Click the `Submit` button.
+        6. Verify that number of imported roles == number of importable roles from step 3.
+        7. Verify that Ansible variables have been imported along with roles.
+        8. Delete an imported role.
+        9. Verify that the role was successfully deleted.
+
+    :expectedresults: All roles are imported successfully. One role is deleted successfully.
+    """
+    with target_sat.ui_session as session:
```
mshriver, line 41: This was very recently changed to a function instead of a property, needs to be `target_sat.ui_session()`

PR #9750, opened 2022-06-29
Add stubs for new host UI, content tab

peterdragun:


tests/foreman/ui/test_host.py
```diff
@@ -2234,6 +2234,69 @@ def test_positive_read_details_page_from_new_ui(
...
+@pytest.mark.stubbed
+def test_positive_update_delete_package():
+    """Update package on host using the new Content tab
+
+    :id: ffc19a40-85f4-4894-a18b-f6d88b2ce377
+
+    :steps:
+        1. Install a package on registered host.
+        2. Navigate to Content tab.
+        3. Check is package is in upgradable state.
```
tstrych, line 2247: `Check is package` makes no sense, please fix the sentence. 
Don't know how much we are playing for an English, but  you are missing a lot of articles (a/the), if you don't mind copy the text to Grammarly and you can fix it in no time :).

tests/foreman/ui/test_host.py
```diff
@@ -2234,6 +2234,69 @@ def test_positive_read_details_page_from_new_ui(
...
+        4. Select package and upgrade via rex.
```
tstrych, line 2248: I believe the traces tab can be tested within this test, as you updated the package, (if it is a service it needs to be restarted once its updated, but that would add another level to this test, so its better to think about it from the start, or we can create a separate test, let me know what you think )

tests/foreman/ui/test_host.py
```diff
@@ -2234,6 +2234,69 @@ def test_positive_read_details_page_from_new_ui(
...
+
+    :expectedresults: Package is updated and deleted
+
+    """
+
+
+@pytest.mark.tier4
+@pytest.mark.stubbed
+def test_positive_apply_erratum():
+    """Apply arratum on host using the new Errata tab
```
tstrych, line 2259: replace `arratum` for `erratum`

tstrych (commented):
Thanks for adding it, please work on minor comments.

tests/foreman/ui/test_host.py
```diff
@@ -2234,6 +2234,69 @@ def test_positive_read_details_page_from_new_ui(
...
+def test_positive_crud_module_streams():
+    """CRUD test for Module streams new UI tab
+
+    :id: 9800a006-49cc-4c0a-aed8-6a32c4bf0eab
+
+    :setup:
+        1. Valid yum repo with Module Streams.
+
+    :steps:
+        1. Create Yum Repository with url contain module-streams
```
tstrych, line 2289: reformulate, e.g `Create Yum Repository which contains module-streams as URL.`
or `Create Yum Repository with URL containing module-streams`

ColeHiggins2 (approved):
ACK

tstrych (approved):
ACK

PR #9752, opened 2022-06-29
Adding Webhooks API tests

Griffin-Sullivan:
Creating a new file and some initial tests for Webhooks API testing. 

Depends on: https://github.com/SatelliteQE/nailgun/pull/834

### DIFF
--- a/tests/foreman/api/test_webhook.py
+++ b/tests/foreman/api/test_webhook.py
@@ -16,19 +16,158 @@
 
 :Upstream: No
 """
+import re
+
 import pytest
 from nailgun import entities
+from requests.exceptions import HTTPError
+from wait_for import TimedOutError
+from wait_for import wait_for
+
+from robottelo.config import settings
+from robottelo.constants import WEBHOOK_EVENTS
+from robottelo.constants import WEBHOOK_METHODS
+from robottelo.datafactory import parametrized
+from robottelo.logging import logger
+
+
+def _read_log(ch, pattern):
+    """Read the first line from the given channel buffer and return the matching line"""
+    # read lines until the buffer is empty
+    for log_line in ch.stdout().splitlines():
+        logger.debug(f'foreman-tail: {log_line}')
+        if re.search(pattern, log_line):
+            return log_line
+    else:
+        return None
+
+
+def _wait_for_log(channel, pattern, timeout=2, delay=0.2):
+    """_read_log method enclosed in wait_for method"""
+    matching_log = wait_for(
+        _read_log,
+        func_args=(
+            channel,
+            pattern,
+        ),
+        fail_condition=None,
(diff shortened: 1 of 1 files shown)


tests/foreman/api/test_webhook.py
```diff
@@ -0,0 +1,173 @@
...
+        """Test positive webhook creation with a valid event
+
+        :id: 9b505f1b-7ee1-4362-b44c-f3107d043a05
+
+        :expectedresults: Webhook is created with event specified
+
+        :CaseImportance: High
+        """
+        hook = entities.Webhooks(event=event).create()
+        assert event in hook.event
```
Gauravtalreja1, line 97: As this test for creating valid events and methods is already being performed in the end-to-end test. Could this test be consolidated?

tests/foreman/api/test_webhook.py
```diff
@@ -0,0 +1,173 @@
...
+        """Test negative webhook creation with an invalid event
+
+        :id: 60cd456a-9943-45cb-a72e-23a83a691499
+
+        :expectedresults: Webhook is not created
+
+        :CaseImportance: High
+        """
+        with pytest.raises(HTTPError):
+            entities.Webhooks(event='invalid_event').create()
```
Gauravtalreja1, line 83: Instead of calling with entities, I would suggest using target_sat/class_target_sat

tests/foreman/api/test_webhook.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        hook = entities.Webhooks(
+            event='actions.katello.repository.sync_succeeded', http_method='GET'
+        ).create()
+        repo = entities.Repository(
+            organization=module_org, content_type='yum', url=settings.repos.yum_0.url
+        ).create()
+        with target_sat.session.shell() as shell:
+            shell.send('foreman-tail')
+            repo.sync()
```
Gauravtalreja1, line 172: Could `repo.sync()` be called before the context manager here?

tests/foreman/api/test_webhook.py
Griffin-Sullivan: Why is that?

tests/foreman/api/test_webhook.py
Griffin-Sullivan: This is being parameterized with all the events from `WEBHOOK_EVENTS`, so it is for ensuring each event listed is valid. This is not being tested in end to end.

tests/foreman/api/test_webhook.py
Gauravtalreja1: It's good to have, but it's not strictly necessary

Gauravtalreja1 (approved):

PR #9638, opened 2022-05-31
Update robottelo to use Broker 0.2+

JacobCallahan:
Aside from the updated requirement, the VMBroker class has been renamed
to Broker. This commit's bulk changes adjust the class to the new name.

PR #9763, opened 2022-07-04
[CherryPick to 6.10] Update robottelo to use Broker 0.2+ (#9638)

jyejare:
Aside from the updated requirement, the VMBroker class has been renamed
to Broker. This commit's bulk changes adjust the class to the new name.

Fixes the failing checks for 6.10.z PRs now for broker conflict between upgrade and robottelo.


pytest_fixtures/core/contenthosts.py
```diff
@@ -221,3 +221,17 @@ def set_remote_execution_by_default(request, module_target_sat):
...
+    request.param = {"rhel_version": version.split('.')[0], "distro": "centos"}
+    with Broker(**host_conf(request), host_classes={'host': ContentHost}) as host:
+        yield host
+
+
+@pytest.fixture
+def oracle_host(request, version):
+    request.param = {"rhel_version": version.split('.')[0], "distro": "oracle"}
+    with Broker(**host_conf(request), host_classes={'host': ContentHost}) as host:
+        yield host
```
Gauravtalreja1, line 237: I guess Convert2RHEL is not supported for 6.10, I'm unsure if this is required for 6.10.z, @shweta83 can confirm?

pytest_fixtures/core/broker.py
```diff
@@ -29,7 +29,7 @@ def _target_sat_imp(request, _default_sat, satellite_factory):
     if request.node.get_closest_marker(name='destructive'):
         new_sat = satellite_factory()
         yield new_sat
-        VMBroker(hosts=[new_sat]).checkin()
+        Broker(hosts=[new_sat]).checkin()
```
Gauravtalreja1, line 32: Could you please include a step to unregister sat_host before checkin?
```suggestion
        new_sat.unregister()
        Broker(hosts=[new_sat]).checkin()
```

pytest_fixtures/core/contenthosts.py
shweta83: Yes @Gauravtalreja1, we have designed test case for RHEL8 as well in `Convert2Rhel`  so it won't work on 6.10.

jyejare: Ok, Make sense ! Even I was confused about adding or removing this. Removing now !

Gauravtalreja1 (approved):

PR #9768, opened 2022-07-08
Remove cleanup.py

sambible:
For Helpers refactor - Removing cleanup.py, and replacing it's remaining calls with the direct calls, or adding the function to the test tile - whichever would be a smaller code impact.

As a bonus, fixing a test that had been broken due to the message format changing, so that I could see it passing with the aforementioned cleanup refactor

(If it makes sense, I can split that piece into a separate PR)


tests/foreman/api/test_location.py
```diff
@@ -27,8 +27,8 @@
 from nailgun import entities
 from requests.exceptions import HTTPError
 
-from robottelo.cleanup import capsule_cleanup
 from robottelo.cli.factory import make_proxy
+from robottelo.cli.proxy import Proxy
```
JacobCallahan, line 31: Same suggestion

tests/foreman/api/test_smartproxy.py
```diff
@@ -22,7 +22,7 @@
 from requests import HTTPError
 
 from robottelo.api.utils import one_to_many_names
-from robottelo.cleanup import capsule_cleanup
+from robottelo.cli.proxy import Proxy
```
JacobCallahan, line 25: You know the thing

JacobCallahan (changes requested):

tests/foreman/cli/test_realm.py
```diff
@@ -31,6 +31,11 @@
 pytestmark = [pytest.mark.run_in_one_thread]
 
 
+def realm_cleanup(realm_id=None):
```
mshriver, line 34: Here its used via `request.addfinalizer`, and you'll want the satellite instance so you can use `target_sat.api.Realm(id=realm_id).delete()`.  Since its getting created in the test and you don't know the ID, I'd just pass the satellite instance in too.

You could use pytest classes here to store `self.realm_id` within the test CALL phase and then write this as a yielding fixture, being able to reference `self.realm_id` instead of depending on the test passing it. Don't think its worth the shuffle though.

mshriver (changes requested):
Couple small things to use target_sat, should be quick and easy to knock this one out!

pytest_fixtures/component/smartproxy.py
```diff
@@ -32,6 +31,6 @@ def module_fake_proxy(request, module_target_sat):
             args['url'] = url
             proxy = module_target_sat.api.SmartProxy(**args).create()
             yield proxy
-            capsule_cleanup(proxy.id)
+            target_sat.cli.Proxy.delete({'id': proxy.id})
```
mshriver, line 34: You can use `module_target_sat` here, no need to include both.

tests/foreman/api/test_location.py
```diff
@@ -61,13 +60,13 @@ class TestLocation:
     # TODO Add coverage for media, realms as soon as they're implemented
 
     @pytest.fixture
-    def make_proxies(self, options=None):
+    def make_proxies(self, target_sat, options=None):
```
mshriver, line 63: At the risk of being annoying, its so small that we should just fix it here. No fixtures should be taking kwargs with defaults - and nothing in this test module is trying to call this function.  This `options` kwarg needs to be removed.

tests/foreman/cli/test_capsule.py
```diff
@@ -40,7 +39,7 @@ def _make_proxy(request, options=None):
 
     @request.addfinalizer
     def _cleanup():
-        capsule_cleanup(proxy['id'])
+        Proxy.delete({'id': proxy['id']})
```
mshriver, line 42: Since this isn't the only use of the import I'll let it slide - but ideally this is also taking `target_sat` from the test function, just like in test_smartproxy.py.

tests/foreman/cli/test_realm.py
```diff
@@ -31,6 +30,11 @@
 pytestmark = [pytest.mark.run_in_one_thread]
 
 
+def realm_cleanup(target_sat, realm_id=None):
```
mshriver, line 33: The callers in this module don't pass target_sat, and this isn't a fixture.

Given that its a single line, it doesn't make sense to wrap it in a function. I'd just replace the calls to this function with `request.addfinalizer(lambda: target_sat.cli.Realm(id=realm_id).delete())` throughout the module. 

Otherwise you have to update every call to pass the target_sat.

mshriver (changes requested):
Couple more things, looking great though, this consistency matters.

pytest_fixtures/component/smartproxy.py
sambible: Ah, I thought I probably could.

tests/foreman/cli/test_capsule.py
sambible: I'll change it, no reason to be inconsistent

tests/foreman/api/test_location.py
```diff
@@ -61,13 +60,13 @@ class TestLocation:
     # TODO Add coverage for media, realms as soon as they're implemented
 
     @pytest.fixture
-    def make_proxies(self, options=None):
+    def make_proxies(self, target_sat):
         """Create a Proxy"""
-        proxy1 = make_proxy(options=options)
-        proxy2 = make_proxy(options=options)
+        proxy1 = make_proxy()
+        proxy2 = make_proxy()
```
JacobCallahan, line 66:
```suggestion
        proxy1 = target_sat.cli_factory.make_proxy()
        proxy2 = target_sat.cli_factory.make_proxy()
```

tests/foreman/api/test_smartproxy.py
```diff
@@ -48,13 +47,13 @@ def module_proxy_attrs(module_target_sat):
     return set(smart_proxy.update_json([]).keys())
 
 
-def _create_smart_proxy(request, **kwargs):
+def _create_smart_proxy(request, target_sat, **kwargs):
     """Create a Smart Proxy and add the finalizer"""
     proxy = entities.SmartProxy(**kwargs).create()
```
JacobCallahan, line 52:
```suggestion
    proxy = target_sat.api.SmartProxy(**kwargs).create()
```

tests/foreman/api/test_smartproxy.py
```diff
@@ -48,13 +47,13 @@ def module_proxy_attrs(module_target_sat):
...
 
     @request.addfinalizer
     def _cleanup():
-        capsule_cleanup(proxy.id)
+        target_sat.cli.Proxy.delete({'id': proxy.id})
```
JacobCallahan, line 56: this might be better done via api

tests/foreman/cli/test_hostgroup.py
```diff
@@ -268,7 +267,7 @@ def test_positive_update_hostgroup(
         def _cleanup():
             with session_puppet_enabled_sat:
                 HostGroup.delete({'id': hostgroup['id']})
-                capsule_cleanup(new_content_source['id'])
+                Proxy.delete({'id': new_content_source['id']})
```
JacobCallahan, line 270: Could (and maybe should) also use the API here
```suggestion
                session_puppet_enabled_sat.cli.Proxy.delete({'id': new_content_source['id']})
```

tests/foreman/cli/test_realm.py
```diff
@@ -88,7 +93,7 @@ def test_positive_realm_info_name(
             'locations': [loc.read().name for loc in module_fake_proxy.location],
         }
     )
-    request.addfinalizer(lambda: realm_cleanup(realm['id']))
+    request.addfinalizer(lambda: target_sat.cli.Realm(realm['id']).delete())
     info = Realm.info({'name': realm['name']})
```
JacobCallahan, line 97: lots of mixes between target_sat's Realm and the default Realm. This could lead to issues

JacobCallahan (changes requested):

tests/foreman/api/test_smartproxy.py
sambible: So does broadly preferring to use target_sat over entities make sense

tests/foreman/cli/test_realm.py
mshriver: I think @JacobCallahan is referring to the fact that this is a destructive test, but I don't think you're deleting the default realm here?

That being said, this test function will be moving with #9697 and you'll have to rebase on top of it.

tests/foreman/api/test_smartproxy.py
mshriver: Yep, great to update when and where we can, but its not strictly required yet.

tests/foreman/cli/test_realm.py
mshriver: destructive test move was merged, @sambible time to rebase - some of these changes will have moved to new modules.

sambible:
Rebased and added a few more fixes ( and shocked that it went well, rebases have been my git kryptonite for a long time )

Going to address a few of the other suggestions.

JacobCallahan (approved):
ACK. Merge pending open comments

sambible:
Does this need a CherryPick label since i had to rebase?

mshriver (approved):

PR #9741, opened 2022-06-27
Adding Webhook CLI tests

Griffin-Sullivan:
This PR is for adding some basic CRUD tests for Webhooks in hammer CLI. I copied the functionality of cli/test_discoveryrule.py, since I liked how the `<component>_factory` fixture works. I am also using this PR to replace https://github.com/SatelliteQE/robottelo/pull/9223.

Results:
============================= test session starts ==============================
platform linux -- Python 3.9.13, pytest-7.1.2, pluggy-0.13.1
shared_function enabled - OFF - scope:  - storage: file
rootdir: /home/gsulliva/Programming/robottelo, configfile: pyproject.toml
plugins: forked-1.3.0, services-2.2.1, cov-2.12.1, xdist-2.5.0, mock-3.7.0, reportportal-5.1.2, ibutsu-2.2.1
collected 60 items

tests/foreman/cli/test_webhook.py ...................................... [ 63%]
......................                                                   [100%]
========================= 60 passed in 392.63s (0:06:32) ========================

PR #9779, opened 2022-07-13
Adding UI Webhook tests

Griffin-Sullivan:
Basic UI testing of Webhooks CRUD functionality. Depends on https://github.com/SatelliteQE/airgun/pull/723

### DIFF
--- a/tests/foreman/ui/test_webhook.py
+++ b/tests/foreman/ui/test_webhook.py
@@ -1,0 +1,81 @@
+"""Test class for Webhook UI
+
+:Requirement: Webhooks
+
+:CaseAutomation: Automated
+
+:CaseLevel: Component
+
+:CaseComponent: Webhooks
+
+:Assignee: gsulliva
+
+:TestType: Functional
+
+:CaseImportance: High
+
+:Upstream: No
+"""
+import pytest
+from fauxfactory import gen_string
+from fauxfactory import gen_url
+
+
+@pytest.mark.tier1
+def test_positive_end_to_end(session):
+    """Perform end to end testing for webhooks.
+
+    :id: 9d2072e7-c6b6-4ecb-ac84-e73f4cf76fe4
+
+    :expectedresults: All expected CRUD actions finished successfully
+
+    :CaseImportance: Critical
+    """
+    hook_name = gen_string('alpha')
+    subscribe_to = 'Host Created'
+    target_url = gen_url(subdomain=gen_string('alpha'), scheme='http')
+    template = 'Katello Publish'
+    http_method = 'GET'
+    new_hook_name = gen_string('alpha')
+    new_subscribe_to = 'Host Destroyed'
+    new_target_url = gen_url(subdomain=gen_string('alpha'), scheme='http')
+    new_template = 'Katello Promote'
+    new_http_method = 'PUT'
+    with session:
+        session.webhook.create(
+            {
+                'general.subscribe_to': subscribe_to,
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_webhook.py
```diff
@@ -0,0 +1,81 @@
+"""Test class for Webhook UI
+
+:CaseAutomation: Automated
+
+:CaseLevel: Component
+
+:CaseComponent: Webhooks
```
Gauravtalreja1, line 7: Please include this component in testimony.yaml to fix failing GHA

Griffin-Sullivan: Just waiting on https://github.com/SatelliteQE/robottelo/pull/9741 and I'll rebase off that because I still need review on the airgun PR before this can be approved.

jawatts: @Griffin-Sullivan @mshriver done

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/ui/test_webhook.py
airgun: 723

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/ui/test_webhook.py
airgun: 723

Griffin-Sullivan:
Seems like production is having some weird problems with filling the text input in Webhook update. Moving this and airgun pr to draft.

Griffin-Sullivan:
Solved issue that I had in production.

tests/foreman/ui/test_webhook.py
```diff
@@ -0,0 +1,83 @@
...
+
+    :id: 9d2072e7-c6b6-4ecb-ac84-e73f4cf76fe4
+
+    :expectedresults: All expected CRUD actions finished successfully
+
+    :CaseImportance: Critical
+    """
+    hook_name = gen_string('alpha')
+    subscribe_to = 'Host Created'
+    target_url = gen_url(subdomain=gen_string('alpha'), scheme=choice(['http', 'https']))
```
mshriver, line 38: How long does this test take to execute? I'd rather parametrize this instead of potentially making the test flaky. There's also no guarantee below that the new target url will be using a different scheme than the original - if that's even intended coverage on the update?

Griffin-Sullivan: It takes 4 minutes to execute. Can you define flaky? Also the intention with the url scheme is that webhooks can not be `ftp`, so I'm just ensuring it generates a valid scheme for the test.

mshriver: Flaky would come from a potential bug associated with use of one type of protocol. So when the test randomly chooses the working protocol, things are fine - but when it picks the broken protocol, randomly, the test fails. You go to run it locally to reproduce the CI failure - and it passes because it randomly chose the working protocol.

It may sound far fetched, but @jaryn just unraveled this exact situation in the bookmark tests.

mshriver: Correct, or just consistently use one.  Since the host isn't real, I don't think there's much coverage provided by using both protocols.

If you want to validate the allowed protocol types, given the execution time, I'd consider having a separate test function to fill the UI field with http/https/ftp/file protocols and check the validation behavior specifically.

Griffin-Sullivan: Agreed. Going to limit to just `http`, since we aren't gaining much coverage.

mshriver (approved):

PR #9782, opened 2022-07-15
[Part-1] Helpers refactor of helpers.py helpers :)

jyejare:
Helpers and their refactors - Part 1:

- Fact: gen_ssh_keypairs is only used in Ssh Key In User Future tests
  - Consideration: The gen_ssh_keypairs is a very generic function doesnt belongs to any specific considerations like endpoint_factory, fixtures, endpoint object, host classes and their mixins, its a pure utility function that doesnt need separate module. Hence putting in robottelo/utils/__init__.py

- Fact: ServerFileDownloader or its object download_server_file is not bing used anywhere in robottelo.
  - Consideration: File_downloader method is a better alternative to ServerFileDownloader and is being widely used since its allows to download the file locally as well as on any server. Hence removing that helper class is a big plus.

- Fact: File_downloader is being used only for downloading the file from url to satellite.
  - Consideration: File_downloader could be completely moved to Host class / mixins.

- Fact: Line_count function is a one line function used in test_logging.py only.
  - Consideration: Line_count function is a one line function and hence can be directly replaced with function code

- Fact: Cut_lines function is a one line function used in test_logging.py only but its lengthy to fit in one line to replace the usage by its code.
  - Consideration: Cut_lines function is a one line function but its lengthy so directly pasted the function in the test module.

- Fact: escape_search is not used anywhere in robottelo except unit testing that function itself.
  - Consideration: Removing the function along with its unit tests.

- Fact: update_dictionary is used once but in depreciated cli.factory.py modules create_object function.
  - Consideration: Remove it.

- Fact: get_data_file is used multiple times but the function is using os module to find the absolute path and then returns that path.
  - Consideration: Remove function and replace its usage by pathlib.PurePath
- Fact: read_data_file is used multiple times but the function is using os module to find the absolute path, read the file and then returns that path file contents.
  - Consideration: Remove function and replace its usage by pathlib.Path

### DIFF
--- a/pytest_fixtures/component/oscap.py
+++ b/pytest_fixtures/component/oscap.py
@@ -1,5 +1,6 @@
 from pathlib import PurePath
 
+
 import pytest
 from fauxfactory import gen_string
 from nailgun import entities
@@ -9,15 +10,15 @@ from robottelo.config import robottelo_tmp_dir
 from robottelo.config import settings
 from robottelo.constants import OSCAP_PROFILE
 from robottelo.constants import OSCAP_TAILORING_FILE
-from robottelo.helpers import get_data_file
+from robottelo.constants import DataFile
 
 
 @pytest.fixture(scope="session")
 def tailoring_file_path(session_target_sat):
     """Return Tailoring file path."""
-    local = get_data_file(OSCAP_TAILORING_FILE)
+    local = DataFile.OSCAP_TAILORING_FILE
     session_target_sat.put(
-        local_path=get_data_file(OSCAP_TAILORING_FILE),
+        local_path=local,
         remote_path=f'/tmp/{OSCAP_TAILORING_FILE}',
     )
     return {'local': local, 'satellite': f'/tmp/{OSCAP_TAILORING_FILE}'}
--- a/pytest_fixtures/component/oscap.py
+++ b/pytest_fixtures/component/oscap.py
@@ -1,6 +1,5 @@
 from pathlib import PurePath
 
-
 import pytest
 from fauxfactory import gen_string
 from nailgun import entities
@@ -8,9 +7,9 @@ from nailgun import entities
 from robottelo.cli.factory import make_scapcontent
 from robottelo.config import robottelo_tmp_dir
 from robottelo.config import settings
+from robottelo.constants import DataFile
 from robottelo.constants import OSCAP_PROFILE
 from robottelo.constants import OSCAP_TAILORING_FILE
-from robottelo.constants import DataFile
 
 
 @pytest.fixture(scope="session")
--- a/robottelo/utils/__init__.py
+++ b/robottelo/utils/__init__.py
@@ -1,7 +1,12 @@
-# General utility functions which does not fit into other util modules
+# General utility functions which does not fit into other util modules OR
+# Independent utility functions that doesnt need separate module
 import os
 import re
 from pathlib import Path
+
+from cryptography.hazmat.backends import default_backend as crypto_default_backend
+from cryptography.hazmat.primitives import serialization as crypto_serialization
+from cryptography.hazmat.primitives.asymmetric import rsa
 
 from robottelo.constants import Colored
 from robottelo.errors import InvalidVaultURLForOIDC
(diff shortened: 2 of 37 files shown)


robottelo/host_helpers/host_mixins.py
```diff
@@ -0,0 +1,32 @@
...
+import os
+
+from robottelo.errors import DownloadFileError
+
+
+class IOHelpers:
+    def download_file(self, file_url, local_path=None, file_name=None):
+        """Downloads file from given fileurl to directory specified by local_path
+        with given file_name on host specified by hostname. Leave hostname as None
+        to download file on the localhost.If remote directory is not specified it
```
jameerpathan111, line 12:
```suggestion
        with given file_name on Satellite. If remote directory is not specified it
```

jameerpathan111 (changes requested):

pytest_fixtures/component/oscap.py
```diff
@@ -9,15 +9,14 @@
...
 from robottelo.constants import OSCAP_PROFILE
 from robottelo.constants import OSCAP_TAILORING_FILE
-from robottelo.helpers import get_data_file
 
 
 @pytest.fixture(scope="session")
 def tailoring_file_path(session_target_sat):
     """Return Tailoring file path."""
-    local = get_data_file(OSCAP_TAILORING_FILE)
+    local = PurePath('tests/foreman/data').joinpath(OSCAP_TAILORING_FILE)
```
mshriver, line 17: I don't think this change is a good one. There are now 46 individual direct references to this file path.

At least defining the path instance as an importable variable would be acceptable, but I don't think this helper is necessarily one that needs to go away.

There's a need/want to separate the file name itself from the local path, so that the file can be referenced on the remote server too.  This can be done with the properties of the Path instance, so maybe creating a Box class with file instances for the available data files would be a good way to directly reference files via a single import?

robottelo/host_helpers/host_mixins.py
```diff
@@ -0,0 +1,32 @@
+# Host helpers related to File Downloading and operations on host objects
+# Replace os with tmpfile library
+import os
+
+from robottelo.errors import DownloadFileError
+
+
+class IOHelpers:
+    def download_file(self, file_url, local_path=None, file_name=None):
```
mshriver, line 9: This looks like overlap with this function:

https://github.com/SatelliteQE/robottelo/blob/87a748f761637ce44bfc979818701c5d4867f553/robottelo/hosts.py#L538

Starting to get concerned about the subclassing being done here, and the potential for duplicate definitions and namespace collisions.

robottelo/utils/__init__.py
```diff
@@ -0,0 +1,42 @@
+# Independent utility functions that doesnt need separate module and
+import base64
+
+from cryptography.hazmat.backends import default_backend as crypto_default_backend
+from cryptography.hazmat.primitives import serialization as crypto_serialization
+from cryptography.hazmat.primitives.asymmetric import rsa
+
+
+def gen_ssh_keypairs():
```
mshriver, line 9: Why not in `robottelo.ssh`?

pytest_fixtures/component/oscap.py
jyejare: Yeah, Even I felt bit weird for this. But now as per your comment, I could create class inheriting `Box` class in `robottelo/constants.py` whereas all the file paths could be instances of it. Something like :

```
class DataFile(Box):
    DATA_DIR = PurePath('tests/foreman/data')
    OSCAP_TAILORING_FILE = DATA_DIR.joinpath(OSCAP_TAILORING_FILE)
```

Access as:
```
from robottelo.constants import DataFile
local = DataFile.OSCAP_TAILORING_FILE
```
??

robottelo/host_helpers/host_mixins.py
jyejare: That's a `put` operation whereas https://github.com/SatelliteQE/robottelo/blob/87a748f761637ce44bfc979818701c5d4867f553/robottelo/hosts.py#L534 is the right method.

I wonder why its just for ContentHost and not in `Host` base class. But even thats the case, we can remove `download_file` and start using `get` altogether.

robottelo/utils/__init__.py
jyejare: The `robottelo.ssh` is host based module like firing command on some host. But this function is a private and public SSH Key generator function which is non-host based. By adding this to the function ssh module, I dont want people to feel its generating ssh keypair for any specific host.

pytest_fixtures/component/oscap.py
mshriver: Yep, exactly the kind of thing I was thinking of.  Callers don't have to import the whole module, and there's a singular point to list/locate data files consistently as PurePath objects.

jyejare: Done !

robottelo/host_helpers/host_mixins.py
jyejare: Done !

mshriver (approved):
Great change @jyejare, a chunk of helpers gone, and better patterns established for construction and use of constants.

mshriver:
This does not need to be cherry-picked into 6.10.z or 6.9.z branches.

robottelo/host_helpers/host_mixins.py
mshriver: Well, both put and get are both already defined there, I was just linking one. 

Going to push a commit to your PR to fix the broken import since this file/class was removed.

JacobCallahan (approved):

PR #9785, opened 2022-07-18
Change foreman service teardown

swadeley:
Hello

1] I think we should remove the `--only=foreman` because I think the aim of this fixture is to return Satellite to its original state, and due to bug [1] we should not assume only foreman needs starting.

2] Is this fixture aptly named?  In stead of `foreman_service_teardown` would `foreman_service_restore` be more accurate?

[1] [Bug 2104134](https://bugzilla.redhat.com/show_bug.cgi?id=2104134) - Stopping foreman also stops dynflow-sidekiq

### DIFF
--- a/pytest_fixtures/core/sys.py
+++ b/pytest_fixtures/core/sys.py
@@ -10,7 +10,7 @@ from robottelo.hosts import SatelliteHostError
 def foreman_service_teardown(target_sat):
     """stop and restart of foreman service"""
     yield target_sat
-    target_sat.execute('satellite-maintain service start --only=foreman')
+    target_sat.execute('satellite-maintain service start')
 
 
 @pytest.fixture


.github/workflows/merge_to_master.yml
```diff
@@ -4,7 +4,7 @@ name: update_robottelo_image
 on:
   push:
     branches:
-      - master
+      - 6.9.z
```
Gauravtalreja1, line 7: Maybe this wasn't meant to be? You might want to try resetting your branch with master

Gauravtalreja1 (changes requested):

.github/workflows/merge_to_master.yml
swadeley: oh dear

omkarkhatavkar (approved):

.github/workflows/merge_to_master.yml
swadeley: Was caused by a failed rebase to master. I deleted and recreated the branch. Thank you.

pytest_fixtures/core/sys.py
```diff
@@ -10,7 +10,7 @@
 def foreman_service_teardown(target_sat):
     """stop and restart of foreman service"""
     yield target_sat
-    target_sat.execute('satellite-maintain service start --only=foreman')
+    target_sat.execute('satellite-maintain service start')
```
Gauravtalreja1, line 13: you might want to update this step here as well, https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/destructive/test_contenthost.py#L74

It would be best to remove this fixture entirely and add the `request.addfinalizer` call in the test itself for teardown, after replacing this last usage of this fixture with target_sat here, https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/destructive/test_foreman_service.py#L23

mshriver: I think having the fixture is fine, so that its easier to use and we're not repeating the call and lambda definition wherever its needed.

Also allows for easier growth of the service teardown cleanup that can be applied broadly. If we want to add a service status assert here too? easy - without updating every use.

mshriver (approved):

swadeley:
Thank you.

PR #9792, opened 2022-07-19
Expand Search tests to more components.

jarovo:
This is gonna make tests less flaky (random behavior) because it makes the test going trough all the defined components rather than just randomly pick one.

This increases running time quite significantly

### DIFF
--- a/tests/foreman/ui/test_bookmarks.py
+++ b/tests/foreman/ui/test_bookmarks.py
@@ -99,7 +99,7 @@ def test_positive_end_to_end(session, ui_entity):
 
 
 @pytest.mark.tier2
-def test_positive_create_bookmark_public(session, random_entity, default_viewer_role, test_name):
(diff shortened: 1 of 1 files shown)


jaryn:
trigger: test-robottelo
pytest: tests/foreman/ui/test_bookmarks.py -n 8

jaryn:
> @jaryn 50 failures, any that you want to address here in this PR? Any BZ's coming out of those failures?

I beg you pardon Mike. I think the title of this may have misled you. This PR is actually making nonparametrized tests not failing randomly as they used to be. This should bring more stable results (although it might be stable failure), making the test debugging more easy. It is improving the observability as one can see the changes regarding each test prametrization. The other plus side is it will certainly be much better for the TFA learning process.

I know management may not like see those failures increased but let me say this is the true state of the things regarding Search component tests.

Thanks for understanding my intentions.

latran (approved):
Ack, but one question.  What's the reasoning for removing the cherry pick label?

PR #9797, opened 2022-07-19
Add skip marker for failed test cases in cv

latran:
There's a bunch of failed test cases due to[ bz-2086957](https://bugzilla.redhat.com/show_bug.cgi?id=2086957).  From discussion with Brad, there's no prioritization on fixing the bz.  I'm adding all these skip markers for awareness.  Let me know what opinions you have on this is since there's a fair number of failures here.


JacobCallahan (approved):
Since these only apply to legacy cv page, these tests may need to be refactored to point to the new page or removed. Skipping is fine for now

pondrejk (approved):

PR #9807, opened 2022-07-20
Fixing the lce creation that requires "prior" as a param and adding more skip markers

latran:
There's a few test cases that failed when calling `repo.setup_content_view(module_org.id)` because `self.satellite.cli_factory.make_lifecycle_environment({'organization-id': org_id})` requires `prior` parameter.  That has been fixed now.

I've also included the skip markers for a few test cases that I passed over because of this issue.

Lastly, it seems only `ui/test_contentview.py` uses `setup_content_view()` being used so this is pretty isolated.

### DIFF
--- a/robottelo/host_helpers/cli_factory.py
+++ b/robottelo/host_helpers/cli_factory.py
@@ -222,6 +222,7 @@ ENTITY_FIELDS = {
     },
     'lifecycle_environment': {
         '_entity_cls': 'LifecycleEnvironment',
+        'prior': 'Library',
         'name': gen_alphanumeric,
     },
     'tailoringfile': {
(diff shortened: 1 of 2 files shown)


Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/ui/test_contentview.py::test_positive_delete_cv_promoted_to_multi_env

Griffin-Sullivan:
Looks like that test is the only one that uses `setup_content_view` and you aren't skipping due to the BZ you added here

latran:
@Griffin-Sullivan good catch.  I added the skip marker to that as well now.

shweta83 (approved):

mshriver (approved):

PR #9820, opened 2022-07-22
Remove closed skip Bug 1735540 - Virt-who-config for kubevirt does not support in API and hammer CLI

yanpliu:
Hey, 
Bug 1735540 - Virt-who-config for kubevirt does not support in API and hammer CLI is closed on Satellite6.11. so remove the skip  case step

### DIFF
--- a/tests/foreman/virtwho/cli/test_kubevirt.py
+++ b/tests/foreman/virtwho/cli/test_kubevirt.py
@@ -48,7 +48,6 @@ def virtwho_config(form_data, target_sat):
     return target_sat.cli.VirtWhoConfig.create(form_data)['general-information']
 
 
-@pytest.mark.skip_if_open('BZ:1735540')
 class TestVirtWhoConfigforKubevirt:
     @pytest.mark.tier2
     def test_positive_deploy_configure_by_id(


shweta83 (approved):
Ack

yanpliu:
> ACK, based on BZ, it seems that it existed on older versions, so including the NoCherrypick label, @yanpliu please feel free to correct

@Gauravtalreja1 I want to know how to dell with the bug which status is closed. For fixed version such as Satellite6.11 and Satellite6.12 branch ,how to resolve it ? For which is existed older versions, how to resolve it? As you known ,the automation throw error "INTERNALERROR> requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://bugzilla.redhat.com/rest/bug?id=1735540&include_fields=id%2Csummary%2Cstatus%2Cresolution%2Ccf_last_closed%2Clast_change_time%2Ccreation_time%2Cflags%2Ckeywords%2Cdupe_of%2Ctarget_milestone%2Ccf_clone_of%2Cclone_ids%2Cdepends_on"

ColeHiggins2 (approved):

mshriver:
@yanpliu That is handled via the branches in robottelo used to test the respective version of satellite.

The master branch right now is used for 6.11 and 6.12 (6.11 branch coming soon). Removing this skip marker will start to execute the test against 6.11 and 6.12 in pipelines. The mark will still be in the 6.10 branch, and the test will continue to skip when run against 6.10 z-stream builds. If the fix is backported to 6.10, then the mark needs removed there too.

yanpliu:
> @yanpliu That is handled via the branches in robottelo used to test the respective version of satellite.
> 
> The master branch right now is used for 6.11 and 6.12 (6.11 branch coming soon). Removing this skip marker will start to execute the test against 6.11 and 6.12 in pipelines. The mark will still be in the 6.10 branch, and the test will continue to skip when run against 6.10 z-stream builds. If the fix is backported to 6.10, then the mark needs removed there too.

@mshriver Thank you very much for your explanation, it makes me clearly.

PR #9834, opened 2022-07-26
Round correctly, i.e. to 2 significant digits

lhellebr:
This is what is actually done in Satellite


tests/foreman/ui/test_computeresource_vmware.py
```diff
@@ -17,6 +17,7 @@
 :Upstream: No
 """
 from random import choice
+from math import log10, floor
```
mshriver, line 20: This fails pre-commit, our import hook only allows one import per line.

mshriver (approved):

PR #9845, opened 2022-07-28
Removed deprecated permission from test cases

latran:
The list of permissions does not include `Environment` or `ConfigGroup` anymore.  Removing these so tests can pass.

Test results:
```
pytest tests/foreman/cli/test_contentview.py -k test_positive_sub_host_with_restricted_user_perm_at_custom_loc
=============== 1 passed, 128 deselected, 2 warnings in 609.99s (0:10:09) ==============

pytest tests/foreman/cli/test_contentview.py -k test_positive_sub_host_with_restricted_user_perm_at_default_loc
============= 1 passed, 128 deselected, 2 warnings in 675.06s (0:11:15) ================
```

### DIFF
--- a/tests/foreman/cli/test_contentview.py
+++ b/tests/foreman/cli/test_contentview.py
@@ -2162,7 +2162,6 @@ class TestContentView:
                the following permissions::
 
                 - (Miscellaneous) [my_organizations]
-                - Environment [view_environments]
                 - Organization [view_organizations],
                 - Katello::Subscription: [
                     view_subscriptions,
@@ -2171,7 +2170,6 @@ class TestContentView:
                     import_manifest,
                     delete_manifest]
                 - Katello::ContentView [view_content_views]
-                - ConfigGroup [view_config_groups]
                 - Hostgroup view_hostgroups
                 - Host [view_hosts, create_hosts, edit_hosts]
                 - Location [view_locations]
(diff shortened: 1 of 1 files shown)


ogajduse:
Does this PR relate to Sat 6.11 or 6.12?

jameerpathan111:
> The list of permissions does not include `Environment` or `ConfigGroup` anymore. Removing these so tests can pass.

Does this have anything to do with Puppet not being enabled on Satellite?

jameerpathan111:
> Hi @jameerpathan111 , these tests do not use Puppet or have anything to do with puppet.

So I checked with @vsedmik and these permissions are available on puppet enabled satellite.
We can remove them as we are running this test on a puppet-disabled satellite.

ogajduse (approved):
ACK based on @jameerpathan111's and @vsedmik's investigation.
Question: Would it make sense to rewrite one or more tests to use the `session_puppet_enabled_sat` fixture instead of `target_sat`?

latran:
It only makes sense if the test is puppet related or if the closed loop has puppet in it or it has significant data related to puppet.  For anything that was split out like this one, I wouldn't be too concern with covering it for puppet.  I'll let @vsedmik make the decision on that.

vsedmik (approved):
I agree with @latran - these tests do not relate to puppet anyhow and even do not use the puppet-related permissions for their functionality. So +1 to remove the permissions and have the tests run against the default target_sat. Though it is possible, it does not make sense to have them run against the puppet-enabled Satellite only for the sake of keeping the permissions in.

pondrejk (approved):

PR #9849, opened 2022-07-29
Replace instances of dogfood registration with repofile downloads

JacobCallahan:
New dogfood no longer supports the previous registration process. Instead, we will be directly using repofiles for dogfood content.

### DIFF
--- a/robottelo/host_helpers/contenthost_mixins.py
+++ b/robottelo/host_helpers/contenthost_mixins.py
@@ -40,7 +40,27 @@ class VersionedContent:
             'cbrhel': constants.OSCAP_PROFILE[f'cbrhel{self._v_major}'],
         }
 
-    # pull from settings.server.repos???
+    @cached_property
+    def dogfood_repofile(self):
+        from robottelo.config import settings
+
+        if self.__class__.__name__ == 'ContentHost':
+            product = 'tools'
+        else:
+            product = self.__class__.__name__.lower()
+        # this may be also available from self.satellite.version
+        split_ver = self.satellite.version.split('.')
+        if len(split_ver) == 2:
+            split_ver.append('0')
+        sat_ver = '.'.join(split_ver[:3])  # keep only major.minor.patch
+        return (
+            f'{settings.repos.ohsnap_repo_host}/api/releases/'
+            f'{sat_ver}/el{self._v_major}/{product}/repo_file'
+        )
+
+    def download_repofile(self):
+        """Downloads the tools, capsule, or satellite repos on the machine"""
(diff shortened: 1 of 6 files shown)


tests/foreman/destructive/test_katello_certs_check.py
```diff
@@ -117,6 +117,7 @@ def test_positive_install_sat_with_katello_certs(certs_vm_setup):
     result = rhel_vm.subscription_manager_attach_pool([settings.subscription.rhn_poolid])[0]
     for repo in getattr(constants, f"OHSNAP_RHEL{version}_REPOS"):
         rhel_vm.enable_repo(repo, force=True)
+    # What is the purpose of this?
```
JacobCallahan, line 120: @omkarkhatavkar this is a question for you, if you don't mind filling me in on the details.

robottelo/hosts.py
```diff
@@ -1385,6 +1352,12 @@ def get_caller():
             login=login,
         )
 
+    @property
+    def satellite(self):
+        if not self._satellite:
+            return self
+        return self._satellite
+
```
Griffin-Sullivan, line 1360: Where is this being used?

ogajduse: I think that it is not being really used right now. `Satellite` inherits from `Capsule`, if this property would not be present in Satellite class, then with `Satellite().satellite` you would instantiate a new Satellite instance instead of just pointing to `self`.
Is there any other reason, Jake?

JacobCallahan: You're absolutely right! That's precisely what this override is for.

peterdragun (approved):
Thanks, that should help with fixing a lot of tests, including some of mine :+1:

robottelo/host_helpers/contenthost_mixins.py
```diff
@@ -40,7 +40,27 @@ def OSCAP(self):
...
+    @cached_property
+    def dogfood_repofile(self):
+        from robottelo.config import settings
+
+        if self.__class__.__name__ == 'ContentHost':
+            product = 'tools'
+        else:
+            product = self.__class__.__name__.lower()
+        # this may be also available from self.satellite.version
+        split_ver = str(settings.server.version.release).split('.')
```
mshriver, line 52: Seems worth trying to use `self.satellite.version` here instead of reading settings.

robottelo/hosts.py
mshriver: Please add docblock explaining this.

mshriver (approved):

PR #9851, opened 2022-07-29
Update the Sat version fallback to 6.12

JacobCallahan:
After branching, we need to make sure we're falling back to the correct version.

### DIFF
--- a/robottelo/constants/__init__.py
+++ b/robottelo/constants/__init__.py
@@ -15,8 +15,8 @@ class Colored(Box):
 
 
 # This should be updated after each version branch
-SATELLITE_VERSION = "6.11"
-SATELLITE_OS_VERSION = "7"
+SATELLITE_VERSION = "6.12"
+SATELLITE_OS_VERSION = "8"
 
 LOCALES = (
     'ca',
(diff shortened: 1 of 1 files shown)


latran (approved):

robottelo/constants/__init__.py
```diff
@@ -15,7 +15,7 @@ class Colored(Box):
 
 
 # This should be updated after each version branch
-SATELLITE_VERSION = "6.11"
+SATELLITE_VERSION = "6.12"
 SATELLITE_OS_VERSION = "7"
```
latran, line 19: This is supposed to be 8 in 6.12 right?

latran (changes requested):

robottelo/constants/__init__.py
JacobCallahan: oohhhh good catch!

latran (approved):

ogajduse (commented):
@JacobCallahan Do you want to change `constants.REPOS` in this PR? I see Satellite version strings there that are good candidates to be updated too.

PR #9858, opened 2022-08-01
Constants cleanup

JacobCallahan:
Most of these constants are no longer being used.
I also updated the ostree fedora constants since those no longer exist.

I additionally included changes revolving around removing the use of constants in favor of just the values themselves.
In these instances, the constant was significantly longer than its value.

### DIFF
--- a/tests/upgrades/test_yum_plugins.py
+++ b/tests/upgrades/test_yum_plugins.py
@@ -28,7 +28,6 @@ from wait_for import wait_for
 from robottelo.api.utils import attach_custom_product_subscription
 from robottelo.api.utils import call_entity_method_with_timeout
 from robottelo.config import settings
-from robottelo.constants import DISTRO_RHEL7
 from robottelo.constants import REPOS
 from robottelo.logging import logger
 from robottelo.upgrade_utility import install_or_update_package
@@ -37,7 +36,6 @@ from robottelo.upgrade_utility import run_goferd
 
 
 DOCKER_VM = settings.upgrade.docker_vm
-CLIENT_OS = DISTRO_RHEL7
 
 
 class TestScenarioYumPluginsCount:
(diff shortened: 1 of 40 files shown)


tests/upgrades/test_yum_plugins.py
```diff
@@ -78,7 +76,7 @@ def _check_yum_plugins_count(self, client_container_id):
     def _create_custom_tools_repos(self, product):
         """Create custom tools repo and sync it"""
 
-        tools_repo_url = settings.repos.sattools_repo[DISTRO_RHEL7]
+        tools_repo_url = settings.repos.sattools_repo['rhel7']
         if None in [tools_repo_url]:
             raise ValueError(f'The Tools Repo URL {self.client_os} is not provided!')
```
shweta83, line 81:
```suggestion
            raise ValueError(f'The Tools Repo URL rhel7 is not provided!')
```

JacobCallahan: That could very well be the right call. However, I cannot determine the origin of `self.client_os`. @ColeHiggins2 do you know?

latran (approved):

tests/upgrades/test_yum_plugins.py
mshriver: yeah .... `self` is the test class, and I don't see this getting set anywhere. I think I'd go with @shweta83 suggestion here since `rhel7` is hardcoded in the key lookup relevant to this statement.

mshriver (approved):

tests/upgrades/test_yum_plugins.py
mshriver: Since this line isn't actually touched by the PR, I'm going to merge as-is. The test owner can address (or @shweta83 you can open a one-liner!) this `client_os` reference.

PR #9864, opened 2022-08-02
Fixed file downloading by avoiding SFTP

jyejare:
Influenced by https://github.com/SatelliteQE/robottelo/pull/9782#discussion_r929434758 in PR #9782 and forgot that SFTP get wont work with HTTP get.

Adding a new `download_file` from the existing code itself without brainstorming much on its improvement  https://github.com/SatelliteQE/robottelo/pull/9782/commits/a6d8f55a11e3f88d4f4c3a9e412905b9bdc446b6#diff-825c5b0e0eb8cd7eeda6528dcbe21f2de66967c2ca7e60ab281cc56d46e9aa45L96-L143 , but that works as expected !

Some tests are failing in CI so this is bit urgent !!!!!!!!

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -1,3 +1,4 @@
+import os
 import re
 import time
 from contextlib import contextmanager
@@ -33,12 +34,14 @@ from robottelo.constants import CUSTOM_PUPPET_MODULE_REPOS_PATH
 from robottelo.constants import CUSTOM_PUPPET_MODULE_REPOS_VERSION
 from robottelo.constants import HAMMER_CONFIG
 from robottelo.constants import SATELLITE_VERSION
+from robottelo.errors import DownloadFileError
 from robottelo.helpers import InstallerCommand
 from robottelo.host_helpers import CapsuleMixins
 from robottelo.host_helpers import ContentHostMixins
 from robottelo.host_helpers import SatelliteMixins
 from robottelo.logging import logger
 from robottelo.utils import validate_ssh_pub_key
+
 
 POWER_OPERATIONS = {
     VmState.RUNNING: 'running',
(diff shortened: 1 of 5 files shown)


jyejare:
trigger: test-robottelo
pytest: tests/foreman/api/test_rhcloud_inventory.py

adarshdubey-star:
Tests failing on PRT

jyejare:
trigger: test-robottelo
pytest: tests/foreman/api/test_rhcloud_inventory.py

robottelo/hosts.py
```diff
@@ -250,6 +253,32 @@ def power_control(self, state=VmState.RUNNING, ensure=True):
...
+            _, file_name = os.path.split(file_url)
+
+        if not local_path:
+            local_path = '/tmp/'
+
+        # download on server
+        result = self.execute(f'wget -O {local_path}{file_name} {file_url}')
+        if result.status != 0:
+            raise DownloadFileError(f'Unable to download {file_name}')
+        return [f'{local_path}{file_name}', file_name]
```
JacobCallahan, line 280: If we use `PurePath`s here we can remove the `os` import and smooth some things out.
```suggestion
        file_name = PurePath(file_name or file_url).name
        local_path = PurePath(local_path or '/tmp') / file_name

        # download on server
        result = self.execute(f'wget -O {local_path.absolute()} {file_url}')
        if result.status != 0:
            raise DownloadFileError(f'Unable to download {file_name}: {result.stderr}')
        return local_path, file_name
```
This would mean the return for this method also changes to be a tuple consisting of the local_path as a PurePath object and the original file_name

JacobCallahan (changes requested):

robottelo/hosts.py
jyejare: @JacobCallahan Shouldnt be a blocker comment but since this is failing tests I didnt put much effort in improving code as I said in description of this RP.

Will look into it tomorrow !

adarshdubey-star (approved):

jyejare:
trigger: test-robottelo
pytest: tests/foreman/api/test_rhcloud_inventory.py

robottelo/hosts.py
jyejare: @JacobCallahan Updated with some fixes from suggestion script and locally lo0ks good ! Also running PRT lets see how it goes !

JacobCallahan (approved):

jyejare:
@adarshdubey-star @JacobCallahan The 1 test failure isn't related to this change, so we should be good to get this in ?

robottelo/hosts.py
```diff
@@ -250,6 +252,29 @@ def power_control(self, state=VmState.RUNNING, ensure=True):
...
+        :param str file_url: The complete server file path from where the
+            file will be downloaded.
+        :param str local_path: Name of directory where file will be saved. If not
+            provided file will be saved in /tmp/ directory.
+        :param str file_name: New name of the Downloaded file else its given from file_url
+
+        :returns: Returns list containing complete file path and name of downloaded file.
+        """
+        file_name = PurePath(file_name or file_url).name
+        local_path = PurePath(local_path or '/tmp') / file_name
```
mshriver, line 270: This should be using `settings.robottelo.tmp_dir`

https://github.com/SatelliteQE/robottelo/blob/master/conf/robottelo.yaml.template#L8

mshriver (changes requested):
We have a tmp dir setting, shouldn't use `/tmp` directly.

robottelo/hosts.py
jyejare: Certainly, we can go for it, but I see many places in robottelo are still using `/tmp` directly or `/var/tmp` as a hardcoded one! 

So can we handle all those places in a separate PR! For this PR moving from `/tmp` to `/var/tmp` might or not cause the problem and for testing again it might take some time !

Again this PR is not improvement PR but a quick fix PR that @adarshdubey-star brought up !

mshriver: I can accept that.

PR #9920, opened 2022-08-05
rex pull-provider tests

pondrejk:
Initial tests for pull provider remote execution. This PR:

- adds a capsule fixture, as the pull-provider feature is enabled via installer per capsule
- adds a helper to register host via global registration, needed to enable pp on client. It's actually the recommended way to register hosts and can be used without additional helpers for many other tasks (e.g. for key transfer or insights installation)
- adds two rex tests for enabling pp client either via conversion script and straight trough global registration. The latter expects changes that are yet to come in sn5, but passed locally on patched sn4

Still, the tests might get flaky due to ocassional job hanging, bz#2115229 is open to keep track on that

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -5,6 +5,7 @@ from functools import cached_property
 from pathlib import Path
 from pathlib import PurePath
 from tempfile import NamedTemporaryFile
+from urllib.parse import urlencode
 from urllib.parse import urljoin
 from urllib.parse import urlunsplit
 
@@ -481,10 +482,10 @@ class ContentHost(Host, ContentHostMixins):
 
     def register(
         self,
-        target=settings.server.hostname,
-        org='Default_Organization',
-        loc='Default_Location',
-        activation_keys=None,
+        target,
+        org,
+        loc,
+        activation_keys,
         setup_insights=False,
         setup_remote_execution=True,
         setup_remote_execution_pull=False,
@@ -505,8 +506,8 @@ class ContentHost(Host, ContentHostMixins):
         using a global registration template.
 
         :param target: Satellite or Capusle hostname to register to, required.
-        :param org: Organization name to register content host for, required.
-        :param loc: Location name to register content host for, required.
+        :param org: Organization to register content host for, required.
+        :param loc: Location to register content host for, required.
         :param activation_keys: Activation key name to register content host
             with, required.
         :param setup_insights: Install and register Insights client, requires OS repo.
@@ -522,73 +523,68 @@ class ContentHost(Host, ContentHostMixins):
         :param ignore_subman_errors: Ignore subscription manager errors.
         :param force: Register the content host even if it's already registered.
         :param insecure: Don't verify server authenticity.
-        :param username: A user name to register the content host with.
-        :param password: The user password.
+        :param username: Satellite admin username
+        :param password: Satellite admin password
         :return: SSHCommandResult instance filled with the result of the
             registration.
         """
-        cmd = 'curl -sS '
-        if username and password:
-            cmd += f'-u {username}:{password} '
-        if insecure:
-            cmd += '--insecure '
-        if target:
-            cmd += f"'https://{target}:9090/register?"
-        else:
-            raise ContentHostError('Target Satellite/Capsule not specified')
-        if activation_keys:
-            cmd += f'activation_keys={activation_keys}'
-        else:
-            raise ContentHostError('Activation key not specified')
-        if org:
-            selected_org = entities.Organization().search(query={'search': f'name="{org}"'})[0]
-            cmd += f'&organization_id={selected_org.id}'
-        else:
-            raise ContentHostError('Organization not specified')
-        if loc:
-            selected_loc = entities.Location().search(query={'search': f'name="{loc}"'})[0]
-            cmd += f'&location_id={selected_loc.id}'
-        else:
-            raise ContentHostError('Location not specified')
-        if setup_insights:
+        insights = (
             # requires OS repo enabled for host
-            cmd += '&setup_insights=true'
-        else:
-            cmd += '&setup_insights=false'
-        if setup_remote_execution:
-            cmd += '&setup_remote_execution=true'
-        else:
-            cmd += '&setup_remote_execution=false'
-        if update_packages:
-            cmd += '&update_packages=true'
-        else:
-            cmd += '&update_packages=false'
-        if lifecycle_environment:
-            selected_lce = entities.LifecycleEnvironment().search(
-                query={'search': f'name="{lifecycle_environment}"'}
-            )[0]
-            cmd += f'&lifecycle_environment_id={selected_lce.id}'
-        if operating_system:
-            selected_os = entities.OperatingSystem().search(
-                query={'search': f'name="{operating_system}"'}
-            )[0]
-            cmd += f'&operating_system_id={selected_os.id}'
-        if setup_remote_execution_pull:
+            f'&setup_insights={str(setup_insights).lower()}'
+            if setup_insights is not None
+            else ''
+        )
+        rex = (
+            f'&setup_remote_execution={str(setup_remote_execution).lower()}'
+            if setup_remote_execution is not None
+            else ''
+        )
+        rex_pull = (
             # requires Satellite Client repo enabled for host
-            cmd += '&setup_remote_execution_pull=true'
-        if packages:
-            cmd += f'&packages={"+".join(packages)}'
-        if repo_gpg_key_url:
-            cmd += f'&repo_gpg_key_url={repo_gpg_key_url.replace("/", "%2F").replace(":", "%3A")}'
-        if repo:
-            cmd += f'&repo={repo.replace("/", "%2F").replace(":", "%3A")}'
-        if remote_execution_interface:
-            cmd += f'&remote_execution_interface={remote_execution_interface}'
-        if ignore_subman_errors:
-            cmd += '&ignore_subman_errors=true'
-        if force:
-            cmd += '&force=true'
-        cmd += "' | bash"
+            f'&setup_remote_execution_pull={str(setup_remote_execution_pull).lower()}'
+            if setup_remote_execution_pull is not None
+            else ''
+        )
+        lce = (
+            f'&lifecycle_environment_id={lifecycle_environment.id}'
+            if lifecycle_environment is not None
+            else ''
+        )
+        os = f'&operating_system_id={operating_system.id}' if operating_system is not None else ''
+        pkgs = f'&packages={"+".join(packages)}' if packages is not None else ''
+        rex_iface = (
+            f'&remote_execution_interface={remote_execution_interface}'
+            if remote_execution_interface is not None
+            else ''
+        )
+        rp = f'&{urlencode({"repo": repo })}' if repo is not None else ''
+        gpg = (
+            f'&{urlencode({"repo_gpg_key_url": repo_gpg_key_url })}'
+            if repo_gpg_key_url is not None
+            else ''
+        )
+        cmd = (
+            'curl -sS '
+            f'-u {username}:{password} '
+            f'{"--insecure " if insecure else ""}'
+            f"'https://{target.hostname}:9090/register?"
+            f'activation_keys={activation_keys}'
+            f'&organization_id={org.id}'
+            f'&location_id={loc.id}'
+            f'{insights}'
+            f'{rex}'
+            f'&update_packages={"true" if update_packages else "false"}'
+            f'{rex_pull}'
+            f'{rex_iface}'
+            f'{lce}'
+            f'{os}'
+            f'{pkgs}'
+            f'{"&ignore_subman_errors=true" if ignore_subman_errors else ""}'
+            f'{"&force=true" if force else ""}'
+            f'{rp}'
+            f'{gpg}'
+            "' | bash"
+        )
         return self.execute(cmd)
 
     def register_contenthost(
@@ -1386,11 +1382,16 @@ class Capsule(ContentHost, CapsuleMixins):
             )
 
     def enable_mqtt(self):
+        installer_opts = {
+            'foreman-proxy-templates': 'true',
+            'foreman-proxy-registration': 'true',
+            'foreman-proxy-plugin-remote-execution-script-mode': 'pull-mqtt',
+        }
+        enable_mqtt_command = InstallerCommand(
+            installer_opts=installer_opts,
+        )
         result = self.execute(
-            'satellite-installer \
-                    --foreman-proxy-templates=true \
-                    --foreman-proxy-registration=true \
-                    --foreman-proxy-plugin-remote-execution-script-mode=pull-mqtt',
+            enable_mqtt_command.get_command(),
             timeout='20m',
         )
         if result.status != 0:
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -479,6 +479,118 @@ class ContentHost(Host, ContentHostMixins):
         if result.status != 0:
             raise ContentHostError('Failed to install the cockpit')
 
+    def register(
+        self,
+        target=settings.server.hostname,
+        org='Default_Organization',
+        loc='Default_Location',
+        activation_keys=None,
+        setup_insights=False,
+        setup_remote_execution=True,
+        setup_remote_execution_pull=False,
+        lifecycle_environment=None,
+        operating_system=None,
+        packages=None,
+        repo=None,
+        repo_gpg_key_url=None,
+        remote_execution_interface=None,
+        update_packages=False,
+        ignore_subman_errors=True,
+        force=True,
+        insecure=True,
+        username=settings.server.admin_username,
+        password=settings.server.admin_password,
+    ):
+        """Registers content host to the Satellite or Capsule server
+        using a global registration template.
+
+        :param target: Satellite or Capusle hostname to register to, required.
+        :param org: Organization name to register content host for, required.
+        :param loc: Location name to register content host for, required.
+        :param activation_keys: Activation key name to register content host
+            with, required.
+        :param setup_insights: Install and register Insights client, requires OS repo.
+        :param setup_remote_execution: Copy remote execution SSH key.
+        :param setup_remote_execution_pull: Deploy pull provider client on host
+        :param lifecycle_environment: Lifecycle environment.
+        :param operating_system: Operating system.
+        :param packages: A list of packages to install on the host when registered.
+        :param repo: Repository to be added before the registration is performed, supply url.
+        :param repo_gpg_key_url: Public key to verify the package signatures, supply url.
+        :param remote_execution_interface: Identifier of the host interface for remote execution.
+        :param update_packages: Update all packages on the host.
+        :param ignore_subman_errors: Ignore subscription manager errors.
+        :param force: Register the content host even if it's already registered.
+        :param insecure: Don't verify server authenticity.
+        :param username: A user name to register the content host with.
+        :param password: The user password.
+        :return: SSHCommandResult instance filled with the result of the
+            registration.
+        """
+        cmd = 'curl -sS '
+        if username and password:
+            cmd += f'-u {username}:{password} '
+        if insecure:
+            cmd += '--insecure '
+        if target:
+            cmd += f"'https://{target}:9090/register?"
+        else:
+            raise ContentHostError('Target Satellite/Capsule not specified')
+        if activation_keys:
+            cmd += f'activation_keys={activation_keys}'
+        else:
+            raise ContentHostError('Activation key not specified')
+        if org:
+            selected_org = entities.Organization().search(query={'search': f'name="{org}"'})[0]
+            cmd += f'&organization_id={selected_org.id}'
+        else:
+            raise ContentHostError('Organization not specified')
+        if loc:
+            selected_loc = entities.Location().search(query={'search': f'name="{loc}"'})[0]
+            cmd += f'&location_id={selected_loc.id}'
+        else:
+            raise ContentHostError('Location not specified')
+        if setup_insights:
+            # requires OS repo enabled for host
+            cmd += '&setup_insights=true'
+        else:
+            cmd += '&setup_insights=false'
+        if setup_remote_execution:
+            cmd += '&setup_remote_execution=true'
+        else:
+            cmd += '&setup_remote_execution=false'
+        if update_packages:
+            cmd += '&update_packages=true'
+        else:
+            cmd += '&update_packages=false'
+        if lifecycle_environment:
+            selected_lce = entities.LifecycleEnvironment().search(
+                query={'search': f'name="{lifecycle_environment}"'}
+            )[0]
+            cmd += f'&lifecycle_environment_id={selected_lce.id}'
+        if operating_system:
+            selected_os = entities.OperatingSystem().search(
+                query={'search': f'name="{operating_system}"'}
+            )[0]
+            cmd += f'&operating_system_id={selected_os.id}'
+        if setup_remote_execution_pull:
+            # requires Satellite Client repo enabled for host
+            cmd += '&setup_remote_execution_pull=true'
+        if packages:
+            cmd += f'&packages={"+".join(packages)}'
+        if repo_gpg_key_url:
+            cmd += f'&repo_gpg_key_url={repo_gpg_key_url.replace("/", "%2F").replace(":", "%3A")}'
+        if repo:
+            cmd += f'&repo={repo.replace("/", "%2F").replace(":", "%3A")}'
+        if remote_execution_interface:
+            cmd += f'&remote_execution_interface={remote_execution_interface}'
+        if ignore_subman_errors:
+            cmd += '&ignore_subman_errors=true'
+        if force:
+            cmd += '&force=true'
+        cmd += "' | bash"
+        return self.execute(cmd)
+
     def register_contenthost(
         self,
         org='Default_Organization',
@@ -1272,6 +1384,17 @@ class Capsule(ContentHost, CapsuleMixins):
             raise CapsuleHostError(
                 f'A core service is not running at capsule host\n{result.stdout}'
             )
+
+    def enable_mqtt(self):
+        result = self.execute(
+            'satellite-installer \
+                    --foreman-proxy-templates=true \
+                    --foreman-proxy-registration=true \
+                    --foreman-proxy-plugin-remote-execution-script-mode=pull-mqtt',
+            timeout='20m',
+        )
+        if result.status != 0:
+            raise SatelliteHostError(f'Failed to enable pull provider: {result.stdout}')
 
 
 class Satellite(Capsule, SatelliteMixins):
--- a/tests/foreman/cli/test_remoteexecution.py
+++ b/tests/foreman/cli/test_remoteexecution.py
@@ -1053,7 +1053,7 @@ class TestPullProviderRex:
         )
         assert result.status == 0, 'Failed to download certificate'
         client_repo = (
-            f'{settings.repos["DOGFOOD_REPO_HOST"]}/pulp/content/'
+            f'{settings.repos["DOGFOOD_REPO_HOST"].replace("http", "https")}/pulp/content/'
             'Satellite_Engineering/QA/Satellite_Client/custom/Satellite_Client_Composes/'
             f'Satellite_Client_RHEL{rhel_contenthost.os_version.major}_x86_64/'
         )
@@ -1063,9 +1063,9 @@ class TestPullProviderRex:
 
         # register host with rex, enable client repo, install katello-agent
         result = rhel_contenthost.register(
-            module_capsule_configured_mqtt.hostname,
-            module_org.name,
-            smart_proxy_location.name,
+            module_capsule_configured_mqtt,
+            module_org,
+            smart_proxy_location,
             module_ak_with_cv.name,
             packages=['katello-agent'],
             repo=client_repo,
@@ -1180,7 +1180,7 @@ class TestPullProviderRex:
         )
         assert result.status == 0, 'Failed to download certificate'
         client_repo = (
-            f'{settings.repos["DOGFOOD_REPO_HOST"]}/pulp/content/'
+            f'{settings.repos["DOGFOOD_REPO_HOST"].replace("http", "https")}/pulp/content/'
             'Satellite_Engineering/QA/Satellite_Client/custom/Satellite_Client_Composes/'
             f'Satellite_Client_RHEL{rhel_contenthost.os_version.major}_x86_64/'
         )
@@ -1190,15 +1190,14 @@ class TestPullProviderRex:
 
         # register host with pull provider rex (SAT-1677)
         result = rhel_contenthost.register(
-            module_capsule_configured_mqtt.hostname,
-            module_org.name,
-            smart_proxy_location.name,
+            module_capsule_configured_mqtt,
+            module_org,
+            smart_proxy_location,
             module_ak_with_cv.name,
             setup_remote_execution_pull=True,
             repo=client_repo,
         )
         assert result.status == 0, f'Failed to register host: {result.stderr}'
-
         # check mqtt client is running
         result = rhel_contenthost.execute('yggdrasil status')
         assert result.status == 0, f'Failed to start yggdrasil on client: {result.stderr}'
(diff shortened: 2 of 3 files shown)


pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k pull

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k test_positive_run_job_on_host_converted_to_pull_provider

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k test_positive_run_job_on_host_converted_to_pull_provider

latran (commented):
Seems there's still failures for PRT run.  Can you recheck?

pondrejk:
Well, snap5 didn't bring all the bits needed for `test_positive_run_job_on_host_registered_to_pull_provider` to run successfully. As for `test_positive_run_job_on_host_converted_to_pull_provider`, the results go like this:

```
tests/foreman/cli/test_remoteexecution.py FF..E                          [100%]
```

The error is a rhel9 host not returned from broker, two failures are from rhel7 and rhel7_fips hosts, where the very last job in the test fails, might be a case of bz#2115229, but I will investigate further.

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k test_positive_run_job_on_host_converted_to_pull_provider

robottelo/hosts.py
```diff
@@ -1551,6 +1663,17 @@ def install_cockpit(self):
...
+    def enable_mqtt(self):
+        result = self.execute(
+            'satellite-installer \
+                    --foreman-proxy-templates=true \
+                    --foreman-proxy-registration=true \
+                    --foreman-proxy-plugin-remote-execution-script-mode=pull-mqtt',
+            timeout='20m',
+        )
+        if result.status != 0:
+            raise SatelliteHostError(f'Failed to enable pull provider on capsule: {result.stdout}')
```
Gauravtalreja1, line 1675:
```suggestion
            raise SatelliteHostError(f'Failed to enable pull provider : {result.stdout}')
```

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -1019,3 +1020,205 @@ def test_positive_rex_against_infra_hosts(
...
+            f'curl -o /etc/pki/ca-trust/source/anchors/satellite-sat-engineering-ca.crt \
+                    {settings.repos["DOGFOOD_REPO_HOST"]}/pub/katello-server-ca.crt; \
+                    update-ca-trust'
+        )
+        assert result.status == 0, 'Failed to download certificate'
+        client_repo = (
+            'https://satellite.sat.engineering.redhat.com/pulp/content/'
+            'Satellite_Engineering/QA/Satellite_Client/custom/Satellite_Client_Composes/'
+            f'Satellite_Client_RHEL{rhel_contenthost.os_version.major}_x86_64/'
+        )
```
Gauravtalreja1, line 1059:
```suggestion
        client_repo = (
            f'{settings.repos["DOGFOOD_REPO_HOST"]}/pulp/content/'
            'Satellite_Engineering/QA/Satellite_Client/custom/Satellite_Client_Composes/'
            f'Satellite_Client_RHEL{rhel_contenthost.os_version.major}_x86_64/'
        )
```

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -1019,3 +1020,205 @@ def test_positive_rex_against_infra_hosts(
...
```
Gauravtalreja1, line 1186:
```suggestion
        client_repo = (
            f'{settings.repos["DOGFOOD_REPO_HOST"]}/pulp/content/'
            'Satellite_Engineering/QA/Satellite_Client/custom/Satellite_Client_Composes/'
            f'Satellite_Client_RHEL{rhel_contenthost.os_version.major}_x86_64/'
        )
```

Gauravtalreja1 (changes requested):

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k test_positive_run_job_on_host_converted_to_pull_provider

pondrejk:
forom prt:
```
py.test tests/foreman/cli/test_remoteexecution.py -k test_positive_run_job_on_host_converted_to_pull_provider --junit-xml=sat-results.xml -o junit_suite_name=sat-result
 platform linux -- Python 3.8.12, pytest-7.1.2, pluggy-1.0.0
 rootdir: /opt/app-root/src/robottelo, configfile: pyproject.toml
 plugins: forked-1.4.0, ibutsu-2.2.4, mock-3.8.2, reportportal-5.1.2, services-2.2.1, xdist-2.5.0
 collected 37 items / 32 deselected / 5 selected
 
 tests/foreman/cli/test_remoteexecution.py ....E                          [100%]
```
1 error is the missing rhel9 checkout

robottelo/hosts.py
```diff
@@ -479,6 +479,118 @@ def install_cockpit(self):
         if result.status != 0:
             raise ContentHostError('Failed to install the cockpit')
 
+    def register(
```
mshriver, line 482: How is this different from the `register_contenthost` function that's defined immediately belowm that uses subscription manager?

robottelo/hosts.py
```diff
@@ -479,6 +479,118 @@ def install_cockpit(self):
...
+        self,
+        target=settings.server.hostname,
```
mshriver, line 484: This should be a required parameter, and should not default to the settings hostname.

Satellite/Capsule instance should be passed, and the `target.hostname` attribute used where necessary within the function.

robottelo/hosts.py
```diff
@@ -479,6 +479,118 @@ def install_cockpit(self):
...
+        org='Default_Organization',
```
mshriver, line 485: We have constants for both of these, they should be used here.

robottelo/hosts.py
```diff
@@ -479,6 +479,118 @@ def install_cockpit(self):
...
+        loc='Default_Location',
+        activation_keys=None,
+        setup_insights=False,
+        setup_remote_execution=True,
```
mshriver, line 489: There is already a `setup_rex` function on the ContentHost class, why have this function also control REX setup?

robottelo/hosts.py
```diff
@@ -479,6 +479,118 @@ def install_cockpit(self):
...
+        operating_system=None,
+        packages=None,
+        repo=None,
+        repo_gpg_key_url=None,
+        remote_execution_interface=None,
+        update_packages=False,
+        ignore_subman_errors=True,
+        force=True,
+        insecure=True,
+        username=settings.server.admin_username,
```
mshriver, line 501: The target satellite/capsule instance attributes should be used for username and password. The target system might not be using the same credentials or user from settings in the scope of the test.

robottelo/hosts.py
```diff
@@ -479,6 +479,118 @@ def install_cockpit(self):
...
+        if target:
+            cmd += f"'https://{target}:9090/register?"
+        else:
+            raise ContentHostError('Target Satellite/Capsule not specified')
+        if activation_keys:
+            cmd += f'activation_keys={activation_keys}'
+        else:
+            raise ContentHostError('Activation key not specified')
+        if org:
+            selected_org = entities.Organization().search(query={'search': f'name="{org}"'})[0]
```
mshriver, line 544: If you're doing this anyway, have the function accept an Org / Location object.

Otherwise the test is using fixtures for org and location objects, passing them by name here, this is then doing another API lookup of that same object just to get its ID.

robottelo/hosts.py
```diff
@@ -479,6 +479,118 @@ def install_cockpit(self):
...
+            cmd += f'&organization_id={selected_org.id}'
+        else:
+            raise ContentHostError('Organization not specified')
```
mshriver, line 547: Is this in case someone overrides the default and passes `None` or `False` for org?  Just remove the default and make it a required parameter.

robottelo/hosts.py
```diff
@@ -479,6 +479,118 @@ def install_cockpit(self):
...
+        :param force: Register the content host even if it's already registered.
+        :param insecure: Don't verify server authenticity.
+        :param username: A user name to register the content host with.
+        :param password: The user password.
+        :return: SSHCommandResult instance filled with the result of the
+            registration.
+        """
+        cmd = 'curl -sS '
+        if username and password:
+            cmd += f'-u {username}:{password} '
```
mshriver, line 532: I think this style of constructing the command is difficult to read and make sense of, and is using `+` to construct the string from f-string parts, instead of using f-string directly.

How about doing whatever parameter validation you need to do ahead of time, and then constructing the string in a more easy to read format.

```
user = target.username
pass = target.password

cmd = (
    f'curl -sS '
    f'{"--insecure " if insecure else ""}'
    f'https://{target.hostname}:{target.port}/register?'
    f'activation_keys={activation_keys}'
    # etc
)
```

robottelo/hosts.py
```diff
@@ -479,6 +479,118 @@ def install_cockpit(self):
...
+        if insecure:
+            cmd += '--insecure '
+        if target:
+            cmd += f"'https://{target}:9090/register?"
+        else:
+            raise ContentHostError('Target Satellite/Capsule not specified')
+        if activation_keys:
```
mshriver, line 539: Why isn't this a required parameter if the function is going to raise an exception with its default of `None`?

robottelo/hosts.py
```diff
@@ -479,6 +479,118 @@ def install_cockpit(self):
...
+        if org:
+            selected_org = entities.Organization().search(query={'search': f'name="{org}"'})[0]
+            cmd += f'&organization_id={selected_org.id}'
+        else:
+            raise ContentHostError('Organization not specified')
+        if loc:
+            selected_loc = entities.Location().search(query={'search': f'name="{loc}"'})[0]
+            cmd += f'&location_id={selected_loc.id}'
+        else:
+            raise ContentHostError('Location not specified')
```
mshriver, line 552: All of these should be required parameters with no default if they're required.

robottelo/hosts.py
```diff
@@ -479,6 +479,118 @@ def install_cockpit(self):
...
+        if setup_insights:
+            # requires OS repo enabled for host
+            cmd += '&setup_insights=true'
```
mshriver, line 555: Since all of these are the same arg, consider using:
`f'&setup_insights={str(setup_insights).lower()}'`

robottelo/hosts.py
```diff
@@ -479,6 +479,118 @@ def install_cockpit(self):
...
+        :return: SSHCommandResult instance filled with the result of the
+            registration.
+        """
+        cmd = 'curl -sS '
+        if username and password:
+            cmd += f'-u {username}:{password} '
+        if insecure:
+            cmd += '--insecure '
+        if target:
+            cmd += f"'https://{target}:9090/register?"
```
mshriver, line 536: You should really consider using `urllib` to construct this URL, instead of manually. You can pass a data structure of the query params and `urlencode` can compose them into a valid query string.

robottelo/hosts.py
```diff
@@ -479,6 +479,118 @@ def install_cockpit(self):
...
+                query={'search': f'name="{operating_system}"'}
+            )[0]
+            cmd += f'&operating_system_id={selected_os.id}'
+        if setup_remote_execution_pull:
+            # requires Satellite Client repo enabled for host
+            cmd += '&setup_remote_execution_pull=true'
+        if packages:
+            cmd += f'&packages={"+".join(packages)}'
+        if repo_gpg_key_url:
+            cmd += f'&repo_gpg_key_url={repo_gpg_key_url.replace("/", "%2F").replace(":", "%3A")}'
```
mshriver, line 582: This *needs* to be using `urlencode`

mshriver, line 582: This *needs* to be using `urlencode`

mshriver, line 582: This *needs* to be using `urlencode`

robottelo/hosts.py
```diff
@@ -1273,6 +1385,17 @@ def capsule_setup(self, sat_host=None, **installer_kwargs):
                 f'A core service is not running at capsule host\n{result.stdout}'
             )
 
+    def enable_mqtt(self):
+        result = self.execute(
+            'satellite-installer \
+                    --foreman-proxy-templates=true \
```
mshriver, line 1391: extra indent here

robottelo/hosts.py
```diff
@@ -1273,6 +1385,17 @@ def capsule_setup(self, sat_host=None, **installer_kwargs):
                 f'A core service is not running at capsule host\n{result.stdout}'
             )
 
+    def enable_mqtt(self):
+        result = self.execute(
+            'satellite-installer \
```
mshriver, line 1390: Its good practice to use the `InstallerCommand` class to construct these.

https://github.com/SatelliteQE/robottelo/blob/master/robottelo/helpers.py#L345

robottelo/hosts.py
pondrejk: This is using global registration api (post 6.10 way of registering hosts), register_contenthost and other functions here mimic the pre-6.10 manual way of registration. Some functions may still have value, but we should actually using globrex wherever possible (out of scope of this PR). 2 reasons for that:
- globrex starts to drift from the legacy method (e.g. it is not pulling keys in an rpm, so automation did hit https://github.com/SatelliteQE/robottelo/pull/9571 but customers didn't)
- new functionality is being added to globrex but not elsewhere. That's why I need to use this method -- `setup_remote_execution_pull` param enables pull client during registration

robottelo/hosts.py
pondrejk: setup_rex is legacy, though I think both have valid use case. You may want to enable rex straight when registering host. Or you already have a registered non-rex host and you just want to enable it afterwards

robottelo/hosts.py
pondrejk: I don't get how it helps, I see sat-installer being called directly more often then not around robottelo. If you could suggest how to rewrite that bit the InstallerCommand way, I'd be grateful

Gauravtalreja1: https://github.com/SatelliteQE/robottelo/blob/master/pytest_fixtures/component/puppet.py#L9-L37

mshriver: @Gauravtalreja1 thanks, that's a great example!

@pondrejk the InstallerCommand class was introduced without converting every use of `satellite-installer` to use the class. We're encouraging people to use it as much as possible though, eventually converting all the uses of it.

I'm not going to call it required for this test, but I'd encourage it if you have a bit of time to make this switch, which should be straightforward in this case.

robottelo/hosts.py
mshriver: Yeah sometimes you have to mix things a bit, and label some variables that only end up getting used once so that the final string construction is readable. Taking a look at the new commit(s) now, thanks for revising!

robottelo/hosts.py
mshriver: @pondrejk ack, gotcha!

I'd like to discuss this further, and with @JacobCallahan make a plan for moving robottelo away from `register_contenthost` pattern.

omkarkhatavkar: I did added support for generating the global registration command in Nailgun, Incase if that worthy to try https://github.com/SatelliteQE/nailgun/pull/795

robottelo/hosts.py
pondrejk: Here I had to revert the previous change, as we need satellite admin username/password here, not the target capsule ssh credentials

robottelo/hosts.py
```diff
@@ -479,6 +480,119 @@ def install_cockpit(self):
...
+        gpg = (
+            f'&{urlencode({"repo_gpg_key_url": repo_gpg_key_url })}'
+            if repo_gpg_key_url is not None
+            else ''
+        )
+        cmd = (
+            'curl -sS '
+            f'-u {username}:{password} '
+            f'{"--insecure " if insecure else ""}'
+            f"'https://{target.hostname}:9090/register?"
```
pondrejk, line 576: Change here over the previous iteration, we need capsule port here, not the ssh port that is in target.port. Capsule port change is afaik not supported, therefore not worrying about the hardcode here

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k pull_provider

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k pull_provider

robottelo/hosts.py
```diff
@@ -479,6 +480,119 @@ def install_cockpit(self):
...
+            else ''
+        )
+        rex_pull = (
+            # requires Satellite Client repo enabled for host
+            f'&setup_remote_execution_pull={str(setup_remote_execution_pull).lower()}'
+            if setup_remote_execution_pull is not None
+            else ''
+        )
+        lce = ''
+        if lifecycle_environment:
```
mshriver, line 549: Like org and location, why accept a string for these entities and then look them up. Take the object as the argument - the calling test function/fixture/etc will have the entity instance that is needed.

Separately, any entity classes that you do use should be referenced through `target.cli` or `target.api`, not through the imported entities module.

robottelo/hosts.py
```diff
@@ -479,6 +480,119 @@ def install_cockpit(self):
...
+                query={'search': f'name="{operating_system}"'}
+            )[0]
+            os = f'&operating_system_id={selected_os.id}'
+        pkgs = f'&packages={"+".join(packages)}' if packages is not None else ''
+        rex_iface = (
+            f'&remote_execution_interface={remote_execution_interface}'
+            if remote_execution_interface is not None
+            else ''
+        )
+        rp = f'&{urlencode({"repo": repo })}' if repo is not None else ''
```
mshriver, line 566: urlencode will do all this formatting for you, it handles the data structure, not just a single field!

```
In [1]: from urllib.parse import urlencode
In [2]: query_params = dict()
In [3]: query_params['update_packages'] = True
In [4]: query_params['organization_id'] = 1
In [5]: query_params['setup_insights'] = True

In [6]: urlencode(query_params)
Out[6]: 'update_packages=True&organization_id=1&setup_insights=True'
```

You just have to deal with whether or not you set a dict key based on if that field should be included or not, but you don't have to try and format every single field individually. Build the dictionary of all the key-value pairs that should be in the query string, and then format that query string with urlencode.

robottelo/hosts.py
mshriver, line 576: As @omkarkhatavkar pointed out, RegistrationCommand entity exists in nailgun, so why are you directly curl-ing the endpoint here?

https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entities.py#L6024

pondrejk: I didn't know about it, but still I'd prefer my version :) Mainly because directly curling the endpoint is what we ask customers to do in the satellite's global registration dialog. Also the entity would need some working on (missing parameters, handling of none states, etc.)

robottelo/hosts.py
pondrejk: I'm not sold on this, in my opinion it would just make the whole thing harder to debug in the future, plus it doesn't reduce the data wrangling. Still I'd need to deal with some sat api quirks myself, e.g.:

```
query_params['packages'] = ['a', 'b', 'c']
urlencode(query_params)
'update_packages=true&organization_id=1&packages=%5B%27a%27%2C+%27b%27%2C+%27c%27%5D'
```
for packages I need `packages=a+b+c`

pondrejk:
from the latest rpr run:
```
 + py.test tests/foreman/cli/test_remoteexecution.py -k pull_provider --junit-xml=sat-results.xml -o junit_suite_name=sat-result
 platform linux -- Python 3.8.12, pytest-7.1.2, pluggy-1.0.0
 rootdir: /opt/app-root/src/robottelo, configfile: pyproject.toml
 plugins: forked-1.4.0, ibutsu-2.2.4, mock-3.8.2, reportportal-5.1.2, services-2.2.1, xdist-2.5.0
 collected 37 items / 27 deselected / 10 selected
 
 tests/foreman/cli/test_remoteexecution.py ..E.E.F....F                   [100%]
```
Errors are machines not handed from broker, failures are both for rhel 9 hosts, one definitely due to bz#2118356, the other one to be determined.

Otherwise I believe I have addressed or at least replied to all comments.

robottelo/hosts.py
mshriver:
```
In [6]: query_params['packages'] = 'a b c'

In [7]: urlencode(query_params)
Out[7]: 'update_packages=True&setup_insights=True&packages=a+b+c'
```

The default quoting method of urlencode is quote_plus, which will substitute `+` for spaces, you just have to provide the value as a space delineated string.

robottelo/hosts.py
mshriver: Ack, understood you're trying to replicate documented patterns for it!

pondrejk: Actually, I don't see how I could use nailgun in this scenario, I'd have to set up a nailgun environment on a host, which doesn't make sense

robottelo/hosts.py
pondrejk: Thanks for pointers, the function is now using InstallerCommand

robottelo/hosts.py
pondrejk: I could modify inputs to the form that urlencode would like, but why? It makes sense to use it for url processing, but for the whole thing I don't see the value. Currently you have at least a notion of what you are creating in that function just by reading the code, you can compare to what you get from satellite, it's just more maintainable. Plus if the default quoting method changes it will create problems in the future

robottelo/hosts.py
pondrejk: updated

robottelo/hosts.py
omkarkhatavkar: you might need the sat object to get that url.

pondrejk:
@Gauravtalreja1 @mshriver @omkarkhatavkar  do you see any obstacles to getting this PR in?

omkarkhatavkar:
@pondrejk Do you have test results?

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k pull

pondrejk:
@Gauravtalreja1 @mshriver @omkarkhatavkar satlab-tower pr 606 got merged, the latest prt run looks like this:

```
py.test tests/foreman/cli/test_remoteexecution.py -k pull --junit-xml=sat-results.xml -o junit_suite_name=sat-result
09:41:55  ============================= test session starts ==============================
09:41:55  platform linux -- Python 3.8.12, pytest-7.1.2, pluggy-1.0.0
09:41:55  shared_function enabled - OFF - scope:  - storage: file
09:41:55  rootdir: /opt/app-root/src/robottelo, configfile: pyproject.toml
09:41:55  plugins: forked-1.4.0, ibutsu-2.2.4, mock-3.8.2, reportportal-5.1.2, services-2.2.1, xdist-2.5.0
09:41:55  collected 41 items / 29 deselected / 12 selected
09:41:55  
16:22:11  tests/foreman/cli/test_remoteexecution.py ....FF......                   [100%]
```

both failures are due to bz#2118356 (failed to install katello-pull-transport-migrate on rhel9 and rhel9_fips). 

I believe there are no blocker left, just note that squash-merge would be best for this PR

robottelo/hosts.py
mshriver: We're getting to some subjective arguments here, but I don't think the readability is any better with the string composition you have here than it would be with creating a dictionary that gets encoded at once.  My argument is essentially boiling down to "we should use standard library functions for their intended purpose". 

The value of the urlencoding on the entire thing is just that - every value will be properly encoded. All the fields will be concatenated, quoted, and encoded. You're not hand-composing `f'&something={something}'` strings over and over.  The current code is only encoding *some* of the values, and randomly generated names/ids coming from elsewhere in the framework could include characters that need to be encoded. Can you always be sure lifecycle_environment_id, operating_system_id, remote_execution_interface, and individual package names will never include characters that need to be encoded in a query string?

The current code is even having to explicitly use `'+'.join(packages)` within the string composition to property quote them, while claiming the urlencode pattern is not readable or understandable. `urlencode` will encode the list of packages in the expected way, by default - including encoding any of the package names themselves. You just use `' '.join(packages)` as the value in the dict of all query params, to space separate them instead of passing them as a python list.

What are you comparing to satellite in this case, and how is it any more maintainable?  Using what I'm suggesting, you would have a dictionary with all the key-value query parameters that is very readable when logged, instead of a collection of 15 variables that are all currently not logged or debuggable. You could then log the resulting URL after being encoded. The SSH library will be logging the curl command that gets executed, so you'll see the URL there and be able to directly compare to something from Satellite.  Right now to debug this, you have to log all 15 variables separately, then look at the ssh command log to see the full URL after all the individual variables are formatted in.

The default quoting method won't change, standard library functions don't just change like that.

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -1019,3 +1020,204 @@ def test_positive_rex_against_infra_hosts(
...
+
+        :expectedresults: Verify the job was successfully ran against the host converted to mqtt
+
+        :CaseImportance: Critical
+
+        :parametrized: yes
+        """
+        result = rhel_contenthost.execute(
+            f'curl -o /etc/pki/ca-trust/source/anchors/satellite-sat-engineering-ca.crt \
+                    {settings.repos["DOGFOOD_REPO_HOST"]}/pub/katello-server-ca.crt; \
```
mshriver, line 1051: Consider using `&&` so that `update-ca-trust` is only run if curl is successful.

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -1019,3 +1020,204 @@ def test_positive_rex_against_infra_hosts(
...
+                    update-ca-trust'
+        )
+        assert result.status == 0, 'Failed to download certificate'
```
mshriver, line 1054: Using `command; command`, if `curl` failed but `update-ca-trust` was successful, this pass will check.

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -1019,3 +1020,204 @@ def test_positive_rex_against_infra_hosts(
...
+        client_repo = (
+            f'{settings.repos["DOGFOOD_REPO_HOST"].replace("http", "https")}/pulp/content/'
```
mshriver, line 1056: Again this is where the standard library's urllib is suggested. You can parse and recompose the URL, directly and explicitly changing the protocol.

Its a stretch, but its possible the FQDN has the string `http` in it, separately from the protocol.

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -1019,3 +1020,204 @@ def test_positive_rex_against_infra_hosts(
...
+        result = rhel_contenthost.register(
+            module_capsule_configured_mqtt,
+            module_org,
+            smart_proxy_location,
+            module_ak_with_cv.name,
+            packages=['katello-agent'],
+            repo=client_repo,
+        )
+        assert result.status == 0, f'Failed to register host: {result.stderr}'
+        # result = rhel_contenthost.execute('subsription-manager identity/status')
```
mshriver, line 1074: Test step not needed?

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -1019,3 +1020,204 @@ def test_positive_rex_against_infra_hosts(
...
+                'job-template': 'Run Command - Script Default',
+                'inputs': 'command=ls',
+                'search-query': f"name ~ {rhel_contenthost.hostname}",
+            }
+        )
+        result = JobInvocation.info({'id': invocation_command['id']})
+        try:
+            assert result['success'] == '1'
+        except AssertionError:
+            result = 'host output: {}'.format(
```
mshriver, line 1096: golfing a bit here, but no need to label `result` here, just call AssertionError and define the string in it's call parenthesis.

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -1019,3 +1020,204 @@ def test_positive_rex_against_infra_hosts(
...
+            {
+                'job-template': 'Service Action - Script Default',
+                'inputs': 'action=status, service=yggdrasild',
+                'search-query': f"name ~ {rhel_contenthost.hostname}",
+            }
+        )
+        result = JobInvocation.info({'id': invocation_command['id']})
+        try:
+            assert result['success'] == '1'
+        except AssertionError:
```
mshriver, line 1215: This is done about 17 times in this module, might be time to write a function to cover the assertion pattern. Optional for this PR.

tests/foreman/cli/test_remoteexecution.py
mshriver: This (and related comment about the assert) are what I'd consider blocking at this point.

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -1019,3 +1020,204 @@ def test_positive_rex_against_infra_hosts(
...
+
+        :expectedresults: Verify the job was successfully ran against the host registered to mqtt
+
+        :CaseImportance: Critical
+
+        :parametrized: yes
+        """
+        result = rhel_contenthost.execute(
+            f'curl -o /etc/pki/ca-trust/source/anchors/satellite-sat-engineering-ca.crt \
+                    {settings.repos["DOGFOOD_REPO_HOST"]}/pub/katello-server-ca.crt; \
```
mshriver, line 1178: Same thing as above about the potential for this to hide an error.

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -1019,3 +1020,204 @@ def test_positive_rex_against_infra_hosts(
...
+                    update-ca-trust'
+        )
+        assert result.status == 0, 'Failed to download certificate'
+        client_repo = (
+            f'{settings.repos["DOGFOOD_REPO_HOST"].replace("http", "https")}/pulp/content/'
+            'Satellite_Engineering/QA/Satellite_Client/custom/Satellite_Client_Composes/'
+            f'Satellite_Client_RHEL{rhel_contenthost.os_version.major}_x86_64/'
+        )
+        # TODO client_repo should be changed to
```
mshriver, line 1187: @JacobCallahan updates here? I think the dogfood repo changes are in?

tests/foreman/cli/test_remoteexecution.py
pondrejk: submitted https://github.com/SatelliteQE/robottelo/issues/9985 for this

tests/foreman/cli/test_remoteexecution.py
pondrejk: I consider this temporary in line with the TODO few lines lower, I'd prefer not to do this at all, but I suppose we will settle on a new/updated constant pattern for the dogfood repo to accommodate different protocols soonish (? @JacobCallahan). Maybe, I can create an additional constant to care for this?

tests/foreman/cli/test_remoteexecution.py
pondrejk: submitted https://github.com/SatelliteQE/robottelo/issues/9985 for this

tests/foreman/cli/test_remoteexecution.py
pondrejk: dropped, thanks

tests/foreman/cli/test_remoteexecution.py
pondrejk: updated, thanks

robottelo/hosts.py
pondrejk: The intent here is to build the url *as Satelllite would do it* with all the quirks and inconsistencies. This can differ form the way urlencode would do it, and more differences can arise in the future. E.g. for multiple AKs sat would do `activation_keys=mykey%2Ctestkey` for multiple packages it would do `packages=one+two`. 

That's why I prefer to explicitly compose those strings, I don't see any harm in that. Going instead with `list input > space separated string > urlencode > + separated string` is just a long way around for the same result. 

Basically, I'd argue that emulating the Satellite's way of building that curl is *not* the intended purpose for urlencode, or at least it doesn't really make things easier. Still you'd have to deal with none states, strange joins and values, etc.  With current impl. you can log the curl command as well + you can read how it's built, non-compliant IDs would be acutally a bug against Sat, you don't want urlencode to take care for them. 

See, we can take this offline or to the separate issue if it's ok with you.

tests/foreman/cli/test_remoteexecution.py
pondrejk: Updated, thanks

mshriver (approved):

PR #9936, opened 2022-08-09
[Part-2] Helpers refactor of helpers.py helpers :)

jyejare:
Refactor:

- Fact: get_available_capsule_port function is used only with satellite server for identifying free ports. The options in that function is never used. Also, this property is already added into capsulemixins which isnt the right place since the operations in functions are happening on satellite.
  - Consideration: Moving under satellite mixins as property would be good since the option is never used. Removing this property from CapsuleMixins as we are finding open ports for capsule on `Satellite` not on `Capsule`.

- Fact: refactor of default_url_on_new_port function is used only with satellite server. Also, this property is already added into capsulemixins which isnt the right place since the operations in the function are happening on satellite.
  - Consideration: Moving under satellite mixins to access it from satellite object. Removing this property from CapsuleMixins for said issue. Also Extra [1] has been performed as part of this commit!

- Fact: Storage class is not being instantiated anywhere or used in robottelo. As well as there is a better replacement for it and thats `Box` which turns dict into object with attributes.
  - Consideration: Removed the class and even its robottelo unit tests.

- Fact:  get_func_name function isnt used anywhere in robottelo and I dont see its useful at all since we have better alternative for it an that is `inspect`.
  - Consideration: Removed!

- Fact: form_repo_url is used only in one test module for content management, the content management test cases are only tested from API.
  - Consideration: Moved the helper directly to the test module and renamed it to `get_published_repo_url` as thats the only module its used and doesnt have general applicability for other tests.

- Fact: create_repo function that creates a repo from given rpms and moves it to the directory to be accessed via HTTP is not being used anywhere. Also there are some similar names defs but are doing different work altogether.
  - Consideration: Removing it as I dont see a direct need of it in robottelo.

- Fact: extract_ui_token and get_web_session used to create UI session on the fly but both of them are no more required.
  - Consideration: Removing both of them because now we have airgun and satellite.ui_session to interact with satellites UI session.

- Fact: check_provisioned_host is being used to verify if the host ip of provisioned host pings from satellite only.
  - Consideration: Moved under contenthost class as parent class for sat and cap classes as general `ping_host` function to check any host with name or IP pings.

- Fact: slugify_component is only used in issue_handlers plugin and its tests.
  - Consideration: Moving to utils as a general purpose function.

- Fact: idgen function is not used anywhere.
  - Consideration: Removed

- Fact: InstalledCommand is used to generate satellite installer command `string` based on given options and its not playing with satellite system itself.
  - Consideration: Created new installer utils and moved `InstallerCommand` utility in that module.

- Fact: get_nailgun_config used in multiple test modules but without passing the user parameter anywhere, but, in tests the auth of user is being substituted. In some tests it makes to return admin user creds and hence wrong assertions are being made in the tests with admin user instead of non admin user.
  - Consideration: get_nailgun_config should be moved to config/__init__.py and actual user creds should be used for assertions with non-admin / admin configs as needed.

Extra:
- Some of these helpers being used in depreciated `cli/factory.py` and those factory functions been used tests. So as part of this effort I have removed concerning function from `cli/factory` and the dependent tests were moved to use `sat.cli_factory`.


jyejare:
trigger: robottelo
pytest: tests/foreman/destructive/test_realm.py --uses-fixtures module_fake_proxy

jyejare:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_realm.py --uses-fixtures module_fake_proxy

jyejare:
Th 6 test failures in commit https://github.com/SatelliteQE/robottelo/pull/9936/commits/ff19a1cdaffa9f4901122d8cf59c5f10ffeeb8d7 are similar to CI run and its not caused by this PR.

jyejare:
trigger: test-robottelo
pytest: tests/foreman/cli/test_organization.py --uses-fixtures proxy

jyejare:
trigger: test-robottelo
pytest: tests/foreman/api/test_smartproxy.py -k test_positive_create_with_name

devendra104:
trigger: test-robottelo
pytest: tests/foreman/api/test_smartproxy.py -k test_positive_create_with_name

devendra104:
Please don't pass any extra details in the PRT comment except the recommended one otherwise job could be failed with any of the reasons, your recent job got failed because of the below comment which had the backquotes(```).

> ```
> All tests passed for this except html based name in smart proxy creation which could be ignored!
> ```

jyejare:
@devendra104 Ohh! I was able to run the PRT successfully and without backquotes. After the successful PRT, I edited the comment and said this PRT was successful so that reviewers know that PRT went well or if it failed, failed with a proper reason.

So we need to disable PRT run on `edited` comments and just run them when we `create`.

robottelo/host_helpers/satellite_mixins.py
```diff
@@ -101,6 +106,96 @@ def md5_by_url(self, url):
         return result.stdout
 
 
+class SystemInfo:
+    """Things that needs access to satellite shell for gaining satellite system configuration"""
+
+    @property
+    def get_available_capsule_port(self):
```
mshriver, line 113: Generally properties are named like nouns, not like actions. `get_something()` is generally a function call to an instance method, whereas `available_capsule_port` is a *property* of a Satellite instance.  Consider if this property had a `setter` method - does setting `get_available_capsule_port` make sense - or does setting `available_capsule_port` make sense?  I'd say the latter.

mshriver (approved):
My only issue is with property naming, which I'd like to see go in, but isn't blocking.

robottelo/host_helpers/satellite_mixins.py
jyejare: Resolved by renaming it to `available_capsule_port` !

tests/foreman/api/test_activationkey.py
```diff
@@ -424,9 +424,8 @@ def test_positive_remove_user():
     """
     password = gen_string('alpha')
     user = entities.User(password=password, login=gen_string('alpha'), admin=True).create()
-    cfg = get_nailgun_config()
-    cfg.auth = (user.login, password)
-    ak = entities.ActivationKey(cfg).create()
+    user_cfg = user_nailgun_config(user.login, password)
+    ak = entities.ActivationKey(server_config=user_cfg).create()
```
JacobCallahan, line 428: We should evaluate if this is compatible with the satellite objects' entities.

robottelo/hosts.py
```diff
@@ -1259,6 +1260,19 @@ def register_to_cdn(self):
...
 
+    def ping_host(self, host):
+        """Check the provisioned host status by pinging the ip of host
+
+        :param host: IP address or hostname of the provisioned host
+        :returns: None
+        :raises: : `ProvisioningCheckError` if the host is not pingable
+        """
+        result = self.execute(
+            f'for i in {{1..60}}; do ping -c1 {host} && exit 0; sleep 20; done; exit 1'
```
JacobCallahan, line 1271: This seems like a weird way to do this. Why not do this instead?

`f'ping -c 60 -i 20 {host}'`

mshriver: Calling ping like that won't stop once the first echo is returned.

tests/foreman/api/test_activationkey.py
jyejare: @JacobCallahan  Updated all files! Please revisit!

Gauravtalreja1 (commented):
Ack, with few suggestions

shweta83 (approved):

tests/foreman/cli/test_capsule.py
```diff
@@ -58,7 +55,11 @@ def test_negative_create_with_url():
     """
     # Create a random proxy
     with pytest.raises(CLIFactoryError, match='Could not create the proxy:'):
-        make_proxy({'url': 'http://{}:{}'.format(gen_string('alpha', 6), gen_string('numeric', 4))})
+        target_sat.cli_factory.make_proxy(
+            {
+                'url': 'http://{}:{}'.format(gen_string('alpha', 6), gen_string('numeric', 4)),
```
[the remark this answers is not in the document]
jyejare: The `format` function looks more appropriate here than fstring makes it more complex. But still I can change it.

Gauravtalreja1 (approved):

jyejare:
Can we merge @JacobCallahan @mshriver @shweta83 and @Gauravtalreja1 ?

JacobCallahan (approved):

PR #9937, opened 2022-08-09
downgrade ibutsu plugin version to resolve dependency issues

rplevka:
Due to the pytest version breaking changes (affecting robottelo code):
https://docs.pytest.org/en/7.1.x/deprecations.html#fspath-argument-for-node-constructors-replaced-with-pathlib-path

and time constraints, I decided to downgrade the ibutsu plugin version for now.


rplevka:
oh, no
https://docs.pytest.org/en/7.1.x/deprecations.html#fspath-argument-for-node-constructors-replaced-with-pathlib-path

PR #9940, opened 2022-08-10
Remove RHEL9 from supportability matrix

JacobCallahan:
We currently are not able to deploy rhel9 hosts. Due to this, we're
going to temporarily disable the entry.

### DIFF
--- a/conf/supportability.yaml
+++ b/conf/supportability.yaml
@@ -1,4 +1,4 @@
 supportability:
   content_hosts:
     rhel:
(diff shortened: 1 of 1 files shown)


mshriver (approved):

ogajduse:
@pondrejk FYI: This is precisely the opposite of what you wanted to do.

mshriver:
@ogajduse @pondrejk the deploy issues are being worked on, and we're expecting to add this back this week. Couple fires going on with subscriptions, and we wanted to clear this before snap6 testing.

PR #9943, opened 2022-08-10
Add Manifester settings and session-scoped fixtures

synkd:
This commit adds two session-scoped fixtures that use Manifester to
generate manifests for use with Satellites. These fixtures can be
inherited by other fixtures (such as module_manifest_org) or passed
directly to test functions. This commit also adds an example
manifester_settings.yaml file to the base Robottelo directory.


pytest_fixtures/component/taxonomy.py
```diff
@@ -1,6 +1,7 @@
 # Content Component fixtures
 import pytest
 
+from manifester import Manifester
```
Griffin-Sullivan, line 4: We probably want to add manifester to requirements.txt?

mshriver: probably :)

manifester_settings.yaml.example
```diff
@@ -0,0 +1,36 @@
+#rhsm-manifester settings
+log_level: "info"
+offline_token: ""
+proxies: {"https": ""}
+manifest_category:
+  golden_ticket:
+    # An offline token can be generated at https://access.redhat.com/management/api
+    offline_token: ""
+    # Value of sat_version setting should be in the form 'sat-6.10'
+    sat_version: "sat-6.10"
```
mshriver, line 10: 6.12 for master branch

JacobCallahan:
This can be updated a bit after https://github.com/SatelliteQE/manifester/pull/10 has been merged/released.

manifester_settings.yaml.example
mshriver: Required: format this using `self.server.version`

synkd: While I take your point that the version number should not be hardcoded, I'm not sure exactly what you're referring to with `self.server.version`. This pattern doesn't appear to be used elsewhere in Robottelo. I see a couple of uses of `settings.server.version.release` - is this what you were referring to, or did you mean something else?

mshriver: Its a `this.server.version` reference actually. Best upstream example is in the capsule template:

https://github.com/SatelliteQE/robottelo/blob/master/conf/capsule.yaml.template#L3

This one is using `robottelo.satellite_version`, but I would recommend using `server.version` fields instead.

synkd: Ok, I'm getting errors with both of these. I've tried a couple of different variants, including:

`sat_version: "@jinja sat-{{this.server.version}}"`
`sat_version: "@format sat-{this.server.version}"`
`sat_version: "@jinja sat-{{this.robottelo.satellite_version}}"`
`sat_version: "@format sat-{this.robottelo.satellite_version}"`

They're all resulting in errors when the expression in braces is evaluated. Is this expected to work when running locally? Might the fact that I have manifester_settings.yaml in the base robottelo directory instead of in `robottelo/conf/` be a factor?

pytest_fixtures/component/taxonomy.py
```diff
@@ -83,6 +84,26 @@ def module_gt_manifest_org(module_target_sat):
...
+# Satellite QE RHSM account. Manifester can be safely used locally with personal 
+# or stage RHSM accounts. 
+
+# Yields a manifest in entitlement mode with subscriptions determined by the 
+# `manifest_category.robottelo_automation` setting in manifester_settings.yaml.
+@pytest.fixture(scope='session')
+def session_entitlement_manifest():
+    manifester = Manifester(manifest_category='robottelo_automation')
+    with manifester.get_manifest() as manifest:
+        yield manifest
```
JacobCallahan, line 98:
```suggestion
    with Manifester(manifest_category='robottelo_automation') as manifest:
        yield manifest
```

pytest_fixtures/component/taxonomy.py
```diff
@@ -83,6 +84,26 @@ def module_gt_manifest_org(module_target_sat):
...
+
+# Yields a manifest in Simple Content Access mode with subscriptions determined by the
+# `manifest_category.golden_ticket` setting in manifester_settings.yaml
+@pytest.fixture(scope='session')
+def session_sca_manifest():
+    manifester = Manifester(manifest_category='golden_ticket')
+    with manifester.get_manifest() as manifest:
+        yield manifest
```
JacobCallahan, line 106:
```suggestion
    with Manifester(manifest_category='golden_ticket') as manifest:
        yield manifest
```

JacobCallahan (changes requested):
newer Manifester behavior

pytest_fixtures/component/taxonomy.py
```diff
@@ -83,6 +84,24 @@ def module_gt_manifest_org(module_target_sat):
     org.manifest_filename = manifest.filename
     return org
 
+# Note: Manifester should not be used with the Satellite QE RHSM account until 
+# subscription needs are scoped and sufficient subscriptions added to the 
+# Satellite QE RHSM account. Manifester can be safely used locally with personal 
+# or stage RHSM accounts. 
+
+# Yields a manifest in entitlement mode with subscriptions determined by the 
+# `manifest_category.robottelo_automation` setting in manifester_settings.yaml.
```
JacobCallahan, line 93: non-blocker, but these comments should really be docstrings for the fixtures

manifester_settings.yaml.example
```diff
@@ -0,0 +1,36 @@
+#rhsm-manifester settings
```
ogajduse, line 1: Prefix all manifester settings with `MANIFESTER` key like we have it in the rest of the robottelo configuration. This might require a change on manifester side.

manifester_settings.yaml.example
ogajduse: _conf/manifester_settings.yaml_:
```yaml
MANIFESTER:
  manifest_category:
    robottelo_automation:
      sat_version: "@jinja sat-{{this.server.version.release}}"
```

```python
settings.manifester.manifest_category.robottelo_automation.sat_version
```

Gives me `sat-6.11`

ogajduse (changes requested):

manifester_settings.yaml.example
ogajduse: Now I see https://github.com/SatelliteQE/manifester/pull/11. So I might be totally off with the above suggestion.

conf/manifest.yaml.example
```diff
@@ -0,0 +1,31 @@
+MANIFESTER:
```
mshriver, line 1: `MANIFEST` here - the path that gets established in the `Settings` instance is dependent on the top level of the map, not the yaml file itself.

Our convention is for the file's top level map key to match the file name.

mshriver (approved):
Requires manifester update and yaml example update, approved otherwise!

Solid introduction of the tool with some basic fixtures, looking forward to the PR applying them to cases that are currently cloning the 'fake' manifest.

JacobCallahan (approved):

ogajduse (approved):
:+1: Elegant solution in combination with SatelliteQE/manifester#11.

PR #9949, opened 2022-08-10
Remove Inflector dependency

JacobCallahan:
The use of Inflector here was overkill for what we needed.
I replace the calls with the result of the calls and removed the dependency and unit test.


JacobCallahan:
@shweta83 I saw this in your presentation and just had to remove it. Once this is merged, I'll rebase your PR to see if there are conflicts.

mshriver (approved):

PR #9950, opened 2022-08-11
Smoothing Teardown in Azure Host provisioning

jyejare:
The azure test cases were failing in teardown due to VM being deleted in yielding VM creation fixture.

Now at the test level, we verify the VMs existence( in case teardown in yielding fixture doesnt work) before we delete it.

### DIFF
--- a/tests/foreman/cli/test_computeresource_azurerm.py
+++ b/tests/foreman/cli/test_computeresource_azurerm.py
@@ -365,7 +365,8 @@ class TestAzureRMFinishTemplateProvisioning:
                 )
                 yield host
                 with satellite_setting('destroy_vm_on_host_delete=True'):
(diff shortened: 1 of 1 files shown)


mshriver (approved):

mshriver:
Taking this a step further, in CFME days we actually implemented a `delete_if_exists` that would do exactly this pattern, and was a bit cleaner/easier to use.  Something we can consider for robottelo, esp for cloud instances where we can use wrapanapi directly to do cleanup.

jyejare:
@mshriver The checking of existence of VM here is from Satellite not from backend but agree we could employ wrapanapi here. We are using wrapanapi only but with if condition , but `delete_if_exists` could be one liner.

PR #9968, opened 2022-08-15
Automation for bz-1848111

latran:
Automation coverage for bz: https://bugzilla.redhat.com/show_bug.cgi?id=1848111

test results:
```
pytest tests/foreman/cli/test_contentview.py -k test_positive_pulp_workers_running_state
========= 1 passed, 129 deselected, 2 warnings in 13.26s ================
```

### DIFF
--- a/tests/foreman/sys/test_pulp3_filesystem.py
+++ b/tests/foreman/sys/test_pulp3_filesystem.py
@@ -15,7 +15,6 @@
 :CaseImportance: High
 
 :Upstream: No
-
 """
 import pytest
 
@@ -27,7 +26,6 @@ def test_selinux_status(target_sat):
     :id: 43218070-ac5e-4679-b74a-3e2bcb497a0a
 
     :expectedresults: SELinux is enabled and there are no denials
-
     """
     # check SELinux is enabled
     result = target_sat.execute('getenforce')
@@ -55,7 +53,6 @@ def test_pulp_directory_exists(target_sat, directory):
     :parametrized: yes
 
     :expectedresults: Directories are present
-
     """
     # check pulp3 directories present
     assert target_sat.execute(f'test -d {directory}').status == 0, f'{directory} must exist.'
@@ -68,7 +65,6 @@ def test_pulp_service_definitions(target_sat):
     :id: e584faea-dede-4895-8d7f-411cb074f6c0
 
     :expectedresults: systemd files are present
-
     """
     # check pulpcore settings
     result = target_sat.execute('systemctl cat pulpcore-content.socket')
@@ -83,3 +79,20 @@ def test_pulp_service_definitions(target_sat):
     # check pulp3 configuration file present
     result = target_sat.execute('test -f /etc/pulp/settings.py')
     assert result.status == 0
+
+
+@pytest.mark.tier1
+def test_pulp_workers_status(target_sat):
+    """Ensure pulpcore-workers are in active (running) state
+
+    :BZ: 1848111
+
+    :id: f0b8b354-1a30-41c3-93d7-fb51e99c82b6
+
+    :customerscenario: true
+
+    :expectedresults: Pulpcore-workers are in active (running) state
+    """
+    output = target_sat.execute('systemctl status pulpcore-worker* | grep Active')
+    result = output.stdout.rstrip().splitlines()
+    assert all('Active: active (running)' in r for r in result)


tests/foreman/cli/test_contentview.py
```diff
@@ -3962,3 +3962,23 @@ def test_positive_inc_update_should_not_fail(self, module_org):
         assert result[2]
         content_view = ContentView.info({'id': content_view['id']})
         assert '1.1' in [cvv_['version'] for cvv_ in content_view['versions']]
+
+    @pytest.mark.tier1
+    def test_positive_pulp_workers_running_state(self, target_sat):
```
mshriver, line 3967: Consider making this part of the pulp service assertions done here:
https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/sys/test_pulp3_filesystem.py#L65

This isn't really related to contentview CLI.

tests/foreman/cli/test_contentview.py
```diff
@@ -3962,3 +3962,23 @@ def test_positive_inc_update_should_not_fail(self, module_org):
...
+        :id: f0b8b354-1a30-41c3-93d7-fb51e99c82b6
+
+        :customerscenario: true
+
+        :expectedresults: 6 pulpcore-workers are in active (running) state
+
+        :SubComponent: Pulp
+        """
+        output = target_sat.execute('systemctl status pulpcore-worker* | grep Active')
+        result = output.stdout.split('\n')
```
mshriver, line 3981: If you use `rstrip()` you won't have to deal with the empty string element in the list.

tests/foreman/cli/test_contentview.py
```diff
@@ -3962,3 +3962,23 @@ def test_positive_inc_update_should_not_fail(self, module_org):
...
+        assert len(result) - 1 == 6
```
mshriver, line 3982: Why assert against 6 specifically? Is there an installer default tied to this assumption?  Any way to check config/answer/log files for what the expected number of pulpcore-worker services is? Cursory search didn't find anything, other than the service definition being dynamic.

There are really a few assertions going on here in addition to the BZ, its good. Combined with the other service test for pulp, I think you could improve them as well by checking their run state in addition to configuration.

Suggestion for a bit more pythonic way to make this assertion (along with the above change to using rstrip).

```
# this might not fit on 100 chars, break up if necessary. rstrip + splitlines instead of explicitly using `\n`.

output = sat.execute('systemctl status pulpcore-worker* |grep "Active:"').stdout.rstrip().splitlines()
assert all('Active: active (running)' in r for r in output)
```

mshriver (changes requested):
see comments

tests/foreman/cli/test_contentview.py
```diff
@@ -3962,3 +3962,23 @@ def test_positive_inc_update_should_not_fail(self, module_org):
...
+
+        :id: f0b8b354-1a30-41c3-93d7-fb51e99c82b6
+
+        :customerscenario: true
+
+        :expectedresults: 6 pulpcore-workers are in active (running) state
+
+        :SubComponent: Pulp
+        """
+        output = target_sat.execute('systemctl status pulpcore-worker* | grep Active')
```
Gauravtalreja1, line 3980: Would suggest using f-m CLI to check service status with `brief` and `only` options rather than systemctl, where you could validate the RC of the command instead of splitting stdout
```suggestion
        output = target_sat.cli.Service.status({'brief': True, 'only': 'pulpcore-worker*'})
```

tests/foreman/cli/test_contentview.py
Gauravtalreja1: pulpcore-worker count is set by the installer flag and currently, it is set to 6 because of the number of CPU cores used. We probably shouldn't keep this number constant in this test, so rstrip should be useful here.

tests/foreman/cli/test_contentview.py
mshriver: This bleeds into 'testing' the CLI for services - not a bad idea, but may influence how the tests are organized. I've suggested verifying the pulpcore-worker states in the existing pulp_filesystem module, if using `Service.status` it would start to make more sense that the test is in a module for services CLI, and pulpcore-worker is just one of the services being validated.

latran:
Thanks for the suggestions @mshriver and @Gauravtalreja1 !  This looks a lot cleaner now.

Test result:
```
pytest tests/foreman/sys/test_pulp3_filesystem.py -k test_pulp_workers_status

================= 1 passed, 6 deselected, 2 warnings in 12.22s ======================
```

tests/foreman/sys/test_pulp3_filesystem.py
```diff
@@ -83,3 +79,20 @@ def test_pulp_service_definitions(target_sat):
...
+
+    :id: f0b8b354-1a30-41c3-93d7-fb51e99c82b6
+
+    :customerscenario: true
+
+    :expectedresults: Pulpcore-workers are in active (running) state
+    """
+    output = target_sat.execute('systemctl status pulpcore-worker* | grep Active')
+    result = output.stdout.rstrip().splitlines()
+    assert all('Active: active (running)' in r for r in result)
```
Griffin-Sullivan, line 98: Never seen `all()` before. Very nice!

mshriver (approved):

PR #9983, opened 2022-08-19
Switching most ContentHost checkouts to container-based hosts

JacobCallahan:
This change is relatively small but targets a much larger impact.
Unless otherwise specified, ContentHost checkouts will use containers, when supported.
This can be disabled at the fixture level by modifying the request params;
the test level by marking the test with `no_containers`;
or at runtime with the --no-containers option which will disable containers for all tests.

By default, any remote execution tests will not use containers.

A new config field has been added to container_repo.yaml to account for the new container host. This could additionally be used for image syncing.
I additionally refactored the content_host.yaml structure to be more clear and support containers


pytest_fixtures/core/contenthosts.py
```diff
@@ -19,20 +19,21 @@ def host_conf(request):
...
+    if not deploy_kwargs:
+        deploy_kwargs = settings.content_host.get(_rhelver).get('vm', {})
+    conf.update(deploy_kwargs)
+    # CI-defined params specific to only a few jobs
     rhel_compose_id = settings.get(f"content_host.deploy_kwargs.{_rhelver}.compose")
     if rhel_compose_id:
         conf['deploy_rhel_compose_id'] = rhel_compose_id
     scenario = settings.get(f"content_host.deploy_kwargs.{_rhelver}.scenario")
     if scenario:
         conf['deploy_scenario'] = scenario
```
JacobCallahan, line 36: @rplevka @ogajduse i'd like to consult the both of you on how to handle this better

JacobCallahan:
Once we agree on the new config structure, I'll submit an MR for the changes in CI

synkd (commented):
My understanding is that RHEL containers managed by docker or podman inherit their subscriptions from the system on which they are hosted. [1] Will this change impact our uses of Red Hat subscriptions in Robottelo?

[1] https://access.redhat.com/solutions/5870841, but note that I'm referring to the statement to this effect in the `Issue` section, as I assume we are using Red Hat-provided docker or podman and thus the rest of the KCS will not apply.

JacobCallahan:
@synkd that is the case for traditional containers. However, these images circumvent that behavior in subscription-manager and are seen as actual content hosts.

pytest_plugins/markers.py
```diff
@@ -19,6 +19,7 @@ def pytest_configure(config):
         "build_sanity: Fast, basic tests that confirm build is ready for full test suite",
         "rhel_ver_list: Filter rhel_contenthost versions by list",
         "rhel_ver_match: Filter rhel_contenthost versions by regexp",
+        "no_containers: Disable container hosts from being used in favor of VMs",
```
omkarkhatavkar, line 22: instead `no_contianers` cant we be just specific such, `run_on_vm` or `only_vm`?

omkarkhatavkar: `no_containers` is the extra layer to disable the behavior dynamically and use the VM.

rplevka: My 0.05€:
I'd prefer `no_containers` to prevent potential confusion of what this mark is doing in case we (very unlikely) introduced bare metal hosts.

conf/content_host.yaml.template
```diff
@@ -1,33 +1,52 @@
...
-            CORES: 1
+  default_rhel_version: 7
+  rhel6:
+    vm:
+      workflow: deploy-base-rhel
+      release: '6.10'
+      memory: 1GiB
+      cores: 1
+    container:
+      container_host: "@format {this.container_repo.registries.auto_hub.fqdn}/rhel6:latest"
```
rplevka, line 10: a little redundancy in the naming isn't it?
how about just `host` since it is a child of `container` key?

optional: wouldn't something like `image_url` describe this better?

rplevka (approved):

pytest_plugins/markers.py
JacobCallahan: we could, but I personally think no_containers is more clear about what it's doing.

conf/content_host.yaml.template
JacobCallahan: `container_host` is the name of the Broker argument for its Container provider. It's the equivalent of `workflow` for AnsibleTower. In the current form, these fields require no additional processing before being passed to Broker.

JacobCallahan: We've actually decided to strip that from the image names in general since there is little value in pushing to automation hub.

pytest_fixtures/core/contenthosts.py
```diff
@@ -180,13 +180,21 @@ def container_contenthost(rhel7_contenthost, target_sat):
 
 @pytest.fixture
 def centos_host(request, version):
-    request.param = {"rhel_version": version.split('.')[0], "distro": "centos"}
-    with Broker(**host_conf(request), host_classes={'host': ContentHost}) as host:
+    request.param = {
+        "rhel_version": version.split('.')[0],
```
mshriver, line 184: Consider creating a version object here to use `.major`? Change from current behavior, def optional.

mshriver (approved):

Griffin-Sullivan (approved):

PR #9986, opened 2022-08-23
Tests for satellite clone

LadislavVasina1:
I have implemented tests for satellite clone. Fixture that provides suitable RHEL for sat (6 cores, 20 GiB) to be installed to has been added too because none of the existing fixtures was suitable for me.
While implementing the tests I have found the wrong keys for rhel content host parametrization and I have also found an incorrect list of repos needed for sat installation on RHEL8 ('ansible-2.9-for-rhel-8-x86_64-rpms' deleted from list). 
image


omkarkhatavkar:

tests/foreman/maintain/test_clone.py
```diff
@@ -0,0 +1,163 @@
...
+    rhel_version = sat_ready_rhel._v_major
+
+    # SATELLITE PART - SOURCE SERVER
+    # Enabling and starting services
+    assert target_sat.execute('satellite-maintain service enable').status == 0
+    assert target_sat.execute('satellite-maintain service start').status == 0
+    # Doing backup
+    backup_result = target_sat.execute(
+        f'satellite-maintain backup {backup_type} --assumeyes /var/backup'
+    )
```
Gauravtalreja1, line 46:
```suggestion
    result = sat_maintain.cli.Backup.run_backup(
        backup_dir='/var/backup',
        backup_type=backup_type,
        options={'assumeyes': True, 'plaintext': True},
    )
```

tests/foreman/maintain/test_clone.py
```diff
@@ -0,0 +1,163 @@
...
+    # Disabling repositories
+    assert sat_ready_rhel.execute('subscription-manager repos --disable=*').status == 0
+    # Getting satellite maintenance repo
+    res = requests.get(
+        f'http://ohsnap.sat.engineering.redhat.com/api/releases/{target_sat.version}/'
+        f'el{rhel_version}/satellite/repo_file'
+    )
+    escaped = res.text.replace('$', '\\$')
+    assert sat_ready_rhel.execute(f'echo "{escaped}" > /etc/yum.repos.d/satellite.repo').status == 0
+    assert sat_ready_rhel.execute('ls /etc/yum.repos.d/satellite.repo').status == 0
```
Gauravtalreja1, line 74: We've a helper available for downloading repofiles available in contenthost_mixin
```suggestion
     sat_ready_rhel.download_repofile()
```

tests/foreman/maintain/test_clone.py
```diff
@@ -0,0 +1,163 @@
...
+import requests
+
+from robottelo import constants
+from robottelo.config import settings
+
+USER = settings.subscription.rhn_username
+PASSWORD = settings.subscription.rhn_password
+SSH_PASS = settings.server.ssh_password
+POOL = settings.subscription.rhn_poolid
+BACKUP_DIR = '/var/backup'
```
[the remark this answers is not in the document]
LadislavVasina1: True I forgot them there. But I can leave SSH_PASS there if it's being used or not?

robottelo/constants/__init__.py
```diff
@@ -247,7 +247,6 @@ class Colored(Box):
 OHSNAP_RHEL8_REPOS = (
     'rhel-8-for-x86_64-baseos-rpms',
     'rhel-8-for-x86_64-appstream-rpms',
-    'ansible-2.9-for-rhel-8-x86_64-rpms',
```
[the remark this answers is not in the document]
vsedmik: I think it's a part of appstream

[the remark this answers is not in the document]
LadislavVasina1: There is no ansible repo in installation in satellite documentation. https://access.redhat.com/documentation/en-us/red_hat_satellite/6.11/html-single/installing_satellite_server_in_a_connected_network_environment/index#:~:text=%23%20subscription%2Dmanager%20repos%20%2D%2Ddisable%20%22*%22-,Enable%20the%20following%20repositories%3A,-%23%20subscription%2Dmanager%20repos%20%2D%2Denable%3Drhel%2D8

tests/foreman/maintain/test_clone.py
LadislavVasina1: As I am looking at _download_repofile_ definition it returns only tools not satellite repofile link
https://github.com/SatelliteQE/robottelo/blob/9609c95caedaa4596fb257620bf663845c269f89/robottelo/host_helpers/contenthost_mixins.py#L43-L50

robottelo/constants/__init__.py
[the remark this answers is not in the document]
ogajduse: We just need `ansible-core` package from AppStream repository. We do not need to enable ansible repository.

tests/foreman/maintain/test_clone.py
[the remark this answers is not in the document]
ogajduse: > True I forgot them there. But I can leave SSH_PASS there if it's being used or not?

Agreed with keeping `SSH_PASS`. We already use something similar in `test_templatesync.py`
https://github.com/SatelliteQE/robottelo/blob/9609c95caedaa4596fb257620bf663845c269f89/tests/foreman/api/test_templatesync.py#L34

tests/foreman/maintain/test_clone.py
ogajduse: @JacobCallahan @jyejare What is your suggestion here? It would be really nice to use this method, but it does not allow specifying the `product` and it relies on the class name instead.

tests/foreman/maintain/test_clone.py
```diff
@@ -0,0 +1,163 @@
...
+
+:Requirement: foreman-maintain
+
+:CaseAutomation: Automated
+
+:CaseLevel: Component
+
+:CaseComponent: SatelliteClone
+
+:Assignee: lpramuk
```
mshriver, line 11: @lpramuk you know @LadislavVasina1 is making you own this module? :wink:

tests/foreman/maintain/test_clone.py
```diff
@@ -0,0 +1,163 @@
...
+USER = settings.subscription.rhn_username
+PASSWORD = settings.subscription.rhn_password
+SSH_PASS = settings.server.ssh_password
+POOL = settings.subscription.rhn_poolid
+BACKUP_DIR = '/var/backup'
+
+
+@pytest.mark.parametrize(
+    "sat_ready_rhel", [7, 8] if settings.server.version.rhel_version < 8 else [8], indirect=True
+)
```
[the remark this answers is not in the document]
mshriver: Given how we execute the tests, I'd not parametrize it regardless.

Otherwise the sat-6.11-rhel7-SatelliteClone pipeline will run the test against a rhel7 sat, and the sat-6.11-rhel8-SatelliteClone pipeline will run it against both - repeating the test from the first pipeline.

The fixture should just use the OS version of satellite that's in settings. Replicate the target satellite's base OS. The separate pipelines will cover the two base OS 'parameters'.

tests/foreman/maintain/test_clone.py
```diff
@@ -0,0 +1,163 @@
...
+
+    # RHEL PART - TARGET SERVER
+    assert sat_ready_rhel.execute('ls /backup/config_files.tar.gz').status == 0
+    # Registering to cdn
+    sat_ready_rhel.register_to_cdn()
+    # Disabling repositories
+    assert sat_ready_rhel.execute('subscription-manager repos --disable=*').status == 0
+    # Getting satellite maintenance repo
+    res = requests.get(
+        f'http://ohsnap.sat.engineering.redhat.com/api/releases/{target_sat.version}/'
```
mshriver, line 69: Read this domain name from settings, shouldn't be hardcoded.

ogajduse: There already is a suggestion by @Gauravtalreja1 mentioning the use of `sat_ready_rhel.download_repofile()`.

robottelo/constants/__init__.py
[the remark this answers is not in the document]
Gauravtalreja1: hmm, weird, el7 requires ansible repo but el8 doesn't in documentation, and I check we do enable it in satlab temp workflows too, so I think better to check with devs

tests/foreman/maintain/test_clone.py
lpramuk: SatelliteClone = lpramuk, 
it could change in future

robottelo/constants/__init__.py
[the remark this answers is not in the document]
lpramuk: @Gauravtalreja1 RHEL8 is my feature you can check my MRs in satlab-tower - MR611
no ansible repo on RHEL8 please

tests/foreman/maintain/test_clone.py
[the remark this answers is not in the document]
ogajduse: I am not sure that I follow your comment 100%.
The support matrix is the following:
```
Satellite 6.11 on RHEL 7 ---clone---> Satellite 6.11 on RHEL 7
Satellite 6.11 on RHEL 7 ---clone---> Satellite 6.11 on RHEL 8
Satellite 6.11 on RHEL 8 ---clone---> Satellite 6.11 on RHEL 8
```
So the `sat-6.11-rhel7-SatelliteClone` pipeline should also test that we can clone from RHEL 7 to RHEL 8.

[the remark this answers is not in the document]
mshriver: ack, good catch @ogajduse, still needs the parametrization in `6.11.z` branch, just not in the master branch.

tests/foreman/maintain/test_clone.py
```diff
@@ -0,0 +1,163 @@
...
+    assert sat_ready_rhel.execute(subscription_mng_cmd).status == 0
+
+    # If we want to do restore on rhel 8 we need to enable satellite module
+    if rhel_version == 8:
+        # Enabling satellite module
+        assert sat_ready_rhel.execute('dnf module enable -y satellite:el8').status == 0
+    # Install satellite-clone
+    assert sat_ready_rhel.execute('yum install satellite-clone -y').status == 0
+    # Run satellite-clone
+    assert sat_ready_rhel.execute('satellite-clone -y', timeout='3h').status == 0
```
omkarkhatavkar, line 162: timeout feels bigger ? is it expected.

LadislavVasina1: While doing normal backup+restore of fresh satellite it took around 1.5hours. I don't know how long it could take if the satellite from which the backup is being made of has a lot of stuff on it. So I set the timeout to 3 hours because of that.

tests/foreman/maintain/test_clone.py
```diff
@@ -0,0 +1,163 @@
...
+    # If we want to do restore on rhel 8 we need to enable satellite module
+    if rhel_version == 8:
+        # Enabling satellite module
+        assert sat_ready_rhel.execute('dnf module enable -y satellite:el8').status == 0
+
+    # Install satellite-clone
+    assert sat_ready_rhel.execute('yum install satellite-clone -y').status == 0
+    # Run satellite-clone
+    assert sat_ready_rhel.execute('satellite-clone -y', timeout='3h').status == 0
+    assert sat_ready_rhel.execute('foreman-maintain health check').status == 0
```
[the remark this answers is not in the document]
LadislavVasina1: `sat_ready_rhel` is ContentHost object not Satellite so I can't use what you are suggesting
`AttributeError: 'ContentHost' object has no attribute 'cli'`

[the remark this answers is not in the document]
ogajduse: @JacobCallahan What do you say to
```python
sat = Satellite(sat_ready_rhel.__dict__)
sat_ready_rhel.cli.Health.check().status == 0
```
?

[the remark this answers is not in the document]
mshriver: Without digging into whether all the kwargs would be handled, I think a simpler approach can be taken.

```
assert sat_ready_rhel.execute('satellite-clone -y', timeout='3h').status == 0
cloned_sat = Satellite(sat_ready_rhel.hostname)
assert cloned_sat.cli.Health.check().status == 0

```

tests/foreman/maintain/test_clone.py
```diff
@@ -0,0 +1,166 @@
...
+import requests
+
+from robottelo import constants
+from robottelo.config import settings
+from robottelo.hosts import Satellite
+
+SSH_PASS = settings.server.ssh_password
+
+
+@pytest.mark.parametrize("sat_ready_rhel", [8], indirect=True)
```
Gauravtalreja1, line 29: Instead of parametrizing, would suggest setting el8 default as similar to cores and memory in `sat_ready_rhel`?

tests/foreman/maintain/test_clone.py
```diff
@@ -0,0 +1,166 @@
...
+@pytest.mark.parametrize('backup_type', ['online', 'offline'])
+def test_positive_clone_backup(target_sat, sat_ready_rhel, backup_type):
+    rhel_version = sat_ready_rhel._v_major
```
Gauravtalreja1, line 32: please add some metadata to these tests

tests/foreman/maintain/test_clone.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # SATELLITE PART - SOURCE SERVER
+    # Enabling and starting services
+    assert target_sat.cli.Service.enable().status == 0
+    assert target_sat.cli.Service.start().status == 0
+    # Doing backup
+    no_pulp_backup_result = target_sat.execute(
+        'satellite-maintain backup offline --skip-pulp-content ' '--assumeyes /var/backup'
+    )
```
Gauravtalreja1, line 106:
```suggestion
    no_pulp_backup_result = target_sat.cli.Backup.run_backup(
        backup_dir='/var/backup',
        backup_type=backup_type,
        options={'assumeyes': True, 'plaintext': True, 'skip-pulp-content': True},
    )
```

tests/foreman/maintain/test_clone.py
```diff
@@ -0,0 +1,166 @@
...
+    assert sat_ready_rhel.execute(f'echo "{escaped}" > /etc/yum.repos.d/satellite.repo').status == 0
+    assert sat_ready_rhel.execute('ls /etc/yum.repos.d/satellite.repo').status == 0
+
+    # Getting repos to enable according to version of used rhel
+    repos_to_enable = getattr(constants, f'OHSNAP_RHEL{rhel_version}_REPOS')
+    subscription_mng_cmd = 'subscription-manager repos'
+    for repo in repos_to_enable:
+        subscription_mng_cmd += f' --enable={repo}'
+    # Enabling repositories
+    assert sat_ready_rhel.execute(subscription_mng_cmd).status == 0
```
Gauravtalreja1, line 149:
```suggestion
    # Enabling repositories
    for repo in getattr(constants, f"OHSNAP_RHEL{rhel_version}_REPOS"):
        sat_ready_rhel.enable_repo(repo, force=True)
```

tests/foreman/maintain/test_clone.py
LadislavVasina1: I want to use `sat_ready_rhel` fixture also for 6.11 if my tests are going to be merged to 6.11 branch.

ogajduse: I would leave the parametrization here because with Sat 6.13 we will be moving to RHEL 9 as a platform. This parametrization would come in handy in this case.

robottelo/hosts.py
```diff
@@ -1140,6 +1140,7 @@ def register_to_cdn(self):
             username=settings.subscription.rhn_username,
             password=settings.subscription.rhn_password,
             releasever=release_version,
+            auto_attach=auto_attach,
```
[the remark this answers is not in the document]
ogajduse: resolving this one as the `auto_attach` option was removed

tests/foreman/maintain/test_clone.py
```diff
@@ -0,0 +1,190 @@
...
+from robottelo import constants
+from robottelo.config import settings
+from robottelo.hosts import Satellite
+
+SSH_PASS = settings.server.ssh_password
+
+
+@pytest.mark.parametrize("sat_ready_rhel", [8], indirect=True)
+@pytest.mark.parametrize('backup_type', ['online', 'offline'])
+def test_positive_clone_backup(target_sat, sat_ready_rhel, backup_type):
```
ogajduse, line 31: Can you merge the two tests into one? There is only a small difference between them which can be put into an `if` conditional block.

ogajduse (changes requested):

tests/foreman/maintain/test_clone.py
```diff
@@ -0,0 +1,190 @@
+"""Tests for satellite-maintain clone
```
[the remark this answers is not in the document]
Gauravtalreja1: +1, It would be best to move these tests under `tests/foreman/destructive/test_sat_clone.py`, as satellite clone is in itself a separate tool and not related to satellite-maintain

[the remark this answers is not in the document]
ogajduse: @lpramuk Can you chime in here?

tests/foreman/maintain/test_clone.py
```diff
@@ -0,0 +1,190 @@
...
+        options={'assumeyes': True, 'plaintext': True, 'skip-pulp-content': True},
+    )
+    assert no_pulp_backup_result.status == 0
+    sat_backup_dir = no_pulp_backup_result.stdout.strip().split()[-2]
+
+    # Copying satellite pulp data to target RHEL
+    assert sat_ready_rhel.execute('mkdir -p /var/lib/pulp').status == 0
+    assert (
+        target_sat.execute(
+            f'''sshpass -p "{SSH_PASS}" scp -o StrictHostKeyChecking=no \
```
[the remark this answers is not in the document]
JacobCallahan: correct. this is what remote_copy is for

tests/foreman/maintain/test_clone.py
```diff
@@ -0,0 +1,190 @@
...
+    )
+
+    # RHEL PART - TARGET SERVER
+    assert sat_ready_rhel.execute('ls /backup/config_files.tar.gz').status == 0
+    # Registering to cdn
+    sat_ready_rhel.register_to_cdn()
+    # Disabling repositories
+    assert sat_ready_rhel.execute('subscription-manager repos --disable=*').status == 0
+    # Getting satellite maintenance repo
+    res = requests.get(
```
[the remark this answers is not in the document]
JacobCallahan: would `target_sat.dogfood_repofile` not work in this case?

tests/foreman/maintain/test_clone.py
JacobCallahan: correct. dogfood_repofile is an attribute of each host object. you can instead get the repofile attribute from a satellite, then manually download it

tests/foreman/maintain/test_clone.py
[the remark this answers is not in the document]
ogajduse: It would not as `sat_ready_rhel` is `ContentHost`, not `Satellite` object.
If `dogfood_repofile` is used on ContentHost, then the client repo is defined instead of the Satellite repo.

tests/foreman/maintain/test_clone.py
ogajduse: > correct. dogfood_repofile is an attribute of each host object. you can instead get the repofile attribute from a satellite, then manually download it

This also seems reasonable. But what if there is a test that installs Satellite on a ContentHost without the need of target_sat? Yes, target_sat is always spined up during the pytest session, but I can imagine such a test case for PIT where we want to test Satellite installation using robottelo, not AAP.

tests/foreman/maintain/test_clone.py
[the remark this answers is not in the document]
ogajduse: > correct. this is what remote_copy is for

@JacobCallahan correct, but note that he is using `-r` parameter. How would you suggest using `remote_copy` in case we need to copy multiple files?

LadislavVasina1:
Adding pytest output so you can clearly see test parametrization
```collected 4 items                                                                                                                                                                                              

<Package destructive>
  <Module test_clone.py>
    <Function test_positive_clone_backup[skip_pulp-online-8]>
    <Function test_positive_clone_backup[skip_pulp-offline-8]>
    <Function test_positive_clone_backup[no_skip_pulp-online-8]>
    <Function test_positive_clone_backup[no_skip_pulp-offline-8]>

**Review by ogajduse — COMMENTED** [review]
I am OK with the current state of the PR, but I would like to get more feedback on my comments here. I will hold my ACK until the majority of them are answered.

File: tests/foreman/maintain/test_clone.py (hunk 16 above)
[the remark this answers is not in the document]
Comment (lpramuk) [reply tests/foreman/maintain/test_clone.py]: `satellite-clone` is maintenance tool coming from maintenance repo 
-> this speaks for having it in `tests/foreman/maintain`

but `satellite-clone` is separate from `satellite-maintain` tool
-> this speaks for having it in another location

So this depends how we define tests under `tests/foreman/maintain` - specific single tool   or   generally maintenance dedicated tests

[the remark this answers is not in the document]
Comment (LadislavVasina1) [reply tests/foreman/maintain/test_clone.py]: @JacobCallahan @Gauravtalreja1 What is your opinion on test placement after Lukas's insight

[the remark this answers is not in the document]
Comment (lpramuk) [reply tests/foreman/maintain/test_clone.py]: or there is yet `tests/foreman/longrun` :smiling_imp:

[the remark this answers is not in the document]
Comment (LadislavVasina1) [reply tests/foreman/maintain/test_clone.py]: After the JFT discussion, I am choosing option 2. `tests/foremain/destructive/test_clone.py` for storing my test.

File: tests/foreman/maintain/test_clone.py (hunk 17 above)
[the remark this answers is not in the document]
Comment (sambible) [reply tests/foreman/maintain/test_clone.py]: Would this be a case for adding a remote_copy_dir function to broker? Seems like something that could be useful in other tests as well.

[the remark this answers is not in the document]
Comment (LadislavVasina1) [reply tests/foreman/maintain/test_clone.py]: @sambible I had the same thought so I have already opened an issue:
https://github.com/SatelliteQE/broker/issues/165

File: tests/foreman/maintain/test_clone.py (hunk 2 above)
Comment (omkarkhatavkar) [reply tests/foreman/maintain/test_clone.py]: @JacobCallahan @mshriver I feel we are making this tightly coupled, even if you call ` Satellite(rhel_vm.hostname).download_repofile()` there is no guarantee that we could get a valid one. 

I'm my case https://github.com/SatelliteQE/robottelo/pull/10033/files#diff-6bf5cc432e0611da570dd30c289a1cfca53618276a13f7c1a04f2981b3bc0708R110 even I wanted to get the satellite repos in the RHEL machine as test verify custom cert satellite installation. This generates the incorrect curl URL as `http://<host>/api/releases/upstream/el8/satellite/repo_file` 

If you see the above URL `upstream` is dynamically generated based on the satellite.version property. I will have to fix this using kwargs now in satellite class. Would it be good to have a generic method and provide more power tests to download the repo?

Comment (Gauravtalreja1) [reply tests/foreman/maintain/test_clone.py]: +1, I think we should make this helper func more generic to download repos based on product arg passed, and not keep this tightly coupled with Satellite object.
With introduction to new repo API's in ohsnap, we do've more product options like satellite, capsule, client, and utils, so I feel it would be good to handle those products too in this helper, not something blocking this PR 

@JacobCallahan @mshriver any thoughts?

File: tests/foreman/maintain/test_clone.py (hunk 17 above)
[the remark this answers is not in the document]
Comment (Gauravtalreja1) [reply tests/foreman/maintain/test_clone.py]: @LadislavVasina1 Thanks for creating a issue, I think we can replace all sshpass related implementation  with remote_copy in separate PR later, resolving a thread

File: tests/foreman/maintain/test_clone.py (hunk 2 above)
Comment (JacobCallahan) [reply tests/foreman/maintain/test_clone.py]: @omkarkhatavkar @Gauravtalreja1 i'm certainly open to the property becoming a method, providing more control.

File: tests/foreman/maintain/test_clone.py (hunk 18 above)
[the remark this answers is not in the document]
Comment (JacobCallahan) [reply tests/foreman/maintain/test_clone.py]: The suggestion was more to use the target_sat to get the correct repofile URL. However, a previous comment expressed a need for even greater control. I have agreed to moving this to a method.

File: tests/foreman/maintain/test_clone.py (hunk 2 above)
Comment (ogajduse) [reply tests/foreman/maintain/test_clone.py]: > @omkarkhatavkar @Gauravtalreja1 i'm certainly open to the property becoming a method, providing more control.

@JacobCallahan Is that something that you would like to work on as an extension of the work that you have done in #9849?

File: tests/foreman/maintain/test_clone.py (hunk 18 above)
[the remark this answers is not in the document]
Comment (ogajduse) [reply tests/foreman/maintain/test_clone.py]: Resolving this thread, let's discuss it in https://github.com/SatelliteQE/robottelo/pull/9986#discussion_r952389184

**Review by ogajduse — APPROVED** [review]
ACK
I do not want to block this PR from merging anymore. However, we should implement the suggestions about `dogfood_repofile` presented in the review comments.

File: tests/foreman/destructive/test_clone.py [hunk 19]
```diff
@@ -0,0 +1,117 @@
...
+    # RHEL PART - TARGET SERVER
+    assert sat_ready_rhel.execute('ls /backup/config_files.tar.gz').status == 0
+    # Registering to cdn
+    sat_ready_rhel.register_to_cdn()
+    # Disabling repositories
+    assert sat_ready_rhel.execute('subscription-manager repos --disable=*').status == 0
+    # Getting satellite maintenace repo
+    res = requests.get(
+        f'{settings.repos.ohsnap_repo_host}/api/releases/{target_sat.version}/'
+        f'el{rhel_version}/satellite/repo_file'
```
lpramuk, line 95: Without snaps this won't work !

These URLs:
```
http://ohsnap/api/releases/6.11.0/el7/satellite/repo_file
http://ohsnap/api/releases/6.11.1/el7/satellite/repo_file
http://ohsnap/api/releases/6.11.2/el7/satellite/repo_file
http://ohsnap/api/releases/6.11.3/el7/satellite/repo_file
```
returns the same repofile pointing to latest 6.11.z  
```
http://dogfood/pulp/repos/Sat6-CI/QA/Satellite_6_11_with_RHEL7_Server/custom/Satellite_6_11_Composes/Satellite_6_11_RHEL7/
```

lpramuk: I would be for using `settings.repos.satmaintenace_repo` and you can then replace `s/RHEL7/RHEL8/` to get URL for other RHEL version

lpramuk (changes requested):
Change the way maintenance repository is set up

tests/foreman/destructive/test_clone.py
lpramuk: Or make developers to fix this behavior

tests/foreman/destructive/test_clone.py
```diff
@@ -0,0 +1,111 @@
...
+from robottelo.config import settings
+from robottelo.hosts import Satellite
+
+SSH_PASS = settings.server.ssh_password
+pytestmark = pytest.mark.destructive
+
+
+@pytest.mark.parametrize("sat_ready_rhel", [8], indirect=True)
+@pytest.mark.parametrize('backup_type', ['online', 'offline'])
+@pytest.mark.parametrize('skip_pulp_content', [True, False], ids=['skip_pulp', 'no_skip_pulp'])
```
Gauravtalreja1, line 31:
```suggestion
@pytest.mark.parametrize('skip_pulp', [False, True], ids=['include_pulp', 'skip_pulp'])
```

tests/foreman/destructive/test_clone.py
```diff
@@ -0,0 +1,111 @@
...
+    # Enabling and starting services
+    assert target_sat.cli.Service.enable().status == 0
+    assert target_sat.cli.Service.start().status == 0
+
+    # Doing backup
+    backup_result = target_sat.cli.Backup.run_backup(
+        backup_dir='/var/backup',
+        backup_type=backup_type,
+        options={'assumeyes': True, 'plaintext': True, 'skip-pulp-content': skip_pulp_content},
+    )
```
Gauravtalreja1, line 61: Are you planning to include any scenario for restoring pulp content? like sync a few repos and verify they exist after clone/restore, will suggest checking `tests/foreman/maintain/test_backup_restore.py::test_positive_backup_restore`

tests/foreman/destructive/test_clone.py
```diff
@@ -0,0 +1,111 @@
...
+    sat_ready_rhel.register_to_cdn()
+    # Disabling repositories
+    assert sat_ready_rhel.execute('subscription-manager repos --disable=*').status == 0
+    # Getting satellite maintenace repo
+    sat_ready_rhel.download_repofile(product='satellite', release=sat_version)
+    # Enabling repositories
+    for repo in getattr(constants, f"OHSNAP_RHEL{rhel_version}_REPOS"):
+        sat_ready_rhel.enable_repo(repo, force=True)
+    # Enabling satellite module
+    assert sat_ready_rhel.execute(f'dnf module enable -y satellite:el{rhel_version}').status == 0
```
Gauravtalreja1, line 98: It would be better to move these steps to the `sat_ready_rhel` fixture where it has all required specs and repos configured, what do you think?

tests/foreman/destructive/test_clone.py
LadislavVasina1: In the future probably yes but not now.

tests/foreman/destructive/test_clone.py
LadislavVasina1: I think that this fixture should be generic so it can also be used in other tests. This also means that I personally would prefer to leave this RHEL setup in the test, because i.e. sometimes you could use this fixture to install a satellite or sometimes you would want to use this just to clone a satellite that requires different repos.

ogajduse: I would definitely not put the setup into the `sat_ready_rhel` fixture. Instead, I would add a new fixture/set of fixtures that are dependent on `sat_ready_rhel`, to perform this setup.
However, I see this as a TODO for future satellite-clone test expansions.

tests/foreman/destructive/test_clone.py
ogajduse: I will resolve this one as users are now [able to specify the snap number](https://github.com/SatelliteQE/robottelo/pull/10079/files#diff-8c51f6ed7f99859d1efd6f74f29cc8e05e54bca2389c18e7e52c0203c1cbee6bR43) while requesting the repofile from Ohsnap.

See satellite/ohsnap/-/issues/30 on our internal GitLab.

tests/foreman/destructive/test_clone.py
```diff
@@ -57,12 +57,12 @@ def test_positive_clone_backup(target_sat, sat_ready_rhel, backup_type, skip_pul
     backup_result = target_sat.cli.Backup.run_backup(
         backup_dir='/var/backup',
         backup_type=backup_type,
-        options={'assumeyes': True, 'plaintext': True, 'skip-pulp-content': skip_pulp_content},
+        options={'assumeyes': True, 'plaintext': True, 'skip-pulp': skip_pulp},
```
Gauravtalreja1, line 60: `skip-pulp-content` option wouldn't change in this case 
```suggestion
        options={'assumeyes': True, 'plaintext': True, 'skip-pulp-content': skip_pulp},
```

LadislavVasina1: True, my bad.

tests/foreman/destructive/test_clone.py
Gauravtalreja1: Yes, you're right, I actually meant to include this steps as part of fixture, and not particularly `sat_ready_rhel` fixture

PR #9987, opened 2022-08-23
Move Http Proxy api tests from test_repository.py to test_http_proxy.py

jameerpathan111:
This PR moves tests present in `tests/foreman/api/test_repository.py` to `tests/foreman/api/test_http_proxy.py`.
It also parameterise the test to run tests for `no_http_proxy`, `auth_http_proxy`, `unauth_http_proxy`.

I'll be adding a few more tests here so this is one of the first pr.

Test result:
```
pytest tests/foreman/api/test_http_proxy.py         
collected 3 items                                                                                                                                                             

tests/foreman/api/test_http_proxy.py ...                                                                                                                                [100%]
================================================================= 3 passed, 68 warnings in 219.80s (0:03:39) ==================================================================
```

### DIFF
--- a/tests/foreman/api/test_http_proxy.py
+++ b/tests/foreman/api/test_http_proxy.py
@@ -16,96 +16,30 @@
 
 :Upstream: No
 """
-import re
-import tempfile
-import time
-from string import punctuation
-from urllib.parse import urljoin
-from urllib.parse import urlparse
-from urllib.parse import urlunparse
-
 import pytest
 from fauxfactory import gen_string
-from nailgun import client
-from nailgun import entities
-from nailgun.entity_mixins import TaskFailedError
-from requests.exceptions import HTTPError
 
 from robottelo import constants
-from robottelo import datafactory
-from robottelo import manifests
-from robottelo.api.utils import call_entity_method_with_timeout
 from robottelo.api.utils import enable_rhrepo_and_fetchid
-from robottelo.api.utils import promote
-from robottelo.api.utils import upload_manifest
 from robottelo.config import settings
-from robottelo.constants import DataFile
-from robottelo.constants import repos as repo_constants
-from robottelo.datafactory import parametrized
-from robottelo.logging import logger
 
 
-@pytest.fixture
-def repo_options(request, module_org, module_product):
-    """Return the options that were passed as indirect parameters."""
-    options = getattr(request, 'param', {}).copy()
-    options['organization'] = module_org
-    options['product'] = module_product
-    return options
-
-
-@pytest.fixture
-def repo_options_custom_product(request, module_org):
-    """Return the options that were passed as indirect parameters."""
-    options = getattr(request, 'param', {}).copy()
-    options['organization'] = module_org
-    options['product'] = entities.Product(organization=module_org).create()
-    return options
-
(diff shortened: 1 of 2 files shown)


peterdragun (approved):

jameerpathan111:
rebased the pr.

tests/foreman/api/test_http_proxy.py
```diff
@@ -0,0 +1,142 @@
...
+def create_http_proxy(sat, org, proxy_type):
+    """
+    Creates HTTP proxy.
+
+    :param str sat: Satellite to use.
+    :param str org: Organization
+    :param str proxy_type: 'auth_http_proxy' or 'unauth_http_proxy' http proxy.
+    """
+    if proxy_type == 'unauth_http_proxy':
+        return sat.api.HTTPProxy(
```
mshriver, line 36: :+1: thanks for using `sat.api` here instead of `entities` import.

tests/foreman/api/test_http_proxy.py
```diff
@@ -0,0 +1,142 @@
...
+    yield http_proxy, request.param
+    target_sat.update_setting('content_default_http_proxy', content_proxy_value)
+    target_sat.update_setting('http_proxy', general_proxy_value)
+
+
+@pytest.mark.tier2
+@pytest.mark.upgrade
+@pytest.mark.run_in_one_thread
+@pytest.mark.parametrize(
+    'function_http_proxy', ['no_http_proxy', 'auth_http_proxy', 'unauth_http_proxy'], indirect=True
```
mshriver, line 76: Consider parametrizing `None`, `True`, and `False` here, and use these strings as the IDs for the parameters.  It could make your logic checks easier to follow, if the parameter is something like `proxy_with_auth`.  Arguable, but string comparisons always feel weak/clunky to me. 

Example of two calls from the fixture:
```
proxy_type = request.param
content_proxy_value = target_sat.update_setting(
    'content_default_http_proxy', http_proxy.name if proxy_type is not None else ''
)
general_proxy_value = target_sat.update_setting(
    'http_proxy', general_proxy if proxy_type is not None else ''
)
```

Example from a line below:
```
http_proxy_policy = 'use_selected_http_proxy' if proxy_with_auth is not None else 'none'
```

tests/foreman/api/test_http_proxy.py
```diff
@@ -0,0 +1,142 @@
...
+            username=settings.http_proxy.username,
+            password=settings.http_proxy.password,
+            organization=[org.id],
+        ).create()
+
+
+@pytest.fixture(scope='function')
+def function_http_proxy(request, module_manifest_org, target_sat):
+    """Create a new HTTP proxy and set related settings based on proxy"""
+    http_proxy = create_http_proxy(target_sat, module_manifest_org, request.param)
```
mshriver, line 54: Since the calls are the same, just with different kwargs, *consider* not having a separate function for the 'create()` call, and just do it here. Create a kwargs dict with if/else blocks, and then expand those kwargs into a single call to `sat.api.HTTPProxy.create()`

mshriver (approved):
Great work combining the test cases into an end-to-end scenario!

PR #9994, opened 2022-08-24
rex cli slight refactor

### LINKED ISSUE #9985: some refactoring for the rex cli module
came up within the review for #9920

- create a function for the assertion pattern:

```
[tests/foreman/cli/test_remoteexecution.py](https://github.com/SatelliteQE/robottelo/pull/9920/files/0c1c30a768189c076b5409c130880edc641bb4ea#diff-6947be6fc255c8920f52b3f0e14ffb8da4c42a1a84c575581b34fc3a168ca6dc)
        result = JobInvocation.info({'id': invocation_command['id']})
        try:
            assert result['success'] == '1'
        except AssertionError:
```
[mshriver] This is done about 17 times in this module, might be time to write a function to cover the assertion pattern. Optional for this PR.

- change the assertion itself
```
[tests/foreman/cli/test_remoteexecution.py](https://github.com/SatelliteQE/robottelo/pull/9920/files/0c1c30a768189c076b5409c130880edc641bb4ea#diff-6947be6fc255c8920f52b3f0e14ffb8da4c42a1a84c575581b34fc3a168ca6dc)
        try:
            assert result['success'] == '1'
        except AssertionError:
            result = 'host output: {}'.format(
```
[mshriver] golfing a bit here, but no need to label result here, just call AssertionError and define the string in it's call parenthesis.

pondrejk:
Closes #9985

Removed an unused fixture and moved a heavily used assertion pattern to a function

### DIFF
--- a/tests/foreman/cli/test_remoteexecution.py
+++ b/tests/foreman/cli/test_remoteexecution.py
@@ -51,24 +51,6 @@ from robottelo.logging import logger
 
 
 @pytest.fixture()
-def fixture_vmsetup(request, module_org, target_sat):
-    """Create VM and register content host"""
-    if '_count' in request.param.keys():
-        with Broker(
-            nick=request.param['nick'],
-            host_classes={'host': ContentHost},
-            _count=request.param['_count'],
-        ) as clients:
-            for client in clients:
-                client.configure_rex(satellite=target_sat, org=module_org)
-            yield clients
-    else:
-        with Broker(nick=request.param['nick'], host_classes={'host': ContentHost}) as client:
-            client.configure_rex(satellite=target_sat, org=module_org)
-            yield client
-
-
-@pytest.fixture()
 def fixture_sca_vmsetup(request, module_gt_manifest_org, target_sat):
     """Create VM and register content host to Simple Content Access organization"""
     if '_count' in request.param.keys():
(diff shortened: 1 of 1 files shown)


pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k test_positive_run_default_job_template_by_ip

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -374,20 +317,8 @@ def test_positive_run_recurring_job_with_max_iterations_by_ip(self, rex_contenth
                 'max-iteration': 2,  # just two runs
             }
         )
-
         result = JobInvocation.info({'id': invocation_command['id']})
-        try:
-            assert result['status'] == 'queued'
```
omkarkhatavkar, line 380: Do you think that `queued` status you need to pass common function ? As you are only checking the status as `1` now.

omkarkhatavkar (approved):

tests/foreman/cli/test_remoteexecution.py
```diff
@@ -100,6 +82,24 @@ def infra_host(request, target_sat, module_capsule_configured):
     yield infra_hosts[request.param]
 
 
+def assert_job_invocation(invocation_command_id, client_hostname, expected_result='success'):
```
omkarkhatavkar, line 85: you can keep expected_result='1' and `assert result['success'] == expected_result` no need for if else block.

pondrejk: that would be nice, but the annoyance here is that I need to check `result['stauts']` instead of `result['success']`

Gauravtalreja1 (approved):

tests/foreman/cli/test_remoteexecution.py
mshriver: I think you're right you'll still need an if block here, but there's a bit of maybe mis-naming going on?

`expected_result` is the variable, but is this really the expected result, or is it the expected status?

Is `status` still set once `success` is set? (leading question, can you assert status still if success is expected?)

The kwarg is a string, but only has two expected states.  It can't be called with an expected status of 'failed', or 'in progress', or anything else - so maybe it should just be a boolean `assert_success` kwarg?  Seems ugly to take `field` and `status` as separate kargs for the caller to indicate which field and what state.

mshriver (approved):

tests/foreman/cli/test_remoteexecution.py
mshriver: Nice! Yeah this cleans up the tests very well and makes for clear separation of intent in what's being called.

PR #10007, opened 2022-08-31
Avoid adding capsule to the http proxy except list

vsedmik:
This was used earlier to bypass the proxying of sat-to-caps communication which was not working properly.
Currently BZ#1984400 solves the issue so the workaround is not needed anymore (even if we kept using of http proxy in our
current automation).

Plus pushed a test skipping until a BZ is resolved.

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -1353,14 +1353,6 @@ class Capsule(ContentHost, CapsuleMixins):
         if result.status:
             raise CapsuleHostError(f'The satellite-capsule package was not found\n{result.stdout}')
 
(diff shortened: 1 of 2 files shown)


vsedmik:
trigger: test-robottelo
pytest: tests/foreman/api/test_contentmanagement.py::TestCapsuleContentManagement

mshriver (approved):

synkd (approved):

latran (approved):

ColeHiggins2 (approved):

Gauravtalreja1 (approved):

PR #10010, opened 2022-09-03
Replace entities with sat object

shweta83:
Replace entities with sat object.

### DIFF
--- a/tests/upgrades/test_activation_key.py
+++ b/tests/upgrades/test_activation_key.py
@@ -94,7 +94,7 @@ class TestActivationKey:
             3. Delete activation key.
 
         :expectedresults: Activation key's entities should be same after upgrade and activation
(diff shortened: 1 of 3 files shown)


tests/upgrades/test_activation_key.py
```diff
@@ -93,28 +94,28 @@ def test_post_crud_activation_key(self, dependent_scenario_name):
             3. Delete activation key.
 
         :expectedresults: Activation key's entities should be same after upgrade and activation
-        key update and delete should work.
+        key update and delettarge should work.
```
peterdragun, line 97:
```suggestion
        key update and delete should work.
```

peterdragun (approved):
ACK, with a minor comment about docstring updated by accident

omkarkhatavkar (approved):
ACK

PR #10011, opened 2022-09-05
Add rhel repos before installing cockpit

peterdragun:
- Depends on https://github.com/SatelliteQE/airgun/pull/743

### DIFF
--- a/pytest_fixtures/component/subscription.py
+++ b/pytest_fixtures/component/subscription.py
@@ -25,6 +25,19 @@ def module_subscribe_satellite(module_target_sat):
     module_target_sat.unregister()
 
 
+@pytest.fixture(scope='class')
+def class_subscribe_satellite(class_target_sat):
+    """Subscribe Satellite to CDN"""
+    class_target_sat.register_to_cdn()
+    # Enable extras repo if os_version is RHEL7
+    if class_target_sat.os_version.major < 8:
+        class_target_sat.enable_repo(
+            f'rhel-{class_target_sat.os_version.major}-server-extras-rpms', force=True
+        )
+    yield class_target_sat
+    class_target_sat.unregister()
+
+
 @pytest.fixture
 def subscribe_satellite(target_sat):
     """Subscribe Satellite to CDN"""
--- a/pytest_fixtures/core/sys.py
+++ b/pytest_fixtures/core/sys.py
@@ -52,19 +52,9 @@ def puppet_proxy_port_range(session_puppet_enabled_sat):
 
 
 @pytest.fixture(scope='class')
-def install_cockpit_plugin(class_target_sat):
-    class_target_sat.download_repofile()
-    class_target_sat.install_cockpit()
-    # TODO remove this change when we start using new host detail view
-    setting_object = class_target_sat.api.Setting().search(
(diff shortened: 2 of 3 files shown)


peterdragun:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_host.py -k test_positive_cockpit
airgun: 743

peterdragun:
trigger: test-robottelo
pytest: -v tests/foreman/destructive/test_host.py -k test_positive_cockpit[rhel9_fips]
airgun: 743

pytest_fixtures/core/sys.py
```diff
@@ -53,18 +53,10 @@ def puppet_proxy_port_range(session_puppet_enabled_sat):
 
 @pytest.fixture(scope='class')
 def install_cockpit_plugin(class_target_sat):
-    class_target_sat.download_repofile()
+    class_target_sat.create_custom_repos(
+        **settings.repos[f'rhel{class_target_sat.os_version.major}_os']
+    )
```
Gauravtalreja1, line 58: It would have been better if you registered with CDN and enabled OHSNAP_RHEL8_REPOS instead of setting up custom RHEL repos

peterdragun:
trigger: test-robottelo
pytest: -v tests/foreman/destructive/test_host.py -k test_positive_cockpit
airgun: 743

pytest_fixtures/component/subscription.py
```diff
@@ -25,6 +25,19 @@ def module_subscribe_satellite(module_target_sat):
...
+def class_subscribe_satellite(class_target_sat):
+    """Subscribe Satellite to CDN"""
+    class_target_sat.register_to_cdn()
+    # Enable extras repo if os_version is RHEL7
+    if class_target_sat.os_version.major < 8:
+        class_target_sat.enable_repo(
+            f'rhel-{class_target_sat.os_version.major}-server-extras-rpms', force=True
+        )
+    yield class_target_sat
+    class_target_sat.unregister()
```
Gauravtalreja1, line 38: We can avoid enabling the El7 repo here since 6.12 is the default on El8, and class_target_sat already has the teardown that unregisters a system, so this fixture seems redundant here for one register step that can be included under class_cockpit_sat.

peterdragun: This was just a copy-paste of other fixtures in this module with changing the scope, as I need a class-scoped fixture. We can definitely drop support for EL7 in the master branch, but that would mean inconsistency in this file and also manual cherry-picking of this change to 6.11 (which is not terrible, that is just a reminder for me).
Personally, I would prefer this to merge as it is and then create separate PR for removing support for EL7, but if you think that makes no sense I can remove it now.

pondrejk (approved):

pytest_fixtures/component/subscription.py
Gauravtalreja1: @peterdragun I agree with you and I leave it up to you how to handle. As subscribe_satellite fixtures just registers (which is done by register_to_cdn) and unregisters (which is done by teardown of broker fixtures), I don't think it's necessary at all :)

Gauravtalreja1 (approved):

PR #10015, opened 2022-09-05
Auto CherryPicking to n-1 only

jyejare:
As discussed in the poll taken in automation channel , we are sticking to auto cherry picking to n-1 by default.

- If PR further needs to auto cherrypick then the auto-cherrypicked PR should contain/add label `CherryPick`.
- If not needed in further branches, then `NoCherryPick` label should be added.

### DIFF
--- a/.github/workflows/auto_cherry_pick.yml
+++ b/.github/workflows/auto_cherry_pick.yml
@@ -7,37 +7,34 @@ on:
       - closed
 
 jobs:
-  branch-matrix:
+  previous-branch:
     if: github.event.pull_request.merged == true && contains(github.event.pull_request.labels.*.name, 'CherryPick')
-    name: Generate a branch matrix to apply cherrypicks
+    name: Calculate previous branch name
     runs-on: ubuntu-latest
     outputs:
-      branches: ${{ steps.set-matrix.outputs.branches }}
+      previous_branch: ${{ steps.set-branch.outputs.previous_branch }}
     steps:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - id: set-matrix
-        run: echo "::set-output name=branches::$(git branch -rl --sort=-authordate 'origin/6.*.z' --format='%(refname:lstrip=-1)' | head -n2 | jq -cnR '[inputs | select(length>0)]')"
-  auto_cherry_picking:
+      - id: set-branch
+        run: echo "::set-output name=previous_branch::$(if [ $GITHUB_BASE_REF  == 'master' ]; then echo $(git branch -rl --sort=-authordate 'origin/6.*.z' --format='%(refname:lstrip=-1)' | head -n1 | jq -cnR '[inputs | select(length>0)]'); else echo ['"6.'$(($(echo $GITHUB_BASE_REF | cut -d. -f2) - 1))'.z"']; fi)"
+
+  auto-cherry-pick:
+    name: Auto Cherry Pick to previous branch
     if: github.event.pull_request.merged == true && contains(github.event.pull_request.labels.*.name, 'CherryPick')
-    name: Auto Cherry picking to branches
-    needs: branch-matrix
+    needs: previous-branch
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        to_branch: ${{ fromJson(needs.branch-matrix.outputs.branches) }}
+        to_branch: ${{ fromJson(needs.previous-branch.outputs.previous_branch) }}
     steps:
-      - name: Checkout Robottelo
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-        if: matrix.to_branch != github.base_ref
       - name: Cherry pick into ${{ matrix.to_branch }}
         uses: jyejare/github-cherry-pick-action@main
         with:
           branch: ${{ matrix.to_branch }}
           labels: |
             Auto_Cherry_Picked
-        # skipping PRs remote target_branch from cherrypicking into itself
-        if: matrix.to_branch != github.base_ref


.github/workflows/auto_cherry_pick.yml
```diff
@@ -7,37 +7,18 @@ on:
...
         with:
           fetch-depth: 0
-        if: matrix.to_branch != github.base_ref
-      - name: Cherry pick into ${{ matrix.to_branch }}
-        uses: jyejare/github-cherry-pick-action@main
+      - name: Cherry pick into 6.11.z
+        uses: jyejare/github-cherry-pick-action@1.0.4
         with:
-          branch: ${{ matrix.to_branch }}
+          branch: 6.11.z
```
peterdragun, line 22: Is there any reason to hardcode the branch? It used to be dynamic, right?

.github/workflows/auto_cherry_pick.yml
```diff
@@ -7,37 +7,18 @@ on:
...
     steps:
       - name: Checkout Robottelo
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
-        if: matrix.to_branch != github.base_ref
-      - name: Cherry pick into ${{ matrix.to_branch }}
-        uses: jyejare/github-cherry-pick-action@main
+      - name: Cherry pick into 6.11.z
+        uses: jyejare/github-cherry-pick-action@1.0.4
```
Gauravtalreja1, line 20: Any reason to hardcode this version here? It would be better to always use the `main` to pick up the latest code
```suggestion
        uses: jyejare/github-cherry-pick-action@main
```

.github/workflows/auto_cherry_pick.yml
```diff
@@ -7,37 +7,18 @@ on:
       - closed
 
 jobs:
-  branch-matrix:
-    if: github.event.pull_request.merged == true && contains(github.event.pull_request.labels.*.name, 'CherryPick')
```
Gauravtalreja1, line 11: Don't we need this condition now?

jyejare: We dont need branch matrix the condition still persists for autocherrypick job.

.github/workflows/auto_cherry_pick.yml
jyejare: We dont want to be on latest because we are not testing everything we are merging and hence tagging works best with the tested code.

.github/workflows/auto_cherry_pick.yml
jyejare: Yeah ! The branch matrix was dynamic earlier but since we are just going for n-1 we dont need complex calculation and then the complex conditioning for skipping cherrypick for self branch.

TBH there is no good algorithm for calculating n-1 branch name based on current branch and hence  we are sticking to the constant/hardcoded branch name!

latran (approved):

.github/workflows/auto_cherry_pick.yml
JacobCallahan: So when this is brought back to 6.11.z, we will hardcode this value to 6.10.z?

jyejare:
@JacobCallahan :

yes ! Very similar concept to other gha's where we are hardcoding and updating them while branching is a part of maintainance now.

JacobCallahan (approved):

.github/workflows/auto_cherry_pick.yml
omkarkhatavkar: @jyejare I think it's good approach to have dynamic code for selection version. Anyways upcoming 6.13 we might need the 6.12 and 6.11 then we again have to add this logic. I would rather keep this logic still present and change the `head -n 1` for now. Thanks adding earlier that dynamic code.

jyejare: Ohk ! Switching to dynamic way of calculating the branch version on demand !

jyejare: @omkarkhatavkar Sorry, but dynamic calculation was suitable when we had n-2 strategy where we were calculating to 2 last branches and then with master we were chrrypicking to latest 2 remote branches(unfortunately `git remote -r` always returns top branches only and not below any branch name), for n-1 we used to skip the top branch and only cherrypick to `n-2'nd`  branch and for n-2 we needed not to cherrypick.

This login doesn't work with n-1 as we always had to calculate which branch is the current one now and then calculate its previous branch from that. The calculations are way more complex than we keep it static and maintain it per branching time.

peterdragun: What about using something like this :thinking: 
```
if [ $(git rev-parse --abbrev-ref HEAD)  == 'master' ]; then $(git branch -rl --sort=-authordate 'origin/6.*.z' --format='%(refname:lstrip=-1)' | head -n2 | jq -cnR '[inputs | select(length>0)]'); else $(($(git rev-parse --abbrev-ref HEAD | cut -d. -f2) - 1)) fi
```
It looks like it is able to do - 1 without hardcoding branch. It is not the nicest line and I believe it could be done better, that was just me playing with bash for a moment.

jyejare: @peterdragun Nop! 

This wont work `git rev-parse --abbrev-ref HEAD)  == 'master'` because right side of this condition will always be the branch name in the PR.

I mean there are ways to do it but they are very complex. I can think of the following logic:
```
1. Find base_branch into which the PR is to be merged.
2. If its master then cherrypick to `6.*.z` topmost using previous math we used.
3. If not master, then do `6.(*-1).z` and find the previous branch.
```

But this is subject to testing and its not easy to test something with GHAs and this will hold us from cherrypicking until we reach to the correct result.

peterdragun: Oh, okay, haven't thought about this issue, but the second and the third points are already resolved by my code, so the only thing to figure out is the first one, find the branch that is the target of the current PR.
That could be probably resolved by: https://stackoverflow.com/questions/62331829/how-to-get-the-target-branch-of-the-github-pull-request-from-an-actions
I don't have much experience with GHA but I believe there must be a way to get name of that branch.

jyejare: @peterdragun And thats the only thing testing will take efforts :)

ogajduse: I have to second @peterdragun here. I find it very inappropriate to dismiss the solution just because `testing will take efforts`.
If the real issue here is that you personally do not have time to test it, let us know and Peter or I can test it and come back with results.

ogajduse (changes requested):
NACK
The target branch should not be hardcoded as long as we keep the branch names in the predictable format `6.X.z`.

.github/workflows/auto_cherry_pick.yml
peterdragun: Testing done.
See:
from master: https://github.com/peterdragun/robottelo/runs/8244910155?check_suite_focus=true
followup from 6.11.z: https://github.com/peterdragun/robottelo/actions/runs/3013576470
Note that the second GHA failed, but it was caused by autocherry pick failed and commit needs manual cherrypick

jyejare: @peterdragun The code is updated with the changes, please review and merge if all looks good !

ogajduse (approved):
Thank you @jyejare and @peterdragun for taking the time to test the dynamic solution.

jyejare:
Merging this with multiple approvals !

PR #10041, opened 2022-09-12
Change os version for fake host

peterdragun:


robottelo/host_helpers/cli_factory.py
```diff
@@ -443,7 +443,7 @@ def make_fake_host(self, options=None):
         if not options.get('operatingsystem') and not options.get('operatingsystem-id'):
             try:
                 options['operatingsystem-id'] = self._satellite.cli.OperatingSys.list(
-                    {'search': 'name="RedHat" AND (major="6" OR major="7")'}
+                    {'search': 'name="RedHat" AND (major="7" OR major="8")'}
```
Gauravtalreja1, line 446: This might also need to be updated here https://github.com/SatelliteQE/robottelo/blob/master/robottelo/cli/factory.py#L876

And please provide some test results

Gauravtalreja1 (changes requested):

pondrejk (approved):

robottelo/host_helpers/cli_factory.py
peterdragun: As there is already a catch for index error that change is not required, a new OS will be created and then the test setup will pass. But sure, we can update it so there is no need to create a new OS.

Test results will be added in a separate comment.

peterdragun:
trigger: test-robottelo
pytest: -v tests/foreman/cli/test_host.py -k test_positive_create_and_update_with_content_source

latran (approved):

Gauravtalreja1 (approved):

PR #10063, opened 2022-09-15
Minor changes to account for container-based hosts

JacobCallahan:
Hosts where katello-agent installation was desired were basing the installation check on whether or not goferd service was running. This new change explicitly runs it on containers.
Also, fixed a param issue involving rex hosts


spusateri (approved):
This seems like good workaround until we know how many customers are using katello-agent and implement something more permanent :+1:

swadeley (approved):
ACK

PR #10064, opened 2022-09-15
Turn container_contenthost back to VM

vsedmik:
Let's move the `container_contenthost` back to a VM as it does not seem legit to run docker inside a docker container.

Passes locally with 6.11.3. With 6.12.0 it fails due to a regression [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=2125244), for which a fix was already merged in upstream, so I don't `skip_if_open`. Also, I want to have it in 6.11.z regardless the BZ state.

### DIFF
--- a/tests/foreman/api/test_contentmanagement.py
+++ b/tests/foreman/api/test_contentmanagement.py
@@ -1114,6 +1114,8 @@ class TestCapsuleContentManagement:
         :parametrized: yes
 
         :CaseLevel: Integration
+
+        :BZ: 2125244
         """
         upstream_names = [
             'quay/busybox',  # schema 1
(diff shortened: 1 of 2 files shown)


tests/foreman/api/test_contentmanagement.py
```diff
@@ -1085,6 +1085,7 @@ def test_positive_sync_kickstart_repo(
         assert sat_pkgs == caps_pkgs
 
     @pytest.mark.tier4
+    @pytest.mark.no_containers
```
ogajduse, line 1088: This marker would not have any effect here.

ogajduse (approved):

JacobCallahan (approved):

PR #10033, opened 2022-09-09
fixed: rhsso tests for RHEL8 and certificate tests

omkarkhatavkar:
rhsso tests for RHEL8 and certificate tests

PR #10079, opened 2022-09-19
Give users greater control over dogfood_repofile

ogajduse:
This feature has been requested in the following discussions:
* https://github.com/SatelliteQE/robottelo/pull/9986#discussion_r961925131
* https://github.com/SatelliteQE/robottelo/pull/9986#discussion_r952389184
* https://github.com/SatelliteQE/robottelo/pull/10033/files#r968362516

### DIFF
--- a/robottelo/host_helpers/contenthost_mixins.py
+++ b/robottelo/host_helpers/contenthost_mixins.py
@@ -40,27 +40,33 @@ class VersionedContent:
             'cbrhel': constants.OSCAP_PROFILE[f'cbrhel{self._v_major}'],
         }
 
-    @cached_property
-    def dogfood_repofile(self):
+    def dogfood_repofile(self, sat_ver=None, product=None):
         from robottelo.config import settings
 
-        if self.__class__.__name__ == 'ContentHost':
-            product = 'tools'
-        else:
-            product = self.__class__.__name__.lower()
-        # this may be also available from self.satellite.version
-        split_ver = self.satellite.version.split('.')
-        if len(split_ver) == 2:
-            split_ver.append('0')
-        sat_ver = '.'.join(split_ver[:3])  # keep only major.minor.patch
+        if not product:
+            if self.__class__.__name__ == 'ContentHost':
+                product = 'tools'
+            else:
+                product = self.__class__.__name__.lower()
+        if not sat_ver:
+            # this may be also available from self.satellite.version
+            sat_ver = self.satellite.version
+        sat_ver = sat_ver.split('.')
+
+        if len(sat_ver) == 2:
+            sat_ver.append('0')
+        sat_ver = '.'.join(sat_ver[:3])  # keep only major.minor.patch
+        # TODO: make URL containing also snap number
         return (
             f'{settings.repos.ohsnap_repo_host}/api/releases/'
             f'{sat_ver}/el{self._v_major}/{product}/repo_file'
         )
 
-    def download_repofile(self):
+    def download_repofile(self, sat_ver=None, product=None):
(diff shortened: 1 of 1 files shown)


JacobCallahan (approved):
Nice!

peterdragun (approved):

JacobCallahan (approved):

omkarkhatavkar (approved):

robottelo/host_helpers/contenthost_mixins.py
```diff
@@ -40,27 +40,33 @@ def OSCAP(self):
...
-        else:
-            product = self.__class__.__name__.lower()
-        # this may be also available from self.satellite.version
-        split_ver = self.satellite.version.split('.')
-        if len(split_ver) == 2:
-            split_ver.append('0')
-        sat_ver = '.'.join(split_ver[:3])  # keep only major.minor.patch
+        if not product:
+            if self.__class__.__name__ == 'ContentHost':
+                product = 'tools'
```
Gauravtalreja1, line 48:
```suggestion
                product = 'client'
```

robottelo/host_helpers/contenthost_mixins.py
```diff
@@ -40,27 +40,33 @@ def OSCAP(self):
             'cbrhel': constants.OSCAP_PROFILE[f'cbrhel{self._v_major}'],
         }
 
-    @cached_property
-    def dogfood_repofile(self):
+    def dogfood_repofile(self, sat_ver=None, product=None, snap=''):
```
Gauravtalreja1, line 43: Non-blocking, just a thought, we should make the default product as `satellite` here, and make `sat_ver` and `snap_ver` as kwarg?

robottelo/host_helpers/contenthost_mixins.py
ogajduse: Good catch! Fixed in the last revision.

robottelo/host_helpers/contenthost_mixins.py
ogajduse: We rely on the host's class name like `Satellite`, `Capsule`, or `ContentHost` to distinguish the product automagically.
If we would specify a default for a product in the method header/definition, the whole conditional block in the method would be useless.
So either user explicitly specifies the product or the method will figure it out based on the host object.
I agree with converting args to kwargs to some degree, but we have only three args here. Is it worth/beneficial to use kwargs?

latran (approved):

robottelo/host_helpers/contenthost_mixins.py
omkarkhatavkar: Just thinking out loud here, do we going to have the satellite version installed on the satellite different from the setting object? 

If that is the case then I support the dynamic way of handling getting a satellite version. otherwise, it seems to be just overhead.

ogajduse: That is a good question for @lpramuk who aims to do upgrades with robottelo in the future.

lpramuk (approved):

robottelo/host_helpers/contenthost_mixins.py
lpramuk: @Gauravtalreja1 we won't be passing mysterious dicts to functions, it is upto to caller to expand the dict before calling function.
```
d = {'sat_ver': '6.12.0', 'snap_ver': '11.0' }
obj.dogfood_repofile(**d)     <- defined as dogfood_repofile(self, sat_ver=None, snap_ver=''):
```
so definitely not  this !
```
obj.dogfood_repofile(d)      <- defined as dogfood_repofile(self, ver_kwargs=None):
```

PR #10110, opened 2022-09-26
Remove entities call for target_sat.api call for LCE

latran:
Replacing all entities call with `target_sat.api` calls now.

### DIFF
--- a/tests/foreman/ui/test_lifecycleenvironment.py
+++ b/tests/foreman/ui/test_lifecycleenvironment.py
@@ -18,7 +18,6 @@
 """
 import pytest
 from airgun.session import Session
-from nailgun import entities
 from navmazing import NavigationTriesExceeded
 
 from robottelo.api.utils import create_role_permissions
@@ -31,11 +30,6 @@ from robottelo.constants import FAKE_1_CUSTOM_PACKAGE_NAME
 from robottelo.constants import FAKE_2_CUSTOM_PACKAGE
 from robottelo.constants import FAKE_3_CUSTOM_PACKAGE_NAME
 from robottelo.datafactory import gen_string
-
-
-@pytest.fixture(scope='module')
-def module_org():
-    return entities.Organization().create()
 
 
 @pytest.mark.upgrade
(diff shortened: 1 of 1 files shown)


shweta83 (approved):

peterdragun (approved):

tests/foreman/ui/test_lifecycleenvironment.py
```diff
@@ -287,8 +287,8 @@ def test_positive_custom_user_view_lce(session, test_name):
     lce_name = gen_string('alpha')
     user_login = gen_string('alpha')
     user_password = gen_string('alpha')
-    org = entities.Organization().create()
-    role = entities.Role(name=role_name).create()
+    org = target_sat.api.Organization().create()
```
synkd, line 290: Non-blocking, but does this test need to create an organization, or can it inherit `module_org` as well?

synkd (approved):

tests/foreman/ui/test_lifecycleenvironment.py
```diff
@@ -152,7 +148,7 @@ def test_positive_search_lce_content_view_packages_by_full_name(session, module_
 
 @pytest.mark.tier3
 @pytest.mark.skipif((not settings.robottelo.REPOS_HOSTING_URL), reason='Missing repos_hosting_url')
-def test_positive_search_lce_content_view_packages_by_name(session, module_org):
+def test_positive_search_lce_content_view_packages_by_name(session, module_org, target_sat):
```
Gauravtalreja1, line 151: Optional: Since target_sat is already included here, a `session` here can also be replaced by `target_sat.ui_session()`

tests/foreman/ui/test_lifecycleenvironment.py
latran: It can.  I may come back later remove all that, but not at this point.

tests/foreman/ui/test_lifecycleenvironment.py
jyejare: Good suggestion but I think thats out of the scope of this PR !

jyejare (approved):

PR #10117, opened 2022-09-28
Fix no_containers checks to run CHosts tests on VMs

jyejare:
The checks to find pytest option and marker for `no_containers` were misleading.

Pytest option to find no_containers option has a wrong lookup value.
AND
Presence of no_containers marker on test node was at a wrong attribute.

Fixed both !

### DIFF
--- a/pytest_fixtures/core/contenthosts.py
+++ b/pytest_fixtures/core/contenthosts.py
@@ -21,7 +21,13 @@ def host_conf(request):
     _rhelver = f"{distro}{params.get('rhel_version', settings.content_host.default_rhel_version)}"
     # check to see if no-containers is passed as an argument to pytest
     deploy_kwargs = {}
-    if not request.config.getoption('no-containers', None) and not params.get('no_containers'):
+    if not any(
+        [
+            request.config.getoption('no_containers'),
+            params.get('no_containers'),
+            request.node.get_closest_marker('no_containers'),
+        ]
+    ):
         deploy_kwargs = settings.content_host.get(_rhelver).get('container', {})
     # if we're not using containers or a container isn't available, use a VM
     if not deploy_kwargs:


pytest_fixtures/core/contenthosts.py
```diff
@@ -21,7 +21,7 @@ def host_conf(request):
     _rhelver = f"{distro}{params.get('rhel_version', settings.content_host.default_rhel_version)}"
     # check to see if no-containers is passed as an argument to pytest
     deploy_kwargs = {}
-    if not request.config.getoption('no-containers', None) and not params.get('no_containers'):
+    if not request.config.getoption('no_containers') and not request.node.get_closest_marker('no_containers'):
```
JacobCallahan, line 24: The params check needs to stay since other fixtures are using it to control whether or not to use containers. Of course, this line will need to be run through black.
```suggestion
    if not request.config.getoption('no_containers') and not params.get('no_containers') and not request.node.get_closest_marker('no_containers'):
```

JacobCallahan (changes requested):
Good find. This was a holdout from the design period where no_containers was controlled via parameterization.

pytest_fixtures/core/contenthosts.py
jyejare: Ohh! Make sense! Reinitiating the removed check !

pytest_fixtures/core/contenthosts.py
```diff
@@ -21,7 +21,9 @@ def host_conf(request):
     _rhelver = f"{distro}{params.get('rhel_version', settings.content_host.default_rhel_version)}"
     # check to see if no-containers is passed as an argument to pytest
     deploy_kwargs = {}
-    if not request.config.getoption('no-containers', None) and not params.get('no_containers'):
+    if not request.config.getoption('no_containers') and not request.node.get_closest_marker(
+        'no_containers'
+    ):
```
Griffin-Sullivan, line 26: @JacobCallahan Why use logical `and` here? Is the expectation for the user to specify the marker and the pytest option if they don't want to use containers? Maybe that selection process is also done somewhere else? Because I feel like we would want `no_containers` if either option or marker were present.

jyejare: Basically we are checking that the switch for spinning up host as containers **from all the possible places** is **ON** :)

I tested it and the condition is very valid !

Griffin-Sullivan: Thanks for clarifying! I wasn't exactly sure what the check was doing.

pytest_fixtures/core/contenthosts.py
jyejare: Fixed !

Griffin-Sullivan (approved):

JacobCallahan (approved):

PR #10118, opened 2022-09-28
[6.11.z] Fix no_containers checks to run CHosts tests on VMs

jyejare:
Cherrypick of commit: 70885f5577dfa2b0f5beadd2db683c490cf2bc98

The checks to find pytest option and marker for `no_containers` were misleading.

Pytest option to find no_containers option has a wrong lookup value.
AND
Presence of no_containers marker on test node was at a wrong attribute.

Fixed both !


swadeley (approved):
ACK

JacobCallahan (approved):

PR #10120, opened 2022-09-28
Remove system_facts

sambible:
Part of the helpers refactor.

This file isn't currently referenced anywhere in Robottelo, and looks to be a pretty old relic. Unless there is value in keeping the constants around, I don't really see any reason to not simply delete this file.

Thoughts?


JacobCallahan:
See if anything needs to be added to the SystemFacts class defined here: https://github.com/SatelliteQE/robottelo/blob/master/robottelo/host_helpers/contenthost_mixins.py#L93

shweta83 (approved):
I think we are good to remove this @sambible @JacobCallahan

Gauravtalreja1 (approved):

jyejare:
@sambible Check the suggestion from @JacobCallahan. Tts very valuable suggestion if the deletions from this PR can fit into what he said. If it doesn't let's continue.

Waiting for your ACK/NACK.

sambible:
The file being proposed for deletion is effectively a custom facts dictionary, with specifics set already. As it says
"""JSON representation for a RHEL server."""

Personally, I don't see any reason to maintain this specific collection of facts. It's not being used by any tests currently, and the code in there doesn't do anything we can't do with the mixin already (i.e. generate a facts file from a RHEL server)

jyejare (approved):

jyejare:
@JacobCallahan Are you good with @sambible comment, then please merge !

PR #10127, opened 2022-09-30
[6.11.z] Temporary workaround for new dynaconf issue

jyejare:
Cherrypick of commit: 884fb93798bc6610dbb10221f00f6fefd2f88b92

Dynaconf's box object's new behavior appears to make pop actions useless. This caused an issue with the way we update the deploy arguments. Our temporary workaround is to make a dictionary copy of the deploy arguments, update them, then assign that to the settings object's deploy arguments. More investigation into the dynaconf side is needed.


jyejare:
Merging this as its priority fix blocking robottelo tests on demand satellite checkout.

PR #10128, opened 2022-09-30
Fix override of deploy args settings

JacobCallahan:
The dynabox list objects used to store these deploy args have effectively become read only in newer versions of Dynaconf.
This change fixes the way we override these settings by using set instead of direct assignment.

### DIFF
--- a/pytest_fixtures/core/broker.py
+++ b/pytest_fixtures/core/broker.py
@@ -74,7 +74,7 @@ def satellite_factory():
     if settings.server.get('deploy_arguments'):
         logger.debug(f'Original deploy arguments for sat: {settings.server.deploy_arguments}')
         resolved = _resolve_deploy_args(settings.server.deploy_arguments)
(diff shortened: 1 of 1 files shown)


pytest_fixtures/core/broker.py
```diff
@@ -99,7 +99,7 @@ def capsule_factory():
     if settings.capsule.get('deploy_arguments'):
         logger.debug(f'Original deploy arguments for cap: {settings.capsule.deploy_arguments}')
         resolved = _resolve_deploy_args(settings.capsule.deploy_arguments)
-        settings.capsule.deploy_arguments = resolved
+        settings.set("capsule.deploy_arguments", resolved)
```
Gauravtalreja1, line 102:
```suggestion
        settings.set('capsule.deploy_arguments', resolved)
```

Gauravtalreja1 (approved):
Ack, with small suggestion

PR #10159, opened 2022-10-12
fixing KeyError in parametrized fixture

pondrejk:
I was using a different key to parametrize the infra_host fixture than in the dict within the function, leading to the KeyError


Gauravtalreja1:
trigger: test-robottelo
pytest: --uses-fixtures infra_host tests/foreman/cli/

Gauravtalreja1 (approved):
ACK, @pondrejk same key is used in 6.11.z's infra_host fixture, so you may need to add a cherrypick label.

mshriver (approved):

PR #10162, opened 2022-10-12
Convert manifest-dependent API tests to manifester

synkd:
This PR will convert all API tests that use manifests to obtain their manifests using manifester. I am planning on pushing one commit per module in an attempt to stay organized. The initial commit converts the `api/test_contentview.py` module. Note that, in this module, the `TestContentViewRedHatOstreeContent` class is decorated with `@pytest.mark.skip_if_open("BZ:1625783")`, but this BZ is closed WONTFIX. With the marker removed, these tests fail, but the failures are unrelated to the use of manifester.

### DIFF
--- a/tests/foreman/api/test_contentview.py
+++ b/tests/foreman/api/test_contentview.py
@@ -821,16 +821,12 @@ class TestContentViewRedHatContent:
     """Tests for publishing and promoting content views."""
 
     @pytest.fixture(scope='class', autouse=True)
-    def initiate_testclass(self, request, module_org, module_cv, module_entitlement_manifest):
+    def initiate_testclass(self, request, module_cv, module_entitlement_manifest_org):
         """Set up organization, product and repositories for tests."""
 
-        entities.Subscription().upload(
-            data={'organization_id': module_org.id},
-            files={'content': module_entitlement_manifest.content},
-        )
         repo_id = enable_rhrepo_and_fetchid(
             basearch='x86_64',
-            org_id=module_org.id,
+            org_id=module_entitlement_manifest_org.id,
             product=PRDS['rhel'],
             repo=REPOS['rhst7']['name'],
             reposet=REPOSET['rhst7'],
@@ -1378,16 +1374,12 @@ class TestContentViewRedHatOstreeContent:
 
     @pytest.mark.run_in_one_thread
     @pytest.fixture(scope='class', autouse=True)
-    def initiate_testclass(self, request, module_org, module_entitlement_manifest):
+    def initiate_testclass(self, request, module_entitlement_manifest_org):
         """Set up organization, product and repositories for tests."""
 
-        entities.Subscription().upload(
-            data={'organization_id': module_org.id},
-            files={'content': module_entitlement_manifest.content},
-        )
         repo_id = enable_rhrepo_and_fetchid(
             basearch=None,
-            org_id=module_org.id,
+            org_id=module_entitlement_manifest_org.id,
             product=PRDS['rhah'],
             repo=REPOS['rhaht']['name'],
             reposet=REPOSET['rhaht'],
--- a/tests/foreman/api/test_contentview.py
+++ b/tests/foreman/api/test_contentview.py
@@ -25,7 +25,6 @@ from fauxfactory import gen_utf8
 from nailgun import entities
 from requests.exceptions import HTTPError
 
-from robottelo import manifests
 from robottelo.api.utils import apply_package_filter
 from robottelo.api.utils import enable_rhrepo_and_fetchid
 from robottelo.config import settings
@@ -823,14 +822,13 @@ class TestContentViewRedHatContent:
 
     @pytest.mark.skip_if_not_set('fake_manifest')
     @pytest.fixture(scope='class', autouse=True)
-    def initiate_testclass(self, request, module_org, module_cv):
+    def initiate_testclass(self, request, module_org, module_cv, module_entitlement_manifest):
         """Set up organization, product and repositories for tests."""
 
-        with manifests.clone() as manifest:
-            entities.Subscription().upload(
-                data={'organization_id': module_org.id},
-                files={'content': manifest.content},
-            )
+        entities.Subscription().upload(
+            data={'organization_id': module_org.id},
+            files={'content': module_entitlement_manifest.content},
+        )
         repo_id = enable_rhrepo_and_fetchid(
             basearch='x86_64',
             org_id=module_org.id,
@@ -1382,14 +1380,13 @@ class TestContentViewRedHatOstreeContent:
     @pytest.mark.run_in_one_thread
     @pytest.mark.skip_if_not_set('fake_manifest')
     @pytest.fixture(scope='class', autouse=True)
-    def initiate_testclass(self, request, module_org):
+    def initiate_testclass(self, request, module_org, module_entitlement_manifest):
         """Set up organization, product and repositories for tests."""
 
-        with manifests.clone() as manifest:
-            entities.Subscription().upload(
-                data={'organization_id': module_org.id},
-                files={'content': manifest.content},
-            )
+        entities.Subscription().upload(
+            data={'organization_id': module_org.id},
+            files={'content': module_entitlement_manifest.content},
+        )
         repo_id = enable_rhrepo_and_fetchid(
             basearch=None,
             org_id=module_org.id,
--- a/tests/foreman/api/test_contentview.py
+++ b/tests/foreman/api/test_contentview.py
@@ -820,7 +820,6 @@ class TestContentViewDelete:
 class TestContentViewRedHatContent:
     """Tests for publishing and promoting content views."""
 
-    @pytest.mark.skip_if_not_set('fake_manifest')
     @pytest.fixture(scope='class', autouse=True)
     def initiate_testclass(self, request, module_org, module_cv, module_entitlement_manifest):
         """Set up organization, product and repositories for tests."""
@@ -1378,7 +1377,6 @@ class TestContentViewRedHatOstreeContent:
     """Tests for publishing and promoting cv with RH ostree contents."""
 
     @pytest.mark.run_in_one_thread
-    @pytest.mark.skip_if_not_set('fake_manifest')
     @pytest.fixture(scope='class', autouse=True)
     def initiate_testclass(self, request, module_org, module_entitlement_manifest):
         """Set up organization, product and repositories for tests."""
--- a/tests/foreman/api/test_errata.py
+++ b/tests/foreman/api/test_errata.py
@@ -23,9 +23,7 @@ import pytest
 from nailgun import entities
 
 from robottelo import constants
-from robottelo import manifests
 from robottelo.api.utils import enable_rhrepo_and_fetchid
-from robottelo.api.utils import upload_manifest
 from robottelo.api.utils import wait_for_tasks
 from robottelo.cli.factory import setup_org_for_a_custom_repo
 from robottelo.cli.factory import setup_org_for_a_rh_repo
@@ -53,18 +51,17 @@ def activation_key(module_org, module_lce):
 
 
 @pytest.fixture(scope='module')
-def rh_repo(module_org, module_lce, module_cv, activation_key):
+def rh_repo(module_entitlement_manifest_org, module_lce, module_cv, activation_key):
     return setup_org_for_a_rh_repo(
         {
             'product': constants.PRDS['rhel'],
             'repository-set': constants.REPOSET['rhst7'],
             'repository': constants.REPOS['rhst7']['name'],
-            'organization-id': module_org.id,
+            'organization-id': module_entitlement_manifest_org.id,
             'content-view-id': module_cv.id,
             'lifecycle-environment-id': module_lce.id,
             'activationkey-id': activation_key.id,
         },
-        force_manifest_upload=True,
     )
 
 
@@ -415,17 +412,14 @@ def test_positive_sorted_issue_date_and_filter_by_cve(module_org, custom_repo, t
 
 
 @pytest.fixture(scope='module')
-def setup_content_rhel6():
+def setup_content_rhel6(module_entitlement_manifest_org):
     """Setup content fot rhel6 content host
     Using `Red Hat Enterprise Virtualization Agents for RHEL 6 Server (RPMs)`
     from manifest, SATTOOLS_REPO for host-tools and yum_9 repo as custom repo.
 
     :return: Activation Key, Organization, subscription list
     """
-    org = entities.Organization().create()
-    with manifests.clone() as manifest:
-        upload_manifest(org.id, manifest.content)
-
+    org = module_entitlement_manifest_org
     rh_repo_id_rhva = enable_rhrepo_and_fetchid(
         basearch='x86_64',
         org_id=org.id,
@@ -825,12 +819,12 @@ def test_errata_installation_with_swidtags(
 
 
 @pytest.fixture(scope='module')
-def rh_repo_module_manifest(module_manifest_org):
+def rh_repo_module_manifest(module_entitlement_manifest_org):
     """Use module manifest org, creates tools repo, syncs and returns RH repo."""
     # enable rhel repo and return its ID
     rh_repo_id = enable_rhrepo_and_fetchid(
         basearch=constants.DEFAULT_ARCHITECTURE,
-        org_id=module_manifest_org.id,
+        org_id=module_entitlement_manifest_org.id,
         product=constants.PRDS['rhel8'],
         repo=constants.REPOS['rhst8']['name'],
         reposet=constants.REPOSET['rhst8'],
@@ -843,24 +837,24 @@ def rh_repo_module_manifest(module_manifest_org):
 
 
 @pytest.fixture(scope='module')
-def rhel8_custom_repo_cv(module_manifest_org):
+def rhel8_custom_repo_cv(module_entitlement_manifest_org):
     """Create repo and publish CV so that packages are in Library"""
     return setup_org_for_a_custom_repo(
         {
             'url': settings.repos.module_stream_1.url,
-            'organization-id': module_manifest_org.id,
+            'organization-id': module_entitlement_manifest_org.id,
         }
     )
 
 
 @pytest.fixture(scope='module')
 def rhel8_module_ak(
-    module_manifest_org, default_lce, rh_repo_module_manifest, rhel8_custom_repo_cv
+    module_entitlement_manifest_org, default_lce, rh_repo_module_manifest, rhel8_custom_repo_cv
 ):
     rhel8_module_ak = entities.ActivationKey(
-        content_view=module_manifest_org.default_content_view,
-        environment=entities.LifecycleEnvironment(id=module_manifest_org.library.id),
-        organization=module_manifest_org,
+        content_view=module_entitlement_manifest_org.default_content_view,
+        environment=entities.LifecycleEnvironment(id=module_entitlement_manifest_org.library.id),
+        organization=module_entitlement_manifest_org,
     ).create()
     # Ensure tools repo is enabled in the activation key
     rhel8_module_ak.content_override(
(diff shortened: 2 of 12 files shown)


synkd:
trigger: test-robottelo
pytest: tests/foreman/api/test_contentview.py::TestContentViewRedHatContent

tests/foreman/api/test_contentview.py
```diff
@@ -823,14 +822,13 @@ class TestContentViewRedHatContent:
 
     @pytest.mark.skip_if_not_set('fake_manifest')
```
Griffin-Sullivan, line 823: Should we remove this skip, since we aren't using `fake_manifest` now?

synkd: The current plan is to leave `fake_manifest` in place as a backup, but yes, we should remove `fake_manifest` skips. I'll make that change.

synkd:
A couple of the content management tests are failing locally during Capsule deployment, but I don't think manifester is the cause. The tests are passing in CI, though, so I'd like to see what results I get with PRT.

synkd:
trigger: test-robottelo
pytest: tests/foreman/api/test_contentmanagement.py -k "test_positive_sync_kickstart_repo or test_positive_sync_kickstart_check_os or test_positive_iso_library_sync"

synkd:
trigger: test-robottelo
pytest: tests/foreman/api/test_contentmanagement.py -k 'test_positive_sync_kickstart_repo or test_positive_sync_kickstart_check_os or test_positive_iso_library_sync'

synkd:
The two failures in the most recent PRT run are not consistent with my local failures. In the PRT run, the two tests that failed both did so while attempting to sync the RHEL 8 BaseOS kickstart repo. Strangely, it looks to me like `TestSatelliteContentManagement::test_positive_sync_kickstart_repo` successfully synced the repo, set the download policy explicitly to immediate, and subsequently failed to sync the same repo. Looking at the Nailgun Repository entity [1], though, it looks to me like the default policy is immediate, so I'm not sure why the test is written in this way. If my reading of the `Repository` entity is correct, then `TestSatelliteContentManagement::test_positive_sync_kickstart_check_os` is also failing to sync the RHEL 8 BaseOS kickstart repo with the immediate download policy.

I will do some additional debugging tomorrow.

[1] https://github.com/SatelliteQE/nailgun/blob/53b6f11cc77eb202bbf2066dbb27c79b85f35e42/nailgun/entities.py#L6145

synkd:
trigger: test-robottelo
pytest: tests/foreman/api/test_contentmanagement.py -k 'test_positive_sync_kickstart_repo'

synkd:
trigger: test-robottelo
pytest: tests/foreman/api/test_contentmanagement.py -k 'test_positive_sync_kickstart_repo or test_positive_sync_kickstart_check_os or test_positive_iso_library_sync'

synkd:
All the `api/test_contentmanagement.py` tests are passing today, so I am chalking up yesterday's failures to a transient CDN issue. Moving on to `api/test_convert2rhel.py`.

Griffin-Sullivan (approved):

synkd:
Just to clarify, this PR is still in progress. The following modules still need to be converted before this PR is taken out of draft:
api/test_convert2rhel.py
api/test_organization.py
api/test_product.py
api/test_reporttemplates.py
api/test_repository.py
api/test_repository_set.py
api/test_errata.py
api/test_syncplan.py
api/test_subscription.py

synkd:
Note: as mentioned in my last commit message, the tests in `api/test_convert2rhel.py` are failing both locally and in CI with the same error.

synkd:
trigger: test-robottelo
pytest: tests/foreman/api -k "test_negative_check_org_endpoint or test_positive_filter_product_list or test_positive_generate_entitlements_report or test_positive_schedule_entitlements_report"

synkd:
trigger: test-robottelo
pytest: tests/foreman/api/ -k 'test_negative_check_org_endpoint or test_positive_filter_product_list or test_positive_generate_entitlements_report or test_positive_schedule_entitlements_report'

synkd:
The two report templates tests failed with:
```
2022-10-19 17:21:22 - nailgun.client - WARNING - Received HTTP 400 response: {"displayMessage":"The specified organization is in Simple Content Access mode. Attaching subscriptions is disabled","errors":["The specified organization is in Simple Content Access mode. Attaching subscriptions is disabled"]}
```
This is likely because the PRT run was initiated before https://github.com/SatelliteQE/robottelo/pull/10179 was merged. Running them again to be sure.

synkd:
trigger: test-robottelo
pytest: tests/foreman/api/test_repository.py -k 'test_positive_recreate_pulp_repositories or test_positive_sync_rh or test_positive_sync_repo_null_contents_changed'

synkd:
trigger: test-robottelo
pytest: tests/foreman/api/test_repository_set.py

synkd:
The three tests in `api/test_errata.py` that use manifests are failing in CI, so I'm not going to run PRT for those tests.

synkd:
trigger: test-robottelo
pytest: tests/foreman/api/test_syncplan.py -k 'test_positive_synchronize_rh_product_future_sync_date or test_positive_synchronize_rh_product_past_sync_date'

synkd:
trigger: test-robottelo
pytest: tests/foreman/api/test_subscription.py::test_positive_subscription_status_disabled

synkd:
Looking into the logs for the PRT run, it appears that the `rhel6:latest` base image is built on RHEL 6.9:

`'[D 221021 22:05:42 container:197] {'id': 'f178b4887043966ecd44a4ffd72fdae9a863e389ca415dc64f53e995b7b5df28', 'image': 'sha256:46d6d65a8cf78763f03b43c62641e019312ebe3d99eaea30ef07b6a8ea94e4d0', 'name': 'satci_c9bdfdf1', 'container_config': {'Hostname': 'f178b4887043', *** 8< SNIP 8< *** 'JENKINS_URL=https://satqe-jenkins-csb-satellite-qe.apps.ocp-c1.prod.psi.redhat.com/job/robottelo-pr-testing/1389/', *** 8< SNIP 8< *** 'Image': 'rhel6:latest', 'Volumes': None, 'WorkingDir': '/root', 'Entrypoint': None, 'OnBuild': None, 'Labels': {'BROKER_ORIGIN': 'rhel_contenthost:/opt/app-root/src/robottelo/pytest_fixtures/core/contenthosts.py', 'BZComponent': 'rhel-server-docker', *** 8< SNIP 8< *** 'Name': 'rhel6', 'Version': '6.9', 'architecture': 'x86_64', 'authoritative-source-url': 'registry.access.redhat.com', 'broker_compatible': 'True', 'build-date': '2018-05-03T23:40:14.946673''`

RHEL 6.10 is the actual latest RHEL 6 y-stream release, and I don't believe that Simple Content Access is supported on RHEL 6.9 (though the version of `subscription-manager` is really the deciding factor there). This seems to be the cause of the failure; sub-man does not recognize that SCA is enabled:
```
[D 221021 22:05:47 hosts:106] f178b4887043 executing command: subscription-manager register --org LVOoUnxoQ --activationkey KmUprsrtQR --force
[D 221021 22:05:50 hosts:108] f178b4887043 command result:
    stdout:
    The system has been registered with ID: 80f1eb34-ae74-426a-95e9-c8c1caa89b3a 
    
    Installed Product Current Status:
    Product Name: Red Hat Enterprise Linux Server
    Status:       Not Subscribed
    
    Unable to find available subscriptions for all your installed products.
    
    stderr:
    
    status: 1
```
If my analysis is correct, then to resolve this issue, we'll need to either use a more current RHEL 6 image or convince the maintainers of the RHEL 6 UBI to update the image that we're pulling now. Either way, manifester does not appear to be the cause of this failure, as far as I can tell.

synkd:
Ok, some additional information on RHEL 6 and SCA: the first version of `subscription-manager` that is SCA-aware is 1.24 [1]. The latest version of `subscription-manager` to ship for RHEL 6 is 1.20. So `api/test_subscription.py::test_positive_subscription_status_disabled` is going to fail on RHEL 6 unless some sort of alternative logic is provided for that case. Consequently, I feel confident in stating that manifester is not the cause of that test failure.

[1] https://access.redhat.com/solutions/5060001

synkd:
trigger: test-robottelo
pytest: tests/foreman/api/test_subscription.py -k "test_sca_end_to_end or test_positive_candlepin_events_processed_by_stomp or test_positive_expired_SCA_cert_handling"

synkd:
trigger: test-robottelo
pytest: tests/foreman/api/test_subscription.py -k 'test_sca_end_to_end or test_positive_candlepin_events_processed_by_stomp or test_positive_expired_SCA_cert_handling'

synkd:
trigger: test-robottelo
pytest: tests/foreman/api/test_subscription.py -k 'test_sca_end_to_end or test_positive_candlepin_events_processed_by_stomp or test_positive_expired_SCA_cert_handling
nailgun: 850

synkd:
trigger: test-robottelo
pytest: tests/foreman/api/test_subscription.py -k 'test_sca_end_to_end or test_positive_candlepin_events_processed_by_stomp or test_positive_expired_SCA_cert_handling'
nailgun: 850

synkd:
`test_subscription.py::test_positive_candlepin_events_processed_by_stomp` is failing locally and in CI with the same error, which is unrelated to manifester.

I'm going to try one additional PRT run with all of the tests modified by this PR, excluding those that are failing in CI.

synkd:
trigger: test-robottelo
pytest: tests/foreman/api/ -k 'test_positive_add_rh or test_positive_add_rh_custom_spin or test_positive_update_rh_custom_spin or test_positive_publish_rh or test_positive_publish_rh_custom_spin or test_positive_promote_rh or test_positive_promote_rh_custom_spin or test_positive_sync_kickstart_repo or test_positive_sync_kickstart_check_os or test_positive_iso_library_sync or test_negative_check_org_endpoint or test_positive_filter_product_list or test_positive_generate_entitlements_report or test_positive_schedule_entitlements_report or test_positive_recreate_pulp_repositories or test_positive_sync_rh or test_positive_sync_repo_null_contents_changed or test_positive_reposet_enable or test_positive_reposet_disable or test_positive_synchronize_rh_product_future_sync_date or test_positive_synchronize_rh_product_past_sync_date or test_positive_subscription_status_disabled or test_positive_refresh or test_positive_create_after_refresh or test_negative_upload or test_positive_delete_manifest_as_another_user or test_positive_subscription_status_disabled or test_sca_end_to_end or test_positive_expired_SCA_cert_handling'
nailgun: 850

synkd:
The one failure in PRT build 1404 was the RHEL 6 run of `test_subscription.py::test_positive_subscription_status_disabled`, which is expected based on my comment above. I'm trying to determine which three content view tests were skipped and why.

synkd:
Running the pytest command locally with the `-rs --setup-plan` options, it looks like the skipped tests were the Red Hat OSTree content view tests:
```
SKIPPED [1] tests/foreman/api/test_contentview.py:1402: BZ:1625783
SKIPPED [1] tests/foreman/api/test_contentview.py:1421: BZ:1625783
SKIPPED [1] tests/foreman/api/test_contentview.py:1439: BZ:1625783
```
I'm not quite sure what those tests matched on, since I did not intend to run them. But overall, I would say that the PRT run looks good. The only failure is expected and not caused by manifester.

JacobCallahan (approved):
ACK, nice!

jyejare:
@synkd You need to rebase !

tests/foreman/api/test_subscription.py
```diff
@@ -58,31 +57,38 @@ def rh_repo(module_org):
...
 
 
-@pytest.mark.skip_if_not_set('fake_manifest')
+@pytest.fixture(scope='function')
+def duplicate_manifest():
+    """Provides a function-scoped manifest that can be used alongside function_entitlement_manifest
+    when two manifests are required in a single test"""
+    with Manifester(manifest_category=settings.manifest.entitlement) as manifest:
+        yield manifest
+
```
Griffin-Sullivan, line 87: Would we want to put this with other fixtures, so it could be used in other scenarios?

Griffin-Sullivan (approved):

tests/foreman/api/test_subscription.py
synkd: Possibly, but so far this is the only module I've encountered that has required a duplicate manifest. If I find another module that requires a duplicate function-scoped manifest, then I'll move this fixture definition to `pytest_fixtures/component/taxonomy.py`.

adarshdubey-star:
@synkd  Is there any specific reason to not switch from entities to target_sat  & getting rid of entities where possible?

adarshdubey-star (approved):

tests/foreman/api/test_errata.py
```diff
@@ -53,18 +51,17 @@ def activation_key(module_org, module_lce):
 
 
 @pytest.fixture(scope='module')
-def rh_repo(module_org, module_lce, module_cv, activation_key):
+def rh_repo(module_entitlement_manifest_org, module_lce, module_cv, activation_key):
```
Gauravtalreja1, line 54: is this change intended?

tests/foreman/api/test_organization.py
```diff
@@ -146,9 +144,7 @@ def test_negative_check_org_endpoint(self):
 
         :CaseImportance: High
         """
-        org = entities.Organization().create()
-        with manifests.clone() as manifest:
-            upload_manifest(org.id, manifest.content)
+        org = module_entitlement_manifest_org
         orgstring = json.dumps(org.read_json())
```
Gauravtalreja1, line 148:
```suggestion
        orgstring = json.dumps(module_entitlement_manifest_org.read_json())
```

tests/foreman/api/test_subscription.py
```diff
@@ -108,19 +111,14 @@ def test_positive_refresh(request):
 
     :CaseImportance: Critical
     """
-    org = entities.Organization().create()
+    org = function_entitlement_manifest_org
     sub = entities.Subscription(organization=org)
-    with manifests.original_manifest() as manifest:
-        upload_manifest(org.id, manifest.content)
```
Gauravtalreja1, line 114: are we good with replacing `manifests.clone()` and `manifests.original_manifest()` both with entitlement_manifest_org?

tests/foreman/api/test_subscription.py
```diff
@@ -108,19 +111,14 @@ def test_positive_refresh(request):
...
-    request.addfinalizer(lambda: sub.delete_manifest(data={'organization_id': org.id}))
```
Gauravtalreja1, line 115: I think if `delete_manifest` was part of the teardown here then we shouldn't remove this line

And Not sure if a similar `delete_manifest` is required for a teardown for global fixtures?

tests/foreman/api/test_subscription.py
```diff
@@ -177,15 +170,17 @@ def test_negative_upload():
         organization.
     """
     orgs = [entities.Organization().create() for _ in range(2)]
-    with manifests.clone() as manifest:
+    with function_entitlement_manifest as manifest:
```
Gauravtalreja1, line 173: function_entitlement_manifest is already in contextmanager, I think Manifestor contextmanager would help here instead.
```suggestion
    with Manifester(manifest_category=settings.manifest.entitlement) as manifest:
```

tests/foreman/api/test_subscription.py
```diff
@@ -224,7 +219,7 @@ def test_positive_delete_manifest_as_another_user(function_org, target_sat):
         verify=False,
     )
     # use the first admin to upload a manifest
-    with manifests.clone() as manifest:
+    with function_entitlement_manifest as manifest:
```
Gauravtalreja1, line 222: Same here?

tests/foreman/api/test_errata.py
synkd: Yes, this is intended. In converting modules with local fixture stacks, I am generally replacing module_org with module_entitlement_manifest_org if the module_org previously had a manifest imported (as was the case here with the `force_manifest_upload=True` option).

tests/foreman/api/test_subscription.py
synkd: `delete_manifest` was only necessary because of the use of `original_manifest()`: https://github.com/SatelliteQE/robottelo/blob/2e5794e62d381aacde94f01c5265937737165424/robottelo/manifests.py#L183

Each use of `function_entitlement_manifest_org` (or `function_entitlement_manifest`) will create a unique manifest used only for that function. Since the org and the manifest are both function-scoped here, we can safely remove this line.

synkd: Regarding teardown for global fixtures, deleting the manifest from the organization should not generally be necessary for function- or module-scoped fixtures. However, note that part of the teardown on the Manifester side is to delete the upstream subscription allocation, which will free up the subscriptions in the RHSM account.

tests/foreman/api/test_subscription.py
synkd: Yes, this should cause no issues and should be more flexible than both `manifests.clone()` and `manifests.original_manifest()`.

tests/foreman/api/test_subscription.py
synkd: My thinking is that calling manifester directly in a test function should only be done in situations where it is required to do so (such as a single test function that requires multiple manifests). That way, we have a centralized location in the fixture definitions in case we need to make changes to how they operate in the future. For that reason, I would prefer to keep this as is. Could you please explain what you see as the functional advantage of making this change?

tests/foreman/api/test_contentview.py
```diff
@@ -821,16 +820,14 @@ def test_positive_delete(self, content_view, name):
...
 
-        with manifests.clone() as manifest:
-            entities.Subscription().upload(
-                data={'organization_id': module_org.id},
-                files={'content': manifest.content},
-            )
+        entities.Subscription().upload(
+            data={'organization_id': module_org.id},
+            files={'content': module_entitlement_manifest.content},
+        )
```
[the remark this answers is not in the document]
synkd: In doing these conversions, I have been trying to minimize the changes that I'm making to tests to keep the scope of this PR in check. However, I agree that your suggestion here makes sense and is more concise, so I'll make that change here and in the other location you point out below.

tests/foreman/api/test_errata.py
synkd: Yes, this is intended. For modules with local fixture stacks, I have been changing module_org to module_entitlement_manifest_org in cases such as this one where a manifest is uploaded to the org, which is done here with the `force_manifest_upload=True` option to `setup_org_for_a_rh_repo`.

tests/foreman/api/test_subscription.py
shweta83: This fixture can be used in other modules with changes coming in other test modules. Moving to taxonomy.py would bring better visibility for this fixture.

tests/foreman/api/test_subscription.py
```diff
@@ -311,7 +310,9 @@ def test_sca_end_to_end(module_ak, rhel7_contenthost, module_org, rh_repo, custo
 
 
 @pytest.mark.tier2
-def test_positive_candlepin_events_processed_by_stomp(rhel7_contenthost, function_org, target_sat):
+def test_positive_candlepin_events_processed_by_stomp(
+    rhel7_contenthost, function_entitlement_manifest, function_org, target_sat
```
shweta83, line 314:
```suggestion
    rhel7_contenthost, function_entitlement_manifest_org, target_sat
```

tests/foreman/api/test_subscription.py
Gauravtalreja1: Sounds great for teardown for global fixtures 👍

and for this particular teardown step, I know it's safe to remove, but I think add/refresh/delete manifest scenarios should be part test_subscription to test api endpoints using `entities.Subscription`, so it would be good to keep this step, right?

synkd: Fair enough, I hadn't considered that deleting specifically after refreshing should be considered a particular test scenario. I'll add the line back in.

tests/foreman/api/test_subscription.py
synkd: At first I implemented this change, then, while debugging locally, I realized that the change actually breaks the test. The first bits of the test are registering a content host to an organization that has not had a manifest uploaded yet and verifying that the host does not have a valid subscription. Then, the test uploads a manifest, attaches a subscription to the host, and verifies that it does have a valid subscription. So we actually need to use separate organization and manifest fixtures in this case.

tests/foreman/api/test_subscription.py
synkd: @Gauravtalreja1 So this test actually requires that we use one manifest. It's testing to ensure that uploading the same manifest to two different organizations results in an error. I did run this test as written in this PR and it passes (i.e. we do get the error when the manifest is uploaded to the second org).

I agree that we could relatively easily search for the string `Manifester` to identify patterns like the one you're suggesting if we need to make updates in the future. However, the reason I'm pushing back on your suggestion is that, while I've been working on the manifester conversions, I have noticed some design patterns in Robottelo that perform ad hoc actions instead of using fixtures, similar to this case (the one that stands out the most is creating an organization in a test function with `entities.Organization.create()` instead of using the `function_org` fixture). While such patterns function perfectly well, in my opinion they add friction to the process of maintaining or refactoring code. This is most noticeable when making changes to a large number of tests like I'm doing with the manifester conversions. Modifying one instance of such a pattern is fairly trivial, but the work adds up when it has to be repeated dozens or hundreds of times.

To give an example of this type of change that's directly relevant here, at some point in the future, we will need to convert all uses of entitlement manifests to Simple Content Access manifests (because entitlement mode will eventually be removed as an option for manifests). To me, it seems simpler to primarily modify the fixtures being used by tests to implement that change and to reserve exceptions for cases in which calling Manifester directly is functionally required.

Let me know if you agree with this; if not, I'm still open to arguments for using the pattern that you suggested.

Gauravtalreja1: @synkd Okay, got it, if this works then it's fine, won't block PR for this suggestion.

Gauravtalreja1 (approved):

PR #10163, opened 2022-10-13
Temporary workaround for dynaconf.vendor.box.exceptions.BoxKeyError: DynaBox object has no attribute host

adarshdubey-star:
1. I am seeing this issue when we are trying to spin two containers as a host which happens due to https://github.com/SatelliteQE/broker/blob/master/broker/broker.py#L363 when it's trying to invoke the host.setup() function in robotello/hosts.py.

```
    def setup(self):
        if not self.blank:
            self.remove_katello_ca()
            self.execute('subscription-manager clean')
```

### DIFF
--- a/tests/foreman/api/test_rhcloud_inventory.py
+++ b/tests/foreman/api/test_rhcloud_inventory.py
@@ -29,6 +29,7 @@ from robottelo.utils.io import get_report_data
 from robottelo.utils.io import get_report_metadata
 
 generate_report_task = 'ForemanInventoryUpload::Async::UploadReportJob'
+pytestmark = [pytest.mark.no_containers]
 
 
 def generate_inventory_report(satellite, org):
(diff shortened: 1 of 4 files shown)


jyejare:
@adarshdubey-star Have you raised an issue in broker ?

adarshdubey-star:
> @adarshdubey-star Have you raised an issue in broker ?

Not yet, will raise it.

adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/api/test_rhcloud_inventory.py tests/foreman/cli/test_rhcloud_inventory.py tests/foreman/cli/test_rhcloud_insights.py tests/foreman/ui/test_rhcloud_inventory.py tests/foreman/ui/test_rhcloud_insights.py

JacobCallahan:
@adarshdubey-star can you elaborate on what the actual issue is?
Supporting information would be the context, tracebacks, and what you believe is going wrong.

JacobCallahan:
OK @adarshdubey-star and I just dug into the issue more and this is a problem with pickle stripping the _cont_inst property from the Host class during multiprocessing checkouts. I'll see if I can come up with a good solution to this today. If not, then we can merge this as a temporary workaround.

JacobCallahan:
Broker fix proposed: https://github.com/SatelliteQE/broker/pull/170

JacobCallahan (changes requested):
NACK, pending testing of proposed Broker changes

PR #10172, opened 2022-10-17
fix runtime error introduced by 884fb93

rplevka:
the current implementation was causing runtime errors since the loop was messing with the dict keys

### DIFF
--- a/pytest_fixtures/core/broker.py
+++ b/pytest_fixtures/core/broker.py
@@ -12,8 +12,8 @@ from robottelo.logging import logger
 
 def _resolve_deploy_args(args_dict):
     # TODO: https://github.com/rochacbruno/dynaconf/issues/690
-    args_dict = args_dict.copy().to_dict()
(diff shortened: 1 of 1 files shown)


pytest_fixtures/core/broker.py
```diff
@@ -12,8 +12,8 @@
 
 def _resolve_deploy_args(args_dict):
     # TODO: https://github.com/rochacbruno/dynaconf/issues/690
-    args_dict = args_dict.copy().to_dict()
-    for key, val in args_dict.items():
+    args_dict_copy = args_dict.copy().to_dict()
+    for key, val in args_dict_copy.items():
```
JacobCallahan, line 16: There shouldn't be a need to store the copy in this version. Additionally, to_dict may also not be needed.
```suggestion
    for key, val in args_dict.copy().to_dict().items():
```

JacobCallahan (approved):
Non-blocking, but something to consider

PR #10190, opened 2022-10-21
added test coverage for clone host-group

omkarkhatavkar:
#### PR Description 

Added the test coverage for the clone host group as [bz 2122261](https://bugzilla.redhat.com/show_bug.cgi?id=2122261)

#### Dependency 
Airgun PR :  https://github.com/SatelliteQE/airgun/pull/773

#### Test Result 

```
Launching pytest with arguments -k test_positive_clone_host_groups /Users/okhatavk/satellite/robottelo/tests/foreman/ui/test_hostgroup.py --no-header --no-summary -q in /Users/okhatavk/satellite/robottelo

collecting ... collected 7 items / 6 deselected / 1 selected

tests/foreman/ui/test_hostgroup.py::test_positive_clone_host_groups 

=========== 1 passed, 6 deselected, 25 warnings in 246.58s (0:04:06) ===========
```

### DIFF
--- a/tests/foreman/ui/test_hostgroup.py
+++ b/tests/foreman/ui/test_hostgroup.py
@@ -194,6 +194,7 @@ def test_positive_create_new_host():
     pass
 
 
+@pytest.mark.tier2
 def test_positive_nested_host_groups(
     session, module_org, module_lce, module_published_cv, module_ak_cv_lce, target_sat
 ):
@@ -253,7 +254,7 @@ def test_positive_nested_host_groups(
             {
                 'host_group.lce': module_lce.name,
                 'host_group.content_view': module_published_cv.name,
-                'host_group.activation_keys': module_ak_cv_lce.name,
+                'activation_keys.activation_keys': module_ak_cv_lce.name,
             },
         )
         child_hostgroup_values = session.hostgroup.read(f'{parent_hg_name}/{child_hg_name}')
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_hostgroup.py
```diff
@@ -264,3 +266,77 @@ def test_positive_nested_host_groups(
...
+            parent_hg_name,
+            {
+                'host_group.name': clone_hg_name,
+            },
+        )
+        assert target_sat.api.HostGroup().search(query={'search': f'name={clone_hg_name}'})
+        clone_hostgroup_values = session.hostgroup.read(clone_hg_name)
+        assert module_ak_cv_lce.name in clone_hostgroup_values['host_group']['lce']
+        assert module_published_cv.name in clone_hostgroup_values['host_group']['content_view']
+        assert module_ak_cv_lce.name in clone_hostgroup_values['host_group']['activation_keys']
```
pondrejk, line 326: should we go over all parameters, not only those that are currently not copied due to the bug? Maybe in the future it will be os architecture that is not copied, so this test could catch that

pondrejk (approved):
ack pending one comment

pondrejk (approved):

tests/foreman/ui/test_hostgroup.py
```diff
@@ -264,3 +266,79 @@ def test_positive_nested_host_groups(
...
+@pytest.mark.tier2
+def test_positive_clone_host_groups(
+    session, module_org, module_lce, module_published_cv, module_ak_cv_lce, target_sat
+):
+    """Verify create, update and delete operation for clone host-group
+
+    :id: 9f02dcc5-98aa-48bd-8114-edd3a0be65c1
+
+    :Steps:
+
```
ogajduse, line 281: unnecessary newline

tests/foreman/ui/test_hostgroup.py
```diff
@@ -264,3 +266,79 @@ def test_positive_nested_host_groups(
...
+        assert module_published_cv.name in clone_hostgroup_values['host_group']['content_view']
+        assert module_ak_cv_lce.name in clone_hostgroup_values['activation_keys']['activation_keys']
+        assert os_name in clone_hostgroup_values['operating_system']['operating_system']
+        assert architecture.name in clone_hostgroup_values['operating_system']['architecture']
+
+        # Update clone host group
+        session.hostgroup.update(
+            clone_hg_name,
+            {
+                'host_group.lce': ENVIRONMENT,
```
ogajduse, line 334: I just wonder if checking against `module_published_cv` and `module_lce` is better than checking against `DEFAULT_CV` and `ENVIRONMENT` here.

ogajduse (approved):
LGTM, comments are non-blocking

tests/foreman/ui/test_hostgroup.py
```diff
@@ -264,3 +266,79 @@ def test_positive_nested_host_groups(
...
+                'host_group.content_view': DEFAULT_CV,
+            },
+        )
+        clone_hostgroup_values = session.hostgroup.read(clone_hg_name)
+        assert ENVIRONMENT in clone_hostgroup_values['host_group']['lce']
+        assert DEFAULT_CV in clone_hostgroup_values['host_group']['content_view']
+
+        # Delete clone host group
+        session.hostgroup.delete(clone_hg_name)
```
vsedmik, line 343: We might want to delete the 'parent_hg' at the end of the test as well (not necessarily via UI).

vsedmik (approved):
ACK, one non-blocking comment to consider.

tests/foreman/ui/test_hostgroup.py
omkarkhatavkar: I wanted to test the `module_ak_cv_lce.name` activation key while cloning the host group. I did not have the activation key created for Default Environment. Hence I design tests like that. I don't see any difference, even if we change to Default Environment, anyways it should be get updated in either case.

tests/foreman/ui/test_hostgroup.py
```diff
@@ -264,3 +266,79 @@ def test_positive_nested_host_groups(
         # Delete nested host group
         session.hostgroup.delete(f'{parent_hg_name}/{child_hg_name}')
         assert not target_sat.api.HostGroup().search(query={'search': f'name={child_hg_name}'})
+
+
+@pytest_skip_if_bug_open('bugzilla', 2122261)
```
Gauravtalreja1, line 271: Use `@pytest.mark.skip_if_open('BZ:2122261')` instead

tests/foreman/ui/test_hostgroup.py
```diff
@@ -264,3 +265,80 @@ def test_positive_nested_host_groups(
...
+                'host_group.content_view': DEFAULT_CV,
+            },
+        )
+        clone_hostgroup_values = session.hostgroup.read(clone_hg_name)
+        assert ENVIRONMENT in clone_hostgroup_values['host_group']['lce']
+        assert DEFAULT_CV in clone_hostgroup_values['host_group']['content_view']
+
+        # Delete parent and clone host group
+        session.hostgroup.delete(parent_hg_name)
+        assert target_sat.api.HostGroup().search(query={'search': f'name={clone_hg_name}'})
```
Gauravtalreja1, line 342: Shouldn't this assert be checking if `parent_hg_name` HostGroup exists after deletion?

Gauravtalreja1 (commented):
Ack, with a question

tests/foreman/ui/test_hostgroup.py
omkarkhatavkar: the intention was that it does not affect cloned host group.

Gauravtalreja1: Okay, it would be better to add extra assert after parent_hg_name delete, non-blocking

Gauravtalreja1 (approved):

tests/foreman/ui/test_hostgroup.py
omkarkhatavkar: no need, as this is already part of `test_positive_end_to_end` test

pondrejk (approved):

PR #10194, opened 2022-10-21
Add a test to check pulp status via pulp-cli

vsedmik:
Since 6.12 the [pulp-cli](https://github.com/pulp/pulp-cli) is installed by default on the Satellite. In this test we check that:
1) The pulp-cli is installed and configured to work properly.
2) The pulp components are alive according to provided status.

### DIFF
--- a/tests/foreman/sys/test_pulp3_filesystem.py
+++ b/tests/foreman/sys/test_pulp3_filesystem.py
@@ -16,6 +16,9 @@
 
 :Upstream: No
 """
+import json
+from datetime import datetime
+
 import pytest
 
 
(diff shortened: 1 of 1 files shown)


vsedmik:
trigger: test-robottelo
pytest: tests/foreman/sys/test_pulp3_filesystem.py::test_pulp_status

Griffin-Sullivan (approved):

tests/foreman/sys/test_pulp3_filesystem.py
```diff
@@ -96,3 +99,28 @@ def test_pulp_workers_status(target_sat):
...
     result = output.stdout.rstrip().splitlines()
     assert all('Active: active (running)' in r for r in result)
+
+
+@pytest.mark.tier1
+def test_pulp_status(target_sat):
+    """Test pulp status via pulp-cli.
+
+    :id: aa7fda5a-cfa1-4fa1-8714-8d5a2e2f89d9
+
```
ogajduse, line 109: Could you add the `:steps:` section to the docstring?

tests/foreman/sys/test_pulp3_filesystem.py
```diff
@@ -96,3 +99,28 @@ def test_pulp_workers_status(target_sat):
...
+    status = json.loads(target_sat.execute('pulp status').stdout)
+    now = datetime.utcnow()
+    assert status['database_connection']['connected']
+    assert status['redis_connection']['connected']
+
+    workers_beats = [
+        datetime.strptime(worker['last_heartbeat'], '%Y-%m-%dT%H:%M:%S.%fZ')
+        for worker in status['online_workers']
+    ]
+    assert all((now - beat).seconds < 20 for beat in workers_beats)
```
ogajduse, line 121: Where does the value `20` come from? I wonder if it is worth creating some kind of global variable to store the value in. However, it might cause more pollution to the constants.

tests/foreman/sys/test_pulp3_filesystem.py
```diff
@@ -96,3 +99,28 @@ def test_pulp_workers_status(target_sat):
...
+    apps_beats = [
+        datetime.strptime(app['last_heartbeat'], '%Y-%m-%dT%H:%M:%S.%fZ')
+        for app in status['online_content_apps']
+    ]
+    assert all((now - beat).seconds < 20 for beat in apps_beats)
```
ogajduse, line 126: Could you add an error message for both assertions?

tests/foreman/sys/test_pulp3_filesystem.py
vsedmik: It comes just from my observations, the heartbeat was never older than 10 seconds during multiple syncs.

tests/foreman/sys/test_pulp3_filesystem.py
vsedmik: Added.

pondrejk (approved):

PR #10199, opened 2022-10-24
getting rid of fake proxy in realm tests

pondrejk:
and as realm was the only remaining module to use that, I removed the fixture as well

### DIFF
--- a/pytest_fixtures/component/smartproxy.py
+++ b/pytest_fixtures/component/smartproxy.py
@@ -1,7 +1,4 @@
 import pytest
-from fauxfactory import gen_string
-
-from robottelo.cli.proxy import CapsuleTunnelError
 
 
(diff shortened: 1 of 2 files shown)


pondrejk:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_realm.py

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_realm.py

pondrejk:
eh, the problem seems to be in fixture resolution order, after module_enroll_idm_and_configure_external_auth re-runs the installer, the module capsule is no longer seen by satellite, looking for the most plausible

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_realm.py

PR #10204, opened 2022-10-26
cherrypick_label check fails for z-streams pr's

omkarkhatavkar:
Enforcing cherrypick labels check fails for the Y or Z stream AutoCherryPicked PR'S  https://github.com/SatelliteQE/robottelo/actions/runs/3326380185/jobs/5500043843

### DIFF
--- a/.github/workflows/required_labels.yml
+++ b/.github/workflows/required_labels.yml
@@ -14,4 +14,4 @@ jobs:
         with:
           mode: exactly
           count: 1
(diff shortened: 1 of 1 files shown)


omkarkhatavkar:
had offline discussion with @jyejare This is not needed as the `CheryPick` or `No-Cherrypick` is an instruction to GHA whether that should be auto cherrypicked down to the previous zStream

PR #10231, opened 2022-11-02
Port/refactor of f-m service tests from testfm

Gauravtalreja1:
**Description:**
1. Port/refactor f-m service subcommand tests from TestFM to robottelo
https://github.com/SatelliteQE/testfm/blob/master/tests/test_service.py

It includes changes from https://github.com/SatelliteQE/robottelo/pull/10227, and it will be rebased when either PR is merged

### DIFF
--- a/robottelo/cli/sm_service.py
+++ b/robottelo/cli/sm_service.py
@@ -27,50 +27,50 @@ class Service(Base):
     command_base = 'service'
 
     @classmethod
-    def start(cls, options=None):
+    def start(cls, options=None, env_var=None):
         """Build satellite-maintain service start"""
         cls.command_sub = 'start'
         options = options or {}
-        return cls.sm_execute(cls._construct_command(options))
+        return cls.sm_execute(cls._construct_command(options), env_var=env_var)
 
     @classmethod
-    def stop(cls, options=None):
+    def stop(cls, options=None, env_var=None):
         """Build satellite-maintain service stop"""
         cls.command_sub = 'stop'
         options = options or {}
-        return cls.sm_execute(cls._construct_command(options))
+        return cls.sm_execute(cls._construct_command(options), env_var=env_var)
 
     @classmethod
-    def restart(cls, options=None):
+    def restart(cls, options=None, env_var=None):
         """Build satellite-maintain service"""
         cls.command_sub = 'restart'
         options = options or {}
-        return cls.sm_execute(cls._construct_command(options))
+        return cls.sm_execute(cls._construct_command(options), env_var=env_var)
 
     @classmethod
-    def status(cls, options=None):
+    def status(cls, options=None, env_var=None):
         """Build satellite-maintain service status"""
         cls.command_sub = 'status'
         options = options or {}
-        return cls.sm_execute(cls._construct_command(options))
+        return cls.sm_execute(cls._construct_command(options), env_var=env_var)
 
     @classmethod
-    def enable(cls, options=None):
+    def enable(cls, options=None, env_var=None):
         """Build satellite-maintain service enable"""
         cls.command_sub = 'enable'
         options = options or {}
-        return cls.sm_execute(cls._construct_command(options))
+        return cls.sm_execute(cls._construct_command(options), env_var=env_var)
 
     @classmethod
-    def disable(cls, options=None):
+    def disable(cls, options=None, env_var=None):
         """Build satellite-maintain service disable"""
         cls.command_sub = 'disable'
         options = options or {}
-        return cls.sm_execute(cls._construct_command(options))
+        return cls.sm_execute(cls._construct_command(options), env_var=env_var)
 
     @classmethod
-    def list(cls, options=None):
+    def list(cls, options=None, env_var=None):
         """Build satellite-maintain service list"""
         cls.command_sub = 'list'
         options = options or {}
-        return cls.sm_execute(cls._construct_command(options))
+        return cls.sm_execute(cls._construct_command(options), env_var=env_var)
--- a/tests/foreman/maintain/test_service.py
+++ b/tests/foreman/maintain/test_service.py
@@ -1,0 +1,267 @@
+"""Test module for satellite-maintain service functionality
+
+:Requirement: foreman-maintain
+
+:CaseAutomation: Automated
+
+:CaseLevel: Component
+
+:CaseComponent: ForemanMaintain
+
+:Assignee: gtalreja
+
+:TestType: Functional
+
+:CaseImportance: Critical
+
+:Upstream: No
+"""
+import pytest
+from fauxfactory import gen_string
+
+from robottelo.config import settings
+from robottelo.constants import HAMMER_CONFIG
+from robottelo.constants import MAINTAIN_HAMMER_YML
+from robottelo.constants import SATELLITE_ANSWER_FILE
+from robottelo.utils.issue_handlers import is_open
+
+pytestmark = pytest.mark.destructive
+
+
+@pytest.fixture
+def missing_hammer_config(request, sat_maintain):
+    """Setup/teardown fixture used by test_positive_service_restart_without_hammer_config"""
+    flag = 'initial_admin_password'
+    sat_maintain.execute(
+        f'sed -i "s/{flag}: {settings.server.admin_password}/{flag}: invalid/g" '
+        f'{SATELLITE_ANSWER_FILE}'
+    )
+    sat_maintain.execute(f'mv {HAMMER_CONFIG} /tmp/foreman.yml')
+    sat_maintain.execute(f'mv {MAINTAIN_HAMMER_YML} /tmp/foreman-maintain-hammer.yml')
+
+    @request.addfinalizer
+    def _finalize():
+        sat_maintain.execute(
+            f'sed -i "s/{flag}: .*/{flag}: {settings.server.admin_password}/g" '
+            f'{SATELLITE_ANSWER_FILE}'
+        )
+        sat_maintain.execute(f'mv /tmp/foreman.yml {HAMMER_CONFIG}')
+        sat_maintain.execute(f'mv /tmp/foreman-maintain-hammer.yml {MAINTAIN_HAMMER_YML}')
+
+
+@pytest.mark.include_capsule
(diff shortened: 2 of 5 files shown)


Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_service.py

tests/foreman/maintain/test_service.py
```diff
@@ -0,0 +1,309 @@
...
+    assert 'FAIL' not in result.stdout
+    assert result.status == 0
+    result = sat_maintain.cli.Service.status(options={'brief': True})
+    logger.info(result.stdout)
+    assert 'FAIL' not in result.stdout
+    assert result.status == 0
+
+
+@pytest.mark.include_capsule
+def test_positive_service_stop_restart(sat_maintain):
```
shubhamsg199, line 124: What do you think of merging this test with https://github.com/SatelliteQE/robottelo/pull/10231/files#diff-122f6911bfcde697ff27cd2f1d73e85bf0ac0e01aacb371e7fc0c7cbbec5a181R90 ?
As only one additional step of restart is present here, can be added in above test

tests/foreman/maintain/test_service.py
```diff
@@ -0,0 +1,309 @@
...
+        sat_maintain.execute(
+            f'sed -i "s/{flag}: invalid/{flag}: {settings.server.admin_password}/g" '
+            f'{satellite_answer_file}'
+        )
+        sat_maintain.execute(f'mv /tmp/foreman.yml {HAMMER_CONFIG}')
+        sat_maintain.execute(f'mv /tmp/foreman-maintain-hammer.yml {maintain_hammer_yml}')
+
+
+@pytest.mark.include_capsule
+def test_positive_service_restart(sat_maintain):
```
shubhamsg199, line 52: Do we need this test as this logic is already tested in test_positive_service_stop_restart

tests/foreman/maintain/test_service.py
```diff
@@ -0,0 +1,309 @@
+"""Test class for satellite-maintain upgrade functionality
```
vsedmik, line 1: this

tests/foreman/maintain/test_service.py
```diff
@@ -0,0 +1,309 @@
...
+
+    :steps:
+        1. Run satellite-maintain service list
+
+    :expectedresults: Satellite services should be listed
+    """
+    result = sat_maintain.cli.Service.list()
+    logger.info(result.stdout)
+    assert 'FAIL' not in result.stdout
+    assert result.status == 0
```
vsedmik, line 86: Do we have an ambition to check what particular services are listed and what is their status in this test case?

vsedmik (approved):
LGTM, just a few comments.

tests/foreman/maintain/test_service.py
Gauravtalreja1: @shubhamsg199 yes, we can combine this tests, and perform stop twice, but I think its better to be explicit, and keep an separate test for it

tests/foreman/maintain/test_service.py
Gauravtalreja1: Updated

tests/foreman/maintain/test_service.py
Gauravtalreja1: Good catch, updated

shubhamsg199 (approved):
Ack

robottelo/cli/sm_service.py
```diff
@@ -27,50 +27,50 @@ class Service(Base):
     command_base = 'service'
 
     @classmethod
-    def start(cls, options=None):
+    def start(cls, options=None, env_var=''):
```
jyejare, line 30:
```suggestion
    def start(cls, options=None, env_var=None):
```
same as comment - https://github.com/SatelliteQE/robottelo/pull/10227#discussion_r1022466973

tests/foreman/maintain/test_service.py
```diff
@@ -0,0 +1,267 @@
...
+        f'sed -i "s/{flag}: {settings.server.admin_password}/{flag}: invalid/g" '
+        f'{satellite_answer_file}'
+    )
+    sat_maintain.execute(f'mv {HAMMER_CONFIG} /tmp/foreman.yml')
+    sat_maintain.execute(f'mv {maintain_hammer_yml} /tmp/foreman-maintain-hammer.yml')
+
+    @request.addfinalizer
+    def _finalize():
+        sat_maintain.execute(
+            f'sed -i "s/{flag}: invalid/{flag}: {settings.server.admin_password}/g" '
```
jyejare, line 45:
```suggestion
            f'sed -i "s/{flag}: .*/{flag}: {settings.server.admin_password}/g" '
```
Match any character than fixed string to return to the original value.

tests/foreman/maintain/test_service.py
```diff
@@ -0,0 +1,267 @@
...
+        3. Run satellite-maintain service start
+        4. Run satellite-maintain service status --brief
+
+    :expectedresults: Satellite services should stop and start.
+    """
+    result = sat_maintain.cli.Service.stop()
+    assert 'FAIL' not in result.stdout
+    assert result.status == 0
+    result = sat_maintain.cli.Service.status(options={'brief': True})
+    assert 'FAIL' in result.stdout
```
jyejare, line 90: Can you add a comment above as to what actually gets into FAIL state ?
OR you could you be specific as `FAIL something> string so that we will not false pass this test in actual FAIL message for service stopping.

tests/foreman/maintain/test_service.py
```diff
@@ -0,0 +1,267 @@
...
+
+    :expectedresults: service should restart.
+
+    :BZ: 1626651
+    """
+    result = sat_maintain.cli.Service.stop()
+    assert 'FAIL' not in result.stdout
+    assert result.status == 0
+    result = sat_maintain.cli.Service.status(options={'brief': True})
+    assert 'FAIL' in result.stdout
```
jyejare, line 122: Same as above.

tests/foreman/maintain/test_service.py
```diff
@@ -0,0 +1,267 @@
...
+        1. Stop foreman service
+        2. Check the status for httpd should not have affect
+        3. Run satellite-maintain health check
+
+    :expectedresults: foreman service should restart correctly
+    """
+    result = sat_maintain.cli.Service.stop(options={'only': 'foreman'})
+    assert result.status == 0
+    assert 'FAIL' not in result.stdout
+    assert 'foreman' in result.stdout
```
jyejare, line 170: I think its good to check the actual status of foreman service through `Service.status` stdout than just chaking FAIL or inclusion of `foreman`.

Same for all the tests in this PR.

tests/foreman/maintain/test_service.py
```diff
@@ -0,0 +1,267 @@
...
+
+def test_positive_service_restart_without_hammer_config(missing_hammer_config, sat_maintain):
+    """Restart services using service restart when hammer config/credentials are missing
+
+    :id: c7518650-d72a-47b1-8d38-42b862f474fc
+
+    :steps:
+        1. Run satellite-maintain service restart
+
+    :expectedresults: Services restart should work even if hammer config/credentials are missing
```
jyejare, line 206: Thats weird behavior of satellite or FM, an attacker (Security) can simply run this command to shutdown the services without satellite credentials.

We should file a bug !

jyejare (changes requested):
ACK pending comments !

tests/foreman/maintain/test_service.py
Gauravtalreja1: f-m services will be independent to hammer credentials in `.hammer/cli.modules.d/foreman.yml`, its not related to satellite credentials :)

tests/foreman/maintain/test_service.py
Gauravtalreja1: Here Service.status also returns same thing as Service.stop, i don't think it will make any sense here

tests/foreman/maintain/test_service.py
Gauravtalreja1: Steps are quite straight forward here, we're stopping all services first then check status for all services

robottelo/cli/sm_service.py
Gauravtalreja1: Updated

tests/foreman/maintain/test_service.py
Gauravtalreja1: Nice suggestion, Thank you @jyejare

tests/foreman/maintain/test_service.py
jyejare: Read my comment again :)

tests/foreman/maintain/test_service.py
jyejare:
```suggestion
    assert 'fail' not in result.stdout.lower()
    assert '<something >foreman <something>' in result.stdout 
```
Be specific with assertions and care about casing in assertions .

tests/foreman/maintain/test_service.py
jyejare: Yeah! But ultimately the impact remains same. I mean if the credentials are not being entertained by F-M then it should be now.

Not blocking your Pr for this though!

tests/foreman/maintain/test_service.py
jameerpathan111: @jyejare It's hard to tell why it could get in fail state and that's for all the foreman-maintain commands.
For this one, mostly it'd fail if a service is not running.
FM shows FAIL/OK for check/command we run and that's what we really care about. 
And if command/check doesn't fail then usually there's no unnecessary o/p so it's very unlikely that we'll get `FAIL` in output and that test will false pass.

```
| displaying dynflow-sidekiq@orchestrator          [FAIL]                       
| displaying foreman                               [FAIL]                       
| displaying httpd                                 [OK]
```

jyejare: @jameerpathan111 That clears everything !

Gauravtalreja1: Thank you @jameerpathan111 @jyejare

PR #10234, opened 2022-11-02
Added test for leapp wf

LadislavVasina1:
I have implemented a short test that checks if WF is working and upgrades rhel satellite.

### DIFF
--- a/conf/server.yaml.template
+++ b/conf/server.yaml.template
@@ -6,7 +6,7 @@ SERVER:
   VERSION:
     # The full release version (6.9.2)
     # Populate with satellite version
-    RELEASE: ''
+    RELEASE:
     # The snap version currently testing (if applicable)
     # SNAP:
     # The source of Satellite packages. Can be one of:
--- a/conf/server.yaml.template
+++ b/conf/server.yaml.template
@@ -5,8 +5,7 @@ SERVER:
   #  - replace.with.satellite.hostname
   VERSION:
     # The full release version (6.9.2)
-    # Populate with satellite version
-    RELEASE:
+    # RELEASE: populate with satellite version
     # The snap version currently testing (if applicable)
     # SNAP:
     # The source of Satellite packages. Can be one of:
--- a/conf/server.yaml.template
+++ b/conf/server.yaml.template
@@ -6,7 +6,7 @@ SERVER:
   VERSION:
     # The full release version (6.9.2)
     # Populate with satellite version
-    RELEASE:
+    RELEASE: ''
     # The snap version currently testing (if applicable)
     # SNAP:
     # The source of Satellite packages. Can be one of:
--- a/conf/server.yaml.template
+++ b/conf/server.yaml.template
@@ -5,7 +5,7 @@ SERVER:
   #  - replace.with.satellite.hostname
   VERSION:
     # The full release version (6.9.2)
-    # RELEASE: populate with satellite version
+    RELEASE:
     # The snap version currently testing (if applicable)
     # SNAP:
     # The source of Satellite packages. Can be one of:
--- a/conf/server.yaml.template
+++ b/conf/server.yaml.template
@@ -5,6 +5,7 @@ SERVER:
   #  - replace.with.satellite.hostname
   VERSION:
     # The full release version (6.9.2)
+    # Populate with satellite version
     RELEASE:
     # The snap version currently testing (if applicable)
     # SNAP:
--- a/tests/foreman/destructive/test_leapp.py
+++ b/tests/foreman/destructive/test_leapp.py
@@ -1,12 +1,12 @@
-"""Tests for leapp upgrade
+"""Tests for leapp upgrade of Satellite
 
 :CaseAutomation: Automated
 
 :CaseLevel: Component
 
-:CaseComponent: LEAPP
+:CaseComponent: Upgrade
 
-:Assignee: shwsingh
+:Assignee: lpramuk
 
 :TestType: Functional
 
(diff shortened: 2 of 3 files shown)


tests/foreman/destructive/test_leapp.py
```diff
@@ -0,0 +1,49 @@
+"""Tests for leapp upgrade
+
+:CaseAutomation: Automated
+
+:CaseLevel: Component
+
+:CaseComponent: LEAPP
```
lpramuk, line 7: -> Upgrade

tests/foreman/destructive/test_leapp.py
```diff
@@ -0,0 +1,49 @@
...
+
+:Assignee: shwsingh
```
lpramuk, line 9: lpramuk

tests/foreman/destructive/test_leapp.py
```diff
@@ -0,0 +1,49 @@
+"""Tests for leapp upgrade
```
lpramuk, line 1: Tests for leapp upgrade of Satellite

tests/foreman/destructive/test_leapp_satellite.py
```diff
@@ -0,0 +1,51 @@
...
+        3. Check results
+
+    :expectedresult:
+        1. Rhel version of the satellite has been updated
+
+    :parametrized: no
+    """
+    # Getting original RHEL version so we can increment it later in the test
+    orig_rhel_ver = target_sat.os_version.major
+    Broker().execute(job_template="satellite-leapp-upgrade", target_vm=target_sat.name)
```
Gauravtalreja1, line 45: is this workflow/JT? can't find this in the tower repo

tests/foreman/destructive/test_leapp_satellite.py
```diff
@@ -0,0 +1,51 @@
...
+    # Recreate the session object within a Satellite object after upgrading
+    target_sat.connect()
+    # Get RHEL version after upgrading
+    result = target_sat.execute('cat /etc/redhat-release | grep -Po "\\d"')
+    # Check if RHEL was upgraded
+    assert result.stdout[0] == str(orig_rhel_ver + 1), 'RHEL was not upgraded'
```
Gauravtalreja1, line 51: result.stdout isn't a list, you'll need `result.stdout.split('\n')[0]`
Also, you could check the major version using `target_sat.os_version.major`

tests/foreman/destructive/test_leapp_satellite.py
LadislavVasina1: see: 
https://infra-ansible-tower-01.infra.sat.rdu2.redhat.com/#/templates/job_template/233/details

tests/foreman/destructive/test_leapp_satellite.py
LadislavVasina1: ad 1) But it is a string and you can access char at index 0 inside that string so, in my opinion, it is correct.
ad 2) `os_version` is a cached property that is not changed after upgrading the RHEL.

tests/foreman/destructive/test_leapp_satellite.py
ogajduse: @Gauravtalreja1 See satlab-tower MR !665.

ogajduse:
> @LadislavVasina1 Good work, adding a few questions/suggestions.
> 
> It seems this test isn't required for 6.12/master, could you update the target branch for this PR accordingly?

I would leave the test in the master so it is not forgotten in the older branches. With Sat 6.14 or 6.15 we will just change the test parametrization in a way that we change `7` to `8` and `8` to `9` - that would allow us to test Leapp upgrade of Satellite from RHEL 8 to RHEL 9.
Also it would be very strange practice to leave the test in the branch for older release only and then resurrect the test again. The component owner of that test might change and the test in the old branch might be forgotten.
However, I see a point in your comment, @Gauravtalreja1. How do we want to treat such test cases? @jyejare

tests/foreman/destructive/test_leapp_satellite.py
```diff
@@ -40,10 +40,12 @@ def test_positive_leapp(target_sat):
 
     :parametrized: no
```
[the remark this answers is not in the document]
ogajduse: Out of curiosity, can you point me to the code where we make it default? I searched but had no luck finding it.

conf/server.yaml.template
```diff
@@ -5,7 +5,7 @@ SERVER:
   #  - replace.with.satellite.hostname
   VERSION:
     # The full release version (6.9.2)
-    # RELEASE: populate with satellite version
+    RELEASE:
```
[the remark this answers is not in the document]
ogajduse: see: https://github.com/SatelliteQE/robottelo/actions/runs/3411376897/jobs/5675494749#step:6:5937

ogajduse:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_leapp_satellite.py -v

ogajduse:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_leapp_satellite.py -v

LadislavVasina1:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_leapp_satellite.py -v

tests/foreman/destructive/test_leapp_satellite.py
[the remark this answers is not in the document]
adarshdubey-star: @ogajduse I think its here https://github.com/SatelliteQE/robottelo/blob/master/testimony.yaml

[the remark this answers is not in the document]
ogajduse: Here it only says what the choices are and whether the token is required. I see no mention about `default`
https://github.com/SatelliteQE/robottelo/blob/3cf798c39ad8fc98a85f3d999a3bab3ccb549978/testimony.yaml?rgh-link-date=2022-11-14T07%3A48%3A55Z#L172-L178

ogajduse (approved):

tests/foreman/destructive/test_leapp_satellite.py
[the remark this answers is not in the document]
Gauravtalreja1: @ogajduse yes, you're right, its not default, but preferably we use `:parametrized: yes` when test is parametrized

tests/foreman/destructive/test_leapp.py
lpramuk: ouch `Upgrades`

PR #10236, opened 2022-11-03
Port/refactor of f-m packages tests from testfm

Gauravtalreja1:
**Description:**
1. Port/refactor f-m packages subcommand tests from TestFM to robottelo
https://github.com/SatelliteQE/testfm/blob/master/tests/test_packages.py

It includes changes from https://github.com/SatelliteQE/robottelo/pull/10227, and it will be rebased when either PR is merged

### DIFF
--- a/tests/foreman/maintain/test_packages.py
+++ b/tests/foreman/maintain/test_packages.py
@@ -1,0 +1,290 @@
+"""Test class for satellite-maintain packages functionality
+
+:Requirement: foreman-maintain
+
+:CaseAutomation: Automated
+
+:CaseLevel: Component
+
+:CaseComponent: ForemanMaintain
+
+:Assignee: gtalreja
+
+:TestType: Functional
+
+:CaseImportance: Critical
+
+:Upstream: No
+"""
+import pytest
+
+from robottelo.config import settings
+from robottelo.utils.installer import InstallerCommand
+
+pytestmark = pytest.mark.destructive
+
+
+@pytest.mark.include_capsule
+def test_positive_fm_packages_lock_unlock(sat_maintain):
+    """Verify whether packages get locked and unlocked for satellite and capsule
+    using foreman_maintain
+
+    :id: d387d8be-10ad-4a62-aeff-3bc6a82e6bae
+
+    :parametrized: yes
+
+    :steps:
+        1. Run satellite-maintain packages lock
+        2. Run satellite-maintain packages status
+        3. Run satellite-maintain packages is-locked
+        4. Run satellite-maintain packages unlock
+        5. Run satellite-maintain packages status
+        6. Run satellite-maintain packages is-locked
+
+    :expectedresults: Packages get locked and unlocked using foreman_maintain
+    """
+    # Packages locking is enabled in installer for satellite, and disabled for capsule by default
+    installer = 'disabled' if sat_maintain.__class__.__name__ == 'Capsule' else 'enabled'
+
+    # Test Package lock command
+    result = sat_maintain.cli.Packages.lock()
+    assert 'FAIL' not in result.stdout
+    assert result.status == 0
+
+    result = sat_maintain.cli.Packages.status()
+    assert 'Packages are locked.' in result.stdout
+    assert f'Automatic locking of package versions is {installer} in installer.' in result.stdout
+    assert 'FAIL' not in result.stdout
+    assert result.status == 0
+
+    result = sat_maintain.cli.Packages.is_locked()
+    assert 'Packages are locked' in result.stdout
+    assert result.status == 0
+
+    # Test package unlock command
+    result = sat_maintain.cli.Packages.unlock()
+    assert 'FAIL' not in result.stdout
+    assert result.status == 0
+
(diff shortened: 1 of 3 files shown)


pondrejk:
@Gauravtalreja1 how this differs from https://github.com/SatelliteQE/robottelo/pull/10231 ?

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_packages.py

tests/foreman/maintain/test_packages.py
```diff
@@ -0,0 +1,318 @@
...
+
+    :id: 354d9940-10f1-4244-9079-fdbd24be49b3
+
+    :parametrized: yes
+
+    :steps:
+        1. Run satellite-maintain packages check-update
+        2. Run satellite-maintain packages update to update the walrus package.
+
+    :BZ: 1803975
```
shubhamsg199, line 265: I see a customer case with this BZ, better to add `:customerscenario: true`

shubhamsg199 (approved):
ACK,  minor suggestions

tests/foreman/maintain/test_packages.py
Gauravtalreja1: @shubhamsg199 updated

tests/foreman/maintain/test_packages.py
```diff
@@ -0,0 +1,293 @@
...
+
+    result = sat_maintain.cli.Packages.status()
+    assert 'Packages are locked.' in result.stdout
+    assert 'Automatic locking of package versions is enabled in installer.' in result.stdout
+    assert 'FAIL' not in result.stdout
+    assert result.status == 0
+
+    result = sat_maintain.cli.Packages.is_locked()
+    assert 'Packages are locked' in result.stdout
+    assert result.status == 0
```
jyejare, line 177: I think this portion is repeating, we should straight jump to the actual agenda of this test.

jyejare: Or you are checking the locking of packages double unnecessary .

tests/foreman/maintain/test_packages.py
```diff
@@ -0,0 +1,293 @@
...
+    assert 'FAIL' not in result.stdout
+    assert 'Nothing to do' not in result.stdout
+    assert 'Packages are locked.' in result.stdout
+    assert 'Automatic locking of package versions is enabled in installer.' in result.stdout
+
+    result = sat_maintain.cli.Packages.update(packages='zsh', options={'assumeyes': True})
+    assert result.status == 0
+    assert 'FAIL' not in result.stdout
+    assert 'Nothing to do' not in result.stdout
```
jyejare, line 194: I am not sure this is working because the step is installing zsh with the latest version and how we could expect it would update again !

This step should be 
```suggestion
    assert 'Nothing to do' in result.stdout
```

tests/foreman/maintain/test_packages.py
```diff
@@ -0,0 +1,293 @@
...
+    assert 'walrus-5.21-1' in result.stdout
+
+    @request.addfinalizer
+    def _finalize():
+        assert sat_maintain.execute('dnf remove -y walrus').status == 0
+        sat_maintain.execute('rm -rf /etc/yum.repos.d/custom_repo.repo')
+
+
+@pytest.mark.stubbed
+def test_positive_fm_packages_sat_installer(sat_maintain):
```
jyejare, line 280: Add :testAutomation: token for not automated.

jyejare (approved):
ACK pending non blocking comments!

tests/foreman/maintain/test_packages.py
Gauravtalreja1: yes, its just to exercise both f-m packages subcommand if they return correct output

tests/foreman/maintain/test_packages.py
Gauravtalreja1: if packages are locked we nothing to do in stdout, but here we use f-m to install package, so it works and we shouldn't get nothing to do, so this assertion is valid

tests/foreman/maintain/test_packages.py
Gauravtalreja1: @jyejare Updated, request to re-review

tests/foreman/maintain/test_packages.py
jyejare: I mean the above test case is already testing it we don't need to reconfirm it in this test!

tests/foreman/maintain/test_packages.py
jyejare: Is this Manual-Only or :NotAutomated:

There is a difference in both of these tokens:

- ManualOnly = That never suppose to Automate and will always be run manually
- NotAutomsted = Not automated currently but should be automated in future.

Screenshot 2022-11-16 at 5 40 15 PM

Gauravtalreja1: @jyejare Yes, I know, ManualOnly is correct here

Issue #10238, opened 2022-11-03
[Failed-AutoCherryPick] - [6.11.z] Host content ownership


Satellite-QE:
#### Auto-Cherry-Pick WorkFlow Failure:
- [Failed Cherrypick Action](https://github.com/SatelliteQE/robottelo/actions/runs/3386822582)
- [Parent Pull Request](https://github.com/SatelliteQE/robottelo/pull/10235)

jyejare:
@shweta83  Please do the manual cherrypick as the auto-cherrypicking is failed !

jyejare:
Ping @shweta83 !

jyejare:
Closing this issue as dependent PR is merged!

PR #10241, opened 2022-11-04
fixing auto assigning to correct user rather SATQE bot

omkarkhatavkar:
#### PR Description 

Currently, issues are getting assigned to the SATQE bot because it is used to create the PRs. e.g. https://github.com/SatelliteQE/robottelo/issues/10238

 With this PR we are not auto-assigning the PR to pr author and it will be used across further processes creating issues. This will resolve the issue.

### DIFF
--- a/.github/auto_assign.yml
+++ b/.github/auto_assign.yml
@@ -1,0 +1,1 @@
+addAssignees: author
--- a/.github/workflows/auto_cherry_pick.yml
+++ b/.github/workflows/auto_cherry_pick.yml
@@ -43,11 +43,14 @@ jobs:
           labels: |
             Auto_Cherry_Picked
             ${{ matrix.to_branch }}
+          assignees: "${{ github.event.pull_request.assignee.login }}"
 
   create-issue:
     runs-on: ubuntu-latest
     if: ${{ always() && contains(join(needs.*.result, ','), 'failure') }}
     needs: auto-cherry-pick
+    env:
+      to-branch: ${{ fromJson(needs.previous-branch.outputs.previous_branch) }}
     steps:
       - name: Create Issue on Failed Auto Cherrypick
         uses: dacbd/create-issue-action@main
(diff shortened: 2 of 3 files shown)


.github/workflows/auto_assign.yaml
```diff
@@ -0,0 +1,17 @@
+name: Auto Assign
```
jyejare, line 1: I really dont get why this action is needed !

I think this is one time process to add assignees to current opened PRs and we are done. Anyways henceforth the assignees will be added automatically.

omkarkhatavkar: We needed this because issues are getting created and assigned to SATQE bot user. This is why because we use SATQE bot user to create the autocherrypick PRs. If the PR is already assigned to Auther automatically, which is a valid case mostly, most people don't do that. As we assigned this automatically we can leverage that to create issues with those assignees.

pondrejk (approved):

.github/auto_assign.yml
```diff
@@ -0,0 +1 @@
+addAssignees: author
```
Gauravtalreja1, line 1: one question, why separate files for config? can't we include this in the action?

omkarkhatavkar: I tested that, and also looked at multiple repos. I haven to seen any around like that.  This is widely used GitHub action

jyejare (approved):

.github/auto_assign.yml
Gauravtalreja1: Okay, I see, I was looking into this action, found setting for reviewers, it would be nice to include reviewers as well to a PR by default

jyejare: @Gauravtalreja1 You mean reviewers as T1 and T2 by default ! Probably I wont because then people will be flooded with email every PR opened !

PR #10251, opened 2022-11-07
fix the broken create issue with adding needs

omkarkhatavkar:
fix the broken create issue with adding needs

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -64,13 +64,3 @@ jobs:
       - name: Analysis (git diff)
         if: failure()
         run: git diff
-
-  # This needed for to create the gh issue in case of failed auto0-cherry-pick
-  add-assignees:
-    name: Add author to assignee
-    if: "!contains(github.event.pull_request.labels.*.name, 'Auto_Cherry_Picked')"
-    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


pondrejk (approved):

Gauravtalreja1 (approved):

.github/workflows/pull_request.yml
```diff
@@ -73,4 +73,5 @@ jobs:
     steps:
```
jyejare, line 73: can you add following in this place, and try rerunning !
```suggestion
    permissions: write-all
    steps:
```

.github/workflows/pull_request.yml
```diff
@@ -9,6 +9,7 @@ env:
     PYCURL_SSL_LIBRARY: openssl
     ROBOTTELO_BUGZILLA__API_KEY: ${{ secrets.BUGZILLA_KEY }}
 
+permissions: write-all
```
jyejare, line 12:
```suggestion
```

omkarkhatavkar:
@jyejare  Although the newly introduced action is currently failing, it seems probable that is because it's trying to do something the target hasn't authorized. Post-merge, the target will have authorized the action, so in theory, will start working. If not, it's trivial to revert.

PR #10274, opened 2022-11-10
Fix Create Issue failing for sequence not expected

jyejare:
Fix the issue of GHA of `create_issue` failing with error `sequence not expected`.


jyejare:
Fixing this on urgent basis the way its failing for creating issues and extra effort needed to create a failed cherry-pick issue!

PR #10299, opened 2022-11-15
Cherry-picking past two Discovery commits to 6.11

Griffin-Sullivan:
One commit added the first new Discovery test for actually discovering and provisioning a host. The second commit was for cleaning up LibvirtDiscovery as it is deprecated.


pytest_fixtures/core/broker.py
```diff
@@ -173,3 +173,91 @@ def module_capsule_configured(module_capsule_host, module_target_sat):
     """Configure the capsule instance with the satellite from settings.server.hostname"""
     module_capsule_host.capsule_setup(sat_host=module_target_sat)
     yield module_capsule_host
+
+
+@pytest.fixture(scope='module')
+def module_capsule_configured_mqtt(module_capsule_configured):
```
Gauravtalreja1, line 179: these mqtt fixtures aren't intended for 6.11.z, so I think you should add only `module_discovery_sat`

Gauravtalreja1 (changes requested):

pytest_fixtures/core/broker.py
Griffin-Sullivan: Yeah I don't know why I did that haha

Gauravtalreja1 (approved):

jyejare (approved):

PR #10346, opened 2022-11-23
Automation coverage for BZ 2077824

Gauravtalreja1:
Description:
1. Adding automation coverage for [BZ 2077824](https://bugzilla.redhat.com/show_bug.cgi?id=2077824)

Test Results:
```
(.robottelo) ➜  robottelo git:(automate-2077824) ✗ pytest -q --disable-pytest-warnings tests/foreman/cli/test_capsule.py -k test_positive_search_nondefault_proxy
.                                                                                                                                                     [100%]
1 passed, 28 deselected, 1 warning in 47.79s
```

### DIFF
--- a/tests/foreman/api/test_smartproxy.py
+++ b/tests/foreman/api/test_smartproxy.py
@@ -17,8 +17,8 @@
 :Upstream: No
 """
 import pytest
+from fauxfactory import gen_string
 from fauxfactory import gen_url
-from nailgun import entities
 from requests import HTTPError
 
 from robottelo.utils.datafactory import parametrized
@@ -35,7 +35,7 @@ def module_proxy_attrs(module_target_sat):
     Every Satellite has a built-in smart proxy, so searching for an
     existing smart proxy should always succeed.
     """
-    smart_proxy = entities.SmartProxy().search(
+    smart_proxy = module_target_sat.api.SmartProxy().search(
         query={'search': f'url = {module_target_sat.url}:9090'}
     )
     # Check that proxy is found and unpack it from the list
@@ -57,7 +57,7 @@ def _create_smart_proxy(request, target_sat, **kwargs):
 
 @pytest.mark.skip_if_not_set('fake_capsules')
 @pytest.mark.tier1
-def test_negative_create_with_url():
+def test_negative_create_with_url(target_sat):
     """Proxy creation with random URL
 
     :id: e48a6260-97e0-4234-a69c-77bbbcde85d6
@@ -69,7 +69,7 @@ def test_negative_create_with_url():
     """
     # Create a random proxy
     with pytest.raises(HTTPError) as context:
-        entities.SmartProxy(url=gen_url(scheme='https')).create()
+        target_sat.api.SmartProxy(url=gen_url(scheme='https')).create()
     assert 'Unable to communicate' in context.value.response.text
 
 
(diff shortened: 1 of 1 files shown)


vsedmik (approved):
The CICQ failure seems unrelated, looks good to me.

tests/foreman/cli/test_capsule.py
```diff
@@ -382,3 +386,25 @@ def test_positive_enable_bmc():
...
+    :id: caf51662-6b4e-11ed-baba-2b9d7b368002
+
+    :expectedresults: Non-default proxy can be searched
+
+    :BZ: 2077824
+
+    :customerscenario: true
+    """
+    proxy = _make_proxy(request, target_sat, options={'name': gen_string('alpha')})
+    path = target_sat.api.Capsule().path() + '?search=id+%21%3D+1'
```
adarshdubey-star, line 405: Non-blocking, what are we concatenating here with the path? and why do we need that?

tests/foreman/cli/test_capsule.py
```diff
@@ -382,3 +386,25 @@ def test_positive_enable_bmc():
...
+    response = client.get(path, auth=get_credentials(), verify=False)
```
adarshdubey-star, line 406: Does the verification need to be true here taking auth into account?

tests/foreman/cli/test_capsule.py
Gauravtalreja1: @adarshdubey-star its a search string used in path to find capsules with id!=1, as mentioned in bz

tests/foreman/cli/test_capsule.py
Gauravtalreja1: for API requests we need false as we don't have ssl certs for sat

adarshdubey-star (approved):
Ack, please check CICQ failures.

tests/foreman/cli/test_capsule.py
jyejare, line 406: Why does this test is written in CLI test module rather than API?

Asked because all the steps and verification you have been dong here with API.

vsedmik: The API counterpart would be in `tests/foreman/api/test_smartproxy.py`

Gauravtalreja1: yes, there is no separate module present yet for API yet

tests/foreman/cli/test_capsule.py
```diff
@@ -382,3 +386,25 @@ def test_positive_enable_bmc():
...
+    assert response.status_code == 200
+    assert len(json.loads(response.text)['results']) == 1
+    assert json.loads(response.text)['results'][0]['name'] == proxy.name
+    assert json.loads(response.text)['results'][0]['url'] == proxy.url
```
ogajduse, line 410: What about using nailgun as we do it in the other modules?
```suggestion
    capsule_search = target_sat.api.Capsule().search(query={'search': 'id != 1'})
    assert len(capsule_search) == 1
    assert capsule_search[0].name == proxy.name
    assert capsule_search[0].url == proxy.url
```

Gauravtalreja1: @ogajduse it was not working for me earlier somehow, its working now, updated

tests/foreman/cli/test_capsule.py
Gauravtalreja1: updated

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/api/test_smartproxy.py -k test_positive_search_nondefault_proxy

jyejare (approved):

PR #9481, opened 2022-04-06
Update hammer-cli options for 6.11 

Gauravtalreja1:
**Description:**
1. Update `hammer-cli` options for 6.11 in hammer_commands.json, generated from running `scripts/hammer_command_tree.py`
2. Fixing `scripts/hammer_command_tree.py` by adding a hostname arg to ssh.command for runnning locally.
3. Removing `.splitlines()` from parse_help helper, which was added after moving from paramiko to ssh2-python.
4. Fixing `cli/test_hammer.py` which wasn't returning all available differences earlier, here's all available differences,

```
>           pytest.fail(format_commands_diff(differences))
E           Failed: hammer ansible roles
E             Added subcommands:
E               * sync
E
E           hammer ansible roles sync (new command)
E             Added options:
E               * proxy-id
E               * organization-id
E               * organization
E               * location-id
E               * location
E               * organization-title
E               * help
E               * role-names
E               * location-title
E
E           hammer architecture create
E             Added options:
E               * organization-id
E               * organization
E               * location-id
E               * organization-title
E               * location
E               * location-title
E
E           hammer architecture delete
E             Added options:
E               * organization-id
E               * organization
E               * location-id
E               * organization-title
E               * location
E               * location-title
E
E           hammer architecture info
E             Added options:
E               * organization-id
E               * organization
E               * location-id
E               * organization-title
E               * location
E               * location-title
E
E           hammer architecture list
E             Added options:
E               * organization-id
E               * organization
E               * location-id
E               * organization-title
E               * location
E               * location-title
E
E           hammer architecture update
E             Added options:
E               * organization-id
E               * organization
E               * location-id
E               * organization-title
E               * location
E               * location-title
E
E           hammer bookmark create
E             Added options:
E               * organization-id
E               * organization
E               * location-id
E               * organization-title
E               * location
E               * location-title
E
E           hammer bookmark delete
E             Added options:
E               * organization-id
E               * organization
E               * location-id
E               * organization-title
E               * location
E               * location-title
E
E           hammer bookmark info
E             Added options:
E               * organization-id
E               * organization
E               * location-id
E               * organization-title
E               * location
E               * location-title
E
E           hammer bookmark list
E             Added options:
E               * organization-id
E               * organization
E               * location-id
E               * organization-title
E               * location
E               * location-title
E
E           hammer bookmark update
E             Added options:
E               * organization-id
E               * organization
E               * location-id
E               * organization-title
E               * location
E               * location-title
E
E           hammer capsule
[...]

PR #10349, opened 2022-11-24
Add basic tests for dependencies

JacobCallahan:
These check some basic functionality for our top-level dependencies. They are not meant to be exhaustive, but rather to catch any obvious changes in behavior.
I also added a helper script for robottelo's config files. The initial helper allows a user to symlink their configs from another directory. We should expand on this script later for config migration etc.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -56,6 +56,9 @@ jobs:
           pytest --collect-only --setup-plan --disable-pytest-warnings -n 2 -m pre_upgrade tests/upgrades/
           pytest --collect-only --setup-plan --disable-pytest-warnings -n 2 -m post_upgrade tests/upgrades/
 
+      - name: Run Robottelo's Tests
+        run: pytest -sv tests/robottelo/
+
       - name: Make Docs
         run: |
           make test-docstrings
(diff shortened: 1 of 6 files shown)


JacobCallahan:
Now would be a good time to add back in the robottelo tests to our GitHub checks. Or at the very least, the dependency tests.
Hold off on merge

JacobCallahan:
@Gauravtalreja1 I commented out a test that is failing and may be related to #9481

mshriver (approved):

tests/robottelo/test_hammer.py
```diff
@@ -109,189 +109,189 @@ def test_parsed_json_match_parsed_csv(self):
         assert hammer.parse_json(json_output) == hammer.parse_csv(csv_ouput_lines)[0]
 
 
-class TestParseHelp:
-    """Tests for parsing hammer help output"""
+# class TestParseHelp:
```
Gauravtalreja1, line 112: @JacobCallahan Fixing this test here in https://github.com/SatelliteQE/robottelo/pull/10375, so this failing test can be uncommented now

Gauravtalreja1 (changes requested):

jyejare:
@JacobCallahan This needs conflicts to resolve.

tests/robottelo/test_hammer.py
JacobCallahan: @Gauravtalreja1 updated, thanks for addressing that.

Gauravtalreja1 (approved):

PR #10351, opened 2022-11-25
Skip manifest import when running test against stage console.redhat.com

jameerpathan111:


ogajduse (approved):

ogajduse (approved):

conf/rh_cloud.yaml.template
```diff
@@ -2,3 +2,4 @@ RH_CLOUD:
   INSTALL_RHC: false
   ORGANIZATION: org_name
   ACTIVATION_KEY: ak_name
+  CRC_ENV: "prod"
```
Gauravtalreja1, line 5: As there are only two possible env's rn, prod, and stage, so we can also make this a boolean, and this will require an updated on satqe-jenkins repo too, non-blocking :)
```suggestion
  CRC_PROD: True
```

tests/foreman/ui/test_rhc.py
```diff
@@ -60,12 +60,13 @@ def fixture_setup_rhc_satellite(request, module_target_sat, module_rhc_org):
     """Create Organization and activation key after successful test execution"""
     yield
     if request.node.rep_call.passed:
-        manifests_path = module_target_sat.download_file(
-            file_url=settings.fake_manifest.url['default']
-        )[0]
-        module_target_sat.cli.Subscription.upload(
-            {'file': manifests_path, 'organization-id': module_rhc_org.id}
-        )
+        if settings.rh_cloud.crc_env == "prod":
```
Gauravtalreja1, line 63:
```suggestion
        if settings.rh_cloud.crc_env == 'prod':
```

Gauravtalreja1 (commented):
Ack, with a suggestion

jyejare (approved):

conf/rh_cloud.yaml.template
jyejare: We might need an extra calculation in CI to set this to True/False when running the job instead of current passing fwd whatever we have set in the parameter.

jameerpathan111: Thanks for the suggestion but I prefer the existing approach.

tests/foreman/ui/test_rhc.py
jameerpathan111: Updated.

PR #10378, opened 2022-12-01
added manifester-related files to gitignore

pondrejk:
I wouldn't like to accidentally push these one day. @synkd please confirm it is correct and safe to do.


synkd (approved):
Thanks for opening this, @pondrejk. I totally agree that these should be in gitignore.

Griffin-Sullivan (approved):

Gauravtalreja1 (approved):

PR #10405, opened 2022-12-07
[6.11.z] Add basic tests for dependencies

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/10349

These check some basic functionality for our top-level dependencies. They are not meant to be exhaustive, but rather to catch any obvious changes in behavior.
I also added a helper script for robottelo's config files. The initial helper allows a user to symlink their configs from another directory. We should expand on this script later for config migration etc.


JacobCallahan (approved):
Added in missing commit and added back in the pexpect requirement since it still exists in this branch

PR #10424, opened 2022-12-13
Add test for the Capsule sync deadlock issue

vsedmik:
This PR adds coverage for [BZ#2062526](https://bugzilla.redhat.com/show_bug.cgi?id=2062526)

For this scenario to run we need ~400 GiB storage on Capsule with RHEL7 repo (or ~120 GiB with RHSCL repo), so I had to use larger `deploy_flavor` for the Capsule.

Originally the scenario was intended to use RHEL7 repo, but due to another [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=2122872) causing OOM extinction and sync task failure, I decided to go with half-sized RHSCL, which was capable to cause the deadlock too.

If/when the gunicorn BZ is resolved, we can get back to the RHEL7 repo and tweak the test for higher load.

### DIFF
--- a/conf/capsule.yaml.template
+++ b/conf/capsule.yaml.template
@@ -1,7 +1,4 @@
 CAPSULE:
-  # Capsule name to be used in the CapsuleVirtualMachine class (will be deprecated)
-  SATELLITE_VERSION_TAG: "@jinja {{this.robottelo.satellite_version | replace('.', '')}}"
-  INSTANCE_NAME: "@format qe-sat{this[capsule].satellite_version_tag}-rhel7-tierX-capsule"
   VERSION:
     # The full release version (6.9.2)
     RELEASE:  # populate with capsule version
@@ -13,6 +10,6 @@ CAPSULE:
     # The base os rhel version where the capsule installed
     # RHEL_VERSION:
   # The Ansible Tower workflow used to deploy a capsule
-  DEPLOY_WORKFLOW: deploy-sat-capsule
+  DEPLOY_WORKFLOW: deploy-capsule
   # Dictionary of arguments which should be passed along to the deploy workflow
-  # DEPLOY_ARGUMENTS:
+  DEPLOY_ARGUMENTS:
--- a/conf/flavors.yaml.template
+++ b/conf/flavors.yaml.template
@@ -1,0 +1,9 @@
+FLAVORS:
+  # 6 CPU, 24 GB RAM, 100 GB disk
+  DEFAULT: satqe-ssd.standard.std
+  # 16 CPU, 32 GB RAM, 160 GB disk
+  LARGE: satqe-ssd.standard.xxxl
+  # 6 CPU, 24 GB RAM, 500 GB disk
+  UPGRADE: satqe-ssd.upgrade.std
+  # 8 CPU, 32 GB RAM, 500 GB disk
+  CUSTOM_DB: satqe-ssd.customerdb.std
--- a/pytest_fixtures/core/sat_cap_factory.py
+++ b/pytest_fixtures/core/sat_cap_factory.py
@@ -44,6 +44,15 @@ def capsule_host(capsule_factory):
     Broker(hosts=[new_cap]).checkin()
 
 
+@pytest.fixture
+def large_capsule_host(capsule_factory):
+    """A fixture that provides a Capsule based on config settings"""
+    new_cap = capsule_factory(deploy_flavor=settings.flavors.custom_db)
+    yield new_cap
+    new_cap.teardown()
+    Broker(hosts=[new_cap]).checkin()
+
+
 @pytest.fixture(scope='module')
 def module_capsule_host(capsule_factory):
     """A fixture that provides a Capsule based on config settings"""
@@ -67,6 +76,13 @@ def capsule_configured(capsule_host, target_sat):
     """Configure the capsule instance with the satellite from settings.server.hostname"""
     capsule_host.capsule_setup(sat_host=target_sat)
     yield capsule_host
+
+
+@pytest.fixture
+def large_capsule_configured(large_capsule_host, target_sat):
+    """Configure the capsule instance with the satellite from settings.server.hostname"""
+    large_capsule_host.capsule_setup(sat_host=target_sat)
+    yield large_capsule_host
 
 
 @pytest.fixture(scope='module')
--- a/tests/foreman/destructive/test_capsulecontent.py
+++ b/tests/foreman/destructive/test_capsulecontent.py
@@ -1,0 +1,93 @@
+"""Capsule-Content related tests, which require destructive Satellite
+
+:Requirement: Capsule-Content
+
+:CaseAutomation: Automated
+
+:CaseLevel: System
+
+:CaseComponent: Capsule-Content
+
+:team: Phoenix-content
+
+:TestType: Functional
+
+:CaseImportance: High
+
+:Upstream: No
+"""
+import pytest
+from fauxfactory import gen_alpha
+
+from robottelo import constants
+
+pytestmark = [pytest.mark.destructive]
+
+
+@pytest.mark.tier4
+@pytest.mark.skip_if_not_set('capsule')
+def test_positive_sync_without_deadlock(
+    target_sat, large_capsule_configured, function_entitlement_manifest_org
+):
+    """Synchronize one bigger repo published in multiple CVs to a blank Capsule.
+    Assert that the sync task succeeds and no deadlock happens.
+
+    :id: 91c6eec9-a582-46ea-9898-bdcaebcea2f0
+
+    :setup:
+        1. A blank external capsule that has not been synced yet (!) with immediate download
+           policy and running multiple (4 and more) pulpcore workers.
+
+    :steps:
+        1. Sync one bigger repository to the Satellite.
+        2. Create a Content View, add the repository and publish it.
+        3. Create several copies of the CV and publish them.
+        4. Add the Library LCE to the Capsule.
+        5. Sync the Capsule.
+
+    :expectedresults:
+        1. Sync passes without deadlock.
+
+    :customerscenario: true
+
+    :BZ: 2062526
+
+    """
+    # Note: As of now BZ#2122872 prevents us to use the originally intended RHEL7 repo because
+    # of a memory leak causing Satellite OOM crash in this scenario. Therefore, for now we use
+    # smaller RHSCL repo instead, which was also capable to hit the deadlock issue, regardless
+    # the lower rpms count. When the BZ is fixed, reconsider upscale to RHEL7 repo or similar.
+    repo_id = target_sat.api_factory.enable_rhrepo_and_fetchid(
+        basearch='x86_64',
+        org_id=function_entitlement_manifest_org.id,
+        product=constants.REPOS['rhscl7']['product'],
+        repo=constants.REPOS['rhscl7']['name'],
+        reposet=constants.REPOSET['rhscl7'],
+        releasever=constants.REPOS['rhscl7']['releasever'],
+    )
+    repo = target_sat.api.Repository(id=repo_id).read()
+    repo.sync(timeout='60m')
+
+    cv = target_sat.publish_content_view(function_entitlement_manifest_org, repo)
+
+    for i in range(4):
+        copy_id = target_sat.api.ContentView(id=cv.id).copy(data={'name': gen_alpha()})['id']
+        copy_cv = target_sat.api.ContentView(id=copy_id).read()
+        copy_cv.publish()
+
+    proxy = large_capsule_configured.nailgun_smart_proxy.read()
+    proxy.download_policy = 'immediate'
+    proxy.update(['download_policy'])
+
+    lce = target_sat.api.LifecycleEnvironment(
+        organization=function_entitlement_manifest_org
+    ).search(query={'search': f'name={constants.ENVIRONMENT}'})[0]
+    large_capsule_configured.nailgun_capsule.content_add_lifecycle_environment(
+        data={'environment_id': lce.id}
+    )
+    result = large_capsule_configured.nailgun_capsule.content_lifecycle_environments()
+    assert len(result['results']) == 1
+    assert result['results'][0]['id'] == lce.id
+
+    sync_status = large_capsule_configured.nailgun_capsule.content_sync(timeout='90m')
+    assert sync_status['result'] == 'success', 'Capsule sync task failed.'
--- a/pytest_fixtures/core/contenthosts.py
+++ b/pytest_fixtures/core/contenthosts.py
@@ -224,7 +224,7 @@ def oracle_host(request, version):
 def sat_ready_rhel(request):
     deploy_args = {
         'deploy_rhel_version': request.param,
-        'deploy_flavor': 'satqe-ssd.standard.std',
+        'deploy_flavor': settings.flavors.default,
         'promtail_config_template_file': 'config_sat.j2',
         'workflow': 'deploy-rhel',
     }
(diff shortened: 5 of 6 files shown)


jyejare (changes requested):
Missing Branch labels in this PR to auto-cherrypick!

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/api/test_contentmanagement.py::TestCapsuleContentManagement::test_positive_sync_without_deadlock

vsedmik:
@JacobCallahan I'm affraid this scenario can't be run against an OSP-deployed Capsule. It requires ~500GB storage on the Capsule, which was done throuhg `increase-storage` WF on RHV, but AFAIK this is not supported on OSP.

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/api/test_capsulecontent.py::TestCapsuleContentManagement::test_positive_sync_without_deadlock

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_capsulecontent.py::test_positive_sync_without_deadlock

vsedmik:
Just to clarify, now the storage requirement is fulfilled by `deploy-flavor` rather than the original `increase-storage` WF.
The PRT seems to be passing ok.

tests/foreman/destructive/test_capsulecontent.py
```diff
@@ -0,0 +1,110 @@
...
+            }
+        )[0]
+        .read()
+    )
+    large_capsule_configured.nailgun_capsule.content_add_lifecycle_environment(
+        data={'environment_id': lce.id}
+    )
+    result = large_capsule_configured.nailgun_capsule.content_lifecycle_environments()
+    assert len(result['results']) == 1
+    assert lce.id in [capsule_lce['id'] for capsule_lce in result['results']]
```
lhellebr, line 107: Why is this a loop when `len(result['results']) == 1`?

lhellebr (changes requested):

pytest_fixtures/core/sat_cap_factory.py
```diff
@@ -46,6 +46,15 @@ def capsule_host(capsule_factory):
     Broker(hosts=[new_cap]).checkin()
 
 
+@pytest.fixture
+def large_capsule_host(capsule_factory):
+    """A fixture that provides a Capsule based on config settings"""
+    new_cap = capsule_factory(deploy_flavor='satqe-ssd.customerdb.std')
```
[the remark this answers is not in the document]
vsedmik: Well, it already is elsewhere. But I can see the point.
What would you recommend instead? Move it to config?

tests/foreman/destructive/test_capsulecontent.py
```diff
@@ -0,0 +1,110 @@
...
+
+    :customerscenario: true
+
+    :BZ: 2062526
+
+    """
+    workers_count = int(
+        large_capsule_configured.execute('ps -u pulp | grep pulpcore-worker | wc -l').stdout
+    )
+    assert workers_count >= 4, 'Not enough pulpcore workers running to test the scenario.'
```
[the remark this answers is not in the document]
vsedmik: I feel reluctant to configure it directly as requirements may change in the future. Right now we always have more than 4 CPUs on Capsule and I think the count will rather grow than decline. We can throw this away, or make a skip instead. You are right that the test failure is inappropriate in this case.

tests/foreman/destructive/test_capsulecontent.py
```diff
@@ -0,0 +1,110 @@
...
+    :id: 91c6eec9-a582-46ea-9898-bdcaebcea2f0
+
+    :setup:
+        1. A blank external capsule that has not been synced yet (!) with immediate download
+           policy and running multiple (4 and more) pulpcore workers.
+
+    :steps:
+        1. Sync one bigger repository to the Satellite.
+        2. Create a Content View, add the repository and publish it.
+        3. Create several copies of the CV and publish them.
```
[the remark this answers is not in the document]
vsedmik: It's usual to have same repo in multiple CVs I think. This way we increase the probability of a deadlock when same packages are synced at the same time. It is required to hit the issue.

tests/foreman/destructive/test_capsulecontent.py
```diff
@@ -0,0 +1,110 @@
...
+
+    :BZ: 2062526
+
+    """
+    workers_count = int(
+        large_capsule_configured.execute('ps -u pulp | grep pulpcore-worker | wc -l').stdout
+    )
+    assert workers_count >= 4, 'Not enough pulpcore workers running to test the scenario.'
+
+    # Note: As of now BZ#2122872 prevents us to use the originally intended RHEL7 repo because
```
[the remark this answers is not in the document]
vsedmik: When that BZ is fixed we would need to tweak the test anyway, as we don't know how would it affect the resources consumption (how good or bad would the fix be).

pytest_fixtures/core/sat_cap_factory.py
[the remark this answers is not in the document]
lhellebr: Yes

vsedmik:
Moved the `deploy_flavor` to config (see jenkins MR 1010), plus updated capsule.yaml.template and addressed @lhellebr 's comments.

lhellebr (approved):

tests/foreman/destructive/test_capsulecontent.py
```diff
@@ -0,0 +1,111 @@
...
+        reposet=constants.REPOSET['rhscl7'],
+        releasever=constants.REPOS['rhscl7']['releasever'],
+    )
+    repo = target_sat.api.Repository(id=repo_id).read()
+    repo.sync(timeout='60m')
+
+    cv = target_sat.api.ContentView(
+        organization=function_entitlement_manifest_org, repository=[repo]
+    ).create()
+    cv.publish()
```
ogajduse, line 80:
```suggestion
    cv = target_sat.publish_content_view(function_entitlement_manifest_org, repo)
```

tests/foreman/destructive/test_capsulecontent.py
```diff
@@ -0,0 +1,111 @@
...
+
+    for i in range(4):
+        copy_id = target_sat.api.ContentView(id=cv.id).copy(data={'name': gen_alpha()})['id']
+        copy_cv = target_sat.api.ContentView(id=copy_id).read()
+        copy_cv.publish()
+
+    proxy = target_sat.api.SmartProxy(id=large_capsule_configured.nailgun_capsule.id).read()
```
ogajduse, line 87:
```suggestion
    proxy = large_capsule_configured.nailgun_smart_proxy.read()
```

conf/capsule.yaml.template
```diff
@@ -13,6 +10,8 @@ CAPSULE:
     # The base os rhel version where the capsule installed
     # RHEL_VERSION:
   # The Ansible Tower workflow used to deploy a capsule
-  DEPLOY_WORKFLOW: deploy-sat-capsule
+  DEPLOY_WORKFLOW: deploy-capsule
   # Dictionary of arguments which should be passed along to the deploy workflow
-  # DEPLOY_ARGUMENTS:
+  DEPLOY_ARGUMENTS:
+    # Flavor for big capsules deployment
+    LARGE_FLAVOR: satqe-ssd.disk.xxl
```
ogajduse, line 17: Your change would add `large_flavor: satqe-ssd.disk.xxl` extra variable to each and every capsule deployment.
Consider defining a list of available flavors somewhere in settings, not necessarily in `capsule.yaml` or in constants, as it could be used by other host types.
```suggestion
  DEPLOY_FLAVORS:
    LARGE: satqe-ssd.disk.xxl  # Flavor for big capsules deployment
  DEPLOY_ARGUMENTS:
```
This is the only blocking comment in this review.

ogajduse (changes requested):

conf/capsule.yaml.template
vsedmik: Great catch! I'll update update it in the comming week.
Thanks @ogajduse !

conf/capsule.yaml.template
```diff
@@ -1,7 +1,4 @@
 CAPSULE:
-  # Capsule name to be used in the CapsuleVirtualMachine class (will be deprecated)
-  SATELLITE_VERSION_TAG: "@jinja {{this.robottelo.satellite_version | replace('.', '')}}"
-  INSTANCE_NAME: "@format qe-sat{this[capsule].satellite_version_tag}-rhel7-tierX-capsule"
```
jyejare, line 4: IS this not required anymore ?

tests/foreman/destructive/test_capsulecontent.py
```diff
@@ -0,0 +1,111 @@
...
+from fauxfactory import gen_alpha
+
+from robottelo import constants
+
+pytestmark = [pytest.mark.destructive]
+
+
+@pytest.mark.tier4
+@pytest.mark.skip_if_not_set('capsule')
+def test_positive_sync_without_deadlock(
```
jyejare, line 29: I dont see case importance changed for this test so its `high`, are you agree with that ?

tests/foreman/destructive/test_capsulecontent.py
```diff
@@ -0,0 +1,111 @@
...
+    :customerscenario: true
+
+    :BZ: 2062526
+
+    """
+    workers_count = int(
+        large_capsule_configured.execute('ps -u pulp | grep pulpcore-worker | wc -l').stdout
+    )
+    if workers_count < 4:
+        pytest.skip('Not enough pulpcore workers running to test the scenario.')
```
jyejare, line 60: Are pulpcore workers depends on the system configuration? if yes, then having a static count of workers does not really make sense ! 

We have to have it dynamic!

jyejare (commented):
ACK pending questions/comments

tests/foreman/destructive/test_capsulecontent.py
ogajduse: The importance of the whole module is set to `high`. Isn't it enough? Do we need to state the importance for every single test explicitly?

jyejare: @ogajduse No, I mean it is not about inheriting the importance from module level but I believe this could come under `Critical` severity as it belongs to capsule syncing and thats one of the vital thing capsule is made for.

So my real question is, would you like to change the importance to critical for this test ?

conf/capsule.yaml.template
vsedmik: I think it's not. I haven't found any references from within robottelo to use this settings.

tests/foreman/destructive/test_capsulecontent.py
vsedmik: I just respected the severity defined for the BZ. I don't oppose to set it higher, but maybe I would rather do that for some basic scenario like RH or KS repos sync to capsule in `tests/foreman/api/test_capsulecontent.py`, than this deadlock one.

tests/foreman/destructive/test_capsulecontent.py
vsedmik: The pulpcore workers count is derived from the CPUs count, from this point of view we could say it's dynamic. However, the number here defines the minimal threshold we need to hit the deadlock issue and it is rather static - we need at least some minimal count of workers to hit the issue and we don't really care if we have 2 or 20 more extra workers.

As I wrote in some comment before, in our deployments we always have more and I don't expect the count of CPUs needed for capsule to decline, so we could probably even remove this check.

jyejare: @vsedmik Go for removal of it or increase the count ?

ogajduse (approved):
Great work @vsedmik!

tests/foreman/destructive/test_capsulecontent.py
vsedmik: All right, I choose removal if you insist on it.

vsedmik:
Added `DEFAULT` flavor and referenced in `sat_ready_rhel` as per @ogajduse's request in the related Jenkins MR.

TODO: After the Jenkins MR is merged, run PRT for some `sat_ready_rhel` test before merging this.

pytest_fixtures/core/sat_cap_factory.py
```diff
@@ -44,6 +44,15 @@ def capsule_host(capsule_factory):
     Broker(hosts=[new_cap]).checkin()
 
 
+@pytest.fixture
+def large_capsule_host(capsule_factory):
+    """A fixture that provides a Capsule based on config settings"""
+    new_cap = capsule_factory(deploy_flavor=settings.flavors.custom_db)
```
Gauravtalreja1, line 50: Large flavor should have been used here instead of custom_db, or is this intended?
```suggestion
    new_cap = capsule_factory(deploy_flavor=settings.flavors.large)
```

conf/flavors.yaml.template
```diff
@@ -0,0 +1,9 @@
+FLAVORS:
+  # 6 CPU, 24 GB RAM, 100 GB disk
+  DEFAULT: satqe-ssd.standard.std
+  # 16 CPU, 32 GB RAM, 160 GB disk
```
Gauravtalreja1, line 4: I think we can keep this yaml as is (without .template) similar to supportability.yaml, wdyt @jyejare @JacobCallahan  ?

tests/foreman/destructive/test_capsulecontent.py
```diff
@@ -0,0 +1,102 @@
...
+        .search(
+            query={
+                'search': (
+                    f'name={constants.ENVIRONMENT} and '
+                    f'organization_id={function_entitlement_manifest_org.id}'
+                )
+            }
+        )[0]
+        .read()
+    )
```
Gauravtalreja1, line 93:
```suggestion
    lce = target_sat.api.LifecycleEnvironment(organization=function_entitlement_manifest_org).search(
            query={'search': f'name={constants.ENVIRONMENT}'}
    )[0]
```

vsedmik: We have one `Library` LCE in `default org` and another in `function_entitlement_manifest_org`, IMHO we need to be explicit here.

pytest_fixtures/core/sat_cap_factory.py
bherrin3: @vsedmik - specifically for DISK-only. There is a capability to deploy any VM and then increase the disk only with add-disk. I can't recall, off the top of my head, if that flavor was made specifically for the scope of this test.

If it was not and that isn't a satqe-ssd flavor, add-disk "methodoloy" should be used here.

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_capsulecontent.py::test_positive_sync_without_deadlock

pytest_fixtures/core/sat_cap_factory.py
vsedmik: Well, it's not DISK-only, the concurrent sync consumes other resources too.

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/installer/test_installer.py::test_satellite_and_capsule_installation

tests/foreman/destructive/test_capsulecontent.py
Gauravtalreja1: yes, just suggesting here to include search as part of target_sat.api.LifecycleEnvironment(organization=xyz) and not while querying

ogajduse: @Gauravtalreja1's suggestion preserves the organization in a form of an argument for `LifecycleEnvironment`. That should scope the `search()` to that organization only.
I believe that the API response will be the same with this patch. Am I correct?

vsedmik: Oh, I overlooked the org specified as LCE argument, looks nice! Let me pull it in.

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_capsulecontent.py::test_positive_sync_without_deadlock

PR #10436, opened 2022-12-15
No-Cherrypick by default

jyejare:
By default the AutoCherrypicked PRs to labeled branches need not to again decide for auto-cherrypicking. The decision has should have already been made in parent PR itself as to what branch the PR will be cherrypicked and assigned branch labels.

### DIFF
--- a/.github/workflows/auto_cherry_pick.yml
+++ b/.github/workflows/auto_cherry_pick.yml
@@ -56,6 +56,7 @@ jobs:
           labels: |
             Auto_Cherry_Picked
             ${{ matrix.label }}
+            No-CherryPick
           assignees: ${{ env.assignee }}
 
       - name: Add Parent PR's PRT comment to Auto_Cherry_Picked PR's


JacobCallahan (approved):

.github/workflows/auto_cherry_pick.yml
```diff
@@ -56,6 +56,7 @@ jobs:
           labels: |
             Auto_Cherry_Picked
             ${{ matrix.label }}
+            No-CherryPick
```
omkarkhatavkar, line 59: Instead of copying the label No-CherryPick, we could restrict the https://github.com/SatelliteQE/robottelo/blob/d7078f2aada41a8a583d2b37eb0021b24a9af7cc/.github/workflows/required_labels.yml#L7 for default branch(master). This will avoid even the adding label and no need to run the workflow as well

JacobCallahan: i like that

jyejare: We could not and should not because we are not always going to cherrypick just from master. We might remove something from 6.13 and in subsequent zStream branches but not in master.

So to support this scenario we have to maintain the ACP GHA in all branches. 

Its better to have some support than no support. :)

omkarkhatavkar (approved):

PR #10457, opened 2022-12-20
fixing the auto-merge issue

omkarkhatavkar:
fixing the https://github.com/SatelliteQE/robottelo/actions/runs/3740676253

also noticed the auto merging is getting triggered https://github.com/SatelliteQE/robottelo/actions/runs/3740710868 here. so avoiding that trigger

### DIFF
--- a/.github/workflows/automerge.yaml
+++ b/.github/workflows/automerge.yaml
@@ -8,10 +8,6 @@ on:
       - ready_for_review
     branches-ignore:
       - master
-  issue_comment:
-    types:
-      - created
-      - deleted
   pull_request_review:
     types:
       - submitted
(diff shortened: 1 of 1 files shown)


.github/workflows/automerge.yaml
```diff
@@ -34,16 +30,10 @@ jobs:
           direction: last
 
       - name: Fail automerge if PRT was not initiated
-        if: steps.find-prt-comment.outputs.comment-body == ""
+        if: steps.find-prt-comment.outputs.comment-body == ''
         run: |
           echo "::error PRT comment not added the PR"
 
-      - name: Wait for PRT checks to get initiated
```
jyejare, line 41: Isnt this needed anymore ?

omkarkhatavkar: we are not going to trigger the event using the comment hence, the user can take time and once PRT starts they can trigger just reloading the label

jyejare: But thats very manual process. If someone forgets to reload the label, your PR will be auto-merged without PRT running/passing.

omkarkhatavkar: if in that case we can add 5 mins delay for every occurrence

omkarkhatavkar: @jyejare if you are okay we can do this then?

omkarkhatavkar: I removed the issue_comment trigger as this only works in the default branch master, we don't want to run it here on the master. To avoid that I'm removing that.   see here  https://github.com/SatelliteQE/robottelo/actions/runs/3740710868

JacobCallahan (approved):
Merge pending resolution of @jyejare's thread

.github/workflows/automerge.yaml
jyejare: Sure! For now we can merge as is.

Later, we can find some GHA that waits for some GHA to start or we can build one if needed :)

jyejare (approved):

PR #10460, opened 2022-12-21
add delay for errata check and fix wait_for parameters

rplevka:
Asserting that the applied erratum is unlisted is hit and miss - we experience issues with timing, so I added `wait_for`.

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -162,18 +162,35 @@ class ContentHost(Host, ContentHostMixins):
     @property
     def nailgun_host(self):
         """If this host is subscribed, provide access to its nailgun object"""
-        if self.subscribed:
+        if self.identity.get('registered_to') == self.satellite.hostname:
             try:
                 host_list = self.satellite.api.Host().search(query={'search': self.hostname})[0]
             except Exception as err:
                 logger.error(f'Failed to get nailgun host for {self.hostname}: {err}')
                 host_list = None
             return host_list
+        else:
+            logger.warning(f'Host {self.hostname} not registered to {self.satellite.hostname}')
 
     @property
     def subscribed(self):
         """Boolean representation of a content host's subscription status"""
         return 'Status: Unknown' not in self.execute('subscription-manager status').stdout
+
+    @property
+    def identity(self):
+        """A Dictionary containing RHSM identity attributes of the host"""
+        id_output = self.execute('subscription-manager identity').stdout
+        id_dict = {}
+        if id_output:
+            id_dict = {
+                i.split(':')[0].replace(' ', '_'): i.split(': ')[1]
+                for i in id_output.split('\n')[:-1]
+            }
+            regged_to = self.subscription_config['server']['hostname']
+            if regged_to:
+                id_dict['registered_to'] = regged_to
+        return id_dict
 
     @property
     def ip_addr(self):
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -1571,8 +1571,18 @@ def test_positive_provision_baremetal_with_uefi_secureboot():
 
 
 @pytest.fixture(scope="function")
-def setup_custom_repo(target_sat, module_org, katello_host_tools_host):
+def setup_custom_repo(target_sat, module_org, katello_host_tools_host, request):
     """Create custom repository content"""
+
+    def restore_sca_setting():
+        """Restore the original SCA setting for module_org"""
+        module_org.sca_enable() if sca_enabled else module_org.sca_disable()
+
+    if module_org.sca_eligible().get('simple_content_access_eligible', False):
+        sca_enabled = module_org.simple_content_access
+        module_org.sca_disable()
+        request.addfinalizer(restore_sca_setting)
+
     # get package details
     details = {}
     if katello_host_tools_host.os_version.major == 6:
@@ -1798,13 +1808,14 @@ def test_positive_erratum_applicability(
                 if errata['installable'] == 'true'
             ],
             handle_exception=True,
-            fail_condition=True,
             timeout=300,
             delay=5,
         )
     except TimedOutError:
-        assert False, f"timed out waiting for erratum \"{setup_custom_repo['security_errata']}\""
-        " to disappear from the list"
+        raise TimedOutError(
+            f"Timed out waiting for erratum \"{setup_custom_repo['security_errata']}\""
+            " to disappear from the list"
+        )
 
 
 @pytest.mark.cli_katello_host_tools
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -26,6 +26,7 @@ from fauxfactory import gen_integer
 from fauxfactory import gen_ipaddr
 from fauxfactory import gen_mac
 from fauxfactory import gen_string
+from wait_for import TimedOutError
 from wait_for import wait_for
 
 from robottelo.cli.activationkey import ActivationKey
@@ -1773,7 +1774,7 @@ def test_positive_erratum_applicability(
     applicable_errata, _ = wait_for(
         lambda: Host.errata_list({'host-id': host_info['id']}),
         handle_exception=True,
-        failure_condition=[],
+        fail_condition=[],
         timeout=120,
         delay=5,
     )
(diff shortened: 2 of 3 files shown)


rplevka:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py -k test_positive_erratum_applicability

ogajduse (approved):
ACK pending PRT and pre-commit failures.

rplevka:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py -k test_positive_erratum_applicability

rplevka:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py -k test_positive_erratum_applicability

rplevka:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py -k test_positive_erratum_applicability

adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py::test_positive_erratum_applicability

tests/foreman/cli/test_host.py
```diff
@@ -1572,6 +1573,8 @@ def test_positive_provision_baremetal_with_uefi_secureboot():
 @pytest.fixture(scope="function")
 def setup_custom_repo(target_sat, module_org, katello_host_tools_host):
     """Create custom repository content"""
+    # disable SCA for the org
+    target_sat.api.Organization(id=module_org.id, simple_content_access=False).update(['simple_content_access'])
```
Gauravtalreja1, line 1577: can you try `module_org.sca_disable()` here instead of an update?

rplevka: so this actually doesn't work:
```
'{"displayMessage":"Current organization does not have a manifest imported.","errors":["Current organization does not have a manifest imported."]}'
```
but updating the attribute directly does

rplevka: I filed a BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=2159974

to workaround this, I decided to override a `module_org` fixture by a custom one, that creates the org with sca set to false.

rplevka:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py -k test_positive_erratum_applicability

robottelo/hosts.py
```diff
@@ -164,19 +164,36 @@ def satellite(self):
...
+        id_dict = {}
+        if id_output:
+            id_dict = {
+                i.split(':')[0].replace(' ', '_'): i.split(': ')[1]
+                for i in id_output.split('\n')[:-1]
+            }
+            regged_to_output = self.execute('subscription-manager config | grep " hostname"').stdout
+            if regged_to_output and len(regged_to_output.split('= ')) > 1:
+                id_dict['registered_to'] = regged_to_output.replace('\n', '').split('= ')[1]
+        return id_dict
```
Gauravtalreja1, line 195: If the host is attached to CDN it returns `[subscription.rhsm.example.com]` instead of just hostname for `registered_to`, can you check?

peterdragun (approved):
Ack, nice work!

robottelo/hosts.py
```diff
@@ -164,19 +164,36 @@ def satellite(self):
     @property
     def nailgun_host(self):
         """If this host is subscribed, provide access to its nailgun object"""
-        if self.subscribed:
+        if self.identity.get('registered_to') == self.satellite.hostname:
```
Gauravtalreja1, line 167: It is likely to fail if the host is registered to a capsule

rplevka: this is simply referring to the optional `satellite` attribute of the `ContentHost` class.
https://github.com/SatelliteQE/robottelo/blob/master/robottelo/hosts.py#L144
there doesn't seem to be any distinction made between satellite and capsule ATM.

I believe this is as closest as we can get to the real state as we can with the current implementation

tests/foreman/cli/test_host.py
```diff
@@ -66,6 +67,11 @@ def module_default_proxy(module_target_sat):
     return module_target_sat.cli.Proxy.list({'search': f'url = {module_target_sat.url}:9090'})[0]
 
 
+@pytest.fixture(scope='module')
+def module_org(module_target_sat):
+    return module_target_sat.api.Organization(simple_content_access=False).create()
```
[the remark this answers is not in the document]
rplevka: We can't simply add this to the test as a step, since the test uses another fixture, that calls `module_org`.
We probably have no other choice than wait till https://bugzilla.redhat.com/show_bug.cgi?id=2159974 is fixed,
or go with my solution (disable SCA for all tests in the module) as a temporary measure.

Funny fact: I need this patch to exist in the older branches, where SCA is not even a concern, however i probably won't be able to implement it before it lands here, in `master` branch

tests/foreman/cli/test_host.py
```diff
@@ -1787,11 +1793,23 @@ def test_positive_erratum_applicability(
...
+            in [
+                errata['erratum-id']
+                for errata in Host.errata_list({'host-id': host_info['id']})
+                if errata['installable'] == 'true'
+            ],
+            handle_exception=True,
+            fail_condition=True,
+            timeout=300,
+            delay=5,
+        )
```
ogajduse, line 1809: The following is more readable to me. I spent a few minutes on it until I realized `fail_condition=True`. Do we achieve the same behavior if we would apply the patch?
```suggestion
        applicable_erratum, _ = wait_for(
            lambda: setup_custom_repo["security_errata"]
            not in [
                errata['erratum-id']
                for errata in Host.errata_list({'host-id': host_info['id']})
                if errata['installable'] == 'true'
            ],
            handle_exception=True,
            timeout=300,
            delay=5,
        )
```

robottelo/hosts.py
ogajduse: I am going to test this scenario using the code from @rplevka's branch next week. I will let you know the result.

JacobCallahan:
@rplevka Can you resolve the merge conflict and revisit this PR? Thanks!

robottelo/hosts.py
ogajduse: @rplevka Now you can use:
```python
regged_to = self.subscription_config['server']['hostname']
if regged_to:
    id_dict['registered_to'] = regged_to_output.replace('\n', '').split('= ')[1]
```

tests/foreman/cli/test_host.py
rplevka: the bz is now resolved, i'm going to use originally proposed `module_org.sca_disable()` with some logic around it (finalizer, etc.)

rplevka:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py::test_positive_erratum_applicability[rhel8]

tests/foreman/cli/test_host.py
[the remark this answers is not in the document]
rplevka: the bugzilla is now fixed and i managed to control SCA for the org on the test level.

rplevka:
@JacobCallahan done.
@Gauravtalreja1 would you mind giving this another look? I believe I managed to address everything and the tests are passing now.

rplevka:
note about automatic cherry-picks:
I chose `6.13` label only on purpose, as the commits in the current form is only applicable in `6.13`. I will raise backport PRs to the older branches manually, after removing the `sca`-related bits

rplevka:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py::test_positive_erratum_applicability

robottelo/hosts.py
ogajduse: If a ContentHost is registered to a capsule, it shows up in the host list in Satellite, so we should be able to find it there.
If a Capsule is registered to Satellite, then `nailgun_host` for the capsule should exist too, but the question is whether we can determine the correct Satellite instance to search the Host entity in.
Answer: We surely can if Capsule's `satellite` property gives us the right Satellite instance.

@Gauravtalreja1 Could you correct me if I am wrong here?

ogajduse: After some testing, I see that if the host is registered to Capsule, `rhsm.conf` contains the Capsule hostname, not the Satellite one. #10547 will not save us, the condition needs to be tweaked.

pondrejk (approved):

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py::test_positive_erratum_applicability

PR #10493, opened 2023-01-06
CI checks with Python 3.11, removed 3.8

jyejare:
- Added CI checks for Python 3.11 and 3.10
- Removed 3.8 support

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, 3.9]
+        python-version: ['3.9', '3.10', '3.11']
     steps:
       - name: Checkout Robottelo
         uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/pull_request.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, 3.9]
+        python-version: ['3.9', '3.10', '3.11']
```
Gauravtalreja1, line 18: Since we already moving to py311, can we exclude py39 from this matrix too?

jyejare: Just to verify the backward compatibility with python 3.9, also few QEs and dependent tools(Nailgun/Airgun) still runs things on Python 3.9. 

Also, we validate 3 python versions compatibility at a time so lets keep 3.9!

Gauravtalreja1: Okay, make sense, soon we can move dependant tools to run on 3.11 too :)

Also, setup.py mentions 3.8, can you update it as well?

JacobCallahan (approved):

PR #10500, opened 2023-01-10
added the check to wait for PRT status using status API

omkarkhatavkar:
#### PR Description 

- Added checkpoint for PRT status using Status API 
- Added checkpoint for other check-runs 
- Marking Failure in case any of the above conditions fails. 
- Verbose logs to identify for the owner what has gone wrong.  

#### Verified and Tested: 
https://github.com/omkarkhatavkar/robottelo/actions/runs/3882465855/jobs/6622647642

### DIFF
--- a/.github/workflows/automerge.yaml
+++ b/.github/workflows/automerge.yaml
@@ -43,9 +43,68 @@ jobs:
         run: |
           echo "Waiting for ~ 10 mins, PRT to be initiated." && sleep 600
 
+      - name: Wait for other status checks to Pass
+        id: waitforstatuschecks
+        uses: lewagon/wait-on-check-action@v1.2.0
+        with:
+          ref: ${{ github.head_ref }}
+          repo-token: ${{ secrets.CHERRYPICK_PAT }}
+          wait-interval: 60
+          running-workflow-name: 'Automerge auto-cherry-picked pr'
+          allowed-conclusions: success,skipped
+
+      - name: is PRT check Passed ?
+        run: |
+          api_endpoint="https://api.github.com/repos/${{github.repository}}/statuses/${{ github.head_ref }}"
+
+          function get_status() {
+            curl -s "$api_endpoint"  | jq -r '.[0].state'
+          }
+
+          function get_context() {
+            curl -s "$api_endpoint" | jq -r '.[0].context'
+          }
+
+          statuses_length=$(curl -s $api_endpoint | jq 'length')
+          if [ $statuses_length -eq 0 ]; then
+              echo "PRT failed to start ! Stopping."
+              exit 1
+          fi
+
+          status=$(get_status)
+          context=$(get_context)
+
+          if [ "$context" != "Robottelo-Runner" ]; then
+            echo "::error Failed to get Robottelo-Runner status"
+            exit 1
+          fi
+
+          counter=0
+          echo "Waiting for PRT to complete....."
+          while [ "$status" != "success" ] && [ "$status" != "failure" ]; do
+            if [ $counter -gt 20 ]; then
+              echo "PRT Timeout"
+              exit 1
+            fi
+            sleep 300
+            status=$(get_status)
+            echo "Robottelo-Runner : $status"
+            counter=$((counter+1))
+          done
+
+          if [ "$status" == "success" ]; then
+            echo "PRT Passed Successfully!"
+          else
+            echo "Robottelo-Runner : $status"
+            echo "::error PRT failed"
+            exit 1
+          fi
+
+
       - id: automerge
         name: Auto merge of cherry-picked PRs.
         uses: "pascalgn/automerge-action@v0.15.5"
+        if: steps.waitforstatuschecks.outputs.status == 'success'
         env:
           GITHUB_TOKEN: "${{ secrets.CHERRYPICK_PAT }}"
           MERGE_LABELS: "AutoMerge_Cherry_Picked, Auto_Cherry_Picked"
@@ -59,4 +118,6 @@ jobs:
           if [ "${{ steps.automerge.outputs.mergeResult }}" == 'merged' ]; then
             echo "Pull request ${{ steps.automerge.outputs.pullRequestNumber }} is Auto Merged !"
           else
-            echo "::error Auto Merge of Pull request ${{ steps.automerge.outputs.pullRequestNumber }} is ${{steps.automerge.outputs.mergeResult}} !"
+            echo "::error Auto Merge for Pull request failed !"
+            exit 1
+          fi


.github/workflows/automerge.yaml
```diff
@@ -43,9 +43,47 @@ jobs:
...
+              fi
+              if [ "$COMMIT_STATUS" == "success" ]; then
+                  echo "PRT Passed Successfully!"
+              else
+                  echo "Robottelo-Runner : $COMMIT_STATUS"
+                  echo "::error PRT failed/timeout!" & exit 1
+              fi
+          fi
+
+      - name: 'Wait for other status checks to Pass'
```
adarshdubey-star, line 73: non-blocking
```suggestion
      - name: Wait for other status checks to Pass
```

.github/workflows/automerge.yaml
```diff
@@ -43,9 +43,47 @@ jobs:
         run: |
           echo "Waiting for ~ 10 mins, PRT to be initiated." && sleep 600
 
+      - name: is PRT check Passed ?
+        run: |
+          counter=0
+          STATUSES=$(curl -s https://api.github.com/repos/${{github.repository}}/statuses/${{ github.head_ref }}  | jq -r 'length')
```
adarshdubey-star, line 49: Just curious about what's happening here with the length ?

omkarkhatavkar: There could be instances where the PRT would not get triggered and the status API would be empty, in that case, we wanted to grace stop the execution. That's why I'm checking the length here

.github/workflows/automerge.yaml
```diff
@@ -43,9 +43,67 @@ jobs:
...
+            curl -s "$api_endpoint"  | jq -r '.[0].state'
+          }
+
+          function get_context() {
+            curl -s "$api_endpoint" | jq -r '.[0].context'
+          }
+
+          statuses_length=$(curl -s $api_endpoint | jq 'length')
+          if [ $statuses_length -eq 0 ]; then
+              echo "PRT failed to start ! Stopping."
```
jyejare, line 60: Not only PRT but this means no checks are started yet, Right ?

.github/workflows/automerge.yaml
```diff
@@ -43,9 +43,67 @@ jobs:
...
+              exit 1
+          fi
+
+          status=$(get_status)
+          context=$(get_context)
+
+          if [ "$context" != "Robottelo-Runner" ]; then
```
jyejare, line 67: I think its not Robottelo-Runner in context but there would be checks as well, so the `comparision` wont work !

.github/workflows/automerge.yaml
```diff
@@ -43,9 +43,67 @@ jobs:
...
+            echo "::error Failed to get Robottelo-Runner status"
+            exit 1
+          fi
+
+          counter=0
+          while [ "$status" != "success" ] && [ "$status" != "failure" ]; do
+            if [ $counter -gt 10 ]; then
```
jyejare, line 74: We should change the counter to bit higher number like 20 or more, because some times test run takes huge time to complete so 50 minutes wont be enough!

.github/workflows/automerge.yaml
```diff
@@ -43,9 +43,67 @@ jobs:
...
+              echo "PRT failed/timeout"
+              exit 1
+            fi
+            echo "Waiting for PRT to complete....."
```
jyejare, line 78: Move this line above while loop else this wll print in loops!

.github/workflows/automerge.yaml
```diff
@@ -43,9 +43,67 @@ jobs:
...
+            curl -s "$api_endpoint" | jq -r '.[0].context'
+          }
+
+          statuses_length=$(curl -s $api_endpoint | jq 'length')
+          if [ $statuses_length -eq 0 ]; then
+              echo "PRT failed to start ! Stopping."
+              exit 1
+          fi
+
+          status=$(get_status)
```
jyejare, line 64: I doubt about this really provides the status only for Robottelo-Runner (PRT) check. You url is pointing to all the statuses not just to PRT.

Even this is applicable for context below.

.github/workflows/automerge.yaml
```diff
@@ -43,9 +43,67 @@ jobs:
...
+
+          if [ "$status" == "success" ]; then
+            echo "PRT Passed Successfully!"
+          else
+            echo "Robottelo-Runner : $status"
+            echo "::error PRT failed"
+            exit 1
+          fi
+
+      - name: Wait for other status checks to Pass
```
jyejare, line 93: I think this should be moved above the PRT check because they will start and finish before PRT. Once we check for checks  then we will only left with PRT to finish.

.github/workflows/automerge.yaml
```diff
@@ -43,9 +43,67 @@ jobs:
...
+          ref: ${{ github.head_ref }}
+          repo-token: ${{ secrets.CHERRYPICK_PAT }}
+          wait-interval: 60
+          running-workflow-name: 'Wait for other status checks to Pass'
+          allowed-conclusions: success,skipped
+
       - id: automerge
         name: Auto merge of cherry-picked PRs.
         uses: "pascalgn/automerge-action@v0.15.5"
+        if: steps.waitforstatuschecks.outputs.status == 'success'
```
jyejare, line 106: You should also be able to echo the PRT check output using `set-output` in shell script. so that you check both of the checks non-PRT and PRT.

jyejare (changes requested):

.github/workflows/automerge.yaml
omkarkhatavkar: Yes, this means only PRT did not kick-off.

.github/workflows/automerge.yaml
omkarkhatavkar: all checks use the checks API, and PRT uses only the status API. This is just an extra layer of checking.

.github/workflows/automerge.yaml
omkarkhatavkar: Yeah may be this can be increased.

.github/workflows/automerge.yaml
omkarkhatavkar: Nice Catch! will update this

.github/workflows/automerge.yaml
omkarkhatavkar: Yes, only PRT uses the old Status API. Checks are running on GitHub are uses the Checks API. No need to add further filters as I already added one layer check earlier. Anyways this wont be required.

.github/workflows/automerge.yaml
omkarkhatavkar: Yeah this good Point we can move this block above

.github/workflows/automerge.yaml
omkarkhatavkar: Although the step is not necessary, as the lewagon/wait-on-check-action@v1.2.01 is responsible for failing the step, the PRT also exits when it fails. There is no need to collect the output result as the program exits at that point.

jyejare: But what if checks passed and PRT did not :) This step will still run.

jyejare (changes requested):

.github/workflows/automerge.yaml
omkarkhatavkar: `exit 1`  will terminate the program with failure. There will not steps will be run.

.github/workflows/automerge.yaml
```diff
@@ -43,9 +43,68 @@ jobs:
...
+              echo "PRT failed to start ! Stopping."
+              exit 1
+          fi
+
+          status=$(get_status)
+          context=$(get_context)
+
+          if [ "$context" != "Robottelo-Runner" ]; then
+            echo "::error Failed to get Robottelo-Runner status"
+            exit 1
```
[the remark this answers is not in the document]
omkarkhatavkar: No this will only check that the robottelo check is been started. It is checking only the context.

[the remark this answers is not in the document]
omkarkhatavkar: once the PRT starts it creates the status as pending. We are looking that the context generated is using robottelo-runner only

.github/workflows/automerge.yaml
omkarkhatavkar: @jyejare this is the latest test result https://github.com/omkarkhatavkar/robottelo/actions/runs/3929759805/jobs/6718953226

PR #10533, opened 2023-01-16
remove cv version from environment with host registered

vijaysawant:
- Remove promoted content view version from environment that is used in association of an Activation key and content-host registration
- Steps
1. Create a content view cv1
2. Add a yum repo to the content view
3. Publish the content view
4. Promote the content view to multiple environment Library -> DEV -> QE
5. Create an Activation key with the QE environment
6. Register a content-host using the Activation key
7. Remove the content view cv1 version from QE environment.  Note: prior to removing, replace the current QE environment of cv1 by DEV and content view cv1 for Content-host and for Activation key.
8. Refresh content-host subscription

### DIFF
--- a/tests/foreman/api/test_contentviewversion.py
+++ b/tests/foreman/api/test_contentviewversion.py
@@ -711,9 +711,12 @@ def test_positive_delete_cv_promoted_to_multi_env(module_org):
         content_view.read()
 
 
-@pytest.mark.stubbed
+@pytest.mark.no_containers
+@pytest.mark.rhel_ver_list([7, 8, 9])
 @pytest.mark.tier3
-def test_positive_remove_cv_version_from_env_with_host_registered():
+def test_positive_remove_cv_version_from_env_with_host_registered(
+    target_sat, default_org, rhel7_contenthost
+):
     """Remove promoted content view version from environment that is used
     in association of an Activation key and content-host registration.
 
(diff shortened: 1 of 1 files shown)


vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/api/test_contentviewversion.py -k test_positive_remove_cv_version_from_env_with_host_registered

tests/foreman/api/test_contentviewversion.py
```diff
@@ -711,9 +711,12 @@ def test_positive_delete_cv_promoted_to_multi_env(module_org):
         content_view.read()
 
 
-@pytest.mark.stubbed
+@pytest.mark.no_containers
+@pytest.mark.rhel_ver_list([7, 8, 9])
 @pytest.mark.tier3
-def test_positive_remove_cv_version_from_env_with_host_registered():
+def test_positive_remove_cv_version_from_env_with_host_registered(
+    target_sat, default_org, rhel7_contenthost
```
synkd, line 718: Most of the tests in this module use the `module_org` fixture. Is there a particular reason that you're using `default_org` in this test?

tests/foreman/api/test_contentviewversion.py
```diff
@@ -747,6 +750,63 @@ def test_positive_remove_cv_version_from_env_with_host_registered():
...
+    )
+    assert rhel7_contenthost.subscribed
+    result = rhel7_contenthost.execute('yum install -y tree')
+    assert result.status == 0
+
+    # assert - content-host exists
+    host = entities.Host().search(query={"search": f'name={rhel7_contenthost.hostname}'})[0]
+    assert host.name == rhel7_contenthost.hostname
+
+    # Remove the content view cv1 version from QE environment. Note: prior to removing,
```
synkd, line 794: The steps and expected results are not 100% clear to me, but I'm interpreting them a bit differently from your implementation here. My reading is that the test is supposed to check that removing the content view version from the QE LCE doesn't affect the host or the activation key. Here's my understanding of the test based on the steps/expected results:
- create the QE and DEV LCEs
- sync the repo, create/publish/promote the content view
- create the activation key
- register host using activation key
- change the LCE of the activation key and the host to Dev
- remove the content view from the QE LCE
- run `subscription-manager refresh` on the host
- assert that the activation key still exists
- assert that the content host still exists
- assert that the activation key and host are in the Dev LCE
- try to install a package from the repo in the content view on the host and assert that the installation was successful

As I mentioned, I don't think it's completely clear what the intention of the test is based on the steps and expected results, so please let me know what you think about this.

tests/foreman/api/test_contentviewversion.py
```diff
@@ -747,6 +750,63 @@ def test_positive_remove_cv_version_from_env_with_host_registered():
...
+    cv1.version[0].promote(data={'environment_ids': lce_qe.id, 'force': True})
+    # Create an Activation key with the QE environment
+    ak_lce = entities.ActivationKey(
+        organization=default_org, environment=lce_qe, content_view=cv1
+    ).create()
+    # assert created ak is associate with lce qe
+    assert ak_lce.environment.id == lce_qe.id
+
+    # Register a content-host using the Activation key
+    rhel7_contenthost.create_custom_repos(rhel7=settings.repos.rhel7_os)
```
synkd, line 780: If I understand create_custom_repo correctly, it generates a repofile for an external repo and places it on the content host. It looks like the repo you're using here is the RHEL 7 compose repo, which is where tree is installed from a few lines down. However, that repo is not in the content view that you've created. Based on my reading of the steps and expected results, this test should install a package from the content view onto the host. Please let me know if I'm misunderstanding this.

tests/foreman/api/test_contentviewversion.py
```diff
@@ -747,6 +750,63 @@ def test_positive_remove_cv_version_from_env_with_host_registered():
...
+        f'subscription-manager register --activationkey={ak_lce.name} \
+        --org="{default_org.label}" --force'
+    )
+    assert rhel7_contenthost.subscribed
+    result = rhel7_contenthost.execute('yum install -y tree')
+    assert result.status == 0
+
+    # assert - content-host exists
+    host = entities.Host().search(query={"search": f'name={rhel7_contenthost.hostname}'})[0]
+    assert host.name == rhel7_contenthost.hostname
```
ogajduse, line 792: Can it be substituted with the following?
```python
assert rhel7_contenthost.nailgun_host
```

tests/foreman/api/test_contentviewversion.py
```diff
@@ -747,6 +750,63 @@ def test_positive_remove_cv_version_from_env_with_host_registered():
...
+    cv1.version[0].promote(data={'environment_ids': lce_dev.id, 'force': True})
+    ak_updated = entities.ActivationKey(
+        id=ak_lce.id, organization=default_org, environment=lce_dev, content_view=cv1
+    ).update(['environment'])
+    rhel7_contenthost.unregister()
+    rhel7_contenthost.execute(
+        f'subscription-manager register --activationkey={ak_updated.name} \
+        --org="{default_org.label}" --force'
+    )
+    rhel7_contenthost.execute('subscription-manager refresh')
```
ogajduse, line 807: Why is `refresh` needed here if you are not performing any other actions on the content host after it?

vsedmik:
Well, yeah. Just skimmed through and all the comments are valid and need to be addressed. But first of all I would start with @synkd 's comment https://github.com/SatelliteQE/robottelo/pull/10533#discussion_r1071449975 as it makes more sense to me than the original docstring steps and it gives better guidance, so I'd stick to Danny's and try to refactor the test accordingly.
Also use of the fixtures that @ogajduse proposed will make your life easier and improve the code readability.
Try to debug with just one `rhel_contenthost` and when everything is ok, extend to `@pytest.mark.rhel_ver_match('[^6]')` as mentioned above.

PR #10540, opened 2023-01-17
Add external-logging option

rplevka:
adds a cli option (`--external_logging`) for enabling external logging (promtail->loki) on the provisioned VMs and Containers.
The plugin simply adds the appropriate extra arg (for vms) or env var (for containers) for all os versions existing in `supportability.yml`, as well as to the extra vars of the satellite server deploy options

### DIFF
--- a/pytest_plugins/external_logging.py
+++ b/pytest_plugins/external_logging.py
@@ -1,0 +1,26 @@
+from robottelo.config import settings
+
+
+def pytest_addoption(parser):
+    """Adds option for enabling external logging"""
+    help_text = '''
+        Flag for enabling promtail on the spawned hosts.
+        This is used for sending of the logs to the external Loki instance
+
+        Usage: --external_logging
+    '''
+    parser.addoption("--external-logging", action="store_true", default=False, help=help_text)
+
+
+def pytest_cmdline_main(config):
+    if not config.getoption('external_logging', False):
+        return
+    settings.set('server.deploy_arguments.promtail_enable', True)
+    settings.set('capsule.deploy_arguments.promtail_enable', True)
+    ch = settings.content_host
+    for os in [i for i in ch if isinstance(ch[i], dict) and ch[i].get('vm')]:
+        ch[os]['vm']['promtail_enable'] = True
+    # update the container env too, if available
+    promtail_var = {'PROMTAIL_ENABLE': "True"}
+    for os in [i for i in ch if isinstance(ch[i], dict) and ch[i].get('container')]:
+        settings.set(f'content_host.{os}.container.environment', promtail_var)
--- a/conftest.py
+++ b/conftest.py
@@ -4,6 +4,7 @@ import pytest
 pytest_plugins = [
     # Plugins
     'pytest_plugins.disable_rp_params',
+    'pytest_plugins.external_logging',
     'pytest_plugins.fixture_markers',
     'pytest_plugins.infra_dependent_markers',
     'pytest_plugins.issue_handlers',


rplevka:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py::test_positive_apply_security_erratum --external-logging

rplevka:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py::test_positive_apply_security_erratum --external-logging

rplevka:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py::test_positive_apply_security_erratum --external-logging

rplevka:
the failing test itself is not caused by the change. The goal of running the test was to ensure the spawned vms and containers will log to the external resource, which they do:

```
$ logcli query '{broker_origin="align_to_satellite:/opt/app-root/src/robottelo/pytest_fixtures/core/xdist.py", hostname="redacted", filename="/var/log/foreman/production.log"}' | grep "Common labels"

2023/01/24 10:37:42 Common labels: {awx_user_name="satqe_auto_droid", broker_origin="align_to_satellite:/opt/app-root/src/robottelo/pytest_fixtures/core/xdist.py", distro_name="RedHat", distro_version="8.7", filename="/var/log/foreman/production.log", hostname="redacted", jenkins_url="https://our.jenkins/job/robottelo-pr-testing/1727/", job="foreman_production", sat_role="satellite", source="app", vm_template="tpl-sat-jenkins-6.13.0-7.0-rhel-8.7"}
```

ogajduse (approved):

rplevka:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py::test_positive_apply_security_erratum --external-logging

rplevka:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py::test_positive_apply_security_erratum --external-logging

omkarkhatavkar:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py::test_positive_apply_security_erratum --external-logging

pytest_plugins/external_logging.py
```diff
@@ -0,0 +1,28 @@
...
+    settings.server.deploy_arguments.promtail_enable = True
+    settings.capsule.deploy_arguments.promtail_enable = True
+    ch = settings.content_host
+    for os in [i for i in ch if isinstance(ch[i], dict) and ch[i].get('vm')]:
+        ch[os]['vm']['promtail_enable'] = True
+    # update the container env too, if available
+    promtail_var = {'PROMTAIL_ENABLE': "True"}
+    for os in [i for i in ch if isinstance(ch[i], dict) and ch[i].get('container')]:
+        env_settings = ch[os]['container'].get('environment', {})
+        ch[os]['container']['environment'] = env_settings
```
omkarkhatavkar, line 27: In single line `ch[os]['container'].setdefault('environment', {})`

pondrejk (approved):

pytest_plugins/external_logging.py
rplevka: thanks! applied.

omkarkhatavkar (approved):

omkarkhatavkar:
@rplevka CI is complaining

pytest_plugins/external_logging.py
```diff
@@ -0,0 +1,27 @@
...
+        Usage: --external_logging
+    '''
+    parser.addoption("--external-logging", action="store_true", default=False, help=help_text)
+
+
+def pytest_collection_modifyitems(config):
+    if not config.getoption('external_logging', False):
+        return
+    settings.server.deploy_arguments.promtail_enable = True
+    settings.capsule.deploy_arguments.promtail_enable = True
```
Gauravtalreja1, line 19: Would suggest using settings.set instead of direct assigning value to override 
```suggestion
    settings.set('server.deploy_arguments.promtail_enable', True)
    settings.set('capsule.deploy_arguments.promtail_enable', True)
```

pytest_plugins/external_logging.py
```diff
@@ -0,0 +1,27 @@
...
+    help_text = '''
+        Flag for enabling promtail on the spawned hosts.
+        This is used for sending of the logs to the external Loki instance
+
+        Usage: --external_logging
+    '''
+    parser.addoption("--external-logging", action="store_true", default=False, help=help_text)
+
+
+def pytest_collection_modifyitems(config):
```
jyejare, line 15: I think the way of enabling logging through `pytest_collection_modify` is not the correct way because we are not really modifying the collection items. With that said, we could go with multiple ways of enabling logging:
1. Directly through broker config adding an option to enable external logging and then that seeds in every `Host object` that is being used in ContentHost, Satellite and Capsule classed as it is.
2. Add the option through `robottelo conf` and then enable external logging through ContentHost object and then seed that will be used in Satellite and Capsule classes as it is.
3. Add option through any conf file and write a pytest autouse session level fixture in conftest.py that just seeds what you are doing here in `pytest_collection_modifyitems`.

There might be more smarter ways that asking CI and individuals to pass `--external-logging` option to pytest and individuals might forget that always to pass that option. So having that setting in conf file(either broker or robottelo confs) could be a much better option.

jyejare (changes requested):

pytest_plugins/external_logging.py
rplevka: > I think the way of enabling logging through pytest_collection_modify is not the correct way because we are not really modifying the collection items.

I'd still rather stick to the hook, as this is a 1-shot function and we don't need to cache it or anything (perfectly fits the hook function description).
For the "incorrect hook" argument - this is easily fixable, as all we need to do is to use a different, more suitable hook, that provides `config` object: `pytest_cmdline_main`.

Not including this inside a config was a little by design, as I'd like to have this feature to be `opt-in`. I believe, the vast minority of users will require the tests to log to external resource and I'd like to prevent people from setting it once in conf and forget about it.

This would generate a lot of logs that will only eat up the storage and won't be used.
from CI perspective, there's nothing easier than appending the option to the pipelines we need this in (we'd need to update CI files even if we used configs anyway).

pytest_plugins/external_logging.py
rplevka: done, thanks @Gauravtalreja1

ogajduse (approved):
Additional ACK after the `pytest_cmdline_main` change that was discussed.

https://github.com/SatelliteQE/robottelo/pull/10540#discussion_r1091776948

jyejare:
@rplevka Lets talk about this in tomorrows Automation Weekly meeting !

jyejare:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py::test_positive_apply_security_erratum --external-logging

rplevka:
@jyejare the logs from the PRT got successfully logged to the ext loki server:
```
$ logcli -q series '{jenkins_url=~".*robottelo-pr-testing/1751.*"}' --since 24h | grep -o 'hostname=[^,]*' | sort | uniq
hostname="bdcff8d5f63e"
hostname="ip-10-0-216-176.<redacted>"
hostname="ip-10-0-217-28.<redacted>"
hostname="ip-10-0-218-196.<redacted>"
```

jyejare (approved):

PR #10556, opened 2023-01-22
Add merge functionality to config_helpers script

JacobCallahan:
This change add the ability for the script to merge in changes from another config directiory. This will most likely be used to merge in changes from the upstream config repository.

I also made some optimisations to the existing symlink functionality.

### DIFF
--- a/scripts/config_helpers.py
+++ b/scripts/config_helpers.py
@@ -2,7 +2,57 @@
 from pathlib import Path
 
 import click
+import deepdiff
+import yaml
 from logzero import logger
+
+
+def merge_nested_dictionaries(original, new, overwrite=False):
+    """Merge two dictionaries together, with the new dictionary taking precedence"""
+    # iterate through all keys in the new dictionary
+    for key, value in new.items():
+        # if the key is not in the original, add it
+        if key not in original:
+            original[key] = value
+        # if the key is in the original, and the value is a dictionary, recurse
+        elif isinstance(value, dict):
+            # use deepdiff to check if the dictionaries are the same
+            if deepdiff.DeepDiff(original[key], value):
+                original[key] = merge_nested_dictionaries(original[key], value, overwrite)
+        elif original[key] == value:
+            continue
+        # if the key is in the original, and the value is a list, ask the user
+        elif overwrite == "ask":
+            choice_prompt = (
+                f"The current value for {key} is {original[key]}.\n"
+                "Please choose an option:\n"
+                "1. Keep the current value\n"
+                f"2. Overwrite the current value with {value}\n"
+            )
+            if isinstance(value, list):
+                choice_prompt += "3. Merge the current value with {value}\n"
+            user_choice = click.prompt(choice_prompt, type=int, default=1, show_default=True)
+            if user_choice == 1:
+                continue
+            elif user_choice == 2:
+                original[key] = value
+            elif user_choice == 3 and isinstance(value, list):
+                original[key] = original[key] + value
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan (commented):
Tested merge functionality and it worked successfully. I'm not exactly sure if the symlink is working for me though.

Griffin-Sullivan:
When I look for the symlinks they aren't showing up.

Griffin-Sullivan (approved):
Nevermind I was misunderstanding the purpose of the symlink. Everything is working correctly. I deleted a file and reran the symlink and it worked.

JacobCallahan:
@Griffin-Sullivan thanks for checking!

scripts/config_helpers.py
```diff
@@ -2,16 +2,66 @@
...
+                original[key] = merge_nested_dictionaries(original[key], value, overwrite)
+        elif original[key] == value:
+            continue
+        # if the key is in the original, and the value is a list, ask the user
+        elif overwrite == "ask":
+            choice_prompt = (
+                f"The current value for {key} is {original[key]}.\n"
+                "Please choose an option:\n"
+                "1. Keep the current value\n"
+                f"2. Overwrite the current value with {value}\n"
```
jyejare, line 30:
```suggestion
                f"2. Overwrite the current value with new/upstream value {value}\n"
```

scripts/config_helpers.py
```diff
@@ -21,19 +71,56 @@ def config():
...
+@click.option(
+    "--from",
+    "from_",
+    type=click.Path(exists=True, file_okay=False),
+    help="The path to the existing config directory",
+)
+@click.option(
+    "--strategy",
+    type=click.Choice(["ask", "overwrite", "preserve"]),
+    help="The strategy to use when merging",
```
jyejare, line 93:
```suggestion
    help="The strategy to use when merging. Choose one from ask, overwrite, or preserve. If not provided then default is ask",
```

jyejare (approved):
Looks Great! 

Left some nitpick comments !

jyejare (commented):
@JacobCallahan  Also why dont we take this config helper to `Makefile` so that we let robottelo users that something like this is available and `make <tab><tab>` this would be shown in their terminal ?

jyejare:
@JacobCallahan  TBH I dont use symlinks between CI and Robottelo files because any changed I make to those file from robottelo appears in CI confs and whenever I want to open a new PR in CI (satelliteqe-jenkins) , the git status just pops these files and it make my life difficult to raise a PR. How do we handle this with symlinks here then ? Or should we altogether remove the symlinks and copy plain and merge when there is a change in CI/upstream ?

JacobCallahan:
@jyejare this change allows you to do either behavior. A user, who doesn't make changes to config, can symlink files they don't touch. Also, they can use this script to copy/merge changes from the ci repo to their local robottelo.

jyejare (approved):

jyejare:
trigger: test-robottelo

jyejare:
@JacobCallahan CI checks looks to be failing. Please look into it so we can merge !

scripts/config_helpers.py
JacobCallahan: I think new/upstream is implied

scripts/config_helpers.py
JacobCallahan: I can't check right now, but are the options not presented by click in the help text?

jyejare:
trigger: test-robottelo

PR #10564, opened 2023-01-25
Add coverage for provisioning test

shubhamsg199:
Adding test coverage for https://bugzilla.redhat.com/show_bug.cgi?id=2105441

- Adding content sync repos for rhel versions.
- Adding rhel9_bos and rhel9_aps to constants.

Test Results:

```
$ pytest -k test_rhel_pxe_provisioning tests/foreman/api/test_provisioning.py -v
======================================== test session starts ====================================================                                                                                                                                 
tests/foreman/api/test_provisioning.py::test_rhel_pxe_provisioning[rhel7-bios] PASSED

tests/foreman/api/test_provisioning.py::test_rhel_pxe_provisioning[rhel7-uefi] PASSED

tests/foreman/api/test_provisioning.py::test_rhel_pxe_provisioning[rhel8-bios] PASSED

tests/foreman/api/test_provisioning.py::test_rhel_pxe_provisioning[rhel8-uefi] PASSED
=================================== 4 passed 6533.60s (1:48:53) ==============================
```
`rhel9-bios` and `rhel9-uefi` tests are failing because of [BZ-2105441](https://bugzilla.redhat.com/show_bug.cgi?id=2105441)


pytest_fixtures/component/provision_pxe.py
```diff
@@ -30,58 +30,85 @@ def module_provisioning_rhel_content(
...
     This fixture sets up kickstart repositories for a specific RHEL version
     that is specified in `request.param`.
     """
     sat = module_provisioning_sat.sat
     rhel_ver = request.param['rhel_version']
-    repo_names = [f'rhel{rhel_ver}']
-    if int(rhel_ver) > 7:
+    repo_names = []
+    if int(rhel_ver) == 7:
+        repo_names.append(f'rhel{rhel_ver}')
```
Griffin-Sullivan, line 42: Would we ever want to test anything < RHEL 7? Might be overkill but we could
```suggestion
    if int(rhel_ver) <= 7:
        repo_names.append(f'rhel{rhel_ver}')
```

robottelo/constants/__init__.py
```diff
@@ -317,7 +319,7 @@ class Colored(Box):
 REPOS = {
     'rhel7': {
         'id': 'rhel-7-server-rpms',
-        'name': 'Red Hat Enterprise Linux 7 Server RPMs x86_64 7Server',
+        'name': 'Red Hat Enterprise Linux 7 Server RPMs x86_64 7.9',
```
Griffin-Sullivan, line 322: For this and the rhel 8 KS repos below, do we want to lock a certain x.y version and have to update it or just use latest? I see the rhel8_bos at the end of the `REPOS` dict has just the whole number `8` for the version. I'm assuming this would grab latest?

Griffin-Sullivan (approved):
Just a few questions. Also should we add the BZ to the docstring for this test?

robottelo/constants/__init__.py
synkd: @Griffin-Sullivan Kickstart repos are only available by specifying y-stream versions; there's no option to just use the latest version. However, with the non-KS repos, you're right that using `8` will grab the latest. This is also true for `7Server`, which should be functionally identical to `7.9`. Because of this, I was curious about this change. @shubhamsg199 This is a non-blocking question, but could you please explain your reasoning for switching from `7Server` to `7.9` here?

pytest_fixtures/component/provision_pxe.py
synkd: Extended Life Cycle Support for RHEL 6 lasts for another 18 months, and RHEL 6 is still a supported Satellite client version, so this is worth considering.

synkd (approved):
Looks good overall. Just a couple of non-blocking add-ons to Griffin's questions. And tagging on to his final question, I do think we should have the BZ in the docstring of the test, particularly since we can expect the test to fail for RHEL 9 until the BZ is resolved.

pytest_fixtures/component/provision_pxe.py
shubhamsg199: Updated.

robottelo/constants/__init__.py
shubhamsg199: Switched back to 7Server. I did this change during my local testing, forgot to change it back :)

pondrejk (approved):
Other reviewers noted missing bz tag in the test docstring, it still remains to be added, otherwise ack

PR #10576, opened 2023-01-30
Adding new Team token in all the plugins and helpers

jyejare:
- Adding new Team token in all the plugins and helpers
- As well as Removing Assignees from Test Cases and Requirements
- TBD


jyejare:
trigger: test-robottelo

pytest_plugins/metadata_markers.py
```diff
@@ -24,8 +24,8 @@ def pytest_addoption(parser):
         help='Comma separated list of component names to include in test collection',
     )
     parser.addoption(
-        '--assignee',
-        help='Comma separated list of assignees to include in test collection',
+        '--team',
+        help='Comma separated list of teams to include in test collection',
```
Gauravtalreja1, line 28: Can we not use the old Assignee field for assigning test modules to teams instead of creating a new field? Just wondering what the rationale is here, non-blocking

pondrejk (approved):
[disclaimer] just reviewed the Endeavour commit

pytest_plugins/metadata_markers.py
jyejare: We could have but their is no point in keeping `assignee` field that should point to person but we want to point it to `team`.

Anyways there is just a difference of `key` name here, the functionality  should remain same.

vsedmik (approved):
Reviewed the Phoenix commit.

jameerpathan111 (approved):
reviewed platform team commit

shweta83:
LGTM for Rocket team.

shweta83 (approved):

JacobCallahan (approved):

PR #10595, opened 2023-01-31
Fixing installer test failures

Griffin-Sullivan:
Saw some failures in the most recent installer component run that needed some quick fixes.

`test_installer_sat_pub_directory_accessibility` and `test_installer_cap_pub_directory_accessibility`

- Both of these were failing because sometimes the command returns `HTTP/2 200 ` (yes with a space) and other times you can get the `HTTP/1.1`. So just adding both to the assertion should be good. Also should be fine for cherry-picking back since we are using `or`

`test_positive_foreman_module`

- We really ought to rename this `test_positive_selinux_foreman_module`
- On RHEL 8, `semodule -l` no longer returns package version, so let's just limit this test to assert both the module and rpm are present on the satellite

Results:
```
configfile: pyproject.toml
collected 1 item

tests/foreman/installer/test_installer.py ..                              [100%]

================== 2 passed, 1 warning in 1277.36s (0:21:17) ===================

configfile: pyproject.toml
collected 1 item

tests/foreman/destructive/test_installer.py .                            [100%]

=================== 1 passed, 1 warning in 661.32s (0:11:01) ===================
```

### DIFF
--- a/tests/foreman/destructive/test_installer.py
+++ b/tests/foreman/destructive/test_installer.py
@@ -57,7 +57,7 @@ def test_installer_sat_pub_directory_accessibility(target_sat):
     https_curl_command = f'curl -i {target_sat.url}/pub/ -k'
     for command in [http_curl_command, https_curl_command]:
         accessibility_check = target_sat.execute(command)
-        assert 'HTTP/1.1 200 OK' in accessibility_check.stdout.split('\r\n')
+        assert 'HTTP/1.1 200 OK' or 'HTTP/2 200 ' in accessibility_check.stdout.split('\r\n')
     target_sat.get(
         local_path='custom-hiera-satellite.yaml',
         remote_path=f'{custom_hiera_location}',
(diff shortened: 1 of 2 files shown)


tests/foreman/destructive/test_installer.py
```diff
@@ -57,7 +57,7 @@ def test_installer_sat_pub_directory_accessibility(target_sat):
     https_curl_command = f'curl -i {target_sat.url}/pub/ -k'
     for command in [http_curl_command, https_curl_command]:
         accessibility_check = target_sat.execute(command)
-        assert 'HTTP/1.1 200 OK' in accessibility_check.stdout.split('\r\n')
+        assert 'HTTP/1.1 200 OK' or 'HTTP/2 200 ' in accessibility_check.stdout.split('\r\n')
```
Gauravtalreja1, line 60: can you pls elaborate on why the command returns `HTTP/2 200`? non-blocking

synkd (approved):
Looks good, and I agree with you about renaming `test_positive_foreman_module`. Are you planning to do that in a separate PR?

pondrejk (approved):

tests/foreman/destructive/test_installer.py
Griffin-Sullivan: My guess is that `http_curl_command` returns the `HTTP/1.1` and `https_curl_command` returns the `HTTP/2`. Or the other way around? Either way we need both because each one is using a different protocol

pondrejk (approved):

tests/foreman/destructive/test_installer.py
ekohl: Correct. Most tools don't implement HTTP/2 over plain text HTTP (AKA `h2c`). None of the browsers do, not sure about curl. `http://` will in pretty much all of the cases use HTTP/1.1 while `https://` can upgrade to HTTP/2.

I think the reason for the space is just a format string. In HTTP/1(.1) it includes a human readable status code but in HTTP 2 I think they leave it out for efficiency. If you use a format string like `'%s %s %s' % (protocol, code, text)`  you will have an empty space when `text` is empty. Though on this part I'm just guessing.

PR #10631, opened 2023-02-03
Fixing false assertion in Infrastructure test

Griffin-Sullivan:
I'm not sure what the original intent of the assertion here was, but as you can see we store `result` as a variable for installing a package and then assert `result.status == 0`. However, a few lines later we had an assertion with the same variable (with no changes) for `result.status == 1`. This was causing a failure for the test as obviously the status is 0. I also disabled SCA as that was causing a failure on 6.13. This has passed on both 6.12 and 6.13.

### DIFF
--- a/tests/foreman/destructive/test_contenthost.py
+++ b/tests/foreman/destructive/test_contenthost.py
@@ -45,6 +45,7 @@ def test_content_access_after_stopped_foreman(target_sat, rhel7_contenthost):
     :parametrized: yes
     """
     org = target_sat.api.Organization().create()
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan:
```
configfile: pyproject.toml
collected 1 item

tests/foreman/destructive/test_contenthost.py .                          [100%]

================== 1 passed, 1 warning in 1116.26s (0:18:36) ===================
```

adarshdubey-star (approved):

ColeHiggins2 (approved):

PR #10390, opened 2022-12-02
Parameterizing Azure Test cases

adarshdubey-star:
`Parameterizing Azure Test cases to function on both Puppet and Non-puppet based Satellite based on Customer usage scenarios.`

`Test results:`

**API:**
```
pytest  tests/foreman/api/test_computeresource_azurerm.py
========================================================================================================= test session starts =========================================================================================================
platform linux -- Python 3.8.6, pytest-7.1.3, pluggy-1.0.0
shared_function enabled - OFF - scope:  - storage: file
rootdir: /home/addubey/work/robottelo, configfile: pyproject.toml
plugins: forked-1.4.0, services-2.2.1, cov-3.0.0, xdist-3.0.2, ibutsu-2.2.4, reportportal-5.1.2, mock-3.10.0
collected 21 items / 1 deselected / 20 selected                                                                                                                                                                                       

tests/foreman/api/test_computeresource_azurerm.py ....................  
```
**CLI:**

```
pytest  tests/foreman/cli/test_computeresource_azurerm.py
======================================================================================= test session starts =======================================================================================
platform linux -- Python 3.8.6, pytest-7.1.3, pluggy-1.0.0
shared_function enabled - OFF - scope:  - storage: file
rootdir: /home/addubey/work/robottelo, configfile: pyproject.toml
plugins: forked-1.4.0, services-2.2.1, cov-3.0.0, xdist-3.0.2, ibutsu-2.2.4, reportportal-5.1.2, mock-3.10.0
collected 22 items                                                                                                                                                                                

tests/foreman/cli/test_computeresource_azurerm.py ......................
```
**UI:**
```

pytest  tests/foreman/ui/test_computeresource_azurerm.py
============================================================================ test session starts ============================================================================
platform linux -- Python 3.8.6, pytest-7.1.3, pluggy-1.0.0
shared_function enabled - OFF - scope:  - storage: file
rootdir: /home/addubey/work/robottelo, configfile: pyproject.toml
plugins: forked-1.4.0, services-2.2.1, cov-3.0.0, xdist-3.0.2, ibutsu-2.2.4, reportportal-5.1.2, mock-3.10.0
collected 4 items                                                                                                                                                           

tests/foreman/ui/test_computeresource_azurerm.py F...    
``` 
`One of the test cases is failing with a known issue not related to change`

Signed-off-by: Adarsh Dubey <addubey@redhat.com>

PR #10650, opened 2023-02-08
hostgroup update test no longer puppet-specific

pondrejk:
The test tried to reach removed hammer subcommands, none of the linked bzs require puppet-enabled satellite. Adjusting for maintainability.

### DIFF
--- a/tests/foreman/cli/test_hostgroup.py
+++ b/tests/foreman/cli/test_hostgroup.py
@@ -229,7 +229,7 @@ def test_negative_create_with_content_source(module_org):
 
 @pytest.mark.run_in_one_thread
 @pytest.mark.tier2
-def test_positive_update_hostgroup(
+def test_positive_update_hostgroup_with_puppet(
     request,
     module_puppet_org,
     env,
@@ -256,7 +256,7 @@ def test_positive_update_hostgroup(
             {
                 'content-source-id': puppet_content_source['id'],
                 'organization-ids': module_puppet_org.id,
-                'environment-id': env.id,
+                'puppet-environment-id': env.id,
                 'query-organization-id': module_puppet_org.id,
             }
         )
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan (approved):

tests/foreman/cli/test_hostgroup.py
```diff
@@ -230,15 +230,11 @@ def test_negative_create_with_content_source(module_org):
 @pytest.mark.run_in_one_thread
 @pytest.mark.tier2
 def test_positive_update_hostgroup(
-    request,
```
Gauravtalreja1, line 233: I think the PR title is a little misleading, Puppet is still supported and there are many other Puppet options available related to Puppet which are available when using `session_puppet_enabled_sat` for hostgroup creation.

From related test failure, it's failing due to `Error: Unrecognised option '--environment-id'`, I'm not sure if an environment is required for host group creation @vsedmik can you confirm this behavior? if not can we try removing this option for create command?

Gauravtalreja1 (changes requested):

tests/foreman/cli/test_hostgroup.py
vsedmik: In general, the original puppet-related `--environment` was deprecated in favour of `--puppet-environment` quite some time ago. I wonder it survived for so long.
When looking into a puppet-enabled SAT, I can see:
```
[root@sat ~]# hammer hostgroup create --help
Usage:
    hammer hostgroup create [OPTIONS]

Options:
...
 --puppet-ca-proxy VALUE                   Name of Puppet CA proxy
 --puppet-ca-proxy-id NUMBER               Puppet CA Capsule id
 --puppet-class[es|-ids] LIST              Names/Ids of associated Puppet classes
 --puppet-environment[-id] VALUE/NUMBER    Name/id of associated Puppet environment
 --puppet-proxy VALUE                      Name of Puppet proxy
 --puppet-proxy-id NUMBER                  Puppet Capsule id
...
```
So `environment-id` -> `puppet-environment-id` might fix it if it fails on HG creation time.
The question is if @pondrejk wants to narrow down the case to bare HostGroup without puppet, or keep the current full scenario with puppet-enabled SAT.

pondrejk: Ok, whatabout a compromise, I'll keep the original puppet satellite-specific test and the new one as well

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_hostgroup.py -k update_hostgroup

pondrejk:
@Gauravtalreja1 @vsedmik prt passed, couls you please re-review?

vsedmik (approved):
Looks good to me!

tests/foreman/cli/test_hostgroup.py
Gauravtalreja1: @pondrejk I think instead of managing 2 similar tests for puppet and non-puppet, it would be nice to have parametrization for this and other tests in this module, similar @adarshdubey-star has implemented in https://github.com/SatelliteQE/robottelo/pull/10390 for Azure tests, non-blocking atm.

Gauravtalreja1 (approved):

PR #10661, opened 2023-02-09
vault-login: check for error code instead of error message

rmynar:
Check for error code instead of error message. The error message is different across locales (it's different in czech, so i didn't get the help_text), error code is still the same.

### DIFF
--- a/scripts/vault_login.py
+++ b/scripts/vault_login.py
@@ -22,7 +22,7 @@ def _vault_command(command: str):
     vcommand = subprocess.run(command, capture_output=True, shell=True)
     if vcommand.returncode != 0:
         verror = str(vcommand.stderr)
-        if 'vault: command not found' in verror:
+        if vcommand.returncode == 127:
             print(f"{Colored.REDDARK}Error! {HELP_TEXT}")
             sys.exit(1)
         elif 'Error revoking token' in verror:


scripts/vault_login.py
```diff
@@ -22,7 +22,7 @@ def _vault_command(command: str):
     vcommand = subprocess.run(command, capture_output=True, shell=True)
     if vcommand.returncode != 0:
         verror = str(vcommand.stderr)
-        if 'vault: command not found' in verror:
+        if vcommand.returncode == 127:
```
jyejare, line 25: Nice Addition ! 

BTW the error messages are being checked to print the right message. But if we have returncodes for every different failure then we should do that for all the errors here !

jyejare (changes requested):

scripts/vault_login.py
rmynar: There is a difference that the "command not found" comes from shell which is usually translated, but other error messages returned by _vault_ which is english only. I also noticed that _vault_ returns code 2 for both implemented catches, so there's no other way to differentiate.
[token revoke](https://github.com/hashicorp/vault/blob/d341378c6353a9e0f776ca024215ea11be8672df/command/token_revoke.go#L170)
[token lookup](https://github.com/hashicorp/vault/blob/d341378c6353a9e0f776ca024215ea11be8672df/command/token_lookup.go#L125)

ColeHiggins2 (approved):
Ack pending @jyejare approval of returncode

scripts/vault_login.py
jyejare: Cool! Lets only have return code check for `command not found`, rest we will check for `error string`. Agree @rmynar ?

rmynar: Agree. Thanks for review.

ogajduse (approved):

PR #10676, opened 2023-02-10
Fix upgrade collection for double pytest plugins reference

jyejare:
Pytest collection with latest pytest version is failing for having pytest plugins under non-root directory which we have for upgrade with `scenario_workers` in `tests/upgrades/` directory.

Error to be exact:
```
Defining 'pytest_plugins' in a non-top-level conftest is no longer supported:
It affects the entire test suite instead of just below the conftest as expected.
  /Users/jitendrayejare/GitRepos/robottelo/tests/upgrades/conftest.py
Please move it to a top level conftest file at the rootdir:
  /Users/jitendrayejare/GitRepos/robottelo
```

So this has been now moved under `pytest_plugins/upgrade/` folder now, so we wont see warnings error in exceptions any more.


Griffin-Sullivan (commented):
I don't think I know enough to approve here, but I have hit this when trying to collect upgrade tests before. Will this allow me to collect them now or is it just going to remove the warning?

jyejare:
@Griffin-Sullivan This will now remove the warning if you are collecting tests from root directory!

jyejare:
trigger: test-robottelo

jyejare:
@JacobCallahan Ping for review !

JacobCallahan (approved):

PR #10679, opened 2023-02-11
Factory instance fixtures file for multipurpose

jyejare:
- New pytest_fixtures file 'factory_fixtures' containing all the fixtures that spins new fresh satellite or capsule instances for on demand from test case
- New pytest plugin `factory_collection` that
  - Adds `factory_instance` marker on the tests using these factory fixtures so that we can filter them out easily
  - Temporary (Until we have upgrade snap template in place) deselect tests using these factory fixtures for upgrade testing in all these cases:
    -  if `--include-factory-upgrades` is not given to the pytest command
    - if test has 'upgrade' marker
    - if test has `destructive` marker
    - Pytest command mentions collecting only upgrade tests with `-m upgrade` specifier


pytest_plugins/factory_collection.py
```diff
@@ -0,0 +1,41 @@
...
+    deselected = []
+    selected = []
+    include_factory_upgrades = config.getvalue('include_factory_upgrades')
+    factory_fixture_names = [m[0] for m in getmembers(factory, isfunction)]
+    for item in items:
+        has_factoryfixture = set(item.fixturenames).intersection(set(factory_fixture_names))
+        if has_factoryfixture:
+            item.add_marker('factory_instance')
+        if 'upgrade' in config.option.markexpr and not config.getvalue('include_factory_upgrades'):
+            if all([item.get_closest_marker('upgrade'), has_factoryfixture]) or item.get_closest_marker('destructive'):
```
Griffin-Sullivan, line 26: Can we combine these together? Or was this done to make it easier to read?

Griffin-Sullivan (approved):
I'm still learning how the files under `pytest_plugins` work, but this makes sense and I don't see any errors in the logic.

pytest_fixtures/core/factory.py
```diff
@@ -0,0 +1,152 @@
...
+            host_class=Capsule,
+            workflow=settings.capsule.deploy_workflow,
+            _count=2,
+            **broker_args,
+        )
+        cap_hosts = wait_for(hosts.checkout, timeout=timeout, delay=delay)
+
+    yield cap_hosts.out
+
+    _ = [cap.teardown() for cap in cap_hosts.out]
```
JacobCallahan, line 113:
```suggestion
    [cap.teardown() for cap in cap_hosts.out]
```

pytest_fixtures/core/factory.py
```diff
@@ -0,0 +1,152 @@
...
+    if 'IDM' in request.param:
+        enroll_idm_and_configure_external_auth(new_sat)
+        yield new_sat
+        disenroll_idm(new_sat)
+    else:
+        enroll_ad_and_configure_external_auth(request, ad_data, new_sat)
+        yield new_sat
+    new_sat.unregister()
+    new_sat.teardown()
+    Broker(hosts=[new_sat]).checkin()
```
JacobCallahan, line 152:
```suggestion
    Broker(hosts=[new_sat]).checkin()

```

pytest_plugins/factory_collection.py
```diff
@@ -0,0 +1,41 @@
...
+def pytest_configure(config):
+    """Register markers related to testimony tokens"""
+    marker = 'factory_instance: Test uses a fresh satellite or Capsule instance deployed by broker'
+    config.addinivalue_line("markers", marker)
+
+
+def pytest_collection_modifyitems(session, items, config):
+    from pytest_fixtures.core import factory
+
+    deselected = []
```
JacobCallahan, line 17: why track what's been deselected?

JacobCallahan (changes requested):

JacobCallahan:
If you're consolidating host-requesting fixtures into one file, I think we need a name more descriptive than `factory`. Maybe `host_factories`?

pytest_plugins/factory_collection.py
jyejare: Earlier I wanted to have two conditions - in the first condition one with an upgrade mark and another condition with a destructive mark. But I need not to have it and they combiningly become the one condition on line 26.  I didnt rethink of making it one by combining it with line 25.

But I would like to keep that as is, as both of these conditions has is its own individuality .

pytest_plugins/factory_collection.py
jyejare: So that it shows in the pytest collection as to how many are deselected.  This is the standard we are following in all our collection plugins, so here as well !

jyejare:
> If you're consolidating host-requesting fixtures into one file, I think we need a name more descriptive than `factory`. Maybe `host_factories`?

@JacobCallahan I didnt like host_factory because people should not feel like they should also have/add content_host / host factories as well. This is very limited to satellite and capsule, and I didnt get the right name for limiting that to only sat and cap and not to content_hosts.

I am very bad at naming !

jyejare:
trigger: test-robottelo

JacobCallahan:
> @JacobCallahan I didnt like host_factory because people should not feel like they should also have/add content_host / host factories as well. This is very limited to satellite and capsule, and I didnt get the right name for limiting that to only sat and cap and not to content_hosts.

naming is certainly tricky, but I think we need something more descriptive than `factory`. Perhaps `sat_factory` or `sat_cap_factory`? I'm also not great at naming, so maybe we need someone else's input.

jyejare:
Naming shouldnt be a blocker here anyways! For now I will keep name as `sat_cap_factory` and lets merge it ASAP because I want to see upgrade real runs for all tiers ASAP. Today would be good day as we are getting snap.

JacobCallahan (approved):
ACK. File naming is important, so thanks for taking the time!

PR #10682, opened 2023-02-13
Add test_positive_mismatched_satellite_fqdn

jameerpathan111:
This pr adds automation coverage for [RFE#1978995](https://bugzilla.redhat.com/show_bug.cgi?id=1978995) (The satellite-installer should display the mismatched FQDN additionally rather than just showing the commands to verify the output).

Test result:
```
pytest tests/foreman/destructive/test_installer.py -k test_positive_mismatched_satellite_fqdn 
collected 3 items / 2 deselected / 1 selected

tests/foreman/destructive/test_installer.py .                            [100%]
================= 1 passed, 2 deselected, 7 warnings in 25.69s =================

```

### DIFF
--- a/tests/foreman/destructive/test_installer.py
+++ b/tests/foreman/destructive/test_installer.py
@@ -17,11 +17,24 @@
 :Upstream: No
 """
 import pytest
+from fauxfactory import gen_domain
+from fauxfactory import gen_string
 
 from robottelo.config import settings
 from robottelo.utils.installer import InstallerCommand
 
 pytestmark = pytest.mark.destructive
+
+
+@pytest.fixture
+def set_random_fqdn(target_sat):
+    shortname = gen_string('alpha')
+    new_domain = gen_domain()
+    target_sat.execute(
+        f'echo "search {new_domain}" >> /etc/resolv.conf; hostnamectl set-hostname {shortname}'
+    )
+    yield shortname, new_domain
+    target_sat.execute(f'hostnamectl set-hostname {target_sat.hostname}')
 
 
 def test_installer_sat_pub_directory_accessibility(target_sat):
@@ -114,3 +127,31 @@ def test_installer_inventory_plugin_update(target_sat):
         )
     )
     assert '(current: true)' in verify_proxy_plugin_flag.stdout
+
+
+def test_positive_mismatched_satellite_fqdn(target_sat, set_random_fqdn):
+    """The satellite-installer should display the mismatched FQDN
+
+    :id: 264910ca-23c8-4192-a993-24bc04994a4c
+
+    :steps:
+        1. Set incorrect satellite hostname.
+        2. assert that output of 'facter fqdn' and 'hostname --fqdn' is different.
+        3. Run satellite-installer.
+        4. Assert that satellite-installer command displays mismatched FQDN.
+
+    :expectedresults: The satellite-installer should display the mismatched FQDN.
+    """
+    shortname, new_domain = set_random_fqdn
+    facter_command_output = target_sat.execute('facter fqdn').stdout.strip()
+    hostname_command_output = target_sat.execute('hostname --fqdn').stdout.strip()
+    assert facter_command_output == f'{shortname}.{new_domain}'
+    assert hostname_command_output == shortname
+    assert facter_command_output != hostname_command_output
+    warning_message = (
+        f"Output of 'facter fqdn' "
+        f"({facter_command_output}) is different from 'hostname -f' "
+        f"({hostname_command_output})"
+    )
+    installer_command_output = target_sat.execute('satellite-installer').stderr
+    assert warning_message in str(installer_command_output)


jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_installer.py -k test_positive_mismatched_satellite_fqdn

jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_installer.py -k test_positive_mismatched_satellite_fqdn

jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_installer.py -k test_positive_mismatched_satellite_fqdn

tests/foreman/destructive/test_installer.py
```diff
@@ -114,3 +126,30 @@ def test_installer_inventory_plugin_update(target_sat):
...
+        2. assert that output of 'facter fqdn' and 'hostname --fqdn' is different.
+        3. Run satellite-installer.
+        4. Assert that satellite-installer command displays mismatched FQDN.
+
+    :expectedresults: The satellite-installer should display the mismatched FQDN.
+
+    :CaseAutomation: Automated
+    """
+    facter_command_output = target_sat.execute('facter fqdn').stdout.strip()
+    hostname_command_output = target_sat.execute('hostname --fqdn').stdout.strip()
```
ehelms, line 147: I think the trick here is, using a shortname rather than an FQDN to generate the difference in the two:

```
[root@sat-6-10-qa-rhel7 vagrant]# hostname test
[root@sat-6-10-qa-rhel7 vagrant]# hostname -f
test
[root@sat-6-10-qa-rhel7 vagrant]# facter fqdn
test.wareagle.example.com
```

ehelms: Oh sorry, I see the `gen_string` is generating a short hostname. Seems then facter is unable to resolve the domain and punting back to just reporting the host. The logic facter uses is linked below, and the tricky part is figuring out which of those code paths is not triggering such that `facter fqdn` should be resolving the domain and thus getting a different output than `hostname`.


@ekohl @evgeni 

https://github.com/puppetlabs/facter/blob/main/lib/facter/resolvers/linux/hostname.rb

ehelms: I am thinking that this may be related to what is in `/etc/resolv.conf`, my local one:

```
# cat /etc/resolv.conf 
# Generated by NetworkManager
search wareagle.example.com
nameserver 192.168.121.1
```

I think that `search` line is what is allowing facter to report back a full FQDN rather than just the short hostname.

jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_installer.py -k test_positive_mismatched_satellite_fqdn

tests/foreman/destructive/test_installer.py
```diff
@@ -17,13 +17,29 @@
...
 
 pytestmark = pytest.mark.destructive
 
 
+@pytest.fixture
+def set_wrong_fqdn(target_sat, request):
+    shortname = gen_string('alpha')
+    new_domain = gen_domain()
+    target_sat.execute(
+        f'echo "search {new_domain}" >> /etc/resolv.conf; hostnamectl set-hostname {shortname}'
```
ekohl, line 34: I had to check, but `man 5 resolv.conf` states:
> If there are multiple search directives, only the search list from the last instance is used.

So this is correct

tests/foreman/destructive/test_installer.py
```diff
@@ -17,13 +17,29 @@
...
+from fauxfactory import gen_string
 
 from robottelo.config import settings
 from robottelo.utils.installer import InstallerCommand
 
 pytestmark = pytest.mark.destructive
 
 
+@pytest.fixture
+def set_wrong_fqdn(target_sat, request):
```
ekohl, line 30: Technically speaking this fixtures sets a shortname as a hostname (as Debian does by default btw). So I don't think this sets it as "wrong", but rather something like `with_random_short_hostname`.

jameerpathan111: updated the name.

tests/foreman/destructive/test_installer.py
jameerpathan111: Updated the test, waiting for the automation result.

ekohl: Because I was interested in how it really worked I also had to look at manpages. `man 1 hostname` states for "THE FQDN" bit:
> Technically: The FQDN is the name getaddrinfo(3) returns for the host name returned by gethostname(2).  The DNS domain name is the part after the first dot.

And then making some selection, though the manpage doesn't specify how it does this. It does say:

> If a machine has multiple network interfaces/addresses or is used in a mobile environment, then it may either have multiple FQDNs/domain names or none at all. Therefore avoid using hostname --fqdn, hostname --domain and dnsdomainname.  hostname --ip-address is subject to the same limitations so it should be avoided as well.

The shortened C version is (removing all error handling and other cases):
```c
struct addrinfo *res;
struct addrinfo hints;
char *p;

memset(&hints, 0, sizeof(struct addrinfo));
hints.ai_socktype = SOCK_DGRAM;
hints.ai_flags = AI_CANONNAME;

p = localhost(); # a wrapper for gethostname() with a safe buffer and cache

getaddrinfo(p, NULL, &hints, &res);

printf("%s\n", res->ai_canonname);
```

See https://sources.debian.org/src/hostname/3.23%2Bnmu1/hostname.c/ for the full source.  At least Fedora 36 uses `hostname` from Debian.

Reading `man 3 getaddrinfo` it states:

> getaddrinfo() returns one or more addrinfo structures

And `addrinfo` contains an entry `ai_next` so I think it's a linked list. That suggests uses the first entry that's returned.

I think the Python equivalent is:
```python
import socket
addrs = socket.getaddrinfo(socket.gethostname(), None, type=socket.SOCK_DGRAM, flags=socket.AI_CANONNAME)
print(addrs[0][3])
```

tests/foreman/destructive/test_installer.py
```diff
@@ -17,13 +17,29 @@
...
+    shortname = gen_string('alpha')
+    new_domain = gen_domain()
+    target_sat.execute(
+        f'echo "search {new_domain}" >> /etc/resolv.conf; hostnamectl set-hostname {shortname}'
+    )
+    yield shortname, new_domain
+
+    @request.addfinalizer
+    def _finalize():
+        target_sat.execute(f'hostnamectl set-hostname {target_sat.hostname}')
```
Griffin-Sullivan, line 40: If this is for tear down, you should be able to just put:
```suggestion
    yield shortname, new_domain
    target_sat.execute(f'hostnamectl set-hostname {target_sat.hostname}')
```
Everything after the yield will run once the test is done.

ekohl (approved):
Not too familiar with Robottelo, but with my Python, pytest and installer knowledge this looks like a valid test.

Griffin-Sullivan (approved):

ehelms (approved):

tests/foreman/destructive/test_installer.py
jameerpathan111: oh right, I forgot. thanks for noticing it.

jameerpathan111: updated the pr.

shubhamsg199 (approved):
Ack

tests/foreman/destructive/test_installer.py
```diff
@@ -114,3 +127,33 @@ def test_installer_inventory_plugin_update(target_sat):
...
+
+    :steps:
+        1. Set incorrect satellite hostname.
+        2. assert that output of 'facter fqdn' and 'hostname --fqdn' is different.
+        3. Run satellite-installer.
+        4. Assert that satellite-installer command displays mismatched FQDN.
+
+    :expectedresults: The satellite-installer should display the mismatched FQDN.
+
+    :CaseAutomation: Automated
```
shubhamsg199, line 145: Already defined at module level.
```suggestion
```

jameerpathan111: removed.

ogajduse (approved):

PR #10694, opened 2023-02-14
fix in test_verify_bugzilla_1078866

pondrejk:
This test is around for a while and failing for a while for it was treating letters from the output as lines. This PR fixes the issue, though imo this test could even be considered for removal.

### DIFF
--- a/tests/foreman/cli/test_organization.py
+++ b/tests/foreman/cli/test_organization.py
@@ -69,7 +69,7 @@ def proxy(target_sat):
 
 
 @pytest.mark.tier2
-def test_verify_bugzilla_1078866():
+def test_positive_no_duplicate_lines():
     """hammer organization <info,list> --help types information
     doubled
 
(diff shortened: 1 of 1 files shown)


pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_organization.py -k test_verify_bugzilla_1078866

Gauravtalreja1 (approved):

Griffin-Sullivan (approved):
I don't think anyone would be upset if you removed it haha. However, if we do keep it, could we consider renaming it from `test_verify_bugzilla_1078866` to `test_positive_no_duplicate_lines` or something more descriptive.

pondrejk:
> I don't think anyone would be upset if you removed it haha. However, if we do keep it, could we consider renaming it from `test_verify_bugzilla_1078866` to `test_positive_no_duplicate_lines` or something more descriptive.

Good point, renamed

omkarkhatavkar (approved):

PR #10696, opened 2023-02-14
Satellite maintain upgrade check negative tests

Griffin-Sullivan:
Introducing a new negative test for checking system requirements on pre-upgrade check. Ideally this will pass on 6.14 when we get a build, but if we catch the failure I can come back and debug. I'm not sure what our current plan for stream is, but I assume we won't try to run this on stream or else it will (probably) fail.

### DIFF
--- a/tests/foreman/maintain/test_upgrade.py
+++ b/tests/foreman/maintain/test_upgrade.py
@@ -18,7 +18,19 @@
 """
 import pytest
 
+from robottelo.config import settings
+
 pytestmark = pytest.mark.destructive
+
+
+def last_y_stream_version(release):
+    """Returns the version of the last Y stream
+
+    Example: 6.13.0 -> 6.12
+    """
+    y = release.split('.')[1]
+    y_minus = str(int(y) - 1)
+    return f"{release.split('.')[0]}.{y_minus}"
 
 
 @pytest.mark.include_capsule
@@ -81,6 +93,60 @@ def test_positive_repositories_validate(sat_maintain):
     assert 'SKIPPED' in result.stdout
     assert 'FAIL' not in result.stdout
     assert skip_message in result.stdout
+
+
+@pytest.mark.no_containers
+@pytest.mark.parametrize(
+    'custom_host',
+    [
+        {'deploy_rhel_version': '8', 'deploy_flavor': 'satqe-ssd.disk.xxxl'},
+        {'deploy_rhel_version': '8', 'deploy_flavor': 'satqe-ssd.standard.std'},
+    ],
+    ids=['default', 'medium'],
+    indirect=True,
+)
+def test_negative_pre_upgrade_tuning_profile_check(request, custom_host):
+    """Negative test that verifies a satellite with less than
+    tuning profile hardware requirements fails on pre-upgrade check.
+
+    :id: 240bb97e-7353-4397-8d0e-228c3593c8cc
+
+    :steps:
+        1. Check out a RHEL instance with less than tuning requirements
+        2. Install Satellite
+        3. Run pre-upgrade check
+
+    :expectedresults: Pre-upgrade check fails.
+    """
+    profile = request.node.callspec.id
+    sat_version = ".".join(settings.server.version.release.split('.')[0:2])
+    # Register to CDN for RHEL8 repos, download and enable last y stream's ohsnap repos,
+    # and enable the satellite module and install it on the host
+    custom_host.register_to_cdn()
+    last_y_stream = last_y_stream_version(settings.server.version.release)
+    custom_host.download_repofile(product='satellite', release=last_y_stream)
+    custom_host.execute('dnf -y module enable satellite:el8 && dnf -y install satellite')
+    # Install without system checks to get around installer checks
+    custom_host.execute(
+        f'satellite-installer --scenario satellite --disable-system-checks --tuning {profile}',
+        timeout='30m',
+    )
+    # Get current Satellite version's repofile
+    custom_host.download_repofile(product='satellite', release=sat_version)
+    # Run satellite-maintain to have it self update to the newest version,
+    # however, this will not actually execute the command after updating
+    custom_host.execute('satellite-maintain upgrade list-versions')
+    # Check that we can upgrade to the new Y stream version
+    result = custom_host.execute('satellite-maintain upgrade list-versions')
+    assert sat_version in result.stdout
+    # Check that the upgrade check fails due to system requirements
+    result = custom_host.execute(
+        f'satellite-maintain upgrade check --target-version {sat_version}', timeout='5m'
+    )
+    assert (
+        f'ERROR: The installer is configured to use the {profile} tuning '
+        'profile and does not meet the requirements.' in result.stdout
+    )
 
 
 @pytest.mark.stubbed
--- a/pytest_fixtures/core/contenthosts.py
+++ b/pytest_fixtures/core/contenthosts.py
@@ -10,6 +10,7 @@ from broker import Broker
 from robottelo import constants
 from robottelo.config import settings
 from robottelo.hosts import ContentHost
+from robottelo.hosts import Satellite
 
 
 def host_conf(request):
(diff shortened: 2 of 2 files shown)


Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_upgrade.py -k test_negative_pre_upgrade_tuning_profile_check

Griffin-Sullivan:
Looks like the test passed on medium and failed on default. 
```
assert 'ERROR: The installer is configured to use the default tuning profile and does not meet the requirements.' in 'Checking for new version of satellite-maintain...\nNothing to update, can\'t find new version of satellite-maintain.\...Red Hat Technical Support.\n\nIn case the failures are false positives, use\n--whitelist="repositories-validate"\n\n\n
```
In my local testing I never had https://github.com/SatelliteQE/robottelo/pull/10696/files#diff-7998553584c990a5fe082e3046c81e1427af57ee13eca5cb8844b90c8eae69d8R153 return only that part. It should return after the full upgrade check is finished. Any ideas for why that would happen in production? Maybe setting a timeout would help?

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_upgrade.py -k test_negative_pre_upgrade_tuning_profile_check

Griffin-Sullivan:
The reason this won't pass CI is because OpenStack uses flavors. Need to edit the test so that it will work there. Putting in draft.

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_upgrade.py -k test_negative_pre_upgrade_tuning_profile_check

bherrin3:
> The reason this won't pass CI is because OpenStack uses flavors. Need to edit the test so that it will work there. Putting in draft.

Needed to move over to an appropriate flavor that included enough base disk. Looks like one was available for testing.

If we need a more permanent solution here, just let me know @Griffin-Sullivan .

ogajduse (approved):
:tada: :+1:

ogajduse:
@Griffin-Sullivan Is cherry-picking to 6.13 only intended?

tests/foreman/maintain/test_upgrade.py
```diff
@@ -83,6 +95,70 @@ def test_positive_repositories_validate(sat_maintain):
...
+        {'deploy_rhel_version': '8', 'deploy_flavor': 'satqe-ssd.disk.xxxl'},
+        {'deploy_rhel_version': '8', 'deploy_flavor': 'satqe-ssd.standard.std'},
+    ],
+    ids=['default', 'medium'],
+    indirect=True,
+)
+def test_negative_pre_upgrade_tuning_profile_check(request, custom_host):
+    """Negative test that verifies a satellite with less than
+    tuning profile hardware requirements fails on pre-upgrade check.
+
```
rmynar, line 111: I would appreciate a list of hw requirements and requirements that are not met in each of the flavors. I think it could be useful for future debugging when requirements or flavor configurations are changed.

Griffin-Sullivan:
> @Griffin-Sullivan Is cherry-picking to 6.13 only intended?

@ogajduse Yes. This is related to a new feature for 6.13, so it should only be ran on 6.13 and beyond.

tests/foreman/maintain/test_upgrade.py
Griffin-Sullivan: I can add the specifications for each flavor to the docstring here, although we do have a list that can be shared if you would prefer that. I'm not exactly sure if we would ever change the flavors? Maybe @ogajduse or @bherrin3 could give more information on that?

tests/foreman/maintain/test_upgrade.py
```diff
@@ -83,6 +95,70 @@ def test_positive_repositories_validate(sat_maintain):
...
+        )
+    else:
+        custom_host.execute(
+            'satellite-installer --scenario satellite --tuning development', timeout='30m'
+        )
+        custom_host.execute(
+            f'sed -i "s/tuning: development/tuning: {profile}/g" '
+            '/etc/foreman-installer/scenarios.d/satellite.yaml'
+        )
+        custom_host.execute('satellite-maintain service restart')
```
ehelms, line 145: What led you to need a service restart here?

Griffin-Sullivan: That's just what I did when I manually tested updating satellite.yaml. I thought it was safe to restart services when manually changing config files. Do you think I can remove this?

ehelms: I think it should be safe to remove it and speed up your test invocation.

ehelms:
LGTM from reviewing the flow of the test itself

tests/foreman/maintain/test_upgrade.py
```diff
@@ -83,6 +95,69 @@ def test_positive_repositories_validate(sat_maintain):
...
+        custom_host.execute(
+            'satellite-installer --scenario satellite --tuning development', timeout='30m'
+        )
+        custom_host.execute(
+            f'sed -i "s/tuning: development/tuning: {profile}/g" '
+            '/etc/foreman-installer/scenarios.d/satellite.yaml'
+        )
+    # Get current Satellite version's repofile and update foreman maintain
+    custom_host.download_repofile(product='satellite', release=sat_version)
+    custom_host.execute('dnf upgrade -y satellite-maintain rubygem-foreman_maintain')
```
lpramuk, line 147: run `satellite-maintain upgrade list-versions` twice instead to trigger an update of `rubygem-foreman_maintain`

tests/foreman/maintain/test_upgrade.py
```diff
@@ -83,6 +95,69 @@ def test_positive_repositories_validate(sat_maintain):
...
+    # Register to CDN for RHEL8 repos, download and enable last y stream's ohsnap repos,
+    # and enable the satellite module and install it on the host
+    custom_host.register_to_cdn()
+    last_y_stream = last_y_stream_version(settings.server.version.release)
+    custom_host.download_repofile(product='satellite', release=last_y_stream)
+    custom_host.execute('dnf -y module enable satellite:el8 && dnf -y install satellite')
+    # If testing <= 6.13 we can run installer with --disable-system-checks,
+    # otherwise we need to set the --tuning to development and then change the tuning
+    # profile manually to get around the installer check
+    if float(sat_version) <= 6.13:
```
lpramuk, line 132: we branch so you don't have to put this if clause

to master put else clause (+ cherrypick to 6.13.z)
when cherrypicking to 6.12.z amend and use part from if clause

lpramuk (changes requested):
Please change some parts

tests/foreman/maintain/test_upgrade.py
Griffin-Sullivan: I felt like it was more reliable to just directly update the package, but if you think having it update itself is better then I can switch.

tests/foreman/maintain/test_upgrade.py
Griffin-Sullivan: You're right I should base this off the branch. However, the if part would be for current release (6.13) and then when we branch 6.14 I'll have to switch it to the else clause. Maybe we should merge with just the if statement part and then when this fails on 6.14 I can switch to use the else part.

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_upgrade.py -k test_negative_pre_upgrade_tuning_profile_check

tests/foreman/maintain/test_upgrade.py
ehelms: The benefit of Lukas's suggestion is you end up following what we tell users to do which helps double check that workflow (https://access.redhat.com/documentation/en-us/red_hat_satellite/6.12/html/upgrading_and_updating_red_hat_satellite/upgrading_satellite#upgrading_a_connected_satellite_server)

Griffin-Sullivan: @ehelms yep that was what I was assuming a user would do. There's really no downside to updating this way. My concern was that reviewers would find it odd to use that command to update.

lpramuk (approved):

PR #10740, opened 2023-02-21
Coverage for execution of custom role in ansible

adarshdubey-star:
Test results :
```
pytest tests/foreman/ui/test_ansible.py -k test_positive_ansible_custom_role
platform linux -- Python 3.11.1, pytest-7.2.1, pluggy-1.0.0
plugins: xdist-3.1.0, services-2.2.1, mock-3.10.0, reportportal-5.1.3, ibutsu-2.2.4, cov-3.0.0
collected 7 items / 6 deselected / 1 selected                                                                                                                               

tests/foreman/ui/test_ansible.py .                                                                                                                                    [100%]

========================================================== 1 passed, 6 deselected, 5 warnings in 754.32s (0:12:34) ==========================================================
```
Signed-off-by: Adarsh Dubey [addubey@redhat.com](mailto:addubey@redhat.com)

### DIFF
--- a/tests/foreman/ui/test_ansible.py
+++ b/tests/foreman/ui/test_ansible.py
@@ -17,9 +17,11 @@
 :Upstream: No
 """
 import pytest
+import yaml
 from fauxfactory import gen_string
 
 from robottelo import constants
+from robottelo.config import robottelo_tmp_dir
 from robottelo.config import settings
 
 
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_ansible.py
```diff
@@ -177,3 +180,84 @@ def test_positive_config_report_ansible(session, target_sat, module_org, rhel_co
...
+        },
+    )
+    target_sat.wait_for_tasks(
+        f'resource_type = JobInvocation and resource_id = {job["id"]}', poll_timeout=1000
+    )
+    result = target_sat.api.JobInvocation(id=job['id']).read()
+    assert result.succeeded == 1
+    with session:
+        session.organization.select(module_org.name)
+        session.location.select(constants.DEFAULT_LOC)
```
Griffin-Sullivan, line 259: The org should already be set to module_org when the test session starts. If you want to use module_location then you won't have to select default location either. See https://github.com/SatelliteQE/robottelo/blob/master/pytest_fixtures/core/ui.py#L23

tests/foreman/ui/test_ansible.py
```diff
@@ -177,3 +180,84 @@ def test_positive_config_report_ansible(session, target_sat, module_org, rhel_co
...
+        assert session.host.search(target_host.name)[0]['Name'] == rhel_contenthost.hostname
+        session.configreport.search(rhel_contenthost.hostname)
```
Griffin-Sullivan, line 261: In the docstring we have:
```
:expectedresults:
        1. Config report should be visible for a custom role run.
```
But we never assert for this. Could you add an assertion that searching for this config report returns what we expect it to return?

adarshdubey-star: The coverage is just related to viewing the report page that is happening with the last few lines where we are dealing with the config report. The deletion of the report at the very last only occurs is the report is generated and the read takes care of the BZ scenario where we get the desired error. But yeah I should make it `generated` instead of `visible`.

tests/foreman/ui/test_ansible.py
adarshdubey-star: I can try removing the org here, about the location I think we should have the default one here as with the core/ui I don't want the module_loc to be selected by default.

Griffin-Sullivan (approved):

tests/foreman/ui/test_ansible.py
```diff
@@ -177,3 +180,83 @@ def test_positive_config_report_ansible(session, target_sat, module_org, rhel_co
...
+        .id
+    )
+    job = target_sat.api.JobInvocation().run(
+        synchronous=False,
+        data={
+            'job_template_id': template_id,
+            'targeting_type': 'static_query',
+            'search_query': f'name = {rhel_contenthost.hostname}',
+        },
+    )
```
Gauravtalreja1, line 251: can we not use UI here to test this scenario completely?

tests/foreman/ui/test_ansible.py
```diff
@@ -177,3 +180,83 @@ def test_positive_config_report_ansible(session, target_sat, module_org, rhel_co
         session.configreport.search(rhel_contenthost.hostname)
         session.configreport.delete(rhel_contenthost.hostname)
         assert len(session.configreport.read()['table']) == 0
+
+
+@pytest.mark.no_containers
+@pytest.mark.rhel_ver_match('9')
```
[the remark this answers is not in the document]
adarshdubey-star: I think to just test it with the latest rhel would be enough!

tests/foreman/ui/test_ansible.py
```diff
@@ -177,3 +180,83 @@ def test_positive_config_report_ansible(session, target_sat, module_org, rhel_co
...
+        4. Assert that the role was assigned to the host successfully
+        5. Run the Ansible playbook associated with that role
+        6. Check if the report is created successfully
+
+    :expectedresults:
+        1. Config report should be generated for a custom role run.
+
+    :BZ: 2155392
+
+    :CaseAutomation: Automated
```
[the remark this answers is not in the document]
adarshdubey-star: I would like to keep this as we need to have a track of it.

tests/foreman/ui/test_ansible.py
```diff
@@ -177,3 +180,83 @@ def test_positive_config_report_ansible(session, target_sat, module_org, rhel_co
...
+    """
+    SELECTED_ROLE = 'custom_role'
+    _, test_file = mkstemp(prefix='task-', suffix='.yaml')
```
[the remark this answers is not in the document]
adarshdubey-star: Sure

tests/foreman/ui/test_ansible.py
adarshdubey-star: We have a new UI for the job_invocation page which is been worked on to add support in airgun till then using those endpoints will break the test hence using the API way. We are testing a similar scenario but with an imported role from the satellite implicitly which is breaking in CI because of the change.
https://github.com/SatelliteQE/robottelo/blob/6ec81260378a3bf2c5a11dc4e2f8065ba41f5221/tests/foreman/ui/test_ansible.py#L128

adarshdubey-star: If we are able to get that change in before release that would be good otherwise I will have to adapt the API way in the above scenario as well for better results.

tests/foreman/ui/test_ansible.py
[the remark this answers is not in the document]
Gauravtalreja1: we already have defined at module level, so it fetches same CaseAutomation, and we add explicitely if its ManualOnly or NotAutomated

Gauravtalreja1 (approved):

PR #10742, opened 2023-02-21
Account for stream as a satellite release version

JacobCallahan:
Since we are setting stream as the satellite release version, we need to fallback to the backup satellite version in config.

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -71,11 +71,9 @@ def get_sat_version():
     try:
         sat_version = Satellite().version
     except (AuthenticationError, ContentHostError, BoxKeyError):
-        if hasattr(settings.server.version, 'release'):
-            sat_version = str(settings.server.version.release)
-        elif hasattr(settings.robottelo, 'satellite_version'):
-            sat_version = settings.robottelo.satellite_version
-        else:
+        if str(sat_version := settings.server.version.get('release')) == 'stream':
+            sat_version = str(settings.robottelo.get('satellite_version'))
+        if not sat_version:
             sat_version = SATELLITE_VERSION
     return Version('9999' if 'nightly' in sat_version else sat_version)
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ deepdiff==6.2.3
 dynaconf[vault]==3.1.11
 fauxfactory==3.1.0
 jinja2==3.1.2
-manifester==0.0.11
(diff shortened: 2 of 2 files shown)


robottelo/hosts.py
```diff
@@ -71,11 +71,9 @@ def get_sat_version():
...
-        if hasattr(settings.server.version, 'release'):
-            sat_version = str(settings.server.version.release)
-        elif hasattr(settings.robottelo, 'satellite_version'):
-            sat_version = settings.robottelo.satellite_version
-        else:
+        if str(sat_version := settings.server.version.get('release')) == 'stream':
+            sat_version = str(settings.robottelo.get('satellite_version'))
+        if not sat_version:
             sat_version = SATELLITE_VERSION
     return Version('9999' if 'nightly' in sat_version else sat_version)
```
devendra104, line 78: `stream` string doesn't comply with the PEP-440 (https://github.com/pypa/packaging/blob/5b34465682d40ddd42dd6cba88e158e1674de7b9/src/packaging/version.py#L113)

jyejare (commented):
ACK pending questions !

robottelo/hosts.py
JacobCallahan: correct. the fix here is to use settings.robottelo.satellite_version if settings.server.version.release is set to stream.

robottelo/hosts.py
```diff
@@ -71,11 +71,9 @@ def get_sat_version():
     try:
         sat_version = Satellite().version
     except (AuthenticationError, ContentHostError, BoxKeyError):
-        if hasattr(settings.server.version, 'release'):
-            sat_version = str(settings.server.version.release)
-        elif hasattr(settings.robottelo, 'satellite_version'):
-            sat_version = settings.robottelo.satellite_version
-        else:
+        if str(sat_version := settings.server.version.get('release')) == 'stream':
+            sat_version = str(settings.robottelo.get('satellite_version'))
```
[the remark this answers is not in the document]
JacobCallahan: correct. we should always be setting settings.robottelo.satellite_version so it can be the fallback here

devendra104 (approved):
ACK

ogajduse:
This PR should have been cherry-picked to `z` branches too. Now it is causing cherry-pick process to fail.

JacobCallahan:
The manifester version could have come back. However, the other change didn't need to.

PR #10786, opened 2023-02-27
[6.11.z] Fix upgrade collection for double pytest plugins reference

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/10676

Pytest collection with latest pytest version is failing for having pytest plugins under non-root directory which we have for upgrade with `scenario_workers` in `tests/upgrades/` directory.

Error to be exact:
```
Defining 'pytest_plugins' in a non-top-level conftest is no longer supported:
It affects the entire test suite instead of just below the conftest as expected.
  /Users/jitendrayejare/GitRepos/robottelo/tests/upgrades/conftest.py
Please move it to a top level conftest file at the rootdir:
  /Users/jitendrayejare/GitRepos/robottelo
```

So this has been now moved under `pytest_plugins/upgrade/` folder now, so we wont see warnings error in exceptions any more.


Satellite-QE:
trigger: test-robottelo

JacobCallahan (approved):

PR #10662, opened 2023-02-09
Automation fix for GCE UI based on feature changes

adarshdubey-star:
`Changes needed for the New GCE CR page `
For better results run with https://github.com/SatelliteQE/airgun/pull/807

Signed-off-by: Adarsh Dubey <addubey@redhat.com>

PR #10792, opened 2023-02-28
Updated test to match new gce provisioning

LadislavVasina1:
These changes fix `test_positive_gce_cloudinit_provision_end_to_end`.
(also fixing `test_positive_gce_provision_end_to_end`)
Since 6.13 gce is no longer being provisioned with project_id, client_email, etc.
It uses the content of json_key instead.

I have also found some issues caused by (probably) missing refactor changes.

`isinstance` check in `api_factory.py`'s class `templateupdate` is not working.
It uses `_satellite` attribute which is not defined in the class thus failing with an error.
I have removed the check because it solves the issue and I think that we don't need to check the object type here because we usually don't check it elsewhere. 

i.e. we don't check role type here, etc.
[https://github.com/shweta83/robottelo/blob/edd5d39b16d8262ebe6b4dfd153443760d4a0e6d/robottelo/host_helpers/api_factory.py#L500](here)


LadislavVasina1:
trigger: test-robottelo
pytest: -k 'test_positive_gce'  tests/foreman/ui/test_host.py

LadislavVasina1:
trigger: test-robottelo
pytest: -k 'test_positive_gce' tests/foreman/ui/test_host.py

robottelo/host_helpers/api_factory.py
```diff
@@ -619,7 +619,7 @@ def disable_syncplan(self, sync_plan):
         assert sync_plan.enabled is False
 
 
-class templateupdate:
+class Templateupdate:
```
ogajduse, line 622:
```suggestion
class TemplateUpdate:
```

ogajduse (approved):

LadislavVasina1:
trigger: test-robottelo
pytest: -k 'test_positive_gce' tests/foreman/ui/test_host.py

tests/foreman/ui/test_host.py
```diff
@@ -1907,13 +1907,15 @@ def gce_resource_with_image(
...
+            json_key_content = f.read()
+
         session.computeresource.create(
             {
                 'name': cr_name,
                 'provider': FOREMAN_PROVIDERS['google'],
-                'provider_content.google_project_id': gce_cert['project_id'],
-                'provider_content.client_email': gce_cert['client_email'],
-                'provider_content.certificate_path': settings.gce.cert_path,
+                'provider_content.json_key': json_key_content,
```
Gauravtalreja1, line 1918: @adarshdubey-star has a PR https://github.com/SatelliteQE/robottelo/pull/10662 open for fixing this and moving this test to the GCE module

LadislavVasina1: Ok, but other changes I made can be merged or not? He will just do a little rebase then.

adarshdubey-star: You can revert the changes for this file @LadislavVasina1, the rest I think @shweta83 can answer as she is working on a similar fix.

LadislavVasina1: reverted

robottelo/host_helpers/api_factory.py
```diff
@@ -628,12 +628,6 @@ def __init__(self, temp):
         :param entities.ProvisioningTemplate temp: entities.ProvisioningTemplate's object
         """
         self.temp = temp
-        if not isinstance(self.temp, self._satellite.api.ProvisioningTemplate):
```
shweta83, line 631: We need to keep this check.

robottelo/host_helpers/api_factory.py
```diff
@@ -628,12 +628,6 @@ def __init__(self, temp):
         :param entities.ProvisioningTemplate temp: entities.ProvisioningTemplate's object
         """
         self.temp = temp
```
shweta83, line 630: if not isinstance(self.temp, entities.ProvisioningTemplate):
            raise TypeError(
                'The template should be of type entities.ProvisioningTemplate, {} given'.format(
                    type(temp)
                )
            )

shweta83 (changes requested):

robottelo/host_helpers/api_factory.py
LadislavVasina1: but `entities` is not defined or is it?

robottelo/host_helpers/api_factory.py
LadislavVasina1: could you provide some reason for me please?

robottelo/host_helpers/api_factory.py
shweta83: You can import it.

ogajduse: @shweta83 Could you please elaborate on why we need to check the object type? I am inquisitive to know why we need it here considering that we do not check types in other places.

ogajduse: As long as the patch works, I am fine.

robottelo/host_helpers/api_factory.py
```diff
@@ -628,7 +629,8 @@ def __init__(self, temp):
...
-        if not isinstance(self.temp, self._satellite.api.ProvisioningTemplate):
+
+        if not isinstance(self.temp, entities.ProvisioningTemplate):
```
Gauravtalreja1, line 633: is `self._satellite.api.ProvisioningTemplate` not working? why use entities?

shweta83: It's not working. So had to use entities here.

Gauravtalreja1: what's an error? It should work, try adding `self.__dict__.update(initiate_repo_helpers(self._satellite))` to `__init__` of APIFactory class, as similar to CLIFactory

shweta83: This does work. Thanks @Gauravtalreja1 . 
@LadislavVasina1 Can you please add the suggested changes by Gaurav.

Gauravtalreja1:
@shweta83 @LadislavVasina1 Can this PR be closed in favor of https://github.com/SatelliteQE/robottelo/pull/10796?
It seems there are many more changes that are being addressed there and can be cherrypicked collectively.

shweta83:
> @shweta83 @LadislavVasina1 Can this PR be closed in favor of #10796? It seems there are many more changes that are being addressed there and can be cherrypicked collectively.

Yes we can close this. I added these changes into the mentioned PR as discussed.

robottelo/host_helpers/api_factory.py
ogajduse: @Gauravtalreja1 Why make it unnecessarily complicated? What is the goal of your suggestion, Gaurav?
I do not find anything wrong with keeping entities imported here.

Gauravtalreja1: @ogajduse Yes, entities work too but just to keep it consistent with other helper classes as we using API property everywhere, and despite what I've discussed with @shweta83 regarding using context manager decorator instead of class for it herein  https://github.com/SatelliteQE/robottelo/pull/10796 where it can work without any issues

shweta83: @Gauravtalreja1 @ogajduse  After debugging and looking at the usage of this, I agree with @ogajduse  that we do not need the check for template type. I have removed the check and updated it to use context manager decorator instead of class.

LadislavVasina1: As I looked through #10796 I realise that my PR is no longer needed. I will close it then.

shweta83: Thanks @LadislavVasina1

Issue #10814, opened 2023-03-01
[Failed-AutoCherryPick] - Implement retry logic in ohsnap helper methods using wait-for library


Satellite-QE:
#### Auto-Cherry-Pick WorkFlow Failure:
- To Branch: 6.11.z
- [Failed Cherrypick Action](https://github.com/SatelliteQE/robottelo/actions/runs/4301919460)
- [Parent Pull Request](https://github.com/SatelliteQE/robottelo/pull/10795)

rplevka:
manual backports:
- [6.12.z](https://github.com/SatelliteQE/robottelo/pull/10817)
- [6.11.z](https://github.com/SatelliteQE/robottelo/pull/10818)

PR #10818, opened 2023-03-01
[6.11.z] Implement retry logic in ohsnap helper methods using wait-for library

rplevka:
implement retry logic in ohsnap helper methods using wait-for library


jyejare:
Fixes #10814 !

jyejare (approved):

PR #10828, opened 2023-03-01
Fix infoblox plugin install tests

Gauravtalreja1:


Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_infoblox.py

tests/foreman/destructive/test_infoblox.py
```diff
@@ -16,21 +16,15 @@
 """
 import pytest
 
-from robottelo.hosts import get_sat_rhel_version
```
adarshdubey-star, line 19: Is get_sat_rhel_version removed?

adarshdubey-star: Also if we have different name for rhel 7 packages we should have it because you are cherry picking it for 6.11 ?

Gauravtalreja1: No it's not, i think we shouldn't remove it, and for 6.11, I'm aware of the changes, I'll update that in cherrypick PR

Gauravtalreja1:
PRT failures are caused in the latest stream build, and it works fine with 6.13 and 6.12 checked locally.

ColeHiggins2 (approved):
ACK Assuming tests pass locally in 6.13 and 6.12

adarshdubey-star (approved):

PR #10836, opened 2023-03-01
[RFE] SAT-11254 Test Automation

damoore044:
test_positive_block_delete_key_in_use (api and cli/test_contentcredentials.py)

-API and CLI test automation
-BZ 2052904 : Blocker now in place when attempting to delete a content credential in use by a product or repository.

Both tests create a new gpg key, a new product with the key, a new repository associated with the product, and attempts to delete the key in use. The tests expect the gpg key to remain associated with the product and repo, and that the gpg, prod, and repo are unmodified.


omkarkhatavkar:

tests/foreman/ui/test_contentcredentials.py
```diff
@@ -90,6 +90,61 @@ def test_positive_end_to_end(session, module_org, gpg_content):
...
+        )[0]
+        # Create new product with the gpg key, and a single associated new repository
+        product = entities.Product(gpg_key=gpg_key, organization=module_org).create()
+        repo = entities.Repository(product=product, url=settings.repos.yum_1.url).create()
+        # Assert the gpg key is associated with the new product and repo
+        values = session.contentcredential.read(gpg_key.name)
+        assert values['details']['name'] == name
+        assert values['details']['content_type'] == CONTENT_CREDENTIALS_TYPES['gpg']
+        assert values['products']['table'][0]['Used as'] == CONTENT_CREDENTIALS_TYPES['gpg']
+        assert values['repositories']['table'][0]['Used as'] == CONTENT_CREDENTIALS_TYPES['gpg']
```
Griffin-Sullivan, line 132: I would move all of this above `with session:` and use the API to create all of this stuff since it's test setup. Essentially the only thing we have to test with the UI here is trying to delete. Even better you could look and see if there are fixtures that already do all of this for you.

tests/foreman/cli/test_contentcredentials.py
```diff
@@ -102,6 +102,46 @@ def test_positive_get_info_by_name(module_org):
...
+    name = gen_string('utf8')
+    product = make_product({'organization-id': module_org.id})
+    repo = make_repository({'product-id': product['id']})
+    gpg_key = make_content_credential(
+        {'key': VALID_GPG_KEY_FILE_PATH, 'name': name, 'organization-id': module_org.id}
+    )
+    # Associate repo with the product, gpg key with product and repo
+    Product.update(
+        {'gpg-key-id': gpg_key['id'], 'id': product['id'], 'organization-id': module_org.id}
+    )
```
Griffin-Sullivan, line 125: Anywhere that we use `Product`, `ContentCredential`, or `Repository` should be accessed through `target_sat`. 
```suggestion
    target_sat.cli.Product.update(
        {'gpg-key-id': gpg_key['id'], 'id': product['id'], 'organization-id': module_org.id}
    )
```

At the top of this test, you should pass in `target_sat` so that we have access to it's CLI class (which gives us all the hammer commands we need like `Product.update()`). If you're wondering why other tests don't do this, it's because they haven't been updated since we started adding the cli object to `target_sat`.

tests/foreman/cli/test_contentcredentials.py
```diff
@@ -102,6 +102,46 @@ def test_positive_get_info_by_name(module_org):
...
+def test_positive_block_delete_key_in_use(module_org):
+    """Create a product and single associated repository. Create a new gpg key and associate
+        it with the product and repository. Attempt to delete the gpg key in use
+
+    :id: 022555fd-e6f2-4c95-80d7-cae26993ca8f
+
+    :expectedresults: deletion of the gpg key is blocked by raising CLIReturnCodeError 70,
+        the key is not deleted, and remains associated with the product and repository
+
+    """
```
Griffin-Sullivan, line 115: I think @ColeHiggins2 said this on the last PR, but you can add the BZ here too.

tests/foreman/ui/test_contentcredentials.py
damoore044: Great thanks for the suggestion there, I'll work on getting this changed

tests/foreman/cli/test_contentcredentials.py
```diff
@@ -102,6 +102,46 @@ def test_positive_get_info_by_name(module_org):
...
+    # Associate repo with the product, gpg key with product and repo
+    Product.update(
+        {'gpg-key-id': gpg_key['id'], 'id': product['id'], 'organization-id': module_org.id}
+    )
+
+    # Attempt to delete gpg in use
+    with pytest.raises(CLIReturnCodeError) as e:
+        ContentCredential.delete({'name': name, 'organization-id': module_org.id})
+    assert 'Cannot delete' in str(e.value)
+
```
ColeHiggins2, line 131: again with the spacing, lets just squish it all down to look pretty

tests/foreman/ui/test_contentcredentials.py
```diff
@@ -90,6 +90,61 @@ def test_positive_end_to_end(session, module_org, gpg_content):
...
+    :expectedresults: deletion of the gpg key is blocked by raising TypeError,
+        the key is not deleted, and remains associated with the product and repository
+
+    :BZ: 2052904
+
+    :CaseLevel: Integration
+    """
+
+    name = gen_string('alpha')
+
```
ColeHiggins2, line 110: here too !

tests/foreman/ui/test_contentcredentials.py
```diff
@@ -90,6 +90,61 @@ def test_positive_end_to_end(session, module_org, gpg_content):
...
+        assert values['details']['name'] == name
+        assert values['details']['content_type'] == CONTENT_CREDENTIALS_TYPES['gpg']
+        assert values['products']['table'][0]['Used as'] == CONTENT_CREDENTIALS_TYPES['gpg']
+        assert values['repositories']['table'][0]['Used as'] == CONTENT_CREDENTIALS_TYPES['gpg']
+
+        # Attempt to delete gpg in use
+        with pytest.raises(TypeError) as e:
+            session.contentcredential.delete(gpg_key.name)
+        assert len(str(e.value)) > 0
+
```
ColeHiggins2, line 138: and here

tests/foreman/ui/test_contentcredentials.py
damoore044: Just pushed changes with api coverage

tests/foreman/api/test_contentcredentials.py
```diff
@@ -227,3 +229,46 @@ def test_positive_delete(module_org):
...
+
+    # Assert the same gpg key is associated with new product and repo
+    assert product.gpg_key.id == gpg_key.id
+    assert repo.gpg_key.id == gpg_key.id
+    assert product.gpg_key.id == repo.gpg_key.id
+
+    # Attempt to delete gpg in use
+    with pytest.raises(HTTPError) as e:
+        gpg_key.delete()
+    assert len(str(e.value)) > 0
```
Griffin-Sullivan, line 265: Do we get any sort of useful error message here? I would prefer to assert something like "Can't delete gpg key because xyz" over asserting that there was at least one error returned. It will be easier to debug failures in the future if there is ever a regression.

tests/foreman/api/test_contentcredentials.py
```diff
@@ -227,3 +229,46 @@ def test_positive_delete(module_org):
     gpg_key.delete()
     with pytest.raises(HTTPError):
         gpg_key.read()
+
+
+@pytest.mark.tier1
```
Griffin-Sullivan, line 234: I would maybe argue that these are `tier3` tests, but I'm not the best at labeling this marker. What do others think? https://github.com/SatelliteQE/robottelo/blob/master/pytest_plugins/markers.py#L11 for context on tiers

Griffin-Sullivan (commented):
If you could look into these and update the PR we can then run PRT to see if it's working in production.

tests/foreman/api/test_contentcredentials.py
damoore044: HTTPError message is `500 Server Error: Internal Server Error for url: https://ip-10-0-167-114.rhos-01.prod.psi.rdu2.redhat.com/katello/api/v2/content_credentials/204`

However nailgun shows 
`nailgun.client - WARNING - Received HTTP 500 response: {"displayMessage":"Cannot delete record because of dependent root_repositories","errors":["Cannot delete record because of dependent root_repositories"]}` -do you know how I could access these nailgun warnings?

tests/foreman/api/test_contentcredentials.py
damoore044: Changed to `tier3` for now, also would like to hear what others think about this one

tests/foreman/api/test_contentcredentials.py
Griffin-Sullivan: if you set line 264 to 
```
result = gpg_key.delete()
```
You might be able to access that error in something like `result.stderr`. I would suggest running in debug mode and looking through the result object for that error. Let me know if you need help setting that up. I can show you how to do it in VS Code if you've never done it before.

ColeHiggins2 (approved):
Looks good to me!

PR #10873, opened 2023-03-03
Upgrade `All Tier tests` deploy arguments for `upgrade template`

jyejare:
- Switching between `non-upgrade deploy arguments` and `upgrade deploy arguments` from satellite_factory
- Server yaml file keeps both `non-upgrade deploy arguments` and `upgrade deploy arguments`
- Deploy_workflow is now part of deploy arguments, but would be popped out before sending to the Broker object

Here is the example of server.yaml file :
```yaml
  DEPLOY_ARGUMENTS:
    # Update after dynaconf https://github.com/rochacbruno/dynaconf/issues/690
    deploy_snap_version: this.server.version.snap
    deploy_sat_version: this.server.version.release
    deploy_rhel_version: this.server.version.rhel_version
    deploy_workflow: deploy-sat-jenkins
    # deploy_template_name: ''
  UPGRADE_DEPLOY_ARGUMENTS:
    deploy_scenario: "satellite-upgrade"
    deploy_sat_version: this.server.version.release
    deploy_snap_version: this.server.version.snap
    deploy_rhel_version: this.server.version.rhel_version
    deploy_workflow: deploy-satellite-upgrade
```


lpramuk:
I dont understand why we need another meta layer of configuration files.
This is not needed at all.

jyejare:
After discussion with @lpramuk , @JacobCallahan and @bherrin3 on Slack Chat, Its understood that:
- There are no different parameters or extra env vars to spin up the upgraded template
- Even though we are spinning up the `zStream` or `yStream` template, the setting of `server.deploy_arguments.deploy_snap_version` to those values retrieves the right stream template
- The only concern we left with was a different deploy_workflow name for upgrade templates, which CI and Teams has to switch between for running the tests on upgraded templates on CI and locally. 
  - For now, we settled on the teams to switch that WF manually instead of what this PR is doing automatically
  - Also, soon @bherrin3 is unifying `deploy_sat_jenkins` WF to pick up the templates for upgrade/GA/Snap so the upgrde WF would no more be there
  - Still if we see any delays in above point 2 or if people keep forgetting about switching the WF in above point 1, then we will revisit this PR.

PR #10863, opened 2023-03-03
[Draft] Seamless component Metadata across tools

jyejare:
TBD

PR #10884, opened 2023-03-06
Convert2RHEL component change to Registration

shweta83:


tests/foreman/api/test_convert2rhel.py
```diff
@@ -6,7 +6,7 @@
 
 :CaseLevel: Integration
 
-:CaseComponent: Convert2rhel
+:CaseComponent: Registration
```
Gauravtalreja1, line 9: Registration seems irrelevant here, Can you add some description about this change?

lhellebr (changes requested):
This file has two tests, both of them test specifically convert2rhel. Since we have a component for that, I don't see any reason not to use it.

Are there plans to drop that component? If yes, may I suggest using a different component, like RemoteExecution or Ansible?

Registration really seems out of place - while registration happens as a part of the test, it's not what is being tested.

Gauravtalreja1:
@shweta83 can this be closed in favour of PR #10863 ?

shweta83:
> @shweta83 can this be closed in favour of PR #10863 ?

Yes we can close this one.

shweta83:
Closing

PR #10919, opened 2023-03-09
adding skip marker for 'scapcontent' and 'oscappolicy' tests due open BZ's

vijaysawant:
- Current tests are failing due to Open BZ's [#2167937](https://bugzilla.redhat.com/show_bug.cgi?id=2167937) & [#2133151](https://bugzilla.redhat.com/show_bug.cgi?id=2133151) , can not create new SCAP content.
- Adding a marker to avoid running these tests from execution.

### DIFF
--- a/tests/foreman/ui/test_oscapcontent.py
+++ b/tests/foreman/ui/test_oscapcontent.py
@@ -35,6 +35,8 @@ def oscap_content_path(module_target_sat):
     return local_file
 
 
+@pytest.mark.skip_if_open("BZ:2167937")
+@pytest.mark.skip_if_open("BZ:2133151")
 @pytest.mark.tier1
 @pytest.mark.upgrade
 def test_positive_end_to_end(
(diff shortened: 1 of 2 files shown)


shubhamsg199 (approved):
Ack

tests/foreman/ui/test_oscapcontent.py
```diff
@@ -35,6 +35,7 @@ def oscap_content_path(module_target_sat):
     return local_file
 
 
+@pytest.mark.skip_if_open("BZ:2167937, 2133151")
```
jyejare, line 38: This wont work! Method `skip_if_open` only accepts a single number, so repeat the marker for every number, as below:
```suggestion
@pytest.mark.skip_if_open("BZ:2167937")
@pytest.mark.skip_if_open("BZ:2133151")
```

jyejare: Repeat this for all below !

jyejare (changes requested):

jyejare (approved):

PR #10950, opened 2023-03-14
Adding coverage for BZ 2142514

Griffin-Sullivan:
Quick assertion to check ansible version 2.13 is installed and clone is working properly.

### DIFF
--- a/tests/foreman/destructive/test_clone.py
+++ b/tests/foreman/destructive/test_clone.py
@@ -44,7 +44,14 @@ def test_positive_clone_backup(target_sat, sat_ready_rhel, backup_type, skip_pul
         1. Satellite-clone ran successfully
 
     :parametrized: yes
+
+    :BZ: 2142514
+
+    :customerscenario: true
     """
+    ansible_ver = target_sat.execute('ansible --version')
+    assert 'core 2.13' in ansible_ver.stdout
+
     rhel_version = sat_ready_rhel._v_major
     sat_version = target_sat.version
 


Griffin-Sullivan:
trigger: test-robottelo

Griffin-Sullivan:
Failures are unrelated to this change.

tests/foreman/destructive/test_clone.py
```diff
@@ -44,7 +44,14 @@ def test_positive_clone_backup(target_sat, sat_ready_rhel, backup_type, skip_pul
         1. Satellite-clone ran successfully
 
     :parametrized: yes
+
+    :BZ: 2142514
+
+    :customerscenario: true
     """
+    ansible_ver = target_sat.execute('ansible --version')
```
adarshdubey-star, line 52: non-blocking 
```suggestion
    ansible_version = target_sat.execute('ansible --version')
```

adarshdubey-star (approved):

tests/foreman/destructive/test_clone.py
```diff
@@ -44,7 +44,14 @@ def test_positive_clone_backup(target_sat, sat_ready_rhel, backup_type, skip_pul
...
+    assert 'core 2.13' in ansible_ver.stdout
```
omkarkhatavkar, line 53: I think we need to have better assertion here, we can not have a hard-coded version. This will start failing once we have upgraded to the ansible version. I feel that we should not have such checkpoints or if you still want at least it should not version specific

omkarkhatavkar (changes requested):

tests/foreman/destructive/test_clone.py
Griffin-Sullivan: yeah I wasn't quite sure where to take it from here. Maybe we could just assert when we run satellite-clone that we don't get the error that there's conflicts with ansible-core? that way it doesn't matter what version it is and we can catch if the failure is caused by ansible.

Griffin-Sullivan: Actually I think we should just close this. The error is not the same as the one I was talking about here. I think the automation we have now should catch the bug just fine.

lpramuk:
We want to have more end-to-end scenarios. 
This test would be out of the case.

Not all BZs neccesarily require coverage

PR #10951, opened 2023-03-14
RFE Automation for new async endpoint on manifest refresh

ColeHiggins2:
SAT-15867
The change here is that the endpoint to retrieve manifests is now split into two: an async endpoint to start "exporting" the manifest, and a second endpoint to download the exported manifest. Previously this was a single, synchronous endpoint.

This tests checks that there are no errors when refreshing a manifest and asserts new logging message to Candlepin

### DIFF
--- a/tests/foreman/api/test_subscription.py
+++ b/tests/foreman/api/test_subscription.py
@@ -480,10 +480,10 @@ def test_positive_async_endpoint_for_manifest_refresh(
     :BZ: 2066323
     """
     target_sat.upload_manifest(module_org.id, session_entitlement_manifest.content)
-    sub = entities.Subscription(organization=module_org)
+    sub = target_sat.Subscription(organization=module_org)
     # set log level to 'debug' and restart services
     target_sat.cli.Admin.logging({'all': True, 'level-debug': True})
-    target_sat.execute('satellite-maintain service restart')
+    target_sat.cli.Service.restart()
     # refresh manifest and assert new log message to confirm async endpoint
     sub.refresh_manifest(data={'organization_id': module_org.id})
     results = target_sat.execute(
@@ -491,4 +491,4 @@ def test_positive_async_endpoint_for_manifest_refresh(
     )
     assert 'Sending GET request to upstream Candlepin' in str(results)
     # set log level back to default
-    target_sat.execute('satellite-installer --reset-foreman-logging-level')
+    target_sat.cli.Admin.logging({'all': True, 'level-production': True})
--- a/tests/foreman/api/test_subscription.py
+++ b/tests/foreman/api/test_subscription.py
@@ -480,7 +480,7 @@ def test_positive_async_endpoint_for_manifest_refresh(
     :BZ: 2066323
     """
     target_sat.upload_manifest(module_org.id, session_entitlement_manifest.content)
-    sub = entities.Subscription(organization=module_org)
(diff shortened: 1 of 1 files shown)


tests/foreman/api/test_subscription.py
```diff
@@ -456,3 +456,39 @@ def test_positive_os_restriction_on_repos():
...
+
+    :customerscenario: true
+
+    :BZ: 2066323
+    """
+    target_sat.upload_manifest(module_org.id, session_entitlement_manifest.content)
+    sub = entities.Subscription(organization=module_org)
+    # set log level to 'debug' and restart services
+    target_sat.cli.Admin.logging({'all': True, 'level-debug': True})
+    target_sat.execute('satellite-maintain service restart')
```
shubhamsg199, line 486: You can use cli here
```suggestion
    target_sat.cli.Service.restart()
```

pondrejk (approved):

shubhamsg199 (approved):
ACK, pending comment

tests/foreman/api/test_subscription.py
```diff
@@ -456,3 +456,39 @@ def test_positive_os_restriction_on_repos():
...
+
+    :expectedresults: Manifest refresh succeeds with no errors and production.log
+        has new debug message
+
+    :customerscenario: true
+
+    :BZ: 2066323
+    """
+    target_sat.upload_manifest(module_org.id, session_entitlement_manifest.content)
+    sub = entities.Subscription(organization=module_org)
```
Gauravtalreja1, line 483:
```suggestion
    sub = target_sat.api.Subscription(organization=module_org)
```

tests/foreman/api/test_subscription.py
```diff
@@ -456,3 +456,39 @@ def test_positive_os_restriction_on_repos():
...
+    target_sat.cli.Admin.logging({'all': True, 'level-debug': True})
+    target_sat.execute('satellite-maintain service restart')
+    # refresh manifest and assert new log message to confirm async endpoint
+    sub.refresh_manifest(data={'organization_id': module_org.id})
+    results = target_sat.execute(
+        'grep "Sending GET request to upstream Candlepin" /var/log/foreman/production.log'
+    )
+    assert 'Sending GET request to upstream Candlepin' in str(results)
+    # set log level back to default
+    target_sat.execute('satellite-installer --reset-foreman-logging-level')
```
Gauravtalreja1, line 494: Instead of running the installer to reset would suggest using the same CLI to revert which will be faster
```suggestion
   target_sat.cli.Admin.logging({'all': True, 'level-production': True})
```

tests/foreman/api/test_subscription.py
ColeHiggins2: I left this as is because Target_sat.api was returning an incorrect dictionary. any ideas ?

Gauravtalreja1: Not sure, it should work as is, what change do you see?

jyejare: @ColeHiggins2 What dictionary do you see @ColeHiggins2 different than `entities.<component>` ?

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/api/test_subscription.py -k test_positive_async_endpoint_for_manifest_refresh

Gauravtalreja1 (approved):

PR #10853, opened 2023-03-02
Mark FAM and satellite-clone tests for PIT

jameerpathan111:
This PR:
- Adds `test_positive_list_tasks` specifically to add coverage for [BZ#2170034](https://bugzilla.redhat.com/show_bug.cgi?id=2170034) in PIT
- Marks FAM tests for PIT

PR #10972, opened 2023-03-16
Adding coverage for BZ 2142514

Griffin-Sullivan:
After discussion with @evgeni I decided to reopen a PR for this customer case. This assertion should catch if the bug is present and save some time.


Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_clone.py -k test_positive_clone_backup

omkarkhatavkar (approved):

evgeni (approved):

tests/foreman/destructive/test_clone.py
```diff
@@ -105,6 +109,8 @@ def test_positive_clone_backup(target_sat, sat_ready_rhel, backup_type, skip_pul
         ).status
         == 0
     )
+    # Assert clone won't fail due to BZ
+    assert sat_ready_rhel.execute('satellite-clone --help').status == 0
```
jameerpathan111, line 113: @Griffin-Sullivan @evgeni I have already covered this scenario in https://github.com/SatelliteQE/robottelo/pull/10853/files#diff-0136ad9f3a21adf5716d5013efd837789362d2624fb5c83f55a226650e0c0470R131 Even though it mentions [BZ#2170034](https://bugzilla.redhat.com/show_bug.cgi?id=2170034) but it's actually covering scenario for [BZ#2142514](https://bugzilla.redhat.com/show_bug.cgi?id=2142514). 
Also, `satellite-clone --help` command didn't reproduce the issue but `satellite-clone --list-tasks` did. 

@Griffin-Sullivan can you update `test_positive_list_tasks` instead and add the customer bz there?

jameerpathan111 (changes requested):

tests/foreman/destructive/test_clone.py
evgeni: Huh, I would have expected (and pretty sure that's how I tested the change) that `--help` has the same issues as `--list-tasks` but down the road either one works as they are both a) quick and b) don't require any customer data to identify a problem.

Given the test is already present, I think we can close this one.

(as for BZ numbers: they are the same "issue", just on different pythons)

lpramuk (changes requested):
So instead of adding new test would you rather improve full cloning tests and add early assert with satellite-clone --help

lpramuk (approved):
I don't think adding satellite-clone --help as a separate test helps anything.
We won't gate the full cloning tests anyway - they will be still running even if your quick test fails. I only see we would be gotten new test that does nothing.

So I like that you've  added an early assert with satellite-clone --help instead
:+1:

lpramuk (changes requested):
lets just have `satellite-clone --list-tasks` instead of `satellite-clone --help`
and we are ready to ship

tests/foreman/destructive/test_clone.py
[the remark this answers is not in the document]
jameerpathan111: Makes sense, if it's gonna save time. 
The only reason I had added a separate test for this was that we didn't want to run e2e test related to satellite-clone for PIT, cause we have a time limitation there(I guess 2 hours). So checking if at least `satellite-clone --list-tasks`  command works made sense, something better than nothing.😄

jameerpathan111 (approved):

shubhamsg199 (approved):
Ack

lpramuk (approved):

PR #11008, opened 2023-03-21
convert `local_report_path` to string

mstrlc:
Convert `local_report_path` to a string so the test doesn't fail on `rhcloud_sat_host.get()` because of incompatible type.

PR #11009, opened 2023-03-21
convert `local_report_path` to string

mstrlc:
Convert `local_report_path` to a string so the test doesn't fail on `rhcloud_sat_host.get() `because of incompatible type.


jameerpathan111 (approved):

adarshdubey-star (approved):
Ack

adarshdubey-star:
@mstrlc any issues or changes in master/stream that are causing the PRT to fail there, https://github.com/SatelliteQE/robottelo/pull/11008

jameerpathan111:
> @mstrlc any issues or changes in master/stream that are causing the PRT to fail there, #11008

We haven't investigated why it's failing for the stream.  Once we know the root cause we'll update the other pr.
Merging this one.

PR #11019, opened 2023-03-22
Moving the ansible_roles_import test_case to destructive

adarshdubey-star:
Fellow tests are getting affected as all the roles are getting imported at once , making some assertion fail as custom role is created with in the same module for some tests , so I think it's better to move this to destructive.

### DIFF
--- a/tests/foreman/destructive/test_ansible.py
+++ b/tests/foreman/destructive/test_ansible.py
@@ -46,3 +46,31 @@ def test_positive_persistent_ansible_cfg_change(target_sat):
     assert param in target_sat.execute(f'cat {ansible_cfg}').stdout.splitlines()
     target_sat.execute('satellite-installer')
     assert param in target_sat.execute(f'cat {ansible_cfg}').stdout.splitlines()
+
+
+def test_positive_import_all_roles(target_sat):
+    """Import all Ansible roles available by default.
+
+    :id: 53fe3857-a08f-493d-93c7-3fed331ed391
+
+    :Steps:
+
+        1. Navigate to the Configure > Roles page.
+        2. Click the `Import from [hostname]` button.
+        3. Get total number of importable roles from pagination.
+        4. Fill the `Select All` checkbox.
+        5. Click the `Submit` button.
+        6. Verify that number of imported roles == number of importable roles from step 3.
+        7. Verify that Ansible variables have been imported along with roles.
+        8. Delete an imported role.
+        9. Verify that the role was successfully deleted.
+
+    :expectedresults: All roles are imported successfully. One role is deleted successfully.
+    """
+    with target_sat.ui_session() as session:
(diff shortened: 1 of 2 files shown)


tests/foreman/destructive/test_ansible.py
```diff
@@ -46,3 +46,31 @@ def test_positive_persistent_ansible_cfg_change(target_sat):
...
+        5. Click the `Submit` button.
+        6. Verify that number of imported roles == number of importable roles from step 3.
+        7. Verify that Ansible variables have been imported along with roles.
+        8. Delete an imported role.
+        9. Verify that the role was successfully deleted.
+
+    :expectedresults: All roles are imported successfully. One role is deleted successfully.
+    """
+    with target_sat.ui_session() as session:
+        assert session.ansibleroles.import_all_roles() == session.ansibleroles.imported_roles_count
```
Gauravtalreja1, line 71: WHY? could you add some description?

adarshdubey-star: Updated.

Gauravtalreja1: @adarshdubey-star can we close this PR as we discussed this last week?

adarshdubey-star: @Gauravtalreja1 I saw that case failing once with the last three runs I have executed the build once again if that's green I will close this one.

adarshdubey-star:
@Gauravtalreja1 @sambible I would like to get this one in as the failure is intermittent.

ogajduse:
@Gauravtalreja1 I see that you have some objections to this PR. What is your view on it?

ogajduse:
@adarshdubey-star Could you briefly mention what the different approaches were?

adarshdubey-star:
> @adarshdubey-star Could you briefly mention what the different approaches were?

Yeah, mainly I went through the fellow test cases to make sure the teardown happened properly for the imported roles,
I also tried to delete the roles directory from cli so that there will be no custom-made or implicit imported roles, but that too didn't help. It became intermittent but its very dicey.

Gauravtalreja1 (approved):

Griffin-Sullivan:
> I also tried to delete the roles directory from cli so that there will be no custom-made or implicit imported roles, but that too didn't help. It became intermittent but its very dicey.

That sounds weird. Could this be a possible bug?

tests/foreman/ui/test_ansible.py
```diff
@@ -25,34 +25,6 @@
 from robottelo.config import settings
 
 
-def test_positive_import_all_roles(target_sat):
```
jyejare, line 28: You could have used fixtures that creates a new satellite where you could have import the roles etc instead of going to the destructive satellite testing.

But ultimately even destructive tests are using new satellite so one and all its the same.

adarshdubey-star: yeah

PR #11084, opened 2023-03-28
skip rex UI tests on 2182353

pondrejk:
see https://github.com/SatelliteQE/airgun/pull/819#issuecomment-1486676435


lhellebr (approved):
ACK as per https://github.com/SatelliteQE/airgun/pull/819#issuecomment-1486856468

Gauravtalreja1 (approved):
Ack, with a question, shouldn't we cpick this to 6.13.z as this bz is opened for this version?

pondrejk:
@Gauravtalreja1 good question, since the 6.13 PR was still open, I edited it directly with this change so no need to cherry-pick here

PR #11090, opened 2023-03-28
Test for katello-tracer install through New Host UI

sambible:
First test using the New Hosts UI for the Tracer tab. Implementation is a bit hacky, not exactly familiar with how installing katello-tracer is done, so pieces from here and there were "appropriated" 

Requires https://github.com/SatelliteQE/airgun/pull/821

### DIFF
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -2621,7 +2621,7 @@ def test_positive_tracer_enable_reload(tracer_install_host, target_sat):
 
     :id: c9ebd4a8-6db3-4d0e-92a2-14951c26769b
 
-    :caseComponent: katello-tracer
+    :caseComponent: Katello-tracer
 
     :Team: Phoenix
 
@@ -2642,18 +2642,18 @@ def test_positive_tracer_enable_reload(tracer_install_host, target_sat):
     )
     with target_sat.ui_session() as session:
         session.organization.select(host['organization_name'])
-        title = session.host_new.get_tracer_title(tracer_install_host.hostname)
-        assert title == "Traces are not enabled"
+        tracer = session.host_new.get_tracer(tracer_install_host.hostname)
+        assert tracer['title'] == "Traces are not enabled"
         session.host_new.enable_tracer(tracer_install_host.hostname)
-        title = session.host_new.get_tracer_title(tracer_install_host.hostname)
-        assert title == "Traces are being enabled"
+        tracer = session.host_new.get_tracer(tracer_install_host.hostname)
+        assert tracer['title'] == "Traces are being enabled"
         wait_for(
-            lambda: session.host_new.get_tracer_title(tracer_install_host.hostname)
+            lambda: session.host_new.get_tracer(tracer_install_host.hostname)['title']
             != "Traces are being enabled",
             timeout=1800,
             delay=5,
             silent_failure=True,
             handle_exception=True,
         )
-        title = session.host_new.get_tracer_title(tracer_install_host.hostname)
-        assert title == "No applications to restart"
+        tracer = session.host_new.get_tracer(tracer_install_host.hostname)
+        assert tracer['title'] == "No applications to restart"
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -21,6 +21,7 @@ import csv
 import os
 import re
 from datetime import datetime
+from time import sleep
 
 import pytest
 import yaml
@@ -257,6 +258,21 @@ def remove_vm_on_delete(target_sat, setting_update):
         target_sat.api.Setting().search(query={'search': 'name=destroy_vm_on_host_delete'})[0].value
     )
     yield
+
+
+@pytest.fixture
+def prepared_host(rex_contenthost, target_sat):
+    """Install katello-host-tools-tracer, create custom
+    repositories on the host"""
+    # create a custom, rhel version-specific OS repo
+    rhelver = rex_contenthost.os_version.major
+    if rhelver > 7:
+        rex_contenthost.create_custom_repos(**settings.repos[f'rhel{rhelver}_os'])
+    else:
+        rex_contenthost.create_custom_repos(
+            **{f'rhel{rhelver}_os': settings.repos[f'rhel{rhelver}_os']}
+        )
+    yield rex_contenthost
 
 
 @pytest.mark.tier2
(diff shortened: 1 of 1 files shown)


sambible:
trigger: test-robottelo
pytest: tests/foreman/ui/test_host.py -v -k 'test_positive_tracer_enable_reload'
airgun: 821

sambible:
trigger: test-robottelo
pytest: tests/foreman/ui/test_host.py -v -k 'test_positive_tracer_enable_reload'
airgun: 821

tests/foreman/ui/test_host.py
```diff
@@ -2597,3 +2613,40 @@ def test_positive_remove_role_in_new_ui(self):
         up on the UI
 
     """
+
+
+@pytest.mark.tier2
+@pytest.mark.rhel_ver_match('8')
```
shubhamsg199, line 2619: Any specific reason we are testing for rhel8 and not rhel7 and rhel9 ?

lhellebr (changes requested):
Make sure the installation of Traces is done correctly, I suspect you forgot to put it in the fixture. Or the fixture's docstring just shouldn't say it enables it because it doesn't?

tests/foreman/ui/test_host.py
sambible: No, this was just an effort to limit the scope of the test to not take too long to run by being parametrized 5 or 6 more times. It shouldn't matter what version of Rhel the host is on, the test doesn't care, it just needs to be able to install katello-tracer

sambible:
This is a weird case, since I actually need it to not be installed on the host, just prepared for installation. This is because the workflow that I'm exercising actually uses rex to install the package, and then turn it on. Maybe I should make that more clear in the docstrings

tests/foreman/ui/test_host.py
```diff
@@ -2597,3 +2613,40 @@ def test_positive_remove_role_in_new_ui(self):
...
+    host = target_sat.api.Host().search(query={'search': prepared_host.hostname})[0].read_json()
+    with target_sat.ui_session() as session:
+        session.organization.select(host['organization_name'])
+        title = session.host_new.get_tracer_title(prepared_host.hostname)
+        assert title == "Traces are not enabled"
+        session.host_new.enable_tracer(prepared_host.hostname)
+        title = session.host_new.get_tracer_title(prepared_host.hostname)
+        assert title == "Traces are being enabled"
+        # Waiting on traces to be enabled. Not optimal, but not sure how else to do it
+        sleep(15)
```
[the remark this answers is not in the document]
sambible: Not really sure what I'd wait_for on here, since the way the test functions is a job gets triggered in the background, and sometime after that job completes the page reloads.

tests/foreman/ui/test_host.py
lhellebr: In that case, I think we usually parametrize the tests with all the possible versions.

tests/foreman/ui/test_host.py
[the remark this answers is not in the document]
lhellebr: How about waiting for the condition you are asserting and fail if after certain time, the waiting hasn't finished?

tests/foreman/ui/test_host.py
sambible: ack, if I remove this decorator it will run on all possible versions, as well as FIPS versions.

tests/foreman/ui/test_host.py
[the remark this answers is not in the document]
sambible: wait_for worked, thanks for the tip - hadn't used it before!

sambible:
trigger: test-robottelo
pytest: tests/foreman/ui/test_host.py -v -k 'test_positive_tracer_enable_reload'
airgun: 821

sambible:
Updated the docstring to reflect that this only sets up the host for katello-tracer installation, which is done during the test. Also updated it to run on all rhel7 and later hosts, as well as incorporated feedback around the wait, using wait_for now

vsedmik (approved):
Looks good to me

ogajduse (commented):
Great to see this addition! I feel this test is doing too less actions considering that it is spinning up a whole fleet of content hosts.
Keep in mind that spinning content hosts takes some time. I would rather see one robust UI test for the tracer rather than a bunch of tests testing just part of the component.
@sambible What are your future plans for the katello-tracer UI testing? Are you open to expanding this test?

lhellebr (approved):
Nice changes, LGTM

tests/foreman/ui/test_host.py
```diff
@@ -259,6 +259,21 @@ def remove_vm_on_delete(target_sat, setting_update):
     yield
 
 
+@pytest.fixture
+def prepared_host(rex_contenthost, target_sat):
```
shubhamsg199, line 263: Non-blocking, would suggest renaming this fixture as `prepared_host` seems something general.

shubhamsg199 (approved):
ACK

tests/foreman/ui/test_host.py
lhellebr: Yes.

sambible:
@ogajduse Absolutely open to modifying/extending this test. I agree that this is a rather simple test, but that's mostly because this is responding to some small UI work in an area that hasn't been tested before. I can definitely see an extension of this test in the future to a more general end to end test, but for now I'm just trying to get enough coverage to satisfy the requirements for a 6.13 release

tests/foreman/ui/test_host.py
```diff
@@ -2597,3 +2612,47 @@ def test_positive_remove_role_in_new_ui(self):
...
+        # Waiting on traces to be enabled. Not optimal, but not sure how else to do it
+        wait_for(
+            lambda: session.host_new.get_tracer_title(prepared_host.hostname)
+            != "Traces are being enabled",
+            timeout=1800,
+            delay=5,
+            silent_failure=True,
+            handle_exception=True,
+        )
+        title = session.host_new.get_tracer_title(prepared_host.hostname)
```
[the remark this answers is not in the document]
sambible: It does return the lambda value, unfortunately that value is a boolean, since I'm doing a comparison. No clean way I can think of to have this function properly and also return the pre comparison value.

JacobCallahan:
@sambible if you raise an issue for the future enhancements, we'll merge it 😉

sambible:
https://github.com/SatelliteQE/robottelo/issues/11112 - done :)

tests/foreman/ui/test_host.py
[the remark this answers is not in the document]
ogajduse: If you look closely at my suggestion, I have moved the comparison to `fail_condition`. But this is non-blocking and could be eventually addressed in #11112.

ogajduse (approved):

ogajduse:
I have put two bullet points to #11112 as recommendations for future improvement. I am merging the PR to unblock testing of the katello-tracer component.

sambible:
going to rebase/merge based on @ogajduse comments.

JacobCallahan:
@sambible we discourage self-merges. Can you also look into the failed cherry-pick?

ogajduse:
@sambible on @sambible's defense, I have suggested self-merge as there were a bunch of approvals and we want that coverage to be included ASAP.

ogajduse:
@JacobCallahan, on @sambible's defense, that was me who suggested self-merge as there were a bunch of approvals and we want that coverage to be included ASAP.

PR #11129, opened 2023-04-03
Cherrypick capsule-sync test

sambible:
Forgot to cherrypick this one to 6.13 last week.


shubhamsg199 (approved):
ACK, please add `:customerscenario: true` as BZ has a customer case attached

PR #11139, opened 2023-04-04
team tags updated for realm

pondrejk:
Well, I'm not 100% sure about this, the fact is that these tests started to appear in Endeavor results pile, which is correct if they truly belong under the Authentication component (CaseComponent clearly takes precedence over Team). Though there used to be a Realm component but I no longer see it in ohsnap, but still exists in BZ

### DIFF
--- a/tests/foreman/cli/test_realm.py
+++ b/tests/foreman/cli/test_realm.py
@@ -8,7 +8,7 @@
 
 :CaseComponent: Authentication
 
-:Team: Rocket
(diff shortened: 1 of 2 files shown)


jyejare (approved):

jyejare:
Yes @peterdragun The team token was wrong, they should be appeared in Endeavour!

Also, Realm was never a component in Ohsnap and hence we moved tests to Authentication based on the tests steps/expected_results.

PR #11158, opened 2023-04-05
Several fixes to test_contenthost and repo_mixins

sambible:
Few updates here:

1. Changing the setup_content fixture to use API instead of CLI. This is because the CLI method of this helper doesn't currently work, and using the API instead provides the same functionality. 
2. Create a fixture that sets the default Host UI to the old one. This is required by all tests in the module, hence the `autouse=True`

Both of these fixes should result in much better results for the Hosts-Content component.

### DIFF
--- a/tests/foreman/ui/test_contenthost.py
+++ b/tests/foreman/ui/test_contenthost.py
@@ -51,9 +51,19 @@ if not setting_is_set('clients') or not setting_is_set('fake_manifest'):
     pytest.skip('skipping tests due to missing settings', allow_module_level=True)
 
 
+@pytest.fixture(scope='module', autouse=True)
+def host_ui_default():
+    settings_object = entities.Setting().search(query={'search': 'name=host_details_ui'})[0]
+    settings_object.value = 'No'
+    settings_object.update({'value'})
+    yield
+    settings_object.value = 'Yes'
+    settings_object.update({'value'})
+
+
 @pytest.fixture(scope='module')
 def module_org():
-    org = entities.Organization().create()
+    org = entities.Organization(simple_content_access=False).create()
     # adding remote_execution_connect_by_ip=Yes at org level
     entities.Parameter(
         name='remote_execution_connect_by_ip',
--- a/robottelo/host_helpers/repository_mixins.py
+++ b/robottelo/host_helpers/repository_mixins.py
@@ -659,7 +659,7 @@ class RepositoryCollection:
         if self.need_subscription:
             # upload manifest only when needed
             if upload_manifest and not self.organization_has_manifest(org_id):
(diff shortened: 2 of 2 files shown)


sambible:
I also expect the setup_content change to fix other tests around robottelo ( I'm not sure how many places it's used ) - since it will always fail currently.

tests/foreman/ui/test_contenthost.py
```diff
@@ -51,9 +51,19 @@
     pytest.skip('skipping tests due to missing settings', allow_module_level=True)
 
 
+@pytest.fixture(scope='module', autouse=True)
```
jyejare, line 54: Why this is `autouse=True` and not on_demand, this would impact all the  tests in this module. Any reason why you would like to keep `host_details_ui` OFF ?

jyejare (commented):
ACK pending question !

Griffin-Sullivan:
Do we know when / if they are removing the old host UI? Only concern here is if 6.14 has only the new one then we will break host UI testing for stream and 6.14 when branching.

sambible:
These tests won't run without the old UI anyways. They will have to be not run on versions of Satellite without the old Host UI, or updated to use the new UI.

tests/foreman/ui/test_contenthost.py
sambible: host_details_ui needs to stay set to 'No' for all of the tests in this module, since they rely on the old host UI, and this is by far the easiest way to let them access it. I thought that autouse true was the easiest way to achieve this, rather than have each host set that property.

Griffin-Sullivan:
> These tests won't run without the old UI anyways

So the tests will skip or fail? I'm just making sure before we merge that this shouldn't be a <= 6.13 only PR. When are we dropping the old UI?

adarshdubey-star:
> Do we know when / if they are removing the old host UI? Only concern here is if 6.14 has only the new one then we will break host UI testing for stream and 6.14 when branching.

@Griffin-Sullivan @sambible  I can see the setting to switch to New UI in the stream for now at least, that implies it will be there for 6.14 but we should confirm this

tests/foreman/ui/test_contenthost.py
jyejare: Yeah, thats the very right way of doing it.

jyejare (approved):

lhellebr (approved):

sambible:
Talked to dev about the timeframe for removing the Old UI, and it won't be until at a minimum Satellite 6.15.

ColeHiggins2:
trigger: test-robottelo

ColeHiggins2 (approved):
ACK

JacobCallahan (approved):

PR #11193, opened 2023-04-07
Rework pagination test

sambible:
Needs a sister Airgun PR to update the pagination widgets. 

Rework/fix of this test - simple to understand stuff, but the change to the first assert is because the current search doesn't actually return all 3 of the results you'd expect, since it respects the per-page setting on the page. This should suffice to determine that we actually did create and find the correct content hosts though, so it should be fine.


tests/foreman/ui/test_contenthost.py
```diff
@@ -1755,20 +1754,18 @@ def test_pagination_multiple_hosts_multiple_pages(session, module_host_template)
...
-        all_fake_hosts_found = session.contenthost.search(
+        # Search for all the hosts by os.
+        fake_hosts_found = session.contenthost.search(
             f'os = {module_host_template.operatingsystem.name}'
         )
-        # Assert dump of fake hosts found includes the higest numbered host created for this test
-        match = re.search(fr'test-{host_num:0>2}', str(all_fake_hosts_found))
-        assert match, 'Highest numbered host not found.'
+        # Assert dump of fake hosts found includes the first numbered host created for this test
+        assert 'test-01' in fake_hosts_found[0]['Name']
```
ColeHiggins2, line 1762: will this always be 'test-01'? looks like we account for test-{host_num:0>2} previously

swadeley: Hi, what is the point of confirming the first host is found?  To prove that pagination is working you need to search for all matching hosts and then assert the host that is on the second page is found.

swadeley:
> Rework/fix of this test - simple to understand stuff, but the change to the first assert is because the current >search doesn't actually return all 3 of the results you'd expect, since it respects the per-page setting on the page. 

But why? This sounds like a regression. When you search for a host by name it should find the hosts beyond the first page. That is what this test is about.

tests/foreman/ui/test_contenthost.py
```diff
@@ -1721,7 +1720,7 @@ def test_pagination_multiple_hosts_multiple_pages(session, module_host_template)
 
     Search for hosts based on operating system and assert that more than one page
     is reported to exist and that more than one page can be accessed. Make some
-    additonal aserts to ensure the pagination widget is working as expected.
+    additonal asserts to ensure the pagination widget is working as expected.
```
adarshdubey-star, line 1723:
```suggestion
    additional asserts to ensure the pagination widget is working as expected.
```

tests/foreman/ui/test_contenthost.py
omkarkhatavkar: I would rather try to fix the issue in pagination or if you need some different behavior please try to override the airgun/views/contenthost.py file. Removing the pagination would not be the right solution here.

lhellebr: We need all the hosts to be present. Not just the first or last one. What if the result is reverse-ordered? What if it's not ordered at all and it returns a random set of found hosts? It's not necessarily defined.

sambible:
This doesn't seem to be a regression of the BZ, since you see the 3 results, you just don't see them on the same page, since you set the per-page setting to less than 3 results. 

The BZ seems to be that it would only return up to the per page value of results, so in this case that would mean you'd only see 3 results - which isn't what happens. You just dont get all the results on the same page.

Griffin-Sullivan:
@sambible what's the status of this PR?

jyejare:
@sambible Looks like the Airgun PR is merged! Would you want to resume on this ?

jyejare:
Ping @sambible , Are we planning to resume on this ?

ColeHiggins2 (commented):
@sambible Any updates on this ?

sambible:
Closing this, going to reeval this test as I think concerns are valid.

PR #11202, opened 2023-04-10
Unblock upgrade all tiers needing factory upgrade instance

jyejare:
We hide upgrade all tier those which needs factory upgraded satellite instance because it was not supported and we were waiting for upgrade_snap_template.

The template is now available and hence all the tests needing the facotry upgrade instance can run safely now. The CI changes have already commenced.

### DIFF
--- a/pytest_plugins/factory_collection.py
+++ b/pytest_plugins/factory_collection.py
@@ -11,30 +11,8 @@ def pytest_configure(config):
 def pytest_collection_modifyitems(session, items, config):
     from pytest_fixtures.core import sat_cap_factory
 
-    deselected = []
-    selected = []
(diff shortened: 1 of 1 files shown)


.github/workflows/autoRelease.yaml
```diff
@@ -0,0 +1,20 @@
+# When new tags gets created a new release would be automatically created
+name: Auto Release on new Tag
```
Griffin-Sullivan, line 2: Do we need to use this same workflow for airgun and nailgun?

jyejare: OOPs , that change shouldnt be part of this PR ! Removing ....

JacobCallahan (approved):

Griffin-Sullivan (approved):

PR #11215, opened 2023-04-11
Removing broken FM tests

Griffin-Sullivan:
I noticed that these tests were causing other foreman-maintain tests to fail. I'm gonna run the test suite on here and see if the pass percentage is better.


Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain

PR #10394, opened 2022-12-02
Convert UI manifest tests to manifester

synkd:
This PR will convert all UI tests that use manifests to obtain those manifests via manifester. I will aim to push one commit for each module that I'm converting. The PR will remain in draft state until I have converted all affected modules. The list of UI modules that I will be converting is as follows:
```
test_repository.py
test_activationkey.py
test_contenthost.py
test_dashboard.py
test_errata.py
test_hostcollection.py
test_host.py
test_organization.py
test_package.py
test_reporttemplates.py
test_rhcloud_insights.py
test_rhcloud_inventory.py
test_rhc.py
test_subscription.py
test_sync.py
```

PR #11216, opened 2023-04-11
[6.13.z] Several fixes to test_contenthost and repo_mixins

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/11158

Few updates here:

1. Changing the setup_content fixture to use API instead of CLI. This is because the CLI method of this helper doesn't currently work, and using the API instead provides the same functionality. 
2. Create a fixture that sets the default Host UI to the old one. This is required by all tests in the module, hence the `autouse=True`

Both of these fixes should result in much better results for the Hosts-Content component.

### DIFF
--- a/robottelo/host_helpers/repository_mixins.py
+++ b/robottelo/host_helpers/repository_mixins.py
@@ -659,7 +659,7 @@ class RepositoryCollection:
         if self.need_subscription:
             # upload manifest only when needed
             if upload_manifest and not self.organization_has_manifest(org_id):
-                self.satellite.upload_manifest(org_id, interface='CLI')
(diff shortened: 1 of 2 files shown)


Satellite-QE:
trigger: test-robottelo

synkd (commented):
Looking at the PRT run in Jenkins, this looks like a manifest issue. I have still not yet completed the UI manifester conversions, but I would be interested to see if the same test passes with this change: https://github.com/SatelliteQE/robottelo/pull/10394/commits/b1cdd0458fe1ad49d02897ed4d73734e03fcf171

omkarkhatavkar:
trigger: test-robottelo

PR #11223, opened 2023-04-12
skip and reduce tracer host tests

pondrejk:
There is BZ#2186188, also I find the api tests a bit redundant, thus removing.

### DIFF
--- a/tests/foreman/api/test_host.py
+++ b/tests/foreman/api/test_host.py
@@ -33,29 +33,9 @@ from nailgun import entities
 from requests.exceptions import HTTPError
 
 from robottelo.config import get_credentials
-from robottelo.config import settings
 from robottelo.constants import DEFAULT_CV
 from robottelo.constants import ENVIRONMENT
 from robottelo.utils import datafactory
-
-
-@pytest.fixture(scope='function')
-def tracer_host(katello_host_tools_tracer_host):
-    # create a custom, rhel version-specific mock-service repo
-    rhelver = katello_host_tools_tracer_host.os_version.major
-    katello_host_tools_tracer_host.create_custom_repos(
-        **{f'mock_service_rhel{rhelver}': settings.repos['MOCK_SERVICE_REPO'][f'rhel{rhelver}']}
-    )
-    katello_host_tools_tracer_host.execute(f'yum -y install {settings.repos["MOCK_SERVICE_RPM"]}')
-    assert (
-        katello_host_tools_tracer_host.execute(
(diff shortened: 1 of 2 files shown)


tests/foreman/api/test_host.py
```diff
@@ -1534,50 +1534,3 @@ def test_positive_bulk_destroy(self, module_org):
...
-        """
-        host = tracer_host.nailgun_host
-        package = settings.repos["MOCK_SERVICE_RPM"]
-        # mark the service log messages for later comparison and downgrade the pkg version
-        service_ver_log_old = tracer_host.execute(f'cat /var/log/{package}/service.log')
-        package_downgrade = tracer_host.execute(f'yum -y downgrade {package}')
-        assert package_downgrade.status == 0
-
-        # tracer should detect a new trace
-        traces = host.traces(data={'host_id': host.id})
```
Gauravtalreja1, line 1568: question, though API and CLI tests look similar, here we execute an endpoint to find new traces and resolve traces, but in CLI we've commands to the same, it doesn't look redundant, so why it should be removed?

pondrejk: The idea here is that since hammer visits those api endpoints anyway, it would be hit as well if anything happened to these endpoints. So the overlap between api and cli is significant

tests/foreman/api/test_host.py
```diff
@@ -1534,50 +1534,3 @@ def test_positive_bulk_destroy(self, module_org):
...
             with pytest.raises(HTTPError):
                 entities.Host(id=host_id).read()
-
-
-class TestHostTraces:
-    """Tests for host tracer"""
-
-    @pytest.mark.tier4
-    @pytest.mark.rhel_ver_match('[^6].*')
-    def test_positive_tracer_list_and_resolve(self, tracer_host):
```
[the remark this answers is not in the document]
pondrejk: done!

pondrejk:
Well, meanwhile the bz got to the on_qa state and will likely be finalized soon, so I removed the skip part and left just the docstring reference

ogajduse (approved):

pondrejk:
@Gauravtalreja1 @ogajduse if you feel strong about keeping the api tests, I could add these list and remediate commands into that cli test that has left, that would get both path tested but would spare us spinning ~6 machines twice

vsedmik (approved):
I think it's a strong argument that spinning up 6 VMs just to test the same endpoints which are called by hammer anyway seems quite redundant. So giving ACK.

Gauravtalreja1 (approved):

PR #11224, opened 2023-04-12
Add test for cloning a locked provisioning template

shubhamsg199:
We have tests for cloning a template but that is cloning a newly created template hence, adding a test for cloning an already locked provisioning template.

### DIFF
--- a/tests/foreman/ui/test_provisioningtemplate.py
+++ b/tests/foreman/ui/test_provisioningtemplate.py
@@ -81,6 +81,34 @@ def test_positive_clone(session, clone_setup):
 
 
 @pytest.mark.tier2
+def test_positive_clone_locked(session, target_sat):
+    """Assure ability to clone a locked provisioning template
+
+    :id: 2df8550a-fe7d-405f-ab48-2896554cda12
+
+    :Steps:
+        1.  Go to Provisioning template UI
+        2.  Choose a locked provisioning template and attempt to clone it
+
+    :expectedresults: The template is cloned
+
+    :CaseLevel: Integration
+    """
+    clone_name = gen_string('alpha')
+    with session:
+        session.provisioningtemplate.clone(
+            'Kickstart default',
+            {
+                'template.name': clone_name,
+            },
+        )
+        pt = target_sat.api.ProvisioningTemplate().search(query={'search': f'name=={clone_name}'})
+        assert pt, (
+            f'Template {clone_name} expected to exist but is not included in the search' 'results'
+        )
+
+
+@pytest.mark.tier2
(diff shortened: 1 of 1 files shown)


shubhamsg199:
trigger: test-robottelo
pytest: tests/foreman/ui/test_provisioningtemplate.py -k test_positive_clone_locked

Griffin-Sullivan (approved):
Looks good! If you wanted to parametrize this you could also change `clone_setup` to take in a param for `locked` or `custom` and then return the Provisioning Template you need for the test. That way you could run this all in one test. However, since the other test has some OS coverage too, I think it's fine to have two separate tests.

jyejare:
The PRT tests was passing, but I rebase this PR since PR checks were failing with `codecov` dependency which we already fixed.

tests/foreman/ui/test_provisioningtemplate.py
```diff
@@ -80,6 +80,35 @@ def test_positive_clone(session, clone_setup):
         assert set(clone_setup['os_list']) == {f'{os.name} {os.major}' for os in assigned_oses}
 
 
+@pytest.mark.tier2
+def test_positive_clone_locked(session, target_sat):
```
jyejare, line 84: Do you think the `caseLevel` and the `caseImportance` for this test case is same as defined at module level? If not please change is accordingly.

At least I doubt on caseLevel to be acceptance here, I think it should be `Integration` or something else like `system` etc.

tests/foreman/ui/test_provisioningtemplate.py
```diff
@@ -80,6 +80,35 @@ def test_positive_clone(session, clone_setup):
...
+        session.provisioningtemplate.clone(
+            'Kickstart default',
+            {
+                'template.name': clone_name,
+            },
+        )
+        pt = target_sat.api.ProvisioningTemplate().search(query={'search': f'name=={clone_name}'})
+        assert (
+            pt
+        ), 'Template {} expected to exist but is not included in the search' 'results'.format(
```
jyejare, line 107: Use `f-strings`

jyejare (commented):
ACK pending minor comments!

tests/foreman/ui/test_provisioningtemplate.py
shubhamsg199: yes, updated

tests/foreman/ui/test_provisioningtemplate.py
shubhamsg199: updated

jyejare (approved):

PR #11237, opened 2023-04-12
Add stream marker

JacobCallahan:
This change introduces a new marker that our associates can use to run new feature automation against stream builds.  This marker should be removed from all tests when branching robottelo. The marker itself should only live in the master branch.

### DIFF
--- a/pytest_plugins/markers.py
+++ b/pytest_plugins/markers.py
@@ -14,6 +14,7 @@ def pytest_configure(config):
         "destructive: Destructive tests",
         "upgrade: Upgrade tests",
         "e2e: End to end tests",
+        "stream: Tests unique to stream builds; purged when robottelo is branched.",
         "pit_server: PIT server scenario tests",
         "pit_client: PIT client scenario tests",
         "run_in_one_thread: Sequential tests",


pytest_plugins/markers.py
```diff
@@ -14,6 +14,7 @@ def pytest_configure(config):
         "destructive: Destructive tests",
         "upgrade: Upgrade tests",
         "e2e: End to end tests",
+        "stream: Tests unique to stream builds; purged when robottelo is branched.",
```
jyejare, line 17: Would you explain in terms of automation how would we purge the `stream` marker on master is branched?

Are we have to manually remove all `stream` markers manually?

If yes, isnt this a very manual and pathetic process we are offering in automation ? Whos responsible for purging?

Gauravtalreja1 (approved):

jyejare (commented):
ACK pending question!

pytest_plugins/markers.py
JacobCallahan: It wouldn't have to manual. A single `sed` command can remove them all, but yes they would need to be removed before branching. It would likely be a good idea to add a `create-branch` type script for all of the manual changes we're also doing around the repo.

sambible (approved):

pytest_plugins/markers.py
jyejare: @JacobCallahan Yeah, the plan is to auto-branch :) !

jyejare (approved):

PR #11238, opened 2023-04-12
Updated target_sat to be module scoped 

ColeHiggins2:
This test was failing in Jenkins with the issue below

```Owner has already imported from another subscription management application```

To fix this, I changed scope and updated target_sat to use module_entitlement_manifest_org

### DIFF
--- a/tests/foreman/api/test_subscription.py
+++ b/tests/foreman/api/test_subscription.py
@@ -458,9 +458,7 @@ def test_positive_os_restriction_on_repos():
     """
 
 
-def test_positive_async_endpoint_for_manifest_refresh(
-    target_sat, module_org, session_entitlement_manifest
-):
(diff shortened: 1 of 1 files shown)


ColeHiggins2:
trigger: test-robottelo
pytest: tests/foreman/api/test_subscription.py -k test_positive_async_endpoint_for_manifest_refresh

JacobCallahan (approved):
The description is misleading, but the content is good. target_sat here is still session-scoped, but now you're using a module-scoped manifest instead of a session-scoped one.

synkd (approved):

sambible (approved):

ColeHiggins2:
> The description is misleading, but the content is good. target_sat here is still session-scoped, but now you're using a module-scoped manifest instead of a session-scoped one.

Right, the manifest scope not the target_sat scope. my apologies

PR #11249, opened 2023-04-13
Test for multiple orgs with the same repo[Repository Rewrite]

ColeHiggins2:
This test is a part of the Repository Rewrite effort [ SAT-13253 ]

The purpose of this test to make sure the metadata for a repository is the same when synced to multiple organizations

Steps:
Create orgs with same repo, sync them, and check if the metadata information is matching all of them

### DIFF
--- a/tests/foreman/api/test_repositories.py
+++ b/tests/foreman/api/test_repositories.py
@@ -217,3 +217,29 @@ def test_negative_upload_expired_manifest(module_org, target_sat):
         "The manifest doesn't exist on console.redhat.com. "
         "Please create and import a new manifest." in error.value.stderr
     )
+
+
+@pytest.mark.tier1
+def test_positive_multiple_orgs_with_same_repo(target_sat):
+    """Test that multiple organizations with the same repository synced have matching metadata
+
+    :id: 39cff8ea-969d-4b8f-9fb4-33b1ba768ff2
+
+    :Steps:
+        1. Create multiple organizations
+        2. Sync the same repository to each organization
+        3. Assert that each repository from each organization contain the same content counts
+
+    :expectedresults: Each repository in each organziation should have the same content counts
+    """
+    repos = []
+    orgs = [target_sat.api.Organization().create() for _ in range(3)]
+    for org in orgs:
+        product = target_sat.api.Product(organization=org).create()
+        repo = target_sat.api.Repository(
+            content_type='yum', product=product, url=settings.repos.module_stream_1.url
+        ).create()
+        repo.sync()
+        repo_counts = target_sat.api.Repository(id=repo.id).read().content_counts
+        repos.append(repo_counts)
+    assert repos[0] == repos[1] == repos[2]
--- a/tests/foreman/api/test_subscription.py
+++ b/tests/foreman/api/test_subscription.py
@@ -448,9 +448,7 @@ def test_positive_os_restriction_on_repos():
     """
 
 
-def test_positive_async_endpoint_for_manifest_refresh(
-    target_sat, function_entitlement_manifest_org
-):
+def test_positive_async_endpoint_for_manifest_refresh(target_sat, module_entitlement_manifest_org):
     """Verify that manifest refresh is using an async endpoint. Previously this was a single,
     synchronous endpoint. The endpoint to retrieve manifests is now split into two: an async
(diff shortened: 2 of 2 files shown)


ColeHiggins2:
trigger: test-robottelo
pytest: tests/foreman/api/test_repositories.py -k test_positive_multiple_orgs_with_same_repo

tests/foreman/api/test_repositories.py
```diff
@@ -190,3 +190,29 @@ def test_positive_sync_kickstart_repo(self, module_entitlement_manifest_org, tar
...
+    """
+    repos = []
+    orgs = [target_sat.api.Organization().create() for _ in range(3)]
+    for org in orgs:
+        product = target_sat.api.Product(organization=org).create()
+        repo = target_sat.api.Repository(
+            content_type='yum', product=product, url=settings.repos.module_stream_1.url
+        ).create()
+        repo.sync()
+        repo_counts = target_sat.api.Repository(id=repo.id).read().content_counts
```
shweta83, line 216: We can add more assertions here and just read all the metadata of repos in place of the count.

tests/foreman/api/test_repositories.py
```diff
@@ -190,3 +190,29 @@ def test_positive_sync_kickstart_repo(self, module_entitlement_manifest_org, tar
...
+        3. Assert that each repository from each organization contain the same content counts
+
+    :expectedresults: Each repository in each organziation should have the same content counts
+    """
+    repos = []
+    orgs = [target_sat.api.Organization().create() for _ in range(3)]
+    for org in orgs:
+        product = target_sat.api.Product(organization=org).create()
+        repo = target_sat.api.Repository(
+            content_type='yum', product=product, url=settings.repos.module_stream_1.url
```
adarshdubey-star, line 213: Non-blocking:
We can extend this to all content types of repositories

Gauravtalreja1 (approved):

tests/foreman/api/test_repositories.py
ColeHiggins2: @shweta83 What other metadata would you want to read? my thought is that content_counts is the best assertion for this case.

tests/foreman/api/test_repositories.py
ColeHiggins2: @adarshdubey-star The repo I am syncing in this case is a yum only repo. parameterized content_type would not be possible unless we test this with multiple custom repos each a different type. that may qualify for a separate test. I can sync with support and find out.

Griffin-Sullivan (approved):

tests/foreman/api/test_repositories.py
synkd: RPM repositories contain a directory with multiple metadata files. This old Pulp doc contains a good summary of each of the files in typical yum repodata under the section `The Yum/RPM Repository` [1]. If the requirement that you're satisfying with this test is to ensure that the metadata is the same when the same repo is synced to multiple orgs, then I agree with Shweta that more than just the package count needs to be checked. Looking at the Repository Nailgun entity [2], it does not look like we're capturing all of that metadata in that entity. However, I would at least add checks against errata and packages to this test. @shweta83 were there other parts of the repodata that you were thinking of here? 

[1] https://docs.pulpproject.org/en/2.19/plugins/pulp_rpm/tech-reference/rpm.html
[2] https://github.com/SatelliteQE/nailgun/blob/84c8c95cedb2f319a710bee93f084ddf76d4bb84/nailgun/entities.py#L6407

synkd (changes requested):

ColeHiggins2:
addressing comments this week

adarshdubey-star:
We can rebase and merge with a PRT run!

ColeHiggins2:
trigger: test-robottelo
pytest: tests/foreman/api/test_repositories.py -k test_positive_multiple_orgs_with_same_repo

tests/foreman/api/test_repositories.py
ColeHiggins2: content_counts in this case returns a dictionary of rpm, erratum, package_groups, srpm, module_stream, etc. So we are asserting all of those are consistent between repos

synkd (approved):

PR #11253, opened 2023-04-14
Use AD 2019. Improve error detection in AD setup fixture.

lhellebr:


sambible (approved):
a very _assertive_ PR - curious though, what's the difference in AD versions?

lhellebr:
AFAIK no external difference. AD on Windows Server 2019 uses 2016 functional level.
https://learn.microsoft.com/en-us/answers/questions/177021/functional-level-windows-2019

I'm not sure why there are two versions in the automation in the first place. Yes, we support AD running on both but OS version isn't that important here plus there are also different versions. The matrix could be exceptionally large if we wanted to test everything, even if it's this insignificant.

pytest_fixtures/component/satellite_auth.py
```diff
@@ -45,7 +45,7 @@ def ldap_cleanup():
 def ad_data():
     supported_server_versions = ['2016', '2019']
 
-    def _ad_data(version='2016'):
+    def _ad_data(version='2019'):
```
Gauravtalreja1, line 48: You might need to update auth fixtures as this could change this condition `ad_data('2019') if '2019' in auth_type else ad_data()`?

omkarkhatavkar:
> AFAIK no external difference. AD on Windows Server 2019 uses 2016 functional level. https://learn.microsoft.com/en-us/answers/questions/177021/functional-level-windows-2019
> 
> I'm not sure why there are two versions in the automation in the first place. Yes, we support AD running on both but OS version isn't that important here plus there are also different versions. The matrix could be exceptionally large if we wanted to test everything, even if it's this insignificant.

In 2019-20, I worked on adding support and testing for this feature. However, at that time, the 2019 AD server was still new and most of our customers were still using the 2016 AD servers. Therefore, it was important for us to include both servers in our test matrix for authentication servers. If you want to reiterate the priority and supportability of this having discussions with the product owner, and programme members we may be able to reduce it.

omkarkhatavkar:
@lhellebr Do you just `CherryPick`, missing the versions, I guess it will be cherry-picked to all previously supported versions.

lhellebr:
> > AFAIK no external difference. AD on Windows Server 2019 uses 2016 functional level. https://learn.microsoft.com/en-us/answers/questions/177021/functional-level-windows-2019
> > I'm not sure why there are two versions in the automation in the first place. Yes, we support AD running on both but OS version isn't that important here plus there are also different versions. The matrix could be exceptionally large if we wanted to test everything, even if it's this insignificant.
> 
> In 2019-20, I worked on adding support and testing for this feature. However, at that time, the 2019 AD server was still new and most of our customers were still using the 2016 AD servers. Therefore, it was important for us to include both servers in our test matrix for authentication servers. If you want to reiterate the priority and supportability of this having discussions with the product owner, and programme members we may be able to reduce it.

2019 was never tested in automation. As you see, 2016 was hardcoded into robottelo. Since the difference shouldn't be big (same functionality level), I decided to just flip 2016 to 2019 and only migrate 2019 to OSP, leaving 2016 behind in RHV that is being deprecated. I think it's enough to support one Windows machine.

pytest_fixtures/component/satellite_auth.py
lhellebr: I think as it is now, the fixture returns 2019 by default. So if anything, that condition can be dropped. I didn't, however, because I want to make only minimal changes to migrate the instance and catch potential problems caused by it.

omkarkhatavkar:
@lhellebr We don't want on that time to run all tests in 2016 and 2019, but still wanted to have support for some 1-2 tests for 2019 at that time. e.g.  https://github.com/SatelliteQE/robottelo/blob/867c00dfebe77c2b0fe517ab3319d82732358370/tests/foreman/ui/test_ldap_authentication.py#L562

Yes 2016 was not hard coded but was default on that time as most of the customers were using that as AD server. We just wanted to have support for 2019 server as it was new.

pytest_fixtures/component/satellite_auth.py
omkarkhatavkar: I think if we are not supporting the 2016 AD server and guessing this is discussed with programme members and the product manager.  I think it's a wise decision to remove the dead code related to the 2016 AD server from robottelo, given that we're no longer supporting it. Although you wanted to keep changes minimal, we were planning to remove this code anyway.

I'd also like to point out that two tests, specifically https://github.com/SatelliteQE/robottelo/blob/8580597f43533d2546c4a6575ea813f8a262a9a1/tests/foreman/destructive/test_ldap_authentication.py#L220 and https://github.com/SatelliteQE/robottelo/blob/867c00dfebe77c2b0fe517ab3319d82732358370/tests/foreman/ui/test_ldap_authentication.py#L562, are currently using both the 2016 and 2019 AD servers. This means that these tests will run twice with the 2019 AD server, which is not the correct behavior.

omkarkhatavkar (changes requested):
mentioned in the comments need to get resolved.

pytest_fixtures/component/satellite_auth.py
```diff
@@ -45,7 +45,7 @@ def ldap_cleanup():
 def ad_data():
     supported_server_versions = ['2016', '2019']
```
omkarkhatavkar, line 46: we can remove 2016 from the list here.

omkarkhatavkar (approved):
ACK, pending on the CI checks.

pytest_fixtures/component/satellite_auth.py
lhellebr: I'm not sure why. It's still somewhat supported in robottelo, we're just missing an instance and are not using it currently.

omkarkhatavkar: okay got it.

omkarkhatavkar (approved):
ACK

PR #11271, opened 2023-04-16
added e2e markers to tests

ColeHiggins2:
Adding end to end pytests markers for:

Errata Management
Lifecycle Environment
Subscription Management

### DIFF
--- a/tests/foreman/ui/test_lifecycleenvironment.py
+++ b/tests/foreman/ui/test_lifecycleenvironment.py
@@ -31,6 +31,7 @@ from robottelo.constants import FAKE_3_CUSTOM_PACKAGE_NAME
 from robottelo.utils.datafactory import gen_string
 
 
+@pytest.mark.e2e
 @pytest.mark.upgrade
 @pytest.mark.tier2
 def test_positive_end_to_end(session):
--- a/tests/foreman/ui/test_errata.py
+++ b/tests/foreman/ui/test_errata.py
@@ -143,6 +143,7 @@ def vm(module_repos_collection_with_setup, rhel7_contenthost, target_sat):
     yield rhel7_contenthost
 
 
+@pytest.mark.e2e
 @pytest.mark.tier3
 @pytest.mark.parametrize('setting_update', ['remote_execution_by_default'], indirect=True)
 @pytest.mark.parametrize(
--- a/tests/foreman/ui/test_subscription.py
+++ b/tests/foreman/ui/test_subscription.py
@@ -65,6 +65,7 @@ def golden_ticket_host_setup(function_entitlement_manifest_org, module_target_sa
     return org, ak
 
 
+@pytest.mark.e2e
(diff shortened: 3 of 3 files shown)


tests/foreman/ui/test_lifecycleenvironment.py
```diff
@@ -31,6 +31,7 @@
 from robottelo.utils.datafactory import gen_string
 
 
+@pytest.mark.e2e
```
omkarkhatavkar, line 34: Invalid End-to-End Test.
 
A CRUD (Create, Read, Update, Delete) test by itself is not an end-to-end test, as it typically only focuses on testing the individual functionalities of creating, reading, updating, and deleting data within the application. However, a CRUD test could be a part of an end-to-end test, which would involve testing the entire workflow of a user interacting with the application and performing the CRUD operations as part of that workflow.

tests/foreman/ui/test_repository.py
```diff
@@ -464,6 +464,7 @@ def test_positive_resynchronize_rpm_repo(session, module_prod):
         assert int(repo_values['content_counts']['Packages']) >= 1
 
 
+@pytest.mark.e2e
```
omkarkhatavkar, line 467: Invalid End to End test

tests/foreman/ui/test_repository.py
```diff
@@ -530,6 +531,7 @@ def test_positive_end_to_end_custom_yum_crud(session, module_org, module_prod):
         assert not session.repository.search(module_prod.name, new_repo_name)
 
 
+@pytest.mark.e2e
```
omkarkhatavkar, line 534: invalid end to end test

omkarkhatavkar (changes requested):
I personally think CRUD tests can not be End to End Tests, Although, a CRUD test could be a part of an end-to-end test, which would involve testing the entire workflow of a user interacting with the application and performing the CRUD operations as part of that workflow.

Griffin-Sullivan (commented):
These are all UI tests. Do we have equivalent tests in API? UI is slower and flaky, so we are pushing for our E2E tests to be API-based. I think @omkarkhatavkar also has some valid concerns, so maybe translating these to API and adding some more E2E functionality would be better.

ColeHiggins2:
> These are all UI tests. Do we have equivalent tests in API? UI is slower and flaky, so we are pushing for our E2E tests to be API-based. I think @omkarkhatavkar also has some valid concerns, so maybe translating these to API and adding some more E2E functionality would be better.

@Griffin-Sullivan This task was to just identify already existing end to end tests. We currently do not have end to end tests in any api for this components. Adding these for the API is a great idea, just not for this PR

ColeHiggins2:
I removed all crud tests @omkarkhatavkar.

omkarkhatavkar (approved):
ACK

pondrejk (approved):

PR #11287, opened 2023-04-17
added e2e marks for Endeavour components

pondrejk:
More info in SAT-16616

### DIFF
--- a/tests/foreman/api/test_ping.py
+++ b/tests/foreman/api/test_ping.py
@@ -21,6 +21,7 @@ import pytest
 pytestmark = [pytest.mark.tier1, pytest.mark.upgrade]
 
 
+@pytest.mark.e2e
 @pytest.mark.build_sanity
 def test_positive_ping(target_sat):
     """Check if all services are running
--- a/tests/foreman/cli/test_hammer.py
+++ b/tests/foreman/cli/test_hammer.py
@@ -188,6 +188,7 @@ def test_positive_check_debug_log_levels(target_sat):
     assert 'log4j.logger.org.candlepin = WARN' in result.stdout
 
 
+@pytest.mark.e2e
 def test_positive_hammer_shell(target_sat):
     """Verify that hammer shell runs a command when input is provided via interactive/bash
 
--- a/tests/foreman/cli/test_hammer.py
+++ b/tests/foreman/cli/test_hammer.py
@@ -30,7 +30,7 @@ from robottelo.logging import logger
 
 HAMMER_COMMANDS = json.loads(DataFile.HAMMER_COMMANDS_JSON.read_text())
 
-pytestmark = [pytest.mark.tier1, pytest.mark.upgrade, pytest.mark.e2e]
+pytestmark = [pytest.mark.tier1, pytest.mark.upgrade]
 
 
 def fetch_command_info(command):
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -302,6 +302,7 @@ def test_positive_search_all_field_sets():
 
 
 # -------------------------- CREATE SCENARIOS -------------------------
+@pytest.mark.e2e
 @pytest.mark.cli_host_create
 @pytest.mark.tier1
 @pytest.mark.upgrade
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -357,6 +357,7 @@ def test_positive_create_and_delete(target_sat, module_lce_library, module_publi
         Host.info({'id': new_host['id']})
 
 
+@pytest.mark.e2e
 @pytest.mark.cli_host_create
 @pytest.mark.tier1
 def test_positive_crud_interface_by_id(target_sat, default_location, default_org):
@@ -945,6 +946,7 @@ def test_negative_create_with_incompatible_pxe_loader():
 
 
 # -------------------------- UPDATE SCENARIOS -------------------------
+@pytest.mark.e2e
 @pytest.mark.cli_host_update
 @pytest.mark.tier1
 def test_positive_update_parameters_by_name(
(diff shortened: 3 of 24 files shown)


tests/foreman/ui/test_domain.py
```diff
@@ -123,6 +123,7 @@ def test_positive_remove_parameter(session, valid_domain_name):
         assert param_name not in [param['name'] for param in params]
 
 
+@pytest.mark.e2e
```
Gauravtalreja1, line 126: We aren't considering UI tests for e2e, just API and CLI

tests/foreman/api/test_ping.py
```diff
@@ -21,6 +21,7 @@
 pytestmark = [pytest.mark.tier1, pytest.mark.upgrade]
 
 
+@pytest.mark.e2e
 @pytest.mark.build_sanity
 def test_positive_ping(target_sat):
     """Check if all services are running
```
Gauravtalreja1, line 27: cli/test_ping can also be considered for e2e, right?

jyejare: Having only tested with one interface should be fine , thats where we settled on right @JacobCallahan ?

tests/foreman/cli/test_host.py
```diff
@@ -302,6 +302,7 @@ def test_positive_search_all_field_sets():
 
 
 # -------------------------- CREATE SCENARIOS -------------------------
+@pytest.mark.e2e
```
jyejare, line 305: There are many tests in this test module which are good candidates for e2e. I dont have exact test cases to let you know but I think the one chosen here is covering less areas than what others are doing.

tests/foreman/api/test_ping.py
pondrejk: Well, probably not strictly e2e, though it felt like a basic gating test that could run against the stream

tests/foreman/cli/test_hammer.py
```diff
@@ -30,7 +30,7 @@
 
 HAMMER_COMMANDS = json.loads(DataFile.HAMMER_COMMANDS_JSON.read_text())
 
-pytestmark = [pytest.mark.tier1, pytest.mark.upgrade, pytest.mark.e2e]
+pytestmark = [pytest.mark.tier1, pytest.mark.upgrade]
```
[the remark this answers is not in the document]
pondrejk: Yeah I was considering this a lot, basically I expect the hammer commands and options to change vividly within stream so it would require a lot of effort to keep these tests green. To me it makes more sense to check the options after we branch when they get stabilized

tests/foreman/ui/test_host.py
```diff
@@ -275,6 +275,7 @@ def tracer_install_host(rex_contenthost, target_sat):
     yield rex_contenthost
 
 
+@pytest.mark.e2e
```
jyejare, line 278: I think there are other tests as well e.g `test_positive_global_registration_end_to_end`  why I could see only one test per module and not more ?

One more example could be `test_positive_crud_module_streams`

tests/foreman/ui/test_domain.py
pondrejk: I didn't know about that, but there is a bunch already marked and I don't have good API and CLI alternative

jyejare (changes requested):
ACK pending comments and questions.

2 things I observed with the PR and I dont expect it to be like that:

1. **Not** more than 1 tests are e2e selected per module
2. Mostly component tests are selected but the **integration tests are ignored**

Is this something been decided and we are following ?

tests/foreman/cli/test_host.py
pondrejk: added a few more, thanks

pondrejk:
> ACK pending comments and questions.
> 
> 2 things I observed with the PR and I dont expect it to be like that:
> 
>     1. **Not** more than 1 tests are e2e selected per module
> 
>     2. Mostly component tests are selected but the **integration tests are ignored**
> 
> 
> Is this something been decided and we are following ?

Take this as a first step, I was ask to quickly mark some tests to get some coverage in. But definitely there will be follow up with better analysis and possibly new tests created

tests/foreman/cli/test_hammer.py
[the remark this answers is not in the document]
lhellebr: Still, some of the tests can be kept, right? Like `test_positive_hammer_shell`

tests/foreman/ui/test_domain.py
lhellebr: Maybe we can mark whatever we want to be e2e and if the decision is strictly not to include UI tests, that can be done on pipeline level?

lhellebr (changes requested):
LGTM.

Some comments added.

I haven't reviewed UI tests since there seems to be an ongoing discussion on whether those are even acceptable to be marked as e2e.

I see there is no external authentication test included. If that's by choice, OK and non-blocking.

I see there is no test for the following components - while I understand some of them shouldn't be part of gating tests, I feel like this is too much coverage missing:
Dashboard (but that's UI)
E-mail
Hooks & Webhooks
Selinux
Roles
LDAP
Notifications
Search

tests/foreman/cli/test_hammer.py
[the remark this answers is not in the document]
pondrejk: Brought that one back

pondrejk:
@lhellebr for *Roles* and *LDAP* there has been some tests already marked before this PR, for *search* and basically all the rest there either is no test coverage at all or the existing tests don't feel right to include as e2e. So these will need more work to create some e2e coverage.

pondrejk:
@jyejare @Gauravtalreja1 mind revisiting?

jyejare (approved):

omkarkhatavkar (approved):
ACK

PR #10930, opened 2023-03-10
[errata] - Upgrade scenario refactor

ogajduse:

PR #11299, opened 2023-04-19
Install katello-agent if it is not installed already

ogajduse:
This functionality was removed in #10930. This PR brings it back but puts a different logic in place.

This fixture now assumes that if satellite-installer RC == 0, katello-agent fixture was installed successfully.

### DIFF
--- a/pytest_fixtures/component/katello_agent.py
+++ b/pytest_fixtures/component/katello_agent.py
@@ -7,11 +7,20 @@ from robottelo.utils.installer import InstallerCommand
 
 @pytest.fixture(scope='module')
 def sat_with_katello_agent(module_target_sat):
-    """Assert that katello-agent feature is enabled on module_target_sat"""
-    result = module_target_sat.install(
-        InstallerCommand(help='| grep foreman-proxy-content-enable-katello-agent')
+    """Enable katello-agent feature on module_target_sat if not enabled yet."""
+    katello_agent_enabled = (
+        'true'
+        in module_target_sat.install(
+            InstallerCommand(help='| grep foreman-proxy-content-enable-katello-agent')
+        ).stdout
     )
-    assert 'true' in result.stdout, 'katello-agent feature is not enabled on Satellite'
+    if not katello_agent_enabled:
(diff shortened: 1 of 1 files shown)


omkarkhatavkar (approved):

pytest_fixtures/component/katello_agent.py
```diff
@@ -7,11 +7,20 @@
...
 def sat_with_katello_agent(module_target_sat):
-    """Assert that katello-agent feature is enabled on module_target_sat"""
-    result = module_target_sat.install(
-        InstallerCommand(help='| grep foreman-proxy-content-enable-katello-agent')
+    """Enable katello-agent feature on module_target_sat if not enabled yet."""
+    katello_agent_enabled = (
+        'true'
+        in module_target_sat.install(
+            InstallerCommand(help='| grep foreman-proxy-content-enable-katello-agent')
+        ).stdout
```
Gauravtalreja1, line 15: Just Curious, Why do we need this check before enabling katello-agent? 

I've played around with this earlier, when you don't register the system and try enabling it using an installer, where the installer fails but this flag is set to true, I don't think it's a good idea to rely on this flag to know if it's enabled.

And instead, we should check this flag after RC for the satellite-installer command to know if its correctly enabled, wdyt?

ogajduse: I get where you are coming from. This fixture is also being used on the Satellite instances that are created from the upgrade template where the katello-agent feature is installed already, so the check is there to save time.

Generally, I would say that we can rely on these flags to determine whether the feature is installed or not. Installer reads the scenario file and prints the values from that file in the help message. I can imagine having a function/property in the Satellite class that returns that answer file and we can read the feature installation state from there.

> And instead, we should check this flag after RC for the satellite-installer command to know if its correctly enabled, wdyt?

We have discussed this with @vsedmik in the chat and the conclusion was that assuming that satellite-installer's RC was 0 is enough - means that the feature was installed successfully.

pondrejk (approved):

PR #11315, opened 2023-04-21
Update puppet e2e tests 

Gauravtalreja1:
Currently, there are 45+ tests running for Puppet so to make that count less and for effective stream testing for Puppet.
1. Combined ConfigReport info and delete tests into a single CRD test
2. Add e2e test in api/test_environment.py similar to cli
3. Mark tests e2e instead of class to avoid -ve tests in smart class parameters test modules.

### DIFF
--- a/tests/foreman/api/test_environment.py
+++ b/tests/foreman/api/test_environment.py
@@ -30,6 +30,83 @@ from robottelo.utils.datafactory import parametrized
 from robottelo.utils.datafactory import valid_environments_list
 
 
+@pytest.mark.e2e
+@pytest.mark.tier1
+@pytest.mark.upgrade
+def test_positive_CRUD_with_attributes(
+    session_puppet_enabled_sat, module_puppet_org, module_puppet_loc
+):
+    """Check if Environment with attributes can be created, updated and removed
+
+    :id: d2187971-86b2-40c9-a93c-66f37691ae2c
+
+    :BZ: 1337947
+
+    :expectedresults:
+        1. Environment is created and has parameters assigned
+        2. Environment can be listed by parameters
+        3. Environment can be updated
+        4. Environment can be removed
+    """
+    puppet_sat = session_puppet_enabled_sat
+    # Create with attributes
+    env_name = gen_string('alpha')
+    environment = puppet_sat.api.Environment(
+        name=env_name, organization=[module_puppet_org], location=[module_puppet_loc]
+    ).create()
+    assert env_name == environment.name
+    assert len(environment.organization) == 1
+    assert environment.organization[0].id == module_puppet_org.id
+    assert len(environment.location) == 1
+    assert environment.location[0].id == module_puppet_loc.id
+
+    # List by name
+    envs = puppet_sat.api.Environment().search(query=dict(search=f'name={env_name}'))
+    assert len(envs) == 1
+    assert envs[0].name == env_name
+
+    # List by org loc id
+    envs = puppet_sat.api.Environment().search(
+        query=dict(
+            search=f'organization_id={module_puppet_org.id} and location_id={module_puppet_loc.id}'
+        )
+    )
+    assert env_name in [res.name for res in envs]
+
+    # List by org loc name
+    envs = puppet_sat.api.Environment().search(
+        query=dict(
+            search=f'organization={module_puppet_org.name} and location={module_puppet_loc.name}'
+        )
+    )
+    assert env_name in [res.name for res in envs]
+
+    # Update org and loc
+    new_org = puppet_sat.api.Organization().create()
+    new_loc = puppet_sat.api.Location().create()
+    env = puppet_sat.api.Environment(id=environment.id, organization=[new_org]).update(
+        ['organization']
+    )
+    assert len(env.organization) == 1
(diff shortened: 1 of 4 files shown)


Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/ --component Puppet -m e2e

tests/foreman/api/test_environment.py
```diff
@@ -30,6 +30,83 @@
...
+    )
+    assert env_name in [res.name for res in envs]
+
+    # List by org loc name
+    envs = puppet_sat.api.Environment().search(
+        query=dict(
+            search=f'organization={module_puppet_org.name} and location={module_puppet_loc.name}'
+        )
+    )
+    assert env_name in [res.name for res in envs]
```
adarshdubey-star, line 82: Do we again need to search and assert by name/id?
Also if we tend to keep this one we must do these steps post-update as well.
```suggestion
```

tests/foreman/api/test_environment.py
```diff
@@ -30,6 +30,83 @@
...
+    env_name = gen_string('alpha')
+    environment = puppet_sat.api.Environment(
+        name=env_name, organization=[module_puppet_org], location=[module_puppet_loc]
+    ).create()
+    assert env_name == environment.name
+    assert len(environment.organization) == 1
+    assert environment.organization[0].id == module_puppet_org.id
+    assert len(environment.location) == 1
+    assert environment.location[0].id == module_puppet_loc.id
+
```
adarshdubey-star, line 62: Should we also assert the name here keeping in mind the search query we are using below?
```suggestion
assert environment.location[0].name ==module_puppet_loc.name
assert environment.organization[0].name ==module_puppet_org.name

```

tests/foreman/cli/test_report.py
```diff
@@ -34,43 +32,30 @@ def run_puppet_agent(session_puppet_enabled_sat):
...
 @pytest.mark.tier1
-def test_positive_info(run_puppet_agent, session_puppet_enabled_sat):
-    """Test Info for Puppet report
+def test_positive_CRD_satellite(run_puppet_agent, session_puppet_enabled_sat):
+    """Test puppet-agent creates a report for satellite when its run, read and delete by its ID
 
     :id: 32646d4b-7101-421a-85e0-777d3c6b71ec
 
-    :expectedresults: Puppet Report Info is displayed
+    :expectedresults: Puppet reports are generated, readable, and deleteable
```
adarshdubey-star, line 42: How is the report generated here?

Gauravtalreja1: reports are generated when puppet-agent is run, which is written in fixture above

tests/foreman/api/test_environment.py
Gauravtalreja1: Yes, I think, this scenario I've written with reference to existing tests of cli, and its good to have

tests/foreman/api/test_environment.py
```diff
@@ -30,6 +30,83 @@
...
+    new_org = puppet_sat.api.Organization().create()
+    new_loc = puppet_sat.api.Location().create()
+    env = puppet_sat.api.Environment(id=environment.id, organization=[new_org]).update(
+        ['organization']
+    )
+    assert len(env.organization) == 1
+    assert env.organization[0].id == new_org.id
+    assert env.organization[0].id != module_puppet_org.id
+
+    env = puppet_sat.api.Environment(id=environment.id, location=[new_loc]).update(['location'])
```
omkarkhatavkar, line 94: why are we making two different calls? one for org and location?

tests/foreman/api/test_environment.py
Gauravtalreja1: I guess this wouldn’t add much value, and you said search query below is already using it so its covered already

tests/foreman/api/test_environment.py
Gauravtalreja1: we can't directly update both in one call i guess, one way is to read first and then update both, so it will be again two calls, and here we update org first and then loc

tests/foreman/api/test_environment.py
adarshdubey-star: Non blocking:
Also if we keep this one we must do these steps post-update.

adarshdubey-star (approved):
Ack, with one non-blocking comment.

tests/foreman/api/test_environment.py
omkarkhatavkar: `puppet_sat.api.Environment(id=environment.id, organization=[new_org], location=[new_loc]).update(['organization', 'location'])`

Gauravtalreja1: It works, updated, thanks for the suggestion!

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/ --component Puppet -m e2e

omkarkhatavkar (approved):
ACK

omkarkhatavkar:
@Gauravtalreja1 can you please fix the CI failures ?

omkarkhatavkar:
CI failure is unrelated, fix it available here for that https://github.com/SatelliteQE/manifester/pull/20

PR #11316, opened 2023-04-21
Hosts: use current Sat. host Organization and location

pnovotny:
UI tests that use the Satellite host now switch to the actual org. and location.

Don't assume that Satellite host is always in the Default Organization and Default Location.

### DIFF
--- a/pytest_fixtures/component/taxonomy.py
+++ b/pytest_fixtures/component/taxonomy.py
@@ -21,6 +21,20 @@ def default_location(session_target_sat):
 
 
 @pytest.fixture
+def current_sat_org(target_sat):
+    """Return the current organization assigned to the Satellite host"""
+    sat_host = target_sat.api.Host().search(query={'search': f'name={target_sat.hostname}'})[0]
+    return sat_host.organization.read().name
+
+
+@pytest.fixture
+def current_sat_location(target_sat):
+    """Return the current location assigned to the Satellite host"""
+    sat_host = target_sat.api.Host().search(query={'search': f'name={target_sat.hostname}'})[0]
+    return sat_host.location.read().name
+
+
+@pytest.fixture
 def function_org(target_sat):
     return target_sat.api.Organization().create()
 
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -34,7 +34,6 @@ from robottelo.config import settings
 from robottelo.constants import ANY_CONTEXT
 from robottelo.constants import DEFAULT_CV
 from robottelo.constants import DEFAULT_LOC
-from robottelo.constants import DEFAULT_ORG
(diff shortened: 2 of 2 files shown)


pnovotny:
trigger: test-robottelo
pytest: tests/foreman/ui/test_host.py -k 'test_positive_host_details_read_templates or test_positive_manage_table_columns'

Griffin-Sullivan (approved):
Looks good! During your debugging, which organization was the host entry for satellite under?

sambible (approved):

lhellebr (approved):
LGTM.
Just out of curiosity, how did it happen that the Satellite host wasnt in Default org/loc?

omkarkhatavkar:
I think there could be a test that removes the sat host object from the default organization. This test is able to fetch the current organization and location. I feel the it is the more dynamic approach.

omkarkhatavkar (approved):

lvrtelov:
> I think there could be a test that removes the sat host object from the default organization. This test is able to fetch the current organization and location. I feel the it is the more dynamic approach.

Omkar can you expand on what would be the end goal of removing the host object from default org?

ogajduse (approved):
I am going to merge this PR as there are enough ACKs and this PR is critical to the 6.13 feature.
@omkarkhatavkar Feel free to discuss potential RFEs for this test or a request for a new test in the Slack automation channel.

pytest_fixtures/component/taxonomy.py
```diff
@@ -20,6 +20,20 @@ def default_location(session_target_sat):
...
+    """Return the current organization assigned to the Satellite host"""
+    sat_host = target_sat.api.Host().search(query={'search': f'name={target_sat.hostname}'})[0]
+    return sat_host.organization.read().name
+
+
+@pytest.fixture
+def current_sat_location(target_sat):
+    """Return the current location assigned to the Satellite host"""
+    sat_host = target_sat.api.Host().search(query={'search': f'name={target_sat.hostname}'})[0]
+    return sat_host.location.read().name
```
Gauravtalreja1, line 34: Nice, I think instead of fixture we can add these to satellie_mixins to call from sat_object similar to the nailgun_proxy property (non-blocking)

And Ideally, sat should be in default org and default loc, if it's not then it could possibly be a bug I assume, so can we add an assertation here as well to check if its default org and loc always if we don't assume?

Gauravtalreja1: Oh @ogajduse merged already, @pnovotny maybe followup PR can be used if this suggestion make sense

pnovotny: @Gauravtalreja1 good point, I'll create new PR and move the functionality from the fixtures to sat_mixins as you suggested.

PR #11332, opened 2023-04-25
Refactor logging

rplevka:
The idea is to unify and add a proper structure to the logs.
- removal of `logzero` and relying purely on native `logging` library
- external lib loggers (broker, manifester) initialization, so they use the same handlers.
- introduction of logger hierarchy per module (adds structure to the logs)
- change file-logging strategy: each fixture and test uses its own log file (makes logging to external resources possible)

### DIFF
--- a/pytest_fixtures/component/katello_certs_check.py
+++ b/pytest_fixtures/component/katello_certs_check.py
@@ -2,7 +2,7 @@
 from pathlib import Path
 
 import pytest
-from broker import Broker
+from broker.broker import Broker
 from fauxfactory import gen_string
 
 from robottelo.constants import CERT_DATA as cert_data
--- a/pytest_fixtures/core/sys.py
+++ b/pytest_fixtures/core/sys.py
@@ -4,6 +4,7 @@ import pytest
 
 from robottelo.config import settings
 from robottelo.hosts import SatelliteHostError
+from robottelo.logging import logger
 
 
 @pytest.fixture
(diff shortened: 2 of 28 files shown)


pytest_fixtures/component/katello_certs_check.py
```diff
@@ -2,7 +2,7 @@
 from pathlib import Path
 
 import pytest
-from broker import Broker
+from broker.broker import Broker
```
JacobCallahan, line 5: Raised a question about this in the Broker PR, so I won't repeat it further in this PR, but this is likely a blocker for me.

pytest_fixtures/core/sys.py
```diff
@@ -24,6 +25,7 @@ def allow_repo_discovery(target_sat):
 @pytest.fixture(autouse=True, scope="session")
 def relax_bfa(session_target_sat):
     """Relax BFA protection against failed login attempts"""
+    logger.info('xxx inside relax_bfa fixture body')
```
JacobCallahan, line 28: natural language is typically easy enough to differentiate in logs without trigger characters
```suggestion
    logger.info('inside relax_bfa fixture body')
```

tests/foreman/api/test_ping.py
```diff
@@ -29,6 +29,7 @@ def test_positive_ping(target_sat):
 
     :expectedresults: Overall and individual services status should be 'ok'.
     """
+    logger.info("xxx inside test body")
```
JacobCallahan, line 32: same comment as before
```suggestion
    logger.info("inside test body")
```

JacobCallahan (changes requested):
I'll reserve comments about the functionality of this PR until I see it demonstrated.

pytest_fixtures/core/sys.py
rplevka: sorry, those are my working logs, so i can easily find and remove them when i'm finished :D

PR #11349, opened 2023-04-25
Adding puppet enabled backup and restore

Griffin-Sullivan:
Adding a test that mostly copies the functionality of `test_positive_backup_restore` for testing backup and restore on a puppet enabled satellite. The reasoning for adding a new test and not parametrizing is because the run time of the test is quite significant (8 tests * 15 - 30 minutes each). If we parametrize that test we now run it 16 times. Instead I thought copying the test and running it once for offline and once for online would be enough coverage.

During my testing I also discovered a bug with our backup and restore tests. In 6.13, `pg_globals.dump` was removed, so I removed that file from the list and updated anything that depended on it (like `REMOTE_SAT_FILES`).

### DIFF
--- a/tests/foreman/maintain/test_backup_restore.py
+++ b/tests/foreman/maintain/test_backup_restore.py
@@ -35,9 +35,8 @@ BACKUP_DIR = "/tmp/"
 BASIC_FILES = {"config_files.tar.gz", ".config.snar", "metadata.yml"}
 CONTENT_FILES = {"pulp_data.tar", ".pulp.snar"}
 OFFLINE_FILES = {"pgsql_data.tar.gz", ".postgres.snar"}
-ONLINE_SAT_FILES = {"candlepin.dump", "foreman.dump", "pulpcore.dump", "pg_globals.dump"}
+ONLINE_SAT_FILES = {"candlepin.dump", "foreman.dump", "pulpcore.dump"}
 ONLINE_CAPS_FILES = {"pulpcore.dump"}
-REMOTE_SAT_FILES = ONLINE_SAT_FILES - {"pg_globals.dump"}
 
 
 NODIR_MSG = "ERROR: parameter 'BACKUP_DIR': no value provided"
@@ -50,7 +49,7 @@ assert_msg = "Some required backup files are missing"
 def get_exp_files(sat_maintain, backup_type):
     if type(sat_maintain) is Satellite:
         if sat_maintain.is_remote_db():
-            expected_files = BASIC_FILES | REMOTE_SAT_FILES
+            expected_files = BASIC_FILES | ONLINE_SAT_FILES
         else:
             expected_files = (
                 BASIC_FILES | OFFLINE_FILES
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_backup_restore.py::test_positive_puppet_backup_restore

tests/foreman/maintain/test_backup_restore.py
```diff
@@ -47,22 +46,10 @@
...
-    if type(sat_maintain) is Satellite:
-        if sat_maintain.is_remote_db():
-            expected_files = BASIC_FILES | REMOTE_SAT_FILES
-        else:
-            expected_files = (
-                BASIC_FILES | OFFLINE_FILES
-                if backup_type == 'offline'
-                else BASIC_FILES | ONLINE_SAT_FILES
-            )
-    else:
```
synkd, line 60: Have you run the other tests in this module to ensure that your changes don't break them? My hunch is that removing this part of the conditional could affect tests in this module that run against Capsules.

synkd (commented):
Looks good overall, but I would like to see confirmation that the changes to `get_exp_files()` won't impact other tests in this module.

tests/foreman/maintain/test_backup_restore.py
Griffin-Sullivan: We can run the backup and restore tests on PRT here and see.

synkd: It looks like the tests with Capsules did indeed fail, but the errors were all with attaching subscriptions, which at first glance does not appear to be connected to this specific change. What do you think, @Griffin-Sullivan?

Griffin-Sullivan: Yeah I was looking through them. I'm gonna do some local runs tomorrow just to double check. I'm not super worried about the change though. It's just removing a constant that is unnecessary.

Griffin-Sullivan: I just realized that I misread `ONLINE_CAPS_FILES` as `ONLINE_SAT_FILES` in this method and that's why I gutted it. I'm now testing it locally with a fix to see if it will work.

Griffin-Sullivan:
All PRT tests are passing.

synkd (approved):

ogajduse (approved):

tests/foreman/maintain/test_backup_restore.py
```diff
@@ -409,6 +408,88 @@ def test_negative_restore_baddir(sat_maintain, setup_backup_tests):
...
+
+    :customerscenario: true
+
+    :BZ: 2158896
+    """
+    # Enable puppet on the Satellite
+    sat_maintain.enable_puppet_satellite()
+
+    subdir = f'{BACKUP_DIR}backup-{gen_string("alpha")}'
+    instance = 'satellite' if type(sat_maintain) is Satellite else 'capsule'
```
JacobCallahan, line 447: Non-blocking, but for the future I'd prefer to use `isinstance` here.

tests/foreman/maintain/test_backup_restore.py
```diff
@@ -409,6 +408,88 @@ def test_negative_restore_baddir(sat_maintain, setup_backup_tests):
...
+        backup_type=backup_type,
+        options={'assumeyes': True, 'plaintext': True, 'skip-pulp-content': False},
+    )
+    assert result.status == 0
+    assert 'FAIL' not in result.stdout
+
+    # Check for expected files
+    backup_dir = re.findall(fr'{subdir}\/{instance}-backup-.*-[0-5][0-9]', result.stdout)[0]
+    files = sat_maintain.execute(f'ls -a {backup_dir}').stdout.split('\n')
+    files = [i for i in files if not re.compile(r'^\.*$').search(i)]
```
JacobCallahan, line 459: also non-blocking but in the future, you may want to consider using `filter` here.

PR #11365, opened 2023-04-27
Auto Release Robottelo

jyejare:
Now, the release for changelogs will be tagged in this repo from the SatQE CI job.

How?
--------
1. Upon receiving UMB message for the new snap for a major downstream release in development, the CI job triggers.
2. The CI job sends an HTTP dispatch request to the `auto release` workflow in this repo (created in this PR)
3. Upon receiving the HTTP request, the workflow runs with new tag name and creates a new tag in the repo.
4. The same tag is then used to create a new release with automatic release notes.

So all the releases here would now be automatically created.

### DIFF
--- a/.github/workflows/dispatch_release.yml
+++ b/.github/workflows/dispatch_release.yml
@@ -1,0 +1,31 @@
+### The auto release workflow triggered through dispatch request from CI
+name: auto-release
+
+# Run on workflow dispatch from CI
+on:
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        type: string
+        description: Name of the tag
+
+jobs:
+  auto-tag-and-release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Git User setup
+        run: "git config --local user.email Satellite-QE.satqe.com && git config --local user.name Satellite-QE"
+
+      - name: Tag latest commit
+        run: "git tag -a ${{ github.event.inputs.tag_name }} -m 'Tagged By SatelliteQE Automation User'"
+
+      - name: Push the tag to the upstream
+        run: "git push ${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}.git --tags"
+
+      - name: create a new release from the tag
+        env:
+          credentials: ${{ secrets.GH_TOKEN }}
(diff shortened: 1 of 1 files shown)


ogajduse (approved):
This is definitely a huge improvement and a nicer solution than the previously presented one. :+1:

However this solution works, I have found the following GH Action that does everything in three steps compared to the 5 steps in your solution.
https://github.com/Bullrich/generate-release-changelog

Take it as a possible enhancement or simplification.  This comment is definitely non-blocking.

jyejare:
@ogajduse We need not to use a github action to generate release notes as the HTTP request for creating a release provides a simple parameter to generate release notes. Also I contributed to action-automatic-releases that releases on tag to genrate-release-notes automatically (https://github.com/marvinpinto/action-automatic-releases/pull/4 ) and the plan in the future is to use that.

.github/workflows/dispatch_release.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Git User setup
+        run: "git config --local user.email Satellite-QE.satqe.com && git config --local user.name Satellite-QE"
+
+      - name: Tag latest commit
+        run: "git tag -a ${{ github.event.inputs.tag_name }} -m 'Tagged By SatelliteQE Automation User'"
+
+      - name: Push the tag to the upstream
+        run: "git push ${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}.git --tags"
+
+      - name: create a new release from the tag
```
omkarkhatavkar, line 28: Just of consistency`Create a new release from the tag`

omkarkhatavkar (approved):
ACK, non blocking comment

Griffin-Sullivan (approved):

.github/workflows/dispatch_release.yml
```diff
@@ -0,0 +1,31 @@
...
+        env:
+          credentials: ${{ secrets.GH_TOKEN }}
+        run: "curl -L -X POST -H \"Authorization: Bearer ${{ secrets.SATQE_GH_TOKEN }}\" ${GITHUB_API_URL}/repos/${GITHUB_REPOSITORY}/releases -d '{\"tag_name\": \"${{ github.event.inputs.tag_name }}\", \"target_commitish\":\"master\", \"name\":\"${{ github.event.inputs.tag_name }}\", \"draft\":false, \"prerelease\":true, \"generate_release_notes\": true}'"
```
omkarkhatavkar, line 31: Add a check here that the release was successful, if not add a slack notification.

PR #11379, opened 2023-04-28
Run Autocherrypick on merged PRs

jyejare:
$subject


JacobCallahan (approved):

ogajduse:
I was wondering if we could make use of some GitHub Apps for some of our processes.
https://github.com/apps/patchback

https://github.com/ansible-collections/community.general/pull/6366#issuecomment-1515162002

jyejare:
@ogajduse We did discuss the need of GHA now outside of this PR in Slack, and we are open to consider apps in future if the current functionality shows limitations.

So could you re-review again and if it looks good, should we just merge it now ?

PR #11086, opened 2023-03-28
Rewrite tests to have one assertion per test

ekohl:
It's a best practice to only have a single assertion per test. The benefit is that all the other unrelated tests still run. This means that you get a full report of all issues, and more precise reporting.

It does this by extracting the systemctl service status checks to a separate, parametrized test. That is also rewritten to use systemctl is-active instead of parsing the output of status. It still adds the systemctl status output as context, which makes debugging easier.

The hammer ping command is then rewritten to collect all not-ok services and assert that it's empty. This gives you a complete list of services that aren't OK with their results, instead of just a single failed service.

Right now this is untested and probably doesn't comply with the Robottelo standards, but it came up in a discussion and showing code is easier than describing it.

PR #11392, opened 2023-05-02
[6.13.z] Rewrite tests to have one assertion per test

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/11086

It's a best practice to only have a single assertion per test. The benefit is that all the other unrelated tests still run. This means that you get a full report of all issues, and more precise reporting.

It does this by extracting the systemctl service status checks to a separate, parametrized test. That is also rewritten to use systemctl is-active instead of parsing the output of status. It still adds the systemctl status output as context, which makes debugging easier.

The hammer ping command is then rewritten to collect all not-ok services and assert that it's empty. This gives you a complete list of services that aren't OK with their results, instead of just a single failed service.

Right now this is untested and probably doesn't comply with the Robottelo standards, but it came up in a discussion and showing code is easier than describing it.

### DIFF
--- a/tests/foreman/installer/test_installer.py
+++ b/tests/foreman/installer/test_installer.py
@@ -1340,14 +1340,33 @@ def test_positive_selinux_foreman_module(target_sat):
 
 @pytest.mark.upgrade
 @pytest.mark.tier1
-def test_positive_check_installer_services(target_sat):
-    """Check if services start correctly
+@pytest.mark.parametrize('service', SATELLITE_SERVICES)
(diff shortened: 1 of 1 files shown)


Satellite-QE:
trigger: test-robottelo
pytest: tests/foreman/installer/test_installer.py::test_positive_check_installer_service_running tests/foreman/installer/test_installer.py::test_positive_check_installer_hammer_ping

ogajduse (approved):

ogajduse:
PRT has passed, merging.

PR #10844, opened 2023-03-02
Improve kerberos cli testing

lhellebr:

PR #11398, opened 2023-05-03
[6.13.z] Cherrypick ad 2019

lhellebr:
Cherrypick and answer to https://github.com/SatelliteQE/robottelo/pull/11253#issuecomment-1512501994 : No :)

### DIFF
--- a/pytest_fixtures/component/satellite_auth.py
+++ b/pytest_fixtures/component/satellite_auth.py
@@ -182,7 +182,7 @@ def ldap_auth_source(
         # entity create with AD settings
         auth_source = module_target_sat.api.AuthSourceLDAP(
             onthefly_register=True,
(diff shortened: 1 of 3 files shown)


omkarkhatavkar (approved):

omkarkhatavkar:
@lhellebr this also needs to be added to 6.12.z and 6.11.z? I think we need to do that as well.

lhellebr:
Yes. I wanted to make that part by part because I wasn't sure how it would work without also cherrypicking https://github.com/SatelliteQE/robottelo/pull/10844 but since this is just reacting to a config change, it *should* work. Since you brought it up, I'm marking with corresponding labels now.

Gauravtalreja1 (approved):

PR #11399, opened 2023-05-03
webhook bz coverage added

pondrejk:
Extending the e2e to check the password wasn't wiped by the webhook update (bz#2119155)

### DIFF
--- a/tests/foreman/ui/test_webhook.py
+++ b/tests/foreman/ui/test_webhook.py
@@ -31,7 +31,7 @@ def test_positive_end_to_end(session, target_sat):
 
     :CaseImportance: Critical
 
-    :BZ: 2112093, 2131771
+    :BZ: 2112093, 2131771, 2119155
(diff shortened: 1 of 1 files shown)


lhellebr (approved):

synkd (approved):

ogajduse (approved):
@pondrejk Since the bug is spread across all supported Satellite versions, would it make sense to cherry-pick the patch also to `6.12.z` and possibly to `6.11.z`?

pondrejk:
> @pondrejk Since the bug is spread across all supported Satellite versions, would it make sense to cherry-pick the patch also to `6.12.z` and possibly to `6.11.z`?

I was thinking I'd do a manual cherrypick as the bz numbers are different for these versions, but maybe that's overthinking as these cloned bzs would be findable trough the parent bz. So adding z-stream tags, thanks

PR #11411, opened 2023-05-04
fix in search bookmark assertion

pondrejk:
This wasn't skipped on 1992652, hence failing since 6.10, then after the bz was closed wontfix half a year ago no amendments were made. So this pr changes the assertions so that the test doesn't rely on an error box that is not shown

### DIFF
--- a/tests/foreman/ui/test_bookmarks.py
+++ b/tests/foreman/ui/test_bookmarks.py
@@ -17,7 +17,6 @@
 :Upstream: No
 """
 import pytest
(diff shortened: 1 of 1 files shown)


pondrejk:
> Looks good! Should we maybe remove 1992652 from the docstring since it was never fixed or leave it for historical purposes?

I guess there's no harm in keeping it there

shweta83 (approved):
Ack

vsedmik (approved):

Gauravtalreja1 (approved):

Issue #11410, opened 2023-05-04
[Failed-AutoCherryPick] - Adding puppet enabled backup and restore


Satellite-QE:
#### Auto-Cherry-Pick WorkFlow Failure:
- To Branch: 6.13.z
- [Failed Cherrypick Action](https://github.com/SatelliteQE/robottelo/actions/runs/4884730843)
- [Parent Pull Request](https://github.com/SatelliteQE/robottelo/pull/11349)

PR #11414, opened 2023-05-04
[6.13.z] Adding puppet enabled backup and restore

Griffin-Sullivan:
Cherry pick for failure: https://github.com/SatelliteQE/robottelo/issues/11410

### DIFF
--- a/tests/foreman/maintain/test_backup_restore.py
+++ b/tests/foreman/maintain/test_backup_restore.py
@@ -35,9 +35,8 @@ BACKUP_DIR = "/tmp/"
 BASIC_FILES = {"config_files.tar.gz", ".config.snar", "metadata.yml"}
 CONTENT_FILES = {"pulp_data.tar", ".pulp.snar"}
 OFFLINE_FILES = {"pgsql_data.tar.gz", ".postgres.snar"}
-ONLINE_SAT_FILES = {"candlepin.dump", "foreman.dump", "pulpcore.dump", "pg_globals.dump"}
+ONLINE_SAT_FILES = {"candlepin.dump", "foreman.dump", "pulpcore.dump"}
 ONLINE_CAPS_FILES = {"pulpcore.dump"}
-REMOTE_SAT_FILES = ONLINE_SAT_FILES - {"pg_globals.dump"}
 
 
 NODIR_MSG = "ERROR: parameter 'BACKUP_DIR': no value provided"
@@ -50,7 +49,7 @@ assert_msg = "Some required backup files are missing"
 def get_exp_files(sat_maintain, backup_type):
     if type(sat_maintain) is Satellite:
         if sat_maintain.is_remote_db():
-            expected_files = BASIC_FILES | REMOTE_SAT_FILES
+            expected_files = BASIC_FILES | ONLINE_SAT_FILES
         else:
             expected_files = (
                 BASIC_FILES | OFFLINE_FILES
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_backup_restore.py::test_positive_backup_restore tests/foreman/maintain/test_backup_restore.py::test_positive_puppet_backup_restore

tests/foreman/maintain/test_backup_restore.py
```diff
@@ -35,9 +35,8 @@
 BASIC_FILES = {"config_files.tar.gz", ".config.snar", "metadata.yml"}
 CONTENT_FILES = {"pulp_data.tar", ".pulp.snar"}
 OFFLINE_FILES = {"pgsql_data.tar.gz", ".postgres.snar"}
-ONLINE_SAT_FILES = {"candlepin.dump", "foreman.dump", "pulpcore.dump", "pg_globals.dump"}
+ONLINE_SAT_FILES = {"candlepin.dump", "foreman.dump", "pulpcore.dump"}
 ONLINE_CAPS_FILES = {"pulpcore.dump"}
-REMOTE_SAT_FILES = ONLINE_SAT_FILES - {"pg_globals.dump"}
```
Gauravtalreja1, line 40: I think this @vsedmik added this earlier for remotedb testing, so are you not going to test backup restore against remotedb?

vsedmik: I think Griffin is fixing the fact that `pg_globals.dump` was removed from 6.13 backups, so now `ONLINE_SAT_FILES == REMOTE_SAT_FILES`, if I got it right.

Griffin-Sullivan: yep @vsedmik is correct. We no longer use `pg_globals.dump`. See https://bugzilla.redhat.com/show_bug.cgi?id=2172540

vsedmik (approved):

tests/foreman/maintain/test_backup_restore.py
```diff
@@ -409,6 +408,89 @@ def test_negative_restore_baddir(sat_maintain, setup_backup_tests):
...
+        3. restore succeeds
+        4. system health check succeeds
+        5. content is present after restore
+
+    :customerscenario: true
+
+    :BZ: 2158896
+    """
+    # Enable puppet on the Satellite
+    sat_maintain.enable_puppet_satellite()
```
Gauravtalreja1, line 444: @Griffin-Sullivan Can we not use session_puppet_enabled_sat here instead or else disable Puppet in the teardown?

Griffin-Sullivan: I guess we could, but I felt that it was safer keeping sat_maintain since that is what the test suite was built around and I didn't want to mess with the test. Although, if we make this change I will now need to update master, so I would rather merge this and if you feel it's necessary to use the fixture then I will add a PR updating master and then cherry pick that to 6.13.z.

Gauravtalreja1: Sounds good, using that fixture could save time and wouldn't conflict with the sat_maintain fixture so other tests could run without errors if the installer fails, and also you could handle disabling Puppet in the teardown as well to workaround

Gauravtalreja1 (approved):

Gauravtalreja1 (approved):

PR #11428, opened 2023-05-05
Update robottelo image when there's change in Airgun/Nailgun repo

jameerpathan111:
Problem statement: Currently, the Robottelo image is only updated when a Robottelo PR is merged, which means that it won't be updated if there are any changes merged in Airgun/Nailgun. This can cause issues when running automation as it might not have the latest code changes from these repositories.

Solution: We can add a new [repository_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch) trigger that [listens for any changes](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#create-a-repository-dispatch-event) in the Airgun/Nailgun repositories. When any PR is merged in those repos, the trigger will run the workflow that updates the Robottelo image. This ensures that the Robottelo image is always up-to-date with the latest code changes.

Depends on https://github.com/SatelliteQE/airgun/pull/843

Additional info: This PR also remove `.github/workflows/merge_to_master.yml` which was used for performing codechecks and building and pushing robottelo image to quay after a pr is merged. Performing codechecks again after merging a pr is not really needed/helpful.

Update: Now using [workflow_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) instead of repository_dispatch.


.github/workflows/merge_to_master.yml
```diff
@@ -1,102 +0,0 @@
...
-          cp .env.example .env
-
-      - name: Pre Commit Checks
-        uses: pre-commit/action@v3.0.0
-
-      - name: Collect Tests
-        run: |
-          pytest --collect-only --disable-pytest-warnings tests/foreman/ tests/robottelo/
-          pytest --collect-only --disable-pytest-warnings -m pre_upgrade tests/upgrades/
-          pytest --collect-only --disable-pytest-warnings -m post_upgrade tests/upgrades/
```
lpramuk, line 52: no more collecting tests ?

.github/workflows/merge_to_master.yml
```diff
@@ -1,102 +0,0 @@
...
-          pip install -U pip
-          pip uninstall -y pycurl
-          pip install --compile --no-cache-dir pycurl
-          pip install -U --no-cache-dir -r requirements.txt -r requirements-optional.txt
-          for conffile in conf/*.yaml.template; do mv -- "$conffile" "${conffile%.yaml.template}.yaml"; done
-          cp broker_settings.yaml.example broker_settings.yaml
-          cp .env.example .env
-
-      - name: Pre Commit Checks
-        uses: pre-commit/action@v3.0.0
```
lpramuk, line 46: no more pre-commit ?

lpramuk (changes requested):
Why some actions are missing after the change ?

lpramuk (approved):

JacobCallahan:
@jameerpathan111 will there be a companion PR opened against nailgun as well?

omkarkhatavkar (approved):

jameerpathan111:
> @jameerpathan111 will there be a companion PR opened against nailgun as well?

yes, I'll raise another pr once we merge this.

jyejare (approved):

jyejare (approved):

jameerpathan111:
Moving this to draft state for now as there might be a few issues with this which Omkar pointed out to me in PM.

.github/workflows/update_robottelo_image.yml
```diff
@@ -0,0 +1,52 @@
...
+
+on:
+  push:
+    branches:
+      - master
+      - 6.*.z
+  repository_dispatch:
+    types:
+      - airgun-pr-merged
+      - nailgun-pr-merged
```
omkarkhatavkar, line 12: If we have cherry-picked this to 6.*.z  branches then any pr merged in the Airgun/Nailgun runs the deploying containers of all the cherrypick branches irrespective of what branch is merged. @jameerpathan111 can you please test that and see if that is not a condition

.github/workflows/update_robottelo_image.yml
```diff
@@ -0,0 +1,60 @@
...
+jobs:
+  robottelo_container:
+    name: Update Robottelo container image on Quay.
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Get branch info
+        id: repo_branch
+        run: |
+          BRANCH="${{ github.event.client_payload.branch }}"
```
omkarkhatavkar, line 23: I would rather use the below approach so that evaluating the ${{ github.event.client_payload.branch }} would not be a problem. Did you test this for both even types?  This comment is non-blocking 
```

    steps:
      - name: Check event type
        if: github.event_name == 'push'
        run: echo "This workflow was triggered by a push event"
      - name: Check event type
        if: github.event_name == 'repository_dispatch'
        run: echo "This workflow was triggered by a repository dispatch event"
```

.github/workflows/update_robottelo_image.yml
```diff
@@ -0,0 +1,60 @@
...
+          if [ -z "$BRANCH" ]; then
+            BRANCH="${GITHUB_REF##*/}"
+          fi
+          echo "REPO_BRANCH=$BRANCH" >> $GITHUB_OUTPUT
```
omkarkhatavkar, line 27: In the case of a push event, this above line will not be executed as evaluating this expression would fail `BRANCH="${{ github.event.client_payload.branch }}"` nothing would run. this is my extensive guess can you confirm once? I'm okay it that is working then

jyejare:
Ping @jameerpathan111 !!

JacobCallahan:
@jameerpathan111 what's the status of this PR?

jameerpathan111:
@jyejare @JacobCallahan I have updated the PRs and now using workflow_dispatch :) Tested it with my dummy repos.

ogajduse (approved):

.github/workflows/update_robottelo_image.yml
```diff
@@ -0,0 +1,46 @@
...
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+
+      - name: Build and push image to Quay
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          file: ./Dockerfile
+          push: true
+          tags: ${{ secrets.QUAY_SERVER }}/${{ secrets.QUAY_NAMESPACE }}/robottelo:${{ steps.image_tag.outputs.IMAGE_TAG }}
```
Gauravtalreja1, line 46: Non-blocking, I remember we had some discussion on this earlier, but can you try using the redhat-actions here instead to push to quay? checkout an example here https://github.com/redhat-actions/push-to-registry#examples

jameerpathan111: @Gauravtalreja1 updated the PR, used `redhat-actions` to create and push image.

jameerpathan111: should work but didn't test

jameerpathan111: Seems to be working, tested with my fork and quay creds.

Gauravtalreja1 (approved):
@jameerpathan111 Thanks for this update and using our own `redhat-actions` to build images/push to quay.

PR #11450, opened 2023-05-10
one more bz tag added

pondrejk:
...for added hammer option in 2154512

### DIFF
--- a/tests/foreman/cli/test_hammer.py
+++ b/tests/foreman/cli/test_hammer.py
@@ -86,7 +86,7 @@ def test_positive_all_options(target_sat):
 
     :expectedresults: All expected options are present
 
-    :BZ: 2119053
+    :BZ: 2119053, 2154512
 
     :customerscenario: true
     """


synkd (approved):

tests/foreman/cli/test_hammer.py
```diff
@@ -86,7 +86,7 @@ def test_positive_all_options(target_sat):
 
     :expectedresults: All expected options are present
 
-    :BZ: 2119053
+    :BZ: 2119053, 2154512
```
Gauravtalreja1, line 89: question: I believe there'll be numerous scenarios that this all-options test has covered, are we planning to include them all? (non-blocking)

shweta83 (approved):

tests/foreman/cli/test_hammer.py
pondrejk: That would certainly highlight the value of this test, but I'm  not sure how I would find those, how long into history to go, etc. I'd like to add more bz as they appear in upcoming releases but I wasn't planning to do a backwards search. That said, if you have a candidate bz in mind I have no problem with adding

PR #11467, opened 2023-05-17
Handle stream versions in version cached_property

Gauravtalreja1:


ogajduse (approved):
Looks sane to me. I would like to have @jyejare's eyes on it too.

JacobCallahan:
@Gauravtalreja1 can you look into the PRT result?

shweta83:
> trigger: test-robottelo pytest: tests/foreman/api/test_provisioning.py::test_rhel_pxe_provisioning provisioning: true

@Gauravtalreja1 Can  you try another test case. This is on_prem. I don't think it will run in PRT.

jyejare (approved):
ACK !

Gauravtalreja1:
> 
> @Gauravtalreja1 Can you try another test case. This is on_prem. I don't think it will run in PRT.

Yes @shweta83 we surely can, few weeks back we added support for it, we add  `provisioning: true` to PRT comment to run on_prem tests, and one imp thing we shouldn't quote reply on PRT comments it triggers a PRT run when it finds related context and fails as its not right format

shweta83:
> > @Gauravtalreja1 Can you try another test case. This is on_prem. I don't think it will run in PRT.
> 
> Yes @shweta83 we surely can, few weeks back we added support for it, we add `provisioning: true` to PRT comments to run on_prem tests, and one imp thing we shouldn't quote reply on PRT comments it triggers a PRT run when it finds related to context and fails as its not right format, here PRT run 2508 is pointing to the same invalid PRT run.
> 
> Whereas, PRT run 2507 ran correctly, and el7 provisioning passed and has el8 and el9 provisioning failures, which are unrelated and are due to new el8.x and 9.x releases, will fix them in follow-up PR as it will require cherrypicks to zstream branches as well

Sure. Thanks @Gauravtalreja1

PR #11479, opened 2023-05-18
Auto select activation key when single ak is while registering host using global registration

shweta83:
Close loop BZ#[1994654](https://bugzilla.redhat.com/show_bug.cgi?id=1994654)


shweta83:
Since the BZ is not yet fixed so moving in draft.

shweta83:
The BZ related to this PR is still in new state so closing this PR.

PR #11481, opened 2023-05-18
Updating service enable/disable to test on stream

Griffin-Sullivan:
See https://bugzilla.redhat.com/show_bug.cgi?id=1995783 for bug description. When https://github.com/theforeman/foreman_maintain/pull/730 lands in stream this should pass and we can verify the BZ.

PR #11514, opened 2023-05-22
Adding better coverage to enable/disable stream test

Griffin-Sullivan:
https://github.com/SatelliteQE/robottelo/pull/11481 got merged before I could investigate. Apparently this test as it was before can pass on Satellite, but the case from the BZ will still fail. The original intent was to have the test case fail and then we can check automation when we get the stream snap with the fix in it. This test will now fail until the fix is added to stream.

### DIFF
--- a/tests/foreman/maintain/test_service.py
+++ b/tests/foreman/maintain/test_service.py
@@ -130,7 +130,6 @@ def test_positive_service_stop_restart(sat_maintain):
     assert result.status == 0
 
 
-@pytest.mark.stream
 @pytest.mark.include_capsule
 def test_positive_service_enable_disable(sat_maintain):
     """Enable/Disable services using satellite-maintain service subcommand
@@ -149,10 +148,17 @@ def test_positive_service_enable_disable(sat_maintain):
 
     :customerscenario: true
     """
+    result = sat_maintain.cli.Service.stop()
+    assert 'FAIL' not in result.stdout
+    assert result.status == 0
     result = sat_maintain.cli.Service.disable()
     assert 'FAIL' not in result.stdout
     assert result.status == 0
     result = sat_maintain.cli.Service.enable()
+    assert 'FAIL' not in result.stdout
+    assert result.status == 0
+    sat_maintain.power_control(state='reboot')
+    result = sat_maintain.cli.Service.status(options={'brief': True, 'only': 'foreman.service'})
     assert 'FAIL' not in result.stdout
     assert result.status == 0
 


Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_service.py -m stream

Griffin-Sullivan:
The current PRT run should fail with `AssertionError: assert 'inactive (dead)' not in '● foreman.s...oreman.org\n'`

tests/foreman/maintain/test_service.py
```diff
@@ -149,12 +149,18 @@ def test_positive_service_enable_disable(sat_maintain):
...
+    assert result.status == 0
     result = sat_maintain.cli.Service.disable()
     assert 'FAIL' not in result.stdout
     assert result.status == 0
     result = sat_maintain.cli.Service.enable()
     assert 'FAIL' not in result.stdout
     assert result.status == 0
+    sat_maintain.power_control(state='reboot')
+    result = sat_maintain.execute('systemctl status foreman.service')
+    assert 'inactive (dead)' not in result.stdout
```
Gauravtalreja1, line 163: We could also use cli.Service.status() here for checking status of foreman.service

Gauravtalreja1: you need to use `foreman-maintain service status -b --only foreman.service` which runs the same systemd command eventually

sambible (approved):

tests/foreman/maintain/test_service.py
```diff
@@ -130,7 +130,6 @@ def test_positive_service_stop_restart(sat_maintain):
     assert result.status == 0
 
 
-@pytest.mark.stream
```
Griffin-Sullivan, line 133: About to branch so no need for this to be stream.

Griffin-Sullivan:
I guess this won't run PRT unless I have the stream marker, so just take my word for it that this gives us the desired failure locally.

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_service.py -k test_positive_service_enable_disable

Griffin-Sullivan:
@Gauravtalreja1 unless the fix has made it in, this PR is supposed to fail PRT with the failure I posted above. This way when the fix is in for the BZ I can check automation to make sure it passed and then I can verify.

Gauravtalreja1:
@Griffin-Sullivan Okay, so do you want me to wait for a fix in the upcoming stream snap or merge as is with the local results that you'll provide?

Griffin-Sullivan:
I'd say merge and I will monitor CI to make sure the test passes when we get the snap with the fix.

Gauravtalreja1 (approved):

PR #11542, opened 2023-05-25
Add CLI test for applied filters

vsedmik:
Satellite 6.14.0 is going to implement indication and listing of filters applied to a CVV. This PR adds a test to ensure the correct functionality in hammer CLI.

Note: The upstream changes tested in this PR should land in the next stream snap (16.0). Expect the PRT to fail until then.

### DIFF
--- a/tests/foreman/cli/test_contentviewfilter.py
+++ b/tests/foreman/cli/test_contentviewfilter.py
@@ -16,6 +16,8 @@
 
 :Upstream: No
 """
+import random
+
 import pytest
 from fauxfactory import gen_string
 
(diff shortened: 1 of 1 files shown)


synkd (approved):
Looks good overall. ACK pending PRT results after the next stream snap, with one non-blocking question about test design.

tests/foreman/cli/test_contentviewfilter.py
```diff
@@ -973,3 +975,87 @@ def test_negative_delete_by_name(self, content_view):
...
+        # Publish first CVV with no filters applied
+        target_sat.cli.ContentView.publish({'id': content_view['id']})
+        cvv = target_sat.cli.ContentView.info({'id': content_view['id']})['versions'][0]
+        cvv_info = target_sat.cli.ContentView.version_info(
+            {'id': cvv['id'], 'include-applied-filters': 'true'}
+        )
+        assert cvv_info['has-applied-filters'] == 'no'
+        assert 'applied-filters' not in cvv_info
+
+        # Randomly add filters to the CV, assert correct CVV info values
```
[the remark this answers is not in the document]
vsedmik: Well, I hope I can. :grinning: 
One reason was that the `has-applied-filters` flag was already tested per filter type (and per test run) in the API counterpart in #11461 (originally the flag was the only thing planned to be implemented in API). But now, with the applied filters even listed via CLI, I felt I need to test that the filters are listed correctly for each published CV version, no matter the filter content type nor inclusion type. Please note that there can be _many_ filters of different (or same, or any combination) types associated with a single CV version. So, what would be the full matrix? No matter the answer it wouldn't be reasonable to test it full, so I decided to go random, gave it 100+ runs to get some statistic (all passed) and pushed for review.
Still, I can see the point of "missing anything on a given automation run". We could change the `k` on line 1014 to limit the risk, or ensure every filter type is present at least once. The price is the test time, as usual.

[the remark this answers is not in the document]
shweta83: Since the functionality is tested in other PR, I think the extent and scope of testing the flag `has-applied-filters` covered in this test should be sufficient.

shweta83 (approved):

rmynar (approved):
I learnt how to use _ as throwaway variable, thanks :-)

tests/foreman/cli/test_contentviewfilter.py
```diff
@@ -973,3 +975,87 @@ def test_negative_delete_by_name(self, content_view):
...
+                    'type': f_type,
+                    'inclusion': random.choice(['true', 'false']),
+                },
+            )
+            filters_applied.append(cvf)
+
+            target_sat.cli.ContentView.publish({'id': content_view['id']})
+            cvv = max(
+                target_sat.cli.ContentView.info({'id': content_view['id']})['versions'],
+                key=lambda x: int(x['id']),
```
lhellebr, line 1029: Version with highest id? That's fishy. Comparing versions would make more sense and be future-proof. I think this will work in every reasonable scenario though. Not blocking.

lhellebr (approved):

tests/foreman/cli/test_contentviewfilter.py
vsedmik: Yes, I need the latest CV version published. Normally I would go with something like `cv_info['versions'][-1]`, but there is a bug so we sometimes get the versions unsorted: https://bugzilla.redhat.com/show_bug.cgi?id=2209758

vsedmik: Aaah, gotcha. It would have to be `float(x['version'])` and we would need to pray the versioning won't change from `X.Y` to `X.Y.Z` anytime soon. Making it less future-proof in my eyes.
```
'9.0' > '10.0'
Out[4]: True
float('9.0') > float('10.0')
Out[5]: False
```

lhellebr: You sure can code it more generally. Comparing it as floats would be really ugly :)
But as I said, this is not blocking.

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/cli/test_contentviewfilter.py::TestContentViewFilter::test_positive_check_filters_applied

Gauravtalreja1 (approved):

PR #11545, opened 2023-05-26
Add test for add/remove ansible_role to host/hostgroup via API

Gauravtalreja1:
Closeloop [BZ#2164400](https://bugzilla.redhat.com/show_bug.cgi?id=2164400)

Dependant PR https://github.com/SatelliteQE/nailgun/pull/927


Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/api/test_ansible.py
nailgun: 927

tests/foreman/api/test_ansible.py
```diff
@@ -123,3 +129,56 @@ def test_positive_ansible_job_on_host(target_sat, module_org, rhel_contenthost):
...
+    host_roles = target_host.list_ansible_roles()
+    assert host_roles[0]['name'] != SELECTED_ROLE
+
+
+@pytest.mark.e2e
+@pytest.mark.stream
+@pytest.mark.tier2
+def test_add_and_remove_ansible_role_hostgroup(target_sat):
+    """
+    Test add and remove functionality for ansible roles in hostgroup via cli
```
Griffin-Sullivan, line 143: Do you mean API? The `expectedresults` also mentions cli.

tests/foreman/api/test_ansible.py
```diff
@@ -123,3 +129,56 @@ def test_positive_ansible_job_on_host(target_sat, module_org, rhel_contenthost):
...
+    :expectedresults:
+        1. Ansible role assign/add/remove functionality should work as expected in cli
+
+    :BZ: 2164400
+    """
+    ROLE_NAMES = [
+        'theforeman.foreman_scap_client',
+        'redhat.satellite.hostgroups',
+        'RedHatInsights.insights-client',
+    ]
```
Griffin-Sullivan, line 163: It seems like we are defining constants on the test case level when traditionally they should be in the constants file. Non-blocking, but I think it would be better to organize the roles as constants in the correct file for all of our ansible role tests.

tests/foreman/api/test_ansible.py
```diff
@@ -123,3 +129,56 @@ def test_positive_ansible_job_on_host(target_sat, module_org, rhel_contenthost):
...
+    ROLES = []
```
Griffin-Sullivan, line 164: Nitpicking but technically a constant in Python shouldn't change value. In this case we are adding to a constant list which is changing it's value. I would expect this to be defined as all lowercase: `roles`.

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/api/test_ansible.py
nailgun: 927

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/api/test_ansible.py -k test_add_and_remove_ansible_role_hostgroup
nailgun: 927

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/api/test_ansible.py -k test_add_and_remove_ansible_role_hostgroup
nailgun: 927

tests/foreman/api/test_ansible.py
Gauravtalreja1: ah, copypasta! Updated.

tests/foreman/api/test_ansible.py
```diff
@@ -123,3 +129,56 @@ def test_positive_ansible_job_on_host(target_sat, module_org, rhel_contenthost):
...
+    hg_name = gen_string('alpha')
+    hg = target_sat.api.HostGroup(name=hg_name).create()
+    assert hg.name == hg_name
+    proxy_id = target_sat.nailgun_smart_proxy.id
+    target_sat.api.AnsibleRoles().sync(data={'proxy_id': proxy_id, 'role_names': [ROLES]})
```
adarshdubey-star, line 169: This might be the issue with the PRT failure.
```suggestion
    target_sat.api.AnsibleRoles().sync(data={'proxy_id': proxy_id, 'role_names': [ROLE_NAMES]})
```

tests/foreman/api/test_ansible.py
Gauravtalreja1: No, it is not a constant, and we use different roles for different tests as per requirements, so its better to keep this specific to test

tests/foreman/api/test_ansible.py
Gauravtalreja1: ah, got it, thanks

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/api/test_ansible.py -k test_add_and_remove_ansible_role_hostgroup
nailgun: 927

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/api/test_ansible.py
nailgun: 927

Griffin-Sullivan (approved):

synkd (approved):

adarshdubey-star (approved):

tests/foreman/api/test_ansible.py
```diff
@@ -123,3 +129,56 @@ def test_positive_ansible_job_on_host(target_sat, module_org, rhel_contenthost):
     )
     result = target_sat.api.JobInvocation(id=job['id']).read()
     assert result.succeeded == 1
+
+    target_sat.api.Host(id=target_host.id).remove_ansible_role(data={'ansible_role_id': role_id})
+    host_roles = target_host.list_ansible_roles()
+    assert len(host_roles) == 0
```
jyejare, line 135: This test is highly vulnerable in the future to fail if any default roles are added to the satellite. Instead, you should verify that `role` in not in list of roles.

tests/foreman/api/test_ansible.py
```diff
@@ -123,3 +129,56 @@ def test_positive_ansible_job_on_host(target_sat, module_org, rhel_contenthost):
...
+    """
+    ROLE_NAMES = [
+        'theforeman.foreman_scap_client',
+        'redhat.satellite.hostgroups',
+        'RedHatInsights.insights-client',
+    ]
+    ROLES = []
+    hg_name = gen_string('alpha')
+    hg = target_sat.api.HostGroup(name=hg_name).create()
+    assert hg.name == hg_name
```
jyejare, line 167: Not really needed to HG creation in othercomponent tests! HG is part of your setup so, just create hg without separate `hg_name` var.

tests/foreman/api/test_ansible.py
```diff
@@ -123,3 +129,56 @@ def test_positive_ansible_job_on_host(target_sat, module_org, rhel_contenthost):
...
+    proxy_id = target_sat.nailgun_smart_proxy.id
+    target_sat.api.AnsibleRoles().sync(data={'proxy_id': proxy_id, 'role_names': ROLE_NAMES})
+    for role in ROLE_NAMES:
+        role_id = target_sat.api.AnsibleRoles().search(query={'search': f'name={role}'})[0].id
+        ROLES.append(role_id)
```
jyejare, line 172: You could have used list comprehension here to generate roles objects without extra ROLES var.

tests/foreman/api/test_ansible.py
```diff
@@ -123,3 +129,56 @@ def test_positive_ansible_job_on_host(target_sat, module_org, rhel_contenthost):
...
+
+    target_sat.api.HostGroup(id=hg.id).assign_ansible_roles(data={'ansible_role_ids': ROLES[:2]})
+    for a, b in zip(target_sat.api.HostGroup(id=hg.id).list_ansible_roles(), ROLE_NAMES[:2]):
```
jyejare, line 175: Nice use of Zip. But provided some meaningful names instead of  a/b.

jyejare (approved):
ACK with non-blocking code improvements comment!

tests/foreman/api/test_ansible.py
Gauravtalreja1: we're already testing this things with default roles, and we need to sync/import roles to the satellite before assigning it to host, also we've handled roles removal in every ansible test, so it should list roles assigned in specific test only.

tests/foreman/api/test_ansible.py
Gauravtalreja1: @jyejare Yes, I knew it would be raised in review, I couldn't find any related names as we're validating roles names from API with defined constant, so wasn't getting any suitable names here, and it will be extend the line unnecessary that's why I preferred a/b, do you've any suggestions for the names? :)

jyejare: a can be `role` and `b` can be `role_name`.

tests/foreman/api/test_ansible.py
Gauravtalreja1: Didn't get you here, hg_name var isn't required in the test but required for setup, I think its better to validate if hg is created before moving ahead with test

tests/foreman/api/test_ansible.py
Gauravtalreja1: ROLES isn't just used for roles object, we need role_ids for ansible endpoints as it only works with ids, but using list comprehension here is nice suggestion, will update it.

tests/foreman/api/test_ansible.py
jyejare: We should never assert test setups, infact we assume it passes. There are tests to validate if hg is being created with the name. Need not to really waste 2 extra lines on it. You can `assert hg` but not really required to check its name is same.

Its optional comment, leaving upto you to accept/ignore.

Gauravtalreja1: Yes, I get it, I removed it, thanks

tests/foreman/api/test_ansible.py
Gauravtalreja1: Updated

tests/foreman/api/test_ansible.py
Gauravtalreja1: Actually, both returns role names only, this names would be little confusing later, so I kept this iterative vars as r1 and r2 instead of a/b, I wouldn't want extend lines here so would like to keep it short, let me know if you still find this vars confusing

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/api/test_ansible.py
nailgun: 927

tests/foreman/api/test_ansible.py
Gauravtalreja1: Actually, both returns role names, so it'll be little confusing if we assign this names, instead I'd prefer something short like r1 and r2 instead for iterative variables, it keeps this loop in single line, let me know if you think it won't make sense?

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/api/test_ansible.py
nailgun: 927

shubhamsg199 (approved):

vsedmik (approved):
Looks good to me. We just need to get the nailgun counterpart merged first, and there is one comment pending.

tests/foreman/api/test_ansible.py
jyejare: ok, lets not waste more time on it :)

Issue #167, opened 2013-12-10
Improve robottelo_runner.py and run-tests.py


elyezer:
Improve the runners to not need to export `PYTHONPATH` when running them.

PR #11553, opened 2023-05-29
leapp scenario 8.8 to 9.2

vijaysawant:
**PR Description**
- Upgrading Rhel host to next major RHEL Realse with job templates
- Path RHEL 8.8 to 9.2

**Steps:**
1. Import a subscription manifest and enable, sync source & target repositories
2. Create CV, add repositories, create lce, ak, etc.
3. Register contenthost using ak
4. Varify target rhel repositories are enable on Satellite
5. Update all packages, install leapp tool and fix inhibitors (in advance)
6. Run Leapp Preupgrade and Leapp Upgrade job template

**Expected output:**
- Update rhel major version to another major version

### DIFF
--- a/tests/foreman/cli/test_leapp_client.py
+++ b/tests/foreman/cli/test_leapp_client.py
@@ -1,0 +1,293 @@
+"""Tests for leapp upgrade of content hosts with Satellite
+
+:Requirement: leapp
+
+:CaseLevel: Integration
+
+:CaseComponent: LeappIntegration
+
+:Team: Rocket
+
+:TestType: Functional
+
+:CaseImportance: High
+
+:CaseAutomation: Automated
+
+:Upstream: No
+"""
+import pytest
+from broker import Broker
+
+from robottelo.config import settings
+from robottelo.constants import PRDS
+from robottelo.hosts import ContentHost
+from robottelo.logging import logger
+
+synced_repos = pytest.StashKey[dict]
+
+RHEL7_VER = '7.9'
+RHEL8_VER = '8.8'
+RHEL9_VER = '9.2'
+
+RHEL_REPOS = {
+    'rhel7_server': {
+        'id': 'rhel-7-server-rpms',
+        'name': f'Red Hat Enterprise Linux 7 Server RPMs x86_64 {RHEL7_VER}',
+        'releasever': RHEL7_VER,
+        'reposet': 'Red Hat Enterprise Linux 7 Server (RPMs)',
+        'product': 'Red Hat Enterprise Linux Server',
+    },
+    'rhel7_server_extras': {
+        'id': 'rhel-7-server-extras-rpms',
+        'name': 'Red Hat Enterprise Linux 7 Server - Extras RPMs x86_64',
+        'releasever': '7',
+        'reposet': 'Red Hat Enterprise Linux 7 Server - Extras (RPMs)',
+        'product': 'Red Hat Enterprise Linux Server',
+    },
+    'rhel8_bos': {
+        'id': 'rhel-8-for-x86_64-baseos-rpms',
+        'name': f'Red Hat Enterprise Linux 8 for x86_64 - BaseOS RPMs {RHEL8_VER}',
+        'releasever': RHEL8_VER,
+        'reposet': 'Red Hat Enterprise Linux 8 for x86_64 - BaseOS (RPMs)',
+    },
+    'rhel8_aps': {
+        'id': 'rhel-8-for-x86_64-appstream-rpms',
+        'name': f'Red Hat Enterprise Linux 8 for x86_64 - AppStream RPMs {RHEL8_VER}',
+        'releasever': RHEL8_VER,
+        'reposet': 'Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)',
+    },
+    'rhel9_bos': {
+        'id': 'rhel-9-for-x86_64-baseos-rpms',
+        'name': f'Red Hat Enterprise Linux 9 for x86_64 - BaseOS RPMs {RHEL9_VER}',
+        'releasever': RHEL9_VER,
+        'reposet': 'Red Hat Enterprise Linux 9 for x86_64 - BaseOS (RPMs)',
+    },
+    'rhel9_aps': {
+        'id': 'rhel-9-for-x86_64-appstream-rpms',
+        'name': f'Red Hat Enterprise Linux 9 for x86_64 - AppStream RPMs {RHEL9_VER}',
+        'releasever': RHEL9_VER,
+        'reposet': 'Red Hat Enterprise Linux 9 for x86_64 - AppStream (RPMs)',
+    },
+}
+
+
+@pytest.fixture(scope='module')
+def module_stash(request):
+    """Module scoped stash for storing data between tests"""
+    # Please refer the documentation for more details on stash
+    # https://docs.pytest.org/en/latest/reference/reference.html#stash
+    request.node.stash[synced_repos] = {}
+    yield request.node.stash
+
+
+@pytest.fixture(scope='module')
+def module_leapp_lce(module_target_sat, module_sca_manifest_org):
+    return module_target_sat.api.LifecycleEnvironment(organization=module_sca_manifest_org).create()
+
+
+@pytest.fixture
+def function_leapp_cv(module_target_sat, module_sca_manifest_org, leapp_repos, module_leapp_lce):
+    function_leapp_cv = module_target_sat.api.ContentView(
+        organization=module_sca_manifest_org
+    ).create()
+    function_leapp_cv.repository = leapp_repos
+    function_leapp_cv = function_leapp_cv.update(['repository'])
+    function_leapp_cv.publish()
+    cvv = function_leapp_cv.read().version[0]
+    cvv.promote(data={'environment_ids': module_leapp_lce.id, 'force': True})
+    function_leapp_cv = function_leapp_cv.read()
+    return function_leapp_cv
+
+
+@pytest.fixture
+def function_leapp_ak(
+    module_target_sat,
+    function_leapp_cv,
+    module_leapp_lce,
+    module_sca_manifest_org,
+):
+    return module_target_sat.api.ActivationKey(
+        content_view=function_leapp_cv,
+        environment=module_leapp_lce,
+        organization=module_sca_manifest_org,
+    ).create()
+
+
+@pytest.fixture
+def leapp_repos(
+    default_architecture,
+    module_stash,
+    upgrade_path,
+    module_target_sat,
+    module_sca_manifest_org,
+):
+    """Enable and sync RHEL BaseOS, AppStream repositories"""
+    source = upgrade_path['source_version']
+    target = upgrade_path['target_version']
+    all_repos = []
+    for rh_repo_key in RHEL_REPOS.keys():
+        release_version = RHEL_REPOS[rh_repo_key]['releasever']
+        if release_version in str(source) or release_version in target:
+            prod = 'rhel' if 'rhel7' in rh_repo_key else rh_repo_key.split('_')[0]
+            if module_stash[synced_repos].get(rh_repo_key, None):
+                logger.info('Repo %s already synced, not syncing it', rh_repo_key)
+            else:
+                module_stash[synced_repos][rh_repo_key] = True
+                repo_id = module_target_sat.api_factory.enable_rhrepo_and_fetchid(
+                    basearch=default_architecture.name,
+                    org_id=module_sca_manifest_org.id,
+                    product=PRDS[prod],
+                    repo=RHEL_REPOS[rh_repo_key]['name'],
+                    reposet=RHEL_REPOS[rh_repo_key]['reposet'],
+                    releasever=release_version,
+                )
+                rh_repo = module_target_sat.api.Repository(id=repo_id).read()
+                all_repos.append(rh_repo)
+                rh_repo.sync(timeout=1800)
+    return all_repos
+
+
+@pytest.fixture
+def verify_target_repo_on_satellite(
+    module_target_sat,
+    function_leapp_cv,
+    module_sca_manifest_org,
+    module_leapp_lce,
+    upgrade_path,
+):
+    """Verify target rhel version repositories have been added in correct CV, LCE on Satellite"""
+    target_rhel_major_ver = upgrade_path['target_version'].split('.')[0]
+    cmd_out = module_target_sat.cli.Repository.list(
+        {
+            'search': f'content_label ~ rhel-{target_rhel_major_ver}',
+            'content-view-id': function_leapp_cv.id,
+            'organization-id': module_sca_manifest_org.id,
+            'lifecycle-environment-id': module_leapp_lce.id,
+        }
+    )
+    repo_names = [out['name'] for out in cmd_out]
+    if target_rhel_major_ver == '9':
+        assert RHEL_REPOS['rhel9_bos']['name'] in repo_names
+        assert RHEL_REPOS['rhel9_aps']['name'] in repo_names
+    else:
+        assert RHEL_REPOS['rhel8_bos']['name'] in repo_names
+        assert RHEL_REPOS['rhel8_aps']['name'] in repo_names
+
+
+@pytest.fixture
+def custom_leapp_host(upgrade_path, module_target_sat, module_sca_manifest_org, function_leapp_ak):
+    """Checkout content host and register with satellite"""
+    deploy_args = {}
+    deploy_args['deploy_rhel_version'] = upgrade_path['source_version']
+    with Broker(
+        workflow='deploy-rhel',
+        host_class=ContentHost,
+        deploy_rhel_version=upgrade_path['source_version'],
+        deploy_flavor=settings.flavors.default,
+    ) as chost:
+        result = chost.register(
+            module_sca_manifest_org, None, function_leapp_ak.name, module_target_sat
+        )
+        assert result.status == 0, f'Failed to register host: {result.stderr}'
+        yield chost
+
+
+@pytest.fixture
+def precondition_check_upgrade_and_install_leapp_tool(custom_leapp_host):
+    """Clean-up directory if in-place upgrade already performed,
+    set rhel release version, update system and install leapp-upgrade"""
+    source_rhel = custom_leapp_host.os_version.base_version
+    custom_leapp_host.run('rm -rf /root/tmp_leapp_py3')
+    custom_leapp_host.run('yum repolist')
+    custom_leapp_host.run(f'subscription-manager release --set {source_rhel}')
+    assert custom_leapp_host.run('yum update -y').status == 0
+    assert custom_leapp_host.run('yum install leapp-upgrade -y').status == 0
+    if custom_leapp_host.run('needs-restarting -r').status == 1:
+        custom_leapp_host.power_control(state='reboot', ensure=True)
+
+
+@pytest.mark.parametrize(
+    'upgrade_path',
+    [
+        # {'source_version': RHEL7_VER, 'target_version': RHEL8_VER},
+        {'source_version': RHEL8_VER, 'target_version': RHEL9_VER},
+    ],
+    ids=lambda upgrade_path: f'{upgrade_path["source_version"]}'
+    f'_to_{upgrade_path["target_version"]}',
+)
+def test_leapp_upgrade_rhel(
+    module_target_sat,
+    custom_leapp_host,
+    upgrade_path,
+    verify_target_repo_on_satellite,
+    precondition_check_upgrade_and_install_leapp_tool,
+):
+    """Test to upgrade RHEL host to next major RHEL release using leapp preupgrade and leapp upgrade
+    job templates
+
+    :id: 8eccc689-3bea-4182-84f3-c121e95d54c3
+
+    :Steps:
+        1. Import a subscription manifest and enable, sync source & target repositories
+        2. Create LCE, Create CV, add repositories to it, publish and promote CV, Create AK, etc.
+        3. Register content host with AK
+        4. Verify that target rhel repositories are enabled on Satellite
+        5. Update all packages, install leapp tool and fix inhibitors
+        6. Run Leapp Preupgrade and Leapp Upgrade job template
+
+    :expectedresults:
+        1. Update RHEL OS major version to another major version
+    """
+    # Fixing known inhibitors for source rhel version 8
+    if custom_leapp_host.os_version.major == 8:
+        # Inhibitor - Firewalld Configuration AllowZoneDrifting Is Unsupported
+        custom_leapp_host.run(
+            'sed -i "s/^AllowZoneDrifting=.*/AllowZoneDrifting=no/" /etc/firewalld/firewalld.conf'
+        )
+    # Run LEAPP-PREUPGRADE Job Template-
+    template_id = (
+        module_target_sat.api.JobTemplate()
+        .search(query={'search': 'name="Run preupgrade via Leapp"'})[0]
+        .id
+    )
+    job = module_target_sat.api.JobInvocation().run(
+        synchronous=False,
+        data={
+            'job_template_id': template_id,
+            'targeting_type': 'static_query',
+            'search_query': f'name = {custom_leapp_host.hostname}',
+        },
+    )
+    module_target_sat.wait_for_tasks(
+        f'resource_type = JobInvocation and resource_id = {job["id"]}', poll_timeout=1800
+    )
+    result = module_target_sat.api.JobInvocation(id=job['id']).read()
+    assert result.succeeded == 1
+
+    # Run LEAPP-UPGRADE Job Template-
+    template_id = (
+        module_target_sat.api.JobTemplate()
+        .search(query={'search': 'name="Run upgrade via Leapp"'})[0]
+        .id
+    )
+    job = module_target_sat.api.JobInvocation().run(
+        synchronous=False,
+        data={
+            'job_template_id': template_id,
+            'targeting_type': 'static_query',
+            'search_query': f'name = {custom_leapp_host.hostname}',
+            'inputs': {'Reboot': 'true'},
+        },
+    )
+    module_target_sat.wait_for_tasks(
+        f'resource_type = JobInvocation and resource_id = {job["id"]}', poll_timeout=1800
+    )
+    result = module_target_sat.api.JobInvocation(id=job['id']).read()
+    assert result.succeeded == 1
+    # Wait for the host to be rebooted and SSH daemon to be started.
+    custom_leapp_host.wait_for_connection()
+
+    custom_leapp_host.clean_cached_properties()
+    new_ver = str(custom_leapp_host.os_version)
+    assert new_ver == upgrade_path['target_version']
--- a/testimony.yaml
+++ b/testimony.yaml
@@ -68,7 +68,7 @@ CaseComponent:
     - katello-agent
     - katello-tracer
     - LDAP
-    - Leappintegration
+    - LeappIntegration
     - LifecycleEnvironments
     - LocalizationInternationalization
     - Logging


vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/cli/test_leapp.py -k test_upgrade_rhel8_to_rehl9

shweta83:
@vijaysawant Can you please update test filename to test_leapp_client.py just to avoid confusion as we test leapp for satellite as well.

tests/foreman/cli/test_leapp.py
```diff
@@ -0,0 +1,210 @@
...
+        'product': PRDS['rhel9'],
+        'distro': 'rhel9',
+        'key': 'rhel9_aps',
+    },
+}
+
+
+@pytest.mark.no_containers
+@pytest.mark.rhel_ver_match('8')
+def test_upgrade_rhel8_to_rehl9(
```
shweta83, line 80:
```suggestion
def test_upgrade_rhel8_to_rhel9(
```

tests/foreman/cli/test_leapp.py
```diff
@@ -0,0 +1,210 @@
...
+        5. Update all packages, install leapp tool and fix inhibitors
+        6. Run Leapp Preupgrade and Leapp Upgrade job template
+    :ExpectedResult:
+        1. Update RHEL OS major version to another major version
+    """
+    # 1. Import a subscription manifest
+    target_sat.upload_manifest(default_org)
+    # Enable rhel8/rhel9 bos, aps repository and add in content view
+    lc_env = target_sat.api.LifecycleEnvironment(organization=default_org).create()
+    c_view = target_sat.api.ContentView(organization=default_org).create()
```
shweta83, line 102: First 3 steps can be part of setup fixture you can create and move it there as it will be common for all the scenarios

tests/foreman/cli/test_leapp.py
```diff
@@ -0,0 +1,210 @@
...
+    )
+    target_sat.wait_for_tasks(
+        f'resource_type = JobInvocation and resource_id = {job["id"]}', poll_timeout=1800
+    )
+    result = target_sat.api.JobInvocation(id=job['id']).read()
+    assert result.succeeded == 1
+    # Resume the rhel_contenthsot with ensure True to ping the virtual machine
+    rhel_contenthost.power_control(state=VmState.RUNNING, ensure=True)
+    rhel_new_ver = rhel_contenthost.run("cat /etc/redhat-release")
+    assert rhel_old_ver != rhel_new_ver
```
shweta83, line 210: Please update the assertion with the particular rhel version it should be updated to.

tests/foreman/cli/test_leapp.py
```diff
@@ -0,0 +1,210 @@
...
+    Job templates
+    :Steps:
+        1. Import a subscription manifest and enable, sync source & target repositories
+        2. Create CV, add repositories, create lce, ak, etc.
+        3. Register contenthost using ak
+        4. Varify target rhel repositories are enable on Satellite
+        5. Update all packages, install leapp tool and fix inhibitors
+        6. Run Leapp Preupgrade and Leapp Upgrade job template
+    :ExpectedResult:
+        1. Update RHEL OS major version to another major version
```
shubhamsg199, line 96:
```suggestion
    Test to upgrade RHEL host to next major RHEL Realse with Leapp Preupgrade and Leapp Upgrade
    Job templates
    
    :Steps:
        1. Import a subscription manifest and enable, sync source & target repositories
        2. Create CV, add repositories, create lce, ak, etc.
        3. Register contenthost using ak
        4. Varify target rhel repositories are enable on Satellite
        5. Update all packages, install leapp tool and fix inhibitors
        6. Run Leapp Preupgrade and Leapp Upgrade job template
        
    :ExpectedResult:
        1. Update RHEL OS major version to another major version
```

tests/foreman/cli/test_leapp.py
```diff
@@ -0,0 +1,210 @@
...
+    target_sat.upload_manifest(default_org)
+    # Enable rhel8/rhel9 bos, aps repository and add in content view
+    lc_env = target_sat.api.LifecycleEnvironment(organization=default_org).create()
+    c_view = target_sat.api.ContentView(organization=default_org).create()
+    all_repos = []
+    for rh_repo_key in ['rhel8_8_bos', 'rhel8_8_aps', 'rhel9_2_bos', 'rhel9_2_aps']:
+        if 'rhel8' in rh_repo_key:
+            prod = PRDS['rhel8']
+        else:
+            prod = PRDS['rhel9']
```
shubhamsg199, line 108:
```suggestion
        prod = PRDS['rhel8'] if 'rhel8' in rh_repo_key else PRDS['rhel9']
```

tests/foreman/cli/test_leapp.py
```diff
@@ -0,0 +1,210 @@
...
+    # 3. Register Host
+    rhel_contenthost.install_katello_ca(target_sat)
+    rhel_contenthost.register_contenthost(org=default_org.label, activation_key=ak.name)
+    rhel_contenthost.add_rex_key(satellite=target_sat)
+    assert rhel_contenthost.subscribed
+    # Verify target rhel version repositories has enabled on Satellite Server
+    cmd_out = target_sat.execute(
+        f"hammer repository list --search 'content_label ~ rhel-9' --content-view {c_view.name} "
+        f"--organization '{default_org.name}' --lifecycle-environment '{lc_env.name}'"
+    )
```
shubhamsg199, line 143: You can use hammer cli here.

tests/foreman/cli/test_leapp.py
shubhamsg199: Please add the test id too

tests/foreman/cli/test_leapp.py
```diff
@@ -0,0 +1,210 @@
...
+        'name': 'Red Hat Enterprise Linux 9 for x86_64 - AppStream RPMs 9.2',
+        'releasever': '9.2',
+        'basearch': 'x86_64',
+        'version': '9',
+        'reposet': CUSTOM_REPOSET['rhel9_2_aps'],
+        'product': PRDS['rhel9'],
+        'distro': 'rhel9',
+        'key': 'rhel9_aps',
+    },
+}
```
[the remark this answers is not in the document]
vijaysawant: Inside constants.py, REPOSET, REPOS are defined only for major version (for example rhel7, rhel8, rhel9, etc) with that release verion only 7.0, 8.0, & 9.0 repos would get enable.
So for minor versions like 7.9, 8.8, 9.2 I needed to re-defined these and these constants are only be using in Leapp sceanrios. So keeping in the same test module as per suggestion from other rocket team member.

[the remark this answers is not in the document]
shubhamsg199: @vijaysawant I think content repos always grab the latest rhel version so with release versions 7, 8, and 9, It will grab 7.9, 8.8, 9.2 versions respectively.

tests/foreman/cli/test_leapp.py
vijaysawant: @shweta83  What different assertions statement can be used instead of `assert old_ver != new_ver` condition?
ex-
`assert "9.2" in str(rhel_new_ver)`
`assert "Red Hat Enterprise Linux release 9.2 (Ootpa)" == str(rhel_new_ver)`

tests/foreman/cli/test_leapp.py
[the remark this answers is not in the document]
vijaysawant: - Talking about the latest versions (i.e 7.9, 8.8 & 9.2 ) - I was also expecting that the latest repos would get enabled and synch if we use robottelo > constants like REPOSET['rhel8_bos'], REPOSET['rhel8_aps'], etc.
- But only repository for major versions got enabled, which ended with reposet error, hence to be more specific with major.minor version we are enabling specific repos explicitly.

vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/cli/test_leapp.py -k test_leapp_client.py

tests/foreman/cli/test_leapp.py
vijaysawant: thanks, added.

tests/foreman/cli/test_leapp.py
vijaysawant: added new fixture

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,227 @@
...
+
+from robottelo.constants import PRDS
+
+CUSTOM_REPOSET = {
+    'rhel8_8_bos': 'Red Hat Enterprise Linux 8 for x86_64 - BaseOS (RPMs)',
+    'rhel8_8_aps': 'Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)',
+    'rhel9_2_bos': 'Red Hat Enterprise Linux 9 for x86_64 - BaseOS (RPMs)',
+    'rhel9_2_aps': 'Red Hat Enterprise Linux 9 for x86_64 - AppStream (RPMs)',
+}
+CUSTOM_REPOS = {
```
vijaysawant, line 30: - remove the extra key:value which are not using / paasing to `api_factory.enable_rhrepo_and_fetchid`
- recheck values for 'distro', 'key' & 'basearch'

vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/cli/test_leapp_client.py -k test_upgrade_rhel8_to_rehl9

tests/foreman/cli/test_leapp.py
```diff
@@ -0,0 +1,210 @@
...
+        },
+    )
+    target_sat.wait_for_tasks(
+        f'resource_type = JobInvocation and resource_id = {job["id"]}', poll_timeout=1800
+    )
+    result = target_sat.api.JobInvocation(id=job['id']).read()
+    assert result.succeeded == 1
+    # Resume the rhel_contenthsot with ensure True to ping the virtual machine
+    rhel_contenthost.power_control(state=VmState.RUNNING, ensure=True)
+    rhel_new_ver = rhel_contenthost.run("cat /etc/redhat-release")
```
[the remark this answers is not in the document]
vijaysawant: `rhel_contenthost.os_version` gives an older rhel version and this leads to an assertion fail, CI execution was also failed due to this

vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/cli/test_leapp_client.py -k test_upgrade_rhel8_to_rehl9

vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/cli/test_leapp_client.py -k test_upgrade_rhel8_to_rehl9

vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/cli/test_leapp_client.py -k test_upgrade_rhel8_to_rehl9

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,203 @@
...
+    )
+    target_sat.wait_for_tasks(
+        f'resource_type = JobInvocation and resource_id = {job["id"]}', poll_timeout=1800
+    )
+    result = target_sat.api.JobInvocation(id=job['id']).read()
+    assert result.succeeded == 1
+    # Resume the rhel_contenthsot with ensure True to ping the virtual machine
+    rhel_contenthost.power_control(state=VmState.RUNNING, ensure=True)
+    rhel_new_ver = rhel_contenthost.run('cat /etc/redhat-release')
+    assert rhel_old_ver != rhel_new_ver
```
ogajduse, line 202: Let me evaluate it:
`Version('8.8') != "Red Hat Enterprise Linux release 9.2 (Plow)"`
You are comparing a Version object with the contents of /etc/redhat-release (a string), which will always be evaluated to `True`.

tests/foreman/cli/test_leapp_client.py
ogajduse, line 1: Can you please break down the code before the pre-upgrade tasks into reusable fixtures?

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,203 @@
...
+        'releasever': '9.2',
+    },
+}
+
+
+@pytest.fixture(scope="module")
+def setup_env(module_target_sat, module_org):
+    """Creating essential things and returning in form of dictiory for use"""
+    # 1. Import a subscription manifest
+    module_target_sat.upload_manifest(module_org)
```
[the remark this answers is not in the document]
synkd: @ogajduse Yes, using the `module_sca_manifest_org` fixture is definitely how I would approach this.

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,203 @@
...
+        'contentview': c_view,
+    }
+
+
+@pytest.mark.no_containers
+@pytest.mark.rhel_ver_match('8')
+def test_upgrade_rhel8_to_rehl9(
+    setup_env,
+    rhel_contenthost,
+):
```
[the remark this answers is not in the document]
synkd: I was initially confused by this suggestion before determining that upgrading from RHEL 8.9 to RHEL 9.3 will only be a supported upgrade path for six months per [1]. Since the supported upgrade path matrix for LEAPP is not exactly intuitive, I think it would be worth linking to one of the documentation sources for this (either [1] or [2]) in the test or module metadata.

[1] https://access.redhat.com/articles/4263361
[2] https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html-single/upgrading_from_rhel_8_to_rhel_9/index#con_supported-upgrade-paths_upgrading-from-rhel-8-to-rhel-9

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,203 @@
...
+    # Fixing inhibitors - download data files to avoid inhibitors
+    rhel_contenthost.run(
+        'curl -k "https://gitlab.cee.redhat.com/oamg/leapp-data/-/raw/stage/data/'
+        '{repomap.json,pes-events.json,device_driver_deprecation_data.json}"'
+        ' -o "/etc/leapp/files/#1"'
+    )
+    rhel_contenthost.run(
+        'sed -i "s/^AllowZoneDrifting=.*/AllowZoneDrifting=no/" /etc/firewalld/firewalld.conf'
+    )
+    # 6.1 Run LEAPP-PREUPGRADE Job Template-
```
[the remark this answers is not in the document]
synkd: The numbers refer to the `:Steps:` in the test metadata. @vijaysawant  If you include the steps as comments in the test body, I would suggest maintaining a 1:1 correspondence between those comments and the test metadata. In other words, if the comments break down step 6 into multiple parts (6.1 and 6.2),. those parts should be included in the test metadata as well.

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,203 @@
...
+            'search_query': f'name = {rhel_contenthost.hostname}',
+            'inputs': {'Reboot': 'true'},
+        },
+    )
+    target_sat.wait_for_tasks(
+        f'resource_type = JobInvocation and resource_id = {job["id"]}', poll_timeout=1800
+    )
+    result = target_sat.api.JobInvocation(id=job['id']).read()
+    assert result.succeeded == 1
+    # Resume the rhel_contenthsot with ensure True to ping the virtual machine
```
synkd, line 199:
```suggestion
    # Resume the rhel_contenthost with ensure True to ping the virtual machine
```

synkd (changes requested):

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,203 @@
...
+    rhel_contenthost.run("yum repolist")
+
+    rhel_contenthost.run("subscription-manager release --set 8.8")
+    result = rhel_contenthost.run("dnf update -y")
+    assert result.status == 0
+    rhel_contenthost.run("yum install leapp-upgrade -y")
+
+    # Fixing inhibitors - download data files to avoid inhibitors
+    rhel_contenthost.run(
+        'curl -k "https://gitlab.cee.redhat.com/oamg/leapp-data/-/raw/stage/data/'
```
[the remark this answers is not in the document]
vijaysawant: I am yet to test that without downloading leapp-data (metadata) is this working as expected, as per the latest hotfix in leapp tool package these 3 files should download implicitly.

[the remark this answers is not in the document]
vijaysawant: line 153,154,etc will be remove in future.

tests/foreman/cli/test_leapp_client.py
vijaysawant: yea that's correct, thanks 
I will use same method to get rhel_old_ver & rhel_new_ver.

vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/cli/test_leapp_client.py -k test_upgrade_rhel8_to_rehl9

tests/foreman/cli/test_leapp_client.py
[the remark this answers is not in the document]
ogajduse: Let me keep this conversation open until you remove the URL from the test.

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,203 @@
...
+            'inputs': {'Reboot': 'true'},
+        },
+    )
+    target_sat.wait_for_tasks(
+        f'resource_type = JobInvocation and resource_id = {job["id"]}', poll_timeout=1800
+    )
+    result = target_sat.api.JobInvocation(id=job['id']).read()
+    assert result.succeeded == 1
+    # Resume the rhel_contenthsot with ensure True to ping the virtual machine
+    rhel_contenthost.power_control(state=VmState.RUNNING, ensure=True)
```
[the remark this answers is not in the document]
vijaysawant: > Besides the typos in line 199, why do we need to `power_control` the VM? I see that reboot is set to true: `'inputs': {'Reboot': 'true'},`, therefore the VM should be running after the `Run upgrade via Leapp` is finished. Waking up the host if the assumption is that the host is up, does not seem to be right.

Host gets disconnected after reboot, in order to get back ssh connection using 'power_control' here.

> Could you create a new `ContentHost` method that would make use of the `connect` method and use it here? The method body will be similar to the contents of the following code snippet

I think it's not required to create a new `wait_for` method, because at line [#L196](https://github.com/SatelliteQE/robottelo/blob/845c03cbf9ac69ee492fb6458e00e5fb09c4e305/tests/foreman/cli/test_leapp_client.py#L196) script waiting till `leapp upgrade` & `reboot` completed, and right after that script loses ssh connection with host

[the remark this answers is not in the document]
vijaysawant: When I check at very first time, I observed once **LEAPP-UPGRADE Job Template** runs successfully and debugger moves to next line, rhel host was up & running and the script didn't have ssh connection with the host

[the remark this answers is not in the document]
ogajduse: I get what your point is, but I believe that you just do not understand what I am asking you for.

You are correct, robottelo looses connection to the SSH daemon after the Leapp job reboots it. That is why you use `power_control`. You are taking an advantage of the fact that `power_control` has the `wait_for ContentHost.connect` logic already implemented. The reason why you should not use `power_control` is that it manipulates with the VM state. The VM should be running, therefore there is no need to ensure that it is really running, the Leapp job should reboot the host, not the test, nor the test should ensure the VM state.
Therefore the only thing that you need to do is to wait until `rhel_contenthost.connect()` succeeds.
And that is why I suggest to create a helper method that would encapsulate the wait logic.

vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/cli/test_leapp_client.py -k test_upgrade_rhel8_to_rhel9

tests/foreman/cli/test_leapp_client.py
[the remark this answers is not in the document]
vijaysawant: Thank you for the suggestion, may I know how I can include the above-supported path [2] in a test?
@synkd

tests/foreman/cli/test_leapp_client.py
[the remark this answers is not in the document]
vijaysawant: done

tests/foreman/cli/test_leapp_client.py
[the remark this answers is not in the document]
vijaysawant: It's working without downloading metadata, as the latest hotfix version of Leapp-tool is being installed on the host.
PRT passed  :smile:

tests/foreman/cli/test_leapp_client.py
vijaysawant: sure, instead of fixture I would say functions, because there are set of small steps needs to be executed sequencially on host or on satellite prior to run Leapp templates

ogajduse: You can achieve sequential execution with fixtures too.

```python
@pytest.fixture
def step_one():
    return 1
    
@pytest.fixture
def step_two(step_one):
    return step_one + 1

@pytest.fixture
def step_three(step_two):
    return step_two + 1

def test_assert_equal_three(step_three):
    assert step_three == 3
```

tests/foreman/cli/test_leapp_client.py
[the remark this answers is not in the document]
ogajduse: Nice, let's close this one then.

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,220 @@
...
+    assert result.status == 0
+    custom_host.run("dnf install leapp-upgrade -y")
+
+    # Fixing inhibitors - download data files to avoid inhibitors
+    custom_host.run(
+        'sed -i "s/^AllowZoneDrifting=.*/AllowZoneDrifting=no/" /etc/firewalld/firewalld.conf'
+    )
+    if custom_host.run('needs-restarting -r').status == 1:
+        custom_host.power_control(state='reboot', ensure=True)
+        custom_host.power_control(state=VmState.RUNNING, ensure=True)
```
ogajduse, line 168: Can you explain why you need to run power_control twice?

Generally, if you reboot the system, it should boot up again, therefore you should not need to ensure that it is really running.

ogajduse (changes requested):

tests/foreman/cli/test_leapp_client.py
vijaysawant: That's a good example for how to use a sequential fixture, I will try to map this with the current scenario, however a current set of functions are not depending on each other and not necessary to return any value, just executing commands on a machine.
Let me open updated code for review once again.

tests/foreman/cli/test_leapp_client.py
vijaysawant: Okay, I think, right after reboot finish with `power_control` at #167, the script gets back ssh connection and no need to use `power_control` again because `wait_for` already inplace inside.
@ogajduse Am I assuming correctly?

tests/foreman/cli/test_leapp_client.py
vijaysawant: Cleared.

tests/foreman/cli/test_leapp_client.py
vijaysawant: Cleared.

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,263 @@
...
+:Upstream: No
+"""
+import pytest
+from wait_for import wait_for
+from wrapanapi.entities.vm import VmState
+
+from robottelo.constants import PRDS
+
+CUSTOM_REPOSET = {
+    'rhel8_8_bos': 'Red Hat Enterprise Linux 8 for x86_64 - BaseOS (RPMs)',
```
lpramuk, line 26: Please merge this dict into CUSTOM_REPOS dict so that you end up with:

```
CUSTOM_REPOS = {
  'rhel8_8_bos': {
     'id':
     'name':
    'reposet':
    'releasever':
  }
...
}
```

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,263 @@
...
+    'rhel8_8_aps': 'Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)',
+    'rhel9_2_bos': 'Red Hat Enterprise Linux 9 for x86_64 - BaseOS (RPMs)',
+    'rhel9_2_aps': 'Red Hat Enterprise Linux 9 for x86_64 - AppStream (RPMs)',
+}
+CUSTOM_REPOS = {
```
lpramuk, line 31: `CUSTOM_REPOS` ?
Wouldn't be `RHEL_REPOS` more appropriate name for this ?

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,263 @@
...
+    # Enable rhel8/rhel9 bos, aps repository and add in content view
+    all_repos = []
+    for rh_repo_key in CUSTOM_REPOS.keys():
+        prod = rh_repo_key.split('_')[0]
+        repo_id = target_sat.api_factory.enable_rhrepo_and_fetchid(
+            basearch=custom_host.arch,
+            org_id=organization.id,
+            product=PRDS[prod],
+            repo=CUSTOM_REPOS[rh_repo_key]['name'],
+            reposet=CUSTOM_REPOSET[rh_repo_key],
```
lpramuk, line 178: with merged `CUSTOM_REPOSET` into `CUSTOM_REPOS` you need to write here 
```
reposet=CUSTOM_REPOS[rh_repo_key]['reposet'],
```

lpramuk (changes requested):
Please merge two data dicts into one

vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/cli/test_leapp_client.py -k test_upgrade_rhel8_to_rhel9

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,262 @@
...
+        content_view=content_view,
+        environment=lifecycle_env,
+        organization=organization,
+    ).create()
+
+
+def register_host_with_satellite(satellite, custom_host, organization, activation_key):
+    """Register content host with satellite"""
+    custom_host.install_katello_ca(satellite)
+    custom_host.register_contenthost(org=organization.label, activation_key=activation_key.name)
```
vijaysawant, line 66: - Use `custom_host.register` instead of `custom_host.register_contenthost` 
- remove `custom_host.install_katello_ca` and `custom_host.add_rex_key`

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,203 @@
...
+        },
+    )
+    target_sat.wait_for_tasks(
+        f'resource_type = JobInvocation and resource_id = {job["id"]}', poll_timeout=1800
+    )
+    result = target_sat.api.JobInvocation(id=job['id']).read()
+    assert result.succeeded == 1
+    # Resume the rhel_contenthsot with ensure True to ping the virtual machine
+    rhel_contenthost.power_control(state=VmState.RUNNING, ensure=True)
+    rhel_new_ver = rhel_contenthost.run('cat /etc/redhat-release')
```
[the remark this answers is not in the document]
vijaysawant: - I tried to implement to [clean cached properties](https://github.com/SatelliteQE/robottelo/pull/11553/files#diff-b93d2c93019656e769060a9f7bd74d6242da37d3b9c571f9ae41bc6224d78dd0R293) in one of older commit, I able to run in debugging mode, but in CI it was returning **KeyError**.
- Looking at the time spend on this PR I have decided to implement this functionality in upcoming quarter for that I have created an [issue](https://github.com/SatelliteQE/robottelo/issues/12115).
- This is the first Leapp upgrade test case and needs to run in CI, so requesting all reviewer to provide valuable comments if those are major or impacting other parts of framework.
- I agree there is scope to make things superior but those would be considered in upcoming Leapp upgrade PR.

tests/foreman/cli/test_leapp_client.py
vijaysawant: - Nice suggestion Lukash, I will take this in upcoming PR.

tests/foreman/cli/test_leapp_client.py
vijaysawant: Thank you for the suggestion and related changes,  I will take this in upcoming PR.

vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/cli/test_leapp_client.py -k test_upgrade_rhel8_to_rhel9

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,259 @@
...
+@pytest.mark.no_containers
+@pytest.mark.parametrize(
+    'custom_host',
+    [
+        {'deploy_rhel_version': '8.8'},
+    ],
+    ids=['RHEL8.8'],
+    indirect=True,
+)
+@pytest.mark.parametrize('target_rhel', ['rhel-9'])
```
Gauravtalreja1, line 139: As you'll be upgrading it rhel9, and skip level upgrade isn't possible, so why parametrizing it? :)

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,259 @@
...
+):
+    """Test to upgrade RHEL host to next major RHEL Realse with Leapp Preupgrade and Leapp Upgrade
+    Job templates
+
+    :id: 7add9b32-a13a-4e65-973c-cd64326b332a
+
+    :Steps:
+        1. Import a subscription manifest and enable, sync source & target repositories
+        2. Create CV, add repositories, create lce, ak, etc.
+        3. Register contenthost using ak
```
Gauravtalreja1, line 153:
```suggestion
        2. Create LCE, Create CV, add repositories to it, publish and promote CV, Create AK, etc.
        3. Register content host with AK
```

Gauravtalreja1 (commented):
Great job @vijaysawant, adding few questions and suggestions

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+def create_activation_key(satellite, content_view, lifecycle_env, organization):
+    """Create activation key uinsg specific entities"""
+    return satellite.api.ActivationKey(
+        content_view=content_view,
+        environment=lifecycle_env,
+        organization=organization,
+    ).create()
+
```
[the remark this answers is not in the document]
vijaysawant: The `create_activation_key` will be reused in other test module (ex. 7 to 8 upgrade) as well

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,259 @@
...
+def precondition_check_upgrade_and_install_leapp_tool(custom_host, source_rhel):
+    """Clean-up directory, set rhel release version, update system and install leapp tool"""
+    # Remove directory if in-place upgrade already performed from RHEL7 to RHEL8
+    custom_host.run("rm -rf /root/tmp_leapp_py3")
+    custom_host.run("dnf clean all")
+    custom_host.run("dnf repolist")
+    custom_host.run(f"subscription-manager release --set {source_rhel}")
+    assert custom_host.run("dnf update -y").status == 0
+    assert custom_host.run("dnf install leapp-upgrade -y").status == 0
```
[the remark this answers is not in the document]
vijaysawant: Is this comment applicable throughout the test file or these 3 lines only?

tests/foreman/cli/test_leapp_client.py
vijaysawant: - The earlier plan was to make the same test case parameterize so both upgrade paths would work, later on, realizes that it would be confusing & decided to make separate test.
- Also `target_rhel` is indicating major version not a actual OS version, it doesn't make any sense to use here in test
- So I will remove this marker and use plane text string wherever requires.

tests/foreman/cli/test_leapp_client.py
lpramuk: @vijaysawant It should be this PR if you want to receive ACKs

vijaysawant: - Not sure about **merging**, we have only one upgrade path implemented which is 8.8 -> 9.2
- I am planning to keep both dictionaries, `CUSTOM_REPOSET` and `CUSOTM_REPOS` separate and these 2 dictionaries will be updated by adding more data related to another upgrade path like 7.9 to 8.9, etc in future.
- Pros: easy to understand major & minor versions which we are using by it's keys
- Cons: Few syntactical changes would be required in for loop between lines 170 to 179 and requires additional conditional statement base upon **source rhel os version**
- Hence I am keeping these 2 dictionaries separate as of now, but Merging both dictionaries can consider once updated.
- Talking about **moving** into the constants, then we can move it, but need to keep them separately because here we are using a specific reponame & release version (`major.minor`), I don't see any other constants which are specific to minor OS version (apart from kickstart)

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,259 @@
...
+
+def verify_target_repo_on_satellite(
+    satellite, content_view, organization, lifecycle_env, target_rhel
+):
+    """Verify target rhel version repositories has enabled on Satellite Server"""
+    cmd_out = satellite.execute(
+        f"hammer repository list --search 'content_label ~ {target_rhel}' "
+        f"--content-view {content_view.name} --organization '{organization.name}' "
+        f"--lifecycle-environment '{lifecycle_env.name}'"
+    )
```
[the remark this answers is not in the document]
vijaysawant: As per rhel upgrade documentation, command has been used

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,259 @@
...
+        assert ("AppStream RPMs 9" in cmd_out.stdout) and ("BaseOS RPMs 9" in cmd_out.stdout)
+    else:
+        # placeholder for target_rhel - rhel-8
+        pass
+
+
+def precondition_check_upgrade_and_install_leapp_tool(custom_host, source_rhel):
+    """Clean-up directory, set rhel release version, update system and install leapp tool"""
+    # Remove directory if in-place upgrade already performed from RHEL7 to RHEL8
+    custom_host.run("rm -rf /root/tmp_leapp_py3")
```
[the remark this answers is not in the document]
vijaysawant: this directory would be present if in-place upgrade already performed

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,259 @@
...
+    custom_host.run("dnf clean all")
+    custom_host.run("dnf repolist")
```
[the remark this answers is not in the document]
vijaysawant: as part of preconditions and checks, some cleanup is necessary if we are performing a fresh rhel system or previously upgraded system

tests/foreman/cli/test_leapp_client.py
vijaysawant: I tried the way suggested, but unfortunately getting an error

> stderr:
    (213, b'Error: There are no enabled repositories in "/etc/yum.repos.d", "/etc/yum/repos.d", "/etc/distro.repos.d".\nCouldn\'t find activation key \'robottelo.hosts.DecClass(auto_attach=True\' (HTTP error code 404: Not Found)\n')
    status: 1

I think we can take this change in upcoming PR's, will that be ok @Gauravtalreja1

tests/foreman/cli/test_leapp_client.py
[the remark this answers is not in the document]
synkd: @vijaysawant That's a good question, but it's one that I don't have a definitive answer to. We don't have an established standard for this sort of thing in the Robottelo or Testimony documentation. My first thought is to include the link in the `:ExpectedResults:` section of the test function docstring with a note that the expected successful RHEL versions can be found in that documentation link. I'm open to other suggestions, though.

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,259 @@
+"""Test for LEAPP cli
+
+:Requirement: leapp
+
+:CaseLevel: Acceptance
+
+:CaseComponent: Leappintegration
```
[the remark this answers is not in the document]
vijaysawant: As per discussion with @jyejare test data should be aligned with testimony data, no need to change for now.

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,257 @@
...
+
+    # 3. Register Host
+    register_host_with_satellite(target_sat, custom_host, organization, ak)
+
+    # 4. Verify target rhel version repositories has enabled on Satellite Server
+    verify_target_repo_on_satellite(target_sat, cv, organization, lc_env, target_rhel='rhel-9')
+
+    # 5. Update all packages and install Leapp utility
+    # Preupgrade conditions and check
+    rhel_old_ver = custom_host.run('cat /etc/redhat-release')
```
ogajduse, line 200: Please use `custom_host.os_version` to fetch the system version.
With #12144 you can clean the cached properties and use `os_version` to fetch the OS version after the upgrade.

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,257 @@
...
+        wait_for(
+            custom_host.connect,
+            fail_condition=lambda res: res is not None,
+            handle_exception=True,
+            raise_original=True,
+            timeout=180,
+            delay=1,
+        )
+    except ConnectionRefusedError:
+        raise ConnectionRefusedError('Timed out waiting for SSH daemon to start on the host')
```
ogajduse, line 254: With the patch from #12144, you can use a one-line function to do what you are doing here on 10 lines. You could use the `wait_for_connection` method.

tests/foreman/cli/test_leapp_client.py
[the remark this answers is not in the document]
ogajduse: #12144 should address it.

tests/foreman/cli/test_leapp_client.py
ogajduse: I believe that there is still room for improvement and newly built tests should use fixtures as much as possible. Please check #12018 and get some inspiration there. That PR is a very nice example of how a test fixture stack should be built.

tests/foreman/cli/test_leapp_client.py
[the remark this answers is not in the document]
jyejare: @Gauravtalreja1 @vijaysawant I have changed the Id for this component in Polarion to `LeappIntegration` as this is the 1st test for this component and changing this now(before PR merge) in prod polarion wont affect any other tests.

@vijaysawant Please change token in testimony.yaml and test as suggested to use capital `I`.

ogajduse (changes requested):

ogajduse:
I would like to see us having stuff in fixtures rather than in functions. Please take a look at the following patch where I remade your functions to fixtures and used them in a test.
https://github.com/ogajduse/robottelo/commit/533cdbc41285187e5ac05c27d4c513641867951c

I would really like to see new tests using fixtures instead of functions.-

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,276 @@
...
+        'reposet': 'Red Hat Enterprise Linux 7 Server (RPMs)',
+        'product': 'Red Hat Enterprise Linux Server',
+    },
+    'rhel7_server_extras': {
+        'id': 'rhel-7-server-extras-rpms',
+        'name': 'Red Hat Enterprise Linux 7 Server - Extras RPMs x86_64',
+        'releasever': '7',
+        'reposet': 'Red Hat Enterprise Linux 7 Server - Extras (RPMs)',
+        'product': 'Red Hat Enterprise Linux Server',
+    },
```
vijaysawant, line 39: **Note:** I added `rhel7` repo data for testing purposes and this will be used in another upgrade path in future, so the review is not needed for these lines ;)

vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/cli/test_leapp_client.py -k test_upgrade_rhel8_to_rhel9

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,298 @@
...
+    'rhel8_8_bos': {
```
lpramuk, line 42: This is moving target. 
Once 8.10/9.4 is released it is going to fail.
We will be changing this from time to time.
So please create a constansts in this module:
```
RHEL7_VER = '7.9'
RHEL8_VER = '8.8'
RHEL9_VER = '9.2'
```
then by changing single constant you accommodate test for future versions

Ofc get rid of minor RHEL versions out of keys in `RHEL_REPOS` dictionary (it is useless there):

`rhel8_8_bos:`  -> `rhel8_bos:`
```
'rhel8_8_bos': {
    ...
    'name': 'Red Hat Enterprise Linux 8 for x86_64 - BaseOS RPMs {RHEL8_VER}',
    'releasever': RHEL8_VER
'rhel8_aps': {
    ...
}

**Review by lpramuk — CHANGES_REQUESTED** [review]
Please create a constants in this moddule for RHEL versions.
RHEL versions are going to be achanging.  9.2 -> 9.4 -> 9.6

File: tests/foreman/cli/test_leapp_client.py (hunk 34 above)
Comment (ogajduse) [reply tests/foreman/cli/test_leapp_client.py:42]: I believe that Lukas means something like this:
https://github.com/ogajduse/robottelo/commit/d937ec7141a734885befb4a2805b3709e3cb653c

**Comment by vijaysawant** [comment]
trigger: test-robottelo
pytest: tests/foreman/cli/test_leapp_client.py -k test_leapp_upgrade_rhel

File: tests/foreman/cli/test_leapp_client.py [hunk 35]
```diff
@@ -0,0 +1,353 @@
...
+        wait_for(
+            custom_leapp_host.connect,
+            fail_condition=lambda res: res is not None,
+            handle_exception=True,
+            raise_original=True,
+            timeout=180,
+            delay=1,
+        )
+    except ConnectionRefusedError:
+        raise ConnectionRefusedError('Timed out waiting for SSH daemon to start on the host')
```
ogajduse, line 349: Please replace it with 
```suggestion
    custom_leapp_host.wait_for_connection()
```

ogajduse (changes requested):
Almost an ACK. :stuck_out_tongue_winking_eye: 
I have left some blocking and non-blocking reviews this time.
Nice work, @vijaysawant :heart:

ogajduse (approved):
Agreed with @vijaysawant to get this PR merged ASAP and address my comments in a follow-up PR.

vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/cli/test_leapp_client.py -k test_leapp_upgrade_rhel

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,353 @@
...
+):
+    logger.info(
+        'Creating Leapp Activation Key for RHEL %s -> %s',
+        upgrade_path['source_version'],
+        upgrade_path['target_version'],
+    )
+    # need to publish and promote cv on lce to get content available on client host
+    function_leapp_cv.publish()
+    cvv = function_leapp_cv.read().version[0]
+    cvv.promote(data={'environment_ids': module_leapp_lce.id, 'force': True})
```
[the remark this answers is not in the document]
vijaysawant: It is required, otherwise could be ended with `Error: Cannot promote environment out of sequence.`

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,375 @@
...
+    module_target_sat.wait_for_tasks(
+        f'resource_type = JobInvocation and resource_id = {job["id"]}', poll_timeout=1800
+    )
+    result = module_target_sat.api.JobInvocation(id=job['id']).read()
+    assert result.succeeded == 1
+    # In case of Upgrade Path A - Rehl-7 to Rhel-8
+    # Inhibitors should fix after running LEAPP-PREUPGRADE Job Template
+    if upgrade_path['source_version'].split('.')[0] == '7':
+        logger.info('Fixing inhibitory for upgrade path : %s', upgrade_path)
+        # 1. Leapp detected loaded kernel drivers which have been removed in RHEL 8
```
ogajduse, line 339: How can you claim that leapp detected some kernel drivers if you do not check for it?

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,375 @@
...
+        custom_leapp_host.run('rmmod floppy')
+        custom_leapp_host.run('rmmod pata_acpi')
+        # 2. Missing required answers in the answer file
```
ogajduse, line 342: Does `1.` and `2.` refer to any documentation? Why are the comments numbered?

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,375 @@
...
+    cvv.promote(data={'environment_ids': module_leapp_lce.id, 'force': True})
+    function_leapp_cv = function_leapp_cv.read()
+    ak = module_target_sat.api.ActivationKey(
+        content_view=function_leapp_cv,
+        environment=module_leapp_lce,
+        organization=module_sca_manifest_org,
+    ).create()
+    # In case of 'rhel7_server_extras' repos, need to enabled (overridden) repository
+    # to get content available on client host
+    if upgrade_path['source_version'].split('.')[0] == '7':
```
ogajduse, line 120:
```suggestion
    if upgrade_path['source_version'] == RHEL7_VER:
```

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,375 @@
...
+    cvv = function_leapp_cv.read().version[0]
+    cvv.promote(data={'environment_ids': module_leapp_lce.id, 'force': True})
+    function_leapp_cv = function_leapp_cv.read()
+    ak = module_target_sat.api.ActivationKey(
+        content_view=function_leapp_cv,
+        environment=module_leapp_lce,
+        organization=module_sca_manifest_org,
+    ).create()
+    # In case of 'rhel7_server_extras' repos, need to enabled (overridden) repository
+    # to get content available on client host
```
ogajduse, line 119:
```suggestion
    # In case of 'rhel7_server_extras' repos, we need to have (overridden) repository enabled
    # to get content available on content host
```

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,375 @@
...
+
+
+@pytest.fixture
+def verify_target_repo_on_satellite(
+    module_target_sat,
+    function_leapp_cv,
+    module_sca_manifest_org,
+    module_leapp_lce,
+    upgrade_path,
+    function_leapp_ak,
```
ogajduse, line 138: You are not using this fixture in the function. Can you either remove it or use `@pytest.mark.usefixtures('function_leapp_ak')`?

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,375 @@
...
+    assert custom_leapp_host.run('yum update -y').status == 0
+    assert custom_leapp_host.run('yum install leapp-upgrade -y').status == 0
+
+
+@pytest.fixture
+def fix_inhibitors(custom_leapp_host):
+    """Fix inhibitors to avoid hard stop of Leapp tool execution"""
+    source_rhel_major_ver = str(custom_leapp_host.os_version.major)
+    logger.info('Fixing inhibitors for source rhel version %s', source_rhel_major_ver)
+    # In case of Upgrade Path A - Rehl-8 to Rhel-9
```
ogajduse, line 198:
```suggestion
    # In case of Upgrade Path A - RHEL 8 to RHEL 9
```

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,375 @@
...
+    if source_rhel_major_ver == '8':
+        # 1. Firewalld Configuration AllowZoneDrifting Is Unsupported
+        custom_leapp_host.run(
+            'sed -i "s/^AllowZoneDrifting=.*/AllowZoneDrifting=no/" /etc/firewalld/firewalld.conf'
+        )
+    else:
+        # In case of Upgrade Path A - Rehl-7 to Rhel-8
```
ogajduse, line 205:
```suggestion
        # In case of Upgrade Path A - RHEL 7 to RHEL 8
```

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,375 @@
...
+    module_target_sat,
+    custom_leapp_host,
+    module_sca_manifest_org,
+):
+    """Register content host with satellite"""
+    logger.info(
+        'Registering Leapp Host with Satellite - RHEL %s', custom_leapp_host.os_version.major
+    )
+    result = custom_leapp_host.register(
+        module_sca_manifest_org, None, function_leapp_ak.name, module_target_sat
```
ogajduse, line 170: Where is `function_leapp_ak` coming from? It is not defined in args.

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,375 @@
...
+    # In case of 'rhel7_server_extras' repos, need to enabled (overridden) repository
+    # to get content available on client host
+    if upgrade_path['source_version'].split('.')[0] == '7':
+        ActivationKey.content_override(
+            {
+                'id': ak.id,
+                'content-label': RHEL_REPOS['rhel7_server_extras']['id'],
+                'value': 'true',
+            }
+        )
```
ogajduse, line 127: Why are you using `cli` when you are using an `api` in the above call?

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,375 @@
...
+
+:CaseAutomation: Automated
+
+:Upstream: No
+"""
+import pytest
+from broker import Broker
+
+from robottelo.cli.activationkey import ActivationKey
+from robottelo.cli.repository import Repository
```
ogajduse, line 23: Please remove these imports and use `module_target_sat.cli.ActivationKey`  and `module_target_sat.cli.Repository` instead.

ogajduse (changes requested):
I have ACKed this PR before b91f5fdddf94a807e6d31001cc9c9affaf67828b was pushed to the branch.

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,375 @@
...
+            'search_query': f'name = {custom_leapp_host.hostname}',
+        },
+    )
+    module_target_sat.wait_for_tasks(
+        f'resource_type = JobInvocation and resource_id = {job["id"]}', poll_timeout=1800
+    )
+    result = module_target_sat.api.JobInvocation(id=job['id']).read()
+    assert result.succeeded == 1
+    # In case of Upgrade Path A - Rehl-7 to Rhel-8
+    # Inhibitors should fix after running LEAPP-PREUPGRADE Job Template
```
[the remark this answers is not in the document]
vijaysawant: - Few inhibitors come after executing `LEAPP-PREUPGRADE` Job template
- In case of rhel-7.9, if we fix any inhibitor before executing `LEAPP-PREUPGRADE`, it disturbs the flow of execution and gives some more inhibitors after execution of `LEAPP-UPGRADE` Job template
- To maintain flow, inhibitors need to be fix as per this order only
- It doesn't looks good to add that stuff related to rhel-7 inside test module

tests/foreman/cli/test_leapp_client.py
vijaysawant: - These are the titles of inhibitor which we can see inside "/var/log/leapp/leapp-report.txt" file.
- Just adding in comment for understanding purpose

tests/foreman/cli/test_leapp_client.py
vijaysawant: I need to check local changes

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,375 @@
...
+):
+    logger.info(
+        'Creating Leapp Activation Key for RHEL %s -> %s',
+        upgrade_path['source_version'],
+        upgrade_path['target_version'],
+    )
+    # need to publish and promote cv on lce to get content available on client host
+    function_leapp_cv.publish()
+    cvv = function_leapp_cv.read().version[0]
+    cvv.promote(data={'environment_ids': module_leapp_lce.id, 'force': True})
```
Gauravtalreja1, line 111: Can we move this CV specific steps into `function_leapp_cv` fixture?

tests/foreman/cli/test_leapp_client.py
Gauravtalreja1, line 119: Can we have a check for this in AK, whether any repo is left disabled then it can be enabled by override?

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,375 @@
...
+        f'resource_type = JobInvocation and resource_id = {job["id"]}', poll_timeout=1800
+    )
+    result = module_target_sat.api.JobInvocation(id=job['id']).read()
+    assert result.succeeded == 1
+    # Wait for the host to be rebooted and SSH daemon to be started.
+    custom_leapp_host.wait_for_connection()
+
+    custom_leapp_host.clean_cached_properties()
+    new_ver = custom_leapp_host.os_version.major
+    assert new_ver == old_ver + 1
```
Gauravtalreja1, line 375: I think instead of old or new version check, it'd make more sense to check host version is same as target_version, right?
```suggestion
    assert custom_leapp_host.os_version == target_version
```

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,375 @@
...
+    # In case of Upgrade Path A - Rehl-7 to Rhel-8
+    # Inhibitors should fix after running LEAPP-PREUPGRADE Job Template
+    if upgrade_path['source_version'].split('.')[0] == '7':
+        logger.info('Fixing inhibitory for upgrade path : %s', upgrade_path)
+        # 1. Leapp detected loaded kernel drivers which have been removed in RHEL 8
+        custom_leapp_host.run('rmmod floppy')
+        custom_leapp_host.run('rmmod pata_acpi')
+        # 2. Missing required answers in the answer file
+        leapp_report_path = "/var/log/leapp/leapp-report.txt"
+        grep_cmd = f"grep 'Title: Missing required answers in the answer file' {leapp_report_path}"
```
Gauravtalreja1, line 344: Use single quotes wherever possible to keep it consistent
```suggestion
        leapp_report_path = '/var/log/leapp/leapp-report.txt'
        grep_cmd = f'grep "Title: Missing required answers in the answer file" {leapp_report_path}'
```

tests/foreman/cli/test_leapp_client.py
ogajduse: Nothing bad about commenting on the code, in fact in some cases it is really appreciated by other readers to have comments in the code.
My point is that I personally do not see value in having the comments numbered. Imagine that you will add a new step between 1 and 2. You would have to change the numbers of all steps that follow after it. It can be a few, but it can also be tens of comments depending on the complexity of the function/test.

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,375 @@
...
+        'id': 'rhel-7-server-extras-rpms',
+        'name': 'Red Hat Enterprise Linux 7 Server - Extras RPMs x86_64',
+        'releasever': '7',
+        'reposet': 'Red Hat Enterprise Linux 7 Server - Extras (RPMs)',
+        'product': 'Red Hat Enterprise Linux Server',
+    },
+    'rhel8_bos': {
+        'id': 'rhel-8-for-x86_64-baseos-rpms',
+        'name': f'Red Hat Enterprise Linux 8 for x86_64 - BaseOS RPMs {RHEL8_VER}',
+        'releasever': RHEL8_VER,
```
Gauravtalreja1, line 63: If we are eventually using the latest rhel version repositories then why setting up releasever and specific version repos? can't we use the constants or VersionedContent directly?

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,375 @@
...
+
+synced_repos = pytest.StashKey[dict]
+
+
+@pytest.fixture(scope='module')
+def module_stash(request):
+    """Module scoped stash for storing data between tests"""
+    # Please refer the documentation for more details on stash
+    # https://docs.pytest.org/en/latest/reference/reference.html#stash
+    request.node.stash[synced_repos] = {}
```
Gauravtalreja1, line 37: Just curious, what's need of pytest stash here?

vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/cli/test_leapp_client.py -k test_leapp_upgrade_rhel

tests/foreman/cli/test_leapp_client.py
vijaysawant: I completely understood what you meant over here, I will remove numbering.

vijaysawant:
rigger: test-robottelo
pytest: tests/foreman/cli/test_leapp_client.py -k test_leapp_upgrade_rhel

tests/foreman/cli/test_leapp_client.py
ogajduse: We should not. 
Red Hat specifies what the [supported upgrade paths](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html-single/upgrading_from_rhel_8_to_rhel_9/index#con_supported-upgrade-paths_upgrading-from-rhel-8-to-rhel-9) are, therefore we should follow them. And that is exactly the purpose of these constants.
At this very moment, we support upgrading to the latest version of RHEL 8 which is 8.8 atm. But once 8.9 gets released, we still want to test 7.9 -> 8.8.
Once RHEL 8.10 gets released, we should manually switch our constants.

tests/foreman/cli/test_leapp_client.py
ogajduse: There is one test in the test module that is parametrized to two upgrade paths. 7to8 and 8to9. If we do not have the stash implemented, the second upgrade path will fail to enable the RHEL 8.8 repository that is already enabled on target_sat.
Stash enables us to store the synced/enabled repos in it so we can skip their enablement and sync before we try to enable and/or sync them.

tests/foreman/cli/test_leapp_client.py
vijaysawant: I think we had discussion around this change back in some comment, not remembered exactly.
But if we move above 3 lines inside `function_leapp_cv`, cv will create > publish & promote to module_lce without any content, Repos are yet to be enable and sync

tests/foreman/cli/test_leapp_client.py
vijaysawant: - This is special case, `rhel7_server_extras` repo is just added to CV but not enabled on the host > AK > content host

tests/foreman/cli/test_leapp_client.py
vijaysawant: - This has been already discussed with Lukash and Ondrej, we want specific repos to be enabled prior to get sync. 
- Constants does have any REPOS defined for specific rhel version.
- These specific rhel repos will be applicable for leapp scenarios so defined in same test module.

tests/foreman/cli/test_leapp_client.py
vijaysawant: @Gauravtalreja1 Do you mean, why we are using pytest stash if we are making test module parameterised?
stash is for storing data between tests, but here we are using single test module.

tests/foreman/cli/test_leapp_client.py
ogajduse: @vijaysawant 
1. Your reply does not answer Gaurav's question.
2. What did the Phoenix team say about this behavior? Is it the expected behavior of Satellite that this repo is not enabled?

tests/foreman/cli/test_leapp_client.py
ogajduse: @Gauravtalreja1 I would not block this PR on it. Your suggestion would bring us more granularity into the fixture structure. Also, I do not see the eventual `function_leapp_cv` fixture being used anywhere else than in `leapp_sat_content` fixture.

robottelo/cli/repository.py
```diff
@@ -77,3 +77,9 @@ def upload_content(cls, options):
         """Upload content to repository."""
         cls.command_sub = 'upload-content'
         return cls.execute(cls._construct_command(options), output_format='csv', ignore_stderr=True)
+
+    @classmethod
+    def list(cls, options):
+        """List repositories"""
+        cls.command_sub = 'list'
+        return cls.execute(cls._construct_command(options), output_format='csv')
```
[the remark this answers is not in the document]
ogajduse: I initially asked @vijaysawant to implement it as he was using `module_target_sat.execute('hammer repository list').
If your statement is true, we should definitely remove the explicit `list` function.

tests/foreman/cli/test_leapp_client.py
Gauravtalreja1: Yes, @vijaysawant extras isn't default enabled i know, this is default behaviour in 6.14, so it'd be better to have a check for this in AK, whether any repo is left disabled then it can be enabled by override?

tests/foreman/cli/test_leapp_client.py
vijaysawant: That was just a merging issue, fixed now.

ogajduse: resolved

tests/foreman/cli/test_leapp_client.py
ogajduse: I would insist on removing this fixture if the test is not using it. Let me know if you hit any issues after its removal.

robottelo/cli/repository.py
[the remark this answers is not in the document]
ogajduse: Addressed.

tests/foreman/cli/test_leapp_client.py
ogajduse: @Gauravtalreja1 Can we resolve this one?

tests/foreman/cli/test_leapp_client.py
ogajduse: I have changed my mind on this after a deeper look.
Maybe the following code could be moved to the `function_leapp_cv` fixture:

```python
    function_leapp_cv.repository = all_repos  # all_repos would become leapp_repos, see below
    function_leapp_cv = function_leapp_cv.update(['repository'])
    logger.info('Repos to be added to the AK: %s', all_repos)
    # Publish, promote content view to lce
    logger.info(f'Publish, promote cv - {function_leapp_cv.name} to lce - {module_leapp_lce.name}')
    function_leapp_cv.publish()
    cvv = function_leapp_cv.read().version[0]
    cvv.promote(data={'environment_ids': module_leapp_lce.id, 'force': True})
    function_leapp_cv = function_leapp_cv.read()
```

In this case,  the `leapp_sat_content` fixture would need to return `all_repos` and it needs to be a dependency of the `function_leapp_cv` fixture.
Also, it would make sense to rename the `leapp_sat_content` to something like `leapp_repos` as it would only manage repositories.

@vijaysawant What do you think about these potential changes? At least IMO this should be easy to implement.

ogajduse (commented):
In addition to what @Gauravtalreja1 I found an extra bit of code to potentially address.

tests/foreman/cli/test_leapp_client.py
Gauravtalreja1: Sure, its non-blocking

tests/foreman/cli/test_leapp_client.py
Gauravtalreja1: ah I see, just checked, I was under impression that we're adding both scenarios here, my bad, sorry, will resolve this comment then, and will nitpick again with this later :)

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,344 @@
...
+    logger.info(
+        'Creating Leapp Activation Key for RHEL %s -> %s',
+        upgrade_path['source_version'],
+        upgrade_path['target_version'],
+    )
+    return module_target_sat.api.ActivationKey(
+        content_view=function_leapp_cv,
+        environment=module_leapp_lce,
+        organization=module_sca_manifest_org,
+    ).create()
```
Gauravtalreja1, line 110: Just to optimize the count of fixture in use, can we move this LCE/CV/AK create fixtures into a `leapp_sat_content` fixture, as this fixtures be used mostly there and not in test directly?
I think AK will be used that too just for registration so it can be returned from `leapp_sat_content` fixture as well, as it returns nothing atm, wdyt?

Gauravtalreja1 (commented):
@vijaysawant Great job on this PR ❤️
adding a just a few suggestions, let me know if need any help around this

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,344 @@
...
+):
+    """Verify target rhel version repositories has enabled on Satellite Server"""
+    target_rhel_major_ver = upgrade_path['target_version'].split('.')[0]
+    cmd_out = module_target_sat.cli.Repository.list(
+        {
+            'search': f'content_label ~ rhel-{target_rhel_major_ver}',
+            'content-view-id': function_leapp_cv.id,
+            'organization-id': module_sca_manifest_org.id,
+            'lifecycle-environment-id': module_leapp_lce.id,
+        }
```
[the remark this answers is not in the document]
vijaysawant: - Whatever you are saying is fine, i.e host (RHEL8.8) will have `target_version` (BaseOS & AppStream) repos listed unless & untill we set release version (using `subscription-manager --set`) to `RHEL9_VER`.
- As per rhel-upgrade document, we have to make sure that correct `target_rhel_major_ver` repos have been added to the correct `CV`, `LCE` on Satellite.

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,344 @@
...
+    'leapp_sat_content',
+    'register_host_with_satellite',
+    'verify_target_repo_on_satellite',
+    'precondition_check_upgrade_and_install_leapp_tool',
+    'fix_inhibitors',
+)
+def test_leapp_upgrade_rhel(
+    module_target_sat,
+    custom_leapp_host,
+    upgrade_path,
```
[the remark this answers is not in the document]
vijaysawant: - Initially `@pytest.mark.usefixtures(...)` included in test module to execute fixtures in expected order. 
- We want some fixtures (pre-requisites) to be executed prior to run test case
- I will work on suggested changes, if suggested changes work as expected then I will include in upcoming commit otherwise will keep as it is.

tests/foreman/cli/test_leapp_client.py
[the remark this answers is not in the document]
ogajduse: I didn't comment on this fixture before because I wanted to get the PR merged ASAP, but I wanted to share my thoughts now. However, since Gaurav opened the discussion here, let me my perspective on it.

I agree with your perspective that if we instruct Satellite to sync these repositories, they should be there. IMHO, testing whether Satellite does its job in terms of repository enablement and sync should be the job of other tests that we have in our test suite. This way, the Leapp & Satellite integration test can focus solely on testing the integration between Leapp and Satellite.

I hope it makes sense. Let me know what you think.

tests/foreman/cli/test_leapp_client.py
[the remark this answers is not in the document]
ogajduse: I, personally see no value and reason in putting fixtures whose return value is not used in the function body to the function arguments. IMO using the `usefixtures` marker makes the code more concise and readable.

There are a few usages of this marker in robottelo and I personally believe that it is a better practice to use the `usefixtures` marker instead of putting the fixture to function arguments if its return value is not used in the test.

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,257 @@
...
+    """Test to upgrade RHEL host to next major RHEL Realse with Leapp Preupgrade and Leapp Upgrade
+    Job templates
+
+    :id: 7add9b32-a13a-4e65-973c-cd64326b332a
+
+    :Steps:
+        1. Import a subscription manifest and enable, sync source & target repositories
+        2. Create LCE, Create CV, add repositories to it, publish and promote CV, Create AK, etc.
+        3. Register content host with AK
+        4. Varify target rhel repositories are enable on Satellite
```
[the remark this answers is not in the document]
ogajduse: This one must have slipped away. Could you please fix it?

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,344 @@
...
+        5. Update all packages, install leapp tool and fix inhibitors
+        6. Run Leapp Preupgrade and Leapp Upgrade job template
+
+    :expectedresults:
+        1. Update RHEL OS major version to another major version
+
+    """
+    logger.info('Running test for upgrade path %s', upgrade_path)
```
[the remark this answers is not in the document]
ogajduse: In fact, I would like to see us using logging more. It can be a great way to debug code and to understand what is happening in a test. and it could improve the speed which we perform TFA with.

I also agree that the logger.info call in this specific case does not provide much value. Robotello now logs the start of each test phase (setup, call, teardown), so there is no need to duplicate that information.

However, I think it is still a good idea to use logging in places where the results of an operation are not immediately obvious. This can help to make tests more readable and easier to debug.

tests/foreman/cli/test_leapp_client.py
ogajduse: I understand your goal of optimizing the number of fixtures in use, and I agree that this could be a good idea in the long run.

However, I think it is important to get the PR merged ASAP, and making the changes you suggest would take some time to change and test. I also think the current structure is more granular and is logically split into individual fixtures. This makes it easier to understand and maintain the code.

[I have also commented](https://github.com/SatelliteQE/robottelo/pull/11553#discussion_r1318867683) on the leapp_sat_content fixture with some potential changes that could make it more efficient. I think these changes would be a good compromise between optimizing the number of fixtures and keeping the code easy to understand and maintain.

tests/foreman/cli/test_leapp_client.py
```diff
@@ -0,0 +1,344 @@
...
+    custom_leapp_host,
+    module_sca_manifest_org,
+    function_leapp_ak,
+):
+    """Register content host with satellite"""
+    logger.info(
+        'Registering Leapp Host with Satellite - RHEL %s', custom_leapp_host.os_version.major
+    )
+    result = custom_leapp_host.register(
+        module_sca_manifest_org, None, function_leapp_ak.name, module_target_sat
```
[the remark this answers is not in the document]
ogajduse: This one-liner used to be a standard function that @vijaysawant was using instead of fixtures. Later on, we improved the structure of the module by turning the helper functions into fixtures.

This is made as a fixture as the `precondition_check_upgrade_and_install_leapp_tool` fixture assumes that the host is already registered; since those two fixtures are explicitly included as fixtures for the test, it is hard to move the registration to the test itself.

Ideally, the test should only contain steps that are testing the actual Satellite & Leapp integration, and host registration is not related to Leapp.

tests/foreman/cli/test_leapp_client.py
vijaysawant: Based upon the @ogajduse [comment](https://github.com/SatelliteQE/robottelo/pull/11553/commits/d6a87722820f60b675d2d9f8602d1eeea053a955#r1262816633) and suggested nice fixture example, all above small tasks breakdown (converted from function) into reusable fixture by considering any new test case would use the fixtures, I could see you have voted for that.

tests/foreman/cli/test_leapp_client.py
vijaysawant: Sure I will test and do necessary changes

vijaysawant:
rigger: test-robottelo
pytest: tests/foreman/cli/test_leapp_client.py -k test_leapp_upgrade_rhel

tests/foreman/cli/test_leapp_client.py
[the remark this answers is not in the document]
Gauravtalreja1: yes, earlier I asked we could move it to either in test itself or in custom_leapp_host fixture, so custom_leapp_host fixture seems like a good place to move.

ogajduse (approved):
Great work on this PR, @vijaysawant! 
I would like to ask any @SatelliteQE/tier-2-reviewers who decide to review this PR to take into account that this PR should be merged ASAP. PRT passes at the time when I am writing this comment. If you think that some parts of the code should be improved, please open an issue in the repository and assign it to @vijaysawant.

I would say :shipit:

tests/foreman/cli/test_leapp_client.py
Gauravtalreja1: Thanks @vijaysawant :)

tests/foreman/cli/test_leapp_client.py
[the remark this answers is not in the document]
Gauravtalreja1: yes, both ways are fine, either on satellite where we're enabling it, or check on host which where satellite will put it eventually(so it's better IMO). 
And, the setup part doesn't seem to be a big deal since until Leapp upgrade works, it looks fine to me, and is non-blocking atm

vijaysawant:
rigger: test-robottelo
pytest: tests/foreman/cli/test_leapp_client.py -k test_leapp_upgrade_rhel

tests/foreman/cli/test_leapp_client.py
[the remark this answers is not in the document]
Gauravtalreja1: Thanks @vijaysawant

Gauravtalreja1 (approved):

PR #11561, opened 2023-05-30
Applied errata template test and custom product fix

sambible:
Test for Host - Applied Errata report template ( Very simple, just checking for the installed errata, and verifying the date field is there, and is correct ) 

This also includes an update to a clifactory method used to generate a custom product for use in Errata tests, as well as used within the new Applied Errata test. In 6.14, custom products were all disabled by default, and this overrides that to true.


sambible:
It's possible to port this test back to 6.13, by removing the last assertion. I'd prefer not to make a clone test, since they do the same thing pretty much. Not sure how best to handle this.

sambible:
trigger: test-robottelo
pytest: tests/foreman/api/test_reporttemplates.py -k test_positive_applied_errata

tests/foreman/api/test_reporttemplates.py
```diff
@@ -369,22 +380,76 @@ def test_negative_create_report_without_name():
...
+            'organization_id': module_org.id,
+            'report_format': "json",
+            'input_values': {
+                'Filter Errata Type': 'all',
+                'Include Last Reboot': 'no',
+                'Status': 'all',
+            },
+        }
+    )
+    print(res)
```
omkarkhatavkar, line 450: looks like a leftover

tests/foreman/api/test_reporttemplates.py
```diff
@@ -369,22 +380,76 @@ def test_negative_create_report_without_name():
 
 
 @pytest.mark.tier2
-@pytest.mark.stubbed
-def test_positive_applied_errata():
+@pytest.mark.rhel_ver_list([7, 8, 9])
```
ogajduse, line 383: I do not like this approach. It will bite us in the ass once we start supporting RHEL 10. I suggest to change it the following way:
```suggestion
@pytest.mark.rhel_ver_match('^(?!6$)\d+$')
```

tests/foreman/api/test_reporttemplates.py
```diff
@@ -369,22 +380,76 @@ def test_negative_create_report_without_name():
...
 
-        1. POST /api/report_templates/:id/generate
+        1. Generate an Applied Errata report
 
     :expectedresults: A report is generated with all applied errata listed
 
     :CaseImportance: Medium
     """
+    ERRATUM_ID = str(settings.repos.yum_6.errata[2])
+    setup_org_for_a_custom_repo(
```
ogajduse, line 403:
```suggestion
    target_sat.cli_factory.setup_org_for_a_custom_repo(
```

tests/foreman/api/test_reporttemplates.py
```diff
@@ -369,22 +380,76 @@ def test_negative_create_report_without_name():
...
+        {
+            'url': settings.repos.yum_9.url,
+            'organization-id': module_org.id,
+            'content-view-id': module_cv.id,
+            'lifecycle-environment-id': module_lce.id,
+            'activationkey-id': activation_key.id,
+        }
+    )
+    rhel_contenthost.install_katello_ca(target_sat)
+    rhel_contenthost.register_contenthost(module_org.label, activation_key.name)
```
ogajduse, line 413: You are using the deprecated way of registering the host. See the conversation [here](https://redhat-internal.slack.com/archives/C04T3JZ5W8K/p1685545990914279).

tests/foreman/api/test_reporttemplates.py
```diff
@@ -369,22 +380,76 @@ def test_negative_create_report_without_name():
...
     :steps:
 
-        1. POST /api/report_templates/:id/generate
+        1. Generate an Applied Errata report
 
     :expectedresults: A report is generated with all applied errata listed
 
     :CaseImportance: Medium
     """
+    ERRATUM_ID = str(settings.repos.yum_6.errata[2])
```
ogajduse, line 402: Is there any reason for the capitalized name?

ogajduse (changes requested):

tests/foreman/api/test_reporttemplates.py
sambible: It's being used in this case as a constant, so python convention

tests/foreman/api/test_reporttemplates.py
sambible: Is there some guidance on the suggested way to do this? I'm not really familiar with Host registration outside of the UI

tests/foreman/api/test_reporttemplates.py
```diff
@@ -68,6 +71,14 @@ def setup_content(module_entitlement_manifest_org, module_target_sat):
     return ak, org
 
 
+@pytest.fixture(scope='module')
+def activation_key(module_org, module_lce):
+    activation_key = entities.ActivationKey(
+        environment=module_lce, organization=module_org
+    ).create()
```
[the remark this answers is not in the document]
vsedmik: Could we reuse the existing fixture [here](https://github.com/SatelliteQE/robottelo/blob/482edeb5fb857fa8d8defd8130735219e44865f5/pytest_fixtures/component/activationkey.py#L14)?

tests/foreman/api/test_reporttemplates.py
```diff
@@ -369,22 +380,76 @@ def test_negative_create_report_without_name():
...
+@pytest.mark.no_containers
+def test_positive_applied_errata(
+    module_org, module_cv, module_lce, activation_key, rhel_contenthost, target_sat
+):
     """Generate an Applied Errata report
 
     :id: a4b577db-141e-4871-a42e-e93887464986
 
-    :setup: User with reporting access rights, some host with applied errata
+    :setup: A Host with some applied errata applied errata
```
vsedmik, line 392:
```suggestion
    :setup: A Host with some applied errata.
```

tests/foreman/api/test_reporttemplates.py
```diff
@@ -369,22 +380,76 @@ def test_negative_create_report_without_name():
...
+    )
+    rt = (
+        target_sat.api.ReportTemplate()
+        .search(query={'search': 'name="Host - Applied Errata"'})[0]
+        .read()
+    )
+    res = rt.generate(
+        data={
+            'organization_id': module_org.id,
+            'report_format': "json",
```
vsedmik, line 442:
```suggestion
            'report_format': 'json',
```
Might be interesting to also check the other formats. However, this is something more general, not tied to this test case. Non-blocking.

tests/foreman/api/test_reporttemplates.py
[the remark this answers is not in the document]
sambible: I'm just removing it - dont' actually need it as a fixture

tests/foreman/api/test_reporttemplates.py
sambible: agreed, this would be something I'd like to see in a broader coverage increase for report templates

sambible:
trigger: test-robottelo
pytest: tests/foreman/api/test_reporttemplates.py -k test_positive_applied_errata

sambible:
Should have addressed everything, so take a look and lgtm

damoore044 (approved):

Griffin-Sullivan (approved):

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/api/test_reporttemplates.py -k test_positive_applied_errata

tests/foreman/api/test_reporttemplates.py
```diff
@@ -369,22 +371,78 @@ def test_negative_create_report_without_name():
...
+            'input_values': {
+                'Filter Errata Type': 'all',
+                'Include Last Reboot': 'no',
+                'Status': 'all',
+            },
+        }
+    )
+    assert res[0]['erratum_id'] == ERRATUM_ID
+    assert res[0]['issued'] == '2012-01-27'
```
jyejare, line 445: I think checking the erratum ID should be enough because there are chances that the package would be replaced in future without keeping in mind that we are also asserting the issue date.

jyejare (changes requested):
ACK pending minor comment.

tests/foreman/api/test_reporttemplates.py
sambible: Checking the issue date has to do with this being a newly added field in 6.14, and this is testing that feature: https://issues.redhat.com/browse/SAT-17071

sambible: I could potentially just assert the existence of the field, but I don't think there's too much harm in checking the content as well.

jyejare: Yeah, testing the field value should be enough instead of the exact value as that would likely fail tomorrow as we change things in lab and regenerate things like RPMs etc.

ogajduse: @jyejare I did not quite get your arguments. Nothing would fail tomorrow as we do not "change things in lab". We also do not "regenerate things like RPMs". And if we do, I would expect to update the constants/config in our test suite to reflect it.

@sambible It is entirely up to you, but if you plan to keep the date there, please put it in `settings.repos.yum_6.errata`.

Ideal would be 
```yaml
  YUM_6:
    URL: '@format {this[robottelo].repos_hosting_url}/needed_errata'
    ERRATA:
    - ID: RHBA-2012:1030
      DATE_ISSUED: 
    - RHEA-2012:0058
    - RHSA-2012:0055
    - RHSA-2012:0056
    - RHSA-2012:0057
```
but that would break in all places where we use errata IDs from YUM_6.

I would go with the following instead.
```yaml
  YUM_6:
    URL: '@format {this[robottelo].repos_hosting_url}/needed_errata'
    ERRATA:
    - RHBA-2012:1030
    - RHEA-2012:0058
    - RHSA-2012:0055
    - RHSA-2012:0056
    - RHSA-2012:0057
    DATE_ISSUED:
    - YYYY-MM-DD
    - YYYY-MM-DD
    - 2012-01-27
    - YYYY-MM-DD
    - YYYY-MM-DD
```

tests/foreman/api/test_reporttemplates.py
```diff
@@ -369,22 +371,78 @@ def test_negative_create_report_without_name():
...
+        }
+    )
+    result = rhel_contenthost.register(module_org, module_location, activation_key.name, target_sat)
+    assert f'The registered system name is: {rhel_contenthost.hostname}' in result.stdout
+    assert rhel_contenthost.subscribed
+    result = rhel_contenthost.run(f'yum install -y {FAKE_1_CUSTOM_PACKAGE}')
+    assert result.status == 0
+    result = rhel_contenthost.run(f'rpm -q {FAKE_1_CUSTOM_PACKAGE}')
+    assert result.status == 0
+    rhel_contenthost.add_rex_key(satellite=target_sat)
```
ogajduse, line 413: The default for `setup_remote_execution` in the `register` method is `True`, therefore you should not need to `add_rex_key` additionally.

tests/foreman/api/test_reporttemplates.py
ogajduse: Could you apply @vsedmik's suggestion to keep the quote type consistent?

ogajduse (changes requested):
I left 3 comments of which one is non-blocking.

ogajduse:
trigger: test-robottelo
pytest: tests/foreman/api/test_reporttemplates.py -k test_positive_applied_errata

tests/foreman/api/test_reporttemplates.py
jyejare: @ogajduse  Forecasting the future is really hard :)

I wont say that we will change or wont change but lets make sure that we dont break things for such changes which are volitile especially if those are not critical. Like checking the date here is not the critical assertion of this test.

Hence I dont agree with your comment ` did not quite get your arguments. Nothing would fail tomorrow as we do not "change things in lab". We also do not "regenerate things like RPMs".`

I am not 100% agree on keeping these dates as part of `configuration` too as they dont seems to be critical. Lets not unnecessarily assert things that are not tied to the test expectation or goal and fail the entire test on it.

jyejare (approved):

tests/foreman/api/test_reporttemplates.py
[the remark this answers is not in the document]
Gauravtalreja1: @sambible this fixture still not removed, and even not used anywhere, can you check?

tests/foreman/api/test_reporttemplates.py
```diff
@@ -369,22 +379,77 @@ def test_negative_create_report_without_name():
 
 
 @pytest.mark.tier2
-@pytest.mark.stubbed
-def test_positive_applied_errata():
+@pytest.mark.rhel_ver_match(r'^(?!6$)\d+$')
```
Gauravtalreja1, line 382:
```suggestion
@pytest.mark.rhel_ver_match('[^6]')
```

tests/foreman/api/test_reporttemplates.py
```diff
@@ -369,22 +379,77 @@ def test_negative_create_report_without_name():
...
+            'activationkey-id': activation_key.id,
+        }
+    )
+    result = rhel_contenthost.register(module_org, module_location, activation_key.name, target_sat)
+    assert f'The registered system name is: {rhel_contenthost.hostname}' in result.stdout
+    assert rhel_contenthost.subscribed
+    result = rhel_contenthost.run(f'yum install -y {FAKE_1_CUSTOM_PACKAGE}')
+    assert result.status == 0
+    result = rhel_contenthost.run(f'rpm -q {FAKE_1_CUSTOM_PACKAGE}')
+    assert result.status == 0
```
Gauravtalreja1, line 420:
```suggestion
    assert rhel_contenthost.execute(f'yum install -y {FAKE_1_CUSTOM_PACKAGE}').status == 0
    assert rhel_contenthost.execute(f'rpm -q {FAKE_1_CUSTOM_PACKAGE}').status == 0
```

tests/foreman/api/test_reporttemplates.py
ogajduse: @Gauravtalreja1 see https://github.com/SatelliteQE/robottelo/pull/11561#discussion_r1211868080.
We have to either come up with some smart solution that would allow us to select all currently supported versions or keep this ugly regex.

sambible:
Removed unnecessary fixture, and cleaned up those asserts.

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/api/test_reporttemplates.py -k test_positive_applied_errata

PR #11571, opened 2023-06-01
Fix get_sat_version to handle stream

Gauravtalreja1:


lpramuk (changes requested):
Stream Satellite has its version and this version is **real**  (definitely not `9999`)

To have real version is one purpose of many why developers introduced us Stream.

NACK in favor of #11572

PR #11573, opened 2023-06-01
Add treeinfo ignore test

vsedmik:
6.14 implements a feature where the `treeinfo` file can be added to the repository's ignorable_content. I that case the `treeinfo` file is not synced (as well as any content referenced by it like `images\`).

This introduces a test that exercises the functionality.

### DIFF
--- a/tests/foreman/api/test_repository.py
+++ b/tests/foreman/api/test_repository.py
@@ -1249,6 +1249,57 @@ class TestRepository:
         assert len(files) == packages_count
         assert constants.RPM_TO_UPLOAD not in files
 
+    @pytest.mark.tier3
+    @pytest.mark.parametrize('policy', ['additive', 'mirror_content_only'])
+    def test_positive_sync_with_treeinfo_ignore(
+        self, target_sat, function_entitlement_manifest_org, policy
+    ):
+        """Verify that the treeinfo file is not synced when added to ignorable content
+        and synced otherwise. Check for applicable mirroring policies.
+
+        :id: d7becf1d-3883-468d-88c4-d513a2e2e90a
+
+        :parametrized: yes
+
+        :steps:
+            1. Enable RHEL8 BaseOS KS repo.
+            2. Add `treeinfo` to ignorable content and sync, check it's missing.
+            3. Remove the `treeinfo` from ignorable content, resync, check again.
+
+        :expectedresults:
+            1. The sync should succeed.
+            2. The treeinfo file should be missing when in ignorable content and present otherwise.
+
+        :customerscenario: true
+
+        :BZ: 2174912, 2135215
+
+        """
+        distro = 'rhel8_bos'
+        repo_id = target_sat.api_factory.enable_rhrepo_and_fetchid(
+            basearch='x86_64',
+            org_id=function_entitlement_manifest_org.id,
+            product=constants.REPOS['kickstart'][distro]['product'],
+            reposet=constants.REPOS['kickstart'][distro]['reposet'],
+            repo=constants.REPOS['kickstart'][distro]['name'],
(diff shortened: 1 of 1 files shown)


tests/foreman/api/test_repository.py
```diff
@@ -1249,6 +1249,57 @@ def test_positive_mirroring_policy(self, target_sat):
...
+        :expectedresults:
+            1. The sync should succeed.
+            2. The treeinfo file should be missing when in ignorable content and present otherwise.
+
+        :customerscenario: true
+
+        :BZ: 2174912, 2135215
+
+        """
+        distro = 'rhel8_bos'
```
omkarkhatavkar, line 1278: is this the only case with rhel8? we should parameterize if that can be tested other rhel versions

vsedmik: Yes, it could be parametrized for other RHELs, I'm just not sure if we want to. I mean, the case here is to ensure we can skip the treeinfo file, no matter where it comes from.
But I am open to any ideas for sure, go persuade me! :)

omkarkhatavkar: If you still want to run this on the RHEL8, I'm good with having it as part. Have that part of the parameter. e.g. https://github.com/SatelliteQE/robottelo/blob/aab4b818e8df670ccd4ea159fed9cf2fd7dbfe00/tests/foreman/ui/test_host.py#L2054

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/api/test_repository.py::TestRepository::test_positive_sync_with_treeinfo_ignore

tests/foreman/api/test_repository.py
synkd: I agree with @vsedmik that it's not necessary to parametrize this for other RHEL versions. This test is verifying that Satellite responds as expected to the file structure of bootable repositories (and I assume other bootable installation trees, e.g. ISOs). This file structure should be consistent across bootable media formats - if this test were to pass against one repo but fail against another, that would indicate an issue with the repository rather than with Satellite.

As for defining the repo in the body of the test vs. in the parameters, I don't currently have a strong opinion. @omkarkhatavkar could you please explain why it would be preferable to define this in the parameters?

shweta83 (approved):

tests/foreman/api/test_repository.py
ogajduse: I share the view with @synkd.
I could imagine having a helper function in robottelo that would tell what is the latest supported RHEL version that the test could use.
The only benefit of using the function would allow us to test the repository content of the latest RHEL and not care about updating the RHEL (repo) version in the test(s).
But my suggestion may make the logic unnecessarily complicated.

ogajduse (approved):
ACK, pending discussion about RHEL version parametrization.

tests/foreman/api/test_repository.py
omkarkhatavkar: My point was will you be able to certify this feature working in RHEL9 by just testing on RHEL8 ? If that is true I'm okay. If not there is need of parameterization

vsedmik: I think so. This feature is not about testing RHELx repo at all. The case here is to ensure that Satellite can ignore the treeinfo file while syncing a repository that contains it. So any repo containing the treeinfo file can be used here, it  does not really matter where it comes from, it does not even need to be a RHEL repo at all. I just picked RHEL as the most straightforward option.

omkarkhatavkar: Okay, make sense !

omkarkhatavkar (approved):

PR #11578, opened 2023-06-01
[6.14.z] Adding better coverage to enable/disable stream test

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/11514

https://github.com/SatelliteQE/robottelo/pull/11481 got merged before I could investigate. Apparently this test as it was before can pass on Satellite, but the case from the BZ will still fail. The original intent was to have the test case fail and then we can check automation when we get the stream snap with the fix in it. This test will now fail until the fix is added to stream.

### DIFF
--- a/tests/foreman/maintain/test_service.py
+++ b/tests/foreman/maintain/test_service.py
@@ -148,10 +148,17 @@ def test_positive_service_enable_disable(sat_maintain):
 
     :customerscenario: true
     """
+    result = sat_maintain.cli.Service.stop()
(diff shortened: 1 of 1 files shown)


Satellite-QE:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_service.py -m stream

Griffin-Sullivan:
Failure is expected. We should merge as is.

adarshdubey-star (approved):

jyejare (approved):

PR #11595, opened 2023-06-05
[6.14.z] Add CLI test for applied filters

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/11542

Satellite 6.14.0 is going to implement indication and listing of filters applied to a CVV. This PR adds a test to ensure the correct functionality in hammer CLI.

Note: The upstream changes tested in this PR should land in the next stream snap (16.0). Expect the PRT to fail until then.

### DIFF
--- a/tests/foreman/cli/test_contentviewfilter.py
+++ b/tests/foreman/cli/test_contentviewfilter.py
@@ -16,6 +16,8 @@
 
 :Upstream: No
 """
+import random
+
 import pytest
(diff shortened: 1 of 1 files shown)


vsedmik:
Looks like the PRT passed 2 times here (#2631, 2632) but the `Fetch the PRT status` action fails, blocking this from getting auto-merged. :(

PR #11609, opened 2023-06-06
Close-loop 1757394

lhellebr:
Requires https://github.com/SatelliteQE/nailgun/pull/928 . Test results included there.

### DIFF
--- a/tests/foreman/api/test_user.py
+++ b/tests/foreman/api/test_user.py
@@ -399,6 +399,37 @@ class TestUser:
         with pytest.raises(HTTPError):
             entities.User(auth_source='').create()
 
+    @pytest.mark.tier1
+    def test_positive_table_preferences(self, module_target_sat):
+        """Create a user, create their Table Preferences, read it
+
+        :id: 10fda3f1-4fee-461b-a413-a4d1fa098a94
+
+        :expectedresults: Table Preferences can be accessed
+
+        :CaseImportance: Medium
+
+        :customerscenario: true
+
+        :BZ: 1757394
+        """
+        existing_roles = entities.Role().search()
+        password = gen_string('alpha')
+        user = entities.User(role=existing_roles, password=password).create()
+        name = "hosts"
+        columns = ["power_status", "name", "comment"]
+        sc = ServerConfig(auth=(user.login, password), url=module_target_sat.url, verify=False)
+        entities.TablePreferences(sc, user=user, name=name, columns=columns).create()
+        table_preferences = entities.TablePreferences(sc, user=user).search()
+        assert len(table_preferences) == 1
+        tp = table_preferences[0]
+        assert hasattr(tp, 'name')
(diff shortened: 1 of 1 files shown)


tests/foreman/api/test_user.py
```diff
@@ -399,6 +399,37 @@ def test_negative_create_with_blank_authorized_by(self):
...
+        :BZ: 1757394
+        """
+        existing_roles = entities.Role().search()
+        password = gen_string('alpha')
+        user = entities.User(role=existing_roles, password=password).create()
+        name = "hosts"
+        columns = ["power_status", "name", "comment"]
+        sc = ServerConfig(auth=(user.login, password), url=module_target_sat.url, verify=False)
+        entities.TablePreferences(sc, user=user, name=name, columns=columns).create()
+        table_preferences = entities.TablePreferences(sc, user=user).search()
```
adarshdubey-star, line 423: non blocking: can we do something like this?
```suggestion
        table_preferences = entities.TablePreferences(sc, 
        user=user,name=name,columns=columns).create().search()
```

tests/foreman/api/test_user.py
```diff
@@ -399,6 +399,37 @@ def test_negative_create_with_blank_authorized_by(self):
...
+        assert len(table_preferences) == 1
+        tp = table_preferences[0]
+        assert hasattr(tp, 'name')
+        assert hasattr(tp, 'columns')
+        assert tp.name == 'hosts'
```
adarshdubey-star, line 428: Non blocking :  Feels a bit redundant assertion , as we are setting the same variable above `name = "hosts"`
```suggestion
```

tests/foreman/api/test_user.py
lhellebr: It's perhaps the same, but I like the original logic better: it clearly shows that I first create something, then I search for a list of those things independently.

tests/foreman/api/test_user.py
lhellebr: We set some local variable in Python above, yes... but this is about asserting that the API indeed returned it!

tests/foreman/api/test_user.py
shweta83: Try to avoid the direct use of entities. We are trying to remove this dependency from robottelo as much as possible.

lhellebr: I know and I did it consistently with the rest of the module:
```
$ grep -o entities tests/foreman/api/test_user.py | wc -l
71
```

adarshdubey-star (approved):
Ack

adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/api/test_user.py -k test_positive_table_preferences
nailgun: 928

adarshdubey-star:
Do we need it for 6.13/6.12 if yes can you please add labels accordingly?

jyejare (approved):

PR #11619, opened 2023-06-07
Add autoflake in precommit check

shweta83:
**Why do we need it?** Existing pre-commit checks doesn't remove unused imports and variables from the code. With autoflake, it remove these unused imports and variables in-place and avoids the manual effort.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,6 +26,18 @@ repos:
   rev: 6.0.0
   hooks:
   - id: flake8
+- repo: https://github.com/PyCQA/autoflake
+  rev: v2.1.1
+  hooks:
+  - id: autoflake
+    "types": [ python ]
+    require_serial: true
+    args:
+      - "--in-place"
+      - "--expand-star-imports"
+      - "--remove-duplicate-keys"
+      - "--remove-unused-variables"
+      - "--remove-all-unused-imports"
 - repo: local
   hooks:
     - id: fix-uuids


adarshdubey-star:
Can you take a look at the checks once

.pre-commit-config.yaml
```diff
@@ -26,6 +26,18 @@ repos:
...
+  hooks:
+  - id: autoflake
+    "types": [ python ]
+    require_serial: true
+    args:
+      - "--in-place"
+      - "--expand-star-imports"
+      - "--remove-duplicate-keys"
+      - "--remove-unused-variables"
+      - "--remove-all-unused-imports"
```
ogajduse, line 40: Can you try configuring autoflake in `pyproject.toml` instead?

jyejare (changes requested):
In CI chcks, One of the autoflake plugin trying to remove `pass` keyword from code which is not necessary to remove as they are placeholders and required to function the functions properly !

Please rectify the issue!

jyejare:
Agree but using pass is a common behavior in programmers and that also tells it is not implemented.


I really dont see we should have any checks to remove it. Probably we can ignore that check from the hook.

ogajduse:
> Agree but using pass is a common behavior in programmers and that also tells it is not implemented.
> 
> I really dont see we should have any checks to remove it. Probably we can ignore that check from the hook.

Agreed, using `pass` is a common habit, but that does not mean that we have to necessarily follow it.

[PEP 257](https://www.python.org/dev/peps/pep-0257/#specification) states
> A docstring is a string literal that occurs as the first statement in a module, function, class, or method definition. Such a docstring becomes the __doc__ special attribute of that object.

That means that if the function has `__doc__` defined, you do not have to explicitly state that the function is no-op by adding `pass` statement.

Also, `pass` does not tell you that the function is not implemented. We rather use the `NotImplementedError` exception for such cases. If the function is to be implemented, we should rather use [Ellipsis](https://docs.python.org/3/library/constants.html#Ellipsis).

There is a [nice thread](https://stackoverflow.com/a/49754261) on this topic on Stack Overflow.

Since autoflake is developed by [PyCQA](https://github.com/PyCQA), I would trust them and go with what their tools suggest.

jyejare:
@ogajduse Sure! And thanks for the nice explanation!  Well still I will keep my NACK to fix the files which has `pass` statement in them and to pass the checks !

jyejare:
@ogajduse @shweta83 The failing checks are still needed to be fixed in this PR else it will fail for all other PRs post merge!

lhellebr:
About `pass` vs `...`, this sums it up quite nicely: https://stackoverflow.com/questions/55274977/when-is-the-usage-of-the-python-ellipsis-to-be-preferred-over-pass
@ogajduse , why do you think "If the function is to be implemented, we should rather use Ellipsis."? I don't agree with that one.
OTOH, I do agree that we should use NotImplementedError more.

ogajduse:
@lhellebr I perhaps did not study it deeper. I do not have a strong opinion on whether to use Ellipsis or `pass`

ogajduse:
I am in favor of closing this PR as there is a plan to include [ruff](beta.ruff.rs/) into our workflow in robottelo in the same way as in broker, nailgun, and airgun.

ogajduse:
Ruff added as a part of #12621

jyejare:
@shweta83 We should now close this as we have ruff in place now! 

Thanks for all your efforts though 👍

PR #11729, opened 2023-06-21
[6.14 RFE] Upgrade test for enabled repos in sca

ColeHiggins2:
SAT-16805

Upgrade scenario for enabled repos in non-sca org

In a non-sca organization, enabled repos with an attached subscription should be set to "Enabled(Override)" after an upgrade to 6.14.

### DIFF
--- a/tests/upgrades/test_repository.py
+++ b/tests/upgrades/test_repository.py
@@ -224,7 +224,14 @@ class TestScenarioCustomRepoOverrideCheck:
 
     @pytest.mark.pre_upgrade
     def test_pre_scenario_custom_repo_sca_toggle(
-        self, target_sat, sat_upgrade_chost, save_test_data, default_location
+        self,
+        target_sat,
+        function_org,
+        function_product,
+        function_lce,
+        sat_upgrade_chost,
+        save_test_data,
+        default_location,
     ):
         """This is a pre-upgrade scenario test to verify that repositories in a non-sca org
         set to "Enabled" should be overridden to "Enabled(Override)" when upgrading to 6.14.
@@ -244,31 +251,33 @@ class TestScenarioCustomRepoOverrideCheck:
             2. Custom Repository is enabled on Host.
 
         """
-        org = target_sat.api.Organization().create()
-        lce = target_sat.api.LifecycleEnvironment(organization=org).create()
+        # org = target_sat.api.Organization().create()
+        # lce = target_sat.api.LifecycleEnvironment(organization=org).create()
 
-        product = target_sat.api.Product(organization=org).create()
-        repo = target_sat.api.Repository(product=product.id, url=settings.repos.yum_1.url).create()
-        repo.sync()
-        content_view = target_sat.publish_content_view(org, repo)
-        content_view.version[0].promote(data={'environment_ids': lce.id})
-        ak = target_sat.api.ActivationKey(
-            content_view=content_view, organization=org.id, environment=lce
+        # product = target_sat.api.Product(organization=org).create()
+        repo = target_sat.api.Repository(
+            product=function_product.id, url=settings.repos.yum_1.url
         ).create()
-        if not target_sat.is_sca_mode_enabled(org.id):
-            subscription = target_sat.api.Subscription(organization=org).search(
-                query={'search': f'name={product.name}'}
+        repo.sync()
+        content_view = target_sat.publish_content_view(function_org, repo)
+        content_view.version[0].promote(data={'environment_ids': function_lce.id})
+        ak = target_sat.api.ActivationKey(
+            content_view=content_view, organization=function_org.id, environment=function_lce
+        ).create()
+        if not target_sat.is_sca_mode_enabled(function_org.id):
+            subscription = target_sat.api.Subscription(organization=function_org).search(
+                query={'search': f'name={function_product.name}'}
             )[0]
             ak.add_subscriptions(data={'subscription_id': subscription.id})
-        sat_upgrade_chost.register(org, default_location, ak.name, target_sat)
+        sat_upgrade_chost.register(function_org, default_location, ak.name, target_sat)
         product_details = sat_upgrade_chost.execute('subscription-manager repos --list')
         assert 'Enabled:   1' in product_details.stdout
 
         save_test_data(
             {
                 'rhel_client': sat_upgrade_chost.hostname,
-                'org_name': org.name,
-                'product_name': product.name,
+                'org_name': function_org.name,
+                'product_name': function_product.name,
                 'repo_name': repo.name,
                 'product_details': product_details.stdout,
             }
--- a/tests/upgrades/test_repository.py
+++ b/tests/upgrades/test_repository.py
@@ -251,10 +251,6 @@ class TestScenarioCustomRepoOverrideCheck:
             2. Custom Repository is enabled on Host.
 
         """
-        # org = target_sat.api.Organization().create()
-        # lce = target_sat.api.LifecycleEnvironment(organization=org).create()
-
-        # product = target_sat.api.Product(organization=org).create()
         repo = target_sat.api.Repository(
             product=function_product.id, url=settings.repos.yum_1.url
         ).create()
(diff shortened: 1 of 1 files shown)


tests/upgrades/test_repository.py
```diff
@@ -203,3 +203,83 @@ def test_post_scenario_custom_repo_check(self, target_sat, pre_upgrade_data):
...
+        ak = target_sat.api.ActivationKey(
+            content_view=content_view, organization=org.id, environment=lce
+        ).create()
+        if not target_sat.is_sca_mode_enabled(org.id):
+            subscription = target_sat.api.Subscription(organization=org).search(
+                query={'search': f'name={product.name}'}
+            )[0]
+            ak.add_subscriptions(data={'subscription_id': subscription.id})
+        sat_upgrade_chost.install_katello_ca(target_sat)
+        sat_upgrade_chost.register_contenthost(org.label, ak.name)
```
Gauravtalreja1, line 246: Can you use global registration here please?

tests/upgrades/test_repository.py
```diff
@@ -203,3 +203,83 @@ def test_post_scenario_custom_repo_check(self, target_sat, pre_upgrade_data):
...
+        product_details = sat_upgrade_chost.execute('subscription-manager repos --list')
+        assert "Enabled:   1" in product_details.stdout
```
Gauravtalreja1, line 248:
```suggestion
        assert 'Enabled:   1' in product_details.stdout
```

tests/upgrades/test_repository.py
```diff
@@ -203,3 +203,83 @@ def test_post_scenario_custom_repo_check(self, target_sat, pre_upgrade_data):
...
+        client_hostname = pre_upgrade_data.get('rhel_client')
+        org_name = pre_upgrade_data.get('org_name')
+        product_name = pre_upgrade_data.get('product_name')
+        repo_name = pre_upgrade_data.get('repo_name')
+        rhel_client = Broker(host_class=ContentHost).from_inventory(
+            filter=f'@inv.hostname == "{client_hostname}"'
+        )[0]
+        result = rhel_client.execute('subscription-manager repo-override --list')
+        assert 'enabled: 1' in result.stdout
+        assert f'{org_name}_{product_name}_{repo_name}' in result.stdout
```
Gauravtalreja1, line 285: Can we fetch this repo label from AK instead?

ColeHiggins2: we are checking that the repo-override gets done here. we could check the ak but can i ask for your reasoning?

tests/upgrades/test_repository.py
```diff
@@ -203,3 +203,82 @@ def test_post_scenario_custom_repo_check(self, target_sat, pre_upgrade_data):
         )[0]
         result = rhel_client.execute(f'yum install -y {FAKE_4_CUSTOM_PACKAGE_NAME}')
         assert result.status == 0
+
+    @pytest.mark.pre_upgrade
+    def test_pre_scenario_custom_repo_check_sca_toggle(
```
jyejare, line 208: Please add a new class for the new scenario, there is no harm in keeping it under one class, but then it's easy to identify them separately as a scenario.

tests/upgrades/test_repository.py
```diff
@@ -203,3 +203,82 @@ def test_post_scenario_custom_repo_check(self, target_sat, pre_upgrade_data):
...
+        self, target_sat, sat_upgrade_chost, save_test_data, default_location
+    ):
+        """This is a pre-upgrade scenario test to verify that repositories in a non-sca org
+        set to "Enabled" should be overridden to "Enabled(Override)" when upgrading to 6.14.
```
jyejare, line 212: Would you like to limit this only for 6.14 ?

tests/upgrades/test_repository.py
```diff
@@ -203,3 +203,82 @@ def test_post_scenario_custom_repo_check(self, target_sat, pre_upgrade_data):
...
+            4. Create Activation Key and add Subscription.
+            5. Create a Content Host, register it, and check Repository Sets for enabled Repository.
+
+        :expectedresults:
+
+            1. Custom Repository is created.
+            2. Custom Repository is enabled on Host.
+
+        """
+        org = target_sat.api.Organization().create()
```
jyejare, line 229: You could use function_org fixture and pass it down to the post via save_test_dat

jyejare: Even we could simply use module_org if we are not conflicting in these 2 scenaeios in pre_upgrade part.

tests/upgrades/test_repository.py
```diff
@@ -203,3 +203,82 @@ def test_post_scenario_custom_repo_check(self, target_sat, pre_upgrade_data):
...
+        lce = target_sat.api.LifecycleEnvironment(organization=org).create()
```
jyejare, line 230: function_lce or module_lce fixture?

tests/upgrades/test_repository.py
```diff
@@ -203,3 +203,82 @@ def test_post_scenario_custom_repo_check(self, target_sat, pre_upgrade_data):
...
+
+        product = target_sat.api.Product(organization=org).create()
```
jyejare, line 232: function_product / module_product fixtures ?

tests/upgrades/test_repository.py
```diff
@@ -203,3 +203,82 @@ def test_post_scenario_custom_repo_check(self, target_sat, pre_upgrade_data):
...
+        repo = target_sat.api.Repository(product=product.id, url=settings.repos.yum_1.url).create()
+        repo.sync()
+        content_view = target_sat.publish_content_view(org, repo)
+        content_view.version[0].promote(data={'environment_ids': lce.id})
+        ak = target_sat.api.ActivationKey(
+            content_view=content_view, organization=org.id, environment=lce
+        ).create()
+        if not target_sat.is_sca_mode_enabled(org.id):
```
jyejare, line 240: I think by default all org has `SCA` enabled, isnt it ?

if yes, what would be the else condition here?

jyejare (changes requested):
ACK pending improvement comments and questions !

tests/upgrades/test_repository.py
ColeHiggins2: for now yes

tests/upgrades/test_repository.py
ColeHiggins2: for this scenario, I want to make sure everything is new and no other tests conflict with it. It might pull more data into product details that is not needed.

tests/upgrades/test_repository.py
ColeHiggins2: same here

tests/upgrades/test_repository.py
ColeHiggins2: same here

tests/upgrades/test_repository.py
jyejare: Function org fixture does the same thing that you are doing here , it saves u that line. Also if that fixture is being asked by other test it will create new org.

tests/upgrades/test_repository.py
jyejare: Same as my last comment

jyejare (commented):
I think we have fixture so that we dont repeat the same steps again in the tests and we should must use them!

jyejare (approved):

tests/upgrades/test_repository.py
```diff
@@ -203,3 +203,105 @@ def test_post_scenario_custom_repo_check(self, target_sat, pre_upgrade_data):
...
+        product_details = sat_upgrade_chost.execute('subscription-manager repos --list')
+        assert 'Enabled:   1' in product_details.stdout
+
+        save_test_data(
+            {
+                'rhel_client': sat_upgrade_chost.hostname,
+                'org_name': function_org.name,
+                'product_name': function_product.name,
+                'repo_name': repo.name,
+                'product_details': product_details.stdout,
```
Griffin-Sullivan, line 278: non-blocking. Not sure if saving this helps for debugging purposes, but this is never used in the post-upgrade.

Griffin-Sullivan (approved):

tests/upgrades/test_repository.py
Gauravtalreja1: yes, I got that, I was asking if we could get this repo label from AK's product-content command instead of making this

PR #11737, opened 2023-06-23
Remove redundant rhev provisioning test

Gauravtalreja1:
1. Already covered https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/cli/test_computeresource_rhev.py#L411
2. Use destroy_vm_on_host_delete setting to cleanup hosts in rhevm when deleted from satellite in teardown

### DIFF
--- a/pytest_fixtures/component/settings.py
+++ b/pytest_fixtures/component/settings.py
@@ -1,19 +1,21 @@
 # Settings Fixtures
 import pytest
-from nailgun import entities
 
 
-@pytest.fixture(scope="function")
-def setting_update(request):
+@pytest.fixture()
+def setting_update(request, target_sat):
     """
     This fixture is used to create an object of the provided settings parameter that we use in
     each test case to update their attributes and once the test case gets completed it helps to
     restore their default value
     """
-    setting_object = entities.Setting().search(query={'search': f'name={request.param}'})[0]
+    key_val = request.param
+    setting, new_value = tuple(key_val.split('=')) if '=' in key_val else (key_val, None)
+    setting_object = target_sat.api.Setting().search(query={'search': f'name={setting}'})[0]
     default_setting_value = setting_object.value
-    if default_setting_value is None:
-        default_setting_value = ''
+    if new_value is not None:
+        setting_object.value = new_value
+        setting_object.update({'value'})
     yield setting_object
     setting_object.value = default_setting_value
     setting_object.update({'value'})
--- a/tests/foreman/cli/test_computeresource_rhev.py
+++ b/tests/foreman/cli/test_computeresource_rhev.py
@@ -380,7 +380,7 @@ def test_negative_add_image_rhev_with_invalid_name(rhev, module_os):
 @pytest.mark.on_premises_provisioning
 @pytest.mark.tier3
 @pytest.mark.rhel_ver_match('[^6]')
-@pytest.mark.parametrize('setting_update', ['destroy_vm_on_host_delete'], indirect=True)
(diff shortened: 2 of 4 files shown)


shweta83 (approved):
Ack

shubhamsg199 (approved):
Ack

tests/foreman/cli/test_computeresource_rhev.py
```diff
@@ -380,8 +380,10 @@ def test_negative_add_image_rhev_with_invalid_name(rhev, module_os):
 @pytest.mark.on_premises_provisioning
 @pytest.mark.tier3
 @pytest.mark.rhel_ver_match('[^6]')
+@pytest.mark.parametrize('setting_update', ['destroy_vm_on_host_delete'], indirect=True)
 def test_positive_provision_rhev_with_host_group(
     request,
+    setting_update,
```
vsedmik, line 386: Do you set the value of the this settings anywhere lower in test, fixture or elsewhere? I just can't see the purpose.

jyejare (changes requested):
ACK pending question/comment!

pondrejk (approved):
ack pending existing comments

tests/foreman/cli/test_computeresource_rhev.py
Gauravtalreja1: Good catch @vsedmik, it was just a copy paste for indirect parametrization we've been using for all tests, I've updated the setting_update  fixture to accept setting name and value through request.param, and it works I tested it locally.
@vsedmik @jyejare Do you mind re-reviewing this PR, Thank you! ❤️

jyejare: I mean the initial question, do we really need this parameterization, Because by default this setting is `True`.

Do you see any test is changing this to False making VMs dropped?

pytest_fixtures/component/settings.py
```diff
@@ -1,19 +1,20 @@
...
-def setting_update(request):
+@pytest.fixture()
+def setting_update(request, target_sat):
     """
     This fixture is used to create an object of the provided settings parameter that we use in
     each test case to update their attributes and once the test case gets completed it helps to
     restore their default value
     """
-    setting_object = entities.Setting().search(query={'search': f'name={request.param}'})[0]
+    setting, value_to_change = request.param
```
jyejare, line 12: There are many tests using this fixture and they dont return tuple, so those tests will fail here with unpacking as we will be simply unpacking single string into tuple.

Handle the unpacking and further process based on parameters please.

jyejare (changes requested):
ACK pending question and change required in tests using fixture

vsedmik (approved):
The fixture update looks nice!
ACK

tests/foreman/cli/test_computeresource_rhev.py
Gauravtalreja1: yeah, we need it, default value for this setting is False which doesn't delete host on CR, and we changing it to True to make sure provisioned host is cleaned up from CR

pytest_fixtures/component/settings.py
Gauravtalreja1: ah, right, I just updated the test to not pass key value as tuple instead I'm now passing key=value and for tests which don't pass value we can keep value as none, and no value gets updated by this fixture as earlier.
Please re-review let me know if you still see any issues with this.

ColeHiggins2 (approved):

jyejare:
trigger: test-robottelo
pytest: tests/foreman/api --uses-fixtures setting_update

jyejare (approved):
The 1 test PRT failure looks unrelated to this change, So PR looks good now!

PR #11783, opened 2023-06-28
register function made more flexible

pondrejk:
The global registration dialog changed since this function was created, it is possible to not specify org and loc, which the register function didn't allow. The function also demanded org supplied as an object, which caused problems in onboarding more tests to it (some use org defined as a dict -- an output from cli helpers). PR changes the function in a way that doesn't require updating the existing callers

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -588,8 +588,8 @@ class ContentHost(Host, ContentHostMixins):
         using a global registration template.
 
         :param target: Satellite or Capusle object to register to, required.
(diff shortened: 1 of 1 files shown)


rmynar (approved):

Griffin-Sullivan (approved):
Do you have any tests we can run on PRT or local results to verify the change doesn't break anything?

synkd (approved):

JacobCallahan (approved):

PR #11788, opened 2023-06-30
[6.14.z] register function made more flexible

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/11783

The global registration dialog changed since this function was created, it is possible to not specify org and loc, which the register function didn't allow. The function also demanded org supplied as an object, which caused problems in onboarding more tests to it (some use org defined as a dict -- an output from cli helpers). PR changes the function in a way that doesn't require updating the existing callers

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -588,8 +588,8 @@ class ContentHost(Host, ContentHostMixins):
         using a global registration template.
 
         :param target: Satellite or Capusle object to register to, required.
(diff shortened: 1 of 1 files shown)


Satellite-QE:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k test_positive_run_job_on_host_registered_to_async_ssh_provider[rhel8]

PR #11799, opened 2023-07-03
katello_host_tools_host fixture fix

pondrejk:
The fixture now uses global registration + it plays nicely with SCA, which fixes a bunch of tests in the host component

### DIFF
--- a/pytest_fixtures/core/contenthosts.py
+++ b/pytest_fixtures/core/contenthosts.py
@@ -128,7 +128,15 @@ def registered_hosts(request, target_sat, module_org):
 def katello_host_tools_host(target_sat, module_org, rhel_contenthost):
     """Register content host to Satellite and install katello-host-tools on the host."""
     repo = settings.repos['SATCLIENT_REPO'][f'RHEL{rhel_contenthost.os_version.major}']
-    target_sat.register_host_custom_repo(module_org, rhel_contenthost, [repo])
+    ak = target_sat.api.ActivationKey(
+        content_view=module_org.default_content_view,
+        max_hosts=100,
+        organization=module_org,
+        environment=target_sat.api.LifecycleEnvironment(id=module_org.library.id),
+        auto_attach=True,
+    ).create()
+
+    rhel_contenthost.register(module_org, None, ak.name, target_sat, repo=repo)
     rhel_contenthost.install_katello_host_tools()
     yield rhel_contenthost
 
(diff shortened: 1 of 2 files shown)


pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_host.py -k test_positive_package_applicability[rhel7]

omkarkhatavkar (approved):

pytest_fixtures/core/contenthosts.py
```diff
@@ -128,7 +128,15 @@ def registered_hosts(request, target_sat, module_org):
 def katello_host_tools_host(target_sat, module_org, rhel_contenthost):
     """Register content host to Satellite and install katello-host-tools on the host."""
     repo = settings.repos['SATCLIENT_REPO'][f'RHEL{rhel_contenthost.os_version.major}']
-    target_sat.register_host_custom_repo(module_org, rhel_contenthost, [repo])
+    ak = target_sat.api.ActivationKey(
+        content_view=module_org.default_content_view,
+        max_hosts=100,
+        organization=module_org,
+        environment=target_sat.api.LifecycleEnvironment(id=module_org.library.id),
+        auto_attach=True,
```
Gauravtalreja1, line 136: Non-Blocking comment, I think it would be better to update `register_host_custom_repo` helper to use GR instead of replacing its usage here, wdyt?

jameerpathan111 (approved):

tests/foreman/cli/test_host.py
```diff
@@ -1635,6 +1636,10 @@ def restore_sca_setting():
             "subscriptions": [{"id": custom_sub.id, "quantity": 1}],
         }
     )
+    # make sure repo is enabled
+    katello_host_tools_host.enable_repo(
+        f'{module_org.name}_{prod.name}_{custom_repo.name}', force=True
+    )
```
Gauravtalreja1, line 1642: GR template adds a custom repo already in /etc/yum.repos.d/ which will be enabled by default, so this step looks redundant

Gauravtalreja1:
@jameerpathan111 We shouldn't merge PRs without T2 reviews, I had few comments though if @pondrejk finds them useful can be addressed in separate PR

tests/foreman/cli/test_host.py
Gauravtalreja1: @pondrejk yes, but this enable_repo func uses subscription-manager to enable this repos, and this custom repos with same content is already enabled from GR templates

PR #11572, opened 2023-06-01
Tweaks for the Satellite version properties

ogajduse:
This patch reverts commit 8ece887 - "Handle stream versions in version cached_property (#11467)".

## Problem statement
#11467 has updated the `version` property to return the `stream` string if the respective Satellite/Capsule is a stream release.
This breaks the compatibility of this method in most places where we use it.

Two places where #11467 broke the logic:

* `is_open` method throws `packaging.version.InvalidVersion: Invalid version: 'stream'` if the Sat release is stream.

* With 6.15 which will start supporting RHEL 9, we will need to change the parametrization of `test_positive_leapp`. There we check for `get_sat_version().minor != 11` which will change to `get_sat_version().minor != 15`.  The patch in #11571 does not help as `get_sat_version` would return `Version('9999')` which has no minor version.

Both use `get_sat_version`.

### Original motivation of #11467
When Sat Stream was 6.14, we executed the `configure-install-sat-provisioning-rhv` WF passing `sat_version=sat_object.version`, where `sat_object.version` property returned `6.14` - which is correct.
But Satlab did not know about 6.14 at that time.

## Solution

* revert #11467 and stick to the original behavior of the `version` property

* Introduce `is_stream` property to simplify the way to distinguish if the release `is_stream`. 

* Change `sat_version=sat.version` to `sat_version='stream' if sat.is_stream else sat.version`. Another solution might be to create symlink `6.Y.yml` in `playbooks/utils/satellite/` that would point to `stream.yml`. This symlink would be created once we branch Satellite from Stream. `Y` equals to `Y version of the release that was branched from Stream` + `1`.

PR #11825, opened 2023-07-07
[6.14.z] Tweaks for the Satellite version properties

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/11572

This patch reverts commit 8ece887 - "Handle stream versions in version cached_property (#11467)".

## Problem statement
#11467 has updated the `version` property to return the `stream` string if the respective Satellite/Capsule is a stream release.
This breaks the compatibility of this method in most places where we use it.

Two places where #11467 broke the logic:

* `is_open` method throws `packaging.version.InvalidVersion: Invalid version: 'stream'` if the Sat release is stream.

* With 6.15 which will start supporting RHEL 9, we will need to change the parametrization of `test_positive_leapp`. There we check for `get_sat_version().minor != 11` which will change to `get_sat_version().minor != 15`.  The patch in #11571 does not help as `get_sat_version` would return `Version('9999')` which has no minor version.

Both use `get_sat_version`.

### Original motivation of #11467
When Sat Stream was 6.14, we executed the `configure-install-sat-provisioning-rhv` WF passing `sat_version=sat_object.version`, where `sat_object.version` property returned `6.14` - which is correct.
But Satlab did not know about 6.14 at that time.

## Solution

* revert #11467 and stick to the original behavior of the `version` property

* Introduce `is_stream` property to simplify the way to distinguish if the release `is_stream`. 

* Change `sat_version=sat.version` to `sat_version='stream' if sat.is_stream else sat.version`. Another solution might be to create symlink `6.Y.yml` in `playbooks/utils/satellite/` that would point to `stream.yml`. This symlink would be created once we branch Satellite from Stream. `Y` equals to `Y version of the release that was branched from Stream` + `1`.

### DIFF
--- a/pytest_fixtures/component/provision_pxe.py
+++ b/pytest_fixtures/component/provision_pxe.py
@@ -152,12 +152,12 @@ def module_provisioning_sat(
     provisioning_domain_name = f"{gen_string('alpha').lower()}.foo"
 
     broker_data_out = Broker().execute(
-        workflow="configure-install-sat-provisioning-rhv",
-        artifacts="last",
+        workflow='configure-install-sat-provisioning-rhv',
+        artifacts='last',
         target_vlan_id=settings.provisioning.vlan_id,
         target_host=sat.name,
         provisioning_dns_zone=provisioning_domain_name,
-        sat_version=sat.version,
+        sat_version='stream' if sat.is_stream else sat.version,
     )
 
     broker_data_out = Box(**broker_data_out['data_out'])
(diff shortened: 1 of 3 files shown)


Satellite-QE:
trigger: test-robottelo
pytest: tests/foreman/api/test_role.py::test_positive_create_user_by_org_admin tests/foreman/cli/test_activationkey.py::test_create_ak_with_syspurpose_set

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/ui/test_branding.py::test_verify_satellite_login_screen_info tests/foreman/cli/test_host.py::test_positive_list_with_nested_hostgroup

Gauravtalreja1 (approved):
Ack, `tests/foreman/ui/test_branding.py::test_verify_satellite_login_screen_info` passed in PRT, and other failure look unrelated to this change.

PR #11875, opened 2023-07-13
Update ui_session to run UI tests with new admin user

Gauravtalreja1:
**Problem:** 
`ui_session` contextmanager runs tests with default admin user, and we need to select organization and select location every time we run tests, and if we don't select default organization/default location will be selected causing tests to fail.
**Solution:** 
Creating a new admin user with assigned organization/location created with common module_org/module_location fixtures, to run UI tests similar to what we've been doing with `session` fixture, using organization/location properties we can set with org/location fixtures. It'll save about ~1minute per UI test to select org and loc.

If folks like this suggestion, we can update all org/location fixtures to set organization/location property and tests which uses `ui_session` contextmanager to remove org/location select steps

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -1674,6 +1674,7 @@ class Satellite(Capsule, SatelliteMixins):
         yield
         self.omitting_credentials = False
 
+    @contextmanager
     def ui_session(self, testname=None, user=None, password=None, url=None, login=True):
         """Initialize an airgun Session object and store it as self.ui_session"""
 
@@ -1686,14 +1687,30 @@ class Satellite(Capsule, SatelliteMixins):
                 if frame.function.startswith('test_'):
                     return frame.function
 
-        return Session(
-            session_name=testname or get_caller(),
-            user=user or settings.server.admin_username,
-            password=password or settings.server.admin_password,
+        test_name = testname or get_caller()
+
+        if getattr(self, 'organization') and getattr(self, 'location') and user is None:
+            user_passwd = gen_string('alphanumeric')
+            new_user = self.api.User(
+                admin=True,
+                default_organization=self.organization,
+                default_location=self.location,
+                description=f'created automatically by airgun for module "{test_name}"',
+                login=f'{test_name}_{gen_string("alphanumeric")}',
+                password=user_passwd,
+            ).create()
+            new_user.password = user_passwd
+
+        yield Session(
+            session_name=test_name,
+            user=user or new_user.login,
+            password=password or new_user.password,
             url=url,
             hostname=self.hostname,
             login=login,
         )
+        if user is None:
+            new_user.delete()
 
     @property
     def satellite(self):
--- a/tests/foreman/ui/test_provisioningtemplate.py
+++ b/tests/foreman/ui/test_provisioningtemplate.py
@@ -60,8 +60,6 @@ def test_positive_clone(module_org, module_location, target_sat, clone_setup):
     """
     clone_name = gen_string('alpha')
     with target_sat.ui_session() as session:
-        session.organization.select(org_name=module_org.name)
-        session.location.select(loc_name=module_location.name)
         session.provisioningtemplate.clone(
             clone_setup['pt'].name,
             {
@@ -134,8 +132,6 @@ def test_positive_end_to_end(module_org, module_location, template_data, target_
         }
     ]
     with target_sat.ui_session() as session:
-        session.organization.select(org_name=module_org.name)
-        session.location.select(loc_name=module_location.name)
         session.provisioningtemplate.create(
             {
                 'template.name': name,
--- a/pytest_fixtures/component/taxonomy.py
+++ b/pytest_fixtures/component/taxonomy.py
@@ -41,7 +41,9 @@ def function_org(target_sat):
 
 @pytest.fixture(scope='module')
 def module_org(module_target_sat):
-    return module_target_sat.api.Organization().create()
+    org = module_target_sat.api.Organization().create()
+    module_target_sat.organization = org
+    yield org
 
 
 @pytest.fixture(scope='class')
@@ -53,7 +55,9 @@ def class_org(class_target_sat):
 
 @pytest.fixture(scope='module')
 def module_location(module_target_sat, module_org):
-    return module_target_sat.api.Location(organization=[module_org]).create()
+    loc = module_target_sat.api.Location(organization=[module_org]).create()
+    module_target_sat.location = loc
+    yield loc
 
 
 @pytest.fixture(scope='class')


Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/ui/test_provisioningtemplate.py

robottelo/hosts.py
```diff
@@ -1662,6 +1662,7 @@ def omit_credentials(self):
         yield
         self.omitting_credentials = False
 
+    @contextmanager
```
jyejare, line 1665: I wonder why it wasnt a context manager earlier and still worked :octocat:

robottelo/hosts.py
```diff
@@ -1674,14 +1675,31 @@ def get_caller():
                 if frame.function.startswith('test_'):
                     return frame.function
 
-        return Session(
-            session_name=testname or get_caller(),
-            user=user or settings.server.admin_username,
-            password=password or settings.server.admin_password,
+        test_name = testname or get_caller()
+
+        if getattr(self, 'organization') and getattr(self, 'location') and user is None:
```
jyejare, line 1680: Check `hasattr` function instead of getattr.

robottelo/hosts.py
```diff
@@ -1674,14 +1675,31 @@ def get_caller():
...
+            login = f'{test_name}_{gen_string("alphanumeric")}'
+            password = gen_string('alphanumeric')
+            new_user = self.api.User(
```
jyejare, line 1683: Why are we creating a new admin user but not use the default one.

The idea is good but would like to understand the reason. This will also impact many tests and those `tests` switching between admin to non-admin and again back to admin will use default admin creds to operate components.

Though this does not sound bad, but probably test does not demand this scenario always.

jyejare (changes requested):
ACK pending questions / comments.

robottelo/hosts.py
```diff
@@ -1674,14 +1675,31 @@ def get_caller():
...
                     return frame.function
 
-        return Session(
-            session_name=testname or get_caller(),
-            user=user or settings.server.admin_username,
-            password=password or settings.server.admin_password,
+        test_name = testname or get_caller()
+
+        if getattr(self, 'organization') and getattr(self, 'location') and user is None:
+            login = f'{test_name}_{gen_string("alphanumeric")}'
```
omkarkhatavkar, line 1681: you are overriding the `login=True`, try to use something different keyword e.g. `user_name`

robottelo/hosts.py
omkarkhatavkar: https://github.com/SatelliteQE/robottelo/blob/67f2a6b67a7c1301c78d09b2281e7b4b6b2afb60/pytest_fixtures/core/ui.py#L9 this getting called when we use `with session:`. The user created and which has the module_org,loc set as default makes the dashboard page is set with org and location default selected

tests/foreman/ui/test_provisioningtemplate.py
```diff
@@ -60,8 +60,6 @@ def test_positive_clone(module_org, module_location, target_sat, clone_setup):
     """
     clone_name = gen_string('alpha')
     with target_sat.ui_session() as session:
-        session.organization.select(org_name=module_org.name)
```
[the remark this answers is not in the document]
omkarkhatavkar: Does login with a new admin user have organization and location selected by default? ` Yes `

We can not set the module_org and module_location as default to admin as we use multiple xdist worker. There could be chances that we could hit the race condition

tests/foreman/ui/test_provisioningtemplate.py
```diff
@@ -134,8 +132,6 @@ def test_positive_end_to_end(module_org, module_location, template_data, target_
         }
     ]
     with target_sat.ui_session() as session:
-        session.organization.select(org_name=module_org.name)
```
omkarkhatavkar, line 137: @Gauravtalreja1 please update the tests from Puppet which has the same pattern

robottelo/hosts.py
```diff
@@ -1674,14 +1675,31 @@ def get_caller():
...
+        yield Session(
+            session_name=test_name,
+            user=user or new_user.login,
+            password=password or new_user.password,
             url=url,
             hostname=self.hostname,
             login=login,
         )
+        if user is None:
+            new_user.delete()
```
omkarkhatavkar, line 1702: creating a user and deleting that user feels cumbersome as this will run every time for each test, we have to handle this a bit better way. Maybe we can use module_user injecting as part of the host object. So that we do no need to delete that user.

robottelo/hosts.py
Gauravtalreja1: Yes, Session class itself is context manager, so it was working, but to yield its object and for handling user deletion as part teardown or using __exit__(), we need this to be contextmanager as well, I don't see any issue with having this here.
and you mentioned `handle post yield in Session class` , do you mean in airgun, can you elaborate on this ?

lhellebr (commented):
I agree with the other comments.

Especially, I'd like to know why don't we just change fixture-created user's default Org + Loc, why do we create a new user each time?

robottelo/hosts.py
Gauravtalreja1: @omkarkhatavkar is right, using default admin user could cause us troubles in case of multiple xdist workers or when running multiple tests with same user.
And for `tests switching between admin to non-admin` we don't have such tests where we switch users, but in ui_session contextmanager we've ability to pass user/password in case of non-admin user, so same way if one would like to pass default admin then they can pass too with its nailgun object.

robottelo/hosts.py
Gauravtalreja1: Good catch, will update

tests/foreman/ui/test_provisioningtemplate.py
Gauravtalreja1: yes, will be updating every test that runs UI tests with  `ui_session` to not select org/loc explicitly, along with updates to org/loc create fixtures too, when @jyejare @JacobCallahan agrees to this

robottelo/hosts.py
Gauravtalreja1: yes, I felt the same but user create/delete will be very quick with API, and even when using module_user fixtures for each module will be same thing, instead we can have some session scoped airgun_automation user created so it'll be per xdist worker, so for each test every time we need to update its organization/location, and user creation/deletion  will be once per session, wdyt?

omkarkhatavkar: yes, I felt the same but user create/delete will be very quick with API which still invalidates the point I'm making here. We should not want to create the user every time. I ack on the module user to create a fixture, that way we will be sharing that user across multiple tests with tests. I do also ack on updating the default_org and default location as part of injecting the ui_session user. (we even can use the admin user here in that case as well, if user parameter is None)

robottelo/hosts.py
jyejare: Forgot the context here, but probably I mean handle that in Session class `__exit__` function as its a context manger.

robottelo/hosts.py
jyejare: Its not yet updated! Do you wish to change ?

robottelo/hosts.py
jyejare: `when running multiple tests with same user.` This will never happen because we have devided tests with xDist workers where workers are not processes on one satellite but the different satellites. 

Now all the tests allocated to one satellite xDist worker will run all its tests serially and hence there wont be any race condition. We could have hit this in the past if thats the case.

jyejare: Why I would go with this solution is two things:
1. Testing tests with non-default admin user would be great enhancement but that also means we wont have single test with default admin that most of the customers do use.
2. Auto-selected org and loc so that wont waste our time in tests to select the org/loc explicitly. 

We have to brainstorm on these two things the pros and cons.

JacobCallahan:
@omkarkhatavkar @jyejare can you revisit the comments you had earlier?

omkarkhatavkar:
@Gauravtalreja1 Can you please update the code based on the review comments?

Gauravtalreja1:
Yes, I will go over the comments and update this PR soon.

robottelo/hosts.py
jyejare: Ping @Gauravtalreja1 !

omkarkhatavkar:
@Gauravtalreja1 This PR has been closed, If you see still need to update this. Please reopen that and work on it.

PR #11879, opened 2023-07-13
Component Audit for Discovery Plugin - 1

adarshdubey-star:
The changes mainly focuses on merging discovery rule scenarios else removing them completely as per audit.


adarshdubey-star:
This PR still requires my attention for docstrings & cleanup

adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/cli/test_discoveryrule.py

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -132,6 +132,9 @@ def test_positive_create_with_search(self, discoveryrule_factory):
         search_query = 'cpu_count = 2'
         rule = discoveryrule_factory(options={'search': search_query})
         assert rule.search == search_query
+        custom_query = 'processor = x86'
+        rule = discoveryrule_factory(options={'search': custom_query})
```
shubhamsg199, line 136: Can we move `discoveryrule_factory` to cli_factory ?

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -172,15 +175,32 @@ def test_positive_create_with_org_loc_id(
         assert class_org.name in rule.organizations
         assert class_location.name in rule.locations
 
+        new_org = Box(make_org())
+        new_loc = Box(make_location())
```
shubhamsg199, line 179:
```suggestion
        new_org = Box(target_sat.cli_factory.make_org())
        new_loc = Box(target_sat.cli_factory.make_location())
```

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -172,15 +175,32 @@ def test_positive_create_with_org_loc_id(
...
+        new_hostgroup = Box(
+            make_hostgroup({'organization-ids': new_org.id, 'location-ids': new_loc.id})
```
shubhamsg199, line 181:
```suggestion
            target_sat.cli_factory.make_hostgroup({'organization-ids': new_org.id, 'location-ids': new_loc.id})
```

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -172,15 +175,32 @@ def test_positive_create_with_org_loc_id(
...
+        )
+        target_sat.cli.DiscoveryRule.update(
+            {
+                'id': rule.id,
+                'organization-ids': new_org.id,
+                'location-ids': new_loc.id,
+                'hostgroup-id': new_hostgroup.id,
+            }
+        )
+        rule = Box(DiscoveryRule.info({'id': rule.id}))
```
shubhamsg199, line 191: same for other occurences
```suggestion
        rule = Box(target_sat.cli.DiscoveryRule.info({'id': rule.id}))
```

shubhamsg199 (commented):
Can you use cli_factory for `make_...` functions ?

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: We can but I think the factory is more of a place for generic fixtures so better to keep it here.

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: updated

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: updated

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: Tried to cover every occurence

shubhamsg199 (approved):
Ack

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -132,6 +128,9 @@ def test_positive_create_with_search(self, discoveryrule_factory):
         search_query = 'cpu_count = 2'
         rule = discoveryrule_factory(options={'search': search_query})
         assert rule.search == search_query
+        custom_query = 'processor = x86'
+        rule = discoveryrule_factory(options={'search': custom_query})
+        assert rule.search == custom_query
```
Griffin-Sullivan, line 133: Was this coverage added as an action item from the component evaluation or moved from another test? We are essentially repeating the test again just with a different query so it's a little repetitive.

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -172,15 +171,34 @@ def test_positive_create_with_org_loc_id(
...
+            {
+                'id': rule.id,
+                'organization-ids': new_org.id,
+                'location-ids': new_loc.id,
+                'hostgroup-id': new_hostgroup.id,
+            }
+        )
+        rule = Box(target_sat.cli.DiscoveryRule.info({'id': rule.id}))
+        assert new_org.name in rule.organizations
+        assert new_loc.name in rule.locations
```
Griffin-Sullivan, line 191: Can also add an assertion here for the hostgroup we updated with, else there's no reason to update with a hostgroup.

Griffin-Sullivan: Same in `test_positive_create_and_update_with_org_loc_name`

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -172,15 +171,34 @@ def test_positive_create_with_org_loc_id(
...
+
     @pytest.mark.tier2
-    def test_positive_create_with_org_loc_name(
-        self, discoveryrule_factory, class_org, class_location, class_hostgroup
+    def test_positive_create_and_update_with_org_loc_name(
```
Griffin-Sullivan, line 194: We could combine this with the test above if you wanted to. Something like using the org.id and loc.name to update and then assert for both. Non-blocking.

Griffin-Sullivan (approved):

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -29,12 +29,8 @@
 from requests import HTTPError
 
 from robottelo.cli.base import CLIReturnCodeError
-from robottelo.cli.discoveryrule import DiscoveryRule
 from robottelo.cli.factory import CLIFactoryError
```
shweta83, line 32: We are going to remove cli.factory and going to replace with cli_factory. You can use cli_factory for Error Codes here.

shweta83:
@adarshdubey-star There are some instances left with `Box(make_discoveryrule())` Can you please update those as well?

tests/foreman/cli/test_discoveryrule.py
shweta83: First search is for default queries or predefined queries and second is the custom query. It would be good to test both types.

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -240,75 +283,37 @@ def test_positive_create_disabled_rule(self, discoveryrule_factory):
...
+        self, name, host_name, discoveryrule_factory
+    ):
+        """Create Discovery Rule with negative parameters.
 
-        :id: a0350dc9-8f5b-4673-be88-a5e35d1f8ca7
+        :id: 0906cf64-ed0b-49af-844f-1af22f81ab94
 
         :expectedresults: Error should be raised and rule should not be created
 
         :CaseImportance: Medium
```
shweta83, line 298: Why `:CaseLevel: Component and  :parametrized: yes` are removed?

tests/foreman/cli/test_discoveryrule.py
Griffin-Sullivan: What is a default / predefined query?

tests/foreman/cli/test_discoveryrule.py
Gauravtalreja1: first cli_factory aren't fixtures, and second we don't have any restrictions on helper functions to keep generic, component specific helpers can be moved to it as well and in future can be used in other components too.

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -172,15 +171,34 @@ def test_positive_create_with_org_loc_id(
         assert class_org.name in rule.organizations
         assert class_location.name in rule.locations
 
+        new_org = Box(target_sat.cli_factory.make_org())
+        new_loc = Box(target_sat.cli_factory.make_location())
```
Gauravtalreja1, line 175: cli_factory methods returns Box object already
```suggestion
        new_org = target_sat.cli_factory.make_org()
        new_loc = target_sat.cli_factory.make_location()
```

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -172,15 +171,34 @@ def test_positive_create_with_org_loc_id(
...
+        )
+        target_sat.cli.DiscoveryRule.update(
+            {
+                'id': rule.id,
+                'organization-ids': new_org.id,
+                'location-ids': new_loc.id,
+                'hostgroup-id': new_hostgroup.id,
+            }
+        )
+        rule = Box(target_sat.cli.DiscoveryRule.info({'id': rule.id}))
```
Gauravtalreja1, line 189: If not output_format is passed to cli methods, then it returns just normal text, I don't think casting it Box will make any change, please check this and I'd suggest to use  csv/json as output_format instead

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -240,75 +283,37 @@ def test_positive_create_disabled_rule(self, discoveryrule_factory):
         rule = discoveryrule_factory(options={'enabled': 'false'})
         assert rule.enabled == 'false'
 
-    @pytest.mark.tier3
+    @pytest.mark.tier1
+    @pytest.mark.parametrize('host_name', **parametrized(invalid_hostnames_list()))
```
Gauravtalreja1, line 287: host_name doens't seem to be used in test, is this intended? and this will unnecessary comination of parametrized tests, please check?

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -240,75 +283,37 @@ def test_positive_create_disabled_rule(self, discoveryrule_factory):
...
-        :CaseLevel: Component
-
-        :parametrized: yes
         """
         with pytest.raises(CLIFactoryError):
             discoveryrule_factory(options={'name': name})
 
-    @pytest.mark.tier3
-    @pytest.mark.parametrize('name', **parametrized(invalid_hostnames_list()))
-    def test_negative_create_with_invalid_hostname(self, name, discoveryrule_factory):
```
Gauravtalreja1, line 263: I think it'll be better to keep this separate test for negative hostnames, instead of having new tests with parametrized cominations of invalid_hostnames_list and invalid_values_list

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -240,75 +283,37 @@ def test_positive_create_disabled_rule(self, discoveryrule_factory):
...
-
-        :CaseImportance: Medium
-        """
         with pytest.raises(CLIFactoryError):
             discoveryrule_factory(options={'hosts-limit': '9999999999'})
 
-    @pytest.mark.tier1
-    def test_negative_create_with_same_name(self, discoveryrule_factory):
-        """Create Discovery Rule with name that already exists
-
```
Gauravtalreja1, line 298: this test wasn't parametrized earlier, I don't think this needs to tested for each value from  invalid_values_list

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -318,178 +323,50 @@ def test_positive_delete(self, discoveryrule_factory):
         :CaseImportance: Critical
         """
         rule = discoveryrule_factory()
-        DiscoveryRule.delete({'id': rule.id})
+        target_sat.cli.DiscoveryRule.delete({'id': rule.id})
```
Gauravtalreja1, line 326: We're already deleting rule in `test_positive_create_with_name` as part of teardown or can have a additional crud/e2e test for it, so can we remove this test?

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -318,178 +323,50 @@ def test_positive_delete(self, discoveryrule_factory):
...
+        target_sat.cli.DiscoveryRule.update(
             {
                 'id': rule.id,
-                'organizations': new_org.name,
-                'locations': new_loc.name,
-                'hostgroup-id': new_hostgroup.id,
+                'name': new_name,
+                'search': new_query,
+                'hostgroup': new_hostgroup.name,
+                'hostname': new_hostname,
```
Gauravtalreja1, line 356: Nice, but this test still doesn't cover `test_positive_update_org_loc_by_id` and `test_positive_update_priority`, can we add that too?

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -614,20 +451,28 @@ def class_user_reader(self, class_user_password, class_org, class_location):
             logger.exception('Exception while deleting class scope user entity in teardown')
 
     @pytest.mark.tier2
-    def test_positive_create_rule_with_non_admin_user(
-        self, class_org, class_location, class_user_password, class_user_manager, class_hostgroup
+    def test_positive_create_view_and_delete_rule_with_non_admin_user(
+        self,
```
Gauravtalreja1, line 455: I think we could've some update permissions assigned to non-admin user, then we could test update for it too, and complete its crud test for non-admin user, wdyt?

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -638,90 +483,27 @@ def test_positive_create_rule_with_non_admin_user(
...
-        rule = Box(
-            DiscoveryRule.with_user(class_user_reader.login, class_user_password).info(
-                {'id': rule.id}
-            )
-        )
-        assert rule.name == rule_name
+            target_sat.cli.DiscoveryRule.info({'id': rule.id})
 
     @pytest.mark.tier2
     def test_negative_delete_rule_with_non_admin_user(
```
Gauravtalreja1, line 499: We could test negative update for this user too, when user doesn't have a proper permissions, wdyt?

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: Default query -  Sugestions you get while creating a discovery rule on satellite webui.
custom query -  Any query that you wish to add that is not part of the predefined suggestions.
Yes, @Griffin-Sullivan it is a bit redundant as it was a suggestion for sd/dev to include some custom queries in our testing.

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: Good catch!

tests/foreman/cli/test_discoveryrule.py
Griffin-Sullivan: Ok we can leave it, but really you would be more concerned about it in the UI tests then.

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: Removing

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: Removed hg as it was irrelevant

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: it is relevant and being used in the test for forming combinations of the invalid host_name

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: Might have been removed while code movement

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: Priority and org/loc are being tested individually

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: I would still keep this here

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: updated

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: I checked and its work fine

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: I think the parametrization checks more combination so good to keep it that way

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: same as above

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: Made it crud

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: That will be a new test case in which check more such things view/create i will try to come up with it in my test case adding PR's

adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/cli/test_discoveryrule.py

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: I tried to update it and the PRT is failing looks like some issue with the import , not investigated deeply,  we can do it with the original code movement anyway.

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: I have updated at someplaces

adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/cli/test_discoveryrule.py

adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/cli/test_discoveryrule.py

tests/foreman/cli/test_discoveryrule.py
shweta83: No issues.

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -115,8 +111,8 @@ def test_positive_create_with_name(self, name, discoveryrule_factory, request, t
         :parametrized: yes
         """
         rule = discoveryrule_factory(options={'name': name, 'priority': gen_int32()})
-        request.addfinalizer(target_sat.api.DiscoveryRule(id=rule.id).delete)
         assert rule.name == name
+        request.addfinalizer(target_sat.api.DiscoveryRule(id=rule.id).delete)
```
shweta83, line 116: Can you add one assertion to check if the rule is successfully deleted?

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -149,14 +148,14 @@ def test_positive_create_with_hostname(self, discoveryrule_factory):
...
+    def test_positive_create_and_update_with_org_loc_id(
+        self, discoveryrule_factory, class_org, class_location, class_hostgroup, target_sat
     ):
-        """Create discovery rule by associating org and location ids
+        """Create discovery rule by associating org and location ids and update
 
         :id: bdb4c581-d27a-4d1a-920b-89689e68a57f
 
-        :expectedresults: Rule was created and with given org & location names.
+        :expectedresults: Rule was created with given org & location names and updated.
```
shweta83, line 158:
```suggestion
        :expectedresults: Rule was created with given org & location ids and updated.
```

adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/cli/test_discoveryrule.py

adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/cli/test_discoveryrule.py

adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/cli/test_discoveryrule.py

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -520,58 +386,18 @@ def test_negative_update_name(self, name, discoveryrule_factory):
...
-            DiscoveryRule.update({'id': rule.id, 'priority': priority})
+            target_sat.cli.DiscoveryRule.update(
+                {
+                    'id': rule.id,
+                    'name': name,
+                    'hostname': '$#@!*',
+                    'hosts-limit': host_limit,
+                    'priority': priority,
+                }
+            )
```
ogajduse, line 400: Merging these four tests into one does not seem correct to me.

CLI tells us that only one parameter has an invalid value. Therefore this test will verify that CLI does not like/refuses only that specific argument.


```
[D 230801 07:00:19 hosts:99] testhost.com executing command: b'LANG=en_US.UTF-8  hammer -v -u admin -p guess --output=csv discovery-rule update --id="88" --name="" --hostname="$#@!*" --hosts-limit="bjSxIiYmwr" --priority="sqPvGlcyFf" '
[D 230801 07:00:20 hosts:101] testhost.com command result:
    stdout:
    
    stderr:
    (134, b"Could not update the rule:\n  Error: Option '--priority': Numeric value is required..\n  \n  See: 'hammer discovery-rule update --help'.\n")
    status: 64
```

I suggest keeping the tests as they were or removing them entirely.

ogajduse (changes requested):

ogajduse:
@adarshdubey-star
It would be great if you can update the following
```diff
diff --git a/tests/foreman/cli/test_discoveryrule.py b/tests/foreman/cli/test_discoveryrule.py
index 66ffc5a49..06d61c091 100644
--- a/tests/foreman/cli/test_discoveryrule.py
+++ b/tests/foreman/cli/test_discoveryrule.py
@@ -592,8 +592,9 @@ class TestDiscoveryRuleRole:
         yield user
         try:
             user.delete()
-        except HTTPError:
-            logger.exception('Exception while deleting class scope user entity in teardown')
+        except HTTPError as err:
+            logger.exception(err)
+            logger.error('Exception while deleting class scope user entity in teardown')
 
     @pytest.fixture(scope='class')
     def class_user_reader(self, class_user_password, class_org, class_location):
@@ -610,8 +611,9 @@ class TestDiscoveryRuleRole:
         yield user
         try:
             user.delete()
-        except HTTPError:
-            logger.exception('Exception while deleting class scope user entity in teardown')
+        except HTTPError as err:
+            logger.exception(err)
+            logger.error('Exception while deleting class scope user entity in teardown')
 
     @pytest.mark.tier2
     def test_positive_create_rule_with_non_admin_user(
```

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: You are correct, instead of removing I am doing some randomization here.

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -258,238 +305,57 @@ def test_negative_create_with_invalid_name(self, name, discoveryrule_factory):
...
-
-        :expectedresults: Rule should be successfully deleted
-
-        :CaseImportance: Critical
-        """
-        rule = discoveryrule_factory()
-        DiscoveryRule.delete({'id': rule.id})
-        with pytest.raises(CLIReturnCodeError):
-            DiscoveryRule.info({'id': rule.id})
-
```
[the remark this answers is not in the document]
adarshdubey-star: It's reborn.

adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/cli/test_discoveryrule.py

shweta83 (approved):
Ack.

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -240,12 +284,15 @@ def test_positive_create_disabled_rule(self, discoveryrule_factory):
...
-    @pytest.mark.tier3
+    @pytest.mark.tier1
+    @pytest.mark.parametrize('host_name', **parametrized(invalid_hostnames_list()))
     @pytest.mark.parametrize('name', **parametrized(invalid_values_list()))
-    def test_negative_create_with_invalid_name(self, name, discoveryrule_factory):
-        """Create Discovery Rule with invalid names
+    def test_negative_create_rule_with_invalid_discovery_params(
+        self, name, host_name, discoveryrule_factory
+    ):
+        """Create Discovery Rule with invalid parameters.
```
Gauravtalreja1, line 293: I don't think it's good practice to test all negative params into single test, as it could break in first with block but there are 3 more scenarios afterward which could get skipped, also there'd be unneccessary parametrization combinations due it, so would be better to keep this negative tests as it

tests/foreman/cli/test_discoveryrule.py
Gauravtalreja1: @adarshdubey-star yes, but can't those be merged with this one?

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -103,20 +99,22 @@ def _create_discoveryrule(org, loc, hostgroup, options=None):
...
+        :expectedresults: Rule should be successfully created.
 
         :CaseImportance: Critical
 
         :parametrized: yes
         """
         rule = discoveryrule_factory(options={'name': name, 'priority': gen_int32()})
-        request.addfinalizer(target_sat.api.DiscoveryRule(id=rule.id).delete)
         assert rule.name == name
+        target_sat.api.DiscoveryRule(id=rule.id).delete()
```
Gauravtalreja1, line 115: Why using API to delete it in CLI module? :)

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -520,58 +386,37 @@ def test_negative_update_name(self, name, discoveryrule_factory):
...
+                {
+                    'id': rule.id,
+                    key: params[key],
+                }
+            )
 
-        :id: 0778dd00-aa19-4062-bdf3-752e1b546ec2
+    @pytest.mark.tier1
+    def test_positive_delete(self, discoveryrule_factory, target_sat):
+        """Delete existing Discovery Rule
```
Gauravtalreja1, line 408: I'd encourage to have a teardown to delete rules in all above tests instead of separate test for it, wdyt?

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -592,8 +437,9 @@ def class_user_manager(self, class_user_password, class_org, class_location):
         yield user
         try:
             user.delete()
-        except HTTPError:
-            logger.exception('Exception while deleting class scope user entity in teardown')
+        except HTTPError as err:
+            logger.exception(err)
+            logger.error('Exception while deleting class scope user entity in teardown')
```
Gauravtalreja1, line 442: Why this block is required even? we don't have try except for other tests in robottelo for user deletion
```suggestion

```

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -610,24 +456,34 @@ def class_user_reader(self, class_user_password, class_org, class_location):
...
```
Gauravtalreja1, line 461: Same here

tests/foreman/cli/test_discoveryrule.py
```diff
@@ -610,24 +456,34 @@ def class_user_reader(self, class_user_password, class_org, class_location):
...
+        self,
+        class_org,
+        class_location,
+        class_user_password,
+        class_user_manager,
+        class_hostgroup,
+        target_sat,
     ):
-        """Create rule with non-admin user by associating discovery_manager role
+        """Create and delete rule with non-admin user by associating discovery_manager role
```
Gauravtalreja1, line 473:
```suggestion
        """Create, update and delete rule with non-admin user by associating discovery_manager role
```

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: yeah I can handle it with upcoming PR's

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: Yeah agree, but what's the issue if we have it here :)

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: same as above :)

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: Right.

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: I tried to keep it that way.

tests/foreman/cli/test_discoveryrule.py
Gauravtalreja1: just unnecessary lines :)

adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/cli/test_discoveryrule.py

tests/foreman/cli/test_discoveryrule.py
adarshdubey-star: updated

adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/cli/test_discoveryrule.py

Gauravtalreja1 (approved):

PR #11878, opened 2023-07-13
Deprecate install_katello_ca method in favor of the register method

ogajduse:
## Problem statement
* robottelo contributors keep using the deprecated method of registration in new tests that they are adding.
* as a consequence, the QE framework keeps using the `install_katello_ca` combined with `register_host` instead of using the Satellite's Global Registration feature.

## Solution
* Add a `DeprecationWarning` to the `install_katello_ca` method. Every time it is called, a deprecation warning is generated.

## Expectations
* When teams add new tests and run the test session, they will be notified about the deprecation if they call `install_katello_ca` in their test/s.
* Once we start seeing the deprecation warnings, we will be more aware of the need to port (most of) the existing tests to Global Registration.

## pytest logs
```
============================================================================================================== warnings summary ==============================================================================================================
tests/foreman/api/test_registration.py::test_positive_allow_reregistration_when_dmi_uuid_changed[rhel6]
tests/foreman/api/test_registration.py::test_positive_allow_reregistration_when_dmi_uuid_changed[rhel7]
tests/foreman/api/test_registration.py::test_positive_allow_reregistration_when_dmi_uuid_changed[rhel7_fips]
tests/foreman/api/test_registration.py::test_positive_allow_reregistration_when_dmi_uuid_changed[rhel8]
tests/foreman/api/test_registration.py::test_positive_allow_reregistration_when_dmi_uuid_changed[rhel8_fips]
tests/foreman/api/test_registration.py::test_positive_allow_reregistration_when_dmi_uuid_changed[rhel9]
tests/foreman/api/test_registration.py::test_positive_allow_reregistration_when_dmi_uuid_changed[rhel9_fips]
  /home/ogajduse/repos/robottelo/robottelo/hosts.py:508: DeprecationWarning: The install_katello_ca method is deprecated, use register_host instead.
    warnings.warn(
```

PR #11880, opened 2023-07-13
Fix the indent of `test_positive_allow_reregistration_when_dmi_uuid_changed`

ogajduse:
Seems that [Seamless component Metadata across tools](https://github.com/SatelliteQE/robottelo/pull/10863) was not that _seamless_ and we did not indent one of the tests correctly. :laughing: 

This PR is fixing it.

### Notes
Found during work on #11878


jyejare:
I think 1.5k lines of code changes can have these silly mistakes which we couldnt even caught by checks.

PR #11894, opened 2023-07-14
Ansible test fixes for selecting only the used role in the test case

adarshdubey-star:
All the roles were being imported causing the locator to be missed for `select all` roles, hence only importing the selected role used in the test case.
**Fixes:**
test_positive_create_and_delete_variable
test_positive_create_variable_with_overrides

### DIFF
--- a/tests/foreman/ui/test_ansible.py
+++ b/tests/foreman/ui/test_ansible.py
@@ -41,13 +41,15 @@ def test_positive_create_and_delete_variable(target_sat):
     :expectedresults: The variable is successfully created and deleted.
     """
     key = gen_string('alpha')
-    role = 'redhat.satellite.activation_keys'
+
+    SELECTED_ROLE = 'redhat.satellite.activation_keys'
+    proxy_id = target_sat.nailgun_smart_proxy.id
+    target_sat.api.AnsibleRoles().sync(data={'proxy_id': proxy_id, 'role_names': [SELECTED_ROLE]})
     with target_sat.ui_session() as session:
-        session.ansibleroles.import_all_roles()
         session.ansiblevariables.create(
             {
                 'key': key,
-                'ansible_role': role,
+                'ansible_role': SELECTED_ROLE,
             }
         )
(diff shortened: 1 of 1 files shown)


adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/ui/test_ansible.py -k test_positive_create_variable_with_overrides

shweta83 (approved):
Ack

shweta83:
Why this fix is not cherrypicked for 6.12.z? I see this test case in 6.12.z

synkd (approved):

Gauravtalreja1 (approved):
Changes LGTM, @adarshdubey-star It'd be helpful if you could update the PR with some information about what changes are being made for future ref, as well as update the commit as well

adarshdubey-star:
> Changes LGTM, @adarshdubey-star It'd be helpful if you could update the PR with some information about what changes are being made for future ref, as well as update the commit as well

Updated @Gauravtalreja1

tests/foreman/ui/test_ansible.py
```diff
@@ -41,13 +41,15 @@ def test_positive_create_and_delete_variable(target_sat):
     :expectedresults: The variable is successfully created and deleted.
     """
     key = gen_string('alpha')
-    role = 'redhat.satellite.activation_keys'
+
+    SELECTED_ROLE = 'redhat.satellite.activation_keys'
+    proxy_id = target_sat.nailgun_smart_proxy.id
+    target_sat.api.AnsibleRoles().sync(data={'proxy_id': proxy_id, 'role_names': [SELECTED_ROLE]})
```
Gauravtalreja1, line 47: @adarshdubey-star Okay, so as per updated description, we were testing importing roles with UI and had a locator issue so we moved to API for importing roles, so I think this isn't the way how it should be fixed, and we should test roles and variable together, but for this ansible variable tests this will be considered as part of setup, so I don't think to block the PR merge on this.

PR #11935, opened 2023-07-19
GHA for T2 only reviews

jyejare:
Now T2 Reviewers ACK is only requested when changes in the framework / core part of Robottelo are being changed by PR, hence to automate this process we are introducing the T2 reviewers GHA to:
- Assign T2 Reviewers
- Assign Framework Changes label to the PR

### DIFF
--- a/.github/workflows/t2_reviews.yaml
+++ b/.github/workflows/t2_reviews.yaml
@@ -1,0 +1,37 @@
+# Github Action to Automatically assign T2 Reviewers to framework touched PRs
+name: Auto assign T2 Reviewers
+
+on:
+  pull_request:
+    paths: ['*', '.github/**', 'conf/**', 'pytest_fixtures/core/*', 'pytest_plugins/**', 'robottelo/**', '!robottelo/constants/*', 'scripts/**']
+
+env:
+  NUMBER: ${{ github.event.number }}
+  REVIEWERS: '["robottelo-tier-2-reviewers"]'
+  LABELS: '["Framework Changes"]'
+
+jobs:
+  assign-reviewers-and-labels:
+    name: Assign T2 Reviewers and Labels to the PR
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add Labels to the PR
+        run: |
+          echo "Adding applicable labels to the PR $NUMBER"
+          curl -L -X POST \
+          -H "Accept: application/vnd.github+json" \
+          -H "Authorization: Bearer ${{ secrets.CHERRYPICK_PAT }}" \
+          -H "X-GitHub-Api-Version: 2022-11-28" \
+          https://api.github.com/repos/SatelliteQE/robottelo/issues/$NUMBER/labels \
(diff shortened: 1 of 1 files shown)


.github/workflows/t2_reviews.yaml
```diff
@@ -0,0 +1,50 @@
+# Github Action to Automatically assign T2 Reviewers to framework touched PRs
+
+name: Auto assign T2 Reviewers
+
+on:
+  pull_request:
+    types: ["opened", "synchronize", "reopened"]
+    paths:
+      - '*'
```
Griffin-Sullivan, line 9: Does this catch all of the files in the root dir?

Griffin-Sullivan (approved):

.github/workflows/t2_reviews.yaml
jyejare: Right!

.github/workflows/t2_reviews.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Assign T2 Reviewers Team to the PR
+        run:
+          echo "Assigning T2 Reviewers to the PR $number"
+          curl -L \
+          -X POST \
+          -H "Accept: application/vnd.github+json" \
+          -H "Authorization: Bearer $GITHUB_TOKEN"\
+          -H "X-GitHub-Api-Version: 2022-11-28" \
+          https://api.github.com/repos/jyejare/robottelo/pulls/$number/requested_reviewers \
```
Gauravtalreja1, line 38: Instead of running this API's endpoints to ask for reviews, can you try exploring this [auto-request-review](https://github.com/marketplace/actions/auto-request-review) action?

Gauravtalreja1: if you don't want to maintain a separate yml config for it, there's [review-request-action action](https://github.com/marketplace/actions/review-request-action) for it too, but it doesn't seem to be maintained since long

jyejare: I didnt just want to assign reviews but also add the labels and I dont want to use 2 different GHAs for simple requests, hence I choose to use a simple curl request which is always pure.

Not only this GHA but I dont find any GHA in the market which will do both for me.

omkarkhatavkar (approved):

.github/workflows/t2_reviews.yaml
Gauravtalreja1: Well, so this are two different operations, you could use two different actions for it, but there isn't any single action which does both. And this isn't matter of having it **pure** or run curl commands, actions provide different features and provides flexibility to run actions in GHA.

jyejare: Yeah ofcourse GHA provides more options but those are only useful when you need them.

Since our requirement is minimum I stayed at lowest necessary thing. We can switch to those GHAs anytime when we feel we should and need.

Also I am bit concern on how much secure does these GHAs are and looking for some mechanism to scan these GHAs for their genuine certificates.

PR #11941, opened 2023-07-20
Fix find-the-parent-prt-comment in GHA

Gauravtalreja1:
As default `direction: first` is used, `peter-evans/find-comment` action finds the first PRT comment from Parent PR, and add it on the cherrypicked PR (example PR https://github.com/SatelliteQE/robottelo/pull/11847), instead we should use last comment for PRT as per previous implementation.


omkarkhatavkar:
I think this needs to be explained to a wider audience as well as this will affect many people. I do agree here.

omkarkhatavkar (approved):

shweta83 (approved):

jameerpathan111 (approved):

PR #11962, opened 2023-07-24
[6.13.z] Sanity Plugin with new sanity tests runs on installer tests satellite

jyejare:
Cherrypick of https://github.com/SatelliteQE/robottelo/pull/11645 in 6.13.z.

### DIFF
--- a/pytest_fixtures/component/puppet.py
+++ b/pytest_fixtures/component/puppet.py
@@ -8,7 +8,10 @@ from robottelo.constants import ENVIRONMENT
 @pytest.fixture(scope='session')
 def session_puppet_enabled_sat(session_satellite_host):
     """Satellite with enabled puppet plugin"""
-    yield session_satellite_host.enable_puppet_satellite()
+    if session_satellite_host:
+        yield session_satellite_host.enable_puppet_satellite()
+    else:
+        yield
 
 
 @pytest.fixture(scope='session')
--- a/pytest_fixtures/core/sat_cap_factory.py
+++ b/pytest_fixtures/core/sat_cap_factory.py
@@ -1,49 +1,75 @@
+from contextlib import contextmanager
+
 import pytest
 from broker import Broker
 from wait_for import wait_for
 
-from pytest_fixtures.core.broker import _resolve_deploy_args
+from robottelo.config import configure_airgun
+from robottelo.config import configure_nailgun
 from robottelo.config import settings
 from robottelo.hosts import Capsule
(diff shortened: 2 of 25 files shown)


pytest_fixtures/component/puppet.py
```diff
@@ -8,7 +8,10 @@
 @pytest.fixture(scope='session')
 def session_puppet_enabled_sat(session_satellite_host):
     """Satellite with enabled puppet plugin"""
-    yield session_satellite_host.enable_puppet_satellite()
+    if session_satellite_host:
+        yield session_satellite_host.enable_puppet_satellite()
+    else:
+        yield
```
Gauravtalreja1, line 14: better to yield nothing if session_satellite_host fails here for this fixture, and test session won't start, right?
```suggestion
    if session_satellite_host:
        yield session_satellite_host.enable_puppet_satellite()
```

pytest_fixtures/core/sat_cap_factory.py
```diff
@@ -95,8 +169,10 @@ def module_capsule_configured(module_capsule_host, module_target_sat):
 @pytest.fixture(scope='session')
 def session_capsule_configured(session_capsule_host, session_target_sat):
     """Configure the capsule instance with the satellite from settings.server.hostname"""
-    session_capsule_host.capsule_setup(sat_host=session_target_sat)
-    yield session_capsule_host
+    if session_capsule_host:
+        session_capsule_host.capsule_setup(sat_host=session_target_sat)
+        yield session_capsule_host
+    yield
```
Gauravtalreja1, line 175: commit this suggestion directly, instead of separate PR :)
```suggestion
    if session_capsule_host:
        session_capsule_host.capsule_setup(sat_host=session_target_sat)
        yield session_capsule_host
```

Gauravtalreja1: @jyejare Can you check this asap as puppet capsule tests are still failing because of this?

pytest_fixtures/component/puppet.py
jyejare: Yeah, here its yielding None, and that has to be explicit with just `yield`. 

Also, this change is required, not exactly remember now why! But let it be as this does not cause any harm!

pytest_fixtures/core/sat_cap_factory.py
jyejare: Changed!

Gauravtalreja1 (approved):

PR #12009, opened 2023-07-25
Adding test for check-update when system is up to date

Griffin-Sullivan:
Automating customer case for `satellite-maintain packages check-update` when system is already up to date. We have to use yum here to update, because you can't use `satellite-maintain packages update --assumeyes` to update all packages.

### DIFF
--- a/tests/foreman/destructive/test_packages.py
+++ b/tests/foreman/destructive/test_packages.py
@@ -1,0 +1,52 @@
+"""Test class for satellite-maintain packages functionality
+
+:Requirement: foreman-maintain
+
+:CaseAutomation: Automated
+
+:CaseLevel: Component
+
+:CaseComponent: ForemanMaintain
+
+:Team: Platform
+
+:TestType: Functional
+
+:CaseImportance: Critical
+
+:Upstream: No
+"""
+import pytest
+
+pytestmark = pytest.mark.destructive
+
+
+def test_positive_all_packages_update(target_sat):
+    """Verify update and check-update work as expected.
+
+    :id: eb8a5611-b1a8-4a18-b80e-56b045c0d2f6
+
+    :steps:
+        1. Run yum update
+        2. Reboot
+        3. Run satellite-maintain packages check-update
+
+    :expectedresults: update should update the packages,
+        and check-update should list no packages at the end.
+
+    :BZ: 2218656
+
+    :customerscenario: true
+    """
+    # Register to CDN for package updates
+    target_sat.register_to_cdn()
+    # Update packages with yum
+    result = target_sat.execute('yum update -y --disableplugin=foreman-protector')
+    assert result.status == 0
+    # Reboot
+    if target_sat.execute('needs-restarting -r').status == 1:
+        target_sat.power_control(state='reboot')
+    # Run check-update again to verify there are no more packages available to update
+    result = target_sat.cli.Packages.check_update()
+    assert 'FAIL' not in result.stdout
+    assert result.status == 0


Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_packages.py::test_positive_all_packages_update

tests/foreman/destructive/test_packages.py
```diff
@@ -0,0 +1,51 @@
...
+    target_sat.register_to_cdn()
+    # Update packages with yum
+    result = target_sat.execute('yum update -y --disableplugin=foreman-protector')
+    assert result.status == 0
+    # Reboot
+    target_sat.power_control(state='reboot')
+    # Run check-update again to verify there are no more packages available to update
+    result = target_sat.cli.Packages.check_update()
+    assert 'FAIL' not in result.stdout
+    assert result.status == 0
```
Gauravtalreja1, line 51: If check-update gets any packages to update, how this is being tested as RC will be still 0 in this case?

tests/foreman/destructive/test_packages.py
```diff
@@ -0,0 +1,51 @@
...
+
+    :customerscenario: true
+    """
+    # Register to CDN for package updates
+    target_sat.register_to_cdn()
+    # Update packages with yum
+    result = target_sat.execute('yum update -y --disableplugin=foreman-protector')
+    assert result.status == 0
+    # Reboot
+    target_sat.power_control(state='reboot')
```
Gauravtalreja1, line 47:
```suggestion
    # Reboot
    if target_sat.execute('needs-restarting -r').status == 1:
        target_sat.power_control(state='reboot')
```

tests/foreman/destructive/test_packages.py
Griffin-Sullivan: Well if we updated all the packages there shouldn't be any packages to update when we run check-update

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_packages.py::test_positive_all_packages_update

tests/foreman/destructive/test_packages.py
synkd: The exit code of `dnf check-update` is 100 if updates are available. [1] Since `satellite-maintain` is just using DNF under the hood for this and passing along the exit code [2], this assertion is a valid test of whether or not updates are available. 

[1] https://dnf.readthedocs.io/en/latest/command_ref.html#check-update-command-label
[2] https://github.com/theforeman/foreman_maintain/blob/ffddc64aacec196a538c383f673a86bbe9407613/lib/foreman_maintain/package_manager/yum.rb#L70

synkd (approved):
ACK pending passing PRT results.

tests/foreman/destructive/test_packages.py
Gauravtalreja1: @synkd well, there's difference between satellite-maintain and dnf commands RC's, iirc there was some BZ around this, so even if you get update or not, check-update returns RC 0, so @Griffin-Sullivan here I think we're ensuring it wrong that no more packages available to update

synkd: @Gauravtalreja1 I just tested this on a stream Satellite, and you're right, `satellite-maintain packages check_update` returns an exit code of 0 even with updates available. This is clearly not the intended behavior, though. I'm wondering if https://github.com/theforeman/foreman_maintain/pull/732/files will address this. It looks to me like line 78 in `lib/foreman_maintain/package_manager/dnf.rb` might set the correct code.

Griffin-Sullivan: Currently testing that PR with packit on stream.

Griffin-Sullivan: Based on my testing, when you have packages available for update, it still returns status 0. We might need to do some stdout checks to see if there are packages listed I guess.

PR #12046, opened 2023-07-28
Component audit - Modify/Delete UI OS tests

shubhamsg199:
This PR aims to refactor Operating System UI tests. There were some OS tests which are combined into one end-to-end test.

Removing tests which are covered already in e2e

### DIFF
--- a/tests/foreman/ui/test_operatingsystem.py
+++ b/tests/foreman/ui/test_operatingsystem.py
@@ -17,49 +17,13 @@
 :Upstream: No
 """
 import pytest
-from nailgun import entities
 
-from robottelo.constants import DEFAULT_TEMPLATE
 from robottelo.constants import HASH_TYPE
 from robottelo.utils.datafactory import gen_string
 
 
-@pytest.fixture(scope='module')
-def module_org():
-    return entities.Organization().create()
-
-
 @pytest.mark.tier2
-def test_positive_update_with_params(session):
-    """Set Operating System parameter
-
-    :id: 05b504d8-2518-4359-a53a-f577339f1ebe
-
-    :expectedresults: OS is updated with new parameter
-
-    :CaseLevel: Integration
-    """
-    name = gen_string('alpha')
-    major_version = gen_string('numeric', 2)
-    param_name = gen_string('alpha')
-    param_value = gen_string('alpha')
-    with session:
-        session.operatingsystem.create(
-            {'operating_system.name': name, 'operating_system.major': major_version}
-        )
-        os_full_name = f'{name} {major_version}'
-        assert session.operatingsystem.search(name)[0]['Title'] == os_full_name
-        session.operatingsystem.update(
-            os_full_name, {'parameters.os_params': {'name': param_name, 'value': param_value}}
-        )
-        values = session.operatingsystem.read(os_full_name)
-        assert len(values['parameters']) == 1
-        assert values['parameters']['os_params'][0]['name'] == param_name
-        assert values['parameters']['os_params'][0]['value'] == param_value
-
-
-@pytest.mark.tier2
-def test_positive_end_to_end(session):
+def test_positive_end_to_end(session, module_org, module_location, target_sat):
     """Create all possible entities that required for operating system and then
     test all scenarios like create/read/update/delete for it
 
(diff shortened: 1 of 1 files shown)


shubhamsg199:
trigger: test-robottelo
pytest: tests/foreman/ui/test_operatingsystem.py

tests/foreman/ui/test_operatingsystem.py
```diff
@@ -118,49 +81,71 @@ def test_positive_end_to_end(session):
...
+        new_family = 'Red Hat CoreOS'
+        new_hash = HASH_TYPE['sha512']
+        new_architecture = target_sat.api.Architecture().create()
+        new_ptable = target_sat.api.PartitionTable(
+            organization=[module_org], location=[module_location], os_family='Redhat'
+        ).create()
+        new_medium = target_sat.api.Media(
+            organization=[module_org], location=[module_location]
+        ).create()
+        new_param_name = gen_string('alpha')
```
Gauravtalreja1, line 98: Can this all new_* vars stuff can be moved to top of test, so test can be better readable? non-blocking

tests/foreman/ui/test_operatingsystem.py
```diff
@@ -118,49 +81,71 @@ def test_positive_end_to_end(session):
...
+                'template.name': template_name,
+                'template.default': True,
+                'type.snippet': False,
+                'template.template_editor.editor': gen_string('alpha'),
+                'type.template_type': 'Provisioning template',
+                'association.applicable_os.assigned': [os['operating_system']['description']],
+                'organizations.resources.assigned': [module_org.name],
+                'locations.resources.assigned': [module_location.name],
+            }
+        )
```
Gauravtalreja1, line 111: I think this as part of setup, so can this be done with API or you want to cover OS assignment on provisioning templates?

tests/foreman/ui/test_operatingsystem.py
```diff
@@ -118,49 +81,71 @@ def test_positive_end_to_end(session):
...
+                'templates.resources': {'Provisioning template': template_name},
+                'operating_system.description': new_description,
+                'operating_system.major': new_major_version,
+                'operating_system.minor': new_minor_version,
+                'operating_system.family': new_family,
+                'operating_system.password_hash': new_hash,
+                'operating_system.architectures.assigned': [new_architecture.name],
+                'partition_table.resources.assigned': [new_ptable.name],
+                'installation_media.resources.assigned': [new_medium.name],
+                'parameters.os_params': {'name': new_param_name, 'value': new_param_value},
```
Gauravtalreja1, line 126: I think in update we should cover updating old `param_name`'s value, instead of adding new param, right?

tests/foreman/ui/test_operatingsystem.py
```diff
@@ -176,7 +161,7 @@ def test_positive_verify_os_name(session):
     name = gen_string('alpha')
     major_version = gen_string('numeric', 2)
     os_full_name = f"{name} {major_version}"
-    entities.OperatingSystem(name=name, major=major_version).create()
+    target_sat.api.OperatingSystem(name=name, major=major_version).create()
     with session:
         values = session.operatingsystem.search(os_full_name)
         assert values[0]['Title'] == os_full_name
```
Gauravtalreja1, line 167: Can this test to search w/ full name also be combined with above test?

tests/foreman/ui/test_operatingsystem.py
```diff
@@ -118,49 +81,71 @@ def test_positive_end_to_end(session):
         assert len(os['parameters']['os_params']) == 1
         assert os['parameters']['os_params'][0]['name'] == param_name
         assert os['parameters']['os_params'][0]['value'] == param_value
+        template_name = gen_string('alpha')
```
Gauravtalreja1, line 84: I think OS created will have DEFAULT_TEMPLATE assigned as per old tests, so can this be asserted after create? and then update template assigned?

tests/foreman/ui/test_operatingsystem.py
shubhamsg199: Yes, but I'll prefer to keep it as while creating OS here we only pass name and major version so the name parameter comes out as `f"{name} {major_version}"`  but in the end-to-end test we pass all the fields  including descriptionso it's different

tests/foreman/ui/test_operatingsystem.py
shubhamsg199: Yes, this can be done but creating provisioning template via API needs a nailgun OS object. I can do so by reading OS via `target_sat.api.ProvisioningTemplate(id=os.id).read()`  but we need `id` which is not available via UI.
Hence, I thought of creating via UI itself

tests/foreman/ui/test_operatingsystem.py
shubhamsg199: We can, but I'll prefer to keep it the same way as this way code looks more structured

tests/foreman/ui/test_operatingsystem.py
shubhamsg199: Updated

tests/foreman/ui/test_operatingsystem.py
shubhamsg199: Updated

adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/ui/test_operatingsystem.py

adarshdubey-star (approved):
Nice refactor.

ColeHiggins2 (approved):
Looks good to me!

Gauravtalreja1 (approved):

PR #12077, opened 2023-08-01
Change certs fixture to use sat_ready_rhel fixture

lpramuk:
Update fixture implementation to OSP in order to get correct flavor for the test `satqe-ssd.standard.std` (instead of `satqe-ssd.memory.xs`)

Marking the test that uses the fixture as non-destructive (just installs satellite on `sat_ready_rhel`)
(execution time dropped from 43min to 33min)

### DIFF
--- a/pytest_fixtures/component/katello_certs_check.py
+++ b/pytest_fixtures/component/katello_certs_check.py
@@ -2,29 +2,18 @@
 from pathlib import Path
 
 import pytest
-from broker import Broker
 from fauxfactory import gen_string
 
 from robottelo.constants import CERT_DATA as cert_data
 from robottelo.hosts import Capsule
-from robottelo.hosts import ContentHost
 
 
 @pytest.fixture
-def certs_vm_setup(request):
-    """Create VM and register content host"""
-    target_cores = request.param.get('target_cores', 1)
-    target_memory = request.param.get('target_memory', '1GiB')
-    with Broker(
-        nick=request.param['nick'],
-        host_class=ContentHost,
-        target_cores=target_cores,
-        target_memory=target_memory,
-    ) as host:
-        cert_data['key_file_name'] = f'{host.hostname}/{host.hostname}.key'
-        cert_data['cert_file_name'] = f'{host.hostname}/{host.hostname}.crt'
-        host.custom_cert_generate(cert_data['capsule_hostname'])
-        yield cert_data, host
+def certs_data(sat_ready_rhel):
+    cert_data['key_file_name'] = f'{sat_ready_rhel.hostname}/{sat_ready_rhel.hostname}.key'
+    cert_data['cert_file_name'] = f'{sat_ready_rhel.hostname}/{sat_ready_rhel.hostname}.crt'
+    sat_ready_rhel.custom_cert_generate(cert_data['capsule_hostname'])
+    yield cert_data
 
 
 @pytest.fixture
--- a/tests/foreman/destructive/test_katello_certs_check.py
+++ b/tests/foreman/destructive/test_katello_certs_check.py
@@ -21,8 +21,6 @@ import re
 
 import pytest
 
-import robottelo.constants as constants
-from robottelo.config import settings
 from robottelo.utils.issue_handlers import is_open
 
 pytestmark = pytest.mark.destructive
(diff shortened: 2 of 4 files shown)


lpramuk:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_katello_certs_check.py::test_positive_install_sat_with_katello_certs

pytest_fixtures/component/katello_certs_check.py
```diff
@@ -2,29 +2,19 @@
...
-    ) as host:
-        cert_data['key_file_name'] = f'{host.hostname}/{host.hostname}.key'
-        cert_data['cert_file_name'] = f'{host.hostname}/{host.hostname}.crt'
-        host.custom_cert_generate(cert_data['capsule_hostname'])
-        yield cert_data, host
+def certs_setup(sat_ready_rhel):
+    host = sat_ready_rhel
+    cert_data['key_file_name'] = f'{host.hostname}/{host.hostname}.key'
+    cert_data['cert_file_name'] = f'{host.hostname}/{host.hostname}.crt'
+    host.custom_cert_generate(cert_data['capsule_hostname'])
```
Gauravtalreja1, line 16: just out of curiosity, why we're generating certs for capsule `capsule.example.com` here? 
if there's specific reason, instead can't we generate it for internal capsule, then we could add sat_ready_rhel obj to cert_data dict and return it as Box obj, so we could access it in test w/o calling sat_ready_rhel, wdyt?

tests/foreman/destructive/test_katello_certs_check.py
```diff
@@ -105,7 +97,8 @@ def test_positive_install_sat_with_katello_certs(certs_vm_setup):
 
     :CaseAutomation: Automated
     """
-    cert_data, rhel_vm = certs_vm_setup
+    cert_data = certs_setup
+    rhel_vm = sat_ready_rhel
     version = rhel_vm.os_version.major
     rhel_vm.download_repofile(product='satellite', release=settings.server.version.release)
     rhel_vm.register_contenthost(
```
Gauravtalreja1, line 104: Can we update it to use `rhel_vm .register_to_cdn()` instead of running 2 functions, register_contenthost and subscription_manager_attach_pool? when el8 repos will be auto-attached so we could also remove enable_repo step. Non-blocking, as its something outside of scope of the PR :)

Gauravtalreja1 (commented):
Ack, with a suggestion

pytest_fixtures/component/katello_certs_check.py
lpramuk: IDK

tests/foreman/destructive/test_katello_certs_check.py
lpramuk: Updated.

lpramuk:
trigger: test-robottelo
pytest: tests/foreman/sys/test_katello_certs_check.py::test_positive_install_sat_with_katello_certs

pytest_fixtures/component/katello_certs_check.py
Gauravtalreja1: @omkarkhatavkar As you owned this component earlier, do you've any insights on this?

ogajduse (approved):
PRT passed. 
Good work @lpramuk!

tests/foreman/sys/test_katello_certs_check.py
```diff
@@ -21,6 +21,54 @@
...
+
+    :expectedresults: Satellite should be installed using the custom certs.
+
+    :CaseAutomation: Automated
+    """
+    version = sat_ready_rhel.os_version.major
+    sat_ready_rhel.download_repofile(product='satellite', release=settings.server.version.release)
+    sat_ready_rhel.register_to_cdn()
+    for repo in getattr(constants, f"OHSNAP_RHEL{version}_REPOS"):
+        sat_ready_rhel.enable_repo(repo, force=True)
```
Gauravtalreja1, line 50: As register_to_cdn auto-attaches the baseos and appstream repos, so we might not need this 
```suggestion

```

tests/foreman/sys/test_katello_certs_check.py
```diff
@@ -21,6 +21,54 @@
...
+    sat_ready_rhel.execute('yum -y update')
```
Gauravtalreja1, line 51:
```suggestion
    sat_ready_rhel.execute('dnf -y update')
```

tests/foreman/sys/test_katello_certs_check.py
```diff
@@ -21,6 +21,54 @@
...
+    result = sat_ready_rhel.execute(
+        'dnf -y module enable satellite:el8 && dnf -y install satellite'
+    )
+    assert result.status == 0
+    command = (
+        'satellite-installer --scenario satellite '
+        f'--certs-server-cert "/root/{certs_data["cert_file_name"]}" '
+        f'--certs-server-key "/root/{certs_data["key_file_name"]}" '
+        f'--certs-server-ca-cert "/root/{certs_data["ca_bundle_file_name"]}" '
```
Gauravtalreja1, line 60: Can we use InstallerCommand here and sat_ready_rhel.install(..)?

tests/foreman/sys/test_katello_certs_check.py
```diff
@@ -21,6 +21,54 @@
...
+        2. Install satellite with custom certs
+        3. Assert output does not report SSL certificate error
+        4. Assert all services are running
+
+
+    :expectedresults: Satellite should be installed using the custom certs.
+
+    :CaseAutomation: Automated
+    """
+    version = sat_ready_rhel.os_version.major
```
Gauravtalreja1, line 46: It'd be nice, if we make it Satellite object then we could use this object to run InstallerCommand and robottelo CLI's with it, wdyt?
```suggestion
    version = sat_ready_rhel.os_version.major
    sat_ready_rhel = Satellite(sat_ready_rhel.hostname)
```

tests/foreman/sys/test_katello_certs_check.py
```diff
@@ -21,6 +21,54 @@
...
+        f'--certs-server-cert "/root/{certs_data["cert_file_name"]}" '
+        f'--certs-server-key "/root/{certs_data["key_file_name"]}" '
+        f'--certs-server-ca-cert "/root/{certs_data["ca_bundle_file_name"]}" '
+    )
+    result = sat_ready_rhel.execute(command, timeout=2200000)
+    assert result.status == 0
+    # assert no hammer ping SSL cert error
+    result = sat_ready_rhel.execute('hammer ping')
+    assert 'SSL certificate verification failed' not in result.stdout
+    assert result.stdout.count('ok') == 8
```
Gauravtalreja1, line 67: I think it is not a good idea to hard code count of ok's, better to use approch to count this similar to cli/test_ping.py

tests/foreman/sys/test_katello_certs_check.py
lpramuk: Updated.

tests/foreman/sys/test_katello_certs_check.py
lpramuk: Enhanced check to match `Status:` and `ok` counts

tests/foreman/sys/test_katello_certs_check.py
lpramuk: Updated to use `InstallerCommand` 
but not `install`
```
        return self.execute(installer_obj.get_command(), timeout=0)
```
timeout=0     vs.    timeout=220000

lpramuk: Further updates.
Now using `install(InstallerCommand())` combo :-)

lpramuk:
trigger: test-robottelo
pytest: tests/foreman/sys/test_katello_certs_check.py::test_positive_install_sat_with_katello_certs

tests/foreman/sys/test_katello_certs_check.py
lpramuk: Updated.

tests/foreman/sys/test_katello_certs_check.py
lpramuk: Reverting back to using just `InstallerCommand()`
By using `install()` where timeout is set to `0` the command has timed out after 10min

0 doesn't mean unlimited as is wrongly stated on many places in the code

lpramuk: @JacobCallahan ^ fyi

ogajduse: Then it seems that
```
By default or if timeout set is zero, blocking calls do not time out. :param timeout: Milliseconds to wait before timeout.
```
statement from ssh2-python docs does not hold true.
https://ssh2-python.readthedocs.io/en/latest/session.html?highlight=Session#ssh2.session.Session.set_timeout

tests/foreman/sys/test_katello_certs_check.py
Gauravtalreja1: @lpramuk this still doesn't seem to be updated, are you sure?

ogajduse: @Gauravtalreja1 what would you like to make a Satellite object? `sat_ready_rhel` is a Satellite object already.

Gauravtalreja1 (approved):

PR #12128, opened 2023-08-04
Registration test dmi_uuid fix

shweta83:
This change updates test_positive_allow_reregistration_when_dmi_uuid_changed to use global registration method in-place of old method of registration.
Also BZ#2229112 is added and this test case will get skipped until the BZ is fixed.

### DIFF
--- a/tests/foreman/api/test_registration.py
+++ b/tests/foreman/api/test_registration.py
@@ -21,7 +21,6 @@ import uuid
 import pytest
 
 from robottelo.constants import CLIENT_PORT
-from robottelo.constants import ENVIRONMENT
 
 pytestmark = pytest.mark.tier1
 
@@ -89,8 +88,10 @@ def test_host_registration_end_to_end(
 
 
 @pytest.mark.tier3
+@pytest.mark.rhel_ver_match('[^6]')
+@pytest.mark.skip_if_open("BZ:2229112")
 def test_positive_allow_reregistration_when_dmi_uuid_changed(
-    module_org, rhel_contenthost, target_sat
+    module_org, rhel_contenthost, target_sat, module_ak_with_synced_repo, module_location
 ):
     """Register a content host with a custom DMI UUID, unregistering it, change
     the DMI UUID, and re-registering it again
@@ -101,18 +102,27 @@ def test_positive_allow_reregistration_when_dmi_uuid_changed(
 
     :customerscenario: true
 
-    :BZ: 1747177
+    :BZ: 1747177,2229112
 
     :CaseLevel: Integration
     """
     uuid_1 = str(uuid.uuid1())
     uuid_2 = str(uuid.uuid4())
-    rhel_contenthost.install_katello_ca(target_sat)
     target_sat.execute(f'echo \'{{"dmi.system.uuid": "{uuid_1}"}}\' > /etc/rhsm/facts/uuid.facts')
-    result = rhel_contenthost.register_contenthost(module_org.label, lce=ENVIRONMENT)
+    command = target_sat.api.RegistrationCommand(
+        organization=module_org,
+        activation_keys=[module_ak_with_synced_repo.name],
+        location=module_location,
+    ).create()
+    result = rhel_contenthost.execute(command)
     assert result.status == 0
     result = rhel_contenthost.execute('subscription-manager clean')
     assert result.status == 0
     target_sat.execute(f'echo \'{{"dmi.system.uuid": "{uuid_2}"}}\' > /etc/rhsm/facts/uuid.facts')
-    result = rhel_contenthost.register_contenthost(module_org.label, lce=ENVIRONMENT)
+    command = target_sat.api.RegistrationCommand(
+        organization=module_org,
+        activation_keys=[module_ak_with_synced_repo.name],
+        location=module_location,
+    ).create()
+    result = rhel_contenthost.execute(command)
     assert result.status == 0


shweta83:
trigger: test-robottelo
pytest: tests/foreman/api/test_registration.py -k test_positive_allow_reregistration_when_dmi_uuid_changed

shweta83:
trigger: test-robottelo
pytest: tests/foreman/api/test_registration.py -k test_positive_allow_reregistration_when_dmi_uuid_changed

tests/foreman/api/test_registration.py
```diff
@@ -107,12 +107,22 @@ def test_positive_allow_reregistration_when_dmi_uuid_changed(
...
+    rhel_contenthost.unregister()
     target_sat.execute(f'echo \'{{"dmi.system.uuid": "{uuid_2}"}}\' > /etc/rhsm/facts/uuid.facts')
-    result = rhel_contenthost.register_contenthost(module_org.label, lce=ENVIRONMENT)
+    command = target_sat.api.RegistrationCommand(
+        organization=module_org,
+        activation_keys=[module_ak_with_synced_repo.name],
+        location=module_location,
+        force=True,
+    ).create()
+    rhel_contenthost.execute(command)
```
shubhamsg199, line 126:
```suggestion
```

Gauravtalreja1: nice catch, it is duplicated line

tests/foreman/api/test_registration.py
```diff
@@ -107,12 +107,22 @@ def test_positive_allow_reregistration_when_dmi_uuid_changed(
...
-    assert result.status == 0
-    result = rhel_contenthost.execute('subscription-manager clean')
+    command = target_sat.api.RegistrationCommand(
+        organization=module_org,
+        activation_keys=[module_ak_with_synced_repo.name],
+        location=module_location,
+    ).create()
+    result = rhel_contenthost.execute(command)
     assert result.status == 0
+    rhel_contenthost.unregister()
```
Gauravtalreja1, line 118: Instead of unregister, can't we force register in next iteration?

tests/foreman/api/test_registration.py
```diff
@@ -88,9 +87,10 @@ def test_host_registration_end_to_end(
     assert CLIENT_PORT == rhel_contenthost.subscription_config['server']['port']
 
 
+@pytest.mark.no_containers
 @pytest.mark.tier3
 def test_positive_allow_reregistration_when_dmi_uuid_changed(
-    module_org, rhel_contenthost, target_sat
+    module_org, rhel_contenthost, target_sat, module_ak_with_synced_repo, module_location
```
Gauravtalreja1, line 93: Can we use module_target_sat instead?

tests/foreman/api/test_registration.py
ogajduse: I would advocate against force unregister in any kind of automation. In my view,  `force` options are to be used by people not programs.

tests/foreman/api/test_registration.py
ogajduse: What difference do you see in module_target_sat and target_sat?
If I am not missing something, they are pointing to the same Satellite object in the end.

tests/foreman/api/test_registration.py
```diff
@@ -107,12 +107,22 @@ def test_positive_allow_reregistration_when_dmi_uuid_changed(
     """
     uuid_1 = str(uuid.uuid1())
     uuid_2 = str(uuid.uuid4())
-    rhel_contenthost.install_katello_ca(target_sat)
     target_sat.execute(f'echo \'{{"dmi.system.uuid": "{uuid_1}"}}\' > /etc/rhsm/facts/uuid.facts')
-    result = rhel_contenthost.register_contenthost(module_org.label, lce=ENVIRONMENT)
-    assert result.status == 0
-    result = rhel_contenthost.execute('subscription-manager clean')
```
ogajduse, line 114: Please do not change the core step of the test!
I recommend reading thoroughly the BZ associated with test cases that you are changing to ensure you are not destroying the test logic.
See https://bugzilla.redhat.com/show_bug.cgi?id=1747177#c19

Please change `unregister` back to `clean`

tests/foreman/api/test_registration.py
```diff
@@ -107,12 +107,22 @@ def test_positive_allow_reregistration_when_dmi_uuid_changed(
...
+    result = rhel_contenthost.execute(command)
     assert result.status == 0
+    rhel_contenthost.unregister()
     target_sat.execute(f'echo \'{{"dmi.system.uuid": "{uuid_2}"}}\' > /etc/rhsm/facts/uuid.facts')
-    result = rhel_contenthost.register_contenthost(module_org.label, lce=ENVIRONMENT)
+    command = target_sat.api.RegistrationCommand(
+        organization=module_org,
+        activation_keys=[module_ak_with_synced_repo.name],
+        location=module_location,
+        force=True,
```
ogajduse, line 124: Previous implementation of the test did not `force` registration? Why do we need to force registration here?

tests/foreman/api/test_registration.py
```diff
@@ -88,9 +87,10 @@ def test_host_registration_end_to_end(
     assert CLIENT_PORT == rhel_contenthost.subscription_config['server']['port']
 
 
+@pytest.mark.no_containers
```
ogajduse, line 90: Could you please justify the usage of a VM? There is a significant test time increase if we use VMs.
Why is container not a good fit for this testcase?

ogajduse (changes requested):
NACK until I know what is the reason for using VM, force registration and until `subscription-manager clean` is not present in the test instead of `unregister`.

@shweta83 I believe that you could have avoided many questions if you would mention the reason why do you do such changes in the PR description.

tests/foreman/api/test_registration.py
shweta83: This was a workaround for the issue but we need to skip the test case until this is fixed. I will add the skip marker.

tests/foreman/api/test_registration.py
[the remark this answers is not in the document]
shweta83: It doesn't need to run for EL6. I will update it

tests/foreman/api/test_registration.py
shweta83: I agree with @ogajduse. We don't need to force register it. It will be better to unregister it first and then register the host after changing dmi_uuid.

tests/foreman/api/test_registration.py
shweta83: I would prefer `target_sat` here as we are changing dmi_uuid of the satellite and also doesn't make difference to use `module_target_sat` here.

tests/foreman/api/test_registration.py
shweta83: Force was not needed. Removed it.

tests/foreman/api/test_registration.py
shweta83: Thanks for pointing this. I missed that comment on the BZ. Changed it to `subscription-manager clean`.

tests/foreman/api/test_registration.py
shweta83: This was not intended. I added it for testing something else. Removed it.

shweta83:
trigger: test-robottelo
pytest: tests/foreman/api/test_registration.py -k test_positive_allow_reregistration_when_dmi_uuid_changed

tests/foreman/api/test_registration.py
Gauravtalreja1: Other tests in this module are using module_target_sat, so ask was just to keep use of it consistent

tests/foreman/api/test_registration.py
```diff
@@ -101,18 +102,27 @@ def test_positive_allow_reregistration_when_dmi_uuid_changed(
 
     :customerscenario: true
 
-    :BZ: 1747177
+    :BZ: 1747177,2229112
```
Gauravtalreja1, line 105:
```suggestion
    :BZ: 1747177, 2229112
```

tests/foreman/api/test_registration.py
```diff
@@ -101,18 +102,27 @@ def test_positive_allow_reregistration_when_dmi_uuid_changed(
...
     assert result.status == 0
     target_sat.execute(f'echo \'{{"dmi.system.uuid": "{uuid_2}"}}\' > /etc/rhsm/facts/uuid.facts')
-    result = rhel_contenthost.register_contenthost(module_org.label, lce=ENVIRONMENT)
+    command = target_sat.api.RegistrationCommand(
+        organization=module_org,
+        activation_keys=[module_ak_with_synced_repo.name],
+        location=module_location,
+    ).create()
+    result = rhel_contenthost.execute(command)
     assert result.status == 0
```
Gauravtalreja1, line 128:
```suggestion
    assert  rhel_contenthost.execute(command).status == 0
```

tests/foreman/api/test_registration.py
```diff
@@ -101,18 +102,27 @@ def test_positive_allow_reregistration_when_dmi_uuid_changed(
...
-    rhel_contenthost.install_katello_ca(target_sat)
     target_sat.execute(f'echo \'{{"dmi.system.uuid": "{uuid_1}"}}\' > /etc/rhsm/facts/uuid.facts')
-    result = rhel_contenthost.register_contenthost(module_org.label, lce=ENVIRONMENT)
+    command = target_sat.api.RegistrationCommand(
+        organization=module_org,
+        activation_keys=[module_ak_with_synced_repo.name],
+        location=module_location,
+    ).create()
+    result = rhel_contenthost.execute(command)
     assert result.status == 0
```
Gauravtalreja1, line 118:
```suggestion
    assert rhel_contenthost.execute(command).status == 0
```

tests/foreman/api/test_registration.py
```diff
@@ -89,8 +88,10 @@ def test_host_registration_end_to_end(
 
 
 @pytest.mark.tier3
+@pytest.mark.rhel_ver_match('[^6]')
+@pytest.mark.skip_if_open("BZ:2229112")
```
Gauravtalreja1, line 92: Can we keep the workaround of this BZ to register twice with condition of `is_open('BZ:2229112')` instead of skipping a test, wdyt?

tests/foreman/api/test_registration.py
Gauravtalreja1: @ogajduse force param is part of registration form, so this is intended for testing by programs, and either we use clean or force both do the same thing for host :)
but I agree with @shweta83 if it isn't required here, as we need to register after changeing uuid, so it makes sense.

tests/foreman/api/test_registration.py
ogajduse: I am a nitpicker, but this one feels like a **real** nitpick, Gaurav.  :laughing:

ogajduse (approved):
:+1:

tests/foreman/api/test_registration.py
shweta83: We should skip this until the issue gets fixed. It would be easier to fix I guess.

tests/foreman/api/test_registration.py
shweta83: Skipping this.

tests/foreman/api/test_registration.py
shweta83: Since it's non-blocking so skipping this.

tests/foreman/api/test_registration.py
shweta83: Since this is Non-blocking so skipping this.

Gauravtalreja1 (approved):

PR #12152, opened 2023-08-07
[6.14.z] Change certs fixture to use sat_ready_rhel fixture

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/12077

Update fixture implementation to OSP in order to get correct flavor for the test `satqe-ssd.standard.std` (instead of `satqe-ssd.memory.xs`)

Marking the test that uses the fixture as non-destructive (just installs satellite on `sat_ready_rhel`)
(execution time dropped from 43min to 33min)

### DIFF
--- a/pytest_fixtures/component/katello_certs_check.py
+++ b/pytest_fixtures/component/katello_certs_check.py
@@ -2,29 +2,18 @@
 from pathlib import Path
 
 import pytest
-from broker import Broker
 from fauxfactory import gen_string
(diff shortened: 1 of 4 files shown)


Satellite-QE:
trigger: test-robottelo
pytest: tests/foreman/sys/test_katello_certs_check.py::test_positive_install_sat_with_katello_certs

PR #12176, opened 2023-08-08
Fix unreliable assertion

sambible:
This simply removes an assert from the test. This portion of the virtualization card doesn't reliably appear on the card, and can cause this test to be flaky.


synkd (commented):
Is there a BZ associated with the inconsistency of displaying the IP address? I could see that issue causing customer frustration.

sambible:
Hmm, not sure. I need to investigate further, as I'm not sure whether this is something we should expect or not. The Virtualization card is dynamically generated based on what the hypervisor returns to Satellite, so this might be a case where it was appropriate to not send this information

synkd:
> Hmm, not sure. I need to investigate further, as I'm not sure whether this is something we should expect or not. The Virtualization card is dynamically generated based on what the hypervisor returns to Satellite, so this might be a case where it was appropriate to not send this information

Yeah, if `public` means "public to the Internet," then I wouldn't expect every VM to have one, but as a user I would expect to see an IP address in the VM info provided on the card. Looking at the code for the VMware card [1], I see that public_ip_address is the only IP address info field on the card, while some of the other provider cards have a private IP address field as well. But I don't know whether there are limitations on the VMware side. Either way, some clarification from dev would be helpful here. 

[1] https://github.com/theforeman/foreman/blob/cc68b5399efb66aa19e9ec487aa6ac18420479a9/webpack/assets/javascripts/react_app/components/HostDetails/Tabs/Details/Cards/Virtualization/VirtVmware.js

sambible:
trigger: test-robottelo

shubhamsg199:
@sambible I checked our VMware instance and the VM which you are importing was `powered off`. I believe that's the reason we don't see `public ip `.  I've `powered on` that VM. Can you check now if the IP is displayed or not?

sambible:
trigger: test-robottelo

sambible:
trigger: test-robottelo

sambible:
Going to close this PR, as this isn't going to be the correct fix for the issue.

sambible:
Finally figured it out, turns out I had a bad assert statement, that would only work if the field was present with a value, but if it didn't have a value, it would fail. Going to change this to simply look for the presence of the key in the dict, which is a cleaner fix for the issue.

sambible:
trigger: test-robottelo
airgun: 933

sambible:
trigger: test-robottelo
airgun: 933

sambible:
trigger: test-robottelo
airgun: 933

sambible:
trigger: test-robottelo
airgun: 933

sambible:
trigger: test-robottelo
airgun: 933

sambible:
trigger: test-robottelo
airgun: 933

sambible:
trigger: test-robottelo
airgun: 933

adarshdubey-star:
Hey @sambible I would suggest including a snippet to check if the machine is powered on as I was running into an issue that the IP is not displayed on the virt_card, somehow the machine gets turned off resulting in the diminishing of the IP.
You can refer this PR - https://github.com/SatelliteQE/robottelo/pull/12295

tests/foreman/ui/test_computeresource_vmware.py
```diff
@@ -593,18 +593,21 @@ def test_positive_virt_card(
...
         assert virt_card['cpus'] == '1'
+        assert virt_card['disk_label'] == 'Hard disk 1'
+        assert 'disk_capacity' in virt_card
+        assert 'partition_capacity' in virt_card
+        assert virt_card['partition_path'] == '/boot'
+        assert 'partition_allocation' in virt_card
         assert virt_card['cores_per_socket'] == '1'
         assert virt_card['firmware'] == 'bios'
-        assert virt_card['hypervisor'] == module_vmware_settings['hypervisor']
+        assert 'hypervisor' in virt_card
```
adarshdubey-star, line 610: we do have hypervisor_name in the vmware.yaml to assert unless the virt_card is giving something else here?

sambible: I added that as part of this test, thinking it would be pretty static. It's already changed 2 or 3 times since I added it, so I'm not sure this will be a reliable thing to assert on.

tests/foreman/ui/test_computeresource_vmware.py
```diff
@@ -593,18 +593,21 @@ def test_positive_virt_card(
...
         host_name = module_vmware_settings['vm_name'] + '.' + domain.name
-        virt_card = session.host_new.get_details(host_name, widget_names='details.virtualization')[
-            'details'
-        ]['virtualization']
+        virt_card = session.host_new.get_virtualization(host_name)['details']
         assert virt_card['datacenter'] == module_vmware_settings['datacenter']
         assert virt_card['cluster'] == module_vmware_settings['cluster']
         assert virt_card['memory'] == '2 GB'
-        assert virt_card['public_ip_address']
+        assert 'public_ip_address' in virt_card
```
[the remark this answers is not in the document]
sambible: No, this is just asserting that the field is present, not that there is a value. I could just check for the presence of a value though

sambible:
Added that snippet you posted @adarshdubey-star  -  seems like a good way to make this test much less flaky! Thanks!

tests/foreman/ui/test_computeresource_vmware.py
sambible: I think I'll just check that there is a hypervisor returned at all

sambible:
trigger: test-robottelo
airgun: 933

shubhamsg199:
I think you'll also need to add wait till the VM is completely Powered ON, As it won't show IP address immediately after powering on.
You could add something like this https://github.com/SatelliteQE/robottelo/blob/f3bc09d2804869b7232a76d64b3f7498c5475eea/tests/foreman/ui/test_computeresource_vmware.py#L318

sambible:
trigger: test-robottelo
airgun: 933

sambible:
trigger: test-robottelo
airgun: 933

ogajduse (approved):

omkarkhatavkar:
trigger: test-robottelo
airgun: 933

sambible:
trigger: test-robottelo
airgun: 933

PR #12216, opened 2023-08-11
Fix capsule sync status persists

vsedmik:
There is some flakiness in this test, it sometimes fail with this error:
```
tests/foreman/api/test_capsulecontent.py:1278: in test_positive_capsule_sync_status_persists
    assert sync_status['last_sync_time'] >= timestamp
E   AssertionError: assert '2023-08-23 23:03:51 UTC' >= '2023-08-23 23:04'
```

As I learned from DEVs, the designed&implemented behaviour is to fall back to audits when all old sync tasks get deleted (which happens on a regular basis on cleanup). So the caps sync time is taken from audits **only** when all sync tasks are missing. This implies that we need to change the test logic, because if we delete the most recent sync task (and keep older ones, which is the current logic) we don't fall back to audits but to the previous task, which is older than our timestamp and the assertion fails.

### DIFF
--- a/tests/foreman/api/test_capsulecontent.py
+++ b/tests/foreman/api/test_capsulecontent.py
@@ -1224,7 +1224,7 @@ class TestCapsuleContentManagement:
             4. Publish CV
             5. Promote to lifecycle env
             6. Sync Capsule
-            7. Delete the task using foreman-rake console
+            7. Delete all sync tasks using foreman-rake console
             8. Verify the status of capsule is still synced
 
         :bz: 1956985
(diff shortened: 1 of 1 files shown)


vsedmik:
trigger: test-robottelo
pytest: tests/foreman/api/test_capsulecontent.py -k test_positive_capsule_sync_status_persists

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/api/test_capsulecontent.py

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/api/test_capsulecontent.py -k test_positive_capsule_sync_status_persists

ColeHiggins2 (approved):

Griffin-Sullivan (approved):

PR #12230, opened 2023-08-14
Fix clone tests to use the right sat_version

Griffin-Sullivan:
All of the satellite-clone tests are failing on stream, because this line resolves to `6.15.0`. Instead we want it to resolve to `stream`, so I opted to use `settings.satellite.version.release`. Only need this for stream, since it's a non-issue on our other branches.

### DIFF
--- a/tests/foreman/destructive/test_clone.py
+++ b/tests/foreman/destructive/test_clone.py
@@ -50,7 +50,7 @@ def test_positive_clone_backup(target_sat, sat_ready_rhel, backup_type, skip_pul
     :customerscenario: true
     """
     rhel_version = sat_ready_rhel._v_major
-    sat_version = target_sat.version
+    sat_version = 'stream' if target_sat.is_stream else target_sat.version
 
     # SATELLITE PART - SOURCE SERVER
     # Enabling and starting services


Griffin-Sullivan:
trigger: test-robottelo

tests/foreman/destructive/test_clone.py
```diff
@@ -50,7 +50,7 @@ def test_positive_clone_backup(target_sat, sat_ready_rhel, backup_type, skip_pul
     :customerscenario: true
     """
     rhel_version = sat_ready_rhel._v_major
-    sat_version = target_sat.version
+    sat_version = settings.server.version.release
```
Gauravtalreja1, line 53: We do've a `is_stream` property as well to handle this correctly and I think it'd be better to rely on satellite rpm instead of this setting.
```suggestion
    sat_version = 'stream' if target_sat.is_stream else target_sat.version
```

Gauravtalreja1 (approved):
Ack, with a non-blocking comment

tests/foreman/destructive/test_clone.py
Griffin-Sullivan: Ok awesome lets try that.

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_clone.py::test_positive_clone_backup[include_pulp-online]

PR #12260, opened 2023-08-16
Duplicate issues are created, as the GHA triggered based on label

omkarkhatavkar:
Based on the PR  https://github.com/SatelliteQE/robottelo/pull/12247 and comment https://github.com/SatelliteQE/robottelo/pull/12247#issuecomment-1680322427, many triggers are getting generated here and logic needs to change here could be written in separate GHA as well. 

image

many events are getting triggered could be a time difference causing this issue.

### DIFF
--- a/.github/workflows/auto_cherry_pick.yml
+++ b/.github/workflows/auto_cherry_pick.yml
@@ -3,7 +3,7 @@ name: auto_cherry_pick_commits
 
 on:
   pull_request_target:
-    types: [closed, labeled]
+    types: [closed]
 
 # Github & Parent PR Env vars
 env:


ogajduse (approved):
@omkarkhatavkar, thank you for removing it. It is indeed a good idea to disable it till we have the implementation more solid. :+1:

ogajduse:
@omkarkhatavkar I see that the CherryPick label is used without a specific branch tag, is that intentional?

Gauravtalreja1 (approved):

omkarkhatavkar:
> @omkarkhatavkar, thank you for removing it. It is indeed a good idea to disable it till we have the implementation more solid. 👍

updated!

PR #12280, opened 2023-08-17
removed wotfix-skipped test

pondrejk:
The blocking bugs were closed as wotfix long time ago, interestingly the test was still skipped in automation. This PR removes it

### DIFF
--- a/tests/foreman/ui/test_audit.py
+++ b/tests/foreman/ui/test_audit.py
@@ -18,7 +18,6 @@ import pytest
 from fauxfactory import gen_string
 from nailgun import entities
 
-from robottelo.constants import ANY_CONTEXT
 from robottelo.constants import ENVIRONMENT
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan (approved):
Is it worth keeping the test and adapting it to pass for the way it currently is in Satellite?

Gauravtalreja1 (approved):

synkd (approved):

pondrejk:
> Is it worth keeping the test and adapting it to pass for the way it currently is in Satellite?

@Griffin-Sullivan not convinced it would make sense, we could reverse the assertion to check the log entry is *not* present, but what kind of bug we could possibly find using such a test?

Griffin-Sullivan:
@pondrejk yeah sounds like that would be a waste of time.

PR #12289, opened 2023-08-17
Assert no version available on satellite-maintain upgrade list-versions for stream

Griffin-Sullivan:
On stream we do not have a version to upgrade to, so this test is failing. Adding a conditional to check on stream that there are no numbers in the output.

### DIFF
--- a/tests/foreman/maintain/test_upgrade.py
+++ b/tests/foreman/maintain/test_upgrade.py
@@ -16,6 +16,8 @@
 
 :Upstream: No
 """
+import re
+
 import pytest
 
 from robottelo.config import settings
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_upgrade.py::test_positive_satellite_maintain_upgrade_list

Griffin-Sullivan:
Ok the initial approach didn't work because capsule showed kernel updates in the output of the command which contained numbers... Gonna need to rethink how to do this.

synkd (commented):
Those kernel versions are showing because somewhere along the line `dnf` installed a new kernel version, but the system was not rebooted, so the system is still running an older kernel version. I wouldn't think this behavior would be exclusive to stream, but I admittedly do not have a sophisticated understanding of the upgrade test workflow, so I'm not sure when `dnf upgrade` gets run in which scenarios.

A couple of ways to work around this issue come to mind. You could reboot the system before running the `list-versions` command, or you could run `grep -v "Security:"` against the output of `list-versions` and then parse the result of that for your assertion. Do you anticipate issues with either of those approaches?

Griffin-Sullivan:
@synkd SUCCESS!

synkd (approved):

PR #12309, opened 2023-08-22
bypass cvv promote to lce if already promoted

vijaysawant:
Upgrade test was failing with Error
`Could not promote the content view: Cannot promote environment out of sequence`
with this fix we bypass cvv promote if already promoted to lce

### DIFF
--- a/robottelo/cli/factory.py
+++ b/robottelo/cli/factory.py
@@ -1691,9 +1691,15 @@ def setup_org_for_a_custom_repo(options=None):
     cvv = ContentView.info({'id': cv_id})['versions'][-1]
     # Promote version to next env
     try:
-        ContentView.version_promote(
-            {'id': cvv['id'], 'organization-id': org_id, 'to-lifecycle-environment-id': env_id}
-        )
+        all_lifecycle_env = ContentView.info({'id': cv_id})['lifecycle-environments']
+        for lce in all_lifecycle_env:
+            if int(lce['id']) == env_id:
+                # already promoted to environment
+                break
+        else:
+            ContentView.version_promote(
+                {'id': cvv['id'], 'organization-id': org_id, 'to-lifecycle-environment-id': env_id}
+            )
     except CLIReturnCodeError as err:
         raise CLIFactoryError(f'Failed to promote version to next environment\n{err.msg}')
     # Create activation key if needed and associate content view with it
--- a/robottelo/cli/factory.py
+++ b/robottelo/cli/factory.py
@@ -1687,16 +1687,13 @@ def setup_org_for_a_custom_repo(options=None):
         ContentView.publish({'id': cv_id})
     except CLIReturnCodeError as err:
         raise CLIFactoryError(f'Failed to publish new version of content view\n{err.msg}')
-    # Get the version id
-    cvv = ContentView.info({'id': cv_id})['versions'][-1]
+    # Get the content view info
+    cv_info = ContentView.info({'id': cv_id})
+    lce_promoted = cv_info['lifecycle-environments']
+    cvv = cv_info['versions'][-1]
     # Promote version to next env
     try:
-        all_lifecycle_env = ContentView.info({'id': cv_id})['lifecycle-environments']
-        for lce in all_lifecycle_env:
-            if int(lce['id']) == env_id:
-                # already promoted to environment
-                break
-        else:
+        if env_id not in [int(lce['id']) for lce in lce_promoted]:
             ContentView.version_promote(
                 {'id': cvv['id'], 'organization-id': org_id, 'to-lifecycle-environment-id': env_id}
             )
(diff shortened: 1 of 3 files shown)


vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/longrun/test_oscap.py -k test_positive_oscap_run_via_ansible

robottelo/cli/factory.py
```diff
@@ -1691,9 +1691,15 @@ def setup_org_for_a_custom_repo(options=None):
...
-        )
+        all_lifecycle_env = ContentView.info({'id': cv_id})['lifecycle-environments']
+        for lce in all_lifecycle_env:
+            if int(lce['id']) == env_id:
+                # already promoted to environment
+                break
+        else:
+            ContentView.version_promote(
+                {'id': cvv['id'], 'organization-id': org_id, 'to-lifecycle-environment-id': env_id}
+            )
```
adarshdubey-star, line 1702: Non blocking - Not tried locally but looks more clean.
```suggestion
if not any(int(lce['id']) == env_id for lce in all_lifecycle_env):
    ContentView.version_promote(
        {'id': cvv['id'], 'organization-id': org_id, 'to-lifecycle-environment-id': env_id}
    )
```

ogajduse:
@jyejare What is the strategy for updating the CLI factory functions?
The file that @vijaysawant touched is in `robottelo/cli/factory.py`, but the same function is defined in `robottelo/host_helpers/cli_factory.py`.
Which one should get updated? One of them or both of them until we deprecate the `robottelo/cli/factory.py`?

jyejare:
The robottelo/host_helpers/cli_factory.py is the right one.

We have deprecated the other one unofficially long ago, now we are officially deprecating it.

vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/longrun/test_oscap.py::test_positive_oscap_run_via_ansible

robottelo/cli/factory.py
```diff
@@ -1691,9 +1691,15 @@ def setup_org_for_a_custom_repo(options=None):
     cvv = ContentView.info({'id': cv_id})['versions'][-1]
     # Promote version to next env
     try:
-        ContentView.version_promote(
-            {'id': cvv['id'], 'organization-id': org_id, 'to-lifecycle-environment-id': env_id}
-        )
+        all_lifecycle_env = ContentView.info({'id': cv_id})['lifecycle-environments']
```
Gauravtalreja1, line 1694: We're running same ContentView.info command for getting cvv and LCE, if you could store in some variable first and use it for CVV and LCE, then we don't run it again

robottelo/cli/factory.py
```diff
@@ -1691,9 +1691,15 @@ def setup_org_for_a_custom_repo(options=None):
...
+        for lce in all_lifecycle_env:
+            if int(lce['id']) == env_id:
+                # already promoted to environment
+                break
```
Gauravtalreja1, line 1698: why break here, if CV is already promoted, we need to go ahead and follow further steps, right?

robottelo/cli/factory.py
Gauravtalreja1, line 1694: here instead of all_lifecycle_env, I'd define var with meaningful name like `lce_promoted`?

robottelo/cli/factory.py
Gauravtalreja1: this looks like a mess, this might not work :-) try something like
```suggestion
       if env_id not in [lce['id'] for lce in all_lifecycle_env]:
            ContentView.version_promote(
                {'id': cvv['id'], 'organization-id': org_id, 'to-lifecycle-environment-id': env_id}
            )
```

robottelo/cli/factory.py
vijaysawant: if no `env_id` matches with `int(lce['id'])` for loop will not break and it will goes inside else to promote at least once.

robottelo/cli/factory.py
vijaysawant: With above suggested approach
**Case1:**
`env_id = 7` and` lce = {['id:'3', 'name': 'A' ], [ 'id':'4', 'name': 'B' ], [ 'id':'5', 'name': 'C' ], [ 'id':'6', 'name': 'D' ]}`

CVV will promote to all lce id (i.e for lce A, B, C & D)

vijaysawant: sorry, I interpreted it in wrong way, it will promote to one environment for `len(lce)` times

vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/cli/test_activationkey.py -k test_positive_ak_with_custom_product_on_rhel6

vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/longrun/test_oscap.py

vijaysawant:
@team rocket, as per discussion I checked that oscap tests from upgrade-all-tiers pipeline for **6.12 zstream** and **6.13 ystream** were failing, hence adding labels

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/longrun/test_oscap.py

shweta83:
trigger: test-robottelo
pytest: tests/foreman/cli/test_activationkey.py -k test_positive_ak_with_custom_product_on_rhel6

Gauravtalreja1 (approved):

PR #12314, opened 2023-08-23
Add Libvirt CR setup helper to fix UI E2E tests

Gauravtalreja1:
Libvirt CR creation doesn't requires this setup but after that while reading and updating compute_profile it fails to connect Libvirt, so for fixing this added a helper which can be used to setup this and will update a few more tests of host provisioning to use this helper in future PR.

### DIFF
--- a/robottelo/host_helpers/satellite_mixins.py
+++ b/robottelo/host_helpers/satellite_mixins.py
@@ -310,6 +310,33 @@ class SystemInfo:
         return result.stdout
 
 
+class ProvisioningSetup:
+    """Provisioning tests setup helper methods"""
+
+    def configure_libvirt_cr(self, server_fqdn=settings.libvirt.libvirt_hostname):
+        """Configures Libvirt ComputeResource to communicate with Satellite
+
+        :param server_fqdn: Libvirt server FQDN
+        :return: None
+        """
+        # Geneate SSH key-pair for foreman user and copy public key to libvirt server
+        self.execute('sudo -u foreman ssh-keygen -q -t rsa -f ~foreman/.ssh/id_rsa -N "" <<< y')
+        self.execute(f'ssh-keyscan -t ecdsa {server_fqdn} >> ~foreman/.ssh/known_hosts')
+        self.execute(
+            f'sshpass -p {settings.server.ssh_password} ssh-copy-id -o StrictHostKeyChecking=no '
+            f'-i ~foreman/.ssh/id_rsa root@{server_fqdn}'
+        )
+        # Install libvirt-client, and verify foreman user is able to communicate with Libvirt server
+        self.register_to_cdn()
+        self.execute('dnf -y --disableplugin=foreman-protector install libvirt-client')
+        assert (
+            self.execute(
+                f'su foreman -s /bin/bash -c "virsh -c qemu+ssh://root@{server_fqdn}/system list"'
+            ).status
+            == 0
+        )
+
+
 class Factories:
     """Mixin that provides attributes for each factory type"""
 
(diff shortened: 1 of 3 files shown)


Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/ui/test_computeresource_libvirt.py

adarshdubey-star (approved):

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/ui/test_computeresource_libvirt.py

robottelo/hosts.py
```diff
@@ -2166,6 +2166,29 @@ def enroll_ad_and_configure_external_auth(self, ad_data):
             self.execute('systemctl daemon-reload && systemctl restart httpd.service').status == 0
         )
 
+    def configure_libvirt_cr(self, server_fqdn=settings.libvirt.libvirt_hostname):
```
jyejare, line 2169: Adding helpers directly in host classes is prohibited, please add them to any of the Mixins or create new mixin class and make that a parent of this class.

robottelo/hosts.py
```diff
@@ -2166,6 +2166,29 @@ def enroll_ad_and_configure_external_auth(self, ad_data):
...
+        """Configures Libvirt ComputeResource to communicate with Satellite
+
+        :param server_fqdn: Libvirt server FQDN
+        :return: None
+        """
+        # Geneate SSH key-pair for foreman user and copy public key to libvirt server
+        self.execute('sudo -u foreman ssh-keygen -q -t rsa -f ~foreman/.ssh/id_rsa -N "" <<< y')
+        self.execute(f'ssh-keyscan -t ecdsa {server_fqdn} >> ~foreman/.ssh/known_hosts')
```
jyejare, line 2177: Should you be scanning for `rsa` key here and not ecdsa ?

jyejare (changes requested):
ACK pending comment and question.

robottelo/hosts.py
Gauravtalreja1: ideally both works as we just need to make env-* server accessible, but here I think it'll be just using it for ssh from foreman user then ecdsa is fine, and I tried using rsa for the same purpose, it adds up unncessary warnings related to it, so I'd stick with scanning ecdsa only.

robottelo/hosts.py
Gauravtalreja1: I wouldn't mind adding it to mixins, but I'm not sure since when we're prohibiting the helpers addition to host classes :)

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/ui/test_computeresource_libvirt.py

robottelo/hosts.py
jyejare: Sorry I missed the key work in my initial comment. I mean to say we dont want to add non-host-classes related helpers into that module directly, that way the file would increase into tones. 

Hence the reason we have Mixins !

robottelo/host_helpers/satellite_mixins.py
```diff
@@ -310,6 +310,33 @@ def validate_pulp_filepath(
         return result.stdout
 
 
+class SetupHelpers:
```
jyejare, line 313: Just setup Helpers doents look good, again people might dump everything here, since every test has helpers for setup purpose mostly.

Name it in way that is purposeful and targetting specific need something like `ProvisioningSetup` or so. The other classes we have are `ContentInfo, Factories, SystemInfo, EnablePluginsSatellite` that targets the specifc classification.

jyejare (changes requested):
Final comment !

robottelo/hosts.py
Gauravtalreja1: yes, I agree on it @jyejare , so I've already updated the PR to move it into mixins

robottelo/host_helpers/satellite_mixins.py
Gauravtalreja1: @jyejare  I think its better to have generic here, if we're going to add specific classes for component specific helpers then we'll end up inheriting SatelliteMixins class from so many classes, so SetupHelpers makes more sense to move setup specific helpers under it

jyejare (approved):

PR #12323, opened 2023-08-23
Add service list check

Griffin-Sullivan:
During the component evaluation SD recommended adding a check for each service being listed to this test as they have found it sometimes will report `[OK]` but miss some services.

### DIFF
--- a/tests/foreman/maintain/test_service.py
+++ b/tests/foreman/maintain/test_service.py
@@ -23,8 +23,32 @@ from robottelo.config import settings
 from robottelo.constants import HAMMER_CONFIG
 from robottelo.constants import MAINTAIN_HAMMER_YML
 from robottelo.constants import SATELLITE_ANSWER_FILE
+from robottelo.hosts import Satellite
 
 pytestmark = pytest.mark.destructive
+
+SATELLITE_SERVICES = [
+    "dynflow-sidekiq@.service",
+    "foreman-proxy.service",
+    "foreman.service",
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_service.py::test_positive_service_list

synkd (approved):

tests/foreman/maintain/test_service.py
```diff
@@ -23,9 +23,33 @@
...
+CAPSULE_SERVICES = [
+    "foreman-proxy.service",
+    "httpd.service",
+    "postgresql.service",
+    "pulpcore-api.service",
+    "pulpcore-content.service",
+    "pulpcore-worker@.service",
+    "redis.service",
+]
+
```
Gauravtalreja1, line 52: Can we move this to constants please? I think some installer tests also uses same constants, so it can be kept at single place. If yes, then can you try if those tests can be updated as well?

Gauravtalreja1 (approved):
Ack, with a suggestion

tests/foreman/maintain/test_service.py
Griffin-Sullivan: The list in installer is a little different that's why I decided to keep the constant in here. If we use this for another file then I'll move it to constants then. https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/installer/test_installer.py#L1296

PR #12175, opened 2023-08-08
Deprecating api/utils.py module

jyejare:
- Deprecated api/utils.py module 
- Almost all the functions in these modules are already moved under API_FACTORY class to access them through `sat.api_factory.some_function()` except for three definitions.
- These three definitions are now moved in this PR (Though there are no tests using those functions, added just in case if it helps). We can remove these if reviewers find that we have similar functionality in other existing definitions.

PR #12325, opened 2023-08-23
[6.14.z] Deprecating api/utils.py module

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/12175

- Deprecated api/utils.py module 
- Almost all the functions in these modules are already moved under API_FACTORY class to access them through `sat.api_factory.some_function()` except for three definitions.
- These three definitions are now moved in this PR (Though there are no tests using those functions, added just in case if it helps). We can remove these if reviewers find that we have similar functionality in other existing definitions.


Satellite-QE:
trigger: test-robottelo

PR #12347, opened 2023-08-24
ISS refactor - batch 2

vsedmik:
Second batch of changes coming from the ISS evaluation

Introduced changes:
1. Added fixtures for importing orgs. These need to be new orgs so they don't reference the original `org` fixtures, which are used for export.
2. `_enable_rhel_repo` helper function was replaced by parametrized fixture and used through the module.
3. Using the fixture above three test cases were merged with `test_positive_export_import_redhat_cv`:
* `test_positive_export_complete_version_rh_repo` - was practically same as the first (export) part of the above case.
* `test_positive_export_import_redhat_cv_with_huge_contents` - same as the above case but bigger (rhscl) repo.
* `test_positive_export_import_kickstart_tree` - was a `stub` but now it's automated by the case above
NOTE: The original plan was to test **all** KS and RHEL repos, however, due to the immediate download policy it would require more than the default 100G. So we need to decide if we want to test the real customer scenarios and go big, or use a small subset to stay bellow 100G.
4. Added RH repo to `test_positive_export_import_default_org_view` to address the PX comment.
5. Moved to SCA org as new default.
6. Some other minor changes like drift towards `target_sat.cli*` and docstrings on touched tests.

### DIFF
--- a/robottelo/constants/__init__.py
+++ b/robottelo/constants/__init__.py
@@ -858,6 +858,7 @@ CUSTOM_PUPPET_MODULE_REPOS_VERSION = '-0.2.0.tar.gz'
 
 PULP_EXPORT_DIR = '/var/lib/pulp/exports/'
 PULP_IMPORT_DIR = '/var/lib/pulp/imports/'
+EXPORT_LIBRARY_NAME = 'Export-Library'
 
 PUPPET_COMMON_INSTALLER_OPTS = {
     'foreman-proxy-puppetca': 'true',
--- a/tests/foreman/cli/test_satellitesync.py
+++ b/tests/foreman/cli/test_satellitesync.py
@@ -18,17 +18,16 @@
 """
 import pytest
 from fauxfactory import gen_string
+from manifester import Manifester
 
 from robottelo.cli.base import CLIReturnCodeError
 from robottelo.cli.content_export import ContentExport
 from robottelo.cli.content_import import ContentImport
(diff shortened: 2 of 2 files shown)


vsedmik:
trigger: test-robottelo
pytest: tests/foreman/cli/test_satellitesync.py -k 'test_positive_export_import_default_org_view or test_positive_export_import_redhat_cv or test_positive_export_complete_library_rh_repo or test_negative_import_redhat_cv_without_manifest or test_positive_export_import_cv_end_to_end or test_positive_export_import_promoted_cv or test_negative_import_same_cv_twice or test_postive_export_import_cv_with_file_content or test_postive_export_import_ansible_collection_repo or test_positive_import_content_for_disconnected_sat_with_existing_content'

sambible (approved):
I think it would be good to push for all the repos, and get some special sauce worked out to allow for it. The closer we can get to testing actual customer systems, the more value tests like that (replicating customer environments) provide. 

Otherwise, changes look good, don't see anything glaringly wrong. ack pending PRT

vsedmik:
> I think it would be good to push for all the repos, and get some special sauce worked out to allow for it. The closer we can get to testing actual customer systems, the more value tests like that (replicating customer environments) provide.
Maybe we could create some session or module scoped big setup for the "Export sync" to allow that. We need create some special sort of setup for "Network sync" anyway, so I would scope that out to a different PR.

pondrejk (approved):

PR #12385, opened 2023-08-28
Remove some redundant tests from Errata-management

damoore044:
-Component Evaluation of Errata Management-

Marked for removal with consideration for merging (with input from DEV, QE, SD, PX):
- `API:test_positive_install_in_host`: Same scenario is tested in a loop during API:test_positive_install_multiple_in_host.
   Would it be worth merging the single errata install test into the beginning of the multiple errata install test?
- `CLI:test_positive_user_permission`: A very similar situation is tested in UI:test_positive_list_permission, showing errata for a new user with permissions. Worth merging?


JacobCallahan (approved):

sambible (approved):
Seems reasonable to me.

PR #12393, opened 2023-08-29
[6.11.z] Add Libvirt CR setup helper to fix UI E2E tests

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/12314

Libvirt CR creation doesn't requires this setup but after that while reading and updating compute_profile it fails to connect Libvirt, so for fixing this added a helper which can be used to setup this and will update a few more tests of host provisioning to use this helper in future PR.


Satellite-QE:
trigger: test-robottelo
pytest: tests/foreman/ui/test_computeresource_libvirt.py

PR #12245, opened 2023-08-15
Update module_synced_repos to explicitly use session_target_sat

Griffin-Sullivan:
When running foreman-maintain tests that have `sat_maintain == session_capsule_configured` it will fail here because `Capsule` does not have an `api` attribute. This update should fix the issue by explicitly calling `session_target_sat`. I also left `sat_maintain` in the fixture since I want this to be done after `sat_maintain` has been created.

PR #12402, opened 2023-08-29
[6.13.z] Use stash to only run module_synced_repos one time

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/12245

When running foreman-maintain tests that have `sat_maintain == session_capsule_configured` it will fail here because `Capsule` does not have an `api` attribute. This update should fix the issue by explicitly calling `session_target_sat`. I also left `sat_maintain` in the fixture since I want this to be done after `sat_maintain` has been created.

### DIFF
--- a/pytest_fixtures/component/maintain.py
+++ b/pytest_fixtures/component/maintain.py
@@ -6,7 +6,19 @@ import pytest
 from robottelo import constants
 from robottelo.config import settings
 from robottelo.constants import SATELLITE_MAINTAIN_YML
+from robottelo.hosts import Capsule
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan:
So it seems that the automerge cherry-pick action failed because it timed out waiting for PRT. I'm not surprised because `test_positive_backup_restore` runs 8 tests that are all quite long. Anyways, you can see it passed 8/8 and is fine to merge.

jyejare (approved):

PR #12405, opened 2023-08-29
Small Adjustment to Virt-card Assert Statements

sambible:
Hopefully last tweak of this PR. These fields that are changing here sometimes (TM) don't appear when running in CI, or sometimes don't have the proper value.

I don't particularly care to check these values, since they are quite dynamic, but I'd like to check for them to not be empty if present ( Aside from Public IP, that one always appears and sometimes does have an empty value )

LGTM.


sambible:
trigger: test-robottelo

sambible:
trigger: test-robottelo

vsedmik (approved):

PR #12415, opened 2023-08-29
Just check for presence of IP address

sambible:
Good grief. Seems that IP address still has a chance to be empty, so I'm just checking that it's a field at all. Please dear lord let this be the last PR to address this dumb card.


sambible:
trigger: test-robottelo

Gauravtalreja1 (approved):

sambible:
Seems to me to be a timing related thing, potentially, since I've also seen it during certain local runs.

sambible:
I'd like to get this in, so this can pass in 6.14 pipeline, and we can monitor results and see what happens on subsequent runs.

vsedmik (approved):

PR #12433, opened 2023-08-31
Change to use function fixtures and explicitly install

sambible:
Some small updates to not leak state between similar tests, and to make sure that repos are enabled.


sambible:
trigger: test-robottelo
pytest: tests/foreman/api/test_reporttemplates.py

PR #12466, opened 2023-09-04
Delete Host record on teardown

### LINKED ISSUE #12438: RFE: Create a log record when we go for host checkin
## Problem statement
It is difficult for the user to distinguish where the host check-in started in the log file - what log records are being logged from the check-in phase.

## Solution
Call the logger once we start checking in and log that the check-in process started.

ogajduse:
Try to delete the host record even if the host is not registered to Satellite

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -227,15 +227,30 @@ class ContentHost(Host, ContentHostMixins):
         return self._satellite
 
     @property
+    def _sat_host_record(self):
+        """Provide access to this host's Host record if it exists."""
+        hosts = self.satellite.api.Host().search(query={'search': self.hostname})
+        if not hosts:
+            logger.debug('No host record found for %s on Satellite', self.hostname)
+            return None
+        return hosts[0]
+
+    def _delete_host_record(self):
+        """Delete the Host record of this host from Satellite."""
+        if h_record := self._sat_host_record:
+            logger.debug('Deleting host record for %s from Satellite', self.hostname)
+            h_record.delete()
+
+    @property
     def nailgun_host(self):
         """If this host is subscribed, provide access to its nailgun object"""
         if self.identity.get('registered_to') == self.satellite.hostname:
             try:
-                host_list = self.satellite.api.Host().search(query={'search': self.hostname})[0]
+                host = self._sat_host_record
             except Exception as err:
                 logger.error(f'Failed to get nailgun host for {self.hostname}: {err}')
-                host_list = None
-            return host_list
+                host = None
+            return host
         else:
             logger.warning(f'Host {self.hostname} not registered to {self.satellite.hostname}')
 
@@ -379,8 +394,8 @@ class ContentHost(Host, ContentHostMixins):
     def teardown(self):
         if not self.blank and not getattr(self, '_skip_context_checkin', False):
             self.unregister()
-            if type(self) is not Satellite and self.nailgun_host:
-                self.nailgun_host.delete()
+            if type(self) is not Satellite:  # do not delete Satellite's host record
+                self._delete_host_record()
 
     def power_control(self, state=VmState.RUNNING, ensure=True):
         """Lookup the host workflow for power on and execute


ogajduse:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py::TestRemoteExecution::test_positive_run_job_effective_user_by_ip[rhel9] tests/foreman/cli/test_remoteexecution.py::TestAnsibleREX::test_positive_run_effective_user_job[rhel9] tests/foreman/cli/test_host.py::test_positive_erratum_applicability[rhel9] tests/foreman/destructive/test_packages.py::test_positive_all_packages_update

ogajduse:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py::TestRemoteExecution::test_positive_run_job_effective_user_by_ip[rhel9] tests/foreman/cli/test_remoteexecution.py::TestAnsibleREX::test_positive_run_effective_user_job[rhel9] tests/foreman/cli/test_host.py::test_positive_erratum_applicability[rhel9] tests/foreman/destructive/test_packages.py::test_positive_all_packages_update

robottelo/hosts.py
```diff
@@ -226,16 +226,33 @@ def satellite(self):
...
+    def _delete_host_record(self):
+        """Delete the Host record of this host from Satellite."""
+        if h_record := self._sat_host_record:
+            h_record.delete()
+        else:
+            logger.warning(
+                f'Attempted to delete Host {self.hostname}, but it was not found in Satellite'
+            )
+
     @property
```
Gauravtalreja1, line 246: I don't think it makes sense to write new properties for same thing, instead shouldn't we cover the other posiblity of host is registered to capsule or to server other than CDN into the nailgun_host property itself, wdyt?

ogajduse: The main issue is that `subscription-manager identity` returns nothing on stdout, therefore the return value of `self.identity` is `None`. Therefore `self.identity.get('registered_to') == self.satellite.hostname` is evaluated as `False` and `nailgun_host` then returns `None`.
I have tried to avoid extending the logic of the nailgun host to keep it simple. Therefore `type(self) is not Satellite and self.nailgun_host` evaluates to `True and False` --> `False` in `teardown`, so the host entity does not get deleted.

I am open to changing it. Feel free to share your ideas around this.

Gauravtalreja1: yes, `subscription-manager identity` returns some uuid and org details in stdout, and identity property returns a dict which also contains the `registered_to: subscription_config['server']['hostname']`, so I think there'll be a some value present always but not sure about self.satellite.hostname, while I don't see any issue with removing nailgun_host from `type(self) is not Satellite and self.nailgun_host` it's fine, but I still feel same nailgun_host property should be fixed for this issue and not in other similar properties

robottelo/hosts.py
```diff
@@ -226,16 +226,33 @@ def satellite(self):
             self._satellite = Satellite()
         return self._satellite
 
+    @property
+    def _sat_host_record(self):
+        """Provide access to this host's Host record if it exists."""
+        host = None
+        with contextlib.suppress(IndexError):
+            host = self.satellite.api.Host().search(query={'search': self.hostname})[0]
```
jameerpathan111, line 234: Can you add `logger.warning/debug` here instead of in `_delete_host_record`, when host record is not found.

ogajduse: addressed in ac03928e4

ogajduse:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py::TestRemoteExecution::test_positive_run_job_effective_user_by_ip[rhel9] tests/foreman/cli/test_remoteexecution.py::TestAnsibleREX::test_positive_run_effective_user_job[rhel9] tests/foreman/cli/test_host.py::test_positive_erratum_applicability[rhel9] tests/foreman/destructive/test_packages.py::test_positive_all_packages_update

jameerpathan111 (approved):

ColeHiggins2 (approved):

PR #12498, opened 2023-09-05
Workaround the issue as long as BZ2228820 is open

lhellebr:


ColeHiggins2 (approved):
Looks good to me

ColeHiggins2:
trigger: test-robottelo
pytest: tests/foreman/api/test_organization.py -k test_positive_create

tests/foreman/api/test_organization.py
```diff
@@ -77,7 +78,10 @@ def test_positive_create(self):
             headers={'content-type': 'text/plain'},
```
Gauravtalreja1, line 78: For workaround here, wouldn't it better to change content-type instead when bz is open with `is_open` and not checking UNSUPPORTED_MEDIA_TYPE, wdyt?

lhellebr: As it is now, we're testing that UNSUPPORTED_MEDIA_TYPE is being returned, or at least ISE 500 if the BZ is open (as opposed to returning something even more wrong, e.g. 200 OK).

Gauravtalreja1: IMO using any other content-type's isn't wrong, and I feel its better option to test something than checking if we hit UNSUPPORTED_MEDIA_TYPE/ISE :)

Its upto you how you want to test it, so its non-blocking.

Gauravtalreja1 (approved):

Gauravtalreja1:
Why you're looking for additional tier1 and tier2 reviews on this? :)

adarshdubey-star:
Please add the description and we can merge this!

lhellebr:
@Gauravtalreja1, because nobody merged and I shouldn't self-merge.

lhellebr:
@adarshdubey-star there's both PR name and commit message, why do you think description is necessary? Is something unclear based on the commit message?

adarshdubey-star:
> @adarshdubey-star there's both PR name and commit message, why do you think description is necessary? Is something unclear based on the commit message?

Nope just a good to have thing :)

PR #12508, opened 2023-09-06
Expect exception instead of false return value

lhellebr:


lhellebr:
trigger: test-robottelo
pytest: tests/foreman/ui/test_organization.py::test_positive_end_to_end

Gauravtalreja1 (approved):

PR #12515, opened 2023-09-06
Debug failover to cloned manifests

synkd:
This PR, in combination with https://github.com/SatelliteQE/manifester/pull/21, addresses some of the errors that have been occurring in CI in cases where Manifester times out when exporting a manifest due to an upstream RHSM issue. These changes should enable Robottelo to successfully fail over to using cloned manifests in those cases.

### DIFF
--- a/robottelo/host_helpers/satellite_mixins.py
+++ b/robottelo/host_helpers/satellite_mixins.py
@@ -141,13 +141,13 @@ class ContentInfo:
         :returns: the manifest upload result
 
         """
-        if manifest is None:
+        if manifest.content is None:
             manifest = clone()
         if timeout is None:
             # Set the timeout to 1500 seconds to align with the API timeout.
             timeout = 1500000
         if interface == 'CLI':
-            if isinstance(manifest.content, (bytes, io.BytesIO)):
+            if hasattr(manifest, 'path'):
                 self.put(f'{manifest.path}', f'{manifest.name}')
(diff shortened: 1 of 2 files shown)


synkd:
Setting the `in progress` label until https://github.com/SatelliteQE/manifester/pull/21 is merged.

jyejare (approved):

JacobCallahan (approved):

JacobCallahan:
@synkd if this is ready, can you add a PRT run?

synkd:
@JacobCallahan since the timeout that would trigger the failover does not happen in upstream RHSM with any discernible pattern, PRT won't reliably tell us whether the change worked. It's more likely to tell us that the change didn't break anything. This is probably worth demonstrating, so I will trigger a run.

I can artificially trigger the failover to cloned manifests on the manifester side by setting the timeout to 0. I'd prefer do this locally and provide some PDB output demonstrating that the control flow is working as expected. Would that work for you?

JacobCallahan:
Well we can test no-break by running a simple test that touches the closest part of manifester

synkd:
trigger: test-robottelo
pytest: tests/foreman/cli/test_contentaccess.py::test_negative_rct_not_shows_golden_ticket_enabled

synkd:
@JacobCallahan any test that uses one of the manifester fixtures from taxonomy.py will adequately test no-break, so we're good on that front with the PRT run above.

PR #12521, opened 2023-09-07
PRT comment adding using curl more standard way

omkarkhatavkar:
Moving away from github action to standard curl for handing the PRT comment

### DIFF
--- a/.github/workflows/auto_cherry_pick.yml
+++ b/.github/workflows/auto_cherry_pick.yml
@@ -77,11 +77,14 @@ jobs:
       - name: Add Parent PR's PRT comment to Auto_Cherry_Picked PR's
         id: add-parent-prt-comment
         if: ${{ always() && steps.cherrypick.outcome == 'success' }}
-        uses: mshick/add-pr-comment@v2
-        with:
-          issue: ${{ steps.cherrypick.outputs.number }}
-          message: ${{ needs.find-the-parent-prt-comment.outputs.prt_comment }}
-          repo-token: ${{ secrets.CHERRYPICK_PAT }}
+        run: |
+          ISSUE_NUMBER=${{ steps.cherrypick.outputs.number }}
+          COMMENT_BODY=${{ needs.find-the-parent-prt-comment.outputs.prt_comment }}
+
+          curl -X POST -H "Authorization: token ${{ secrets.CHERRYPICK_PAT }}" \
+            -H "Accept: application/vnd.github.v3+json" \
+            "https://api.github.com/repos/${{ github.repository }}/issues/${ISSUE_NUMBER}/comments" \
+            -d "{\"body\":\"$COMMENT_BODY\"}"
 
       - name: is autoMerging enabled for Auto CherryPicked PRs ?
         if: ${{ always() && steps.cherrypick.outcome == 'success' && contains(github.event.pull_request.labels.*.name, 'AutoMerge_Cherry_Picked') }}


.github/workflows/auto_cherry_pick.yml
```diff
@@ -63,11 +63,15 @@ jobs:
       - name: Add Parent PR's PRT comment to Auto_Cherry_Picked PR's
         id: add-parent-prt-comment
         if: ${{ always() && steps.cherrypick.outcome == 'success' }}
-        uses: mshick/add-pr-comment@v2
-        with:
-          issue: ${{ steps.cherrypick.outputs.number }}
-          message: ${{ needs.find-the-parent-prt-comment.outputs.prt_comment }}
```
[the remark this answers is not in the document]
omkarkhatavkar: while pasting the comment it adds more statement such as `<!-- add-pr-comment:add-pr-comment -->` hence removing that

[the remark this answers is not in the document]
Gauravtalreja1: yeah, I see, and we're reducing the dependancy of 3rd party GHAs by using GH APIs, so ACK, +1 to this.

Gauravtalreja1 (approved):

.github/workflows/auto_cherry_pick.yml
```diff
@@ -63,11 +63,15 @@ jobs:
...
-          repo-token: ${{ secrets.CHERRYPICK_PAT }}
+        run: |
+          ISSUE_NUMBER=${{ steps.cherrypick.outputs.number }}
+          GITHUB_TOKEN="${{ secrets.CHERRYPICK_PAT }}"
```
jyejare, line 68: Not sure! But this must be overriding the default GITHUB_TOKEN comes from Github, if not or anyways its good to rename it just to avoid the confusion !

jyejare (approved):
ACK ! But has got some question !

.github/workflows/auto_cherry_pick.yml
Gauravtalreja1: @omkarkhatavkar Can you check this and update PR accordingly?

omkarkhatavkar: @jyejare , @Gauravtalreja1  we are not overriding the `GITHUB_TOKEN` for this workflow, we are just passing the "${{ secrets.CHERRYPICK_PAT }}" our own Satellie_user token, you know it why we need that if we use the `GITHUB_TOKEN` then the PR would have comment from the GitHubAction user, while we are using `Satellite` user everywhere.

jyejare: @omkarkhatavkar I got that! I just wanted to reiterate that we could have named the `GITHUB_TOKEN` env var to something else for CHERRYPICK_PAT, cause this will actually override the `GITHUB_TOKEN` which in the feature you might want to use in this GHA.

I mean is the naming of env var here deliberate?

omkarkhatavkar: I can directly pass the CHERRYPICK_PAT instead the variable as well. If that work for you ?

jyejare: Sure! Use it directly :)

pondrejk (approved):

.github/workflows/auto_cherry_pick.yml
omkarkhatavkar: @jyejare tested here https://github.com/omkarkhatavkar/robottelo/actions/runs/6693395618/job/18184437416?pr=158, token is not `****` looks okay to me

jyejare: Sure!

PR #12529, opened 2023-09-07
Fix test_satellite_and_capsule_installation

jameerpathan111:


rmynar (approved):

robottelo/hosts.py
```diff
@@ -129,9 +129,9 @@ def setup_capsule(satellite, capsule, registration_args=None, installation_args=
         f'sshpass -p "{capsule.password}" scp -o "StrictHostKeyChecking no" '
         f'{file} root@{capsule.hostname}:{file}'
     )
-    capsule.install_katello_ca(sat_hostname=satellite.hostname)
+    capsule.install_katello_ca(satellite)
```
rmynar, line 132: Due to some messy cherrypicking the caller (_test_satellite_and_capsule_installation_) uses _org_ as 3rd positional argument of _setup_capsule_ and gets mistaken as _registration_args_. To fix it we need to remove the _org_ from test or add the positional argument into _setup_capsule_. I prefer the second choice - it seems safe because _test_satellite_and_capsule_installation_ is the only caller.
`def setup_capsule(satellite, capsule, org, registration_args=None, installation_args=None):`

rmynar (changes requested):

rmynar (approved):

pondrejk (approved):

PR #12571, opened 2023-09-11
change in import/export directory context

pondrejk:
Updating the fixture as BZ#2232162 evolved into documentation fixture. The new recommended way is to either use a pre-defined export dir or use a different context if custom one is needed. I chose the second option as the first one is already tested in bats tests.


pondrejk:
trigger: test-robottelo
pytest: tests/foreman/api/test_templatesync.py -k  test_positive_export_and_import_with_metadata

lhellebr (approved):
This fix is in accord with devs' recommendation in BZ2232162, LGTM. I suggest testing the other way (changing the path to `/var/lib/foreman/...`) as well.

Gauravtalreja1 (approved):

PR #12572, opened 2023-09-11
[6.14.z] change in import/export directory context

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/12571

Updating the fixture as BZ#2232162 evolved into documentation fixture. The new recommended way is to either use a pre-defined export dir or use a different context if custom one is needed. I chose the second option as the first one is already tested in bats tests.


Satellite-QE:
trigger: test-robottelo
pytest: tests/foreman/api/test_templatesync.py -k  test_positive_export_and_import_with_metadata

PR #12535, opened 2023-09-07
Fix test_negative_pre_upgrade_tuning_profile_check

jameerpathan111:
test_negative_pre_upgrade_tuning_profile_check is failing with `Unrecognised option '--disable-system-checks'`

PR #12578, opened 2023-09-12
Fix test_negative_pre_upgrade_tuning_profile_check

jameerpathan111:
Cherry-pick for https://github.com/SatelliteQE/robottelo/pull/12535 😅

Can't provide test results. Tests are marked to be skipped for the stream, as we still don't have upgrade support.

### DIFF
--- a/tests/foreman/maintain/test_upgrade.py
+++ b/tests/foreman/maintain/test_upgrade.py
@@ -19,7 +19,7 @@
 import pytest
 
 from robottelo.config import settings
-from robottelo.constants import SATELLITE_VERSION
+from robottelo.constants import INSTALLER_CONFIG_FILE, SATELLITE_VERSION
 
 
 def last_y_stream_version(release):
@@ -69,10 +69,6 @@ def test_positive_satellite_maintain_upgrade_list(sat_maintain):
 
 
 @pytest.mark.include_capsule
-@pytest.mark.skipif(
-    (settings.server.version.release == 'stream'),
-    reason='Upgrade path is not available for stream yet',
-)
 def test_positive_repositories_validate(sat_maintain):
     """Test repositories-validate pre-upgrade check is
      skipped when system is subscribed using custom activationkey.
@@ -113,10 +109,6 @@ def test_positive_repositories_validate(sat_maintain):
     ids=['default', 'medium'],
     indirect=True,
 )
-@pytest.mark.skipif(
-    (settings.server.version.release == 'stream'),
-    reason='Upgrade path is not available for stream yet',
-)
 def test_negative_pre_upgrade_tuning_profile_check(request, custom_host):
     """Negative test that verifies a satellite with less than
     tuning profile hardware requirements fails on pre-upgrade check.
@@ -140,10 +132,15 @@ def test_negative_pre_upgrade_tuning_profile_check(request, custom_host):
     )
     custom_host.download_repofile(product='satellite', release=last_y_stream)
     custom_host.execute('dnf -y module enable satellite:el8 && dnf -y install satellite')
-    # Install without system checks to get around installer checks
+    # Install with development tuning profile to get around installer checks
     custom_host.execute(
-        f'satellite-installer --scenario satellite --disable-system-checks --tuning {profile}',
+        'satellite-installer --scenario satellite --tuning development',
         timeout='30m',
+    )
+    # Change to correct tuning profile (default or medium)
+    custom_host.execute(
+        f'sed -i "s/tuning: development/tuning: {profile}/g" {INSTALLER_CONFIG_FILE};'
+        f'satellite-installer'
     )
     # Get current Satellite version's repofile
     custom_host.download_repofile(
@@ -154,10 +151,10 @@ def test_negative_pre_upgrade_tuning_profile_check(request, custom_host):
     custom_host.execute('satellite-maintain upgrade list-versions')
     # Check that we can upgrade to the new Y stream version
     result = custom_host.execute('satellite-maintain upgrade list-versions')
-    assert sat_version in result.stdout
+    assert SATELLITE_VERSION in result.stdout
     # Check that the upgrade check fails due to system requirements
     result = custom_host.execute(
-        f'satellite-maintain upgrade check --target-version {sat_version}', timeout='5m'
+        f'satellite-maintain upgrade check --target-version {SATELLITE_VERSION}', timeout='5m'
     )
     assert (
         f'ERROR: The installer is configured to use the {profile} tuning '
--- a/robottelo/constants/__init__.py
+++ b/robottelo/constants/__init__.py
@@ -1738,6 +1738,7 @@ VMWARE_CONSTANTS = {
 HAMMER_CONFIG = "~/.hammer/cli.modules.d/foreman.yml"
 HAMMER_SESSIONS = "~/.hammer/sessions"
 
+INSTALLER_CONFIG_FILE = '/etc/foreman-installer/scenarios.d/satellite.yaml'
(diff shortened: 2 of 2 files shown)


ogajduse:
We can not run PRT here since there is no support for 6.14 -> 6.15 upgrades.

ogajduse (approved):

tests/foreman/maintain/test_upgrade.py
```diff
@@ -146,11 +147,15 @@ def test_negative_pre_upgrade_tuning_profile_check(request, custom_host):
     )
     custom_host.download_repofile(product='satellite', release=last_y_stream)
     custom_host.execute('dnf -y module enable satellite:el8 && dnf -y install satellite')
-    # Install without system checks to get around installer checks
+    # Install with development tuning profile to get around installer checks
     custom_host.execute(
-        f'satellite-installer --scenario satellite --disable-system-checks --tuning {profile}',
+        'satellite-installer --scenario satellite --tuning development',
```
Gauravtalreja1, line 152: is development supported for downstream or is this being used for negative scenario?

tests/foreman/maintain/test_upgrade.py
```diff
@@ -146,11 +147,15 @@ def test_negative_pre_upgrade_tuning_profile_check(request, custom_host):
...
         timeout='30m',
     )
+    # Change to correct tuning profile (default or medium)
+    custom_host.execute(
+        f'sed -i "s/tuning: development/tuning: {profile}/g" {INSTALLER_CONFIG_FILE}'
+    )
```
Gauravtalreja1, line 158: Shouldn't we rerun installer after tuning profile in INSTALLER_CONFIG_FILE? I don't think it'll make any sense for pre-upgrade checks to run if we don't run installer?

tests/foreman/maintain/test_upgrade.py
lhellebr:
```
# satellite-installer --scenario satellite --tuning
ERROR: Invalid option supplied for --tuning. Please choose from one of: default, medium, large, extra-large, extra-extra-large, development

See: 'satellite-installer --help'
[root@ip-10-0-168-174 ~]# satellite-installer --scenario satellite --tuning development
2023-09-12 05:03:04 [NOTICE] [root] Loading installer configuration. This will take some time.
2023-09-12 05:03:09 [NOTICE] [root] Running installer with log based terminal output at level NOTICE.
2023-09-12 05:03:09 [NOTICE] [root] Use -l to set the terminal output log level to ERROR, WARN, NOTICE, INFO, or DEBUG. See --full-help for definitions.
[...]
```

tests/foreman/maintain/test_upgrade.py
lhellebr: I think the point of this test is to check whether the pre-upgrade checks catch the discrepancy. The question is, do the checks really use the config file? I'd suppose they do even though it's not specified explicitly. Otherwise, I think this test would fail on the last assert.

Also, this is just a cherrypick.

lhellebr (approved):

tests/foreman/maintain/test_upgrade.py
Gauravtalreja1: yeah, but in reality this should be an main PR and cherrypick got merged https://github.com/SatelliteQE/robottelo/pull/12535, and main PR is called cherrypick here, lol :D

jameerpathan111: yeah, cause I wanted to test the fix, which I can't do with master branch, so I raised it first against 6.14 🤷

jameerpathan111: and tbh I don't like the current approach of us doing TFA for 6.14 and raising a fix against master branch and then running into a bunch of the issues that we never faced in automation cause we never ran the test with stream :|

jameerpathan111: I don't think it's required to run installer, and like Lukas mentioned the assertion would fail.

jameerpathan111: But if we run installer, I think it'll fail.

jameerpathan111:
```
trigger: test-robottelo
pytest: tests/foreman/maintain/test_upgrade.py -k test_negative_pre_upgrade_tuning_profile_check
env:
    ROBOTTELO_server__deploy_arguments__deploy_sat_version: '6.14.0'
```

lpramuk:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_upgrade.py -k test_negative_pre_upgrade_tuning_profile_check

tests/foreman/maintain/test_upgrade.py
```diff
@@ -146,11 +147,16 @@ def test_negative_pre_upgrade_tuning_profile_check(request, custom_host):
     )
     custom_host.download_repofile(product='satellite', release=last_y_stream)
     custom_host.execute('dnf -y module enable satellite:el8 && dnf -y install satellite')
-    # Install without system checks to get around installer checks
+    # Install with development tuning profile to get around installer checks
```
lpramuk, line 150: if you are changing this test to use `development` profile the you should also handle tests parametrization ` ids=['default', 'medium'],` 

and all asserts for `{profile}`

lpramuk (changes requested):
Make your changes complete 
You switch over to development profile but you are not removing test parametrization `[default,medium`  and also you have to accommodate all assert for `{profile}` which is now not parametrized

tests/foreman/maintain/test_upgrade.py
lpramuk: Funny part is green PRT for this NEGATIVE test - failing on something else :D

jameerpathan111:
> > ```
> > trigger: test-robottelo
> > pytest: tests/foreman/maintain/test_upgrade.py -k test_negative_pre_upgrade_tuning_profile_check
> > env:
> >     ROBOTTELO_server__deploy_arguments__deploy_sat_version: '6.14.0'
> >     ROBOTTELO_server__deploy_arguments__deploy_snap_version: '16.0'
> >     ROBOTTELO_server__version__release: '6.14.0'
> > ```
> 
> Why are you testing against 6.14.0 while you choose not to cherrypick to `6.14.z` branch ? I would be for 6.14.z cherrypick
> 
> Green negative test in 6.14.z doesn't mean automatically it does its job right 😉

Because I can't test it with stream as it doesn't have upgrade support and because Gaurav wanted to see how the test would do with the changes he suggested. :D

Griffin-Sullivan:
We have the upgrade path in stream now so we can remove the skip and try on stream if you'd like

tests/foreman/maintain/test_upgrade.py
Griffin-Sullivan: We are only using the development profile to get around installer checks for system requirements like RAM and CPU cores. We have to do this in >= 6.14 because `--disable-system-checks` was removed. This test is still designed to make sure the default and medium profiles enforce their requirements upon an upgrade check.

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_upgrade.py -k test_negative_pre_upgrade_tuning_profile_check

Griffin-Sullivan:
The failure here is due to a bug that @ehelms is working on. Putting this in draft till it's fixed.

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_upgrade.py -k test_negative_pre_upgrade_tuning_profile_check

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_upgrade.py -k test_negative_pre_upgrade_tuning_profile_check

PR #12584, opened 2023-09-12
change in upgrade marks for hammer tests

pondrejk:
`test_positive_all_options` is consistently failing in the upgrade run as the set of enabled plugins on upgraded satellite is different than in the regular runs. Also it does not provide much value over existence tests.


lhellebr (commented):
`the set of enabled plugins on upgraded satellite is different than in the regular runs` - why? Is that ok?

Gauravtalreja1 (approved):

pondrejk:
@lhellebr I guess it is given by the historical record of upgrades, e.g. puppet used to be enabled by default in older versions of sat, so the upgraded images also have it enabled, while fresh installs have it off.

synkd (approved):

PR #12605, opened 2023-09-13
Fix regression from cloned manifest failover

synkd:
This PR adds a check to resolve a regression in `upload_manifest` introduced by https://github.com/SatelliteQE/robottelo/pull/12515.

### DIFF
--- a/robottelo/host_helpers/satellite_mixins.py
+++ b/robottelo/host_helpers/satellite_mixins.py
@@ -141,8 +141,9 @@ class ContentInfo:
         :returns: the manifest upload result
 
         """
-        if manifest.content is None:
-            manifest = clone()
+        if not isinstance(manifest, (bytes, io.BytesIO)):
+            if manifest.content is None:
+                manifest = clone()
         if timeout is None:
             # Set the timeout to 1500 seconds to align with the API timeout.
             timeout = 1500000


synkd:
trigger: test-robottelo
pytest: tests/foreman/api/test_subscription.py -k "test_positive_refresh or test_positive_create"

synkd:
The above PRT run tests two scenarios, one in which a manifester org fixture uploads a manifest via API and in one in which a test uploads a manifest via API. Assuming these pass, I'll trigger another PRT run to verify CLI upload functionality.

synkd:
trigger: test-robottelo
pytest: tests/foreman/api/test_subscription.py -k 'test_positive_refresh or test_positive_create'

synkd:
The API scenarios are passing.

synkd:
trigger: test-robottelo
pytest: tests/foreman/cli/test_contentaccess.py::test_negative_rct_not_shows_golden_ticket_enabled

synkd:
@jyejare The `upload_manifest` helper is basically a series of workarounds to account for the fact that we have multiple inconsistent ways of upload manifests. Here are the ones I'm aware of:
- The manifester org fixtures attempt to pass `manifest.content` to the `upload_manifest` helper, which then attempts to upload that as a manifest directly via the API. `manifest.content` is type `bytes`, so this scenario was tested by `tests/foreman/api/test_subscription.py::test_positive_refresh`, which uses the `function_entitlement_manifest_org` fixture. This is also that I overlooked with my previous PR (somewhat embarrassingly, as it's the most common scenario) and thus the scenario that was failing in the sanity tests. 
- Some tests, such as `tests/foreman/api/test_subscription.py::test_positive_create`, pass a manifest fixture (as opposed to a manifest org fixture) directly to the `upload_manifest` helper. This means that they pass a `Manifester` object to the helper. In this case, `manifest.content` is uploaded via API.
- Other tests (such as tests/foreman/cli/test_contentaccess.py::test_negative_rct_not_shows_golden_ticket_enabled) pass a manifest fixture to `upload_manifest` but specify the `CLI` interface. In these functions, the helper will check the attributes of the `manifest` argument to try to determine whether the manifest is a cloned manifest or not. This is necessary due to the differences in attributes between cloned manifest objects and manifester manifest objects. Once that determination is made, the helper uses `put` to copy the manifest file to the Satellite via SFTP and then upload it using the CLI.

Is this all a bit convoluted? Yes, I think so. We could simplify it somewhat by standardizing on passing either the whole manifest object or the content attribute of that object to `upload_manifest`, but if we did that, `upload_manifest` would still need to be able to account for the possibility of a failed manifest export from RHSM and for the difference between uploading via API vs CLI. I'm open to suggestions on that front, but in the interest of quickly fixing tests that use manifest org fixtures, I do think the current PR should be merged as-is.

robottelo/host_helpers/satellite_mixins.py
```diff
@@ -141,8 +141,9 @@ def upload_manifest(self, org_id, manifest=None, interface='API', timeout=None):
         :returns: the manifest upload result
 
         """
-        if manifest.content is None:
-            manifest = clone()
+        if not isinstance(manifest, (bytes, io.BytesIO)):
+            if manifest.content is None:
+                manifest = clone()
```
Gauravtalreja1, line 146: Can we instead check manifest has content atribute set using hasattr here? non-blocking
```suggestion
        if hasattr(manifest, 'content')
            manifest = clone()
```

synkd: This suggestion would substitute cloned manifests for manifester manifests in some situations where doing so is not necessary. For example: https://github.com/SatelliteQE/robottelo/blob/7a5840a9b83fa9fb0902c7a2ff13ec51b2b93677/tests/foreman/cli/test_contentaccess.py#L191
In this test, the object passed as the `manifest` argument to `upload_manifest` will have a `content` attribute if manifester succeeded and would then get swapped out for a cloned manifest at this point even with a valid manifest from manifester.

Gauravtalreja1: ah gotcha, my bad, suggested condition should be something like ` if hasattr(manifest, 'content') is False:`, so it can work when when manifest.content is passed as an argument to upload_manifest() and will know if manifest is passed which will have a content so won't clone it, right?

Gauravtalreja1 (approved):

jyejare:
Merging this on priority as the whole automation is stuck on the issue! The comments as improvements in `upload_manifest` function could be taken care later.

PR #12625, opened 2023-09-14
Removing Auto Release WF from non-master branch

jyejare:
Removing the WF for `Automatic Releases` from non-master branches as only master branch commits are considered in Github Releases!

This was been cherrypicked into 6.14.z during the Branching process which should have been removed !


ogajduse (approved):
as I mentioned in SatelliteQE/nailgun#995,
> If we have a documentation for branching, adding an extra step to remove the file from the newly branched .z branch would be a good idea.

jyejare:
@ogajduse We will soon have auto branching where we will make sure to delete it !

Gauravtalreja1 (approved):

Gauravtalreja1:
@jyejare As we need to remove it from all supported .z branches, shouldn't we cherrypick this commit onto those branches?

jyejare:
@Gauravtalreja1 It's only in 6.14.z branch now as it's introduced in 6.14.z time only.

PR #12647, opened 2023-09-18
Add test for satellite-maintain's self updating ability

Griffin-Sullivan:
Unstubbing this test and modifying the docstring to be more accurate and easy to understand what is being tested.

### DIFF
--- a/tests/foreman/maintain/test_upgrade.py
+++ b/tests/foreman/maintain/test_upgrade.py
@@ -171,27 +171,31 @@ def test_negative_pre_upgrade_tuning_profile_check(request, custom_host):
     )
 
 
-@pytest.mark.stubbed
 @pytest.mark.include_capsule
-def test_positive_self_update_for_zstream(sat_maintain):
-    """Test satellite-maintain self-upgrade to update maintain packages from zstream repo.
+def test_positive_self_update_maintain_package(sat_maintain):
+    """satellite-maintain attempts to update itself when a command is run
 
     :id: 1c566768-fd73-4fe6-837b-26709a1ebed9
 
     :parametrized: yes
 
     :steps:
-        1. Run satellite-maintain upgrade check/run command.
-        2. Run satellite-maintain upgrade check/run command with disable-self-upgrade option.
+        1. Run satellite-maintain upgrade list-versions/check/run command.
+        2. Run satellite-maintain upgrade list-versions/check/run command
+            with disable-self-upgrade option.
 
     :expectedresults:
-        1. Update satellite-maintain package to latest version and gives message to re-run command.
-        2. If disable-self-upgrade option is used then it should skip self-upgrade step for zstream
+        1. satellite-maintain tries to update rubygem-foreman_maintain to a newer available version
+        2. If disable-self-upgrade option is used then it should skip self-upgrade step
 
     :BZ: 1649329
-
-    :CaseAutomation: ManualOnly
     """
+    result = sat_maintain.cli.Upgrade.list_versions()
+    assert result.status == 0
+    assert 'Checking for new version of satellite-maintain...' in result.stdout
+    result = sat_maintain.cli.Upgrade.list_versions(options={'disable-self-upgrade': True})
+    assert result.status == 0
+    assert 'Checking for new version of satellite-maintain...' not in result.stdout
 
 
 @pytest.mark.stubbed


Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_upgrade.py::test_positive_self_update_maintain_package

synkd (commented):
Any thoughts on checking whether a package was actually updated when foreman-maintain reports that it did so? I realize that whether updates are available is going to depend on the state of our Satellite templates vs. what's in the repos, so this check would not be deterministic unless we did something like downgrade the foreman-maintain package before running the `list-versions` command. But I would like to see this test verify that foreman-maintain does what it says it does if it reports a change, rather than just relying on stdout.

Griffin-Sullivan:
@synkd I'll need to do some investigation to do this reliably. Will also need to be destructive.

lpramuk:
@Griffin-Sullivan package never gets updated as repo content always matches installed version.
Still your asserts regarding the message `Checking for new version of satellite-maintain...` are correct.

Griffin-Sullivan:
So you're saying we can't do what @synkd said? @lpramuk

lpramuk:
@Griffin-Sullivan do you want to test upgrade? -> upgrade-scenarios 

In component automation package never gets updated why would you check if it was upgraded ?

lpramuk (approved):

Griffin-Sullivan:
I actually found a place where we will reliably have a new version of foreman-maintain every time we list-versions so I'm adding an assertion there.

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_upgrade.py::test_positive_self_update_maintain_package tests/foreman/maintain/test_upgrade.py::test_negative_pre_upgrade_tuning_profile_check

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_upgrade.py::test_positive_self_update_maintain_package tests/foreman/maintain/test_upgrade.py::test_negative_pre_upgrade_tuning_profile_check

Griffin-Sullivan:
Also removing the skip for upgrade tests on stream, because we have the 6.15 upgrade path now.

tests/foreman/maintain/test_upgrade.py
```diff
@@ -171,27 +171,30 @@ def test_negative_pre_upgrade_tuning_profile_check(request, custom_host):
...
+    """satellite-maintain attempts to update itself when a command is run
 
     :id: 1c566768-fd73-4fe6-837b-26709a1ebed9
 
     :parametrized: yes
 
     :steps:
-        1. Run satellite-maintain upgrade check/run command.
-        2. Run satellite-maintain upgrade check/run command with disable-self-upgrade option.
+        1. Run satellite-maintain upgrade list-versions command.
```
Gauravtalreja1, line 183: self update runs in all of the upgrade subcommands
```suggestion
        1. Run satellite-maintain upgrade list-versions/check/run command.
```

tests/foreman/maintain/test_upgrade.py
```diff
@@ -171,27 +171,30 @@ def test_negative_pre_upgrade_tuning_profile_check(request, custom_host):
...
+        2. Run satellite-maintain upgrade list-versions command with disable-self-upgrade option.
```
Gauravtalreja1, line 184:
```suggestion
        2. Run satellite-maintain upgrade list-versions/check/run command with disable-self-upgrade option.
```

tests/foreman/maintain/test_upgrade.py
```diff
@@ -171,27 +171,30 @@ def test_negative_pre_upgrade_tuning_profile_check(request, custom_host):
     )
 
 
-@pytest.mark.stubbed
 @pytest.mark.include_capsule
-def test_positive_self_update_for_zstream(sat_maintain):
-    """Test satellite-maintain self-upgrade to update maintain packages from zstream repo.
+def test_positive_self_update_maintain_package(sat_maintain):
```
Gauravtalreja1, line 175: so there's confusion in f-m package updates, major version package upgrade is called self-upgrade, so to avoid confusion, can we add for zstream back? 
```suggestion
def test_positive_self_update_maintain_package_for_zstream(sat_maintain):
```

tests/foreman/maintain/test_upgrade.py
```diff
@@ -171,27 +171,30 @@ def test_negative_pre_upgrade_tuning_profile_check(request, custom_host):
...
-        1. Run satellite-maintain upgrade check/run command.
-        2. Run satellite-maintain upgrade check/run command with disable-self-upgrade option.
+        1. Run satellite-maintain upgrade list-versions command.
+        2. Run satellite-maintain upgrade list-versions command with disable-self-upgrade option.
 
     :expectedresults:
-        1. Update satellite-maintain package to latest version and gives message to re-run command.
-        2. If disable-self-upgrade option is used then it should skip self-upgrade step for zstream
+        1. satellite-maintain tries to update to a new version
+        2. If disable-self-upgrade option is used then it should skip self-upgrade step
```
Gauravtalreja1, line 188:
```suggestion
        2. If disable-self-upgrade option is used then it should skip self-update step
```

tests/foreman/maintain/test_upgrade.py
```diff
@@ -171,27 +171,30 @@ def test_negative_pre_upgrade_tuning_profile_check(request, custom_host):
...
     :steps:
-        1. Run satellite-maintain upgrade check/run command.
-        2. Run satellite-maintain upgrade check/run command with disable-self-upgrade option.
+        1. Run satellite-maintain upgrade list-versions command.
+        2. Run satellite-maintain upgrade list-versions command with disable-self-upgrade option.
 
     :expectedresults:
-        1. Update satellite-maintain package to latest version and gives message to re-run command.
-        2. If disable-self-upgrade option is used then it should skip self-upgrade step for zstream
+        1. satellite-maintain tries to update to a new version
```
Gauravtalreja1, line 187:
```suggestion
        1. satellite-maintain tries to update rubygem-foreman_maintain to a newer available version
```

Gauravtalreja1 (commented):
Ack, with a few suggestions

tests/foreman/maintain/test_upgrade.py
Griffin-Sullivan: I'm opting to leave this as self-upgrade since that is what the option is called

tests/foreman/maintain/test_upgrade.py
Griffin-Sullivan: well technically the auto update could run z-stream or y-stream upgrade for foreman-maintain. The self-upgrade command only applies to y-stream, while auto-update can do any of the upgrades.

sambible (approved):
I agree with your responses to Gaurav, so I'm going to approve and merge

sambible:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_upgrade.py::test_positive_self_update_maintain_package

tests/foreman/maintain/test_upgrade.py
Gauravtalreja1: well, they are not same, for y-stream self-upgrade enables next version repository and use `dnf --enablerepo..` to update the maintain packages, and for auto-update its just updates maintain packages if its available in the enabled repo. Let me know wdyt, I don't want to block this PR just for this.

PR #12649, opened 2023-09-19
Alternate Content Source UI tests

LadislavVasina1:
In this PR I have implemented two tests. One end-to-end UI test tests the creation of some ACS types and actions on them and the second one tests the creation of all possible ACS types.

### DIFF
--- a/tests/foreman/ui/test_acs.py
+++ b/tests/foreman/ui/test_acs.py
@@ -23,7 +23,6 @@ from robottelo.constants import REPO_TYPE
 from robottelo.constants.repos import CUSTOM_FILE_REPO
 from robottelo.utils.datafactory import gen_string
 
-
 ssl_name, product_name, product_label, product_description, repository_name = (
     gen_string('alpha') for _ in range(5)
 )
--- a/tests/foreman/ui/test_acs.py
+++ b/tests/foreman/ui/test_acs.py
@@ -78,21 +78,22 @@ def test_acs_positive_end_to_end(session, ui_acs_setup):
     :id: 047452cc-5a9f-4473-96b1-d5b6830b7d6b
 
     :steps:
-    1. Select an organization
-    2. Create ACSes (all supported types/combinations)
-    3. Create ACS on which deletion is going to be tested
-    4. Test deletion
-    5. Test refresh
-    6. Test renaming and changing description
-    7. Test editing capsules
-    8. Test editing urls and subpaths
-    9. Test editing credentials
-    10. Test editing products
+        1. Select an organization
+        2. Create ACSes (all supported types/combinations)
+        3. Create ACS on which deletion is going to be tested
+        4. Test deletion
+        5. Test refresh
+        6. Test renaming and changing description
+        7. Test editing capsules
+        8. Test editing urls and subpaths
+        9. Test editing credentials
+        10. Test editing products
 
 
-    :expectedresults: This test should create all supported types (10)
-    of the Aleternate Content Sources one by one and asserts that actions
-    were made correctly on them.
+    :expectedresults:
+        This test should create all supported types (10)
+        of the Aleternate Content Sources one by one and asserts that actions
+        were made correctly on them.
     """
 
     target_sat, module_sca_manifest_org = ui_acs_setup
--- a/tests/foreman/ui/test_acs.py
+++ b/tests/foreman/ui/test_acs.py
@@ -78,7 +78,7 @@ class TestAllAcsTypes:
         """
         This test creates all possible ACS types.
 
-        :id cbd0f4e6-2151-446a-90d3-69c6935a0c91
+        :id: cbd0f4e6-2151-446a-90d3-69c6935a0c91
 
         :parametrized: yes
 
--- a/tests/foreman/ui/test_acs.py
+++ b/tests/foreman/ui/test_acs.py
@@ -30,7 +30,7 @@ repos_to_enable = ['rhae2.9_el8']
 
 
 @pytest.fixture(scope='class')
-def acs_setup(class_target_sat, module_sca_manifest_org):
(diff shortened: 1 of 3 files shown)


LadislavVasina1:
trigger: test-robottelo
pytest: -k test_acs_positive_end_to_end tests/foreman/ui/test_acs.py
airgun: 977

ogajduse (approved):
LGTM code-wise, but not an ACS expert though.

LadislavVasina1:
trigger: test-robottelo
pytest: -k test_acs_positive_end_to_end tests/foreman/ui/test_acs.py
airgun: 977

vsedmik (approved):
ACK, the test just does what I would expect it to do - tests the UI.
Good job @LadislavVasina1 !

LadislavVasina1:
trigger: test-robottelo
pytest: -k test_acs_positive_end_to_end tests/foreman/ui/test_acs.py
airgun: 977

tests/foreman/ui/test_acs.py
```diff
@@ -0,0 +1,357 @@
...
+        2. Create ACSes (all supported types/combinations)
+        3. Create ACS on which deletion is going to be tested
+        4. Test deletion
+        5. Test refresh
+        6. Test renaming and changing description
+        7. Test editing capsules
+        8. Test editing urls and subpaths
+        9. Test editing credentials
+        10. Test editing products
+
```
pondrejk, line 91: Codewise I see no problems but I'm not convinced of this all being in a single test. If there is a bug in creation of just of the types in step 2 we never get to testing other things, with 10 types it is not so improbable.

IIUC the ACSes in edit and renamed steps are just selected randomly, and there are special ones created to test deletion anyway. 

My suggestion is to have a separate test for step 2 -- `check all types can be created`. 

And then have a separate `e2e test` for the other steps, just move the deletion part at the very end. 

Bonus suggestion: consider what it would take to have ACS configuration for the individual types defined in some structure and then use that to parametrize the tests. For the `check all types can be created` test it would ensure all types are always tested (currently, if the first one fails others are not visited). For the `e2e` test, we could pick one or more types for coverage, or a random one perhaps :)

LadislavVasina1:
"trigger": "test-robottelo"
"pytest": "tests/foreman/ui/test_acs.py -k 'test_check_all_acs_types_can_be_created or test_acs_positive_end_to_end'"

LadislavVasina1:
trigger: test-robottelo
pytest: tests/foreman/ui/test_acs.py -k 'test_check_all_acs_types_can_be_created or test_acs_positive_end_to_end'

tests/foreman/ui/test_acs.py
```diff
@@ -0,0 +1,497 @@
...
+                            base_url='https://test.com',
+                            subpaths=['test/'],
+                            manual_auth=True,
+                            username='test',
+                            password='test',
+                            verify_ssl=True,
+                            ca_cert=ssl_name,
+                        )
+
+                case 'rhui':
```
rplevka, line 243: this one is not `cnt_type` specific. Does it mean that it would run exactly the same test for each `cnt_type` parameter?
Isn't that a sign of incorrect design?

LadislavVasina1: `file` content type is not allowed for `rhui` thus it is skipped at the beginning of the test and only `yum` test is being run with `rhui` acs type

rplevka (changes requested):
I would suggest to re-do the structure of the test, to separate the parametrization logic from the test body:

you define a function to generate the payload data and nice test id strings:
```
def gen_params():
    a = {
        '_common': {
            'use_http_proxies': True,
        },
        'custom': {
            '_common': {
                'custom_type': True,
                'base_url': 'https://test.com/',
                'subpaths': ['test/'],
            },
            'yum': {
                'content_type': 'yum'
            },
            'file': {}
        },
        'simplified': {
            '_common': {
                'simplified_type': True
            },
            'yum': {
                'content_type': 'yum',
                'name': 'simpleYum',
                'description': 'simpleYumDesc',
                'capsules_to_add': class_target_sat.hostname,
                'products_to_add': [constants.REPOS[repo]['product'] for repo in repos_to_enable]
            },
            'file': {
                'content_type': 'file',
                'name': 'simpleFile',
                'description': 'simpleFileDesc',
                'add_all_capsules': True,
                'products_to_add': product_name,
            },
        },
        'rhui': {
            'yum': {
                'rhui_type': True,
                'name': 'rhuiYumNoneAuth',
                'base_url': 'https://test.com/pulp/content',
                'subpaths': ['test/', 'test2/'],
            },
        }
    }
    ids = []
    vals = []
    for acs in a.keys():
        if not acs.startswith('_'):
            for cnt in a[acs]:
                if not cnt.startswith('_'):
                    scenario = a[acs][cnt] | a.get('_common', {}) | a[acs].get('_common', {})   
                    ids.append(f'{acs}_{cnt}')
                    vals.append(scenario)
    return (vals, ids)
```

and here's how you use it in your test
```
@pytest.mark.parametrize('scenario, gen_params()[0], ids=gen_params()[1])    # you put a call to the function into the parametrize call like so
def test_param(scenario):
    ....
    session.acs.create_new_acs(**scenario)                 # and you access the function payload per current scenario like so
    pass
```

tests/foreman/ui/test_acs.py
rplevka: ok, i just asked so i know if these parameters can be used in all scenarios for the purpose of my PoC.

LadislavVasina1:
trigger: test-robottelo
pytest: tests/foreman/ui/test_acs.py

LadislavVasina1:
trigger: test-robottelo
pytest: tests/foreman/ui/test_acs.py

PR #12483, opened 2023-09-05
Fix : ZStream Dependency Auto Merging

jyejare:
Fixing the AutoMerge of Dependabot AutoCherrypicked PRs are not running.

Attempt to change:
- The `pull_request_target` event
- The Author check for PR now removed and just checking for `dependency` label
- Separate GHA for zStream than master as the things varies for zStream for dependecy PRs auto-merge

Good if this is merged along with https://github.com/SatelliteQE/robottelo/pull/12682 that copies dependencies label to zStream cherrypicked PRs !

PR #12682, opened 2023-09-21
Copy dependencies label for dependencies PRs

jyejare:
AutoCherryPick is currently missed adding `dependencies` label to the AutoCherrypicked Dependabot PRs and hence dependabot zStream cherrypicked PRs are not automerging as its missing the label for `dependencies`.

There is another reason its missing automerging the zStream PRs for which the PR is already raised : #12483 !

Verifed its working in my local testing repository:

- The GHA run finding and setting dependencies label: https://github.com/jyejare/test-acp/actions/runs/6260767022/job/16999485038?pr=17 
- The PR opened with dependencies label since parent PR has dependencies label: https://github.com/jyejare/test-acp/pull/18

### DIFF
--- a/.github/workflows/auto_cherry_pick.yml
+++ b/.github/workflows/auto_cherry_pick.yml
@@ -10,6 +10,7 @@ env:
   assignee: ${{ github.event.pull_request.assignee.login }}
   title: ${{ github.event.pull_request.title }}
   number: ${{ github.event.number }}
+  is_dependabot_pr: ''
 
 jobs:
 
@@ -52,6 +53,12 @@ jobs:
         with:
           fetch-depth: 0
 
+      ## Set env var for dependencies label PR
+      - name: Set env var is_dependabot_pr to `dependencies` to set the label
+        if: contains(github.event.pull_request.labels.*.name, 'dependencies')
+        run: |
+          echo "is_dependabot_pr=dependencies" >> $GITHUB_ENV
+
       ## CherryPicking and AutoMerging
       - name: Cherrypicking to zStream branch
         id: cherrypick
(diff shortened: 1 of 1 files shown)


.github/workflows/auto_cherry_pick.yml
```diff
@@ -10,6 +10,7 @@ env:
   assignee: ${{ github.event.pull_request.assignee.login }}
   title: ${{ github.event.pull_request.title }}
   number: ${{ github.event.number }}
+  is_dependencies: ''
```
omkarkhatavkar, line 13: `is_dependabot_pr` would be nice wording.

.github/workflows/auto_cherry_pick.yml
```diff
@@ -52,6 +53,12 @@ jobs:
         with:
           fetch-depth: 0
 
+      ## Set env var for dependencies label PR
+      - name: Set env var is_dependency to `dependencies` to set the label
+        if: contains(github.event.pull_request.labels.*.name, 'dependencies')
```
omkarkhatavkar, line 58: maybe good to check if the PR author is dependabot, then add the `dependencies` label.

jyejare: Not sure why its so important to check author for dependencies. Anyone with immediate need want to open a PR for bumping the dependency version can also tag PRs with just `dependencies` for automerge !

We otherwise dont tag PRs with `dependencies` label !

omkarkhatavkar: I was thinking just one more layer of checking, It's non-blocking.

jyejare:
@omkarkhatavkar Should we merge then ?

omkarkhatavkar (approved):

PR #12686, opened 2023-09-21
Switch import formatting to use Ruff

JacobCallahan:
This now drops the old standard of one import per line in favor of a less noisy standard (isort).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,13 +23,22 @@ select = [
     # "C90", # mccabe
     "E",  # pycodestyle
     "F",  # flake8
+    "I", # isort
     # "Q",  # flake8-quotes
     "UP",  # pyupgrade
     "W",  # pycodestyle
 ]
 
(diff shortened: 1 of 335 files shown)


pyproject.toml
ogajduse, line 1: What do you think about keeping the isort config the same as in broker?
https://github.com/SatelliteQE/broker/blob/dd2471729379f06cb7c809dfbbc116ed5e5a91e7/pyproject.toml#L174-L179

ogajduse: I also see that I duplicated the `tool.black` section. Can you please remove it?

JacobCallahan: good point. will update and see what happens

omkarkhatavkar:
trigger: test-robottelo

ogajduse (approved):
PRT passed :heavy_check_mark:

omkarkhatavkar (approved):

PR #12691, opened 2023-09-21
Remove stubbed FM case that is no longer needed

Griffin-Sullivan:
I'm removing this stubbed case because the steps for updating satellite-maintain here are outdated. We now rely on satellite-maintain's ability to self-upgrade either by running a command like `satellite-maintain upgrade list-versions` which is being tested in https://github.com/SatelliteQE/robottelo/pull/12647 or by using `satellite-maintain self-upgrade`. Dev is planning to make some changes to these commands as well which will get there own automation when it's ready. For now we can just remove this outdated stubbed case.

### DIFF
--- a/tests/foreman/maintain/test_packages.py
+++ b/tests/foreman/maintain/test_packages.py
@@ -273,22 +273,3 @@ def test_positive_fm_packages_update(request, sat_maintain):
     def _finalize():
         assert sat_maintain.execute('dnf remove -y walrus').status == 0
         sat_maintain.execute('rm -rf /etc/yum.repos.d/custom_repo.repo')
-
-
-@pytest.mark.stubbed
-def test_positive_fm_packages_sat_installer(sat_maintain):
-    """Verify satellite-installer is not executed after install/update
-    of satellite-maintain/rubygem-foreman_maintain package
-
-    :id: d73971a1-68b4-4ab2-a87c-76cc5ff80a39
-
-    :steps:
-        1. satellite-maintain packages install/update satellite-maintain/rubygem-foreman_maintain
-
-    :BZ: 1825841
-
-    :expectedresults: satellite-installer shouldn't be executed after install/update
-        of satellite-maintain/rubygem-foreman_maintain package
-
-    :CaseAutomation: ManualOnly
-    """


tests/foreman/maintain/test_packages.py
```diff
@@ -273,22 +273,3 @@ def test_positive_fm_packages_update(request, sat_maintain):
...
         assert sat_maintain.execute('dnf remove -y walrus').status == 0
         sat_maintain.execute('rm -rf /etc/yum.repos.d/custom_repo.repo')
-
-
-@pytest.mark.stubbed
-def test_positive_fm_packages_sat_installer(sat_maintain):
-    """Verify satellite-installer is not executed after install/update
-    of satellite-maintain/rubygem-foreman_maintain package
-
-    :id: d73971a1-68b4-4ab2-a87c-76cc5ff80a39
```
Gauravtalreja1, line 283: why removing this test, could you add some description?

Griffin-Sullivan: yeah sorry I'll update the description.

tests/foreman/maintain/test_packages.py
```diff
@@ -273,22 +273,3 @@ def test_positive_fm_packages_update(request, sat_maintain):
...
-
-    :steps:
-        1. satellite-maintain packages install/update satellite-maintain/rubygem-foreman_maintain
-
-    :BZ: 1825841
-
-    :expectedresults: satellite-installer shouldn't be executed after install/update
-        of satellite-maintain/rubygem-foreman_maintain package
-
```
Gauravtalreja1, line 292: I remember I've added this test in past, it verifies installer doesn't run after we install/update mentioned package, because of an entry in whitelist packages in /etc/yum/pluginconf.d/foreman-protector.whitelist
And IMO, this is unrelated to self-update or self-upgrade workflow, so it doesn't make sense to remove this test in favour of that test.

Gauravtalreja1 (changes requested):

tests/foreman/maintain/test_packages.py
Griffin-Sullivan: This is an outdated way of updating the satellite-maintain package. We no longer tell customers to do this in our documentation. I don't see the need to have a stubbed case in our automation if users aren't supposed to do this. I am willing to at least remove the cherry-pick to other branches and just do this for master if you prefer that.

Gauravtalreja1: Yeah, this isn't a documented way, it was before self-update or self-upgrade features existed iirc, so that's what BZ was about. We still support this way of update too and customers can try this anyway, so I think there's no harm if we keep this stubbed test OR this can be easily automated just by checking installer doesn't run after package update

ehelms:
As this is a manual test, I vote for removal.

sambible (commented):
`We still support this way of update`
If it's not in our documentation is this true? How do we support it if not within the documentation?

Gauravtalreja1:
It's not directly documented now, but current self-update and self-upgrade mechanism uses the same thing with f-m's package_manager and foreman-protector plugin.
Originally it was added as stubbed in TestFM where this wasn't possible to automate, but now it can be automated with robottelo with steps to verify installer don't run when update maintain packages with f-m CLI. 
And, I'll let @Griffin-Sullivan and his team decide, how they'd like to handle this automation or removal as they are owning it now.

PR #12706, opened 2023-09-22
Fixes for stream, 6.14.z; failure ErrataManagement in api:test_errata

damoore044:
Had to enable custom repos, which are disabled by default. 

Repos were not available to the organization or contenthost, so I updated to use `function_org` and `function_lce`,
setting up the org and repos with those fixtures, using the method `setup_org_for_a_custom_repo`.

* Added a quick `sleep()` before asserting applicable_errata_count, 
   as it seems there is some delay after errata install task is done, especially in prt runs.

### DIFF
--- a/tests/foreman/api/test_errata.py
+++ b/tests/foreman/api/test_errata.py
@@ -17,6 +17,7 @@
 :Upstream: No
 """
 # For ease of use hc refers to host-collection throughout this document
+from copy import copy
 from time import sleep
 
 from nailgun import entities
@@ -158,11 +159,11 @@ def test_positive_install_in_hc(module_org, activation_key, custom_repo, target_
 
 
 @pytest.mark.tier3
-@pytest.mark.rhel_ver_list([7, 8, 9])
+@pytest.mark.rhel_ver_match(r'^(?!6$)\d+$')
 @pytest.mark.no_containers
 @pytest.mark.e2e
 def test_positive_install_multiple_in_host(
-    module_org, activation_key, custom_repo, rhel_contenthost, target_sat
+    function_org, rhel_contenthost, target_sat, function_lce
 ):
     """For a host with multiple applicable errata install one and ensure
     the rest of errata is still available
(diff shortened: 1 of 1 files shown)


damoore044:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py -k test_positive_install_multiple_in_host

damoore044:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py -k test_positive_install_multiple_in_host

damoore044:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py -k test_positive_install_multiple_in_host

damoore044:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py -k test_positive_install_multiple_in_host

damoore044:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py -k test_positive_install_multiple_in_host

damoore044:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py -k test_positive_install_multiple_in_host

tests/foreman/api/test_errata.py
```diff
@@ -183,31 +184,53 @@ def test_positive_install_multiple_in_host(
...
             },
         )['id']
         target_sat.wait_for_tasks(
             search_query=(f'label = Actions::RemoteExecution::RunHostsJob and id = {task_id}'),
             search_rate=20,
             max_tries=15,
         )
-        applicable_errata_count -= 1
-        assert rhel_contenthost.applicable_errata_count == applicable_errata_count
+        sleep(10)
```
sambible, line 232: What's the reasoning for this sleep? Can we do a wait_for on the assert instead? Sleep shouldn't be necessary here.

sambible (changes requested):

tests/foreman/api/test_errata.py
damoore044: A wait_for on the assert was consistently failing prt, and inconsistent fail locally. 
I had tried something like this :
`wait_for(lambda: rhel_ch.applicable_errata_count == app_err_count - 1, timeout=600, delay=10)`
But would occasionally fail to complete lambda in time?

damoore044: The sleep method was the most consistent way for prt to pass. Two methods doing similar tasks in api:errata, `_fetch_available_errata` and `_validate_errata_counts` are using `sleep()` before getting counts

ianballou: Since it looks like this task is installing an errata and then immediately checking the new errata status, I think the sleep makes sense since it could take some time for that whole orchestration to finish.

The host errata install task completes, so the info should also be reported back, but the applicable errata calculation is asynchronous and may not be complete.

The best solution would be to poll on the errata applicability calculation, but we may not have the data to determine exactly which task to poll.

ianballou: And I mean sleep or some sort of delay, I'm not sure what the `wait_for` is that @sambible mentioned.

sambible: I'm fine in this case then. Sleep can be a decent tool in certain cases, and it sounds like this one is in that situation.

sambible (approved):

tests/foreman/api/test_errata.py
damoore044: Thanks for the insight @ianballou! `wait_for` is a method we use that will fail if a condition `lambda:` could not be met within a specified timeout (seconds), delay is the time between each check. 

Lots of tests in robottelo use this and I did so recently in `report_templates:installable_errata` test here:
https://github.com/SatelliteQE/robottelo/blob/0ddb4b712f659b78791e5a6437dca4bf140e8ce6/tests/foreman/api/test_reporttemplates.py#L809

tests/foreman/api/test_errata.py
```diff
@@ -158,11 +159,11 @@ def test_positive_install_in_hc(module_org, activation_key, custom_repo, target_
 
 
 @pytest.mark.tier3
-@pytest.mark.rhel_ver_list([7, 8, 9])
+@pytest.mark.rhel_ver_match(r'^(?!6$)\d+$')
```
Griffin-Sullivan, line 162: Can you translate this for me?

synkd (approved):

vsedmik (approved):

tests/foreman/api/test_errata.py
damoore044: @Griffin-Sullivan this was recommended in other parameterized tests I've written.
It will automatically select any version higher than rhel6, so it should find 10 once it is out!
I admit it's a bit difficult on the eyes xD

damoore044: @sambible oops, you are right thanks!

ColeHiggins2 (approved):

tests/foreman/api/test_errata.py
Griffin-Sullivan: I'm not 100% sure, but I think you could also do it like this but just replace the 6 with a 7. https://github.com/SatelliteQE/robottelo/blob/0ddb4b712f659b78791e5a6437dca4bf140e8ce6/tests/foreman/ui/test_repositories.py#L22

Either way, I see other instances like this, so no harm in doing it this way.

Griffin-Sullivan (approved):

damoore044:
Got merged before I pushed updated to use the preferred `thel_contenthost.register( )` method.
Will open a follow-up pr for that soon

tests/foreman/api/test_errata.py
damoore044: @Griffin-Sullivan I like your way a lot better! I'll push to this pr #12851

PR #12722, opened 2023-09-25
Workaround nonexistent auth source in upgrade template

lhellebr:


ColeHiggins2 (approved):
Looks good to me, lets get a prt test going

adarshdubey-star (approved):

PR #12764, opened 2023-09-27
Missing content ID test

sambible:
Simple test for a pretty complicated BZ. Basically, this creates a repository and then deletes the Upstream URL, which in the BZ case triggered the missing content ID issues. It then goes on to make sure you can both update the URL, and delete the repo, cases which were impacted in the BZ. 

The sleep at the beginning is a simple solution to a bit of a weird problem, where when creating a repo, an asynchronous task to generate the metadata is kicked off. If we adjust the URL before that's done, it will collide with that task and error out. Sleeping for 5s gets around this case.

### DIFF
--- a/tests/foreman/api/test_repository.py
+++ b/tests/foreman/api/test_repository.py
@@ -1554,6 +1554,38 @@ class TestRepositorySync:
         )
         assert len(os)
 
+    @pytest.mark.tier2
+    @pytest.mark.parametrize(
+        'repo_options',
+        **datafactory.parametrized(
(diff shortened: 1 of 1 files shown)


JacobCallahan (approved):

ColeHiggins2:
trigger: test-robottelo
pytest: tests/foreman/api/test_repository.py -k test_missing_content_id

ColeHiggins2 (approved):
ack pending results

Griffin-Sullivan (approved):

PR #12765, opened 2023-09-27
Add in the caching mechanism to the Satellite's api property

JacobCallahan:
This should have always existed but was overlooked. It will dramatically decrease the amount of time it takes to access the api property since it doesn't need to be reconstructed each time the property is accessed.

Before
------
%timeit mysat.api.Host
3.62 ms ± 52.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

After
-----
%timeit mysat.api.Host
189 ns ± 0.943 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -1735,6 +1735,7 @@ class Capsule(ContentHost, CapsuleMixins):
                     except AttributeError:
                         # not everything has an mro method, we don't care about them
                         pass
+        self._cli._configured = True
         return self._cli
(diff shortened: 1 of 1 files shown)


jyejare:
trigger: test-robottelo
pytest: -m build_sanity tests/foreman/

omkarkhatavkar (approved):

jyejare (approved):

PR #12796, opened 2023-10-03
Add Ruff pytest standards

JacobCallahan:
This is a big one! Most of the changes are automatic, but a number of these are manual. I've deselected rules relating to fixture naming, but we can have a conversation later about whether we want to adopt the underscore naming conventions for fixtures.

### DIFF
--- a/pytest_fixtures/component/satellite_auth.py
+++ b/pytest_fixtures/component/satellite_auth.py
@@ -35,7 +35,7 @@ def default_ipa_host(module_target_sat):
     return IPAHost(module_target_sat)
 
 
-@pytest.fixture()
+@pytest.fixture
 def ldap_cleanup():
     """this is an extra step taken to clean any existing ldap source"""
     ldap_auth_sources = entities.AuthSourceLDAP().search()
@@ -44,7 +44,7 @@ def ldap_cleanup():
         for user in users:
             user.delete()
         ldap_auth.delete()
-    yield
+    return
 
 
 @pytest.fixture(scope='session')
@@ -104,7 +104,7 @@ def open_ldap_data():
     }
 
 
-@pytest.fixture(scope='function')
+@pytest.fixture
 def auth_source(ldap_cleanup, module_org, module_location, ad_data):
     ad_data = ad_data()
     return entities.AuthSourceLDAP(
@@ -127,7 +127,7 @@ def auth_source(ldap_cleanup, module_org, module_location, ad_data):
     ).create()
 
 
-@pytest.fixture(scope='function')
+@pytest.fixture
 def auth_source_ipa(ldap_cleanup, default_ipa_host, module_org, module_location):
     return entities.AuthSourceLDAP(
         onthefly_register=True,
@@ -149,7 +149,7 @@ def auth_source_ipa(ldap_cleanup, default_ipa_host, module_org, module_location)
     ).create()
 
 
-@pytest.fixture(scope='function')
+@pytest.fixture
 def auth_source_open_ldap(ldap_cleanup, module_org, module_location, open_ldap_data):
     return entities.AuthSourceLDAP(
         onthefly_register=True,
@@ -259,7 +259,7 @@ def ldap_auth_source(
     else:
         ldap_data['server_type'] = LDAP_SERVER_TYPE['UI']['posix']
         ldap_data['attr_login'] = LDAP_ATTR['login']
-    yield ldap_data, auth_source
+    return ldap_data, auth_source
 
 
 @pytest.fixture
@@ -459,7 +459,7 @@ def configure_hammer_no_negotiate(parametrized_enrolled_sat):
 
 
 @pytest.mark.external_auth
-@pytest.fixture(scope='function')
+@pytest.fixture
 def hammer_logout(parametrized_enrolled_sat):
     """Logout in Hammer."""
     result = parametrized_enrolled_sat.cli.Auth.logout()
--- a/tests/foreman/api/test_bookmarks.py
+++ b/tests/foreman/api/test_bookmarks.py
@@ -83,7 +83,7 @@ def test_positive_create_with_query(controller):
 
 
 @pytest.mark.tier1
-@pytest.mark.parametrize('public', (True, False))
+@pytest.mark.parametrize('public', [True, False])
 @pytest.mark.parametrize('controller', CONTROLLERS)
 def test_positive_create_public(controller, public):
     """Create a public bookmark
@@ -368,7 +368,7 @@ def test_negative_update_empty_query(controller):
 
 
 @pytest.mark.tier1
-@pytest.mark.parametrize('public', (True, False))
+@pytest.mark.parametrize('public', [True, False])
 @pytest.mark.parametrize('controller', CONTROLLERS)
 def test_positive_update_public(controller, public):
     """Update a bookmark public state to private and vice versa
--- a/tests/foreman/api/test_capsulecontent.py
+++ b/tests/foreman/api/test_capsulecontent.py
@@ -1177,7 +1177,7 @@ class TestCapsuleContentManagement:
         result = module_capsule_configured.nailgun_capsule.content_lifecycle_environments()
         # there can and will be LCEs from other tests and orgs, but len() >= 2
         assert len(result['results']) >= 2
-        assert lce1.id and lce2.id in [capsule_lce['id'] for capsule_lce in result['results']]
+        assert {lce1.id, lce2.id}.issubset([capsule_lce['id'] for capsule_lce in result['results']])
 
         # Create a Content View, add the repository and publish it.
         cv = target_sat.api.ContentView(
--- a/tests/foreman/api/test_classparameters.py
+++ b/tests/foreman/api/test_classparameters.py
@@ -137,10 +137,10 @@ class TestSmartClassParameters:
             2. Error raised for invalid default value.
         """
         sc_param = module_puppet['sc_params'].pop()
+        sc_param.override = True
+        sc_param.parameter_type = test_data['sc_type']
+        sc_param.default_value = test_data['value']
         with pytest.raises(HTTPError) as context:
-            sc_param.override = True
-            sc_param.parameter_type = test_data['sc_type']
-            sc_param.default_value = test_data['value']
             sc_param.update(['override', 'parameter_type', 'default_value'])
         assert sc_param.read().default_value != test_data['value']
         assert 'Validation failed: Default value is invalid' in context.value.response.text
@@ -370,9 +370,9 @@ class TestSmartClassParameters:
         session_puppet_enabled_sat.api.OverrideValue(
             smart_class_parameter=sc_param, match='domain=example.com', value=gen_string('alpha')
         ).create()
+        sc_param.parameter_type = 'boolean'
+        sc_param.default_value = gen_string('alpha')
         with pytest.raises(HTTPError) as context:
-            sc_param.parameter_type = 'boolean'
-            sc_param.default_value = gen_string('alpha')
             sc_param.update(['parameter_type', 'default_value'])
         assert (
             'Validation failed: Default value is invalid, Lookup values is invalid'
--- a/tests/foreman/api/test_http_proxy.py
+++ b/tests/foreman/api/test_http_proxy.py
@@ -293,7 +293,7 @@ def test_positive_sync_proxy_with_certificate(request, target_sat, module_org, m
     # Create and fetch new cerfiticate
     target_sat.custom_cert_generate(proxy_host)
     cacert = target_sat.execute(f'cat {cacert_path}').stdout
-    assert 'BEGIN CERTIFICATE' and 'END CERTIFICATE' in cacert
+    assert 'END CERTIFICATE' in cacert
 
     # Create http-proxy and repository
     http_proxy = target_sat.api.HTTPProxy(
--- a/tests/foreman/api/test_media.py
+++ b/tests/foreman/api/test_media.py
@@ -41,7 +41,7 @@ class TestMedia:
     @pytest.mark.tier1
     @pytest.mark.upgrade
     @pytest.mark.parametrize(
-        'name, new_name',
+        ('name', 'new_name'),
         **parametrized(list(zip(valid_data_list().values(), valid_data_list().values())))
     )
     def test_positive_crud_with_name(self, module_org, name, new_name):
--- a/tests/foreman/api/test_repositories.py
+++ b/tests/foreman/api/test_repositories.py
@@ -65,10 +65,10 @@ def test_negative_disable_repository_with_cv(module_entitlement_manifest_org, ta
     with pytest.raises(HTTPError) as error:
         reposet.disable(data=data)
         # assert error.value.response.status_code == 500
-        assert (
-            'Repository cannot be deleted since it has already been '
-            'included in a published Content View' in error.value.response.text
-        )
+    assert (
+        'Repository cannot be deleted since it has already been '
+        'included in a published Content View' in error.value.response.text
+    )
 
 
 @pytest.mark.tier1
--- a/tests/foreman/api/test_rhcloud_inventory.py
+++ b/tests/foreman/api/test_rhcloud_inventory.py
@@ -108,7 +108,8 @@ def test_rhcloud_inventory_api_e2e(
     infrastructure_type = [
         host['system_profile']['infrastructure_type'] for host in json_data['hosts']
     ]
-    assert 'physical' and 'virtual' in infrastructure_type
+    assert 'physical' in infrastructure_type
+    assert 'virtual' in infrastructure_type
     # Verify installed packages are present in report.
     all_host_profiles = [host['system_profile'] for host in json_data['hosts']]
     for host_profiles in all_host_profiles:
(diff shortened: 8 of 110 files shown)


JacobCallahan:
There are a few ruff checks that will fail in this initial version until I can contact component owners to decide how we want to handle the recommendations.

sambible:
`335 changes`
Need to make a story on our board to review this bad boy

JacobCallahan:
@sambible please do!

tests/foreman/api/test_http_proxy.py
```diff
@@ -293,7 +293,8 @@ def test_positive_sync_proxy_with_certificate(request, target_sat, module_org, m
     # Create and fetch new cerfiticate
     target_sat.custom_cert_generate(proxy_host)
     cacert = target_sat.execute(f'cat {cacert_path}').stdout
-    assert 'BEGIN CERTIFICATE' and 'END CERTIFICATE' in cacert
+    assert 'BEGIN CERTIFICATE'
+    assert 'END CERTIFICATE' in cacert
```
jyejare, line 297: Same issue as above.

tests/foreman/api/test_rhcloud_inventory.py
```diff
@@ -108,7 +108,8 @@ def test_rhcloud_inventory_api_e2e(
     infrastructure_type = [
         host['system_profile']['infrastructure_type'] for host in json_data['hosts']
     ]
-    assert 'physical' and 'virtual' in infrastructure_type
+    assert 'physical'
+    assert 'virtual' in infrastructure_type
```
jyejare, line 112: Same issue here as above

jyejare (changes requested):
Round 1 Review !

pytest_fixtures/component/satellite_auth.py
```diff
@@ -343,8 +343,8 @@ def module_enroll_idm_and_configure_external_auth(module_target_sat):
     ipa_host.disenroll_idm()
 
 
-@pytest.mark.external_auth
-@pytest.fixture
+@pytest.mark.external_auth()
```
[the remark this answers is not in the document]
JacobCallahan: correct, each mark is a decorator. adding the parenthesis for them is more for forcing consistency. though we could likely force consistency without the parenthesis. this is up for debate in this PR.
https://docs.astral.sh/ruff/rules/pytest-incorrect-mark-parentheses-style/

tests/foreman/api/test_acs.py
```diff
@@ -23,9 +23,9 @@
 from robottelo.constants.repos import PULP_FIXTURE_ROOT, PULP_SUBPATHS_COMBINED
 
 
-@pytest.mark.e2e
-@pytest.mark.tier2
-@pytest.mark.upgrade
+@pytest.mark.e2e()
+@pytest.mark.tier2()
+@pytest.mark.upgrade()
```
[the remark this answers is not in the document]
JacobCallahan: yep. see above

tests/foreman/api/test_capsulecontent.py
```diff
@@ -1176,7 +1176,8 @@ def test_positive_sync_CV_to_multiple_LCEs(
         result = module_capsule_configured.nailgun_capsule.content_lifecycle_environments()
         # there can and will be LCEs from other tests and orgs, but len() >= 2
         assert len(result['results']) >= 2
-        assert lce1.id and lce2.id in [capsule_lce['id'] for capsule_lce in result['results']]
+        assert lce1.id
+        assert lce2.id in [capsule_lce['id'] for capsule_lce in result['results']]
```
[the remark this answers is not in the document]
JacobCallahan: This is formatted correctly or the original test author was mistaken about what they were testing. There are two checks going on here, so it is a separation of them to better clarify where the failure is.

1. check that lce1.id is a truthy value
2. check that lce2.id is in the results list

Also, be careful with your recommended change since it doesn't do what you expect. That first statement will only ever return the value of lce2.id. Consider the following:
```python
>>> (1 and 2)
2
>>> (1 or 2)
1
```

pytest_fixtures/component/satellite_auth.py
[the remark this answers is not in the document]
sambible: I'm definitely in favor of the parens for consistency. The only downside is it looks a little more cluttered visually, but it feels substantially more consistent to be specifically declaring that these are all functions, even if there aren't any arguments

tests/foreman/api/test_http_proxy.py
JacobCallahan: Same answer as above, with a slight modification. It now appears the test author is incorrect in their assumption of how this code works. Consider the following:
```python
>>> 1 and 2 in [2, 3, 4]
True
```

tests/foreman/api/test_bookmarks.py
```diff
@@ -367,8 +367,8 @@ def test_negative_update_empty_query(controller):
     assert bm.query != ''
 
 
-@pytest.mark.tier1
-@pytest.mark.parametrize('public', (True, False))
+@pytest.mark.tier1()
+@pytest.mark.parametrize('public', [True, False])
```
sambible, line 371: This is interesting, any particular reason for the change here

tests/foreman/api/test_rhcloud_inventory.py
JacobCallahan: same answer as the previous two :)
if we agree this is a test logic issue, I can fix them to be the intended behavior.

tests/foreman/api/test_classparameters.py
```diff
@@ -137,15 +137,15 @@ def test_negative_update_parameter_type(self, test_data, module_puppet):
             2. Error raised for invalid default value.
         """
         sc_param = module_puppet['sc_params'].pop()
+        sc_param.override = True
+        sc_param.parameter_type = test_data['sc_type']
+        sc_param.default_value = test_data['value']
```
sambible, line 142: I guess it's taking out the stuff from the raises block that's not a function call. idk how I feel about this stylistically

tests/foreman/api/test_media.py
```diff
@@ -38,10 +38,10 @@ class TestMedia:
     def class_media(self, module_org):
         return entities.Media(organization=[module_org]).create()
 
-    @pytest.mark.tier1
-    @pytest.mark.upgrade
+    @pytest.mark.tier1()
+    @pytest.mark.upgrade()
     @pytest.mark.parametrize(
-        'name, new_name',
+        ('name', 'new_name'),
```
sambible, line 44: Does putting this in a tuple have any affect on the functionality here? I don't think so, but just curious

tests/foreman/api/test_role.py
```diff
@@ -1238,6 +1238,7 @@ def test_positive_taxonomies_control_to_superadmin_without_org_admin(
         entities.User(id=user.id).delete()
         with pytest.raises(HTTPError):
             user_role.read()
+        with pytest.raises(HTTPError):
```
sambible, line 1241: This looks worse to me than just having 2 calls in the same raises block

sambible (changes requested):

tests/foreman/api/test_bookmarks.py
JacobCallahan: https://docs.astral.sh/ruff/rules/pytest-parametrize-values-wrong-type

tests/foreman/api/test_classparameters.py
JacobCallahan: the point of this change is to make sure that only the call that raises the error should be in the context manager. there are no functional reasons why these extra lines should be under that context.

tests/foreman/api/test_media.py
JacobCallahan: this looks to have been a bug in the original test since this isn't a valid parameterization. my guess is that since we're using nailgun in the test, `name` and `new_name` being `None` result in nailgun creating values on its own.

tests/foreman/api/test_repositories.py
```diff
@@ -64,14 +64,14 @@ def test_negative_disable_repository_with_cv(module_entitlement_manifest_org, ta
     data = {'basearch': 'x86_64', 'releasever': '7Server', 'product_id': rh_repo.product.id}
     with pytest.raises(HTTPError) as error:
         reposet.disable(data=data)
-        # assert error.value.response.status_code == 500
-        assert (
-            'Repository cannot be deleted since it has already been '
-            'included in a published Content View' in error.value.response.text
-        )
+    # assert error.value.response.status_code == 500
```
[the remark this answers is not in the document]
JacobCallahan: formatting is correct. this context manager doesn't clear the contents of the context variable on exit. this allows us to perform checks after we exit the context, and will be our enforced style moving forward.

tests/foreman/api/test_role.py
JacobCallahan: which is more functionally correct for the test, though?
shouldn't both raise HTTPErrors?
what if the `user_role.read()` doesn't raise an HTTPError but `user.read()` does?

tests/foreman/api/test_repositories.py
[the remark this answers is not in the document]
sambible: TIL! That's pretty cool.

tests/foreman/api/test_role.py
sambible: You are definitely right, this is more correct. I'm assuming the second statement here hasn't ever executed, actually, since the first statement will always raise an error ( it's been deleted ) and you'd leave the ocntext

Griffin-Sullivan (approved):
Changes to Platform owned tests / fixtures look good to me!

tests/foreman/api/test_capsulecontent.py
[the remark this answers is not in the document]
jyejare: Also its not 2 assertions, its one assertion to verify both items are in the list!

pytest_fixtures/component/satellite_auth.py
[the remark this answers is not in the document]
jyejare: Yeah I mean if its really accepting params thats fine but we can keep it optional if it isnt taking params, looks bit weird !

tests/foreman/api/test_http_proxy.py
jyejare: Yeah there has to be double in operation or set operations to find the items in other list.

tests/foreman/api/test_rhcloud_inventory.py
jyejare: Its a test logic issue for sure!

tests/foreman/api/test_http_proxy.py
JacobCallahan: for test failure clarity, i think having both as separate assertions would be better

pytest_fixtures/component/satellite_auth.py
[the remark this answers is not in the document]
jyejare: In the general world(Learning material etc), pytest marks without params are represented without parenthesis so good to follow that so that any new person contributes in robottelo should not have confusion !

I dont want to press to much on it though!

tests/foreman/api/test_http_proxy.py
jyejare: Yeah for now we can leave it as two separate operations and later Component owner can refactor if needed.

tests/foreman/longrun/test_provisioning_computeresource.py
```diff
@@ -144,8 +144,7 @@ def test_positive_provision_vmware_with_host_group(
             "start=1".format(VMWARE_CONSTANTS['cluster'], vmware.vmware_datacenter),
             'ip': None,
             'mac': None,
-            'interface': "compute_network={},"
-            "compute_type=VirtualVmxnet3".format(vmware.vmware_net_id),
+            'interface': f"compute_network={vmware.vmware_net_id}," "compute_type=VirtualVmxnet3",
```
adarshdubey-star, line 147: @shubhamsg199 @Gauravtalreja1 this looks good to me , need your second eye here.

tests/foreman/ui/test_discoveryrule.py
```diff
@@ -155,14 +155,14 @@ def test_negative_delete_rule_with_non_admin_user(
         location=[module_location],
     ).create()
     with Session(user=reader_user.login, password=reader_user.password) as session:
-        with pytest.raises(ValueError):
+        with pytest.raises(ValueError):  # noqa: PT011 - TODO Adarsh determine better exception
             session.discoveryrule.delete(dr.name)
         dr_val = session.discoveryrule.read_all()
         assert dr.name in [rule['Name'] for rule in dr_val]
```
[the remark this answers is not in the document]
adarshdubey-star: Sure I will have a look and address this in my PR's

tests/foreman/longrun/test_provisioning_computeresource.py
shubhamsg199: Yep, Looks good, anyways these long run tests have been migrated to `tests/foreman/cli/test_computeresource_vmware.py` with these changes

adarshdubey-star (approved):
Changes to the rocket team components look valid to me.

tests/foreman/api/test_capsulecontent.py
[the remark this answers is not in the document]
vsedmik: Hey, all of you are right! :D
Yes, the intention was obviously test that both LCEs are assigned to the Capsule and yes, I was mistaken and `lce1 and lce2` doesn't work as I expected.
So we should rather have something like
```
assert all(item in [capsule_lce['id'] for capsule_lce in result['results']] for item in [lce1.id, lce2.id])
```
Unless I am mistaken again or you have much nicer way to test that.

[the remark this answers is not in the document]
vsedmik: Or
```
assert {lce1.id, lce2.id}.issubset([capsule_lce['id'] for capsule_lce in result['results']])
```
might be more pythonic?

shubhamsg199 (approved):
Changes to Rocket team tests and fixtures look good to me.

tests/foreman/api/test_rhcloud_inventory.py
jameerpathan111: @JacobCallahan @ogajduse it's test logic, as @jyejare mentioned. both the `virtual` and `physical` should be present in `infrastructure_type`.

tests/upgrades/test_classparameter.py
```diff
@@ -130,9 +130,10 @@ def test_pre_puppet_class_parameter_data_and_type(
         self._validate_value(data, sc_param)
 
     @pytest.mark.post_upgrade(depend_on=test_pre_puppet_class_parameter_data_and_type)
+    @pytest.mark.usefixtures('_clean_scenario')
     @pytest.mark.parametrize('count', list(range(1, 10)))
     def test_post_puppet_class_parameter_data_and_type(
-        self, count, _clean_scenario, class_pre_upgrade_data, class_target_sat
+        self, count, class_pre_upgrade_data, class_target_sat
```
Gauravtalreja1, line 136: Just wondering, if this change to use usefixtures here made by ruff ?

tests/foreman/api/test_http_proxy.py
damoore044: Looking back on this, it seems unnecessary to check for both the beginning and end of the certificate, given this is an http_proxy test, it's assumed the generated cert would have beginning and end, no? Perhaps we could just make this a single assertion- 
`assert 'BEGIN CERTIFICATE' in cacert`

tests/upgrades/test_classparameter.py
JacobCallahan: this was a manual change based on one of the new rules ruff enforces.

lpramuk (approved):

jyejare (changes requested):
ACK pending minor comments !

tests/foreman/api/test_http_proxy.py
```diff
@@ -293,7 +293,7 @@ def test_positive_sync_proxy_with_certificate(request, target_sat, module_org, m
     # Create and fetch new cerfiticate
     target_sat.custom_cert_generate(proxy_host)
     cacert = target_sat.execute(f'cat {cacert_path}').stdout
-    assert 'BEGIN CERTIFICATE' and 'END CERTIFICATE' in cacert
+    assert 'END CERTIFICATE' in cacert
```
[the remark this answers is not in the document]
JacobCallahan: Correct, see https://github.com/SatelliteQE/robottelo/pull/12796#discussion_r1358208022

tests/foreman/ui/test_contenthost.py
```diff
@@ -1006,7 +1006,8 @@ def test_module_stream_actions_on_content_host(session, default_location, vm_mod
         )
         assert module_stream[0]['Name'] == FAKE_2_CUSTOM_PACKAGE_NAME
         assert module_stream[0]['Stream'] == stream_version
-        assert 'Enabled' and 'Installed' in module_stream[0]['Status']
+        assert 'Enabled'
```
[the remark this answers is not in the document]
JacobCallahan: good catch, fixed

jyejare (approved):

jyejare:
@JacobCallahan I was planning to merge it now ! And unfortunately I see conflicts holding it up ! Please resolve them :(

PR #12810, opened 2023-10-04
Hosts: move from RHEL 7 to RHEL 8 for glob. registration

pnovotny:
~~Use all supported RHEL versions for global registration tests,
i.e., RHEL 7,8,9, including FIPS.~~

Update:
Move from RHEL 7 to RHEL 8 hosts for non-crucial global registration tests.

Also fixed a forgotten assert in `test_global_registration_token_restriction`.

### DIFF
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -1313,7 +1313,7 @@ def test_global_registration_form_populate(
 def test_global_registration_with_capsule_host(
     session,
     capsule_configured,
-    rhel7_contenthost,
+    rhel8_contenthost,
     module_org,
     module_location,
     module_product,
@@ -1342,7 +1342,7 @@ def test_global_registration_with_capsule_host(
 
     :CaseAutomation: Automated
     """
-    client = rhel7_contenthost
+    client = rhel8_contenthost
     repo = target_sat.api.Repository(
         url=settings.repos.yum_1.url,
         content_type=REPO_TYPE['yum'],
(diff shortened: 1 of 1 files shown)


tests/foreman/ui/test_host.py
```diff
@@ -1313,7 +1313,7 @@ def test_global_registration_form_populate(
 def test_global_registration_with_capsule_host(
     session,
     capsule_configured,
-    rhel7_contenthost,
```
Gauravtalreja1, line 1316: We're covering all of the host registrations in tests/foreman/api/test_registration.py::test_host_registration_end_to_end
With UI there's no significant change in the registration process, so there's not much we would get if we cover all versions, and I think covering only rhel7 or rhel8 contenthost makes sense

pnovotny: Fair point, there is also UI e2e test for registration at tests/foreman/ui/test_registration.py::test_positive_global_registration_end_to_end
that runs all the supported RHELs.

Given that, instead of using more RHEL versions here, I would at least update `rhel7_contenthost` to `rhel8_contenthost` as RHEL 8 is IMO more common and recommended nowadays.

Gauravtalreja1: yes sure, it'll make sense

pnovotny: Done, moved to RHEL 8.

pnovotny:
trigger: test-robottelo
pytest: tests/foreman/ui/test_host.py -k 'test_global_registration_with_capsule_host or test_global_registration_with_gpg_repo_and_default_package or test_global_registration_token_restriction'

Gauravtalreja1 (approved):

PR #12813, opened 2023-10-04
Nailgun ssl cert verification

dosas:
This PR adds the possibility to overwrite the `verify` settings of nailgun's [ServerConfig ](https://github.com/SatelliteQE/nailgun/blob/master/nailgun/config.py#L219) 

According to the setttings given by [requests](https://requests.readthedocs.io/en/latest/user/advanced/#ssl-cert-verification)

Backwards compatibility is kept by defaulting to `False` if the new settings option is not available.

### DIFF
--- a/robottelo/config/__init__.py
+++ b/robottelo/config/__init__.py
@@ -99,6 +99,18 @@ def get_url():
     return urlunsplit((scheme, hostname, '', '', ''))
 
 
+def get_ssl_cert_verify():
+    """Return the SSL certificate or setting to verify HTTPS requests.
+
+    :return: Certificate path or boolean verify setting
+    :rtype: bool or str
+    """
+    try:
+        return settings.server.verify_ca
+    except AttributeError:
+        return False
+
+
 def user_nailgun_config(username=None, password=None):
     """Return a NailGun configuration file constructed from default values.
 
@@ -110,7 +122,7 @@ def user_nailgun_config(username=None, password=None):
 
     """
     creds = (username, password)
-    return ServerConfig(get_url(), creds, verify=False)
+    return ServerConfig(get_url(), creds, verify=get_ssl_cert_verify())
 
 
 def setting_is_set(option):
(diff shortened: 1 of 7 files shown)


omkarkhatavkar:
test-robottelo

omkarkhatavkar:
"trigger": "test-robottelo"

jyejare (approved):

robottelo/config/__init__.py
```diff
@@ -99,6 +99,18 @@ def get_url():
...
 
 
+def get_ssl_cert_verify():
+    """Return the SSL certificate or setting to verify HTTPS requests.
+
+    :return: Certificate path or boolean verify setting
+    :rtype: bool or str
+    """
+    try:
+        return settings.server.verify_ca
```
jyejare, line 109: You can validate settings from https://github.com/SatelliteQE/robottelo/blob/master/robottelo/config/validators.py to return the `False` value by default and hence need not to handle exception here explicitly!

jyejare: Also, you don't need this function in that case, just directly call the settings object wherever applicable

jyejare (commented):
NonBlocking comments!

jyejare:
trigger: test-robottelo

robottelo/config/__init__.py
dosas: Thanks, in that case I will adapt it

omkarkhatavkar (approved):
ACK

omkarkhatavkar:
@jyejare We need to add the same in the CI PR to make this work, otherwise regression would have impact

jyejare:
@omkarkhatavkar Not required, The validator should handle it and return False value automatically! We can merge it as is but good keep CI in sync !

jyejare:
@omkarkhatavkar Opened CI MR 1134 !

jyejare (approved):

PR #12825, opened 2023-10-05
[6.12.z] Remove stubbed FM case that is no longer needed

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/12691

I'm removing this stubbed case because the steps for updating satellite-maintain here are outdated. We now rely on satellite-maintain's ability to self-upgrade either by running a command like `satellite-maintain upgrade list-versions` which is being tested in https://github.com/SatelliteQE/robottelo/pull/12647 or by using `satellite-maintain self-upgrade`. Dev is planning to make some changes to these commands as well which will get there own automation when it's ready. For now we can just remove this outdated stubbed case.

### DIFF
--- a/tests/foreman/maintain/test_packages.py
+++ b/tests/foreman/maintain/test_packages.py
@@ -270,22 +270,3 @@ def test_positive_fm_packages_update(request, sat_maintain):
     def _finalize():
         assert sat_maintain.execute('dnf remove -y walrus').status == 0
         sat_maintain.execute('rm -rf /etc/yum.repos.d/custom_repo.repo')
-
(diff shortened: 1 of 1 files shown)


jyejare:
@Gauravtalreja1 Does the new change made for all living zStreams as well like 6.12, 6.13, 6.14 as well or just developing version ?

Griffin-Sullivan:
Maybe @ehelms would have advise here.

ehelms:
All versions.

jyejare (approved):

PR #12838, opened 2023-10-06
Fixing loadbalancer e2e test

Griffin-Sullivan:
Modifying this test to focus more on the content host being able to get content even if the capsule it's registered to is down.

### DIFF
--- a/tests/foreman/destructive/test_capsule_loadbalancer.py
+++ b/tests/foreman/destructive/test_capsule_loadbalancer.py
@@ -17,6 +17,7 @@
 :Upstream: No
 """
 import pytest
+from wrapanapi import VmState
 
 from robottelo.config import settings
 from robottelo.constants import CLIENT_PORT, DataFile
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_capsule_loadbalancer.py::test_loadbalancer_install_package

tests/foreman/destructive/test_capsule_loadbalancer.py
```diff
@@ -175,30 +176,27 @@ def loadbalancer_setup(
...
         registration.
 
     :CaseLevel: Integration
     """
+    # Register content host
     url = f'https://{loadbalancer_setup["setup_haproxy"]["haproxy"].hostname}'
-    server_url = f'{url}:8443/rhsm'
-    base_url = f'{url}/pulp/content'
 
     result = rhel7_contenthost.download_install_rpm(
```
synkd, line 201: Non-blocking: since you're refactoring this test, consider replacing installing the `katello-ca-consumer` package and registering with sub-man with registering via global registration.

synkd (commented):
I think the functionality that you added to this test makes a lot of sense; we should definitely be testing whether the load balancer is correctly proxying yum transactions to a second Capsule if the first Capsule is unavailable. However, I'm curious as to why you removed the existing functionality from the test instead of writing a new test to capture this scenario. 

Before your changes, if I'm reading correctly, the test was effectively checking that the HAProxy server's round robin algorithm functions as expected with registration requests. With your changes, we're no longer testing that functionality, since the other test in this module specifies `target` parameters at registration time that should override the round robin algorithm.

Could you please explain your thinking a bit more regarding removing functionality vs. adding a new test?

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_capsule_loadbalancer.py::test_loadbalancer_install_package

ehelms:
> Before your changes, if I'm reading correctly, the test was effectively checking that the HAProxy server's round robin algorithm functions as expected with registration requests. With your changes, we're no longer testing that functionality, since the other test in this module specifies `target` parameters at registration time that should override the round robin algorithm.
> 
> Could you please explain your thinking a bit more regarding removing functionality vs. adding a new test?

This was at my recommendation. I did not feel that our tests job should be to test whether round-robin works for HAProxy and that the steps involved here are not of a real use case nature. And this test was proving flaky.

First and foremost, users care that if a host is registered that it can receive it's content no matter what is happening behind the load-balancer. Secondly, that if a host registers it will register no matter what is happening behind the load-balancer. So we could keep the registration test -- but I would opt for the same strategy of taking down a Capsule and ensuring that registration continues to work.

Griffin-Sullivan:
> So we could keep the registration test -- but I would opt for the same strategy of taking down a Capsule and ensuring that registration continues to work.

We do have https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/destructive/test_capsule_loadbalancer.py#L270 so I can edit that after this.

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_capsule_loadbalancer.py::test_loadbalancer_install_package

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_capsule_loadbalancer.py::test_loadbalancer_install_package

synkd (approved):
ACK pending PRT results.

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_capsule_loadbalancer.py::test_loadbalancer_install_package

sambible (approved):

PR #12840, opened 2023-10-06
Add a mechanism to swap nailgun versions on demand

JacobCallahan:
Satellite._swap_nailgun("x.y.z") can be used to change out the operating version of nailgun.

This actually changes out the installed version of nailgun on the fly, so this should **only** be used in upgrades.

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -1752,6 +1752,18 @@ class Satellite(Capsule, SatelliteMixins):
         self._api = type('api', (), {'_configured': False})
         self._cli = type('cli', (), {'_configured': False})
 
+    def _swap_nailgun(self, new_version):
+        """Install a different version of nailgun from GitHub and invalidate the module cache."""
+        import sys
+
+        from pip._internal import main as pip_main
+
+        pip_main(['uninstall', '-y', 'nailgun'])
+        pip_main(['install', f'https://github.com/SatelliteQE/nailgun/archive/{new_version}.zip'])
+        self._api = type('api', (), {'_configured': False})
+        to_clear = [k for k in sys.modules.keys() if 'nailgun' in k]
+        [sys.modules.pop(k) for k in to_clear]
+
     @property
     def api(self):
         """Import all nailgun entities and wrap them under self.api"""
@@ -1759,7 +1771,7 @@ class Satellite(Capsule, SatelliteMixins):
             self._api = type('api', (), {'_configured': False})
         if self._api._configured:
             return self._api
-
+        from nailgun import entities as _entities  # use a private import
         from nailgun.config import ServerConfig
         from nailgun.entity_mixins import Entity
 
@@ -1779,7 +1791,7 @@ class Satellite(Capsule, SatelliteMixins):
             verify=False,
         )
         # add each nailgun entity to self.api, injecting our server config
-        for name, obj in entities.__dict__.items():
+        for name, obj in _entities.__dict__.items():
             try:
                 if Entity in obj.mro():
                     #  create a copy of the class and inject our server config into the __init__


robottelo/hosts.py
```diff
@@ -1752,14 +1752,26 @@ def __init__(self, hostname=None, **kwargs):
...
+        """Install a different version of nailgun from GitHub and invalidate the module cache."""
+        import sys
+
+        from pip._internal import main as pip_main
+
+        pip_main(['uninstall', '-y', 'nailgun'])
+        pip_main(['install', f'https://github.com/SatelliteQE/nailgun/archive/{new_version}.zip'])
+        self._api = type('api', (), {'_configured': False})
+        to_clear = [k for k in sys.modules.keys() if 'nailgun' in k]
+        [sys.modules.pop(k) for k in to_clear]
```
jyejare, line 1765: Does this clearance should be done after uninstalling the previous Nailgun version and before installing new nailgun version ?

robottelo/hosts.py
```diff
@@ -1752,14 +1752,26 @@ def __init__(self, hostname=None, **kwargs):
         self._api = type('api', (), {'_configured': False})
         self._cli = type('cli', (), {'_configured': False})
 
+    def _swap_nailgun(self, new_version):
```
jyejare, line 1755: Would you like this to be a private function or public?

I assume this function would be called with the `upgrade_process` fixture(or whatever we will create for upgrades). Agree that `upgrade_process` could be part of the target
-sat but not necessary we need to tie `_swap_nailgun` method with satellite object it could be independent fixture or utils function, since its being operated on pytest host machine and not on satellite.

robottelo/hosts.py
```diff
@@ -1752,14 +1752,26 @@ def __init__(self, hostname=None, **kwargs):
...
+
     @property
     def api(self):
         """Import all nailgun entities and wrap them under self.api"""
         if not self._api:
             self._api = type('api', (), {'_configured': False})
         if self._api._configured:
             return self._api
-
+        from nailgun import entities as _entities  # use a private import
```
jyejare, line 1774: Not sure how this would help with nailgun switching, could you help us to understand ?

jyejare (commented):
Some comments and Questions !

robottelo/hosts.py
jyejare: This implementation would likely cleanup the bits of newly installed nailgun as well hence above comment, correct me if I am wrong .

JacobCallahan: the order here is less important than you might think. the current modules are live in memory, so this only really helps for new imports (like what we do when recreating the `Satellite.api` attributes.

robottelo/hosts.py
JacobCallahan: `entities` are already imported on line 23 of this file, so we need a new alias to use the new import.

robottelo/hosts.py
jyejare: Interesting how popping out the nailgun specific modules from sys.modules would still make nailgun module available.

Do you mean the newly installed nailgun isnt the part of `sys.modules` and only unistalled is ?

robottelo/hosts.py
jyejare: But wouldnt simply removing the old nailgun and installing a new nailgun point the entities in a new nailgun module ? Do we really need to differ here ?

jyejare (commented):
ACK pending questions/doubts!

robottelo/hosts.py
JacobCallahan: no because `entities` is a live object already loaded into memory in this namespace so it will persist. You can test this out locally.
```python
import sys
from nailgun import entities
print("===== Before =====")
print(id(entities), id(entities.ActivationKey))
to_clear = [k for k in sys.modules.keys() if 'nailgun' in k]
[sys.modules.pop(k) for k in to_clear]
print("===== After =====")
print(id(entities), id(entities.ActivationKey))
```

JacobCallahan:
@jyejare re: 

> "Do you mean the newly installed nailgun isnt the part of sys.modules and only unistalled is ?"

Until a module is loaded via an import statement, it won't be added to `sys.modules`

jyejare (approved):

ColeHiggins2 (approved):

PR #12844, opened 2023-10-09
Change read to read_legacy_ui

sambible:
Changes a few things here to unblock this test:

- sets `install_katello_agent` to false
- selects the default location, where hosts are being registered
- uses the read_legacy_ui, instead of read, since the test is referencing the old UI


sambible:
trigger: test-robottelo
pytest: tests/foreman/ui/test_activationkey.py -k 'test_positive_end_to_end_register'

Griffin-Sullivan (approved):

PR #12845, opened 2023-10-09
Fix packages status

Griffin-Sullivan:
Previous package install commands are setting the command_end to the name of the package. This is causing a failure in one of our install tests. This makes sure that the status command does not pass along any package names even after an install.

### DIFF
--- a/robottelo/cli/sm_packages.py
+++ b/robottelo/cli/sm_packages.py
@@ -51,6 +51,7 @@ class Packages(Base):
     def status(cls, options=None):
         """Build satellite-maintain packages status"""
         cls.command_sub = 'status'
+        cls.command_end = None
         options = options or {}
         return cls.sm_execute(cls._construct_command(options))
 


Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_packages.py::test_positive_fm_packages_install

robottelo/cli/sm_packages.py
```diff
@@ -51,6 +51,7 @@ def is_locked(cls, options=None):
     def status(cls, options=None):
         """Build satellite-maintain packages status"""
         cls.command_sub = 'status'
+        cls.command_end = None
```
Gauravtalreja1, line 54: `command_end = None` is defined already in Base class and we're just inheriting the Base class herein Packages class, not sure what diff it makes?

Griffin-Sullivan: The test above fails because other commands in the sm_packages.py file like `install` and `update` set `cls.command_end = packages`. So when the command is constructed it passes the old command_end with the package name when it should not pass anything. This just makes sure we don't pass anything to the end of the status command.

synkd (approved):

ehelms (approved):

ColeHiggins2 (approved):

robottelo/cli/sm_packages.py
Gauravtalreja1: Okay, I see no issues with hardcoding `cls.command_end = None` as it's working for you.

Gauravtalreja1 (approved):

PR #12851, opened 2023-10-09
Update to errata-stream fix, use up to date register method for rhel_contenthost

damoore044:
Previous pr was merged just a bit before I could get this in. The old method works and passes, 
but should use `.register()` going forward with the removal of katello-agent.

### DIFF
--- a/tests/foreman/api/test_errata.py
+++ b/tests/foreman/api/test_errata.py
@@ -163,7 +163,7 @@ def test_positive_install_in_hc(module_org, activation_key, custom_repo, target_
 @pytest.mark.no_containers
 @pytest.mark.e2e
 def test_positive_install_multiple_in_host(
-    function_org, rhel_contenthost, target_sat, function_lce
+    target_sat, rhel_contenthost, function_org, function_lce
 ):
     """For a host with multiple applicable errata install one and ensure
     the rest of errata is still available
@@ -199,9 +199,14 @@ def test_positive_install_multiple_in_host(
             'activationkey-id': ak.id,
         }
     )
-    # Install katello-ca, register content-host, enable all repos:
-    rhel_contenthost.install_katello_ca(target_sat)
-    rhel_contenthost.register_contenthost(function_org.name, ak.name, function_lce.name)
+    rhel_contenthost.register(
+        org=function_org,
+        activation_keys=ak.name,
+        lifecycle_environment=function_lce,
+        target=target_sat,
+        force=True,
+        loc=None,
+    )
     assert rhel_contenthost.subscribed
     # Installing outdated custom packages:
     for package in constants.FAKE_9_YUM_OUTDATED_PACKAGES:
@@ -209,7 +214,6 @@ def test_positive_install_multiple_in_host(
         assert rhel_contenthost.run(f'yum install -y {package}').status == 0
         assert rhel_contenthost.run(f'rpm -q {package}').status == 0
     rhel_contenthost.add_rex_key(satellite=target_sat)
-    rhel_contenthost.run(r'subscription-manager repos --enable \*')
     # Each errata will be installed sequentially,
     # after each install, applicable-errata-count should drop by one.
     for errata in constants.FAKE_9_YUM_SECURITY_ERRATUM:
@@ -229,7 +233,7 @@ def test_positive_install_multiple_in_host(
             search_rate=20,
             max_tries=15,
         )
-        sleep(10)
+        sleep(20)
         assert rhel_contenthost.applicable_errata_count == pre_errata_count - 1
 
 
--- a/tests/foreman/api/test_errata.py
+++ b/tests/foreman/api/test_errata.py
@@ -225,8 +225,7 @@ def test_positive_install_multiple_in_host(target_sat, rhel_contenthost, module_
             search_rate=20,
             max_tries=15,
         )
-        rhel_contenthost.run('subscription-manager repos')
-        sleep(10)
+        sleep(20)
         assert rhel_contenthost.applicable_errata_count == pre_errata_count - 1
 
 
--- a/tests/foreman/api/test_errata.py
+++ b/tests/foreman/api/test_errata.py
@@ -17,7 +17,6 @@
 :Upstream: No
 """
 # For ease of use hc refers to host-collection throughout this document
-from copy import copy
 from time import sleep
 
 from nailgun import entities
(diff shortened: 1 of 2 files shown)


damoore044:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py -k test_positive_install_multiple_in_host

damoore044:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py -k test_positive_install_multiple_in_host

tests/foreman/api/test_errata.py
```diff
@@ -199,17 +199,21 @@ def test_positive_install_multiple_in_host(
...
     )
-    # Install katello-ca, register content-host, enable all repos:
-    rhel_contenthost.install_katello_ca(target_sat)
-    rhel_contenthost.register_contenthost(function_org.name, ak.name, function_lce.name)
+    rhel_contenthost.register(
+        org=function_org,
+        activation_keys=ak.name,
+        lifecycle_environment=function_lce,
+        target=target_sat,
+        force=True,
```
vijaysawant, line 207: I guess, without `force=True` contenthost registration should work.
Can you please try by removing this parameter?

vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py -k test_positive_install_multiple_in_host

vsedmik (commented):
Looks like `sleep(20)` is still not enough in some cases, rerun PRT

vsedmik (approved):
PRT rerun all passed this time :thinking: 
Looks good, just Vijay's comment is pending

tests/foreman/api/test_errata.py
```diff
@@ -199,17 +199,21 @@ def test_positive_install_multiple_in_host(
             'activationkey-id': ak.id,
         }
     )
-    # Install katello-ca, register content-host, enable all repos:
-    rhel_contenthost.install_katello_ca(target_sat)
-    rhel_contenthost.register_contenthost(function_org.name, ak.name, function_lce.name)
+    rhel_contenthost.register(
+        org=function_org,
+        activation_keys=ak.name,
+        lifecycle_environment=function_lce,
```
Gauravtalreja1, line 205: when we've AK for registration then there's no need to LCE, and here your AK is created from function_lce already, so it can be removed too?
```suggestion
        activation_keys=ak.name,
```

tests/foreman/api/test_errata.py
```diff
@@ -199,17 +199,21 @@ def test_positive_install_multiple_in_host(
...
+        target=target_sat,
+        force=True,
+        loc=None,
```
Gauravtalreja1, line 208: just a nitpick, can you keep both org and loc args together

tests/foreman/api/test_errata.py
```diff
@@ -199,17 +199,21 @@ def test_positive_install_multiple_in_host(
...
+    )
     assert rhel_contenthost.subscribed
     # Installing outdated custom packages:
     for package in constants.FAKE_9_YUM_OUTDATED_PACKAGES:
         rhel_contenthost.run(f'yum remove -y {str(package.split("-", 1)[0])}')
         assert rhel_contenthost.run(f'yum install -y {package}').status == 0
         assert rhel_contenthost.run(f'rpm -q {package}').status == 0
     rhel_contenthost.add_rex_key(satellite=target_sat)
```
Gauravtalreja1, line 216: when using Global registration REX is setup as default, which includes copying key as well, so this step will no longer required
```suggestion
```

tests/foreman/api/test_errata.py
```diff
@@ -199,17 +199,21 @@ def test_positive_install_multiple_in_host(
...
-    rhel_contenthost.run(r'subscription-manager repos --enable \*')
```
Gauravtalreja1, line 212: Custom repos will be disabled by default on host, so are you sure this step is not required?

tests/foreman/api/test_errata.py
```diff
@@ -229,7 +233,7 @@ def test_positive_install_multiple_in_host(
             search_rate=20,
             max_tries=15,
         )
-        sleep(10)
+        sleep(20)
```
Gauravtalreja1, line 236: why sleep time?

tests/foreman/api/test_errata.py
damoore044: @vsedmik any thoughts? I think Jeremy recommended against this when using `.register()` ?

tests/foreman/api/test_errata.py
damoore044: failed prt see previous pr, needs this delay.
https://github.com/SatelliteQE/robottelo/pull/12706

tests/foreman/api/test_errata.py
damoore044: syntax issue, arguments set to None have to be after any explicit arguments.

tests/foreman/api/test_errata.py
damoore044: There is an asynchronous task to wait for after errata install, to update the list of applicable errata

tests/foreman/api/test_errata.py
damoore044: I will try without lce, the method `setup_org_for_a_custom_repo` still uses an lce passed or will create a new one

tests/foreman/api/test_errata.py
damoore044: Also if custom repos are disabled as the test exists now, how could it pass locally and prt?

vsedmik: I think because of [this](https://github.com/SatelliteQE/robottelo/blob/757f2a696ccb17ebcc2a98e7e5b2f2e925334262/robottelo/host_helpers/cli_factory.py#L676), which is executed by `target_sat.cli_factory.setup_org_for_a_custom_repo(...)`

tests/foreman/api/test_errata.py
Gauravtalreja1: yeah, so we argument for org can be kept at last

tests/foreman/api/test_errata.py
```diff
@@ -184,10 +184,7 @@ def test_positive_install_multiple_in_host(
 
     :CaseLevel: System
     """
-    ak = target_sat.api.ActivationKey(
-        organization=function_org,
-        environment=function_lce,
-    ).create()
+    ak = target_sat.api.ActivationKey(organization=function_org).create()
```
Gauravtalreja1, line 187: AK will require either function_lce or cv for contents, here only register doesn't requires lce when there's a AK

tests/foreman/api/test_errata.py
Gauravtalreja1: Oh my bad, I missed we're handling this in that helper, thanks !

tests/foreman/api/test_errata.py
damoore044: I received an error at registration without passing lce-
`HTTP error (500 - Internal Server Error): At least one activation key must have a lifecycle environment and content view assigned to it\n') `

Gauravtalreja1: yes, that must be due to this change, either lce or cv should be assigned to AK

damoore044:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py -k test_positive_install_multiple_in_host

damoore044:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py -k test_positive_install_multiple_in_host

damoore044:
trigger: test-robottelo
pytest: tests/foreman/api/test_errata.py -k test_positive_install_multiple_in_host -cli-log-level=DEBUG

Gauravtalreja1 (approved):

PR #12870, opened 2023-10-10
CLI return code error

vijaysawant:
tests are failing for CLI return code error, added no_containers marker to check behaviour


vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/ui/test_contenthost.py -k test_install_modular_errata

PR #12884, opened 2023-10-11
Add gitleaks to pre-commit hooks

JacobCallahan:
This will prevent secrets from being committed to the repository. Handling it in pre-commit because a check in actions is too late.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,29 +1,33 @@
 # configuration for pre-commit git hooks
 
 repos:
-- repo: https://github.com/pre-commit/pre-commit-hooks
-  rev: v4.4.0
-  hooks:
-  - id: trailing-whitespace
-    exclude: tests/foreman/data/
-  - id: check-yaml
-  - id: debug-statements
-- repo: https://github.com/psf/black
-  rev: 22.10.0
-  hooks:
-  - id: black
-- repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.0.277
-  hooks:
-    - id: ruff
-      args: [--fix, --exit-non-zero-on-fix]
-- repo: local
-  hooks:
-    - id: fix-uuids
-      name: Robottelo Custom Fix UUIDs script
-      description: This hook runs the scripts/fix_uuids.sh script
-      language: script
-      entry: scripts/fix_uuids.sh
-      verbose: true
-      types: [text]
-      require_serial: true
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan (approved):
Seems to be working. I added a fake private key file into the installer dir and it caught it. It also passes normally without.

```
pre-commit run --all-files                                                                                                                                 
trim trailing whitespace.................................................Passed
check yaml...............................................................Passed
debug statements (python)................................................Passed
black....................................................................Passed
ruff.....................................................................Passed
Robottelo Custom Fix UUIDs script........................................Passed
- hook id: fix-uuids
- duration: 0.09s

Running in uuid-fix mode...
No empty :id: was found
No duplicated :id: was found
No :id: missing spaces was found

Detect hardcoded secrets.................................................Failed
- hook id: gitleaks
- exit code: 1

○
    │╲
    │ ○
    ○ ░
    ░    gitleaks

Finding:     REDACTED
Secret:      REDACTED
RuleID:      private-key
Entropy:     6.002793
File:        tests/foreman/installer/key.pem
Line:        1
Fingerprint: tests/foreman/installer/key.pem:private-key:1

11:19AM INF 1 commits scanned.
11:19AM INF scan completed in 65ms
11:19AM WRN leaks found: 1
```

ogajduse (commented):
LGTM, sadly gitleaks does not catch internal URLs that we leak sometimes. It would be great to have a tool that could detect it. Perhaps a [custom gitleaks rule](https://github.com/gitleaks/gitleaks/#configuration) would do the trick.

```python
# the following is something that gitleaks do not detect using its default config
internal_url = "https://secret.server.internal.redhat.com/apisecret.json"
logger.info("Our internal URL is %s" % internal_url
```

.pre-commit-config.yaml
```diff
@@ -27,3 +27,7 @@ repos:
       verbose: true
       types: [text]
       require_serial: true
+- repo: https://github.com/gitleaks/gitleaks
+  rev: v8.16.1
```
Gauravtalreja1, line 31: Can we update it to latest version?
```suggestion
  rev: v8.18.0
```

Gauravtalreja1 (commented):
Great addition @JacobCallahan, I think we should enforce pre-commit now while committing the changes like satlab-tower repo, so we could get this hook working before committing anything locally, else if we leak anything it'll just be highlighted on the failed GHA

.pre-commit-config.yaml
JacobCallahan: updated, good catch!

ColeHiggins2 (approved):
Looks good to me! cant wait to feel safer :)

Gauravtalreja1 (approved):

PR #12896, opened 2023-10-12
Testing stream fixes for errata:cli

damoore044:


damoore044:
trigger: test-robottelo
pytest: tests/foreman/cli/test_errata.py -k test_positive_list_affected_chosts_by_erratum_restrict_flag -cli-log-level=DEBUG

damoore044:
trigger: test-robottelo
pytest: tests/foreman/cli/test_errata.py -k test_positive_list_affected_chosts_by_erratum_restrict_flag

tests/foreman/cli/test_errata.py
```diff
@@ -699,39 +729,40 @@ def test_positive_list_affected_chosts_by_erratum_restrict_flag(
 
     :CaseAutomation: Automated
     """
-
     # Uninstall package so that only the first errata applies.
     for host in errata_hosts:
+        host.execute(r'subscription-manager repos --enable \*')
+        host.execute(r'yum-config-manager --enable \*')
```
vsedmik, line 735: This shouldn't be necessary I guess as you already enable them in `errata_hosts` fixture (which is the good way I would say).

tests/foreman/cli/test_errata.py
```diff
@@ -699,39 +729,40 @@ def test_positive_list_affected_chosts_by_erratum_restrict_flag(
...
         'errata-restrict-installable': 0,
-        'content-view-id': module_cv.id,
         'lifecycle-environment-id': module_lce.id,
         'organization-id': module_entitlement_manifest_org.id,
         'per-page': PER_PAGE_LARGE,
     }
     errata_ids = get_errata_ids(param)
     assert set(REPO_WITH_ERRATA['errata_ids']).issubset(
-        errata_ids
+        set(errata_ids)
```
vsedmik, line 765: Not necessary to cast, `issubset` takes any iterable. But also no harm to have it explicitly. Just saying.

vsedmik (changes requested):
Looks good, just two questions and one non-blocking comment.

tests/foreman/cli/test_errata.py
```diff
@@ -349,27 +364,37 @@ def cv_publish_promote(cv, org, lce):
...
+    sat.cli.ContentViewFilter.delete(
         {
             'content-view-id': cv.id,
             'id': filter_id,
             'organization-id': org.id,
         }
     )
-    cv_publish_promote(cv, org, lce)
+    cv = cv.read()
+    cv_publish_promote(sat, cv, org, lce, force=True)
```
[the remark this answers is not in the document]
damoore044: Locally in test teardown I received errors for promoting to next env out of sequence, use force to bypass. 
Only on `cv_filter_cleanup` I have it forced, could it be doing something out of order?

tests/foreman/cli/test_errata.py
damoore044: I did see an odd error here related to subset of the list where there were duplicate list entries. I wasn't too sure about it but can test more without `set(errata_ids)`

tests/foreman/cli/test_errata.py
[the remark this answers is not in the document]
damoore044: Looks like it is passing, i think the assertion fail here was related to the `module_ak_cv_lce` fixture that I'm no longer using.

tests/foreman/cli/test_errata.py
damoore044: Looks like it's passing without casting, i think the assertion fail here was related to the `module_ak_cv_lce` fixture that I'm no longer using.

damoore044:
trigger: test-robottelo
pytest: tests/foreman/cli/test_errata.py -k test_positive_list_affected_chosts_by_erratum_restrict_flag

tests/foreman/cli/test_errata.py
vsedmik: Might be related to Python version, in 3.10.11 I can see
```
set([2,3]).issubset([1,2,3,4])
Out[3]: True
set([2,3]).issubset([1,3,4])
Out[4]: False
```
Okay so keep it.

damoore044:
Build [4398]: `In progress`

tests/foreman/cli/test_errata.py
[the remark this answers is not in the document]
vsedmik: Oh, maybe I know. Is the function called twice with the same CV in some test (or its teardown)?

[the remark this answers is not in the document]
damoore044: Yes, we call it once in this test at [L-818](https://github.com/SatelliteQE/robottelo/blob/4e38a466e9bd08ab90e368399e23e323bc0216e6/tests/foreman/cli/test_errata.py#L818)
`cv_publish_promote(target_sat, module_cv, module_entitlement_manifest_org, module_lce)`

Then we call once more during teardown (`def cleanup(): cv_filter_cleanup()` at [L-800](https://github.com/SatelliteQE/robottelo/blob/4e38a466e9bd08ab90e368399e23e323bc0216e6/tests/foreman/cli/test_errata.py#L800)) , 
to publish&promote a version without the `cv_filter` once it has been deleted.

`cv_filter_cleanup` calls the same `cv_publish_promote` that the test did prior.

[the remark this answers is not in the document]
damoore044: for cleanup instead of publishing/promoting the new version without `cv_filter`,
could instead just call `target_sat.cli.ContentView.remove_version` for the ver. created in the test ?

damoore044:
trigger: test-robottelo
pytest: tests/foreman/cli/test_errata.py -k test_positive_list_affected_chosts_by_erratum_restrict_flag

tests/foreman/cli/test_errata.py
damoore044: Okay I'll put this back in to be safe

vsedmik (approved):

tests/foreman/cli/test_errata.py
```diff
@@ -225,6 +239,7 @@ def errata_hosts(register_hosts):
             result = host.execute(f'yum install -y {old_package}')
             if result.status != 0:
                 pytest.fail(f'Failed to install {old_package}: {result.stdout} {result.stderr}')
+        host.execute('subscription-manager repos')
```
Griffin-Sullivan, line 242: What's the point of running this line? If it's for debugging / logging I would say use `logger.info()`

Griffin-Sullivan (approved):
One comment otherwise looks good to me!

tests/foreman/cli/test_errata.py
damoore044: It shouldn't be necessary here since it is correctly called earlier before removing and applying errata- [L735](https://github.com/SatelliteQE/robottelo/blob/3f98a1578cb6a0ec71e9f772653a7b7f29bae4c0/tests/foreman/cli/test_errata.py#L735). 
My understanding is this is called after host registration to update subscription-manager package status and errata applicability.

damoore044:
trigger: test-robottelo
pytest: tests/foreman/cli/test_errata.py -k test_positive_list_affected_chosts_by_erratum_restrict_flag

tests/foreman/cli/test_errata.py
Griffin-Sullivan: Ok makes sense thanks!

PR #12908, opened 2023-10-13
Add UEFI HTTP Boot provisioning test

shubhamsg199:
Adding UEFI HTTP Boot provisioning test. The test will be skipped till the BZ#2242925 is fixed

### DIFF
--- a/pytest_fixtures/component/provision_pxe.py
+++ b/pytest_fixtures/component/provision_pxe.py
@@ -313,6 +313,7 @@ def pxe_loader(request):
         'bios': {'vm_firmware': 'bios', 'pxe_loader': 'PXELinux BIOS'},
         'uefi': {'vm_firmware': 'uefi', 'pxe_loader': 'Grub2 UEFI'},
         'ipxe': {'vm_firmware': 'bios', 'pxe_loader': 'iPXE Embedded'},
+        'http_uefi': {'vm_firmware': 'uefi', 'pxe_loader': 'Grub2 UEFI HTTP'},
     }
     return Box(PXE_LOADER_MAP[getattr(request, 'param', 'bios')])
 
--- a/tests/foreman/api/test_provisioning.py
+++ b/tests/foreman/api/test_provisioning.py
@@ -16,12 +16,51 @@
 
 :Upstream: No
 """
+import re
+
 from fauxfactory import gen_string
 import pytest
-from wait_for import wait_for
+from wait_for import TimedOutError, wait_for
 
 from robottelo.config import settings
+from robottelo.logging import logger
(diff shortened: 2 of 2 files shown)


jyejare (approved):
ACK, looks good!

Could we please run PRT if possible for provisioning!

Gauravtalreja1 (approved):
Ack, great job @shubhamsg199 

@jyejare PRT will skip this tests as he's using `pytest.mark.skip_if_open("BZ:2242925")`

rplevka (changes requested):
I see no assertions that would test that the host is indeed being provisioned by HTTPBoot.
Satellite used to deploy the pxelinux and grub config files along the others no matter what os bootloader you selected. So I am wondering whether the host is indeed using the http boot here and not just fallbacking to any other method.
This test does not do any kind of introspection into the boot and installer procedures at all and simply blindly waits till the host build status is being flipped in satellite.
I'd personally  expect more assertions especially in this area as there are several points where the provisioning might go wrong.
Having a failed test of this kind, not knowing what happened sounds unoptimal to me.

Thoughts?

shubhamsg199:
> I see no assertions that would test that the host is indeed being provisioned by `HTTPBoot`. Satellite used to deploy the `pxelinux` and `grub` config files along the others no matter what kind of OS bootloader you selected. So I am wondering whether the host is indeed using the `HTTPBoot` here and not just falling back to any other method. This test does not do any kind of introspection into the boot and installer procedures at all and simply blindly waits till the host build status is being flipped in satellite. I'd personally expect more assertions especially in this area as there are several points where the provisioning might go wrong (invalid provisoning token being passed, proxy not reachable, rendered template being wrong,...). Having a failed test of this kind, not knowing what happened sounds non-optimal to me.
> 
> I'd suggest to add some checkpoints, like listening to incoming HTTP requests to the HTTP proxy to ensure, that a correct request indeed arrived and that the proxy correctly responded to it. We could as well issue the call to proxy manually and see if we get the properly rendered template with the given provisioning token (at least just log it or save it as an artifact for the eventual TFA).
> 
> Thoughts?

If we select Grub2 UEFI HTTP as pxe loader, the booting always starts with PXE over ipv4 then ipv6 and then over HTTP ipv4 and ipv6. However, PXE over ipv4 tries to download the `grubx64.efi` file but it fails, so I guess the Bootloader here matters.
I'll try to implement some functions maybe to read logs to verify the HTTP requests(similar to what. we have in discovery tests)

rplevka (approved):
thanks for implementing it.
there are few more log patterns you might test down the road, e.g. request for the ks file, etc. but i am not going to block the PR on this.

thank you

PR #12909, opened 2023-10-13
Fix vault makescripts with capture output

jyejare:
Missing `Capture Output` made `make vault scripts` to break as it was trying to read and print STDOUT which was never returned due to `capture_output` was not set.

Fix:
- Set `capture_output` to True in subprocess running vault command.
- Alter logging of some events to appropriate level.

This also addresses the REGEX for Vault_enablement key in `.env` when its the first line in `.env` file.

### DIFF
--- a/robottelo/utils/vault.py
+++ b/robottelo/utils/vault.py
@@ -25,7 +25,7 @@ class Vault:
     def setup(self):
         if self.env_path.exists():
             self.envdata = self.env_path.read_text()
-            is_enabled = re.findall('\nVAULT_ENABLED_FOR_DYNACONF=(.*)', self.envdata)
+            is_enabled = re.findall('^(?:.*\n)*VAULT_ENABLED_FOR_DYNACONF=(.*)', self.envdata)
             if is_enabled:
                 self.vault_enabled = is_enabled[0]
             self.export_vault_addr()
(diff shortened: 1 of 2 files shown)


jyejare:
trigger: test-robottelo
pytest: tests/foreman/api/test_product.py -k test_positive_create_with_label

robottelo/utils/vault.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, env_file='.env'):
     def setup(self):
         if self.env_path.exists():
             self.envdata = self.env_path.read_text()
-            is_enabled = re.findall('\nVAULT_ENABLED_FOR_DYNACONF=(.*)', self.envdata)
+            is_enabled = re.findall('^(?:.*\n)*VAULT_ENABLED_FOR_DYNACONF=(.*)', self.envdata)
```
lpramuk, line 28: Please replace it by 
`re.findall('^VAULT_URL_FOR_DYNACONF=(.*)', self.envdata, re.MULTILINE)` 

`^` and MULTILINE is what we need

lpramuk (changes requested):
Change RE to use `^` and `MULTILINE`

robottelo/utils/vault.py
jyejare: With the regex, it has now, it does not need `re.MULTILINE` also the `^` start of the line is covered already.

In future PR we will integrate DotEnv python lib as suggested by @ogajduse !

PR #12923, opened 2023-10-17
Fixture collection splitting restructured

jyejare:
The processing of the list of fixtures in the command line for in Pytest Plugin `Fixture Collection` is restructured.

Earlier:
```
# pytest --uses-fixtures fix1 fix2 fix3 tests/foreman
```

Now:
```
# pytest --uses-fixtures fix1,fix2,fix3 tests/foreman
```

Issue and Fix:
---------------
Earlier pytest is used to process all the non specific arguments after `--uses-fixtures` option , e.g in example above it used to consider `tests/foreman` as fixture as well :)

Now, it process only a string after an option and splits string by comma to obtain the list of fixtures to collect the tests for all those fixtures.

### DIFF
--- a/pytest_plugins/fixture_collection.py
+++ b/pytest_plugins/fixture_collection.py
@@ -13,7 +13,7 @@ def pytest_addoption(parser):
 
         example: pytest tests/foreman --uses-fixtures target_sat module_target_sat
     '''
-    parser.addoption("--uses-fixtures", nargs='+', help=help_text)
+    parser.addoption("--uses-fixtures", nargs='?', help=help_text)
 
 
 def pytest_collection_modifyitems(items, config):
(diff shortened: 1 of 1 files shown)


pytest_plugins/fixture_collection.py
```diff
@@ -22,17 +22,18 @@ def pytest_collection_modifyitems(items, config):
         return
 
     filter_fixtures = config.getvalue('uses_fixtures')
+    fixtures_list = filter_fixtures.split(',') if ',' in filter_fixtures else [filter_fixtures]
```
Gauravtalreja1, line 25: Shouldn't we support both spaces and commas for uses_fixtures here? 
```suggestion
    filter_fixtures = config.getvalue('uses_fixtures')
    fixtures = ' '.join(filter_fixtures.split(',')) if ',' in filter_fixtures else filter_fixtures
```

jyejare: Nop only comma, to support the space cli provides the way already which I m removing in the PR as its not working as its suppose to work.

And adding comma separated support.

Gauravtalreja1 (approved):

PR #12869, opened 2023-10-10
[6.12.z] Optional pytest Vault login and code nitpicking

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/12822

This PR brings optional logging to the vault from pytest session requested by the community. 

- Multiple options to bypass vault login,  where pytest session won't insist on logging into Vault. :
  - No `.env` file
  - `VAULT_ENABLED_FOR_DYNACONF` option is set to `False` in `.env` file
  - Commented out VAULT_ENABLED_FOR_DYNACONF` option in `.env` file
  - No `VAULT_ENABLED_FOR_DYNACONF` option in `.env` file
- Also made some code nitpicking changes !

PR #12822, opened 2023-10-05
Optional pytest Vault login and code nitpicking

jyejare:
This PR brings optional logging to the vault from pytest session requested by the community. 

- Multiple options to bypass vault login,  where pytest session won't insist on logging into Vault. :
  - No `.env` file
  - `VAULT_ENABLED_FOR_DYNACONF` option is set to `False` in `.env` file
  - Commented out VAULT_ENABLED_FOR_DYNACONF` option in `.env` file
  - No `VAULT_ENABLED_FOR_DYNACONF` option in `.env` file
- Also made some code nitpicking changes !

PR #12960, opened 2023-10-22
Fix vault login when VAULT_ENABLED_FOR_DYNACONF on 1st line

lpramuk:
When `.env` files contains `VAULT_ENABLED_FOR_DYNACONF=true` on very first line then this setting is ignored and vault login doesn't work

With merge wait till we merge cherrypicks of #12822 to `6.13.z` (#12868) and `6.12.z` (#12869)

### DIFF
--- a/robottelo/utils/vault.py
+++ b/robottelo/utils/vault.py
@@ -25,7 +25,7 @@ class Vault:
     def setup(self):
         if self.env_path.exists():
             self.envdata = self.env_path.read_text()
-            is_enabled = re.findall('\nVAULT_ENABLED_FOR_DYNACONF=(.*)', self.envdata)
+            is_enabled = re.findall('^VAULT_ENABLED_FOR_DYNACONF=(.*)', self.envdata, re.MULTILINE)
             if is_enabled:
                 self.vault_enabled = is_enabled[0]
             self.export_vault_addr()
@@ -34,7 +34,7 @@ class Vault:
         del os.environ['VAULT_ADDR']
 
     def export_vault_addr(self):
-        vaulturl = re.findall('VAULT_URL_FOR_DYNACONF=(.*)', self.envdata)[0]
+        vaulturl = re.findall('^VAULT_URL_FOR_DYNACONF=(.*)', self.envdata, re.MULTILINE)[0]
 
         # Set Vault CLI Env Var
         os.environ['VAULT_ADDR'] = vaulturl


lpramuk:
trigger: test-robottelo
pytest: tests/foreman/api/test_product.py

lpramuk:
trigger: test-robottelo
pytest: tests/foreman/api/test_repository_set.py

rmynar (changes requested):
I think the original idea was to ignore comments, which had the side-effect of ignoring first line.
Current fix may also match commented value.
I propose to use something like `^[^#]*[ ]*VAULT_ENABLED_FOR_DYNACONF`.

ogajduse:
Wouldn't #12909 solve your issue?

Anyway, I would personally prefer using https://github.com/theskumar/python-dotenv library for .env file read & write operations.

lpramuk:
> I think the original idea was to ignore comments, which had the side-effect of ignoring first line. Current fix may also match commented value. I propose to use something like `^[^#]*[ ]*VAULT_ENABLED_FOR_DYNACONF`.

It seeks after a setting not after commented setting as you suggest

robottelo/utils/vault.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, env_file='.env'):
     def setup(self):
         if self.env_path.exists():
             self.envdata = self.env_path.read_text()
-            is_enabled = re.findall('\nVAULT_ENABLED_FOR_DYNACONF=(.*)', self.envdata)
+            is_enabled = re.findall('VAULT_ENABLED_FOR_DYNACONF=(.*)', self.envdata)
```
rmynar, line 28: I had to play with this a bit...
I suggest using `^[\s]*VAULT_ENABLED_FOR_DYNACONF=(.*)`

robottelo/utils/vault.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, env_file='.env'):
...
             if is_enabled:
                 self.vault_enabled = is_enabled[0]
```
rmynar, line 30: `self.vault_enabled = is_enabled[-1]`

rmynar (changes requested):
When we consider following example file:
`VAULT_ENABLED_FOR_DYNACONF=false`
`#VAULT_ENABLED_FOR_DYNACONF=false`
`# VAULT_ENABLED_FOR_DYNACONF=false`
`   VAULT_ENABLED_FOR_DYNACONF=true`
With proposed changes line 1 matches but it is not used. Lines 2 and 3 don't match (they start with hash sign). Value on line 4 is used because it is last occurence.

However as @ogajduse metioned, there's a conflict with another PR and it would be better idea to use a library to manage .env configuration.

lpramuk:
trigger: test-robottelo
pytest: tests/foreman/api/test_repository_set.py

robottelo/utils/vault.py
lpramuk: `^` works only when it is on the first line since
`^` means start of the string not start of the line

robottelo/utils/vault.py
lpramuk: first or last  - what is the difference when you expect just single ocurrence

when it is duplicate the last would override prevoius ones .... hrm

robottelo/utils/vault.py
lpramuk: but using flag `re.MULTILINE` it can be changed to expected behavior -  `^` means start of the line

robottelo/utils/vault.py
rmynar: it's just a good practice to use the last value if there are multiple entries of one key. but it's best practice to don't repeat keys in config files.

the original problem (unable to read value when key is on first line) is fixed - approved

rmynar (approved):

jyejare:
@rmynar All the occurrences you mentioned should be handled well in #12909, please review/test that PR.

@ogajduse I was thinking for it but TBH we has a use case where we even replace the commented line of TOKEN env var which IMHO wont be able to do using dotenv lib.

jyejare:
@lpramuk I have enabled auto-merge on the https://github.com/SatelliteQE/robottelo/pull/12909 so you can close this PR.

12909 handles this issue and also have additional and important fix!

pondrejk (approved):
ack but needs conflict resolution

jyejare:
Hey @peterdragun @ColeHiggins2 and @lpramuk , we dont need this PR anymore, this has been already taken in PR #12909

lpramuk:
@jyejare you rejected my suggestion in your #12909 so this PR is still valid

JacobCallahan (approved):

JacobCallahan:
@lpramuk can you resolve the merge conflict? Afterward, ping me in slack and we'll get this merged!

PR #12961, opened 2023-10-23
satellite and capsule installation improvements

rmynar:
Applying changes from Component evaluation
- added _pit_ marker for Satellite and Capsule installation tests
- added various log checks (and removed stubbed tescases)
- removed redundant assertions

### DIFF
--- a/tests/foreman/installer/test_installer.py
+++ b/tests/foreman/installer/test_installer.py
@@ -1306,10 +1306,30 @@ def extract_help(filter='params'):
 def common_sat_install_assertions(satellite):
     sat_version = 'stream' if satellite.is_stream else satellite.version
     assert settings.server.version.release == sat_version
+
+    # no errors/failures in /var/log/messages
+    # ignore warnings (some messages contain string "error" but they're still only warnings)
+    # ignore "kernel: GPT: Use GNU Parted to correct GPT errors."
+    result = satellite.execute(
+        r'grep -i "error" /var/log/messages | grep -v "\] WARNING" | grep -v GPT'
+    )
+    assert len(result.stdout) == 0
+    # no errors/failures in /var/log/foreman/production.log
+    # ignore warnings
+    result = satellite.execute(r'grep -i "error" /var/log/foreman/production.log | grep -v "\[W\|"')
+    assert len(result.stdout) == 0
+    # no errors/failures in /var/log/foreman-installer/satellite.log
     result = satellite.execute(
         r'grep "\[ERROR" --after-context=100 /var/log/foreman-installer/satellite.log'
     )
     assert len(result.stdout) == 0
+    # no errors/failures in /var/log/httpd/*
+    result = satellite.execute(r'grep -iR "error" /var/log/httpd/*')
+    assert len(result.stdout) == 0
+    # no errors/failures in /var/log/candlepin/*
+    result = satellite.execute(r'grep -iR "error" /var/log/candlepin/*')
+    assert len(result.stdout) == 0
+
     result = satellite.cli.Health.check()
     assert 'FAIL' not in result.stdout
 
@@ -1332,7 +1352,6 @@ def install_satellite(satellite, installer_args):
         InstallerCommand(installer_args=installer_args).get_command(),
         timeout='30m',
     )
-    common_sat_install_assertions(satellite)
 
 
 @pytest.fixture(scope='module')
@@ -1361,6 +1380,7 @@ def sat_non_default_install(module_sat_ready_rhels):
 
 @pytest.mark.e2e
 @pytest.mark.tier1
+@pytest.mark.pit
 def test_capsule_installation(sat_default_install, cap_ready_rhel, default_org):
     """Run a basic Capsule installation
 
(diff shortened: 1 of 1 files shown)


rmynar:
"trigger": "test-robottelo"
"pytest": "tests/foreman/installer/test_installer.py -k 'test_satellite_installation or test_capsule_installation'"

tests/foreman/installer/test_installer.py
```diff
@@ -1306,10 +1306,30 @@ def extract_help(filter='params'):
 def common_sat_install_assertions(satellite):
     sat_version = 'stream' if satellite.is_stream else satellite.version
     assert settings.server.version.release == sat_version
+
+    # no errors/failures in /var/log/messages
+    # ignore warnings (some messages contain string "error" but they're still only warnings)
+    # ignore "kernel: GPT: Use GNU Parted to correct GPT errors."
+    result = satellite.execute(
+        r'grep -i "error" /var/log/messages | grep -v "\] WARNING" | grep -v GPT'
+    )
```
Griffin-Sullivan, line 1315: Seems like the test failed here. Did you come up with these assertions? Did support mention the coverage here wasn't good enough?

rmynar:
trigger: test-robottelo
pytest: tests/foreman/installer/test_installer.py -k 'test_satellite_installation or test_capsule_installation'

tests/foreman/installer/test_installer.py
rmynar: The idea to check syslog for errors was there in a stubbed testcase. I finally implemented it. But it returned lots of false positives. For example error messages of applications not related to foreman/satellite that were logged even before the installation.  Or some minor warnings and info messages containg the word "error". So instead of syslog I decided to use journalctl which allows advanced filtering - see my next commit. Suggestions to adjust the filtering are welcome - we can implement them within this PR or later.

rmynar: And latest PRT finished successfully :-)

Griffin-Sullivan (approved):

tests/foreman/installer/test_installer.py
```diff
@@ -1361,6 +1378,7 @@ def sat_non_default_install(module_sat_ready_rhels):
 
 @pytest.mark.e2e
 @pytest.mark.tier1
+@pytest.mark.pit
```
ogajduse, line 1381: Great to see this inclusion, @rmynar! Please get in touch with me after this PR is merged. I would like to do a test execution of the PIT pipeline to see how the test behaves in it.

tests/foreman/installer/test_installer.py
```diff
@@ -1306,10 +1306,28 @@ def extract_help(filter='params'):
...
     assert settings.server.version.release == sat_version
+
+    # no errors/failures in journald
+    result = satellite.execute(
+        r'journalctl -q --no-pager -b -p err -u "pulpcore" -u "foreman" -u "foreman-proxy" -u "httpd" -u "postgresql" -u "pulpcore-api" -u "pulpcore-worker"'
+    )
+    assert len(result.stdout) == 0
+    # no errors/failures in /var/log/foreman/production.log
+    # ignore warnings
+    result = satellite.execute(r'grep -i "error" /var/log/foreman/production.log | grep -v "\[W\|"')
```
pondrejk, line 1317: the second grep should be `grep -vE "\[W\|"` so that the regexp is understood, otherwise it will always return the empty list. That being said, what about the Info lines that can contain mentions of error as well

pondrejk (changes requested):

tests/foreman/installer/test_installer.py
rmynar: now I grep lines with severity "E" (error) only. and I added the extended regexp option. please check

rmynar:
trigger: test-robottelo
pytest: tests/foreman/installer/test_installer.py -k 'test_satellite_installation or test_capsule_installation'

pondrejk (approved):
ack pending passing prt

rmynar:
trigger: test-robottelo
pytest: tests/foreman/installer/test_installer.py -k 'test_satellite_installation or test_capsule_installation'

rmynar:
previous prt failure was caused by product bug, so i entered a bz and made the assertion conditional depending on bz status. current prt passed successfully

tests/foreman/installer/test_installer.py
```diff
@@ -1395,14 +1416,29 @@ def test_capsule_installation(sat_default_install, cap_ready_rhel, default_org):
     assert sat_default_install.api.Capsule().search(
         query={'search': f'name={cap_ready_rhel.hostname}'}
     )[0]
+
+    # no errors/failures in journald
+    result = cap_ready_rhel.execute(
+        r'journalctl -q --no-pager -b -p err -u "pulpcore" -u "foreman" -u "foreman-proxy" -u "httpd" -u "postgresql" -u "pulpcore-api" -u "pulpcore-worker"'
```
ogajduse, line 1422: If I am not wrong, you did not cover all services in this list. I would recommend you store all service names in a list that could be placed in constants and generate the command based on it. The list should ideally contain all services that satellite-maintain checks in its `service status` check.

tests/foreman/installer/test_installer.py
```diff
@@ -1395,14 +1416,29 @@ def test_capsule_installation(sat_default_install, cap_ready_rhel, default_org):
...
+    # no errors/failures /var/log/foreman-installer/capsule.log
     result = cap_ready_rhel.execute(
         r'grep "\[ERROR" --after-context=100 /var/log/foreman-installer/capsule.log'
     )
     assert len(result.stdout) == 0
+    # no errors/failures in /var/log/httpd/*
+    result = cap_ready_rhel.execute(r'grep -iR "error" /var/log/httpd/*')
+    assert len(result.stdout) == 0
+    # no errors/failures in /var/log/foreman-proxy/*
+    result = cap_ready_rhel.execute(r'grep -iR "error" /var/log/foreman-proxy/*')
```
ogajduse, line 1439: It would also be great to have more context if an error occurs. I like the approach above with `--after-context`.
Can you please consider adding it to the newly added commands? Eventually, you could use `--context` that would give us a context of `n` lines above as well as below the match.

tests/foreman/installer/test_installer.py
```diff
@@ -1791,6 +1762,7 @@ def test_installer_cap_pub_directory_accessibility(capsule_configured):
 @pytest.mark.tier1
 @pytest.mark.build_sanity
 @pytest.mark.first_sanity
+@pytest.mark.pit_client
```
ogajduse, line 1765: If you want to keep the `pit_client` marker here, you have to ensure that `lru_sat_ready_rhel` has a configurable non-hardcoded workflow name coming from settings that is not `settings.server.deploy_workflow`. It should ideally be `settings.content_host.rhelX.vm.workflow`. Also `deploy_rhel_version` should be configured either to `8`  - simply X version only to ensure we are not requesting the released version of RHEL instead of the non-released one. To get an insight on how pytest is configured in PIT testing, please see `automationPITPipeline.groovy`.


https://github.com/SatelliteQE/robottelo/blob/0a0966ba4a895bf91635eac4b0a7e44b2aaed92d/robottelo/hosts.py#L71-L80

ogajduse (changes requested):

tests/foreman/installer/test_installer.py
ogajduse: Opened #13138 for this.

tests/foreman/installer/test_installer.py
ogajduse: non-blocking ^

ogajduse (approved):

jameerpathan111 (approved):

PR #12963, opened 2023-10-23
[6.15 RFE] Prepare users for SCA-ONLY Hammer

ColeHiggins2:
6.15 Features automation: SAT-20202

This test is verifying that hammer notifies users that Simple Content Access will be required for all organizations in Satellite 6.16

***As of now, Stream Satellite is showing ```Katello 4.12``` instead of ```Satellite 6.16```. Keeping in Draft state until changes are made***

### DIFF
--- a/tests/foreman/cli/test_subscription.py
+++ b/tests/foreman/cli/test_subscription.py
@@ -299,12 +299,12 @@ def test_positive_prepare_for_sca_only_hammer(function_org, target_sat):
     org_results = target_sat.execute(
         f'hammer organization update --id {function_org.id} --simple-content-access false'
     )
-    assert 'Simple Content Access will be required for all organizations in Katello 4.12' in str(
+    assert 'Simple Content Access will be required for all organizations in Satellite 6.16' in str(
         org_results.stderr
     )
     sca_results = target_sat.execute(
         f'hammer simple-content-access disable --organization-id {function_org.id}'
     )
-    assert 'Simple Content Access will be required for all organizations in Katello 4.12' in str(
+    assert 'Simple Content Access will be required for all organizations in Satellite 6.16' in str(
         sca_results.stderr
     )
--- a/tests/foreman/cli/test_subscription.py
+++ b/tests/foreman/cli/test_subscription.py
@@ -284,3 +284,27 @@ def test_positive_auto_attach_disabled_golden_ticket(
     with pytest.raises(CLIReturnCodeError) as context:
         target_sat.cli.Host.subscription_auto_attach({'host-id': host_id})
     assert "This host's organization is in Simple Content Access mode" in str(context.value)
+
+
+@pytest.mark.tier2
+def test_positive_prepare_for_sca_only_hammer(function_org, target_sat):
+    """Verify that upon certain actions, Hammer notifies users that Simple Content Access
+        will be required for all organizations in Satellite 6.16
+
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_subscription.py
```diff
@@ -284,3 +284,27 @@ def test_positive_auto_attach_disabled_golden_ticket(
...
+        will be required for all organizations in Satellite 6.16
+    """
+    org_results = target_sat.execute(
+        f'hammer organization update --id {function_org.id} --simple-content-access false'
+    )
+    assert 'Simple Content Access will be required for all organizations in Satellite 6.16' in str(
+        org_results.stderr
+    )
+    sca_results = target_sat.execute(
+        f'hammer simple-content-access disable --organization-id {function_org.id}'
```
Gauravtalreja1, line 306: Can we update it to use robottelo cli method `SimpleContentAccess.disable()`?

tests/foreman/cli/test_subscription.py
```diff
@@ -284,3 +284,27 @@ def test_positive_auto_attach_disabled_golden_ticket(
...
+    """Verify that upon certain actions, Hammer notifies users that Simple Content Access
+        will be required for all organizations in Satellite 6.16
+
+    :id: 8753adb9-28ff-4a16-b04e-0af532452f32
+
+    :expectedresults: Hammer returns a message notifying users that Simple Content Access
+        will be required for all organizations in Satellite 6.16
+    """
+    org_results = target_sat.execute(
+        f'hammer organization update --id {function_org.id} --simple-content-access false'
```
Gauravtalreja1, line 300: same here

Gauravtalreja1:
@ColeHiggins2 As 6.15.z branching is done already, you'll need to cherrypick this to 6.15.z and flip the labels on this PR

damoore044:
trigger: test-robottelo
pytest: tests/foreman/cli/test_subscription.py -k test_positive_prepare_for_sca_only_ak
airgun: 1026

ColeHiggins2:
DO not merge until tested on downstream satellite

vsedmik (commented):
Should we also get/assert that warning on CREATE operation?
```
hammer organization create --name OWCA --simple-content-access False
```

jyejare:
@ColeHiggins2 Moving this to draft state as per your last comments! Please remove the from draft when its ready to merge !

tests/foreman/cli/test_subscription.py
```diff
@@ -284,3 +284,27 @@ def test_positive_auto_attach_disabled_golden_ticket(
...
+    )
+    assert 'Simple Content Access will be required for all organizations in Satellite 6.16' in str(
```
jyejare, line 302:
```suggestion
    assert 'Simple Content Access will be required for all organizations in Satellite' in str(
```
Could we eliminate satellite version so that it won't fail in the future due to version change in the string ?

jyejare (approved):
Looks good with a non-blocking comment !

tests/foreman/cli/test_subscription.py
Gauravtalreja1: @ColeHiggins2 as you need to verify output from command's stderr, and you're not able to use cli.SimpleContentAccess for this because of ignore_stderr=True argument, you could remove it from here and update the PR for the same https://github.com/SatelliteQE/robottelo/blob/master/robottelo/cli/simple_content_access.py#L33

PR #12978, opened 2023-10-25
Stream fix errata:UI (test_end_to_end)

damoore044:
Refactor errata e2e with updated standards, contenthost through Global Registration, enable repositories with subscription manager. 

Remove `setting_update` of remote_execution_by_default:True ([BZ 2029192](https://bugzilla.redhat.com/show_bug.cgi?id=2029192)), this user setting has 
been removed for 6.15.0, and by default is set to True.

Used `request.addfinalizer` for cleanup of entities modified in the new fixture `registered_contenthost`, to prevent conflicts between parameterized runs.
~Depends on navigation fix in [airgun 1018](https://github.com/SatelliteQE/airgun/pull/1018).~ Merged

### DIFF
--- a/robottelo/cli/factory.py
+++ b/robottelo/cli/factory.py
@@ -1692,9 +1692,10 @@ def setup_org_for_a_custom_repo(options=None):
         raise CLIFactoryError(f'Failed to publish new version of content view\n{err.msg}')
     # Get the content view info
     cv_info = ContentView.info({'id': cv_id})
-    lce_promoted = cv_info['lifecycle-environments']
     assert len(cv_info['versions']) > 0
-    cvv = sorted(cvv['id'] for cvv in cv_info['versions'])[-1]
+    cv_info['versions'].sort(key=lambda version: version['id'])
+    cvv = cv_info['versions'][-1]
+    lce_promoted = cv_info['lifecycle-environments']
     # Promote version to next env
     try:
         if env_id not in [int(lce['id']) for lce in lce_promoted]:
--- a/robottelo/cli/factory.py
+++ b/robottelo/cli/factory.py
@@ -1693,12 +1693,17 @@ def setup_org_for_a_custom_repo(options=None):
     # Get the content view info
     cv_info = ContentView.info({'id': cv_id})
     lce_promoted = cv_info['lifecycle-environments']
-    cvv = cv_info['versions'][-1]
+    assert len(cv_info['versions']) > 0
+    cvv = sorted(cvv['id'] for cvv in cv_info['versions'])[-1]
     # Promote version to next env
     try:
         if env_id not in [int(lce['id']) for lce in lce_promoted]:
             ContentView.version_promote(
-                {'id': cvv['id'], 'organization-id': org_id, 'to-lifecycle-environment-id': env_id}
+                {
+                    'id': cvv['id'],
+                    'organization-id': org_id,
+                    'to-lifecycle-environment-id': env_id,
+                }
             )
     except CLIReturnCodeError as err:
         raise CLIFactoryError(f'Failed to promote version to next environment\n{err.msg}')
--- a/tests/foreman/ui/test_errata.py
+++ b/tests/foreman/ui/test_errata.py
@@ -64,30 +64,6 @@ def _generate_errata_applicability(hostname):
     host.errata_applicability(synchronous=False)
 
 
-def _install_client_package(client, package):
-    """Install a package in virtual machine client.
-    Errata applicability regenerated after successful yum execution.
-
-    :param client: The Virtual machine client.
-    :param package: the package to install in virtual machine client.
-    :returns: True if package installed successfully, False otherwise.
-    """
-    result = client.execute(f'yum install -y {package}')
-    return result.status == 0
-
-
-def _remove_client_package(client, package):
-    """Remove a package in virtual machine client.
-    Errata applicability regenerated after successful yum execution.
-
-    :param client: The Virtual machine client.
-    :param package: the package (full version, or name) to remove from virtual machine client.
-    :returns: True if a package was removed successfully, False otherwise.
-    """
-    result = client.execute(f'yum erase -y {package}')
-    return result.status == 0
-
-
 def _set_setting_value(setting_entity, value):
     """Set setting value.
 
@@ -174,7 +150,7 @@ def registered_contenthost(
     rhel_contenthost,
     module_org,
     module_lce,
-    module_published_cv,
+    module_cv,
     module_target_sat,
     request,
     repos=[CUSTOM_REPO_URL],
@@ -193,7 +169,6 @@ def registered_contenthost(
     custom_products = []
     custom_repos = []
     for repo_url in repos:
-        rhel_contenthost.create_custom_repos(custom_repo=repo_url)
         # Publishes a new cvv, associates org, ak, cv, with custom repo:
         custom_repo_info = module_target_sat.cli_factory.setup_org_for_a_custom_repo(
             {
@@ -201,17 +176,17 @@ def registered_contenthost(
                 'organization-id': module_org.id,
                 'lifecycle-environment-id': module_lce.id,
                 'activationkey-id': activation_key.id,
-                'content-view-id': module_published_cv.id,
+                'content-view-id': module_cv.id,
             }
         )
         custom_products.append(custom_repo_info['product-id'])
         custom_repos.append(custom_repo_info['repository-id'])
 
     # Promote newest version with all content
-    module_published_cv = module_published_cv.read()
-    module_published_cv.version.sort(key=lambda version: version.id)
-    module_published_cv.version[-1].promote(data={'environment_ids': module_lce.id})
-    module_published_cv = module_published_cv.read()
+    module_cv = module_cv.read()
+    module_cv.version.sort(key=lambda version: version.id)
+    module_cv.version[-1].promote(data={'environment_ids': module_lce.id})
+    module_cv = module_cv.read()
 
     result = rhel_contenthost.register(
         activation_keys=activation_key.name,
@@ -235,33 +210,33 @@ def registered_contenthost(
     @request.addfinalizer
     # Cleanup for in between parameterized runs
     def cleanup():
-        nonlocal rhel_contenthost, module_published_cv, custom_repos, custom_products, activation_key
+        nonlocal rhel_contenthost, module_cv, custom_repos, custom_products, activation_key
         rhel_contenthost.unregister()
         activation_key.delete()
         # Remove CV from all lifecycle-environments
         module_target_sat.cli.ContentView.remove_from_environment(
             {
-                'id': module_published_cv.id,
+                'id': module_cv.id,
                 'organization-id': module_org.id,
                 'lifecycle-environment-id': module_lce.id,
             }
         )
         module_target_sat.cli.ContentView.remove_from_environment(
             {
-                'id': module_published_cv.id,
+                'id': module_cv.id,
                 'organization-id': module_org.id,
                 'lifecycle-environment': 'Library',
             }
         )
         # Delete all CV versions
-        module_published_cv = module_published_cv.read()
-        for version in module_published_cv.version:
+        module_cv = module_cv.read()
+        for version in module_cv.version:
             version.delete()
         # Remove repos from CV, delete all custom repos and products
         for repo_id in custom_repos:
             module_target_sat.cli.ContentView.remove_repository(
                 {
-                    'id': module_published_cv.id,
+                    'id': module_cv.id,
                     'repository-id': repo_id,
                 }
             )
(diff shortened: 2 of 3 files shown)


damoore044:
trigger: test-robottelo
pytest: tests/foreman/ui/test_errata.py -k test_end_to_end

damoore044:
trigger: test-robottelo
pytest: tests/foreman/ui/test_errata.py -k test_end_to_end

damoore044:
trigger: test-robottelo
pytest: tests/foreman/ui/test_errata.py -k test_end_to_end

damoore044:
trigger: test-robottelo
pytest: tests/foreman/ui/test_errata.py -k test_end_to_end
airgun: 1018

damoore044:
trigger: test-robottelo
pytest: tests/foreman/ui/test_errata.py -k test_end_to_end
airgun: 1018

damoore044:
trigger: test-robottelo
pytest: tests/foreman/ui/test_errata.py -k test_end_to_end

tests/foreman/ui/test_errata.py
```diff
@@ -163,37 +173,89 @@ def vm(module_repos_collection_with_setup, rhel7_contenthost, target_sat):
...
+
+    for custom_repo_id in custom_repos:
+        custom_repo = module_target_sat.api.Repository(id=custom_repo_id).read()
+        assert custom_repo
+        result = custom_repo.sync()['humanized']
+        assert (
+            len(result['errors']) == 0
+        ), f'Failed to sync custom repository [id: {custom_repo_id}]:\n{str(result["errors"])}'
+
+    yield rhel_contenthost
```
damoore044, line 234:
```suggestion
    return rhel_contenthost
```

tests/foreman/ui/test_errata.py
```diff
@@ -163,37 +173,89 @@ def vm(module_repos_collection_with_setup, rhel7_contenthost, target_sat):
...
+        custom_repos.append(custom_repo_id)
+
+    result = rhel_contenthost.register(
+        activation_keys=activation_key.name,
+        lifecycle_environment=module_lce,
+        target=module_target_sat,
+        org=module_org,
+        loc=None,
+    )
+    assert result.status == 0, f'Failed to register host:\n{result.stdout}'
```
JacobCallahan, line 223: is stdout the most appropriate to use here or would stderr be better for the error message?

damoore044: right, the full output is probably unnecessary here, I'll push a change to `result.stderr`

damoore044:
trigger: test-robottelo
pytest: tests/foreman/ui/test_errata.py -k test_end_to_end

JacobCallahan (approved):

tests/foreman/ui/test_errata.py
```diff
@@ -75,6 +78,20 @@ def _install_client_package(client, package, errata_applicability=False):
...
+    """Remove a package in virtual machine client.
+
+    :param client: The Virtual machine client.
+    :param package: the package (full version, or name) to remove from virtual machine client.
+    :param errata_applicability: If True, force host to generate errata applicability.
+    :returns: True if a package was removed successfully, False otherwise.
+    """
+    result = client.execute(f'yum remove -y {package}')
+    if errata_applicability:
+        _generate_errata_applicability(client.hostname)
```
jeremylenz, line 91: Unless the host is configured not to upload the package profile on yum transactions (which is not the default and very unlikely), a `yum -y remove` should always lead to an automatic applicability calculation for that host. I'm thinking this explicit call may never be needed?

tests/foreman/ui/test_errata.py
```diff
@@ -163,37 +173,89 @@ def vm(module_repos_collection_with_setup, rhel7_contenthost, target_sat):
...
+                'lifecycle-environment-id': module_lce.id,
+                'activationkey-id': activation_key.id,
+                'content-view-id': content_view.id,
+            }
+        )['repository-id']
+        custom_repos.append(custom_repo_id)
+
+    result = rhel_contenthost.register(
+        activation_keys=activation_key.name,
+        lifecycle_environment=module_lce,
```
jeremylenz, line 218: Is `rhel_contenthost.register` the one that uses global registration?

I know this isn't in the scope of this PR, but the `lifecycle_environment` option probably needs to be removed, in that case. We recently removed it from the web UI.

jeremylenz (commented):
A couple thoughts below, for what they're worth :)

In general we're moving away from the term "content host" in favor of simply "host."

tests/foreman/ui/test_errata.py
damoore044: Thanks! the existing `_install_client_package()` also has an `errata_applicability` option.
Would an automatic applicability calculation occur for `yum -y install` as well? If so I can update that method too.

jeremylenz: Yeah, it should be for any package action.

tests/foreman/ui/test_errata.py
damoore044: Yes, `contenthost.register()` ([hosts.py@693](https://github.com/SatelliteQE/robottelo/blob/83ca28419c129aad304d49c89f758af36b9a74b7/robottelo/hosts.py#L693)), should make use of a Global Registration template. 
The older `contenthost.register_contenthost()` needs katello_ca before registering.

I'll try without passing `lce` for registration 👍

damoore044: `lifecycle-environment` doesn't need to be passed to `register()`.
Thanks @jeremylenz !

damoore044:
trigger: test-robottelo
pytest: tests/foreman/ui/test_errata.py -k test_end_to_end

damoore044:
trigger: test-robottelo
pytest: tests/foreman/ui/test_errata.py -k test_end_to_end

tests/foreman/ui/test_errata.py
```diff
@@ -163,37 +169,88 @@ def vm(module_repos_collection_with_setup, rhel7_contenthost, target_sat):
...
+    """RHEL ContentHost registered in satellite,
+    Using SCA and global registration.
+
+    :param repos: list of upstream URLs for custom repositories,
+        default to CUSTOM_REPO_URL
+    """
+    content_view = module_target_sat.api.ContentView(
+        organization=module_org,
+        environment=[module_lce],
+    ).create()
```
vsedmik, line 189: Can we go with `module_cv` here instead (the same one used in test body)?

tests/foreman/ui/test_errata.py
```diff
@@ -163,37 +169,88 @@ def vm(module_repos_collection_with_setup, rhel7_contenthost, target_sat):
...
+
+    activation_key = module_target_sat.api.ActivationKey(
+        organization=module_org,
+        environment=module_lce,
+    ).create()
+
+    custom_repos = []
+    for repo_url in repos:
+        rhel_contenthost.create_custom_repos(custom_repo=repo_url)
```
vsedmik, line 198: I don't think we should need to create custom repos at the host side as we should consume the repos through the Satellite.

tests/foreman/ui/test_errata.py
```diff
@@ -221,46 +281,96 @@ def test_end_to_end(
...
+    # Capture product and repository with the desired content
+    assert len(product_list := module_target_sat.api.Product(organization=module_org).search()) > 0
+    _product = product_list[-1].read()
+    assert len(_product.repository) > 0
+    _repository = _product.repository[-1].read()
+    # Remove custom package if present, then install outdated version
+    _remove_client_package(registered_contenthost, FAKE_1_CUSTOM_PACKAGE_NAME)
+    assert _install_client_package(registered_contenthost, FAKE_1_CUSTOM_PACKAGE)
+    assert (
+        appl_errata := registered_contenthost.applicable_errata_count
```
vsedmik, line 294: PRT just blows my mind here with `E       assert 1 == 1`
I though I understood how the walrus operator works, also locally this works. No idea what's wrong in PRT's python version walrus implementation. Maybe try without walrus.

robottelo/cli/factory.py
```diff
@@ -1693,12 +1693,17 @@ def setup_org_for_a_custom_repo(options=None):
     # Get the content view info
     cv_info = ContentView.info({'id': cv_id})
     lce_promoted = cv_info['lifecycle-environments']
-    cvv = cv_info['versions'][-1]
+    assert len(cv_info['versions']) > 0
+    cvv = sorted(cvv['id'] for cvv in cv_info['versions'])[-1]
```
Gauravtalreja1, line 1697: Non-blocking, Sorting is fine here, but why we're explicitly hitting this issue in stream(seen a couple of PRs for this), so just wondering if are you aware of any related changes in stream? @damoore044 @vsedmik @sambible

tests/foreman/ui/test_errata.py
```diff
@@ -61,17 +64,27 @@ def _generate_errata_applicability(hostname):
     host.errata_applicability(synchronous=False)
 
 
-def _install_client_package(client, package, errata_applicability=False):
+def _install_client_package(client, package):
     """Install a package in virtual machine client.
+    Errata applicability regenerated after successful yum execution.
```
Gauravtalreja1, line 69: Without errata applicability check here, why we keeping functions for just single install and remove tasks, can't we get rid of them instead?

tests/foreman/ui/test_errata.py
```diff
@@ -221,46 +285,105 @@ def test_end_to_end(
...
+        # UTC timing for install task and session
+        install_start = datetime.strptime(task_status['started_at'], _UTC_format)
+        install_end = datetime.strptime(task_status['ended_at'], _UTC_format)
+        assert (install_end - install_start).total_seconds() <= 60
+        assert (install_end - datetime_utc_start).total_seconds() <= 600
+
+        # Find bulk generate applicability task
+        results = module_target_sat.wait_for_tasks(
+            search_query=(
+                f'"Bulk generate applicability for host {registered_contenthost.hostname}"'
```
Gauravtalreja1, line 361:
```suggestion
                f'Bulk generate applicability for host {registered_contenthost.hostname}'
```

robottelo/cli/factory.py
vsedmik: @Gauravtalreja1  The issue is not specific to stream, I've seen it in older versions before and [this BZ](https://bugzilla.redhat.com/show_bug.cgi?id=2209758) was reported already. However, it feels to happen more frequently in stream - same passes in z-streams but fails in stream.

tests/foreman/ui/test_errata.py
damoore044: I had issues during registration (error for no enabled repos) before using this method, I'll try again without. Could have been related to some other part of setup.

tests/foreman/ui/test_errata.py
damoore044: I had a problem promoting a published version of `module_cv`,  
I've implemented `module_published_cv` and promote the newest version after associating repos/content.

tests/foreman/ui/test_errata.py
damoore044: @JacobCallahan and I had some discussion around this, there must be some sort of ambiguity in the assignment I am not seeing. I see success assigning the `applicable_errata` first, and then arresting on the next line

damoore044:
trigger: test-robottelo
pytest: tests/foreman/ui/test_errata.py -k test_end_to_end

damoore044:
trigger: test-robottelo
pytest: tests/foreman/ui/test_errata.py -k test_end_to_end

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/ui/test_errata.py -k test_end_to_end

Griffin-Sullivan (approved):

tests/foreman/ui/test_errata.py
```diff
@@ -163,37 +145,132 @@ def vm(module_repos_collection_with_setup, rhel7_contenthost, target_sat):
...
+        result = custom_repo.sync()['humanized']
+        assert (
+            len(result['errors']) == 0
+        ), f'Failed to sync custom repository [id: {custom_repo_id}]:\n{str(result["errors"])}'
+
+    yield rhel_contenthost
+
+    @request.addfinalizer
+    # Cleanup for in between parameterized runs
+    def cleanup():
```
vsedmik, line 212: It looks that in this cleanup we just roll back all the changes so that we have the module-scoped fixtures (mostly `module_cv`) ready for the next RHEL contenthost (via parametrization). What is the idea behind this need? Isn't it just fine to go with function-scoped fixtures instead module-scoped?

damoore044: I could switch them to function fixtures, but may not be a simple change. 
Many of the errata tests already run using module scope fixtures, so was keeping those for consistency.

vsedmik: Ok, no worries, merged already.

damoore044:
Future Notes: Will probably also work without extensive cleanup, by using function-scoped cv fixture or custom. 
Vlad made a good point today however- the teardown is nice testing to have, of a graceful cleanup of entities being used to host content w/ registered host. 
I wonder if we have any other testing around cleanup for content & entities of a reg. host? Looking into that

PR #12988, opened 2023-10-26
[6.14.z] Add test to ensure katello-agent availability in Client repos

vsedmik:
Katello-agent was deprecated and removed from 6.15 and upstream repos, but we still need to have it available in Client repos for older Satellite versions which support it. This test just verifies that.

PR #13008, opened 2023-10-31
Fix Client repos in constants

vsedmik:
"Cherry-pick" to `6.14.z` is done in #12988. Also, see the PRT results there.


ogajduse (approved):

ogajduse:
Thanks for raising it. Could you please cherry-pick the patch to `6.13.z` and `6.12.z`? I would say that labeling this one with the respective tags should do.

Gauravtalreja1 (approved):
Ack, Thank you @vsedmik

PR #13016, opened 2023-10-31
Check if VAULT_ADDR env var exists before deleting it

dosas:
Currently this throws an error in teardown if `VAULT_ADDR` is not set.

### DIFF
--- a/robottelo/utils/vault.py
+++ b/robottelo/utils/vault.py
@@ -31,7 +31,8 @@ class Vault:
             self.export_vault_addr()
 
     def teardown(self):
-        del os.environ['VAULT_ADDR']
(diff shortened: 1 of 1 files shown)


omkarkhatavkar:

jyejare:
Setting the VAULT_ADDR is mandatory and is being done in Context manager itself.

Whats unsetting that var for you ?

dosas:
I still had some dummy values set in .env file for
```
VAULT_MOUNT_POINT_FOR_DYNACONF
VAULT_PATH_FOR_DYNACONF
```
That is why it executed that path. 
Removing the file fixed it for me.

jyejare (approved):

PR #13027, opened 2023-11-01
Remove BROKER_DIRECTORY from robottelo config

JacobCallahan:
This is something that broker handles on its own and, due to import timing, isn't handled correctly in robottelo anyway. Removing the field and config mechanism should eliminate confusion around this behavior.

### DIFF
--- a/conf/broker.yaml.template
+++ b/conf/broker.yaml.template
@@ -1,7 +1,5 @@
 BROKER:
-    # The path where your broker settings and inventory are located
-    # If you leave it blank, the default is the output of `broker --version`
-    BROKER_DIRECTORY:
+    # Broker has its own config which you can find by running `broker --version`
     HOST_WORKFLOWS:
         POWER_CONTROL: vm-power-operation
         EXTEND: extend-vm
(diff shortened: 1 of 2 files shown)


conf/broker.yaml.template
```diff
@@ -1,7 +1,5 @@
 BROKER:
-    # The path where your broker settings and inventory are located
-    # If you leave it blank, the default is the output of `broker --version`
-    BROKER_DIRECTORY:
+    # Broker has its own config which you can find by running `broker --version`
```
jyejare, line 2: Many of us generally likes to keep conf files in robottelo root directory, so what if we could have made a broker to choose the conf file path rather than keeping it at fixed location.

jyejare (commented):
ACK pending question !

ColeHiggins2 (approved):

conf/broker.yaml.template
JacobCallahan: you can do that today, but it still wouldn't work correctly with robottelo setting it. just `export BROKER_DIRECTORY=/path/to/robottelo`

vsedmik (approved):

PR #13043, opened 2023-11-03
Markers as test property in reports

jyejare:
Test Markers as test metadata in Polarion, RP and Ibutsu reporting tools.

These markers would be used in RP and ibutsu to filter the results based on markers !


jyejare:
trigger: test-robottelo

jyejare:
All tests are passing ! Have an ACK and @rplevka confirmed that Report Portal is ready to take a new property.

PR #13070, opened 2023-11-10
updating the fedora image to 38 as latest using python 3.12

omkarkhatavkar:
#### Problem Statement:
Using the latest Fedora 39 image with Python 3.12 as the default version is causing failures in building the SSH2 Python library, leading to an unsuccessful build of Robottelo.

#### Solution:
Pin the Dockerfile version for Robottelo to Fedora 38, which uses Python 3.11, to resolve compatibility issues and enable a successful build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM fedora
+FROM fedora:38
 MAINTAINER https://github.com/SatelliteQE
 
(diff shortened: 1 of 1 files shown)


JacobCallahan (changes requested):
ACK on this change, but can you also remove the line where it downloads the config file for broker?

JacobCallahan (approved):

PR #13082, opened 2023-11-14
Added SCA enabled manifest in activation key and removed duplicate fixture

shweta83:
### Problem Statement
Fix Registration test cases failing because of changed activation key fixture.
There was duplicate of the fixture `module_activation_key` so removed it from test_host.py and using sca_enabled_manifest in pytest_fixtures/component/activationkey.py

### Solution
Updated activation key fixture in the tests

### DIFF
--- a/pytest_fixtures/component/activationkey.py
+++ b/pytest_fixtures/component/activationkey.py
@@ -6,13 +6,12 @@ from robottelo.utils.manifest import clone
 
 
 @pytest.fixture(scope='module')
-def module_activation_key(module_entitlement_manifest_org, module_target_sat):
+def module_activation_key(module_sca_manifest_org, module_target_sat):
     """Create activation key using default CV and library environment."""
     activation_key = module_target_sat.api.ActivationKey(
-        auto_attach=True,
-        content_view=module_entitlement_manifest_org.default_content_view.id,
-        environment=module_entitlement_manifest_org.library.id,
-        organization=module_entitlement_manifest_org,
+        content_view=module_sca_manifest_org.default_content_view.id,
+        environment=module_sca_manifest_org.library.id,
+        organization=module_sca_manifest_org,
     ).create()
     return activation_key
 
--- a/tests/foreman/api/test_registration.py
+++ b/tests/foreman/api/test_registration.py
@@ -30,9 +30,9 @@ pytestmark = pytest.mark.tier1
 @pytest.mark.e2e
 @pytest.mark.no_containers
 def test_host_registration_end_to_end(
-    module_org,
+    module_entitlement_manifest_org,
     module_location,
-    module_ak_with_synced_repo,
+    module_activation_key,
     module_target_sat,
     module_capsule_configured,
     rhel_contenthost,
@@ -50,9 +50,10 @@ def test_host_registration_end_to_end(
 
     :customerscenario: true
     """
+    org = module_entitlement_manifest_org
     command = module_target_sat.api.RegistrationCommand(
-        organization=module_org,
-        activation_keys=[module_ak_with_synced_repo.name],
+        organization=org,
+        activation_keys=[module_activation_key.name],
         location=module_location,
     ).create()
 
@@ -66,13 +67,13 @@ def test_host_registration_end_to_end(
 
     # Update module_capsule_configured to include module_org/module_location
     nc = module_capsule_configured.nailgun_smart_proxy
-    module_target_sat.api.SmartProxy(id=nc.id, organization=[module_org]).update(['organization'])
+    module_target_sat.api.SmartProxy(id=nc.id, organization=[org]).update(['organization'])
     module_target_sat.api.SmartProxy(id=nc.id, location=[module_location]).update(['location'])
 
     command = module_target_sat.api.RegistrationCommand(
         smart_proxy=nc,
-        organization=module_org,
-        activation_keys=[module_ak_with_synced_repo.name],
+        organization=org,
+        activation_keys=[module_activation_key.name],
         location=module_location,
         force=True,
     ).create()
@@ -92,7 +93,11 @@ def test_host_registration_end_to_end(
 @pytest.mark.tier3
 @pytest.mark.rhel_ver_match('[^6]')
 def test_positive_allow_reregistration_when_dmi_uuid_changed(
-    module_org, rhel_contenthost, target_sat, module_ak_with_synced_repo, module_location
+    module_sca_manifest_org,
+    rhel_contenthost,
+    target_sat,
+    module_activation_key,
+    module_location,
 ):
     """Register a content host with a custom DMI UUID, unregistering it, change
     the DMI UUID, and re-registering it again
@@ -109,10 +114,11 @@ def test_positive_allow_reregistration_when_dmi_uuid_changed(
     """
     uuid_1 = str(uuid.uuid1())
     uuid_2 = str(uuid.uuid4())
+    org = module_sca_manifest_org
     target_sat.execute(f'echo \'{{"dmi.system.uuid": "{uuid_1}"}}\' > /etc/rhsm/facts/uuid.facts')
     command = target_sat.api.RegistrationCommand(
-        organization=module_org,
-        activation_keys=[module_ak_with_synced_repo.name],
+        organization=org,
+        activation_keys=[module_activation_key.name],
         location=module_location,
     ).create()
     result = rhel_contenthost.execute(command)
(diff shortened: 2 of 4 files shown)


tests/foreman/api/test_registration.py
```diff
@@ -90,7 +91,11 @@ def test_host_registration_end_to_end(
 @pytest.mark.tier3
 @pytest.mark.rhel_ver_match('[^6]')
 def test_positive_allow_reregistration_when_dmi_uuid_changed(
-    module_org, rhel_contenthost, target_sat, module_ak_with_synced_repo, module_location
+    module_entitlement_manifest_org,
+    rhel_contenthost,
+    target_sat,
```
vijaysawant, line 96: `module_entitlement_manifest_org`, `module_activation_key` are using module scope satellite
```suggestion
    module_target_sat,
```

tests/foreman/cli/test_registration.py
```diff
@@ -62,14 +63,14 @@ def test_host_registration_end_to_end(
...
+            'organization-ids': org.id,
             'location-ids': module_location.id,
         }
     )
     result = rhel_contenthost.register(
-        module_org,
+        org,
         module_location,
-        module_ak_with_synced_repo.name,
+        [module_activation_key.name],
```
vijaysawant, line 73: is it required to pass in list form?

vijaysawant (commented):
General comments related to fixture and one query from `rhel_contenthost.register(...)`

pytest_fixtures/component/activationkey.py
```diff
@@ -6,8 +6,14 @@
 
 
 @pytest.fixture(scope='module')
-def module_activation_key(module_org, module_target_sat):
-    return module_target_sat.api.ActivationKey(organization=module_org).create()
+def module_activation_key(module_entitlement_manifest_org, module_target_sat):
```
rmynar, line 9: There's a fixture with same name and similar content in here: https://github.com/SatelliteQE/robottelo/blob/5405d12dfb0e66ec879707b1af85c11bb9afc901/tests/foreman/ui/test_host.py#L101
Can we have just one fixture for both?

shweta83: @rmynar The one in test_host.py has more steps which is specific for the module.
For `SCA mode manifest activation key`, I don't see it being used at many places. We can add one if needed.

tests/foreman/cli/test_registration.py
shweta83: Yes. Since it accepts multiple values so this needs to be passed in list.

tests/foreman/api/test_registration.py
shweta83: Since this test changes uuid so I wanted to keep it function scoped. I am not entirely sure but I guess the one used in test function will override the other one.

shweta83:
trigger: test-robottelo
pytest: tests/foreman/api/test_registration.py

pytest_fixtures/component/activationkey.py
vsedmik, line 9:
```suggestion
def module_activation_key(module_sca_manifest_org, module_target_sat):
```

tests/foreman/api/test_registration.py
```diff
@@ -28,9 +28,9 @@
 @pytest.mark.e2e
 @pytest.mark.no_containers
 def test_host_registration_end_to_end(
-    module_org,
+    module_entitlement_manifest_org,
```
vsedmik, line 31:
```suggestion
    module_sca_manifest_org,
```

tests/foreman/api/test_registration.py
```diff
@@ -90,7 +91,11 @@ def test_host_registration_end_to_end(
 @pytest.mark.tier3
 @pytest.mark.rhel_ver_match('[^6]')
 def test_positive_allow_reregistration_when_dmi_uuid_changed(
-    module_org, rhel_contenthost, target_sat, module_ak_with_synced_repo, module_location
+    module_entitlement_manifest_org,
```
vsedmik, line 94:
```suggestion
    module_sca_manifest_org,
```

tests/foreman/cli/test_registration.py
```diff
@@ -27,9 +27,9 @@
 @pytest.mark.e2e
 @pytest.mark.no_containers
 def test_host_registration_end_to_end(
-    module_org,
+    module_entitlement_manifest_org,
```
vsedmik, line 30:
```suggestion
    module_sca_manifest_org,
```

vsedmik (changes requested):
I would encourage to use the SCA fixture instead, unless you test the entitlement subscriptions specifically, which doesn't seem to be this case.

The reason is that in 6.16 the non-SCA orgs will be removed and all tests using the entitlement orgs/manifests (yes, there are many) will need to be migrated to SCA. If you use it now, you won't need to fix it later. Talking about ~3 months horizon max and not sooner than 6.14 branching (next week).

pytest_fixtures/component/activationkey.py
```diff
@@ -6,8 +6,14 @@
...
+    activation_key = module_target_sat.api.ActivationKey(
+        auto_attach=True,
```
vsedmik, line 11: `auto_attach` shouldn't be needed in case of SCA

shweta83:
trigger: test-robottelo
pytest: tests/foreman/api/test_registration.py

tests/foreman/api/test_registration.py
shweta83: @vijaysawant Are you good with this? We can resolve this conversation.

pytest_fixtures/component/activationkey.py
shweta83: I updated the fixture to use sca enabled manifest.

pytest_fixtures/component/activationkey.py
shweta83: Removed.

shweta83:
> @shweta83 can you revisit this PR?

Updated the PR.

tests/foreman/api/test_registration.py
rplevka: @shweta83 I agree that this test should use its own satellite since it is tweaking it.

rplevka (approved):

Gauravtalreja1 (changes requested):
ACK for the changes
@shweta83 could you fix the wrong commit message linked for this PR? and update the PR with some meaningful commit message/PR title to change s/entilement_manifest/sca_manifest change or something like that, instead of generic message of fixing registration test

pytest_fixtures/component/activationkey.py
vijaysawant: I agree with your point @vsedmik that `6.16` onwards only SCA based orgs will be used and we need to use organization fixture accordingly which supports this feature. 
However this suggested changes is applicable in registration scenario of `6.12.z`, `6.13.z`, `6.14.z` and there may be customers who uses **SCA disbale org** & entitlement mode manifest while registration, so I wouldn't suggest to use of `SCA mode manifest org` and would ask to make separate change for later versions, 6.15.z and stream.
@shweta83

Gauravtalreja1:
@shweta83 Could you update the commit message same as the PR title? :)

shweta83:
> ACK for the changes @shweta83 could you fix the wrong commit message linked for this PR? and update the PR with some meaningful commit message/PR title to change s/entilement_manifest/sca_manifest change or something like that, instead of generic message of fixing registration test

Updated.

shweta83:
> @shweta83 Could you update the commit message same as the PR title? :)

Updated.

shweta83:
trigger: test-robottelo
pytest: tests/foreman/api/test_registration.py -k test_positive_allow_reregistration_when_dmi_uuid_changed

pytest_fixtures/component/activationkey.py
shweta83: @vijaysawant We have option to enable SCA in older versions, but in new newer versions, it will be enabled by default. There are many test cases where the manifest with SCA disabled are used so we are good from coverage pov.

Gauravtalreja1 (approved):

pytest_fixtures/component/activationkey.py
vijaysawant: Use of `*_sca_manifest_org` over `*_entitlement_manifest_org` would have been consider by looking at `e2e` marker and label `6.12.z`, `6.13.z` & `6.14.z`
If we have another tests which are covering SCA-disable org scenario for `6.12.z`, `6.13.z` & `6.14.z` Satellite versions then we should be good with this PR change.

ACK

PR #13083, opened 2023-11-14
Add test for SCA subscription report

vsedmik:
### Problem Statement
Satellite 6.15 implemented new report template for SCA-enabled orgs. We need to verify the report is rendered correctly.

### Solution
This PR adds a test which generates the template and checks it contains the correct values.

### DIFF
--- a/tests/foreman/api/test_reporttemplates.py
+++ b/tests/foreman/api/test_reporttemplates.py
@@ -16,6 +16,8 @@
 
 :Upstream: No
 """
+import re
+
 from broker import Broker
 from fauxfactory import gen_string
 import pytest
@@ -24,6 +26,7 @@ from wait_for import wait_for
 
 from robottelo.config import settings
 from robottelo.constants import (
+    DEFAULT_ARCHITECTURE,
     DEFAULT_SUBSCRIPTION_NAME,
     FAKE_1_CUSTOM_PACKAGE,
     FAKE_1_CUSTOM_PACKAGE_NAME,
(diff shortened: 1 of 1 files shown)


vsedmik:
trigger: test-robottelo
pytest: tests/foreman/api/test_reporttemplates.py -k test_positive_installed_products

tests/foreman/api/test_reporttemplates.py
```diff
@@ -836,3 +839,94 @@ def test_positive_installable_errata(
...
+        'basearch': DEFAULT_ARCHITECTURE,
+        'product': REPOS['rhae2']['product'],
+        'name': REPOS['rhae2']['name'],
+        'reposet': REPOS['rhae2']['reposet'],
+        'releasever': None,
+    }
+    repo_id = target_sat.api_factory.enable_sync_redhat_repo(rh_repo, org.id)
+    cv = target_sat.api_factory.cv_publish_promote(cv_name, lce_name, repo_id, org.id)
+    ak = target_sat.api.ActivationKey(
+        content_view=cv, organization=org, environment=cv.environment[-1]
```
vijaysawant, line 893: Just an query.. 
What if we pass `environment=lce_name` directly instead of `environment=cv.environment[-1]`?

As `CV` has been published & `CVV` has been promoted to respective `LCE`, I think for `AK` creation getting `LCE` details **explicitly** is not needed (like this way `cv.environment[-1]`) unless and untill we are going to use previously published/promoted `CV`/`CVV`

tests/foreman/api/test_reporttemplates.py
```diff
@@ -836,3 +839,94 @@ def test_positive_installable_errata(
...
+    )
+    assert report, 'No report generated.'
+    assert report[0]['Host Name'] == rhel_contenthost.hostname, 'Incorrect host was reported.'
+    assert report[0]['Organization'] == org.name, 'Incorrect org was reported.'
+    assert report[0]['Lifecycle Environment'] == lce_name, 'Incorrect LCE was reported.'
+    assert report[0]['Content View'] == cv_name, 'Incorrect CV was reported.'
+    assert report[0]['Role'] == sys_tags['role'], 'Incorrect role was reported.'
+    assert report[0]['Usage'] == sys_tags['usage'], 'Incorrect usage was reported.'
+
+    # Get the installed products via rake and compare them with report
```
vijaysawant, line 920: non-blocking
```suggestion
    # Get the installed products via rake and compare them with generated report
```

tests/foreman/api/test_reporttemplates.py
```diff
@@ -836,3 +839,94 @@ def test_positive_installable_errata(
...
+    rake = target_sat.execute(
+        f'echo "Host.find_by(name: \'{rhel_contenthost.hostname}\').'
+        'subscription_facet.installed_products" | foreman-rake console'
+    )
+    assert rake.status == 0, f'Rake call failed with this output:\n({rake.stdout}).'
+
+    pattern = re.compile(r'name: "(.*?)".*?cp_product_id: "(.*?)"')
+    matches = pattern.findall(rake.stdout)
+    products = [f"{match[0]} ({match[1]})" for match in matches]
```
vijaysawant, line 929: Just to maintain consistency
```suggestion
    products = [f'{match[0]} ({match[1]})' for match in matches]
```

vijaysawant (commented):
query at `ActivationKey(...).create()` with additional non-blocking suggestions

tests/foreman/api/test_reporttemplates.py
vsedmik: In that case it will fail to create the AK with `Received HTTP 404 response: {"displayMessage":"Couldn't find environment 'Abc'","errors":["Couldn't find environment 'Abc'"]}`

I think we expect a `LifecycleEnvironment` object or it's id [here](https://github.com/SatelliteQE/nailgun/blob/cf4ee07b826b4f4cba0c0ce5c91c664f0a009bd7/nailgun/entities.py#L215)

vijaysawant: Okay, thanks for detailing the issue, this is something different error.

vijaysawant:
looks good to me.. :+1:

pondrejk (approved):

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/api/test_reporttemplates.py -k test_positive_installed_products

PR #13107, opened 2023-11-16
use one fixture instead of multiple and remove problematic one

vijaysawant:
### Problem Statement
_Hosts-Content component -> CLI tests_ were failing with  `AttributeError` because of fixture `module_gt_manifest_org` which is being used inside moduled scoped fixture `rh_repo_gt_manifest` of [repository.py](https://github.com/SatelliteQE/robottelo/blob/master/pytest_fixtures/component/repository.py#L36)

> 	`failed on setup with "AttributeError: '_io.BytesIO' object has no attribute 'filename'"`

The fixture `rh_repo_gt_manifest` simply creates new org with GT menifest, create rhst repo, sync and return the respective repo. Rest of fixtures from test module uses `module_sca_manifest_org` which creates module org with menifest.
The above setup were problematic.

### Solution
Collect scattered fixtures into single module scoped fixture and use it.

### Related Issues
No

PR #13123, opened 2023-11-17
[6.13.z] use one fixture instead of multiple and remove problematic one

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/13107

### Problem Statement
_Hosts-Content component -> CLI tests_ were failing with  `AttributeError` because of fixture `module_gt_manifest_org` which is being used inside moduled scoped fixture `rh_repo_gt_manifest` of [repository.py](https://github.com/SatelliteQE/robottelo/blob/master/pytest_fixtures/component/repository.py#L36)

> 	`failed on setup with "AttributeError: '_io.BytesIO' object has no attribute 'filename'"`

The fixture `rh_repo_gt_manifest` simply creates new org with GT menifest, create rhst repo, sync and return the respective repo. Rest of fixtures from test module uses `module_sca_manifest_org` which creates module org with menifest.
The above setup were problematic.

### Solution
Collect scattered fixtures into single module scoped fixture and use it.

### Related Issues
No

### DIFF
--- a/pytest_fixtures/component/repository.py
+++ b/pytest_fixtures/component/repository.py
@@ -33,24 +33,6 @@ def module_product(module_org, module_target_sat):
 
 
 @pytest.fixture(scope='module')
-def rh_repo_gt_manifest(module_gt_manifest_org, module_target_sat):
-    """Use GT manifest org, creates RH tools repo, syncs and returns RH repo."""
-    # enable rhel repo and return its ID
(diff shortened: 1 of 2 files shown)


damoore044:
trigger: test-robottelo
pytest: tests/foreman/cli/test_contentaccess.py -k 'test_positive_list_installable_updates or test_positive_erratum_installable or test_negative_unregister_and_pull_content'

PR #13162, opened 2023-11-22
[RFE] Check RHEL Lifecycle status in ContentHost UI

damoore044:
Extend UI::test_contenthost.py::`test_positive_end_to_end`: 
- To check the new `'RHEL Lifecycle status'` of the rhel host, and check `host status`.

Two helper methods:
- `get_supported_rhel_versions()` : to get all supported rhel versions for contenthosts in satellite.
- `get_rhel_lifecycle_support(rhel_version)` : to get expected RHEL Lifecycle status string, 
based on provided `rhel_version#` vs all supported `versions#` discovered.

### DIFF
--- a/tests/foreman/ui/test_contenthost.py
+++ b/tests/foreman/ui/test_contenthost.py
@@ -102,8 +102,38 @@ def run_remote_command_on_content_host(command, vm_module_streams):
     return result
 
 
+def get_supported_rhel_versions():
+    """Helper to get the supported base rhel versions for contenthost.
+    return: a list of integers
+    """
+    return [
+        ver for ver in settings.supportability.content_hosts.rhel.versions if isinstance(ver, int)
+    ]
+
+
+def get_rhel_lifecycle_support(rhel_version):
+    """Helper to get what the Lifecycle Support Status should be,
+       based on provided rhel version.
+    :param rhel_version: integer of the current base rhel version
+    :return: string with the expected status of rhel version support
+    """
+    rhels = get_supported_rhel_versions()
+    rhel_lifecycle_status = 'Unknown'
+    if rhel_version not in rhels:
+        return rhel_lifecycle_status
+    elif (len(rhels) - 1) - rhels.index(rhel_version) <= 1:
+        rhel_lifecycle_status = 'Full support'
+    elif (len(rhels) - 1) - rhels.index(rhel_version) == 2:
+        rhel_lifecycle_status = 'Approaching end of maintenance support'
+    elif (len(rhels) - 1) - rhels.index(rhel_version) >= 3:
+        rhel_lifecycle_status = 'End of maintenance support'
+
+    return rhel_lifecycle_status
+
+
 @pytest.mark.e2e
 @pytest.mark.tier3
+@pytest.mark.parametrize('setting_update', ['new_hosts_page'], indirect=True)
 @pytest.mark.parametrize(
     'module_repos_collection_with_manifest',
     [
@@ -120,7 +150,14 @@ def run_remote_command_on_content_host(command, vm_module_streams):
     indirect=True,
 )
 @pytest.mark.no_containers
-def test_positive_end_to_end(session, default_location, module_repos_collection_with_manifest, vm):
+def test_positive_end_to_end(
+    vm,
+    session,
+    module_org,
+    setting_update,
+    default_location,
+    module_repos_collection_with_manifest,
+):
     """Create all entities required for content host, set up host, register it
     as a content host, read content host details, install package and errata.
 
@@ -135,11 +172,27 @@ def test_positive_end_to_end(session, default_location, module_repos_collection_
 
     :CaseImportance: Critical
     """
+    property_name = setting_update.name
+    _distro = module_repos_collection_with_manifest.distro
+    host_rhel_version = None
+    if _distro.startswith('rhel'):
+        host_rhel_version = int(_distro[4:])
+    rhel_status = get_rhel_lifecycle_support(host_rhel_version if not None else 0)
+
     result = vm.run(f'yum -y install {FAKE_1_CUSTOM_PACKAGE}')
     assert result.status == 0
     startdate = datetime.utcnow().strftime('%m/%d/%Y')
     with session:
+        session.settings.update(f'name = {property_name}', 'Yes')
         session.location.select(default_location.name)
+        session.organization.select(module_org.name)
+        # Use new hosts UI, only check the details and host status
+        host_details = session.host.get_details(vm.hostname, widget_names='properties')
+        host_status = session.host_new.get_host_statuses(vm.hostname)
+        # Use old hosts UI for remainder
+        session.settings.update(f'name = {property_name}', 'No')
+        assert rhel_status in host_status['RHEL lifecycle']['Status']
+        assert rhel_status in host_details['properties']['properties_table']['RHEL lifecycle']
         # Ensure content host is searchable
         assert session.contenthost.search(vm.hostname)[0]['Name'] == vm.hostname
         chost = session.contenthost.read(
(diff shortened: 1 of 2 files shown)


damoore044:
trigger: test-robottelo
pytest: tests/foreman/ui/test_contenthost.py -k test_positive_end_to_end

damoore044:
trigger: test-robottelo
pytest: tests/foreman/ui/test_contenthost.py -k test_positive_end_to_end

damoore044:
trigger: test-robottelo
pytest: tests/foreman/ui/test_contenthost.py -k test_positive_end_to_end

tests/foreman/ui/test_contenthost.py
```diff
@@ -135,11 +137,30 @@ def test_positive_end_to_end(session, default_location, module_repos_collection_
...
+        rhel_lifecycle_status = 'Full support'
+
     result = vm.run(f'yum -y install {FAKE_1_CUSTOM_PACKAGE}')
     assert result.status == 0
     startdate = datetime.utcnow().strftime('%m/%d/%Y')
     with session:
         session.location.select(default_location.name)
+        session.organization.select(module_org.name)
+        # Use new host UI, check the status and RHEL Lifecycle in details
+        session.settings.update(f'name = {"new_hosts_page"}', 'Yes')
```
vsedmik, line 154: I might rather use the `settings_update` fixture instead. You don't need to care about setting the value back to whatever it was before - the fixture does it for you after the test finishes.
Also, now if any of the asserts between the updates fail, your target_sat would stay switched to `new_hosts_page: Yes`, which may cause the tests executed after this one to fail.

tests/foreman/ui/test_contenthost.py
```diff
@@ -135,11 +137,30 @@ def test_positive_end_to_end(session, default_location, module_repos_collection_
 
     :CaseImportance: Critical
     """
+    rhel_lifecycle_status = 'Unknown'
+    _distro = module_repos_collection_with_manifest.distro
+    if _distro == 'rhel7':
+        rhel_lifecycle_status = 'Approaching end of maintenance support'
+    elif _distro == 'rhel8':
+        rhel_lifecycle_status = 'Full support'
```
vsedmik, line 145: Do we have another way to check the current support status so that we don't need to hardcode it?

damoore044: will look into this soon once I can get passing

damoore044:
trigger: test-robottelo
pytest: tests/foreman/ui/test_contenthost.py -k test_positive_end_to_end

damoore044:
trigger: test-robottelo
pytest: tests/foreman/ui/test_contenthost.py -k test_positive_end_to_end

tests/foreman/ui/test_contenthost.py
damoore044: Would you say the two new helper function address this @vsedmik ?

damoore044:
trigger: test-robottelo

tests/foreman/ui/test_contenthost.py
```diff
@@ -135,11 +172,28 @@ def test_positive_end_to_end(session, default_location, module_repos_collection_
...
         session.location.select(default_location.name)
+        session.organization.select(module_org.name)
+        # Use new hosts UI, only check the details and host status
+        host_details = session.host.get_details(vm.hostname, widget_names='properties')
+        host_status = session.host_new.get_host_statuses(vm.hostname)
+        # Use old hosts UI for remainder
+        session.settings.update(f'name = {property_name}', 'No')
+        assert rhel_status in host_status['RHEL lifecycle']['Status']
+        assert rhel_status in host_details['properties']['properties_table']['RHEL lifecycle']
+        # TODO: Use UI in old 'All Hosts' page, add 'RHEL Lifecycle status' column, assert the status.
```
sambible, line 196: I'd either do this as part of this PR, or not do it - don't leave a TODO comment like this in the test, unless you have a really good reason.

sambible (changes requested):

tests/foreman/ui/test_contenthost.py
damoore044: Looks like we have coverage for all hosts page, adding columns, in
`tests/foreman/ui/test_host/test_positive_manage_table_columns` 
and some other host tests. I will extend the new column in a separate pr!

tests/foreman/ui/test_contenthost.py
vsedmik: Not sure if I read it correctly, this way we grant last two (numerical) entries in the yaml as `Full support`, and entries left lower and lower support level?
Proposal: Let's not depend on the position (though I don't expect it would change to anything like `[8,9,7,6]`, rendering 6 and 7 as `Full support`) but rather it's value like this.
```
        rhels = get_supported_rhel_versions().sort(reverse=True)
        rhel_lifecycle_status = 'Unknown'
        if rhel_version not in rhels:
            return rhel_lifecycle_status
        elif rhels.index(rhel_version) <= 1:
            rhel_lifecycle_status = 'Full support'
        elif rhels.index(rhel_version) == 2:
            rhel_lifecycle_status = 'Approaching end of maintenance support'
        elif rhels.index(rhel_version) >= 3:
            rhel_lifecycle_status = 'End of maintenance support'
        return rhel_lifecycle_status
```

tests/foreman/ui/test_contenthost.py
```diff
@@ -120,7 +150,14 @@ def run_remote_command_on_content_host(command, vm_module_streams):
...
 @pytest.mark.no_containers
-def test_positive_end_to_end(session, default_location, module_repos_collection_with_manifest, vm):
+def test_positive_end_to_end(
+    vm,
+    session,
+    module_org,
+    setting_update,
+    default_location,
+    module_repos_collection_with_manifest,
+):
```
lhellebr, line 160: Can you please:
1) Make the first paragraph of a docstring more high-level, describe in one simple sentence what the test tests, and
2) Add `:steps:` where you describe it more in detail?

tests/foreman/ui/test_contenthost.py
```diff
@@ -135,11 +172,27 @@ def test_positive_end_to_end(session, default_location, module_repos_collection_
...
+    host_rhel_version = None
+    if _distro.startswith('rhel'):
+        host_rhel_version = int(_distro[4:])
+    rhel_status = get_rhel_lifecycle_support(host_rhel_version if not None else 0)
+
     result = vm.run(f'yum -y install {FAKE_1_CUSTOM_PACKAGE}')
     assert result.status == 0
     startdate = datetime.utcnow().strftime('%m/%d/%Y')
     with session:
+        session.settings.update(f'name = {property_name}', 'Yes')
```
lhellebr, line 186: Is this necessary? `new_hosts_page` should already be set to `True` by fixture `setting_update`, shouldn't it?

tests/foreman/ui/test_contenthost.py
```diff
@@ -135,11 +172,27 @@ def test_positive_end_to_end(session, default_location, module_repos_collection_
...
         session.location.select(default_location.name)
+        session.organization.select(module_org.name)
+        # Use new hosts UI, only check the details and host status
+        host_details = session.host.get_details(vm.hostname, widget_names='properties')
+        host_status = session.host_new.get_host_statuses(vm.hostname)
+        # Use old hosts UI for remainder
```
lhellebr, line 192: Why do you need to use both new UI and old UI? Is it because you can't do the whole workflow in the new one? If that's the case, I think it's a major UX bug.

If it's because lack of support in Airgun, I think it would be better to either use only old UI, or (preferably) add the support to Airgun.

damoore044: Can't do the whole workflow in new UI yet, as the contenthost parts fails navigation with new UI, I believe there will need to be some airgun fixes, not completely sure why it won't pass with new UI setting enabled through the entire test.
This test is a bit older written for old UI, so I think for an extension we should keep this on the old UI?
We could plan to update this to use nav in `host_new.py` in the future

damoore044:
trigger: test-robottelo

damoore044:
trigger: test-robottelo

damoore044:
trigger: test-robottelo

damoore044:
trigger: test-robottelo

damoore044:
trigger: test-robottelo

vsedmik (approved):

PR #13168, opened 2023-11-23
Changes for 6.15.z new branch

jyejare:
### Problem Statement
New 6.15.z branch 

### Solution
- Dependabot labels are updated for new branch
- Removed dispatch release GHA from 6.15.z as we are releasing only master changes
- Airgun and Nailgun Requirements uses 6.15.z branch
- Constants are using new version now
- Stream tests removed
- Setup.py uses new version

### Dependent Issues
- Nailgun: https://github.com/SatelliteQE/nailgun/pull/1033
- Airgun: https://github.com/SatelliteQE/airgun/pull/1044

### Related Issues
- Master PR: https://github.com/SatelliteQE/robottelo/pull/13169


lpramuk (changes requested):
Please change also branch (master -> 6.15.z) for GHA

lpramuk:
> Please change also branch (master -> 6.15.z) for GHA

Oh it's there already 
https://github.com/SatelliteQE/robottelo/blob/b46a6958776ac649576d97f74c60f97303de6919/.github/workflows/update_robottelo_image.yml#L7-L8

omkarkhatavkar (approved):

jyejare:
@omkarkhatavkar Yeah will fix that, some SSL issues locally in my setup for running precommit

jyejare:
@omkarkhatavkar CI checks are passing now !

robottelo/constants/__init__.py
```diff
@@ -18,7 +18,7 @@ class Colored(Box):
 # This should be updated after each version branch
 SATELLITE_VERSION = "6.15"
 SATELLITE_OS_VERSION = "8"
-SAT_NON_GA_VERSIONS = ['6.14', '6.15']
+SAT_NON_GA_VERSIONS = ['6.15', '6.16']
```
Gauravtalreja1, line 21: This is used mainatain sat and capsule repo setup tests, it contains a check for non-GA version repo check, keeping it won't harm but it will be unnecessary check for 6.15.z, right? @jameerpathan111 @Griffin-Sullivan  
```suggestion
SAT_NON_GA_VERSIONS = ['6.15']
```

.github/workflows/dispatch_release.yml
```diff
@@ -1,31 +0,0 @@
-### The auto release workflow triggered through dispatch request from CI
-name: auto-release
-
```
Gauravtalreja1, line 3: Do we not need releases for zstream?

robottelo/constants/__init__.py
jyejare: @jameerpathan111 @Griffin-Sullivan @Gauravtalreja1 There is condition based on this constant in `test_advanced.py` even that has to be removed and cherrypicked ?

Should I just remove this change from this PR and later someone will raise a followup PR for it ?

.github/workflows/dispatch_release.yml
jyejare: We dont do releases for zStream only master since we are mostly interested in Framework changes and they are being done in master and cherrypicked so does not make sense to have zReleases !

robottelo/constants/__init__.py
Gauravtalreja1: I think just removing 6.16 from constants should be fine

Gauravtalreja1 (approved):

PR #13169, opened 2023-11-23
Changes to Master due to new 6.15 and 6.16

jyejare:
### Problem Statement
New 6.15 downstream and master points to stream that is 6.16

### Solution
- Dependabot.yaml cherrypicks to 6.15
- Robottelo conf and constants now uses 6.16 and 6.15 satellite versions 

### Related Issues
https://github.com/SatelliteQE/robottelo/pull/13168

### DIFF
--- a/conf/robottelo.yaml.template
+++ b/conf/robottelo.yaml.template
@@ -15,9 +15,9 @@ ROBOTTELO:
   RUN_ONE_DATAPOINT: false
   # Satellite version supported by this branch
   # UNDR version is used for some URL composition
-  SATELLITE_VERSION: "6.15"
+  SATELLITE_VERSION: "6.16"
   # The Base OS RHEL Version(x.y) where the satellite would be installed
-  RHEL_VERSION: "7.9"
+  RHEL_VERSION: "8.9"
   # Dynaconf and Dynaconf hooks related options
   SETTINGS:
     GET_FRESH: true
--- a/robottelo/constants/__init__.py
+++ b/robottelo/constants/__init__.py
@@ -16,9 +16,9 @@ class Colored(Box):
 
 
 # This should be updated after each version branch
-SATELLITE_VERSION = "6.15"
+SATELLITE_VERSION = "6.16"
 SATELLITE_OS_VERSION = "8"
-SAT_NON_GA_VERSIONS = ['6.14', '6.15']
+SAT_NON_GA_VERSIONS = ['6.15', '6.16']
 
 # Default system ports
 HTTPS_PORT = '443'
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,7 @@ updates:
     labels:
       - "CherryPick"
       - "dependencies"
+      - "6.15.z"
       - "6.14.z"
       - "6.13.z"
       - "6.12.z"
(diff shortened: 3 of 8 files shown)


robottelo/constants/__init__.py
```diff
@@ -16,9 +16,9 @@ class Colored(Box):
 
 
 # This should be updated after each version branch
-SATELLITE_VERSION = "6.15"
+SATELLITE_VERSION = "6.16"
 SATELLITE_OS_VERSION = "8"
```
lpramuk, line 20: Or "9"  right ? :wink:

jyejare: Not yet confirmed I believe, anyways OS Version would be provided by CI (either 8 or 9) for executing tests in CI! This is been managed by conf files!

If it supports on RHEL 9 , we can switch this drop down version to 9.

omkarkhatavkar (approved):

JacobCallahan (changes requested):
Please remove any current stream markers

jyejare:
@JacobCallahan I did not because I would like Teams to decide and remove the markers on the tests that they think are not applicable for new stream version.

There is possibility that the feature is divided into two parts and hence the test/feature is still applicable with new stream version.

conf/robottelo.yaml.template
```diff
@@ -15,7 +15,7 @@ ROBOTTELO:
   RUN_ONE_DATAPOINT: false
   # Satellite version supported by this branch
   # UNDR version is used for some URL composition
-  SATELLITE_VERSION: "6.15"
+  SATELLITE_VERSION: "6.16"
   # The Base OS RHEL Version(x.y) where the satellite would be installed
   RHEL_VERSION: "7.9"
```
Gauravtalreja1, line 20:
```suggestion
  RHEL_VERSION: "8.9"
```

robottelo/constants/__init__.py
Gauravtalreja1: Could we use settings.robottelo.rhel_version here provided by CI instead of this constant, so this can be managed by configs for each job for 8/9?

jyejare: @Gauravtalreja1 Upon looking the usage of this variable it is being used in deciding(returning) capsule repo in `SatelliteCapsuleRepository` class of `repository_mixins.py` module.

We should deprecate the use of this variable and switch the url def to use rhel_version  conf settings.

jyejare: I dont block this PR on it as that change has to be cherrypicked back to all zBranches but this PR is not for cherrypicking

jyejare:
@JacobCallahan @Gauravtalreja1 Comments addressed !

JacobCallahan (approved):

JacobCallahan:
@Gauravtalreja1 please merge if you're satisfied with the changes.

robottelo/constants/__init__.py
Gauravtalreja1: agree @jyejare , we could start with master for this deprecation or else you could handle this in different PR which we can cherrypick to older branch, but it's no blocker

Gauravtalreja1 (approved):

PR #13177, opened 2023-11-23
Create a distinction between the product-ready OS and the product WFs

ogajduse:
### Problem Statement
There is no distinction between content host deployment workflows and workflows to deploy the operating system on which the product is supposed to be installed. That causes issues in the PIT server testing scenario.
We would like to add a test to install Satellite and Capsule to the test collection, but we need to have a separate content host deployment workflow and the `sat_ready_rhel` and `cap_ready_rhel` workflows. That is due to the need to install the product on top of a non-released OS that we deploy using a workflow that is different from the one that we use for the content host deployment.

### Solution
Create a distinction between the two workflows in the configuration and use the configuration.

### Related Issues
satelliteqe/satellite-jenkins!1173


conf/capsule.yaml.template
```diff
@@ -10,6 +10,8 @@ CAPSULE:
     # The base os rhel version where the capsule installed
     # RHEL_VERSION:
   # The Ansible Tower workflow used to deploy a capsule
-  DEPLOY_WORKFLOW: deploy-capsule
+  DEPLOY_WORKFLOW:
+    PRODUCT: deploy-satellite  # workflow to deploy OS with product running on top of it
```
rplevka, line 14:
```suggestion
    PRODUCT: deploy-capsule  # workflow to deploy OS with product running on top of it
```

robottelo/config/validators.py
```diff
@@ -23,7 +23,9 @@
         ),
         Validator('server.admin_password', default='changeme'),
         Validator('server.admin_username', default='admin'),
-        Validator('server.deploy_workflow', must_exist=True),
+        Validator('server.deploy_workflow', must_exist=True, is_type_of=dict),
```
rplevka, line 26: i am playing with an idea, that accepting the old format could be accepted too (for backward compatibility). in such scenario, the value would be automatically treated as `PRODUCT` workflow.

conf/capsule.yaml.template
ogajduse: fixed

robottelo/config/validators.py
ogajduse: Just a note: we can use Dynaconf hooks for that. https://www.dynaconf.com/advanced/?h=hook#hooks-for-the-solution

conf/capsule.yaml.template
```diff
@@ -10,6 +10,8 @@ CAPSULE:
     # The base os rhel version where the capsule installed
     # RHEL_VERSION:
   # The Ansible Tower workflow used to deploy a capsule
-  DEPLOY_WORKFLOW: deploy-capsule
+  DEPLOY_WORKFLOW:
```
JacobCallahan, line 13: Now that this is a parent of multiple workflows, I think this reads better.
This suggestion also applies to the server config
```suggestion
  DEPLOY_WORKFLOWS:
```

rmynar:
+1 to rename DEPLOY_WORKFLOW to DEPLOY_WORKFLOWS.

This solution also seems to be suitable for FIPS testing.

rmynar (approved):

pytest_fixtures/core/sat_cap_factory.py
```diff
@@ -245,12 +245,13 @@ def parametrized_enrolled_sat(
 
 
 def get_deploy_args(request):
+    '''Get deploy arguments for Satellite base OS deployment. Should not be used for Capsule.'''
```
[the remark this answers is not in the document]
ogajduse: Fixed

conf/capsule.yaml.template
ogajduse: Thanks for the suggestion. Applied.

ogajduse:
Config migration(s) were introduced in the latest changes. This mechanism is pretty simple and smoothes the transition from the old configuration.
I would like to hear your feedback on it.

robottelo/config/validators.py
ogajduse: @rplevka I have introduced config migration. Please see the latest changes.

conf/migrations.py
```diff
@@ -0,0 +1,33 @@
...
+                    f'rhel{Version(str(sat_rhel_version)).major}'
+                ].vm.workflow,
+            }
+            logger.info(
+                f'Migrated {product_type}.DEPLOY_WORKFLOW to {product_type}.DEPLOY_WORKFLOWS'
+            )
+        else:
+            logger.info(
+                f'No migration needed for {product_type}.DEPLOY_WORKFLOW to {product_type}.DEPLOY_WORKFLOWS. DEPLOY_WORKFLOWS is already set.'
+            )
```
JacobCallahan, line 33: If no migration is needed, I'd rather it be silent.

conf/dynaconf_hooks.py
```diff
@@ -33,7 +38,22 @@ def write_cache(path, data):
...
+    from conf import migrations
+
+    migration_functions = [
+        mf for mf in getmembers(migrations, isfunction) if mf[0].startswith('migration_')
+    ]
+    for name, func in migration_functions:
+        logger.debug(f'Running {name}')
+        func(settings, data)
+        logger.debug(f'Finished running {name}')
+    logger.info('Finished running config migration hooks')
```
JacobCallahan, line 56: can you add some comments here explaining what this is for and where to define the migrations?

JacobCallahan (changes requested):
neat solution with the migration, so just a couple of minor things!

conf/migrations.py
ogajduse: removed

JacobCallahan (approved):

PR #13187, opened 2023-11-24
Add coverage for BZ#2092039 and update few docstrings

vsedmik:
### Problem Statement
Coverage for [BZ#2092039](https://bugzilla.redhat.com/show_bug.cgi?id=2092039) is missing.

### Solution
This PR. Also adds few docstring fixes.

### DIFF
--- a/tests/foreman/cli/test_satellitesync.py
+++ b/tests/foreman/cli/test_satellitesync.py
@@ -579,8 +579,6 @@ class TestContentViewSync:
             1. CV version custom contents has been exported to directory.
             2. All The exported custom contents has been imported in org/satellite.
 
-        :CaseImportance: High
-
         :CaseLevel: System
 
         :BZ: 1832858
@@ -772,8 +770,6 @@ class TestContentViewSync:
             1. Filtered CV version custom contents has been exported to directory
             2. Filtered exported custom contents has been imported in org/satellite
 
-        :CaseImportance: High
-
         :CaseLevel: System
         """
         exporting_cv_name = importing_cvv = gen_string('alpha')
@@ -952,8 +948,6 @@ class TestContentViewSync:
         :BZ: 1655239, 2040870
 
         :customerscenario: true
-
-        :CaseImportance: High
 
         :CaseLevel: System
         """
@@ -1184,13 +1178,10 @@ class TestContentViewSync:
         :id: 4cc69666-407f-4d66-b3d2-8fe2ed135a5f
 
         :steps:
-
-            1. Import a cv with a path that doesn't exist
+            1. Import a CV with a path that doesn't exist.
 
         :expectedresults:
-
-            1. Error 'Unable to sync repositories, no library repository found' should be
-                displayed
+            1. Error 'Unable to sync repositories, no library repository found' should be displayed.
         """
         export_folder = gen_string('alpha')
         import_path = f'{PULP_IMPORT_DIR}{export_folder}'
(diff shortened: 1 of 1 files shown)


synkd (approved):

tests/foreman/cli/test_satellitesync.py
```diff
@@ -2022,6 +2010,97 @@ def test_positive_export_import_incremental_yum_repo(
...
+        )
+        assert '1.0' in target_sat.validate_pulp_filepath(
+            function_sca_manifest_org, PULP_EXPORT_DIR
+        )
+
+        # Run import and verify the product and repo is created in the importing org
+        import_path = target_sat.move_pulp_archive(function_sca_manifest_org, export['message'])
+        target_sat.cli.ContentImport.repository(
+            {'organization-id': function_import_org_with_manifest.id, 'path': import_path},
+            timeout=300000,
```
Griffin-Sullivan, line 2065: Is this a 300000 second timeout? If so, I think we should turn it down.

Griffin-Sullivan (approved):
I appreciate all of the docstring updates. Makes it much cleaner. Just one question about the timeout for both imports. Otherwise this is ready to merge.

tests/foreman/cli/test_satellitesync.py
vsedmik: Should be in miliseconds I guess (at least from what I can see elsewhere grepping for `\.cli\..*timeout`)

Griffin-Sullivan: I followed the steps to here https://github.com/SatelliteQE/broker/blob/master/broker/hosts.py#L139. It's possible VS Code led me astray, so I'll double check.

Griffin-Sullivan: Might be good to tag @JacobCallahan here for confirmation.

tests/foreman/cli/test_satellitesync.py
```diff
@@ -2022,6 +2010,97 @@ def test_positive_export_import_incremental_yum_repo(
...
+        indirect=True,
+    )
+    def test_positive_export_import_mismatch_label(
+        self,
+        target_sat,
+        export_import_cleanup_function,
+        config_export_import_settings,
+        function_sca_manifest_org,
+        function_import_org_with_manifest,
+        function_synced_rhel_repo,
```
lhellebr, line 2026: Is it necessary to sync the whole RHEL repo?

lhellebr (commented):
LGTM pending my comment and the timeout uncertainity.

tests/foreman/cli/test_satellitesync.py
vsedmik: Well, this is rather bad naming. It should have been called `function_synced_rh_repo` - it does not sync the RHEL repo but the RH repo you provide in parameter, Red Hat Ansible Engine in this case. Which needs to be synced with `immediate` download policy specifically for export/import purpose.

tests/foreman/cli/test_satellitesync.py
JacobCallahan: @Griffin-Sullivan is correct, that timeout is in seconds. In fact, Broker also allows you to use more clear timeout formatting so instead of specifying a long timeout like that, you can just pass `timeout="5m"`
https://github.com/SatelliteQE/broker/blob/master/broker/session.py#L99
https://github.com/SatelliteQE/broker/blob/master/broker/helpers.py#L417

vsedmik: Ok, changed to the nicer format, let's wait for the PRT.

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/cli/test_satellitesync.py -k mismatch

lhellebr (approved):

PR #13253, opened 2023-12-04
eol version check

pondrejk:
### Problem Statement
This supplements tests for the EOL banner feature. The eol date for the version is defined in the satellite-lifecycle package (so that there is the same solution for connected and disconnected sats). This test aims to check the date in the package is correctly updated. 

### Solution
Comparing the date from the package with the date from the lifecycle api.

Keeping in draft state, as the EOL banner just got to 6.15, but not yet to older releases and the lifecycle api contains data only for released versions

### Related Issues

satellite-jenkins MR 1185

### DIFF
--- a/tests/foreman/api/test_eol_banner.py
+++ b/tests/foreman/api/test_eol_banner.py
@@ -1,0 +1,55 @@
+"""Test module for EOL tracking
+
+:Requirement: Dashboard
+
+:CaseAutomation: Automated
+
+:CaseComponent: Dashboard
+
+:Team: Endeavour
+
+:CaseImportance: High
+"""
+
+from datetime import datetime
+
+import pytest
+import requests
+import yaml
+
+from robottelo.config import settings
+from robottelo.constants import LIFECYCLE_METADATA_FILE
+from robottelo.logging import logger
+
+
+@pytest.mark.tier2
+def test_positive_check_eol_date(target_sat):
+    """Check if the EOL date for the satellite version
+
+    :id: 1c2f0d19-a357-4461-9ace-edb468f9ca5c
+
+    :expectedresults: EOL date from satellite-lifecycle package is accurate
+    """
+    current_version = '.'.join(target_sat.version.split('.')[0:2])
+    output = yaml.load(target_sat.execute(rf'cat {LIFECYCLE_METADATA_FILE}').stdout, yaml.Loader)
+    logger.debug(f'contents of {LIFECYCLE_METADATA_FILE} :{output}')
+    eol_datetime = datetime.strptime(output['releases'][current_version]['end_of_life'], '%Y-%m-%d')
+    result = requests.get(settings.subscription.lifecycle_api_url, verify=False)
+    if result.status_code != 200:
+        result.raise_for_status()
+    versions = result.json()['data'][0]['versions']
+    version = [v for v in versions if v['name'] == current_version]
+    if len(version) > 0:
+        api_date = [
+            (p['date_format'], p['date'])
+            for p in version[0]['phases']
+            if p['name'] == 'Maintenance support'
+        ]
+        if api_date[0][0] == 'string':
+            assert eol_datetime.strftime("%B, %Y") in api_date[0][1]
+        elif api_date[0][0] == 'date':
+            assert eol_datetime.strftime("%Y-%m-%dT%H:%M:%S.%f")[:-3] + 'Z' == api_date[0][1]
+        else:
+            pytest.fail("Unexpcted date format returned")
+    else:
+        pytest.skip("The Satellite version is not GA yet")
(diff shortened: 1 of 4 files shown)


omkarkhatavkar:
@pondrejk can you please work on the review comments as the PR is getting stale status

Satellite-QE:
**PRT Result**
```
Build Number: 6900
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/api/test_eol_banner.py --external-logging
Test Result : ================== 1 passed, 9 warnings in 623.95s (0:10:23) ===================
```

tests/foreman/api/test_eol_banner.py
```diff
@@ -0,0 +1,49 @@
...
+    :id: 1c2f0d19-a357-4461-9ace-edb468f9ca5c
+
+    :expectedresults: EOL date from satellite-lifecycle package is accurate
+    """
+    current_version = '.'.join(target_sat.version.split('.')[0:2])
+    output = yaml.load(target_sat.execute(rf'cat {LIFECYCLE_METADATA_FILE}').stdout, yaml.Loader)
+    eol_datetime = datetime.strptime(output['releases'][current_version]['end_of_life'], '%Y-%m-%d')
+    result = requests.get(settings.subscription.lifecycle_api_url, verify=False)
+    if result.status_code != 200:
+        raise requests.HTTPError(f'{settings.subscription.lifecycle_api_url} is not accessible')
```
rplevka, line 37:
```suggestion
    result.raise_for_status()
```
https://docs.python-requests.org/en/latest/api/#requests.Response.raise_for_status

tests/foreman/api/test_eol_banner.py
```diff
@@ -0,0 +1,49 @@
...
+    versions = result.json()['data'][0]['versions']
+    version = [v for v in versions if v['name'] == current_version]
+    if len(version) > 0:
```
rplevka, line 40: you are missing `else` part of the conditional. if for some reason the API response would not include the version you are expecting, the test would probably just pass silently (@lhellebr's suggested change would address the problem, but it would fail with a nasty index error in such scenario, so I'd prefer handling it in a civilized way).

The same applies to your `api_date[0][0] == 'string'` checks below

tests/foreman/api/test_eol_banner.py
```diff
@@ -0,0 +1,49 @@
...
+@pytest.mark.tier2
+def test_positive_check_eol_date(target_sat):
+    """Check if the EOL date for the satellite version
+
+    :id: 1c2f0d19-a357-4461-9ace-edb468f9ca5c
+
+    :expectedresults: EOL date from satellite-lifecycle package is accurate
+    """
+    current_version = '.'.join(target_sat.version.split('.')[0:2])
+    output = yaml.load(target_sat.execute(rf'cat {LIFECYCLE_METADATA_FILE}').stdout, yaml.Loader)
```
rplevka, line 33: I guess you know better, but since this test revolves around data inside this file, I'd probably debug-logged its contains for cases it would be rendered empty or invalid and yaml parser would fail on it.

Would probably make TFA much easier.

Optional - no blocker, resolve this comment at will.

rplevka (changes requested):
I believe there are some serious flaws in the assertion logic that might lead to false-negative test results in some scenarios.

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/api/test_eol_banner.py

tests/foreman/api/test_eol_banner.py
```diff
@@ -0,0 +1,53 @@
...
+    version = [v for v in versions if v['name'] == current_version]
+    if len(version) > 0:
+        api_date = [
+            (p['date_format'], p['date'])
+            for p in version[0]['phases']
+            if p['name'] == 'Maintenance support'
+        ]
+        if api_date[0][0] == 'string':
+            assert eol_datetime.strftime("%B, %Y") in api_date[0][1]
+        if api_date[0][0] == 'date':
```
rplevka, line 50: Turn this into an `else if`
And add one more, "catch all" `else` to cover eventual, different values for the fields with raising some meaningful error.
Otherwise, such scenarios might pass unnoticed.

rplevka (changes requested):
One more unhandled `else`

pondrejk:
hello @rplevka @lhellebr mind revisiting?

tests/foreman/api/test_eol_banner.py
```diff
@@ -0,0 +1,55 @@
...
+    versions = result.json()['data'][0]['versions']
+    version = [v for v in versions if v['name'] == current_version]
+    if len(version) > 0:
+        api_date = [
+            (p['date_format'], p['date'])
+            for p in version[0]['phases']
+            if p['name'] == 'Maintenance support'
+        ]
+        if api_date[0][0] == 'string':
+            assert eol_datetime.strftime("%B, %Y") in api_date[0][1]
```
lhellebr, line 49: Can we test for equality?

lhellebr (approved):

rplevka (approved):

pnovotny (approved):
LGTM

PR #13284, opened 2023-12-06
Fix subnet tests

Gauravtalreja1:
### Problem Statement
After https://github.com/SatelliteQE/robottelo/pull/12796, we identified a failure for one of negative tests is using invalid_missing_attributes dict to validate if subnet can be updated, after pytest.raises we're also validating if subnet isn't updated, there are invalid keys in use to check subnet key.

### Solution
1. Fixing the negative test with using exact values to check if subnet isn't updated with invalid values
2. Updating tests to have the teardown
3. Updating tests to use fauxfactory.gen_netmask instead of fixed netmasks

### DIFF
--- a/tests/foreman/cli/test_subnet.py
+++ b/tests/foreman/cli/test_subnet.py
@@ -19,7 +19,7 @@
 import random
 import re
 
-from fauxfactory import gen_choice, gen_integer, gen_ipaddr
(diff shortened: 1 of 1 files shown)


Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_subnet.py

jameerpathan111 (approved):

shubhamsg199 (approved):
Ack

PR #13325, opened 2023-12-07
Default foreman cache changed from file to redis

rmynar:
Default Foreman cache was changed from file to redis. We have to adjust tests to reflect this change.

### DIFF
--- a/tests/foreman/installer/test_installer.py
+++ b/tests/foreman/installer/test_installer.py
@@ -1385,7 +1385,7 @@ def sat_non_default_install(module_sat_ready_rhels):
     installer_args = [
         'scenario satellite',
         f'foreman-initial-admin-password {settings.server.admin_password}',
-        'foreman-rails-cache-store type:redis',
+        'foreman-rails-cache-store type:file',
         'foreman-proxy-content-pulpcore-hide-guarded-distributions false',
     ]
     install_satellite(module_sat_ready_rhels[1], installer_args)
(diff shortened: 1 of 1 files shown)


tests/foreman/installer/test_installer.py
```diff
@@ -1468,7 +1468,7 @@ def test_foreman_rails_cache_store(sat_non_default_install):
 
     :steps:
         1. Install Satellite.
-        2. Verify that foreman-redis package is installed.
+        2. Verify that foreman-redis package is not installed.
```
Griffin-Sullivan, line 1471: Could we change the comment below `Verify foreman-rails-cache-store option works` to `does not work`. Also can we change the docstring and test name to reflect this is testing non-default cache. like `test_foreman_non_default_rails_cache_store`

Griffin-Sullivan (commented):
Just a quick couple of changes to make it obvious what the non-default test is. Everything else looks good

adarshdubey-star (approved):
Ack Code-wise pending comments from Griffin.

tests/foreman/installer/test_installer.py
rmynar: Actually, the installer option (foreman-rails-cache-store) has to work. It should allow users to install with non default cache. The default cache is what has been changed from "file" to "redis". I slightly updated comments and docstring. I hope it's better now. Please check.

pondrejk (approved):

PR #13365, opened 2023-12-11
Update yum repo for ACSes

vsedmik:
### Problem Statement
`rpm-zchunk` repo was removed from fixtures, which makes some ACS tests to fail:
```
"description"=>"ACS remote for url \'https://fixtures.pulpproject.org/rpm-zchunk/\' raised an error \'404: Not Found\'. Please check your ACS remote configuration."
```

### Solution
Use another yum repo, preferably one with the new zst metadata compression.

### DIFF
--- a/robottelo/constants/repos.py
+++ b/robottelo/constants/repos.py
@@ -2,7 +2,7 @@
 
 PULP_FIXTURE_ROOT = 'https://fixtures.pulpproject.org/'
 PULP_SUBPATHS_COMBINED = {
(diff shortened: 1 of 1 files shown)


vsedmik:
> Should this have any effect on other tests using a yum repo? Or do those also not care about what repo is being used.

@sambible Nope, these are used just in combination with `PULP_FIXTURE_ROOT`, in ACS modules specifically AFAIK.

sambible (approved):

pondrejk (approved):

PR #13393, opened 2023-12-12
Candore dependency to use with robottelo image

jyejare:
### Problem Statement

Candore isnt available in robottelo image to run extraction and comparision of pre and post-data in candore CI pipeline using robottelo image for data population using robottelo-cli.

### Solution

Install candore dependency for robottelo to use in image.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,7 @@
 
 betelgeuse==1.10.0
 broker[docker]==0.4.1
+candore==1.0.1
 cryptography==41.0.7
 deepdiff==6.7.1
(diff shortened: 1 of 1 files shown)


Gauravtalreja1 (approved):

ogajduse (changes requested):
Candore is not a direct dependency of robottelo, therefore I would prefer to see candore being installed in the Containerfile/Dockerfile in an extra `pip install ` step.
If we do not use candore in robottelo, I would keep it separate from robottelo's direct dependencies.

PR #11544, opened 2023-05-26
Remove cli.factory dependencies from the repo

shweta83:
This PR will remove the dependency of cli.factory from robottelo and cli_factory will be used in place of that.
All the test modules using cli.factory method should be replaced with cli_factory methods as we are deprecating cli.factory.

Old methods -> New methods

make_activation_key -> make_activation_key
make_architecture -> make_architecture
make_content_view -> make_content_view
make_content_view_with_credentials -> content_view_with_credentials
make_content_view_filter -> content_view_filter
make_content_view_filter_rule -> content_view_filter_rule
make_discoveryrule -> discoveryrule
make_content_credential -> make_content_credential
make_location -> make_location
make_model -> model
make_partition_table -> make_partition_table
make_product -> make_product
make_product_with_credentials -> product_with_credentials
make_product_wait -> make_product_wait
make_repository_with_credentials -> repository_with_credentials
make_role -> role
make_filter -> make_filter
make_scap_policy -> scap_policy
make_subnet -> subnet
make_sync_plan -> sync_plan
make_host -> make_host
make_fake_host -> make_fake_host
make_host_collection -> host_collection
make_job_invocation -> job_invocation
make_job_invocation_with_credentials -> Removed as not used anywhere
make_job_template -> job_template
make_user -> user
make_usergroup -> usergroup
make_usergroup_external --> usergroup_external
make_compute_resource -> compute_resource
make_org -> make_org
make_org_with_credentials -> org_with_credentials
make_realm -> realm
make_reporttemplate -> reporttemplate
make_os -> os
make_scapcontent -> scapcontent
make_domain -> make_domain
make_hostgroup -> hostgroup
make_medium -> make_medium
make_environment -> ?
make_lifecycle_environment -> make_lifecycle_environment
make_tailoringfile -> tailoringfile
make_template -> make_template
make_template_input -> template_input
make_virt_who_config -> virt_who_config
activationkey_add_subscription_to_repo -> activationkey_add_subscription_to_repo
setup_org_for_a_custom_repo -> setup_org_for_a_custom_repo
_setup_org_for_a_rh_repo -> _setup_org_for_a_rh_repo
setup_org_for_a_rh_repo -> setup_org_for_a_rh_repo
add_role_permissions -> add_role_permissions
setup_cdn_and_custom_repositories -> setup_cdn_and_custom_repositories
setup_cdn_and_custom_repos_content -> setup_cdn_and_custom_repos_content
make_http_proxy -> Not used anywhere

PR #13399, opened 2023-12-12
cli org test fixes

pondrejk:
### Problem Statement

module_org can also contain default location which surprised a test before. Also, `hammer organization create` accepts `--users` not `--user` and certainly not `--pasword`. 

Somehow the previous make_organization factory method didn't fail with that (just debug statement), not so with target_sat.cli_factory. So in this case, https://github.com/SatelliteQE/robottelo/pull/11544 helped to uncover the problem

### Solution
Checking just the presence of associated locations, not their count. Using correct parameters for make_org


pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_organization.py -k admin

synkd (approved):

Gauravtalreja1 (approved):

Gauravtalreja1 (approved):

PR #13421, opened 2023-12-13
Vertical menu search test

pnovotny:
New module and test case for vertical menu search functionality.

TC: `test_positive_vertical_navigation_search_end_to_end`

Requires: https://github.com/SatelliteQE/airgun/pull/1076

### DIFF
--- a/tests/foreman/ui/test_search.py
+++ b/tests/foreman/ui/test_search.py
@@ -1,0 +1,213 @@
+"""Test classes for Search component
+
+:Requirement: Search
+
+:CaseAutomation: Automated
+
+:CaseComponent: Search
+
+:Team: Endeavour
+
+:CaseImportance: Medium
+"""
+from collections import namedtuple
+
+from fauxfactory import gen_string
+import pytest
+
+from robottelo.config import settings
+
+SearchData = namedtuple('SearchData', ['expected_items', 'not_expected_items'])
+
+
+def menu_search_should_find(search_func: callable, values: list[str]):
+    """Helper function that does a vertical menu search for each value in `values` list
+    and makes sure it finds a corresponding match.
+    """
+    for search_string in values:
+        results = search_func(search_string)
+        assert results, f'No result found for search string {search_string}!'
+        for result in results:
+            assert (
+                search_string.lower() in result.lower()
+            ), f'Search string {search_string} does not match result {result}!'
+
+
+def menu_search_should_not_find(search_func: callable, values: list[str]):
+    """Helper function that does a vertical menu search for each value in `values` list
+    and makes sure it does not find any match.
+    """
+    for search_string in values:
+        results = search_func(search_string)
+        assert not results, f'Search string {search_string} should not return any results!'
+
+
+@pytest.fixture(scope='module')
+def auditor_module_user(
+    module_target_sat, module_org, module_location, default_org, default_location
+):
+    """User with Auditor role. This user has significantly limited menu access.
+    Expected or not expected data for searching the menu are appended as `search_for` attribute.
+    """
+    auditor_role = module_target_sat.api.Role().search(query={'search': 'name="Auditor"'})[0]
+    password = gen_string('alphanumeric')
+    user = module_target_sat.api.User(
+        admin=False,
+        location=[module_location],
+        organization=[module_org],
+        role=[auditor_role],
+        password=password,
+    ).create()
+    user.password = password
+    user.search_for = SearchData(
+        expected_items=[
+            'audits',
+            'job',
+            'tasks',
+            'bookmark',
+            'about',
+            module_org.name,
+            module_location.name,
+        ],
+        not_expected_items=[
+            '-',
+            'dashboard',
+            'facts',
+            'subscriptions',
+            'product',
+            'content',
+            'sync plan',
+            'host',
+            'role',
+            'compute',
+            'domain',
+            'user',
+            'settings',
+            default_org.name,
+            default_location.name,
+        ],
+    )
+
+    yield user
+
+    user.delete()
+
+
+@pytest.fixture(scope='module')
+def admin_user(module_target_sat, module_org, module_location, default_org, default_location):
+    """Admin user.
+    Expected or not expected data for searching the menu are appended as `search_for` attribute.
+    """
+    admin_user = module_target_sat.api.User().search(
+        query={'search': f'login={settings.server.admin_username}'}
+    )[0]
+    admin_user.password = settings.server.admin_password
+    admin_user.search_for = SearchData(
+        expected_items=[
+            '-',
+            'dashboard',
+            'facts',
+            'subscriptions',
+            'product',
+            'content',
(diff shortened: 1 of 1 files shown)


pnovotny:
trigger: test-robottelo
pytest: tests/foreman/ui/test_search.py -k test_positive_vertical_navigation_search_end_to_end
airgun: 1076

pnovotny:
trigger: test-robottelo
pytest: tests/foreman/ui/test_search.py -k test_positive_vertical_navigation_search_end_to_end
airgun: 1076

pnovotny:
trigger: test-robottelo
pytest: tests/foreman/ui/test_search.py -k test_positive_vertical_navigation_search_end_to_end
airgun: 1076

pnovotny:
trigger: test-robottelo
pytest: tests/foreman/ui/test_search.py -k test_positive_vertical_navigation_search_end_to_end
airgun: 1076
env:
  ROBOTTELO_ui__record_video: true

pnovotny:
trigger: test-robottelo
pytest: tests/foreman/ui/test_search.py -k test_positive_vertical_navigation_search_end_to_end
airgun: 1076
env:
  ROBOTTELO_ui__record_video: true

pondrejk (approved):

tests/foreman/ui/test_search.py
```diff
@@ -0,0 +1,175 @@
...
+
+    :Setup:
+        Create a non-admin user with Auditor role.
+        Create a custom organization and location.
+
+    :Steps:
+        1. Perform a search in the vertical navigation menu with following inputs:
+            1a. valid characters
+            1b. invalid characters
+            1c. current user organization and location
```
lhellebr, line 87: So this should be able to search orgs and locs by their names. What other entities should it be able to search for? I see it doesn't work for job templates. But I wouldn't expect it to work.

tests/foreman/ui/test_search.py
```diff
@@ -0,0 +1,175 @@
...
+            1d. custom organization and location not assigned to the user
+        2. Perform the actions as:
```
lhellebr, line 89: I think user should be parametrized. So you will get two tests collected, with their respective users and fields expected to be/not be found.

tests/foreman/ui/test_search.py
```diff
@@ -0,0 +1,175 @@
...
+        'tasks',
+        'bookmark',
+        'about',
+    ]
+
+    with module_target_sat.ui_session() as session:
+        search = session.host.search_menu
+
+        results = search(' ')
+        assert len(results) == 10, f'Maximum number of search results is 10, got {len(results)}!'
```
lhellebr, line 136: Make 10 a constant?

tests/foreman/ui/test_search.py
```diff
@@ -0,0 +1,175 @@
...
+            *admin_menu_items,
+            *auditor_menu_items,
+            default_org.name,
+            default_location.name,
+            module_org.name,
+            module_location.name,
+        ]
+        menu_search_should_find(search, valid_search_strings)
+
+        menu_search_should_not_find(search, invalid_characters)
```
lhellebr, line 149: Why only invalid characters? Search for some queries consisting of valid characters that you however don't expect to be found.

lhellebr (commented):
Overally good but some comments left

tests/foreman/ui/test_search.py
pnovotny: It searches only inside organizations and locations. And the User menu. This is due to design where org., loc. and user are part of the menu, but it's displayed on the top in the "desktop" layout. 
If you narrow the browser window - like a mobile layout - the three menus are moved from top to the left vertical menu. And therefore the search also applies to them.

tests/foreman/ui/test_search.py
pnovotny: Ok, I've refactored the test to be parametrized. Passing some user-specific data - mainly fixtures - was a bit head-scratcher, but in the end I hope it's still readable and not overly complex.

tests/foreman/ui/test_search.py
pnovotny: Done.

pnovotny:
trigger: test-robottelo
pytest: tests/foreman/ui/test_search.py -k test_positive_vertical_navigation_search_end_to_end

pnovotny:
trigger: test-robottelo
pytest: tests/foreman/ui/test_search.py -k test_positive_vertical_navigation_search_end_to_end
env:
ROBOTTELO_ui__record_video: true

lvrtelov:
@pnovotny can we get this merge soon? it's open for a while now

pnovotny:
trigger: test-robottelo
pytest: tests/foreman/ui/test_search.py -k test_positive_vertical_navigation_search_end_to_end
airgun: 1233
env:
ROBOTTELO_ui__record_video: true

Satellite-QE:
**PRT Result**
```
Build Number: 5702
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/ui/test_search.py -k test_positive_vertical_navigation_search_end_to_end
Test Result : ================= 2 passed, 33 warnings in 1047.12s (0:17:27) ==================
```

tests/foreman/ui/test_search.py
```diff
@@ -0,0 +1,213 @@
...
+    for search_string in values:
+        results = search_func(search_string)
+        assert results, f'No result found for search string {search_string}!'
+        for result in results:
+            assert (
+                search_string.lower() in result.lower()
+            ), f'Search string {search_string} does not match result {result}!'
+
+
+def menu_search_should_not_find(search_func: callable, values: list[str]):
```
Griffin-Sullivan, line 36: not a blocker but I'm curious if running the above function with `not` in front of it would achieve the same result?

Griffin-Sullivan (approved):

sambible (approved):

tests/foreman/ui/test_search.py
sambible: I'd agree, there's probably a way to do this without a separate helper. Doesn't really affect the test overall though, more of a style difference.

pnovotny: I get that the first three lines of the two functions are pretty similar, but merging both functions together to one and adding some "bool" argument, i.e., expect_to_find=True/False, would lead to more if..else branching and thus to less readability, IMO. Esp. comparing to how many lines of code we save by that.

Using `not menu_search_should_find()` here won't help, because `menu_search_should_find()` doesn't return bool, but does asserts instead - so a test failure is easily identifiable in the results.
Doing something like this
```
with pytest.raises(AssertionError):
    menu_search_should_find()
```
would solve it, but then there is a case that AssertionError is thrown by the second assert (line 31) and that would be wrong.

pnovotny:
trigger: test-robottelo
pytest: tests/foreman/ui/test_search.py -k test_positive_vertical_navigation_search_end_to_end
airgun: 1250

PR #13475, opened 2023-12-15
extra check not recording property incase of video recording is false

omkarkhatavkar:
### Problem Statement
Currently, every time we record a video URL is a JUnit report, which is not necessary in case the video recording is false.

### Solution
Adding extra layer check to see record_property is not null and video recording is true

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -1849,8 +1849,9 @@ class Satellite(Capsule, SatelliteMixins):
             video_url = settings.ui.grid_url.replace(
                 ':4444', f'/videos/{ui_session.ui_session_id}.mp4'
             )
-            self.record_property('video_url', video_url)
-            self.record_property('session_id', ui_session.ui_session_id)
+            if self.record_property is not None and settings.ui.record_video:
+                self.record_property('video_url', video_url)
+                self.record_property('session_id', ui_session.ui_session_id)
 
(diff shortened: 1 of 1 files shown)


robottelo/hosts.py
```diff
@@ -1849,8 +1849,9 @@ def get_caller():
             video_url = settings.ui.grid_url.replace(
                 ':4444', f'/videos/{ui_session.ui_session_id}.mp4'
             )
-            self.record_property('video_url', video_url)
-            self.record_property('session_id', ui_session.ui_session_id)
+            if not self.record_property and settings.ui.record_video:
```
jyejare, line 1852: IMHO check for`self.record_property` to be none is not required because this could have been already populated when nested session would be runinng for a test of same satellite ?

omkarkhatavkar: I wanted to prevent a scenario where the test passes, but the record property for that specific SAT instance doesn't get retrieved somehow. Omitting this condition could potentially lead to test failures. 

To address this, I can include a warning message or logging mechanism in case such a situation occurs. Another approach is to detect this issue through failing UI tests when the UI recording session cannot be obtained. This will just avoid the side effects incase.

JacobCallahan (approved):

jyejare (approved):

PR #13489, opened 2023-12-18
Refactor and simplify ping test

rmynar:
Cache status was added to `hammer ping` output and there can be more cache servers configured so the output syntax has changed. I changed and simplified the test so it can parse current output and it also meets test description.

sample output:
```
database:         
    Status:          ok
    Server Response: Duration: 0ms
cache:            
    servers: 
     1) Status:          ok
        Server Response: Duration: 0ms
candlepin:        
    Status:          ok
    Server Response: Duration: 23ms
...
```
PS: I'm not aware of negative scenario for this testcase (i.e. scenario when some of Satellite services report status "FAIL"). We may add it in future.


rmynar:
trigger: test-robottelo
pytest: tests/foreman/cli/test_ping.py

tests/foreman/cli/test_ping.py
```diff
@@ -42,21 +42,13 @@ def test_positive_ping(target_sat, switch_user):
...
     result = target_sat.execute(f"su - {'postgres' if switch_user else 'root'} -c 'hammer ping'")
     assert result.stderr[1].decode() == ''
-    assert result.status == 0
 
-    status_count = 0
-    ok_count = 0
-    # Exclude message from stdout for services candlepin_events and katello_events
-    result.stdout = [line for line in result.stdout.splitlines() if 'message' not in line]
+    # Filter lines containing "Status: "
+    statuses = [line for line in result.stdout.splitlines() if 'status' in line.lower()]
```
JacobCallahan, line 47: There's a disconnect between the comment and the actual check. Personally, I think the comprehension is clear enough to not need a comment.

tests/foreman/cli/test_ping.py
```diff
@@ -42,21 +42,13 @@ def test_positive_ping(target_sat, switch_user):
...
-    # example [1, 2, 3, 4, 5, 6] will return [(1, 2, 3), (4, 5, 6)]
-    # only the status line is relevant for this test
-    for _, status, _ in zip(*[iter(result.stdout)] * 3):
-        status_count += 1
-
-        if status.split(':')[1].strip().lower() == 'ok':
-            ok_count += 1
+    # Get count of total status lines and lines containing OK
+    status_count = len(statuses)
+    ok_count = len([status for status in statuses if 'ok' in status.lower()])
```
JacobCallahan, line 51: I'm not sure what possible statuses could be encountered, but a check for just an `"ok"` sub-string seems like it could lead to false positives.

JacobCallahan (changes requested):

tests/foreman/cli/test_ping.py
rmynar: I slightly changed the comment and also the filter expression. Now I filter "status:" (added the colon)

JacobCallahan (approved):

pondrejk (approved):

PR #13491, opened 2023-12-19
Add Ruff rule group B (flake8-bugbear)

JacobCallahan:
This change includes the rule addition and mostly manual changes to existing code in order to bring it in line with the new rules and avoid as many ignores as is reasonable.

I added a number of {{from err}} statements to the end of exceptions that currently include the error message. The error message being included in the raised string can be removed later, but keeping for consistency for now.

In cases where I raised from None, the raised exception is clear enough that the additional context of the parent exception isn't helpful.

cli/test\_ping.py has a zip that should probably be strict.


tests/foreman/cli/test_ping.py
```diff
@@ -52,7 +52,7 @@ def test_positive_ping(target_sat, switch_user):
     # iterate over the lines grouping every 3 lines
     # example [1, 2, 3, 4, 5, 6] will return [(1, 2, 3), (4, 5, 6)]
     # only the status line is relevant for this test
-    for _, status, _ in zip(*[iter(result.stdout)] * 3):
+    for _, status, _ in zip(*[iter(result.stdout)] * 3, strict=False):  # should this be strict?
```
ogajduse, line 55: I did some solid archeology and I found that aedae75647180f150ce26801cdeb869e5bdb7270 introduced the `zip` logic. @rmynar is changing it in #13489 anyway.

To answer the question in the comment, I would say that it should not be strict since
```python
(Pdb++) foo = zip(*[iter(result.stdout)] * 3, strict=True)
(Pdb++) for ff in foo: print(ff)
('database:         ', '    Status:          ok', '    Server Response: Duration: 0ms')
('candlepin:        ', '    Status:          ok', '    Server Response: Duration: 17ms')
('candlepin_auth:   ', '    Status:          ok', '    Server Response: Duration: 14ms')
('candlepin_events: ', '    Status:          ok', '    Server Response: Duration: 0ms')
('katello_events:   ', '    Status:          ok', '    Server Response: Duration: 0ms')
('pulp3:            ', '    Status:          ok', '    Server Response: Duration: 77ms')
('pulp3_content:    ', '    Status:          ok', '    Server Response: Duration: 39ms')
('foreman_tasks:    ', '    Status:          ok', '    Server Response: Duration: 3ms')
*** ValueError: zip() argument 2 is shorter than argument 1
```

ogajduse (approved):

tests/foreman/cli/test_ping.py
JacobCallahan: Hmmm... i wonder that's being cutoff by the zip then.
Will change strict to False for this, thanks for digging into this @ogajduse!

robottelo/host_helpers/cli_factory.py
```diff
@@ -717,13 +724,15 @@ def _setup_org_for_a_rh_repo(self, options=None):
...
         else:
             env_id = options['lifecycle-environment-id']
-        # If manifest does not exist, clone and upload it
-        if len(self._satellite.cli.Subscription.exists({'organization-id': org_id})) == 0:
-            with clone() as manifest:
-                try:
-                    self._satellite.upload_manifest(org_id, manifest.content)
-                except CLIReturnCodeError as err:
-                    raise CLIFactoryError(f'Failed to upload manifest\n{err.msg}')
+        # Clone manifest and upload it
```
synkd, line 727: It looks to me like this change will cause tests that pass `force_use_cdn=True` to the `setup_org_for_a_rh_repo` method to always use cloned manifests. Is that intended?

synkd (commented):
Looks good overall, just have one manifest-related question that I would like answered before ACKing.

robottelo/host_helpers/cli_factory.py
JacobCallahan: great catch! I'm not sure how that conditional was deleted, but i don't see a reason why it should have been. put back in place in the latest change.

jyejare (approved):

synkd (approved):

PR #13505, opened 2023-12-19
fix cli errata tests and update cli factory method

vijaysawant:
### Problem Statement
**Com eval -** 2 tests were failing due to cli factory method > `setup_org_for_a_rh_repo` with error mentioned below
`test_positive_list_filter_by_org_sort_by_date`
`test_positive_list_filter_by_cve`

**Error:**
> E   AttributeError: 'Manifest' object has no attribute 'path'

### Solution
Basically due to old way of cloning and uploading manifest into the satellite didn't work, also as per `6.15.z` and later version of satellite `*_sca_manifest_org` fixture is supposed to be used.
With this fix above 2 tests and couple of few tests are expected to get fix.

### Related Issues
The below 2 tests will get benifit of the change made in cli_factory and will pass accordingly.
`cli.test_activationkey.py > test_positive_subscription_quantity_attached`
`cli.test_host.py > test_negative_without_attach_with_lce`

### DIFF
--- a/robottelo/host_helpers/cli_factory.py
+++ b/robottelo/host_helpers/cli_factory.py
@@ -691,7 +691,7 @@ class CLIFactory:
 
         1. Checks if organization and lifecycle environment were given, otherwise
             creates new ones.
-        2. Clones and uploads manifest.
+        2. If manifest does not exist, clone and upload it.
         3. Enables RH repo and synchronizes it.
         4. Checks if content view was given, otherwise creates a new one and
             - adds the RH repo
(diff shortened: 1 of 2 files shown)


vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/cli/test_errata.py -k "test_positive_list_filter_by_org_sort_by_date" -k "test_positive_list_filter_by_cve"

damoore044:
trigger: test-robottelo
pytest: tests/foreman/cli/test_errata.py -k "test_positive_list_filter_by_org_sort_by_date or test_positive_list_filter_by_cve"

damoore044 (approved):
nice work on these!

robottelo/host_helpers/cli_factory.py
```diff
@@ -717,15 +717,6 @@ def _setup_org_for_a_rh_repo(self, options=None):
             env_id = self.make_lifecycle_environment({'organization-id': org_id})['id']
         else:
             env_id = options['lifecycle-environment-id']
-        # Clone manifest and upload it
```
vsedmik, line 720: One thing to consider: This removal means that the Org provided via `options` must have some manifest already added, otherwise it would fail on line 731/722 I guess. Same situation would happen when no Org is provided - new one is created on line 713, but no manifest is added.
To address this let's make the Org param mandatory OR upload some manifest here like we do [in fixtures](https://github.com/SatelliteQE/robottelo/blob/83ca28419c129aad304d49c89f758af36b9a74b7/pytest_fixtures/component/taxonomy.py#L102).

Whatever path you choose, please update the docstring too (L694).

vsedmik (commented):
Nice to see those tests passing again, just one small concern about the manifest to consider.

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/cli/test_errata.py -k 'test_positive_list_filter_by_org_sort_by_date or test_positive_list_filter_by_cve'

vsedmik (approved):
ACK, PRT pending

robottelo/host_helpers/cli_factory.py
vijaysawant: `self._satellite.cli.Subscription.exists({'organization-id': org_id})`  will return dictionary if manifest already uploaded, otherwise it will return empty dictionary

PR #13544, opened 2023-12-21
Add CU scenario for puma worker count

Griffin-Sullivan:
### Problem Statement
Missing coverage for BZ 2025760.

### Solution
Add new test to cover the scenario for changing the puma worker count.

### DIFF
--- a/tests/foreman/destructive/test_installer.py
+++ b/tests/foreman/destructive/test_installer.py
@@ -16,6 +16,8 @@
 
 :Upstream: No
 """
+import random
+
 from fauxfactory import gen_domain, gen_string
 import pytest
 
@@ -147,3 +149,40 @@ def test_positive_installer_certs_regenerate(target_sat):
     )
     assert result.status == 0
     assert 'FAIL' not in target_sat.cli.Base.ping()
+
+
+def test_positive_installer_puma_worker_count(target_sat):
+    """Installer should set the puma worker count and thread max without having to manually
+    restart the foreman service.
+
+    :id: d0e7d958-dd3e-4962-bf5a-8d7ec36f3485
+
+    :steps:
+        1. Check how many puma workers there are
+        2. Select a new worker count that is less than the default
+        2. Change answer's file to have new count for puma workers
+        3. Run satellite-installer --foreman-foreman-service-puma-workers new_count --foreman-foreman-service-puma-threads-max new_count
+
+    :expectedresults: aux should show there are only new_count puma workers after installer runs
+
+    :BZ: 2025760
+
+    :customerscenario: true
+    """
+    count = int(target_sat.execute('pgrep --full "puma: cluster worker" | wc -l').stdout)
+    worker_count = str(random.randint(1, count - 1))
+    result = target_sat.install(
+        InstallerCommand(
+            foreman_foreman_service_puma_workers=worker_count,
+            foreman_foreman_service_puma_threads_max=worker_count,
+        )
+    )
+    assert result.status == 0
+    result = target_sat.execute(f'grep "foreman_service_puma_workers" {SATELLITE_ANSWER_FILE}')
+    assert worker_count in result.stdout
+    result = target_sat.execute('ps aux | grep -v grep | grep -e USER -e puma')
+    for i in range(count):
+        if i < int(worker_count):
+            assert f'cluster worker {i}' in result.stdout
+        else:
+            assert f'cluster worker {i}' not in result.stdout


Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_installer.py::test_positive_installer_puma_worker_count

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_installer.py::test_positive_installer_puma_worker_count

tests/foreman/destructive/test_installer.py
```diff
@@ -147,3 +147,39 @@ def test_positive_installer_certs_regenerate(target_sat):
...
+    :expectedresults: aux should show there are only 5 puma workers after installer runs
+
+    :BZ: 2025760
+
+    :customerscenario: true
+    """
+    result = target_sat.execute('ps aux | grep -v grep | grep -e USER -e puma')
+    for i in range(9):
+        assert f'cluster worker {i}' in result.stdout
+    target_sat.execute(f'sed -i "/foreman_service_puma_workers: / s/$/5/" {SATELLITE_ANSWER_FILE}')
```
Gauravtalreja1, line 172: why updating answers file first then running installer, and we are already verifying installer works with RC so we can rely on it , right? or verify answers file after installer run?

tests/foreman/destructive/test_installer.py
```diff
@@ -147,3 +147,39 @@ def test_positive_installer_certs_regenerate(target_sat):
...
+        2. Change answer's file to have 5 puma workers
+        3. Run satellite-installer --foreman-foreman-service-puma-workers 5 --foreman-foreman-service-puma-threads-max 5
+
+    :expectedresults: aux should show there are only 5 puma workers after installer runs
+
+    :BZ: 2025760
+
+    :customerscenario: true
+    """
+    result = target_sat.execute('ps aux | grep -v grep | grep -e USER -e puma')
```
Gauravtalreja1, line 169:
```suggestion
    result = target_sat.execute('ps aux | grep -e USER -e "puma: cluster worker"')
```

tests/foreman/destructive/test_installer.py
Griffin-Sullivan: So I don't need this line because the installer will do this work? We just need to verify the installer changed these lines after it's been run?

tests/foreman/destructive/test_installer.py
Griffin-Sullivan: I'm just following what the BZ did https://bugzilla.redhat.com/show_bug.cgi?id=2025760. I'd prefer to keep as is.

tests/foreman/destructive/test_installer.py
Gauravtalreja1: yes, I think so

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_installer.py::test_positive_installer_puma_worker_count

Gauravtalreja1 (approved):

tests/foreman/destructive/test_installer.py
```diff
@@ -147,3 +147,40 @@ def test_positive_installer_certs_regenerate(target_sat):
...
+    for i in range(9):
```
pondrejk, line 170: this feels a bit fragile towards future default value changes, would it make sense to first source the default setting from the answer file?

pondrejk (approved):
minor comment

tests/foreman/destructive/test_installer.py
Griffin-Sullivan: The answer file is blank which is what makes this create 9 puma workers by default. We could instead count the amount of puma workers and then range off of the len?

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_installer.py::test_positive_installer_puma_worker_count

tests/foreman/destructive/test_installer.py
Griffin-Sullivan: Ok so I learned this piece of code determines worker count https://github.com/theforeman/puppet-foreman/blob/908ed638790f3cfc6e2f5f62e1b023c907a9913c/manifests/config.pp#L83C1-L92. Therefore, I changed the test to record what the current worker count is, run the installer with worker count set to a number less than that count, and then assert it worked.

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_installer.py::test_positive_installer_puma_worker_count

tests/foreman/destructive/test_installer.py
```diff
@@ -147,3 +149,44 @@ def test_positive_installer_certs_regenerate(target_sat):
...
+
+    :BZ: 2025760
+
+    :customerscenario: true
+    """
+    result = target_sat.execute('ps aux | grep -v grep | grep -e USER -e puma')
+    count = 0
+    for line in result.stdout.splitlines():
+        if 'puma: cluster worker' in line:
+            count += 1
```
JacobCallahan, line 175: likely easier to do it this way
```suggestion
    count = int(target_sat.execute('pgrep --full "puma: cluster worker" | wc -l').stdout)
```

tests/foreman/destructive/test_installer.py
```diff
@@ -147,3 +149,44 @@ def test_positive_installer_certs_regenerate(target_sat):
...
+    )
+    assert result.status == 0
+    result = target_sat.execute(f'grep "foreman_service_puma_workers" {SATELLITE_ANSWER_FILE}')
+    assert worker_count in result.stdout
+    result = target_sat.execute('ps aux | grep -v grep | grep -e USER -e puma')
+    for i in range(count):
+        if i < int(worker_count):
+            assert f'cluster worker {i}' in result.stdout
+        else:
+            assert f'cluster worker {i}' not in result.stdout
```
JacobCallahan, line 192: see above for better process counting

JacobCallahan (changes requested):

tests/foreman/destructive/test_installer.py
```diff
@@ -147,3 +149,44 @@ def test_positive_installer_certs_regenerate(target_sat):
...
+def test_positive_installer_puma_worker_count(target_sat):
+    """Installer should set the puma worker count and thread max without having to manually
+    restart the foreman service.
+
+    :id: d0e7d958-dd3e-4962-bf5a-8d7ec36f3485
+
+    :steps:
+        1. Check how many puma workers there are by default (9)
+        2. Change answer's file to have 5 puma workers
+        3. Run satellite-installer --foreman-foreman-service-puma-workers 5 --foreman-foreman-service-puma-threads-max 5
```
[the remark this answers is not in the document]
Griffin-Sullivan: No....

tests/foreman/destructive/test_installer.py
```diff
@@ -147,3 +149,44 @@ def test_positive_installer_certs_regenerate(target_sat):
...
+
+    :customerscenario: true
+    """
+    result = target_sat.execute('ps aux | grep -v grep | grep -e USER -e puma')
+    count = 0
+    for line in result.stdout.splitlines():
+        if 'puma: cluster worker' in line:
+            count += 1
+    numbers = range(1, count)[:-1]
+    worker_count = str(random.choice(numbers))
```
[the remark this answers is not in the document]
Griffin-Sullivan: yeah I'll add your suggestion and a comment

tests/foreman/destructive/test_installer.py
Griffin-Sullivan: What I liked about this loop is it also asserted that there were certain workers not in the output

JacobCallahan (approved):

PR #13589, opened 2024-01-03
Add test case for ISS export with history ID

vsedmik:
### Problem Statement
Old negative stub for exports with "future datetime" was obsoleted since datetime is not used anymore. Instead the `from-history-id` option can be used, which is not covered either.

### Solution
This PR adds a test case to export repo using `from-history-id`. It's not positive nor negative as it tests both. Also exercises exports listing.

### DIFF
--- a/robottelo/cli/content_export.py
+++ b/robottelo/cli/content_export.py
@@ -31,7 +31,7 @@ class ContentExport(Base):
     command_requires_org = True
 
     @classmethod
-    def list(cls, output_format='json', options=None):
+    def list(cls, options=None, output_format='json'):
         """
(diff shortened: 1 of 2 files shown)


vsedmik:
trigger: test-robottelo
pytest: tests/foreman/cli/test_satellitesync.py -k history

pondrejk (approved):

vsedmik:
Rebased.

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/cli/test_satellitesync.py -k history

Griffin-Sullivan (approved):

PR #13593, opened 2024-01-03
[6.14.z] Markers as test property in reports

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/13043

Test Markers as test metadata in Polarion, RP and Ibutsu reporting tools.

These markers would be used in RP and ibutsu to filter the results based on markers !


jyejare:
**Merging this PR on priority, since:**

1. The PRT test execution is passing for other branches so dont need to run here again as the change doesnt differs in versions.
2. PR #13599 has to be immediately merged for issues in Reporting tools and this PR would fail the cherrypicking!

PR #13599, opened 2024-01-04
Fixing the missing markers as independent properties

jyejare:
### Problem Statement
In PR #13043, Missed adding filtered markers as independent test properties impacting reporting tools like RP in reporting/filtering abilities.

### Solution
This PR corrects that by adding only filtered markers as test properties.

### Related Issues
No.

### DIFF
--- a/pytest_plugins/metadata_markers.py
+++ b/pytest_plugins/metadata_markers.py
@@ -119,12 +119,15 @@ def pytest_collection_modifyitems(items, config):
         markers_prop_data = []
         exclude_markers = ['parametrize', 'skipif', 'usefixtures', 'skip_if_not_set']
         for marker in item.iter_markers():
-            prop = marker.name
-            if prop in exclude_markers:
+            proprty = marker.name
+            if proprty in exclude_markers:
                 continue
             if marker_val := next(iter(marker.args), None):
-                prop = '='.join([prop, str(marker_val)])
-            markers_prop_data.append(prop)
+                proprty = '='.join([proprty, str(marker_val)])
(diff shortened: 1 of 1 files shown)


pytest_plugins/metadata_markers.py
```diff
@@ -119,12 +119,15 @@ def pytest_collection_modifyitems(items, config):
         markers_prop_data = []
         exclude_markers = ['parametrize', 'skipif', 'usefixtures', 'skip_if_not_set']
         for marker in item.iter_markers():
-            prop = marker.name
-            if prop in exclude_markers:
+            proprty = marker.name
```
omkarkhatavkar, line 122: typo here.

omkarkhatavkar (approved):
ACK, pending minor typo

pytest_plugins/metadata_markers.py
jyejare: Its deliberately done as `property` is a reserved keyword in python :)

jyejare: Variable could be of any name though, including typing mistakes !

rplevka: `PEP8` suggests to append a trailing underscore rather than introducing a spelling error in such cases

https://peps.python.org/pep-0008/#descriptive-naming-styles
```
...
single_trailing_underscore_: used by convention to avoid conflicts with Python keyword
```

jyejare: OOPs! Its merged !

PR #13601, opened 2024-01-04
[6.12.z] Markers as test property in reports (#13043)

jyejare:
### Problem Statement
Cherrypick failed for https://github.com/SatelliteQE/robottelo/pull/13043 to 6.12.z

### Solution
Manually cherrypicking !

### Related Issues
None


jyejare:
**Merging this PR on priority, since:**

1. The PRT test execution is passing for other branches so dont need to run here again as the change doesnt differs in versions.
2. PR #13599 has to be immediately merged for issues in Reporting tools and this PR would fail the cherrypicking!

PR #13602, opened 2024-01-04
Component Evaluation - add RHEL 8 and 9 host versions

lhellebr:


lhellebr:
trigger: test-robottelo
pytest: tests/foreman/ui/test_host.py::test_global_re_registration_host_with_force_ignore_error_options

tests/foreman/ui/test_host.py
```diff
@@ -1535,8 +1535,9 @@ def test_global_registration_upgrade_subscription_manager(
 
 @pytest.mark.tier3
 @pytest.mark.usefixtures('enable_capsule_for_registration')
+@pytest.mark.rhel_ver_list([7, 8, 9])
```
Gauravtalreja1, line 1538:
```suggestion
@pytest.mark.rhel_ver_match('[^6].*')
```

tests/foreman/ui/test_host.py
```diff
@@ -1535,8 +1535,9 @@ def test_global_registration_upgrade_subscription_manager(
 
 @pytest.mark.tier3
 @pytest.mark.usefixtures('enable_capsule_for_registration')
```
Gauravtalreja1, line 1537: we can remove enable_capsule_for_registration ?

lhellebr: I'd prefer not to. As it is now, it would fail with:
```
widgetastic_patternfly4.formselect.FormSelectDisabled: FormSelect('.//*[contains(@data-ouia-component-type,"PF4/FormSelect") and @data-ouia-component-id="reg_smart_proxy"]') is not enabled
```
It can be changed so that WebUI doesn't try to fill it but I think it's more future-proof now. Now, `default_smart_proxy` is used which amounts to the Satellite itself but it may change one time and then no fix will be necessary here.

lhellebr:
Thanks for the comments, I've addressed them. This fails for 6.15 due to some unrelated navigation issues. It passes on 6.14.

Gauravtalreja1 (approved):

PR #13614, opened 2024-01-04
Fix and update installer ping test

Griffin-Sullivan:
### Problem Statement
This test is failing on stream because hammer ping now has a couple new lines that are nested for the cache

### Solution
Simplify the assertion so it's much easier.

### DIFF
--- a/tests/foreman/installer/test_installer.py
+++ b/tests/foreman/installer/test_installer.py
@@ -1628,18 +1628,10 @@ def test_positive_check_installer_hammer_ping(target_sat):
     """
     # check status reported by hammer ping command
     result = target_sat.execute('hammer ping')
-    test_result = {}
-    service = None
-    for line in result.stdout.strip().replace(' ', '').split('\n'):
-        if line.split(':')[0] not in ('Status', 'ServerResponse', 'message'):
-            service = line.split(':')[0]
-            test_result[service] = {}
-        else:
-            key, value = line.split(":", 1)
-            test_result[service][key] = value
-
-    not_ok = {svc: result for svc, result in test_result.items() if result['Status'] != 'ok'}
-    assert not not_ok
+    assert result.status == 0
+    for line in result.stdout.split('\n'):
+        if 'Status' in line:
+            assert 'ok' in line
 
 
 @pytest.mark.e2e


Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/installer/test_installer.py::test_positive_check_installer_hammer_ping

tests/foreman/installer/test_installer.py
```diff
@@ -1627,19 +1627,10 @@ def test_positive_check_installer_hammer_ping(target_sat):
...
-        else:
-            key, value = line.split(":", 1)
-            test_result[service][key] = value
-
-    not_ok = {svc: result for svc, result in test_result.items() if result['Status'] != 'ok'}
-    assert not not_ok
+    result = target_sat.cli.Base.ping()
+    for line in result.split('\n'):
+        if 'Status' in line:
+            assert 'ok' in line
```
lhellebr, line 1633: This is a weaker assertion. For example, when I stop the foreman services, I get this output and the assertion won't fail for it:

```
Error: Failed to open TCP connection to <FQDN>:443 (Connection refused - connect(2) for "<FQDN>" port 443)
```

lhellebr (changes requested):

Griffin-Sullivan:
@lhellebr during local testing I found that the command won't execute if services are unreachable / down. This causes the test to fail and give you the output you pasted above. So I think it should work fine the way it is.

sambible (approved):

tests/foreman/installer/test_installer.py
```diff
@@ -1627,19 +1627,10 @@ def test_positive_check_installer_hammer_ping(target_sat):
...
-        if line.split(':')[0] not in ('Status', 'ServerResponse', 'message'):
-            service = line.split(':')[0]
-            test_result[service] = {}
-        else:
-            key, value = line.split(":", 1)
-            test_result[service][key] = value
-
-    not_ok = {svc: result for svc, result in test_result.items() if result['Status'] != 'ok'}
-    assert not not_ok
+    result = target_sat.cli.Base.ping()
```
rmynar, line 1630: I would prefer the original code `target_sat.execute('hammer ping')`. It allows to investigate stdout, stderr and return code.
It seems that `cli.Base.ping()` simply returns only stdout on success or raises _CLIReturnCodeError_ on failure (ping returns non-zero value if any service returns status FAIL). Further assertions are redundant.

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/installer/test_installer.py::test_positive_check_installer_hammer_ping

rmynar (approved):

lhellebr:
What you're trying to do goes against the sense of this test. Think about it, what is the purpose of the second assertion? It will only evaluate if return code of the previous command is 0.
You state `the command won't execute if services are unreachable`, I think what you mean is it executes and returns non-zero status.
Now, point of this test is testing `hammer ping`, not only to the level of it returning 0 when it should but also parsing its output and having some expectations on it. The original version didn't do it ideally but it was still a stronger assertion than the new version. If you conclude that "if return value is 0 and there is not a single line saying `Status: ok` means this still should pass", then you can almost remove the second assertion entirely.

rmynar:
I think that the purpose of this change is mainly to parse the new output of hammer ping. For hammer ping we have separate tests (https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/cli/test_ping.py)

lhellebr:
That being said, you're the owner of the test so feel free to merge without my ACK if you feel I'm being too harsh :)

Griffin-Sullivan:
I was unaware of the other ping test. I'm not sure I understand the reason we have both tests. I feel inclined to remove this test since it's just a different version of the other test in test_ping.py. Thoughts @rmynar since you did the eval for this file?

rmynar:
I was also unaware of the other ping test when I did the eval. However I think this one is kind of sanity test for the Installer component (i.e. "No need to continue further testing when some service fails to run after installation/upgrade"). I'm not against removing this test or moving the assertion into another testcase. Current changes seem sufficient to adapt the test to the new hammer ping output.

PR #13647, opened 2024-01-08
fix assert_job_invocation_status calls

pondrejk:
### Problem Statement
 assert_job_invocation_status requires satellite as a first argument (since recent refactor), but the calls were not providing it, hence ~3 test failures

### Solution
Updating the calls to include target_sat. For reason unknown, this problem occurs only in master and 6.15


pondrejk:
trigger: test-robottelo
pytest:  tests/foreman/cli/test_remoteexecution.py  -k test_positive_run_recurring_job_with_max_iteration

Gauravtalreja1 (approved):

PR #13694, opened 2024-01-09
Adding coverage for BZ2221621

rmynar:
Added new testcase to cover BZ 2221621 (and it's clone BZ 2238363)

BZ reported in Sat 6.13. Attempting to run installer with invalid certificates on existing installation causes breakage of current setup - hammer ping reports failures. After fix the installer returns error and leaves current setup (hammer ping reports OK statuses)

### DIFF
--- a/tests/foreman/destructive/test_installer.py
+++ b/tests/foreman/destructive/test_installer.py
@@ -179,3 +179,44 @@ def test_positive_installer_puma_worker_count(target_sat):
             assert f'cluster worker {i}' in result.stdout
         else:
             assert f'cluster worker {i}' not in result.stdout
+
+
+def test_negative_handle_invalid_certificate(cert_setup_destructive_teardown):
+    """Satellite installer should not do any harmful changes to existing satellite after attempt
+    to use invalid certificates.
+
+    :id: 97b72faf-4684-4d8c-ae0e-1ebd5085620b
+
+    :steps:
+        1. Launch satellite installer and attempt to use invalid certificates
+
+    :expectedresults: Satellite installer should fail and Satellite should be running
+
+    :BZ: 2221621, 2238363
+
+    :customerscenario: true
+
+    """
+    cert_data, satellite = cert_setup_destructive_teardown
+
+    # check if satellite is running
+    result = satellite.execute('hammer ping')
+    assert result.status == 0, f'Hammer Ping failed:\n{result.stderr}'
+
+    # attempt to use invalid certificates
+    result = satellite.install(
+        InstallerCommand(
+            'certs-update-server',
+            'certs-update-server-ca',
+            scenario='satellite',
+            certs_server_cert='/root/certs/invalid.crt',
+            certs_server_key='/root/certs/invalid.key',
+            certs_server_ca_cert=f'"/root/{cert_data["ca_bundle_file_name"]}"',
+        )
+    )
+    # installer should fail with non-zero value
+    assert result.status != 0
+    assert "verification failed" in result.stdout
+
+    result = satellite.execute('hammer ping')
+    assert result.status == 0, f'Hammer Ping failed:\n{result.stderr}'


tests/foreman/destructive/test_installer.py
```diff
@@ -140,3 +140,41 @@ def test_positive_installer_certs_regenerate(target_sat):
...
+
+    # check if satellite is running
+    result = satellite.execute('hammer ping')
+    assert result.status == 0, f'Hammer Ping failed:\n{result.stderr}'
+
+    # attempt to use invalid certificates
+    command = (
+        'satellite-installer --scenario satellite '
+        f'--certs-server-cert "/root/certs/invalid.crt" '
+        f'--certs-server-key "/root/certs/invalid.key" '
```
JacobCallahan, line 171:
```suggestion
        '--certs-server-cert "/root/certs/invalid.crt" '
        '--certs-server-key "/root/certs/invalid.key" '
```

JacobCallahan (changes requested):
You can also use the InstallerCommand class from robottelo.utils.installer instead of constructing the command by string. This isn't a blocker suggestion, though.
https://github.com/SatelliteQE/robottelo/blob/master/robottelo/utils/installer.py#L4

rmynar:
trigger: test-robottelo
pytest: tests/foreman/installer/test_installer.py -k 'test_handle_invalid_certificate'

rmynar:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_installer.py -k 'test_handle_invalid_certificate'

tests/foreman/destructive/test_installer.py
```diff
@@ -140,3 +140,43 @@ def test_positive_installer_certs_regenerate(target_sat):
...
+
+    # attempt to use invalid certificates
+    result = satellite.install(
+        InstallerCommand(
+            'certs-update-server',
+            'certs-update-server-ca',
+            scenario='satellite',
+            certs_server_cert='"/root/certs/invalid.crt"',
+            certs_server_key='"/root/certs/invalid.key"',
+            certs_server_ca_cert=f'"/root/{cert_data["ca_bundle_file_name"]}"',
```
JacobCallahan, line 175: now that we're not composing a larger string, we should be able to simplify the quotes
```suggestion
            certs_server_cert='/root/certs/invalid.crt',
            certs_server_key='/root/certs/invalid.key',
            certs_server_ca_cert=f'/root/{cert_data["ca_bundle_file_name"]}',
```

JacobCallahan (changes requested):
very close! one minor change

tests/foreman/destructive/test_installer.py
rmynar: We can remove the pair of quotes around constant strings, however I would like to keep the quotes around _ca_cert_. Otherwise the command could be broken if _ca_cert_ filename contains space(s) - it's inprobable, but possible.

JacobCallahan (approved):

tests/foreman/destructive/test_installer.py
```diff
@@ -140,3 +140,43 @@ def test_positive_installer_certs_regenerate(target_sat):
...
+
+    # attempt to use invalid certificates
+    result = satellite.install(
+        InstallerCommand(
+            'certs-update-server',
+            'certs-update-server-ca',
+            scenario='satellite',
+            certs_server_cert='/root/certs/invalid.crt',
+            certs_server_key='/root/certs/invalid.key',
+            certs_server_ca_cert=f'"/root/{cert_data["ca_bundle_file_name"]}"',
```
Gauravtalreja1, line 175:
```suggestion
            certs_server_ca_cert=f'/root/{cert_data["ca_bundle_file_name"]}',
```

rmynar: I already discussed that with @JacobCallahan. Removing these quotes might lead to broken installer command if _ca_bundle_file_name_ contains space(s).

tests/foreman/destructive/test_installer.py
```diff
@@ -140,3 +140,43 @@ def test_positive_installer_certs_regenerate(target_sat):
...
+        )
+    )
+    # installer should fail with non-zero value
+    assert result.status != 0
```
ogajduse, line 179: Can you add an extra check after this assertion to check that STDOUT/STDERR contains the expected failure message? That would ensure that if the installer failed, it did not fail on a different error.

rmynar: That seems reasonable. I added a check for "verification failed". Let's check...

tests/foreman/destructive/test_installer.py
lhellebr: I agree with Radek, these quotes are appropriate.

ColeHiggins2 (approved):
ack pending comments

ogajduse:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_installer.py -k 'test_handle_invalid_certificate'

tests/foreman/destructive/test_installer.py
```diff
@@ -179,3 +179,44 @@ def test_positive_installer_puma_worker_count(target_sat):
             assert f'cluster worker {i}' in result.stdout
         else:
             assert f'cluster worker {i}' not in result.stdout
+
+
+def test_handle_invalid_certificate(cert_setup_destructive_teardown):
```
pondrejk, line 184: to be consistent with naming around robottelo I suggest `test_negative_handle...`

pondrejk (approved):
small suggestion

tests/foreman/destructive/test_installer.py
rmynar: Ok, let it be `negative`...

ogajduse (approved):
PRT passed, and comments were addressed. Merging

PR #13716, opened 2024-01-10
[6.15.z] Removing changes associated with Bugbear rule B019

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/13715

This PR reverts the change from lru_cache to cachedmethod. The primary goal of the original change was to avoid memory leaks, due to lru_cache holding instances of a class alive longer than they should be.
However, for our test framework, the workaround to avoid these memory leaks really aren't worth it.

I also added an lru_cache to cli_factory.__getattr__, improving its recurring lookup time substantially.

Finally, I corrected a type check that was erroneously converted in the initial pass.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,6 +32,7 @@ select = [
 ]
 
 ignore = [
+    "B019",  # lru_cache can lead to memory leaks - acceptable tradeoff
     "E501",  # line too long - handled by black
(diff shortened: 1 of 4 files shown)


Satellite-QE:
trigger: test-robottelo
pytest: tests/foreman/cli -k test_host_registration_end_to_end

JacobCallahan (approved):

PR #13754, opened 2024-01-11
Adding fix for discoveryrule UI

adarshdubey-star:
Going with the API way for provisioning as the host created is a fake discovered host and its provisioning invokes a warning on the UI, as the subnet is randomly generated.

### DIFF
--- a/tests/foreman/ui/test_discoveryrule.py
+++ b/tests/foreman/ui/test_discoveryrule.py
@@ -210,6 +210,8 @@ def test_positive_list_host_based_on_rule_search_query(
     target_sat.api_factory.create_discovered_host(options={'physicalprocessorcount': cpu_count + 1})
     provisioned_host_name = f'{host.domain.read().name}'
     with session:
+        session.organization.select(org_name=module_org.name)
+        session.location.select(loc_name=module_location.name)
         values = session.discoveryrule.read_all()
         assert discovery_rule.name in [rule['Name'] for rule in values]
         values = session.discoveryrule.read_discovered_hosts(discovery_rule.name)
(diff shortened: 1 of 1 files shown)


adarshdubey-star:
trigger: test-robottelo
pytest: tests/foreman/ui/test_discoveryrule.py -k test_positive_list_host_based_on_rule_search_query

synkd (approved):

tests/foreman/ui/test_discoveryrule.py
```diff
@@ -218,8 +220,8 @@ def test_positive_list_host_based_on_rule_search_query(
         assert values['table'][0]['IP Address'] == ip_address
         assert values['table'][0]['CPUs'] == str(cpu_count)
         # auto provision the discovered host
-        session.discoveredhosts.apply_action('Auto Provision', [discovered_host['name']])
-        assert not session.discoveredhosts.search('name = "{}"'.format(discovered_host['name']))
+        result = target_sat.api.DiscoveredHost(id=discovered_host['id']).auto_provision()
+        assert f'provisioned with rule {discovery_rule.name}' in result['message']
```
Gauravtalreja1, line 224: why changing this to API in UI module?

adarshdubey-star: No, we shouldn't be doing it, because we have real hosts as well on which  I want that warning to be caught from airgun if there is a flaw. Also, we are testing the UI auto_provision in other places as well , so its just a simple change.

shweta83: @adarshdubey-star Can you please share any test where we are testing the UI interface for auto_provision

adarshdubey-star: https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/ui/test_discoveredhost.py#L152

Gauravtalreja1 (approved):

PR #13821, opened 2024-01-18
[6.13.z] Random Fixes in pytest plugins

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/13820

### Problem Statement
1. In factory collection plugin, The factory instance marker was being added to test cases using `satellite_factory` and `caspule_factory` fixtures, which is anyways being used in all test cases through target_sat hence all the tests were being collected/discollected.
2. In reqs plugin, the message to update the reqs was wrong.

### Solution
1. Now, the `satellite_factory` and `caspule_factory` fixtures would be ignored from factory fixtures and rest would be taken into account to filter the tests.
2. Updated the message to show the right messsage.


Satellite-QE:
trigger: test-robottelo

PR #13856, opened 2024-01-22
cli oscap fix name already taken

pondrejk:
### Problem Statement
some tests were relying on the default factory title for the newly created scap contents, which caused a bunch of 'name already taken' failures for parametrized tests

### Solution
just create the title within the test body


pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_oscap.py -k test_positive_delete_scap_content_with_id

Gauravtalreja1 (approved):

PR #12989, opened 2023-10-26
removing the from nailgun import entities and moving target_sat, module_target_sat

omkarkhatavkar:
#### Key Changes:

- **Direct Imports Removed**: We've cleaned up our codebase by eliminating direct imports from the nailgun package.

- **Using target_sat, module_target_sat, class_target_sat**: Modified the API calls and using the `target_sat.api.` pattern for all API calls.

PR #13871, opened 2024-01-23
fix in test_permission

pondrejk:
### Problem Statement
https://github.com/SatelliteQE/robottelo/pull/12989 swapped `entities` for `target_sat.api` but:

```
(Pdb) entities.ActivationKey
<class 'nailgun.entities.ActivationKey'>
(Pdb) target_sat.api.ActivationKey
<class 'robottelo.hosts.Satellite.api.<locals>.inject_config.<locals>.DecClass'>
```
which breaks the if condition in this test

### Solution

Swapping back to entities, as it is still heavily used in that module. Deferring further changes to component evaluation

### DIFF
--- a/tests/foreman/api/test_permission.py
+++ b/tests/foreman/api/test_permission.py
@@ -374,7 +374,7 @@ class TestUserRole:
         self.give_user_permission(_permission_name(entity_cls, 'update'), target_sat)
         # update() calls read() under the hood, which triggers
         # permission error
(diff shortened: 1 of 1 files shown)


pondrejk:
trigger: test-robottelo
pytest: tests/foreman/api/test_permission.py -k test_positive_check_update[2]

ogajduse (approved):

PR #13902, opened 2024-01-25
Add host teardown for smart classparameters test

shubhamsg199:
### Problem Statement
test_positive_test_hidden_parameter_value is failing due to no teardown present for the host in test_positive_list

### Solution
Add teardown for the host deletion

### DIFF
--- a/tests/foreman/cli/test_classparameters.py
+++ b/tests/foreman/cli/test_classparameters.py
@@ -66,6 +66,7 @@ class TestSmartClassParameters:
     @pytest.mark.e2e
     def test_positive_list(
         self,
+        request,
         session_puppet_enabled_sat,
         module_puppet_org,
         module_puppet_loc,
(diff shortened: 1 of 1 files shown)


Gauravtalreja1 (approved):
ACK for the teardown change, @shubhamsg199 could you update commit message/PR title to some meaningful message, and run PRT here please?

lhellebr (approved):
Have you considered making a host a fixture instead?

ColeHiggins2 (approved):

rplevka (changes requested):
Thanks for fixing the test!
I agree with @lhellebr here.
Let's follow the practice of separating the environment setup from the test itself by moving the host creation to a fixture.
It seems pretty straightforward and should not take too much effort.

shubhamsg199:
trigger: test-robottelo
pytest: tests/foreman/cli/test_classparameters.py

shubhamsg199:
> Thanks for fixing the test! I agree with @lhellebr here. Let's follow the practice of separating the environment setup from the test itself by moving the host creation to a fixture. It seems pretty straightforward and should not take too much effort.

@rplevka The host creation here is  just used by one test so I would rather keep it as is and can add the fixture if needed in future

Gauravtalreja1 (approved):
ACK, Moving host setup and teardown to separate fixture isn't required atm I feel

PR #13923, opened 2024-01-29
Remove BZ check which is migrated to Jira and Add new BZ check causing selinux denials

shweta83:
### Problem Statement
The BZ in the test case has been migrated to jira so the check doesn't work.
The older BZ is entirely unrelated to the test  case so removed it but we ran into another issue and raised a BZ for it -> BZ#2263294
### Solution
Removing the check from the test.

### DIFF
--- a/tests/foreman/sys/test_pulp3_filesystem.py
+++ b/tests/foreman/sys/test_pulp3_filesystem.py
@@ -27,15 +27,13 @@ def test_selinux_status(target_sat):
 
     :expectedresults: SELinux is enabled and there are no denials
 
-    :customerscenario: true
-
-    :BZ: 2131031
+    :BZ: 2263294
     """
     # check SELinux is enabled
     result = target_sat.execute('getenforce')
     assert 'Enforcing' in result.stdout
     # check there are no SELinux denials
-    if not is_open('BZ:2131031'):
+    if not is_open('BZ:2263294'):
         result = target_sat.execute('ausearch --input-logs -m avc -ts today --raw')
         assert result.status == 1, 'Some SELinux denials were found in journal.'
 


shweta83:
trigger: test-robottelo
pytest: tests/foreman/sys/test_pulp3_filesystem.py -k test_selinux_status

shweta83:
The test is failing with some unknown selinux denials and it is in discussion, so moving this PR to draft until the issue gets resolved.

shweta83:
> > The test is failing with some unknown selinux denials and it is in discussion, so moving this PR to draft until the issue gets resolved.
> 
> Since this has been migrated to Jira, has the Jira issue been resolved? If not, then we'll need to block the test logic on that issue. This is something that @jameerpathan111 has volunteered to take on.

The issue migrated to Jira is closed and it is a different issue from the test. I can block this issue but we need a new bug for it.

tests/foreman/sys/test_pulp3_filesystem.py
```diff
@@ -29,15 +28,14 @@ def test_selinux_status(target_sat):
 
     :customerscenario: true
 
-    :BZ: 2131031
+    :BZ: 2263294
```
Gauravtalreja1, line 31:
```suggestion
    :BZ: 2131031, 2263294
```

tests/foreman/sys/test_pulp3_filesystem.py
```diff
@@ -16,9 +16,8 @@
 
 import pytest
 
-from robottelo.utils.issue_handlers import is_open
-
 
+@pytest.mark.skip_if_open("BZ:2263294")
```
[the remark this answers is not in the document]
shweta83: The test case is not doing much. So I added the skip for entire test

tests/foreman/sys/test_pulp3_filesystem.py
shweta83: The remove BZ is migrated and also the bug is unrelated the test case

tests/foreman/sys/test_pulp3_filesystem.py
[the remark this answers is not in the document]
Gauravtalreja1: yes, I get it, but even if there's a single step, I'd not skip the test

tests/foreman/sys/test_pulp3_filesystem.py
Gauravtalreja1: It looks related to selinux only, and this is ancient test, so we could either keep it but its upto you

shweta83: I would still say that the bug is unrelated to the test so would like to remove the BZ. It would make confusion in future to relate the bug with the test case.

tests/foreman/sys/test_pulp3_filesystem.py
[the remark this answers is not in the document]
shweta83: No problem, Updated.

shweta83:
trigger: test-robottelo
pytest: tests/foreman/sys/test_pulp3_filesystem.py -k test_selinux_status

Gauravtalreja1 (approved):

PR #13817, opened 2024-01-17
ensure mqtt capsule is not bypassed

pondrejk:
### Problem Statement
With global proxy setting on, it is possible that if the mqtt capsule is not reached, the default one is used instead. In such case we wouldn't find out there is a problem. 

### Solution
This PR forces the registered through capsule for the tests using external capsule

PR #13931, opened 2024-01-30
[6.15.z] ensure mqtt capsule is not bypassed

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/13817

### Problem Statement
With global proxy setting on, it is possible that if the mqtt capsule is not reached, the default one is used instead. In such case we wouldn't find out there is a problem. 

### Solution
This PR forces the registered through capsule for the tests using external capsule


Satellite-QE:
trigger: test-robottelo
pytest: tests/foreman/cli/test_remoteexecution.py -k test_positive_run_job_on_host_registered_to_pull_provider[rhel9-no_global_proxy]

PR #13947, opened 2024-01-31
making dependency_merge.yml same as other z-stream branches

omkarkhatavkar:
### Problem Statement
as dependency_merge.yml using same configuration as the master causing failure in the automerge.   

### Solution
making dependency_merge.yml same as other z-stream branches

### Related Issues
https://github.com/SatelliteQE/robottelo/actions/runs/7720921376

### DIFF
--- a/.github/workflows/dependency_merge.yml
+++ b/.github/workflows/dependency_merge.yml
@@ -1,46 +1,32 @@
-name: Dependabot Auto Merge
-on: pull_request
-
-permissions:
-  pull-requests: write
+name: Dependabot Auto Merge - ZStream
+on:
+  pull_request_target:
+    branches-ignore:
+      - master
 
 jobs:
   dependabot:
     name: dependabot-auto-merge
     runs-on: ubuntu-latest
-    if: github.event.pull_request.user.login == 'dependabot[bot]'
+    if: |
+      contains(github.event.pull_request.labels.*.name, 'dependencies')
+
     steps:
-      - name: Dependabot metadata
-        id: metadata
-        uses: dependabot/fetch-metadata@v1
+      - id: find-prt-comment
+        name: Find the prt comment
+        uses: peter-evans/find-comment@v2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency_merge.yml
```diff
@@ -51,10 +37,10 @@ jobs:
 
       - name: Wait for other status checks to Pass
         id: waitforstatuschecks
-        uses: lewagon/wait-on-check-action@v1.3.3
+        uses: lewagon/wait-on-check-action@v1.3.1
```
Gauravtalreja1, line 40: Can we keep it latest?
```suggestion
        uses: lewagon/wait-on-check-action@v1.3.3
```

.github/workflows/dependency_merge.yml
```diff
@@ -51,10 +37,10 @@ jobs:
...
         with:
           ref: ${{ github.head_ref }}
-          repo-token: ${{ secrets.CHERRYPICK_PAT }}
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
Gauravtalreja1, line 43: is this change `s/CHERRYPICK_PAT/GITHUB_TOKEN` intended?

.github/workflows/dependency_merge.yml
```diff
@@ -1,46 +1,32 @@
...
+on:
+  pull_request_target:
+    branches-ignore:
+      - master
 
 jobs:
   dependabot:
     name: dependabot-auto-merge
     runs-on: ubuntu-latest
-    if: github.event.pull_request.user.login == 'dependabot[bot]'
```
Gauravtalreja1, line 11: can't we just add OR condition to check user is either  dependabot or Satellite bot here instead?

.github/workflows/dependency_merge.yml
omkarkhatavkar: making as uniform as https://github.com/SatelliteQE/robottelo/blob/6.14.z/.github/workflows/dependency_merge.yml please look at it

PR #13554, opened 2023-12-22
bz2220965

lhellebr:
### Problem Statement
https://bugzilla.redhat.com/show_bug.cgi?id=2220965

### Solution
We need to change the UI test sesison fixture to be able to run with non-admin user.

PR #13962, opened 2024-01-31
fix in ui_host fixture

pondrejk:
### Problem Statement
https://github.com/SatelliteQE/robottelo/pull/13554 breaks every unparametrized session fixture with `AttributeError: 'SubRequest' object has no attribute 'param'`

### Solution

request.param is not expected at all times


lhellebr (approved):
LGTM. I thought I tested for this case but obviously not well enough. What test did this fail with? We should run PRT with it.

rplevka (approved):
thanks for catching that

ogajduse (approved):

PR #13970, opened 2024-02-01
setup_org_for_a_rh_repo fix

lhellebr:
### Problem Statement
```
AttributeError: 'Manifest' object has no attribute 'path'
```

### Solution
Follow idiom of creating a temporary file and into it, copying an in-memory object, then using it as a file.

I'm not sure how and when it got broken, I hit it when looking at the results of `test_negative_without_attach_with_lce`. I haven't yet tested extensively to make sure this doesn't break function `_setup_org_for_a_rh_repo` in other cases but it seems like it couldn't have worked.

### DIFF
--- a/robottelo/host_helpers/cli_factory.py
+++ b/robottelo/host_helpers/cli_factory.py
@@ -726,10 +726,14 @@ class CLIFactory:
         # If manifest does not exist, clone and upload it
         if len(self._satellite.cli.Subscription.exists({'organization-id': org_id})) == 0:
             with clone() as manifest:
-                self._satellite.put(manifest.path, manifest.name)
+                remote_name = gen_alpha()
+                _, temporary_local_manifest_path = mkstemp(prefix='manifest-', suffix='.zip')
+                with open(temporary_local_manifest_path, 'wb') as file_handler:
+                    file_handler.write(manifest.content.read())
+                    self._satellite.put(temporary_local_manifest_path, remote_name)
             try:
                 self._satellite.cli.Subscription.upload(
-                    {'file': manifest.name, 'organization-id': org_id}
+                    {'file': remote_name, 'organization-id': org_id}
                 )
             except CLIReturnCodeError as err:
                 raise CLIFactoryError(f'Failed to upload manifest\n{err.msg}') from err
--- a/robottelo/host_helpers/cli_factory.py
+++ b/robottelo/host_helpers/cli_factory.py
@@ -726,17 +726,7 @@ class CLIFactory:
         # If manifest does not exist, clone and upload it
         if len(self._satellite.cli.Subscription.exists({'organization-id': org_id})) == 0:
             with clone() as manifest:
-                remote_name = gen_alpha()
-                _, temporary_local_manifest_path = mkstemp(prefix='manifest-', suffix='.zip')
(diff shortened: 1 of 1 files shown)


robottelo/host_helpers/cli_factory.py
```diff
@@ -726,10 +726,14 @@ def _setup_org_for_a_rh_repo(self, options=None):
         # If manifest does not exist, clone and upload it
         if len(self._satellite.cli.Subscription.exists({'organization-id': org_id})) == 0:
             with clone() as manifest:
-                self._satellite.put(manifest.path, manifest.name)
+                remote_name = gen_alpha()
```
Gauravtalreja1, line 729: why not use upload_manifest helper here instead?
```
   with clone() as manifest:
        self._satellite.upload_manifest(org.id, manifest.content)
```

lhellebr:
Regression test, on a different test that uses `setup_org_for_a_rh_repo` and didn't crash with the `AttributeError: 'Manifest' object has no attribute 'path'` error before:
```
pytest tests/foreman/cli/test_activationkey.py::test_positive_create_content_and_check_enabled
```
It successfully fails the same way it fails without this PR, a few lines after a call for `setup_org_for_a_rh_repo` .

robottelo/host_helpers/cli_factory.py
lhellebr: Of course, that's much better. Thanks!

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_activationkey.py::test_positive_add_redhat_and_custom_products tests/foreman/cli/test_host.py::test_negative_without_attach_with_lce tests/foreman/cli/test_reporttemplates.py::test_positive_generate_hostpkgcompare

Gauravtalreja1:
@lhellebr Could you check the failing PRT results?

Gauravtalreja1 (approved):

robottelo/host_helpers/cli_factory.py
```diff
@@ -726,13 +726,7 @@ def _setup_org_for_a_rh_repo(self, options=None):
...
         if len(self._satellite.cli.Subscription.exists({'organization-id': org_id})) == 0:
             with clone() as manifest:
-                self._satellite.put(manifest.path, manifest.name)
-            try:
-                self._satellite.cli.Subscription.upload(
-                    {'file': manifest.name, 'organization-id': org_id}
-                )
-            except CLIReturnCodeError as err:
-                raise CLIFactoryError(f'Failed to upload manifest\n{err.msg}') from err
+                self._satellite.upload_manifest(org_id, manifest.content)
```
Gauravtalreja1, line 729: @lhellebr I missed this earlier, similar changes are required for `setup_org_for_a_rh_repo` helper, could you update it too?

lhellebr: Can we leave it for another PR? This PR is part of CE effort so I'm just doing what is required for that, changing as little as possible to minimize possibility of breaking stuff.

Gauravtalreja1: @lhellebr Sure, please make a note of it or create a issue for it for future, and when you raise a PR to fix it, let me know, thank you!

PR #13979, opened 2024-02-02
preserving PRT result as part of github comment & PRT pass/fail label

omkarkhatavkar:
### Problem Statement
Currently, the PRT result gets clean after a new commit and there is no way for PR to know what happened with the result.  PRT passed/failed labels also need to be added automatically this helps in reviewing the PR quickly and merging them 

### Solution
Add the dispatch workflow that takes care of this thing from the Jenkins CI update the GitHub comment with all details and apply the filter.  

### Testing 
I was able to test this on the my fork of robottelo here https://github.com/omkarkhatavkar/robottelo/actions/runs/7755989443/job/21152444058

### DIFF
--- a/.github/workflows/prt_result.yml
+++ b/.github/workflows/prt_result.yml
@@ -1,0 +1,84 @@
+### The prt result workflow triggered through dispatch request from CI
+name: post-prt-result
+
+# Run on workflow dispatch from CI
+on:
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        type: string
+        description: pr number for PRT run
+      build_number:
+        type: string
+        description: build number for PRT run
+      pytest_result:
+        type: string
+        description: pytest summary result line
+      build_status:
+        type: string
+        description: status of jenkins build e.g. success, unstable or error
+      prt_comment:
+        type: string
+        description: prt pytest comment triggered the PRT checks
+
+
+jobs:
+  post-the-prt-result:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Add last PRT result into the github comment
+        id: add-prt-comment
+        if: ${{ always() && github.event.inputs.pytest_result != '' }}
+        uses: thollander/actions-comment-pull-request@v2
+        with:
+          message: |
+            **PRT Result**
+            ```
+            Build Number: ${{ github.event.inputs.build_number }}
+            Build Status: ${{ github.event.inputs.build_status }}
+            PRT Comment: ${{ github.event.inputs.prt_comment }}
+            Test Result : ${{ github.event.inputs.pytest_result }}
+            ```
+          pr_number: ${{ github.event.inputs.pr_number }}
+          GITHUB_TOKEN: ${{ secrets.CHERRYPICK_PAT }}
+
+      - name: Add the PRT passed/failed labels
+        id: prt-status
+        if: ${{ always() && github.event.inputs.build_status != '' }}
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.CHERRYPICK_PAT }}
+          script: |
+            const prNumber = ${{ github.event.inputs.pr_number }};
+            const buildStatus = "${{ github.event.inputs.build_status }}";
+            const labelToAdd = buildStatus === "SUCCESS" ? "PRT-Passed" : "PRT-Failed";
+            github.rest.issues.addLabels({
+              issue_number: prNumber,
+              owner: context.repo.owner,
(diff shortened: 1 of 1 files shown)


synkd (commented):
This is a very useful addition to the Robottelo GH actions. I think it's a great idea to preserve the history of PRT requests in a PR's comments. However, I do have a few questions about your test results in https://github.com/omkarkhatavkar/robottelo/pull/158:

- In that PR, the build number for each run is reported as `10`. Is this due to repeatedly retriggering the same check from the same PRT run?
- There's some inconsistency in the reported build status, with a mix of SUCCESS, FAILURE, and UNSTABLE reported for a build with a failed test. It looks like your most recent testing has converged on this result being reported as UNSTABLE, which I think is fine. But would it be possible to see results from a build showing a status of SUCCESS when all tests have passed?
- Which user account(s) will post the results comment to PRs when this is merged into Robottelo?

omkarkhatavkar:
> This is a very useful addition to the Robottelo GH actions. I think it's a great idea to preserve the history of PRT requests in a PR's comments. However, I do have a few questions about your test results in [omkarkhatavkar#158](https://github.com/omkarkhatavkar/robottelo/pull/158):
 
Thanks for seeing value in this PR (Contribution)

> * In that PR, the build number for each run is reported as `10`. Is this due to repeatedly retriggering the same check from the same PRT run?

I was testing this that's why it would have repeatedly shown build number 10, In real case it will be the build number coming from the Jenkins PRT run.  

> * There's some inconsistency in the reported build status, with a mix of SUCCESS, FAILURE, and UNSTABLE reported for a build with a failed test. It looks like your most recent testing has converged on this result being reported as UNSTABLE, which I think is fine. But would it be possible to see results from a build showing a status of SUCCESS when all tests have passed?

I was just testing some dummy data, But the thumb rule is that if the test passes without any issues and the build is a Success it will present the status as **PRT Passed**, else **PRT Failed**. Adding more color code it will look bit better    


> * Which user account(s) will post the results comment to PRs when this is merged into Robottelo?
The result will be posted by the token which I'm using here which is SatelliteQE user.

omkarkhatavkar:
https://github.com/omkarkhatavkar/robottelo/pull/158#issuecomment-1927023967 This one the last comment by me. I think this looks much better as well

synkd (approved):

.github/workflows/prt_result.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: Add the PRT passed/failed labels
+        id: prt-status
+        if: ${{ always() && github.event.inputs.build_status != '' }}
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.CHERRYPICK_PAT }}
+          script: |
+            const prNumber = ${{ github.event.inputs.pr_number }};
+            const buildStatus = "${{ github.event.inputs.build_status }}";
```
jyejare, line 53:
```suggestion
            const buildStatus = ${{ github.event.inputs.build_status }};
```

jyejare (approved):
Looks Great! Some nitpick comments.

.github/workflows/prt_result.yml
omkarkhatavkar: We need to do this in string format. I was facing issues if not having in the String.

jyejare:
I have an ACK.

omkarkhatavkar:
@jyejare Can we merge this beast ?

.github/workflows/prt_result.yml
```diff
@@ -0,0 +1,83 @@
...
+            const labelToAdd = buildStatus === "SUCCESS" ? "PRT-Passed" : "PRT-Failed";
+            github.rest.issues.addLabels({
+              issue_number: prNumber,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              labels: [labelToAdd]
+            });
+      - name: Remove failed label on test pass or vice-versa
```
jyejare, line 62: Last comment, can we also do removal of label using CHERRYPICK_PAT token , i.e from SATELLITEQE user ?

Else we will see as SatelliteQE user adding the label and Github Action removing the label. 

Not a blocking comment, but just a nitpick comment !

jyejare (commented):
ACK pending comment !

omkarkhatavkar:
@jyejare Could you please create labels for `"PRT-Passed"` and `"PRT-Failed"`? Please choose some nice colours to represent them.

jyejare:
@omkarkhatavkar Added the labels for both! Now we should be good to merge if no complaints !!

PR #13994, opened 2024-02-05
Errata coverage for CV Incremental Update

damoore044:
### Purpose
- Add coverage for incremental update, that it is available when expected, and 
- that the expected host's content-view and environment(s) are are updated as expected, and
- the available errata/packages to host are updated by the incremental version as expected.
- we can install updated packages and see the expected (filtered-security only) applicable errata/count drops, and 
the applicable packages count drops.
- modification to helper method `cv_publish_promote`, which did not always filter environment ids for promotion as expected.
### Coverage reason
We already have some testing for if any incremental update is needed on a single view/environment, `test_positive_incremental_update_required`:
- But the existing test doesn't check the successful update and new CV version itself (contained errata/content). Or how the registered host's applicability is affected by the new version.
### PRT Case
```
trigger: test-robottelo
```

### DIFF
--- a/tests/foreman/api/test_errata.py
+++ b/tests/foreman/api/test_errata.py
@@ -19,6 +19,7 @@ import pytest
 from robottelo.config import settings
 from robottelo.constants import (
     DEFAULT_ARCHITECTURE,
+    DEFAULT_CV,
     FAKE_1_CUSTOM_PACKAGE,
     FAKE_2_CUSTOM_PACKAGE,
     FAKE_2_CUSTOM_PACKAGE_NAME,
@@ -130,6 +131,11 @@ def _validate_errata_counts(host, errata_type, expected_value, timeout=120):
                 expected_value,
             )
         )
+
+
+def _fetch_library_environment_for_org(sat, org):
+    search = {'search': f'name="Library" and organization_id={org.id}'}
+    return sat.api.LifecycleEnvironment().search(query=search)[0].read()
 
 
 def _fetch_available_errata(host, expected_amount=None, timeout=120):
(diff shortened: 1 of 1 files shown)


damoore044:
trigger: test-robottelo

damoore044:
trigger: test-robottelo

tests/foreman/api/test_errata.py
```diff
@@ -1430,3 +1464,247 @@ def rh_repo_module_manifest(module_sca_manifest_org, module_target_sat):
...
+
+    # create multiple linked environments
+    for n in range(number_of_lces):
+        new_lce = target_sat.api.LifecycleEnvironment(
+            organization=module_sca_manifest_org,
+            prior=lce_list[n],
+        ).create()
+        lce_list.append(new_lce)
+    assert len(lce_list) == number_of_lces + 1
+    # colelct default CV for org
```
sambible, line 1528:
```suggestion
    # collect default CV for org
```

tests/foreman/api/test_errata.py
```diff
@@ -1430,3 +1464,247 @@ def rh_repo_module_manifest(module_sca_manifest_org, module_target_sat):
...
+    host = rhel8_contenthost.nailgun_host.read()
+    response = target_sat.api.Host().bulk_available_incremental_updates(
+        data={
+            'organization_id': module_sca_manifest_org.id,
+            'included': {'ids': [host.id]},
+            'errata_ids': FAKE_9_YUM_SECURITY_ERRATUM,
+        },
+    )
+    assert (
+        not response
```
sambible, line 1599: Is there a particular reason you are asserting this here? Would it be better to assert what you actually expect to be returned? I'd think you'd know how many updates should be available.

sambible (changes requested):
Small comment change, and a question, but overall looks like a through test. Good work.

tests/foreman/api/test_errata.py
damoore044: right, if there are no incremental updates, response will be just `[]`, I can change to that

damoore044:
trigger: test-robottelo

damoore044:
trigger: test-robottelo

damoore044:
trigger: test-robottelo

damoore044:
trigger: test-robottelo

tests/foreman/api/test_errata.py
```diff
@@ -1430,3 +1468,253 @@ def rh_repo_module_manifest(module_sca_manifest_org, module_target_sat):
...
+    host_app_packages = rhel8_contenthost.applicable_package_count
+    response = target_sat.api.Host().bulk_available_incremental_updates(
+        data={
+            'organization_id': module_sca_manifest_org.id,
+            'included': {'ids': [host.id]},
+            'errata_ids': FAKE_9_YUM_SECURITY_ERRATUM,
+        },
+    )
+    assert (
+        response
```
vijaysawant, line 1631: Can't we check empty list here?
```suggestion
        response != []
```

tests/foreman/api/test_errata.py
```diff
@@ -458,18 +496,18 @@ def cv_publish_promote(sat, org, cv, lce=None, needs_publish=True):
     return _result
 
 
-def _publish_and_wait(sat, org, cv):
+def _publish_and_wait(sat, org, cv, search_rate=5, max_tries=12):
```
vijaysawant, line 499: - Though function `_publish_and_wait` is for internal use only, I would suggest to use value `search_rate=1` & `max_tries=10` as per reference at L29  in [host_helplers/capsule_mixins.py::CapsuleInfo.wait_for_tasks](https://github.com/SatelliteQE/robottelo/blob/master/robottelo/host_helpers/capsule_mixins.py#L29) 
- You can pass expected values from function call like this 
`_publish_and_wait(sat, org, cv, search_rate=5, max_tries=12)`
- I don't understant how to explain the scenario but we can consider best practices, coding standards and relate here.
- Optional, non blocking.

vijaysawant (changes requested):
suggested few changes

tests/foreman/api/test_errata.py
vijaysawant: I believ you have already addressed the @sambible 's comment, I am resolving this conversation.

tests/foreman/api/test_errata.py
damoore044: In prior reviews from Jake I've been told that it's good practice to just `assert response/result`, instead of asserting != specific empty value. @JacobCallahan any thoughts here?

tests/foreman/api/test_errata.py
damoore044: Thanks! I've implemented it, to properly default to expected `search_rate=1, max_tries=10`,
if no other values were to be passed, as referenced by L29 in the link you sent.

damoore044:
trigger: test-robottelo

tests/foreman/api/test_errata.py
damoore044: I'm thinking of the possibility of the response being `None`, if there were some unexpected issue, 
then `assert None != []` would pass, and some ambiguous value error would probably be had in an assertion below?

damoore044:
trigger: test-robottelo

damoore044:
trigger: test-robottelo

tests/foreman/api/test_errata.py
vijaysawant: Cool, thank @damoore044

tests/foreman/api/test_errata.py
JacobCallahan: correct, since we are expecting some kind of populated list, then a simple boolean assertion on the list itself should be fine.

vijaysawant: I will leave it upto you how to aseert, it is just an api call (`bulk_available_incremental_updates`) and output/response can assert with any recieved values inside dict.

vijaysawant:
Build 5652 failed due to broken setup
`failed on setup with "ModuleNotFoundError: No module named 'docker'"`

damoore044:
trigger: test-robottelo

vijaysawant (approved):

damoore044:
### Build [5655]: Pass
Only change since was rebasing of an already included CP, #14020 that fixed a broker issue, no code changes.

tests/foreman/api/test_errata.py
```diff
@@ -414,40 +420,72 @@ def package_applicability_changed_as_expected(
...
+                lce_ids = [lce.id]
         else:
-            lce_ids = [_lce.id for _lce in lce]
-
+            # a list of ids or instances was passed
+            if isinstance(lce[0], int):
+                lce_ids = lce
+            else:
+                # will fail here if list of lces is not int or has no id
+                lce_ids = [env.id for env in lce]
```
Gauravtalreja1, line 453:
```suggestion
            lce_ids = lce if isinstance(lce[0], int) else [env.id for env in lce]
```

damoore044:
trigger: test-robottelo

tests/foreman/api/test_errata.py
damoore044: I implemented this, and the same logic for the non-list case above!

damoore044:
trigger: test-robottelo

ColeHiggins2 (approved):
Ack pending @sambible approval

sambible (approved):
looks good, lgtm

PR #14017, opened 2024-02-08
Force keyscan to use ipv4

pondrejk:
### Problem Statement
a bunch or templatesync tests failing with `Host key verification failed` when exporting templates to gitea. Turns out that ssh-keyscan opts for ipv6 connenction, where it gets no response

### Solution
Force ssh-keyscan to go via ipv6. Also updated filter to use existing template

### DIFF
--- a/pytest_fixtures/component/templatesync.py
+++ b/pytest_fixtures/component/templatesync.py
@@ -66,7 +66,7 @@ def git_pub_key(session_target_sat, git_port):
     id = res.json()['id']
     # add ssh key to known host
     session_target_sat.execute(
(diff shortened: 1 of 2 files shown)


pondrejk:
trigger: test-robottelo
pytest: tests/foreman/cli/test_templatesync.py -k test_positive_update_templates_in_git[non_empty_repo-ssh]

jyejare:
@pondrejk Did we know/find the root cause whats making `ssh-keyscan` to use IPv6 and from when ? Asking since we are making changes into Infra that should/shouldnt impact the test failures !

ColeHiggins2 (approved):
Ack pending comments

jyejare:
Yup, I was doubtful about hence thought to go for root cause.

PR #14020, opened 2024-02-08
Temporarily add docker and paramiko deps

JacobCallahan:
Since we're currently installing broker from github, we can't install the extra deps. Until it is back, we need to explicitly add in docker and its ssh support layer paramiko.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,11 +4,13 @@ betelgeuse==1.11.0
 # broker[docker]==0.4.1 - Temporarily disabled, see below
 cryptography==42.0.2
 deepdiff==6.7.1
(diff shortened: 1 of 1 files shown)


JacobCallahan:
trigger: test-robottelo

lpramuk (approved):

omkarkhatavkar (approved):

Gauravtalreja1:
Would you like to apply or cherrypick this temporary solution to all branches? Perhaps we can wait for Py3.12 for zstream branches?

JacobCallahan:
PRT failing due to the tests not existing yet. However, @damoore044 verified the changes locally.

PR #14021, opened 2024-02-08
[6.12.z] Temporarily add docker and paramiko deps

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/14020

Since we're currently installing broker from github, we can't install the extra deps. Until it is back, we need to explicitly add in docker and its ssh support layer paramiko.


Satellite-QE:
trigger: test-robottelo

JacobCallahan (approved):

PR #14025, opened 2024-02-08
Add test coverage for BZ:2192939

Gauravtalreja1:
### Problem Statement
Missing test coverage for BZ:2192939, which verifies config report generated for host gets Puppet origin correctly, when reports are without any messages.

### Solution
 1. Add test to validate config reports gets generated for host and has Puppet origin set correctly, when reports are without any messages.
2. Updating configure_puppet helper to not run puppet agent if run_puppet_agent=False, after we sign agent's certificate.
3. Furthermore, additional validation is added for the 'puppetmaster_fqdn' host fact, when puppetmaster is Sat/Cap server, and existing tests are updated accordingly.

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -983,7 +983,7 @@ class ContentHost(Host, ContentHostMixins):
                 f'Failed to put hostname in ssh known_hosts files:\n{result.stderr}'
             )
 
-    def configure_puppet(self, proxy_hostname=None):
+    def configure_puppet(self, proxy_hostname=None, run_puppet_agent=True):
         """Configures puppet on the virtual machine/Host.
         :param proxy_hostname: external capsule hostname
         :return: None.
@@ -1023,12 +1023,14 @@ class ContentHost(Host, ContentHostMixins):
         self.execute('/opt/puppetlabs/bin/puppet agent -t')
         proxy_host = Host(hostname=proxy_hostname)
         proxy_host.execute(f'puppetserver ca sign --certname {cert_name}')
-        # This particular puppet run would create the host entity under
-        # 'All Hosts' and let's redirect stderr to /dev/null as errors at
-        #  this stage can be ignored.
-        result = self.execute('/opt/puppetlabs/bin/puppet agent -t 2> /dev/null')
-        if result.status:
-            raise ContentHostError('Failed to configure puppet on the content host')
+
+        if run_puppet_agent:
+            # This particular puppet run would create the host entity under
+            # 'All Hosts' and let's redirect stderr to /dev/null as errors at
+            #  this stage can be ignored.
+            result = self.execute('/opt/puppetlabs/bin/puppet agent -t 2> /dev/null')
+            if result.status:
+                raise ContentHostError('Failed to configure puppet on the content host')
 
     def execute_foreman_scap_client(self, policy_id=None):
         """Executes foreman_scap_client on the vm to create security audit report.
--- a/tests/foreman/cli/test_report.py
+++ b/tests/foreman/cli/test_report.py
@@ -127,6 +127,7 @@ def test_positive_run_puppet_agent_generate_report_when_no_message(
     :customerscenario: true
 
     :BZ: 2192939, 2257327, 2257314
+    :parametrized: yes
     """
     sat = session_puppet_enabled_sat
     client = rhel_contenthost
--- a/tests/foreman/cli/test_report.py
+++ b/tests/foreman/cli/test_report.py
@@ -14,6 +14,7 @@
 import random
 
 import pytest
+from wait_for import wait_for
 
 from robottelo.exceptions import CLIReturnCodeError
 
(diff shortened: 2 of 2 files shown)


Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_report.py

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_report.py

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_report.py

tests/foreman/cli/test_report.py
```diff
@@ -82,11 +82,63 @@ def test_positive_install_configure_host(
     for client, puppet_proxy in zip(content_hosts, puppet_infra_host, strict=True):
         client.configure_puppet(proxy_hostname=puppet_proxy.hostname)
         report = session_puppet_enabled_sat.cli.ConfigReport.list(
-            {'search': f'host~{client.hostname},origin=Puppet'}
+            {'search': f'host~{client.hostname}'}
```
jyejare, line 85: Are we now relying completely on puppet for reporting, dont we have choice here for Ansible facts etc.

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_report.py::test_positive_run_puppet_agent_generate_report_when_no_message

Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_report.py::test_positive_run_puppet_agent_generate_report_when_no_message

tests/foreman/cli/test_report.py
```diff
@@ -82,11 +83,68 @@ def test_positive_install_configure_host(
...
+
+    :expectedresults:
+        1. Configuration passes without errors.
+        2. Config report is created and contains Puppet origin
+        3. Facts are acquired.
+
+    :customerscenario: true
+
+    :BZ: 2192939, 2257327, 2257314
+    """
```
shubhamsg199, line 130:
```suggestion
    :parametrized: yes
    """
```

shubhamsg199 (approved):

robottelo/hosts.py
```diff
@@ -1023,12 +1023,14 @@ def configure_puppet(self, proxy_hostname=None):
         self.execute('/opt/puppetlabs/bin/puppet agent -t')
```
ekohl, line 1023: Nowadays (since Puppet 6) if you want to bootstrap you use `puppet ssl bootstrap`. Then you can sign it. You may want to pass `--waitforcert 0` as well to prevent it waiting for the server to sign it.

robottelo/hosts.py
```diff
@@ -1023,12 +1023,14 @@ def configure_puppet(self, proxy_hostname=None):
...
-        #  this stage can be ignored.
-        result = self.execute('/opt/puppetlabs/bin/puppet agent -t 2> /dev/null')
-        if result.status:
-            raise ContentHostError('Failed to configure puppet on the content host')
+
+        if run_puppet_agent:
+            # This particular puppet run would create the host entity under
+            # 'All Hosts' and let's redirect stderr to /dev/null as errors at
+            #  this stage can be ignored.
+            result = self.execute('/opt/puppetlabs/bin/puppet agent -t 2> /dev/null')
```
ekohl, line 1031: Using `-t` isn't quite a good test. While testing the original BZ I noticed it is NOT a reproducer for the issue.

`-t` (or `--test`) is a short hand for multiple options. Quoting the man page:

```
* --test:
  Enable the most common options used for testing. These are 'onetime',
  'verbose', 'no-daemonize', 'no-usecacheonfailure', 'detailed-exitcodes',
  'no-splay', and 'show_diff'.
```

The `--verbose` part always creates entries in the report. If you only want to run puppet it should be sufficient to use `puppet agent --onetime --no-daemonize --no-splay`.

tests/foreman/cli/test_report.py
```diff
@@ -82,11 +83,69 @@ def test_positive_install_configure_host(
...
+
+    :BZ: 2192939, 2257327, 2257314
+    :parametrized: yes
+    """
+    sat = session_puppet_enabled_sat
+    client = rhel_contenthost
+    client.configure_puppet(proxy_hostname=sat.hostname, run_puppet_agent=False)
+    # Run either 'puppet agent --detailed-exitcodes' or 'systemctl restart puppet'
+    # to generate Puppet config report for host without any messages
+    assert client.execute('/opt/puppetlabs/bin/puppet agent --detailed-exitcodes').status == 0
```
ekohl, line 137: This starts a daemon. You want to use `--onetime --no-daemonize --no-splay` as arguments to prevent that.

tests/foreman/cli/test_report.py
```diff
@@ -82,11 +83,69 @@ def test_positive_install_configure_host(
...
+    wait_for(
+        lambda: sat.cli.ConfigReport.list({'search': f'host~{client.hostname}'}) != [],
+        timeout=300,
+        delay=30,
+    )
+    report = sat.cli.ConfigReport.list({'search': f'host~{client.hostname}'})
```
ekohl, line 143: This is a list of reports while the name implies it is a singular entity. I'd name it `reports`

tests/foreman/cli/test_report.py
```diff
@@ -82,11 +83,69 @@ def test_positive_install_configure_host(
...
+    assert len(report)
+    assert report[0]['origin'] == 'Puppet'
```
ekohl, line 145: This will pass if there a report with the origin puppet. How about asserting every report has an origin instead?
```python
assert None not in [report['origin'] for report in reports]
```

tests/foreman/cli/test_report.py
```diff
@@ -82,11 +83,69 @@ def test_positive_install_configure_host(
     for client, puppet_proxy in zip(content_hosts, puppet_infra_host, strict=True):
         client.configure_puppet(proxy_hostname=puppet_proxy.hostname)
         report = session_puppet_enabled_sat.cli.ConfigReport.list(
-            {'search': f'host~{client.hostname},origin=Puppet'}
+            {'search': f'host~{client.hostname}'}
         )
         assert len(report)
+        assert report[0]['origin'] == 'Puppet'
         facts = session_puppet_enabled_sat.cli.Fact.list({'search': f'host~{client.hostname}'})
         assert len(facts)
```
ekohl, line 91: This is a rather useless assertion. There's no guarantee that the `puppetmaster_fqdn` fact is present.

tests/foreman/cli/test_report.py
```diff
@@ -82,11 +83,69 @@ def test_positive_install_configure_host(
     for client, puppet_proxy in zip(content_hosts, puppet_infra_host, strict=True):
         client.configure_puppet(proxy_hostname=puppet_proxy.hostname)
         report = session_puppet_enabled_sat.cli.ConfigReport.list(
-            {'search': f'host~{client.hostname},origin=Puppet'}
+            {'search': f'host~{client.hostname}'}
         )
         assert len(report)
+        assert report[0]['origin'] == 'Puppet'
         facts = session_puppet_enabled_sat.cli.Fact.list({'search': f'host~{client.hostname}'})
```
ekohl, line 90: Why not search specifically for the `puppetmaster_fqdn` fact?

tests/foreman/cli/test_report.py
```diff
@@ -82,11 +83,69 @@ def test_positive_install_configure_host(
...
         assert len(facts)
+        assert [f for f in facts if f['fact'] == 'puppetmaster_fqdn'][0][
+            'value'
+        ] == puppet_proxy.hostname
```
ekohl, line 94: I'd write this as:
```python
assert [f['value'] for f in facts if f['fact'] == 'puppetmaster_fqdn'] == [puppet_proxy.hostname]
```

This won't fail with an `IndexError` if the `puppetmaster_fqdn` fact isn't present.

ekohl (commented):
I was out for a few days so a bit late to the party, but I think this needs some work. At least how `puppet agent` is invoked is wrong and needs to change.

PR #14028, opened 2024-02-09
Integrate FAM pipeline into robottelo

Griffin-Sullivan:
### Problem Statement
Current FAM misc pipeline has been broken for a couple releases. This is due to some version changes, but also we were not testing things as accurately as possible. For instance, we were cloning upstream and running the develop branch.

### Solution
Integrate the FAM pipeline into robottelo. This now runs everything on the satellite against itself, so we know that the collection being packaged in Satellite is working correctly.

### Note
The expectation is not that this will pass 100%. Getting all of the modules and roles to pass will take individual setup that needs to be done in robottelo, AAP, or upstream in the test playbook itself. For now I would like to see that the tests are executing as expected and the results show helpful output on failure (ie: the actual ansible tasking logs). After merging, we will need to evaluate as teams how to remediate the issues based on component and team.


Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/sys/test_fam.py::test_positive_run_modules_and_roles

Griffin-Sullivan:
Not a terrible start (30/86 failed). You can see though that the failure output is not very helpful since it doesn't give you the full ansible play logs. I'm gonna research if there's a way to force Jenkins or pytest to show everything.

tests/foreman/sys/test_fam.py
```diff
@@ -32,6 +41,48 @@ def sync_roles(target_sat):
...
+    )
+    module_target_sat.execute(
+        f'''sed -i 's/rhsm_username:.*/rhsm_username: "{settings.subscription.rhn_username}"/g' {config_file}'''
+    )
+    module_target_sat.execute(
+        f'''sed -i 's|subscription_manifest_path:.*|subscription_manifest_path: "data/{module_sca_manifest.name}.zip"|g' {config_file}'''
+    )
+    module_target_sat.execute(
+        f'''sed -i 's/rhsm_password:.*/rhsm_password: "{settings.subscription.rhn_password}"/g' {config_file}'''
+    )
```
evgeni, line 77: given the amount of changes we need to perform here, would it be easier/cleaner if we'd deploy this file (`server.yml`) from a template and completely ignore the example we ship in the package?

tests/foreman/sys/test_fam.py
```diff
@@ -32,6 +41,48 @@ def sync_roles(target_sat):
...
+
+    # Handle color output formatting
+    module_target_sat.execute('export ANSIBLE_NOCOLOR=True && export PY_COLORS=0')
```
evgeni, line 80: does `.execute()` always start a new ssh/shell process? if so, those `export`s won't do anything

Griffin-Sullivan: I'm actually not sure. Good question for @JacobCallahan

tests/foreman/sys/test_fam.py
evgeni: No idea ;)
I was hoping there is just a way in robotello?

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/sys/test_fam.py::test_positive_run_modules_and_roles

Griffin-Sullivan:
Reran cause I realized I needed to add some more test_playbooks to run. This is now running the right amount of tests.

tests/foreman/sys/test_fam.py
evgeni: Looking at https://github.com/SatelliteQE/broker/blob/master/broker/hosts.py, it seems the ssh session is kept alive (if possible).

I'd probably still move that to the actual test invocation below.

tests/foreman/sys/test_fam.py
evgeni: I've found https://github.com/SatelliteQE/robottelo/blob/97d9117d74835665cad06b59d39ab5802a63b0ec/robottelo/hosts.py#L881-L892

evgeni: or https://github.com/SatelliteQE/robottelo/blob/97d9117d74835665cad06b59d39ab5802a63b0ec/robottelo/hosts.py#L1003-L1018

tests/foreman/sys/test_fam.py
```diff
@@ -32,6 +40,48 @@ def sync_roles(target_sat):
...
+    module_target_sat.execute(
+        f"sed -i 's/pytest -vv/pytest -v --color=no /g' {FAM_ROOT_DIR}/Makefile"
```
evgeni, line 81: you want `-vv` here, which will make it output all ansible output (as you wanted)

Griffin-Sullivan:
trigger: test-robottelo
pytest: tests/foreman/sys/test_fam.py::test_positive_run_modules_and_roles

tests/foreman/sys/test_fam.py
```diff
@@ -32,6 +40,47 @@ def sync_roles(target_sat):
...
+    module_target_sat.execute(
+        f'''sed -i 's|subscription_manifest_path:.*|subscription_manifest_path: "data/{module_sca_manifest.name}.zip"|g' {config_file}'''
+    )
+    module_target_sat.execute(
+        f'''sed -i 's/rhsm_password:.*/rhsm_password: "{settings.subscription.rhn_password}"/g' {config_file}'''
+    )
+
+    # Handle color output formatting
+    module_target_sat.execute(
+        f"sed -i 's/pytest -vv/pytest -vv --color=no /g' {FAM_ROOT_DIR}/Makefile"
```
evgeni, line 80: setting `PY_COLORS=0` below should be sufficient to force colors off. Sure this is required?

tests/foreman/sys/test_fam.py
```diff
@@ -71,3 +120,20 @@ def test_positive_import_run_roles(sync_roles, target_sat):
...
+def test_positive_run_modules_and_roles(module_target_sat, setup_fam, ansible_module):
+    """Run all FAM modules and roles on the Satellite
+
+    :id: b595756f-627c-44ea-b738-aa17ff5b1d39
+
+    :expectedresults: All modules and roles run successfully
+    """
+    # Execute test_playbook
+    result = module_target_sat.execute(
+        'export ANSIBLE_NOCOLOR=True && export PY_COLORS=0 && '
```
evgeni, line 136: Both [Ansible](https://docs.ansible.com/ansible/latest/reference_appendices/config.html#ansible-nocolor) and [Pytest](https://docs.pytest.org/en/8.0.x/reference/reference.html#envvar-NO_COLOR) support [`NO_COLOR`](https://no-color.org/) so it should be sufficient to set `NO_COLOR=True`

tests/foreman/sys/test_fam.py
```diff
@@ -71,3 +120,20 @@ def test_positive_import_run_roles(sync_roles, target_sat):
...
+        f'. ~/localenv/bin/activate && cd {FAM_ROOT_DIR} && make livetest_{ansible_module}'
+    )
+    assert 'PASSED' in result.stdout
```
evgeni, line 139: I think you want to also assert exit code being zero, just to be sure?

evgeni:
I think I know why we're missing the "real" output in the test results.

The pytest that is running the "outer" tests is not called with `-vv`:
```
18:36:18  + pytest tests/foreman/sys/test_fam.py::test_positive_run_modules_and_roles --external-logging --junit-xml=sat-results.xml -o junit_suite_name=sat-result
```

tests/foreman/sys/test_fam.py
Griffin-Sullivan: ok and I'm not sure. I never made those changes originally

Griffin-Sullivan:
> I think I know why we're missing the "real" output in the test results.
> 
> The pytest that is running the "outer" tests is not called with `-vv`:
> 
> ```
> 18:36:18  + pytest tests/foreman/sys/test_fam.py::test_positive_run_modules_and_roles --external-logging --junit-xml=sat-results.xml -o junit_suite_name=sat-result
> ```

Interesting. I'll try changing it. Let me update everything and we'll rerun.

Griffin-Sullivan:
trigger: test-robottelo
pytest: -vv tests/foreman/sys/test_fam.py::test_positive_run_modules_and_roles

Satellite-QE:
**PRT Result**
```
Build Number: 5721
Build Status: UNSTABLE
PRT Comment: pytest -vv tests/foreman/sys/test_fam.py::test_positive_run_modules_and_roles
Test Result : =========== 41 failed, 69 passed, 137 warnings in 3850.63s (1:04:10) ===========
```

evgeni:
Well, at least it didn't eat up the output now

tests/foreman/sys/test_fam.py
Griffin-Sullivan: We would still have to replace the hostname and manifest name during the test setup. For now I'd leave this as is even though it's ugly and we can change in the future.

evgeni (approved):

tests/foreman/sys/test_fam.py
```diff
@@ -32,6 +40,42 @@ def sync_roles(target_sat):
...
+@pytest.fixture(scope='module')
+def setup_fam(module_target_sat, module_sca_manifest):
+    # Execute AAP WF for FAM setup
+    Broker().execute(workflow='fam-test-setup', source_vm=module_target_sat.name)
+
+    # Edit Makefile to not try to rebuild the collection when tests run
+    module_target_sat.execute(f"sed -i '/^live/ s/$(MANIFEST)//' {FAM_ROOT_DIR}/Makefile")
+
+    # Upload manifest to test playbooks directory
+    module_target_sat.put(f'{module_sca_manifest.path}', f'{module_sca_manifest.name}')
```
JacobCallahan, line 52: why the f-strings here?

JacobCallahan (changes requested):
great addition! just one minor thing

JacobCallahan (approved):

ogajduse (approved):
Co-reviewed with @jameerpathan111 
LGTM, let's see how it behaves in the production.

jameerpathan111 (approved):

PR #14036, opened 2024-02-12
addFinalizer in test body

pondrejk:
### Problem Statement
Turns out that `@request.addFinalizer` when used in test body (not in fixture) is dependent on the execution order, so basically when some assertion defined in the test before the finalizer part fails, finalizer is not added and so not executed.

This exactly happened in `test_positive_disable_hammer_defaults` in upgrades, breaking a lot of other tests in the same run. Big kudos to @lpramuk for isolating the issue there.

### Solution

Moving the addFinalizer part to the start of the test, fixes the problem. Prt run won't reveal much in this case, but running the previous version of the test with induced assertion error leaves satellite with:

```
~]# hammer defaults list
----------------|------
PARAMETER       | VALUE
----------------|------
organization_id | 41   
----------------|------
```

...which means that the finalizer didn't run. With this pr, the defaults are properly cleaned up:

```
~]# hammer defaults list
----------|------
PARAMETER | VALUE
----------|------
```

### Related Issues

I did a quick search for this pattern across robottelo and I found quite a bit of cases where `@request.addfinalizer` is added towards the end of the test body, with some preceding assertions that could possibly break it:

- [x] https://github.com/SatelliteQE/robottelo/blob/d9c5ea85cf4c12c9a0b8e0cde269fc26c0814378/tests/foreman/cli/test_errata.py#L692
- [x] https://github.com/SatelliteQE/robottelo/blob/d9c5ea85cf4c12c9a0b8e0cde269fc26c0814378/tests/foreman/api/test_discoveryrule.py#L188
- [x] https://github.com/SatelliteQE/robottelo/blob/d9c5ea85cf4c12c9a0b8e0cde269fc26c0814378/tests/foreman/api/test_computeresource_libvirt.py#L256
- [x] https://github.com/SatelliteQE/robottelo/blob/d9c5ea85cf4c12c9a0b8e0cde269fc26c0814378/tests/foreman/api/test_parameters.py#L58
- [x] https://github.com/SatelliteQE/robottelo/blob/d9c5ea85cf4c12c9a0b8e0cde269fc26c0814378/tests/foreman/maintain/test_maintenance_mode.py#L146
- [x] https://github.com/SatelliteQE/robottelo/blob/d9c5ea85cf4c12c9a0b8e0cde269fc26c0814378/tests/foreman/api/test_http_proxy.py#L316
- [x] https://github.com/SatelliteQE/robottelo/blob/d9c5ea85cf4c12c9a0b8e0cde269fc26c0814378/tests/foreman/maintain/test_packages.py#L219
- [x] https://github.com/SatelliteQE/robottelo/blob/d9c5ea85cf4c12c9a0b8e0cde269fc26c0814378/tests/foreman/ui/test_computeresource_gce.py#L217
- [x] https://github.com/SatelliteQE/robottelo/blob/d9c5ea85cf4c12c9a0b8e0cde269fc26c0814378/tests/foreman/maintain/test_advanced.py#L104
- [x] https://github.com/SatelliteQE/robottelo/blob/d9c5ea85cf4c12c9a0b8e0cde269fc26c0814378/tests/foreman/destructive/test_capsule_loadbalancer.py#L234
- [x] https://github.com/SatelliteQE/robottelo/blob/d9c5ea85cf4c12c9a0b8e0cde269fc26c0814378/tests/foreman/maintain/test_health.py#L219
- [x] https://github.com/SatelliteQE/robottelo/blob/d9c5ea85cf4c12c9a0b8e0cde269fc26c0814378/tests/foreman/ui/test_ansible.py#L233

This doesn't include `request.addFinalizer(...)` calls that likely suffer from the same problem. Now, I could go and reorder all those occurrences, but I feel we should discuss this team wide and maybe abolish this practice all together

### DIFF
--- a/tests/foreman/api/test_computeresource_libvirt.py
+++ b/tests/foreman/api/test_computeresource_libvirt.py
@@ -113,9 +113,9 @@ def test_positive_create_with_name_description(
         location=[module_location],
         url=LIBVIRT_URL,
     ).create()
+    request.addfinalizer(compresource.delete)
     assert compresource.name == name
     assert compresource.description == name
-    request.addfinalizer(compresource.delete)
 
 
 @pytest.mark.tier2
@@ -134,9 +134,9 @@ def test_positive_create_with_orgs_and_locs(request, module_target_sat):
     compresource = module_target_sat.api.LibvirtComputeResource(
         location=locs, organization=orgs, url=LIBVIRT_URL
     ).create()
+    request.addfinalizer(compresource.delete)
     assert {org.name for org in orgs} == {org.read().name for org in compresource.organization}
     assert {loc.name for loc in locs} == {loc.read().name for loc in compresource.location}
-    request.addfinalizer(compresource.delete)
 
 
 @pytest.mark.tier2
@@ -175,8 +175,8 @@ def test_negative_create_with_same_name(request, module_target_sat, module_org, 
     cr = module_target_sat.api.LibvirtComputeResource(
         location=[module_location], name=name, organization=[module_org], url=LIBVIRT_URL
     ).create()
-    assert cr.name == name
     request.addfinalizer(cr.delete)
+    assert cr.name == name
     with pytest.raises(HTTPError):
         module_target_sat.api.LibvirtComputeResource(
             name=name,
@@ -245,18 +245,15 @@ def test_negative_update_same_name(request, module_target_sat, module_org, modul
     compresource = module_target_sat.api.LibvirtComputeResource(
         location=[module_location], name=name, organization=[module_org], url=LIBVIRT_URL
     ).create()
+    request.addfinalizer(compresource.delete)
     new_compresource = module_target_sat.api.LibvirtComputeResource(
         location=[module_location], organization=[module_org], url=LIBVIRT_URL
     ).create()
+    request.addfinalizer(new_compresource.delete)
     new_compresource.name = name
     with pytest.raises(HTTPError):
         new_compresource.update(['name'])
     assert new_compresource.read().name != name
-
-    @request.addfinalizer
-    def _finalize():
-        compresource.delete()
-        new_compresource.delete()
 
 
 @pytest.mark.tier2
--- a/tests/foreman/api/test_discoveryrule.py
+++ b/tests/foreman/api/test_discoveryrule.py
@@ -171,8 +171,11 @@ def test_positive_multi_provision_with_rule_limit(
 
     :CaseImportance: High
     """
+
     discovered_host1 = module_target_sat.api_factory.create_discovered_host()
+    request.addfinalizer(module_target_sat.api.Host(id=discovered_host1['id']).delete)
     discovered_host2 = module_target_sat.api_factory.create_discovered_host()
+    request.addfinalizer(module_target_sat.api.Host(id=discovered_host2['id']).delete)
     rule = module_target_sat.api.DiscoveryRule(
         max_count=1,
         hostgroup=module_discovery_hostgroup,
@@ -181,14 +184,6 @@ def test_positive_multi_provision_with_rule_limit(
         organization=[discovery_org],
         priority=1000,
     ).create()
+    request.addfinalizer(rule.delete)
     result = module_target_sat.api.DiscoveredHost().auto_provision_all()
     assert '1 discovered hosts were provisioned' in result['message']
-
-    # Delete discovery rule
-    @request.addfinalizer
-    def _finalize():
-        rule.delete()
-        module_target_sat.api.Host(id=discovered_host1['id']).delete()
-        module_target_sat.api.DiscoveredHost(id=discovered_host2['id']).delete()
-        with pytest.raises(HTTPError):
-            rule.read()
--- a/tests/foreman/api/test_discoveryrule.py
+++ b/tests/foreman/api/test_discoveryrule.py
@@ -175,7 +175,7 @@ def test_positive_multi_provision_with_rule_limit(
     discovered_host1 = module_target_sat.api_factory.create_discovered_host()
     request.addfinalizer(module_target_sat.api.Host(id=discovered_host1['id']).delete)
     discovered_host2 = module_target_sat.api_factory.create_discovered_host()
-    request.addfinalizer(module_target_sat.api.Host(id=discovered_host2['id']).delete)
+    request.addfinalizer(module_target_sat.api.DiscoveredHost(id=discovered_host2['id']).delete)
     rule = module_target_sat.api.DiscoveryRule(
         max_count=1,
         hostgroup=module_discovery_hostgroup,
--- a/tests/foreman/api/test_http_proxy.py
+++ b/tests/foreman/api/test_http_proxy.py
@@ -303,6 +303,11 @@ def test_positive_sync_proxy_with_certificate(request, target_sat, module_org, m
 
     :customerscenario: true
     """
+
+    @request.addfinalizer
+    def _finalize():
+        target_sat.custom_certs_cleanup()
+
     # Cleanup any existing certs that may conflict
     target_sat.custom_certs_cleanup()
     proxy_host = settings.http_proxy.auth_proxy_url.replace('http://', '').replace(':3128', '')
--- a/tests/foreman/cli/test_errata.py
+++ b/tests/foreman/cli/test_errata.py
@@ -688,7 +688,7 @@ def test_positive_list_affected_chosts_by_erratum_restrict_flag(
             'inclusion': 'false',
         }
     )
-    request.addfinalizer(
+    request.addfinalizer(lambda:
         cv_filter_cleanup(
             target_sat, cv_filter['filter-id'], module_cv, module_sca_manifest_org, module_lce
         )
--- a/tests/foreman/cli/test_errata.py
+++ b/tests/foreman/cli/test_errata.py
@@ -688,16 +688,11 @@ def test_positive_list_affected_chosts_by_erratum_restrict_flag(
             'inclusion': 'false',
         }
     )
-
-    @request.addfinalizer
-    def cleanup():
+    request.addfinalizer(
         cv_filter_cleanup(
-            target_sat,
-            cv_filter['filter-id'],
-            module_cv,
-            module_sca_manifest_org,
-            module_lce,
+            target_sat, cv_filter['filter-id'], module_cv, module_sca_manifest_org, module_lce
         )
+    )
 
     # Make rule to hide the RPM that creates the need for the installable erratum
     target_sat.cli_factory.content_view_filter_rule(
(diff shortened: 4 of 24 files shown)


Griffin-Sullivan (approved):
I see it's on the agenda for tomorrow, but yeah each team will need to fix these.

lpramuk (approved):
LGTM
I've enjoyed collab with endeavour team. Finding this problematic test :wink:

tests/foreman/cli/test_hammer.py
```diff
@@ -136,6 +136,13 @@ def test_positive_disable_hammer_defaults(request, function_product, target_sat)
...
+
+    @request.addfinalizer
+    def _finalize():
+        target_sat.cli.Defaults.delete({'param-name': 'organization_id'})
+        result = target_sat.execute('hammer defaults list')
+        assert str(function_product.organization.id) not in result.stdout
+
     target_sat.cli.Defaults.add(
         {'param-name': 'organization_id', 'param-value': function_product.organization.id}
     )
```
Gauravtalreja1, line 148: Yes, Good catch, I've seen this behaviour earlier while writing some test, just one suggestion here, to add finalizer at least after we add entity for better readability
```suggestion
    target_sat.cli.Defaults.add(
        {'param-name': 'organization_id', 'param-value': function_product.organization.id}
    )
    @request.addfinalizer
    def _finalize():
        target_sat.cli.Defaults.delete({'param-name': 'organization_id'})
        result = target_sat.execute('hammer defaults list')
        assert str(function_product.organization.id) not in result.stdout
```

JacobCallahan (approved):

pondrejk:
I'm drafting this so that I could fix other occurrences as discussed on the meeting today

tests/foreman/api/test_computeresource_libvirt.py
```diff
@@ -241,6 +241,12 @@ def test_negative_update_same_name(request, module_target_sat, module_org, modul
 
     :CaseImportance: High
     """
+
+    @request.addfinalizer
+    def _finalize():
+        compresource.delete()
+        new_compresource.delete()
+
```
Gauravtalreja1, line 249: we can move this below after we create compresource and new_compresource?

tests/foreman/api/test_discoveryrule.py
```diff
@@ -171,6 +171,15 @@ def test_positive_multi_provision_with_rule_limit(
 
     :CaseImportance: High
     """
+
+    @request.addfinalizer
+    def _finalize():
+        rule.delete()
```
Gauravtalreja1, line 177: same here

tests/foreman/ui/test_computeresource_gce.py
```diff
@@ -245,6 +246,14 @@ def test_positive_gce_cloudinit_provision_end_to_end(
 
     :expectedresults: Host is provisioned successfully
     """
+
+    @request.addfinalizer
+    def _finalize():
+        gcehost = sat_gce.api.Host().search(query={'search': f'name={hostname}'})
+        if gcehost:
+            gcehost[0].delete()
+        googleclient.disconnect()
```
Gauravtalreja1, line 255: Can we move it at least below after variables get defined? functionally I think it wouldn't affect as we're checking if host exists here first before delete

tests/foreman/ui/test_computeresource_gce.py
```diff
@@ -161,6 +161,14 @@ def test_positive_gce_provision_end_to_end(
 
     :expectedresults: Host is provisioned successfully
     """
+
+    @request.addfinalizer
+    def _finalize():
+        gcehost = sat_gce.api.Host().search(query={'search': f'name={hostname}'})
+        if gcehost:
```
Gauravtalreja1, line 168: same here..

tests/foreman/cli/test_hammer.py
pondrejk: @Gauravtalreja1 It can be done for finalizers that touch just one thing, but as long as you change more items, you can't have them all defined before, because it would defer the point of making sure finalizer is always added. This may be odd to read, but on the other hand it also communicates that the finalizer behaves differently from the rest of the test code, meaning it is not executed first but needs to be added as early as possible

tests/foreman/cli/test_errata.py
```diff
@@ -623,6 +623,17 @@ def test_positive_list_affected_chosts_by_erratum_restrict_flag(
...
+
+    @request.addfinalizer
+    def cleanup():
+        cv_filter_cleanup(
+            target_sat,
+            cv_filter['filter-id'],
+            module_cv,
+            module_sca_manifest_org,
+            module_lce,
+        )
```
JacobCallahan, line 635: What would happen if the finalizer here was called before one of those arguments was created, cv_filter for example?

tests/foreman/maintain/test_advanced.py
```diff
@@ -251,6 +253,17 @@ def test_positive_sync_plan_with_hammer_defaults(request, sat_maintain, module_o
...
+
+    @request.addfinalizer
+    def _finalize():
+        sat_maintain.cli.Defaults.delete({'param-name': 'organization_id'})
+        sync_plans[1].delete()
+        sync_plan = sat_maintain.api.SyncPlan(organization=module_org.id).search(
+            query={'search': f'name="{sync_plans[0]}"'}
+        )
+        if sync_plan:
+            sync_plans[0].delete()
```
JacobCallahan, line 265: similar question as before. if the finalizer is called before sync_plans is populated (with at least two entries), then this will error.

JacobCallahan (changes requested):

tests/foreman/cli/test_errata.py
pondrejk: yeah it is a confusing syntax compared to what we are normally used to in python. The function marked with `@request.addfinalizer` within the test function is *executed* **after** the test runs, so you can use variables even before you define them (!). But at the same time, the position of the block within the test determines when the finalizer is *added*. So I place these blocks at the start to be sure the finalizer is added at all times. The position in test matters for addition but not for execution

lhellebr: And I think that's what Jacob asks about. What if the finalizer is added (here, at the start) and then something fails after it is added, but before e.g. `cv_filter` variable is defined. I suppose the finalizer would throw an exception?

JacobCallahan: correct @lhellebr. at that point, you'd raise an exception because the name `cv_filter` is referenced before definition. Moving this to right after everything is defined is safer.

pondrejk: @JacobCallahan @lhellebr I see, yeah I suppose it would raise an exception, but that's not different from "normal" finalizer behavior if defined in fixture. Consider the two alternatives when having finalizer in test body: 

a) tests fails in the middle, finalizer added before -> result: finalizer added and executed, exception raised if referencing stuff that didn't get created 
b) test fails in the middle, finalizer added after -> result: finalizer not added nor executed, not cleaning up anything that got created before failure, possibly causing interference with other tests

Situation (b) actually happens in the upgrade run, and it took considerable time and effort just to find out what is going on. Therefore I'd much rather have situation (a), which would be easier to debug

pondrejk:
@Gauravtalreja1 @JacobCallahan @lhellebr hello, based on your feedback I changed ordering where it made sense. Could you please revisit?

tests/foreman/api/test_discoveryrule.py
```diff
@@ -171,8 +171,11 @@ def test_positive_multi_provision_with_rule_limit(
 
     :CaseImportance: High
     """
+
     discovered_host1 = module_target_sat.api_factory.create_discovered_host()
+    request.addfinalizer(module_target_sat.api.Host(id=discovered_host1['id']).delete)
     discovered_host2 = module_target_sat.api_factory.create_discovered_host()
+    request.addfinalizer(module_target_sat.api.Host(id=discovered_host2['id']).delete)
```
Gauravtalreja1, line 178:
```suggestion
    request.addfinalizer(module_target_sat.api.DiscoveredHost(id=discovered_host2['id']).delete)
```

tests/foreman/api/test_http_proxy.py
```diff
@@ -303,6 +303,11 @@ def test_positive_sync_proxy_with_certificate(request, target_sat, module_org, m
 
     :customerscenario: true
     """
+
+    @request.addfinalizer
+    def _finalize():
+        target_sat.custom_certs_cleanup()
```
Gauravtalreja1, line 309: Can we keep only one request.addfinalizer method in this test?

tests/foreman/cli/test_errata.py
```diff
@@ -861,17 +856,11 @@ def test_host_errata_search_commands(
             'inclusion': 'false',
         }
     )
-
-    @request.addfinalizer
-    def cleanup():
+    request.addfinalizer(
```
Gauravtalreja1, line 859:
```suggestion
    request.addfinalizer(lambda:
```

tests/foreman/destructive/test_capsule_loadbalancer.py
```diff
@@ -219,6 +220,7 @@ def test_loadbalancer_install_package(
         if loadbalancer_setup['setup_capsules']['capsule_1'].hostname in result.stdout
         else loadbalancer_setup['setup_capsules']['capsule_2']
     )
+    request.addfinalizer(registered_to_capsule.power_control(state=VmState.RUNNING, ensure=True))
```
Gauravtalreja1, line 223:
```suggestion
    request.addfinalizer(lambda: registered_to_capsule.power_control(state=VmState.RUNNING, ensure=True))
```

tests/foreman/api/test_discoveryrule.py
JacobCallahan: Are these not identical actions?

Gauravtalreja1: no, when we provision discovered_host with rule in next step, discovered host will be moved to host entity

tests/foreman/api/test_http_proxy.py
pondrejk: that kind of contradicts with other concerns voiced in comments, if there are more elements to remove and you want to make sure all are marked to be finalized immediately after creation but not before, then you end up with multiple finalizers

tests/foreman/cli/test_hammer.py
```diff
@@ -136,6 +136,13 @@ def test_positive_disable_hammer_defaults(request, function_product, target_sat)
 
     :BZ: 1640644, 1368173
     """
+
+    @request.addfinalizer
+    def _finalize():
+        target_sat.cli.Defaults.delete({'param-name': 'organization_id'})
+        result = target_sat.execute('hammer defaults list')
```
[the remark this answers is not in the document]
pondrejk: I'd like to keep this specific case as it is, as it has broken many tests in the upgrade run

tests/foreman/api/test_http_proxy.py
Gauravtalreja1: I didn't understand this, do we really need to finalizers calling `target_sat.custom_certs_cleanup()` this specific test?
Initially, I would've held the merge on this, but now that it's self-merged, could you please take care of this in a separate PR if you agree with the above?

Gauravtalreja1: @pondrejk ^

lhellebr (commented):
This is already merged but I discussed this with Peter and I ACK the final result where the order of actions makes more sense.

PR #14045, opened 2024-02-12
[6.14.z] Add test coverage for BZ:2192939

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/14025

### Problem Statement
Missing test coverage for BZ:2192939, which verifies config report generated for host gets Puppet origin correctly, when reports are without any messages.

### Solution
 1. Add test to validate config reports gets generated for host and has Puppet origin set correctly, when reports are without any messages.
2. Updating configure_puppet helper to not run puppet agent if run_puppet_agent=False, after we sign agent's certificate.
3. Furthermore, additional validation is added for the 'puppetmaster_fqdn' host fact, when puppetmaster is Sat/Cap server, and existing tests are updated accordingly.

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -993,7 +993,7 @@ class ContentHost(Host, ContentHostMixins):
                 f'Failed to put hostname in ssh known_hosts files:\n{result.stderr}'
             )
 
-    def configure_puppet(self, proxy_hostname=None):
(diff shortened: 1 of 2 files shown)


Satellite-QE:
trigger: test-robottelo
pytest: tests/foreman/cli/test_report.py::test_positive_run_puppet_agent_generate_report_when_no_message

PR #14074, opened 2024-02-14
Move HTTP Proxy tests under correct component

vsedmik:
### Problem Statement
While HTTP Proxy does exist in BZ as a unique component and robotello has `{api|cli|ui}/http_proxy.py` modules too, the test cases still belong under the `Repository` component for some reason.

This doesn't seem logical and makes it more difficult to do TFA, so I believe we should create a separate component in robottelo and CI and flip the tests there.

### Solution
This PR (and another one in CI).

### DIFF
--- a/testimony.yaml
+++ b/testimony.yaml
@@ -56,6 +56,7 @@ CaseComponent:
     - ForemanMaintain
     - Hammer
     - Hammer-Content
+    - HTTPProxy
     - HostCollections
     - HostForm
     - HostGroup
--- a/tests/foreman/api/test_http_proxy.py
+++ b/tests/foreman/api/test_http_proxy.py
@@ -2,7 +2,7 @@
 
 :Requirement: HttpProxy
 
-:CaseComponent: Repositories
+:CaseComponent: HTTPProxy
 
 :team: Phoenix-content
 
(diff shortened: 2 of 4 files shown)


testimony.yaml
```diff
@@ -56,6 +56,7 @@ CaseComponent:
     - ForemanMaintain
     - Hammer
     - Hammer-Content
+    - HttpProxy
```
Gauravtalreja1, line 59: @vsedmik You might need to add this components.yml in CI repo to run this, and IIRC for CI we just remove spaces from component names provided in ohsnap for each team, so I'd suggest following a similar naming convention here, @devendra104 can also confirm this
```suggestion
    - HTTPProxy
```

vsedmik: I have the CI MR up already. I considered the `HttpProxy` better readable, but can change it if necessary. @devendra104 ?

devendra104: I agree with you @vsedmik but good to keep it as similar to the Ohsnap's defined component name until something exceptional.

vsedmik: Updated, thanks for the feedback!

Gauravtalreja1 (approved):

PR #14093, opened 2024-02-15
remove the PRT labels for new commits

omkarkhatavkar:
### Problem Statement
Currently, the labels `PRT-Passed` and `PRT-Failed` are added based on the results of the Pull Request Testing (PRT). However, when contributors update the pull request with new commits, these labels might remain, leading to potential confusion. There is a need to automatically remove these labels when new commits are added to the pull request.

### Solution
To address this issue, we propose implementing a solution using GitHub Actions (GHA). The solution involves periodically checking the GitHub status API to monitor for new commits on the pull request. Upon detecting a new commit, the workflow will automatically remove the `PRT-Passed` and `PRT-Failed` labels from the pull request. This ensures that the labels accurately reflect the current status of the pull request, reducing confusion and maintaining clarity for reviewers and contributors.

### Test Result 
https://github.com/omkarkhatavkar/robottelo/actions/runs/7916581527/job/21610780591?pr=216

### DIFF
--- a/.github/workflows/prt_labels.yml
+++ b/.github/workflows/prt_labels.yml
@@ -1,0 +1,49 @@
+name: Remove the PRT label, for the new commit
+
+on:
+  pull_request:
+    types: ["synchronize"]
+
+jobs:
+    prt_labels_remover:
+      name: remove the PRT label when amendments or new commits added to PR
+      runs-on: ubuntu-latest
+      if: "(contains(github.event.pull_request.labels.*.name, 'PRT-Passed') || contains(github.event.pull_request.labels.*.name, 'PRT-Failed'))"
+      steps:
+        - name: Avoid the race condition as PRT result will be cleaned
+          run: |
+            echo "Avoiding the race condition if prt result will be cleaned" && sleep 60
+
+        - name: Fetch the PRT status
+          id: prt
+          uses: omkarkhatavkar/wait-for-status-checks@main
+          with:
+            ref: ${{ github.head_ref }}
+            context: 'Robottelo-Runner'
+            wait-interval: 2
+            count: 5
+
+        - name: remove the PRT Passed/Failed label, for new commit
+          if: always() && ${{steps.prt.outputs.result}} == 'not_found'
+          uses: actions/github-script@v7
+          with:
+            github-token: ${{ secrets.CHERRYPICK_PAT }}
+            script: |
+              const prNumber = '${{ github.event.number }}';
+              const issue = await github.rest.issues.get({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: prNumber,
+              });
+              const labelsToRemove = ['PRT-Failed', 'PRT-Passed'];
+              const labelsToRemoveFiltered = labelsToRemove.filter(label => issue.data.labels.some(({ name }) => name === label));
+              if (labelsToRemoveFiltered.length > 0) {
+                await Promise.all(labelsToRemoveFiltered.map(async label => {
+                  await github.rest.issues.removeLabel({
+                    issue_number: prNumber,
+                    owner: context.repo.owner,
+                    repo: context.repo.repo,
+                    name: label
+                  });
+                }));
+              }


JacobCallahan (approved):

omkarkhatavkar:
Dont worry about the failing check here as It will be going to get added. It will work only after getting merge because the github token for `SatQE` is inside the repo.

.github/workflows/prt_labels.yml
```diff
@@ -0,0 +1,44 @@
...
+          with:
+            ref: ${{ github.head_ref }}
+            context: 'Robottelo-Runner'
+            wait-interval: 2
+            count: 5
+
+        - name: remove the PRT Passed/Failed label, for new commit
+          if: always() && ${{steps.prt.outputs.result}} == 'not_found'
+          uses: actions/github-script@v7
+          with:
```
Gauravtalreja1, line 24: With this change, will it be possible to delete the previous PRT result comment as well, when new commit appears?
In case of rebase, I think we shouldn't delete, but for new commit we could delete it to avoid confusion for results or do you consider to handle it differently, wdyt @omkarkhatavkar (cc @jyejare @JacobCallahan) ?

omkarkhatavkar: Consider the case where you have two different PRT comments with different tests (collection) and will have different test results. `rebase` would not delete the labels.

jyejare: @Gauravtalreja1 Good Question/Ask. As @omkarkhatavkar stated, deleting the previous PRT result wouldn't be required in all the cases.

But as part of this request, we could do multiple things:
- Add a comment on PR that the PR is refreshed, need to rerun the PRT.
- OR If no PRT specific label on PR, block the PR merge.
- etc.

omkarkhatavkar: @Gauravtalreja1 if you are okay with change can you please resolve this discussion ?

Gauravtalreja1: @omkarkhatavkar Yes, I'm okay with the current changes, but I feel it can be surely considered for future, so I think we can have a issue created, or we can keep this thread

And about the first suggestion of @jyejare I think its not good idea as adding such comments will unnecessarily increase the comment history, and for second suggestion, I feel we shouldn't block the PR merge on PRT results ever, as some PRs might not require results like easy fix PRs, syntax fixing PRs, or some dependabot PRs which gets automerged when basic GHA checks pass

Gauravtalreja1 (approved):

.github/workflows/prt_labels.yml
omkarkhatavkar: @Gauravtalreja1 As discussed in yesterday discussion @JacobCallahan also agreed with majority of people having no concern about having multiple PRT result comments. As well as github also have feature of grouping similar comments. I personally dont feel we have to track this as there no action item right now.

omkarkhatavkar: @Gauravtalreja1 auto merge is not enabled on the master, we never merge the PR on the master. Yesterday I talked about the cherry-picks PR. Adding the automerge will help in merging the PR's quickly. I dont think we would like to merge blocking based on the PRT Failure Result. Current label structure provide more detail to help people as author and reviewer

jyejare: @Gauravtalreja1 @omkarkhatavkar I was just throwing ideas if we really want to strict the PRT results / labels and maintain accurate history. But the real thing is we dont want to delete the RPT history for now.

jyejare (approved):

.github/workflows/prt_labels.yml
```diff
@@ -0,0 +1,44 @@
+name: Remove the PRT label, for the new commit
+
+on:
+  pull_request:
+    types: ["opened", "synchronize"]
```
jyejare, line 5: @omkarkhatavkar only `synchronize` should be good enough because anyways we dont have any PRT labels on new PR.

.github/workflows/prt_labels.yml
```diff
@@ -0,0 +1,44 @@
...
+
+jobs:
+    prt_labels:
```
jyejare, line 8:
```suggestion
    prt_labels_remover:
```

.github/workflows/prt_labels.yml
```diff
@@ -0,0 +1,44 @@
...
+      name: remove the PRT label, if it is a new commit
```
jyejare, line 9:
```suggestion
      name: remove the PRT label when amendments or new commits added to PR
```

.github/workflows/prt_labels.yml
```diff
@@ -0,0 +1,44 @@
...
+          id: prt
+          uses: omkarkhatavkar/wait-for-status-checks@main
+          with:
+            ref: ${{ github.head_ref }}
+            context: 'Robottelo-Runner'
+            wait-interval: 2
+            count: 5
+
+        - name: remove the PRT Passed/Failed label, for new commit
+          if: always() && ${{steps.prt.outputs.result}} == 'not_found'
```
jyejare, line 22: If I am not wrong, I think you can add another check here that searches for labels on the PR that has string `PRT`. That would save your time below and abort the operation if no PRT ran at all on PR hence no PRT labels.

jyejare (changes requested):
ACK pending comments and suggestions!

.github/workflows/prt_labels.yml
omkarkhatavkar: updated

.github/workflows/prt_labels.yml
omkarkhatavkar: updated

.github/workflows/prt_labels.yml
omkarkhatavkar: added at the higher level !

.github/workflows/prt_labels.yml
```diff
@@ -0,0 +1,45 @@
...
```
jyejare, line 23: Is this check really needed to verify if the PRT outputs result is not found ? 

I think we can keep it simple, On PR refresh just remove the labels regardless of PRT status.

jyejare (commented):
ACK pending another suggestion. Sorry for double review :)

.github/workflows/prt_labels.yml
jyejare: But then, this will never work in case when: 
1. PRT is run has a check for PRT in PR checks
2. Added a new commit, that would run this workflow of label removals

And this is the most common case in robottelo. So question, Does PRT check is always removed and ASAP new commit/ammendments is pushed to PR on all PRs ?

omkarkhatavkar: > 1. PRT is run has a check for PRT in PR checks
First of all this check not to see if PRT is passed or Failed. It is checking the that PRT is `Not Found` (which literally means there is new commit as PRT result goes away with status API). This is pretty quick check that will run after new commit. If someone does the PRT comment even in that case it will done. This is seperate GHA. So it will never disturb the main PRT cherry-picks or auto-merge GHA's      

> 2. Added a new commit, that would run this workflow of label removals
Isn't that we want. We wanted to remove the labels related to `PRT-Passed/Failed`  in case new commit as the code has been updated. Previous PRT result not valid now. 

> . So question, Does PRT check is always removed and ASAP new commit/ammendments is pushed to PR on all PRs ?
That's true. Hence I added the check in PRT GHA that get's me the `Not Found` Result. Leverage that result to remove the `PRT-Passed/Failed` labels. 

I know it's bit tricky to understand. I will have more direct in person conversation with you so that it become bit easier to grab some internal details.

omkarkhatavkar: offline discussion, adding some more sleep would really help in for avoiding the race condition

PR #14100, opened 2024-02-16
Add flake8-simplify and flake8-return to ruff config

ogajduse:
This PR adds two new rulesets - flake8-simplify and flake8-return.

### DIFF
--- a/pytest_fixtures/component/satellite_auth.py
+++ b/pytest_fixtures/component/satellite_auth.py
@@ -271,11 +271,12 @@ def auth_data(request, ad_data, ipa_data):
         ad_data['attr_login'] = LDAP_ATTR['login_ad']
         ad_data['auth_type'] = auth_type
         return ad_data
-    elif auth_type == 'ipa':
+    if auth_type == 'ipa':
         ipa_data['server_type'] = LDAP_SERVER_TYPE['UI']['ipa']
         ipa_data['attr_login'] = LDAP_ATTR['login']
         ipa_data['auth_type'] = auth_type
         return ipa_data
+    return None
 
 
 @pytest.fixture(scope='module')
--- a/conf/dynaconf_hooks.py
+++ b/conf/dynaconf_hooks.py
@@ -157,7 +157,7 @@ def get_dogfood_satclient_repos(settings):
 
 
 def get_ohsnap_repo_url(settings, repo, product=None, release=None, os_release=None, snap=''):
-    repourl = dogfood_repository(
+    return dogfood_repository(
         settings.ohsnap,
         repo=repo,
         product=product,
@@ -165,4 +165,3 @@ def get_ohsnap_repo_url(settings, repo, product=None, release=None, os_release=N
         os_release=os_release,
         snap=snap,
     ).baseurl
-    return repourl
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,8 @@ select = [
     "I", # isort
     # "Q",  # flake8-quotes
     "PT",  # flake8-pytest
+    "RET",  # flake8-return
+    "SIM",  # flake8-simplify
     "UP",  # pyupgrade
     "W",  # pycodestyle
 ]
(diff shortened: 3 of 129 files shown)


pytest_fixtures/component/satellite_auth.py
```diff
@@ -271,11 +271,12 @@ def auth_data(request, ad_data, ipa_data):
         ad_data['attr_login'] = LDAP_ATTR['login_ad']
         ad_data['auth_type'] = auth_type
         return ad_data
-    elif auth_type == 'ipa':
+    if auth_type == 'ipa':
         ipa_data['server_type'] = LDAP_SERVER_TYPE['UI']['ipa']
         ipa_data['attr_login'] = LDAP_ATTR['login']
         ipa_data['auth_type'] = auth_type
         return ipa_data
+    return None
```
jyejare, line 279: We dont really need to return None, it by default returns None if nothing else returned.
```
 def maxfun():
   ...:     sd = 'lop'

In [4]: type(maxfun())
Out[4]: NoneType

```

jameerpathan111 (approved):

pytest_fixtures/component/satellite_auth.py
ogajduse: You are correct, that is an implicit behavior of Python. 
The documentation of the rule that checks missing explicit `return None` statements in functions that return NoneType and other types sums it well.

https://docs.astral.sh/ruff/rules/implicit-return/#why-is-this-bad
> Why is this bad?
> 
> The lack of an explicit return statement at the end of a function that can return non-None values can cause confusion. Python implicitly returns None if no other return value is present. Adding an explicit return None can make the code more readable by clarifying intent.

In this case, the author of this fixture decided to implement the function in a way that it could return `NoneType` and also `Box`, therefore I see a benefit in the explicit `return None` statement because it might not be obvious that it could also return None at the first glance.

omkarkhatavkar:
trigger: test-robottelo

Satellite-QE:
**PRT Result**
```
Build Number: 5820
Build Status: UNSTABLE
PRT Comment: pytest null
Test Result : =========================== short test summary info ============================
```

ogajduse:
PRT failure is unrelated to the changes in this PR:
`AttributeError: 'bytes' object has no attribute 'content'`

lhellebr (approved):
Obviously I haven't checked all the automatically made changes but including these rulesets lgtm.

ogajduse:
Due to long wait time, the conflict in `robottelo/cli/hammer.py` must be resolved.

lpramuk (approved):

pytest_fixtures/component/satellite_auth.py
lpramuk: I like these `return None` :heart: 
They often indicate flaws in our code.

jyejare: @lpramuk Could you elaborate on `indicate flows` ?

LadislavVasina1: I see no problem with returning `None` either. As @ogajduse stated by citing the documentation, it is good to return that `None`. Having some stuff explicitly stated could benefit the programmer so he does not have to think about all the other things and possibly forget about something.  The other thing in my opinion is that minimal code is not always a good thing to do. What do you think?

lpramuk: Sometimes is important to rely on return type and explicit `return None` helps to realize this fact.

This PR chases READABILITY of code, not minimalistic code.

jyejare: @LadislavVasina1 @lpramuk @ogajduse Ok! If you all agrees that explicit return improves the readability and thats going to be helpful, I would ACK in favor of `us`.

@LadislavVasina1 But in general, we shouldnt enforce those rules just because they are there, but should be looked for `how these could be useful to us or not`. 

Since you all are in favor of keeping it, I will Agree!

PR #13999, opened 2024-02-06
cli e2e fix

pondrejk:
### Problem Statement
e2e failing with key error

### Solution
updating the assert

PR #14109, opened 2024-02-16
zip_longest instead of zip in parse_csv

pondrejk:
### Problem Statement

The Ruff code standard B905 requires the zip strict value to be explicitly stated, it does not require it to be True. 

### Solution

This PR sets the value to what it (implicitly) was before https://github.com/SatelliteQE/robottelo/pull/13491 thus removing the need for https://github.com/SatelliteQE/robottelo/pull/13737 https://github.com/SatelliteQE/robottelo/pull/13797 https://github.com/SatelliteQE/robottelo/pull/13999 https://github.com/SatelliteQE/robottelo/pull/13989 and additional changes discussed in the last mentioned PR

[edit]: using zip_longest to not drop any data and adding log about the mismatch

### DIFF
--- a/robottelo/cli/hammer.py
+++ b/robottelo/cli/hammer.py
@@ -1,7 +1,10 @@
 """Helpers to interact with hammer command line utility."""
 import csv
+from itertools import zip_longest
 import json
 import re
+
+from robottelo.logging import logger
 
 
 def _normalize(header):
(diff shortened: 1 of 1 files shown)


pondrejk:
trigger: test-robottelo

JacobCallahan:
Does this work in the situations all of those linked PRs were trying to solve or does it just mask the problem? The reason I ask is because there is a difference in zipped iterables where data from one is lost by turning off strict mode.

Satellite-QE:
**PRT Result**
```
Build Number: 5778
Build Status: SUCCESS
PRT Comment: pytest null
Test Result : ======== 11 passed, 6009 deselected, 99 warnings in 1635.28s (0:27:15) =========
```

pondrejk:
> Does this work in the situations all of those linked PRs were trying to solve or does it just mask the problem? The reason I ask is because there is a difference in zipped iterables where data from one is lost by turning off strict mode.

It does work in the sense that it unblocks tests. Without the need to work around the csv sniffing, we are left only with jobinvocation hammer output that does not fit. There is a bz for it, so maybe I can add some logic around that.

pondrejk:
So I made a change to zip_longest, so that data are kept in case of mismatch

ogajduse:
An observation from my investigation on #13989:
We can completely ditch `zip` if we use `csv.DictReader` from the standard library. https://docs.python.org/3/library/csv.html#csv.DictReader

pondrejk:
closing in favor of https://github.com/SatelliteQE/robottelo/pull/13989

PR #14133, opened 2024-02-20
SCA refactor of Subscription HTTP proxy test

vsedmik:
### Problem Statement
As 6.16 is SCA-only we need to get rid of entitlements and subscriptions related logic in our tests but keep the approach user take for this or similar scenario.

### Solution
Get rid of entitlements/subscriptions. Also I propose to change this test case to `e2e` since it syncs a RH repo, publishes and promotes it to an LCE, registers a content host via AK and consumes the content in it. Case remain parametrized to run with authenticated, unauthenticated and without (this param could be probably omited) HTTP proxy.

One minor update in `enable_sync_redhat_repo` so that we can use repos from constants which define `arch` key instead of `basearch` or even no arch at all.

**Open question:** Should the test id be changed since it doesn't cover the subscription manipulation at all?

### PRT test Cases example
trigger: test-robottelo
pytest: tests/foreman/api/test_http_proxy.py -k install_content_with_http_proxy

### DIFF
--- a/robottelo/host_helpers/api_factory.py
+++ b/robottelo/host_helpers/api_factory.py
@@ -144,7 +144,7 @@ class APIFactory:
         """Enable the RedHat repo, sync it and returns repo_id"""
         # Enable RH repo and fetch repository_id
         repo_id = self.enable_rhrepo_and_fetchid(
-            basearch=rh_repo['basearch'],
+            basearch=rh_repo.get('basearch', rh_repo.get('arch', DEFAULT_ARCHITECTURE)),
             org_id=org_id,
             product=rh_repo['product'],
             repo=rh_repo['name'],
--- a/tests/foreman/api/test_http_proxy.py
+++ b/tests/foreman/api/test_http_proxy.py
@@ -125,6 +125,7 @@ def test_positive_end_to_end(setup_http_proxy, module_target_sat, module_manifes
     assert docker_repo['result'] == 'success'
 
 
+@pytest.mark.e2e
 @pytest.mark.upgrade
 @pytest.mark.rhel_ver_match('8')
 @pytest.mark.run_in_one_thread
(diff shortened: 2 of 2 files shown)


vsedmik:
trigger: test-robottelo
pytest: tests/foreman/api/test_http_proxy.py -k install_content_with_http_proxy

Satellite-QE:
**PRT Result**
```
Build Number: 5802
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/api/test_http_proxy.py -k install_content_with_http_proxy
Test Result : ========== 3 passed, 5 deselected, 112 warnings in 987.15s (0:16:27) ===========
```

jameerpathan111:
> Get rid of entitlements/subscriptions. Also I propose to change this test case to `e2e` since it syncs a RH repo, publishes and promotes it to an LCE, registers a content host via AK and consumes the content in it. Case remain parametrized to run with authenticated, unauthenticated and without (this param could be probably omited) HTTP proxy.

> **Open question:** Should the test id be changed since it doesn't cover the subscription manipulation at all?
>
@vsedmik I think it's better to keep it the same as it's still in a way testing the HTTP proxy functionalities.
The only thing I'd recommend checking is whether the code from `function_sca_manifest_org` gets executed before or after the Http Proxy is set. If it's not happening then I'd even recommend adding the manifest upload code to the test itself.

vsedmik:
> The only thing I'd recommend checking is whether the code from `function_sca_manifest_org` gets executed before or after the Http Proxy is set. If it's not happening then I'd even recommend adding the manifest upload code to the test itself.

As far as I can see the manifest is uploaded AFTER the HTTP proxy is set. So from this point of view we should be safe I guess. But good point anyway!

jameerpathan111 (approved):

synkd (approved):

PR #14205, opened 2024-02-28
Fix pytest deprecation warning

dosas:
Applying a mark to a fixture function never had any effect but is a common user error.

see: https://docs.pytest.org/en/stable/deprecations.html#applying-a-mark-to-a-fixture-function

### DIFF
--- a/pytest_fixtures/component/host.py
+++ b/pytest_fixtures/component/host.py
@@ -21,7 +21,6 @@ def module_model():
     return entities.Model().create()
 
 
-@pytest.mark.skip_if_not_set('clients', 'fake_manifest')
 @pytest.fixture(scope="module")
 def setup_rhst_repo(module_target_sat):
     """Prepare Satellite tools repository for usage in specified organization"""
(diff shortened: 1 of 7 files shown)


jyejare:
@dosas Thanks for raising the PR. Since its a warning now and errors in future versions, we will have to address this sooner.

The appropriate solution I feel here is to use those markers directly on tests/test_classes/test_modules.

We shall get back to on this soon!!

lhellebr (approved):

lhellebr:
Maybe there should be another PR putting those marks to the appropriate location but this one is good as it is, removing something that has no effect.

jyejare:
@lhellebr Even I dont see any changes needed from PR diff that holds this PR from merging. But still need to assess once with non-diff code if the removals has any indirect impact! 

I shall have that checked by EOD today!

jyejare (approved):
Confirming these that removing the markers `custom` or `inbuilt` does not have any impact as they were not really being used by tests or any inhouse pytest plugins. 

@dosas Did you try to cover all such occurrences throughout robottelo fixtures? or Do you us to be completing this ?

dosas:
I ran 
```
pytest --collect-only -q tests -W error
```
Without any errors, so all instances should be covered

jyejare (approved):

PR #14217, opened 2024-02-28
Flip repos_collection fixture to SCA

vsedmik:
### Problem Statement
As 6.16 is SCA-only we need to flip from `entitlement` to `SCA` mode. This fixture is still using entitlement, breaking all tests that use it in setup.

### Solution
Flip it to SCA.

### DIFF
--- a/pytest_fixtures/component/repository.py
+++ b/pytest_fixtures/component/repository.py
@@ -227,7 +227,7 @@ def module_repos_collection_with_setup(request, module_target_sat, module_org, m
 
 @pytest.fixture(scope='module')
 def module_repos_collection_with_manifest(
-    request, module_target_sat, module_entitlement_manifest_org, module_lce
+    request, module_target_sat, module_sca_manifest_org, module_lce
 ):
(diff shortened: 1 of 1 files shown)


sambible (approved):

sambible:
You beat me to this change, haha!

jyejare (approved):

vsedmik:
> @vsedmik This change can be done for 6.15.z as well. Wdyt?

@shweta83 For 6.15 I would keep the `entitlement` mode since there are still tests which use this fixture and test subscription-related actions and fields (like sub status, auto-attach, etc). For 6.16 it's a _must_ however and those tests must be converted anyway for 6.16.

shweta83:
> > @vsedmik This change can be done for 6.15.z as well. Wdyt?
> 
> @shweta83 For 6.15 I would keep the `entitlement` mode since there are still tests which use this fixture and test subscription-related actions and fields (like sub status, auto-attach, etc). For 6.16 it's a _must_ however and those tests must be converted anyway for 6.16.

Ok. I am good with this.

shweta83 (approved):
Ack

PR #14275, opened 2024-03-07
Broker dependency updated to resolve ssh2-python limitation/packagever issue

jyejare:
### Problem Statement
Checks in Robottelo PRs failing due to last broker version 0.4.5 where we have seen rate limit issue as well as packageName issue for new repo ssh2-python312

### Solution
The issue  is resolved in latest 0.4.7 version.


JacobCallahan:
Correct. The submodules aren't supported in this way by either pypi or pip. I have retracted that version of Broker for now. Will revert that change and push a new 0.4.6.
I'll start looking into a longer term solution to all of this next week.

shweta83 (approved):
Ack

pondrejk (approved):

PR #14314, opened 2024-03-08
remove log check causing false negatives

rmynar:
Removing the common assertions from fixture teardown. Current setup lead to an error in last test of a test module, when some error appeared in logs during testing.

Example:
```
test_a
...
test_n PASSED or FAILED but creates error entries in logs
...
test_z PASSED or FAILED but throws an ERROR due to error entries in logs
```

### DIFF
--- a/tests/foreman/installer/test_installer.py
+++ b/tests/foreman/installer/test_installer.py
@@ -1381,8 +1381,7 @@ def sat_default_install(module_sat_ready_rhels):
         f'foreman-initial-admin-password {settings.server.admin_password}',
     ]
     install_satellite(module_sat_ready_rhels[0], installer_args)
-    yield module_sat_ready_rhels[0]
-    common_sat_install_assertions(module_sat_ready_rhels[0])
+    return module_sat_ready_rhels[0]
 
 
 @pytest.fixture(scope='module')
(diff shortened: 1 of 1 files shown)


rmynar:
trigger: test-robottelo
pytest: tests/foreman/installer/test_installer.py -k 'test_content_guarded_distributions_option or test_positive_check_installer_hammer_ping'

rmynar:
trigger: test-robottelo
pytest: tests/foreman/installer/test_installer.py -k "test_content_guarded_distributions_option or test_positive_check_installer_hammer_ping"

rmynar:
@lhellebr There were some errors in candlepin.log after _test_content_guarded_distributions_option_ which actually passed. The errors appear after some manifest operation, they're not caused by installer. The errors say `ERROR org.candlepin.audit.DefaultEventMessageReceiver - Unable to process message 99: argument "content" is null` and it seems to be fixed [here](https://github.com/candlepin/candlepin/pull/4473)
Consider this as a hotfix to prevent error of last test in module on teardown. It could be useful to check the logs after each test, but only the part of the logs which appeared since the test started. We're trying to find a way to do it.
We still check logs after [clean installation](https://github.com/SatelliteQE/robottelo/blob/520775beba534e59380c422952f67caa8481a7fc/tests/foreman/installer/test_installer.py#L1805)

lhellebr (approved):

ogajduse (approved):
@rmynar I would like to see these checks to be back soon so the tests can potentially catch errors.

PR #14377, opened 2024-03-12
Use manifester from fixture

dosas:
### Problem Statement

There were two tests that were not using manifester from fixture

### Solution
 see PR

### DIFF
--- a/tests/foreman/api/test_repositories.py
+++ b/tests/foreman/api/test_repositories.py
@@ -11,7 +11,6 @@
 :CaseImportance: Critical
 
 """
-from manifester import Manifester
 from nailgun.entity_mixins import call_entity_method_with_timeout
 import pytest
 from requests.exceptions import HTTPError
@@ -187,7 +186,7 @@ def test_positive_sync_kickstart_repo(module_sca_manifest_org, target_sat):
 
 
 @pytest.mark.tier1
-def test_negative_upload_expired_manifest(module_org, target_sat):
+def test_negative_upload_expired_manifest(module_org, target_sat, function_entitlement_manifest):
     """Upload an expired manifest and attempt to refresh it
 
     :id: d6e652d8-5f46-4d15-9191-d842466d45d0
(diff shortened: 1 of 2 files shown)


synkd (commented):
We should check with the authors of these tests, but I believe it was by design that they did not use Manifester via fixtures in these cases. `function_entitlement_manifest` returns a `Request` object, not a `Manifester` object, so the Manifester instance methods are not available to it or the other Manifester fixtures. If a test requires Manifester instance methods, it's currently necessary to instantiate Manifester locally in order to access those methods.

@omkarkhatavkar @ColeHiggins2 can you confirm that these tests are intentionally instantiating Manifester objects?

dosas:
@synkd It makes sense to ask the original authors.

But this function https://github.com/SatelliteQE/robottelo/blob/master/pytest_fixtures/component/taxonomy.py#L226 IMHO returns an Instance of `Manifester`

vsedmik:
Apart from the question above, in `master` (targets Satellite 6.16) we need to replace `*_entitlement_manifest` by `*_sca_manifest`, since 6.16 will be SCA-only so we are going to refactor these tests anyway.

In other words, this PR needs to be raised against `6.15.z` if we want to use the proposed changes.

dosas:
https://github.com/SatelliteQE/robottelo/pull/14405

dosas:
https://github.com/SatelliteQE/robottelo/pull/14405

dosas:
@vsedmik Isn't that an argument for this PR to go into master? Now you would have to change the fixture in 3 places. After this fix the change would only be in one place.

vsedmik:
> @vsedmik Isn't that an argument for this PR to go into master? Now you would have to change the manifester in 3 places. After this fix the change would only be in one place.

@dosas My thought was it needs to be resolved in master differently anyway in a separate PR. Your changes appeared to me reasonable, but for `6.15.z` and older branches.

dosas:
Alright, I created and linked the PR for 6.15.z. Feel free to close/merge the PR you see fit.

PR #14405, opened 2024-03-14
[6.15.z] Use manifester from fixture

dosas:
https://github.com/SatelliteQE/robottelo/pull/14377

We should check with the authors of these tests, but I believe it was by design that they did not use Manifester via fixtures in these cases. function_entitlement_manifest returns a Request object, not a Manifester object, so the Manifester instance methods are not available to it or the other Manifester fixtures. If a test requires Manifester instance methods, it's currently necessary to instantiate Manifester locally in order to access those methods.

@omkarkhatavkar @ColeHiggins2 can you confirm that these tests are intentionally instantiating Manifester objects?

### DIFF
--- a/tests/foreman/api/test_repositories.py
+++ b/tests/foreman/api/test_repositories.py
@@ -11,7 +11,6 @@
 :CaseImportance: Critical
 
 """
-from manifester import Manifester
 from nailgun.entity_mixins import call_entity_method_with_timeout
 import pytest
 from requests.exceptions import HTTPError
@@ -187,7 +186,7 @@ def test_positive_sync_kickstart_repo(module_sca_manifest_org, target_sat):
 
 
 @pytest.mark.tier1
-def test_negative_upload_expired_manifest(module_org, target_sat):
+def test_negative_upload_expired_manifest(module_org, target_sat, function_entitlement_manifest):
     """Upload an expired manifest and attempt to refresh it
 
     :id: d6e652d8-5f46-4d15-9191-d842466d45d0
@@ -199,10 +198,9 @@ def test_negative_upload_expired_manifest(module_org, target_sat):
 
     :expectedresults: Manifest refresh should fail and return error message
     """
-    manifester = Manifester(manifest_category=settings.manifest.entitlement)
-    manifest = manifester.get_manifest()
+    manifest = function_entitlement_manifest.get_manifest()
     target_sat.upload_manifest(module_org.id, manifest.content)
-    manifester.delete_subscription_allocation()
+    function_entitlement_manifest.delete_subscription_allocation()
     with pytest.raises(CLIReturnCodeError) as error:
         target_sat.cli.Subscription.refresh_manifest({'organization-id': module_org.id})
     assert (
--- a/tests/upgrades/test_subscription.py
+++ b/tests/upgrades/test_subscription.py
@@ -11,11 +11,9 @@
 :CaseImportance: High
 
 """
-from manifester import Manifester
 import pytest
 
 from robottelo import constants
-from robottelo.config import settings
 from robottelo.hosts import ContentHost
 
 
@@ -144,7 +142,9 @@ class TestSubscriptionAutoAttach:
 
     @pytest.mark.parametrize('pre_upgrade_data', ['rhel7', 'rhel8', 'rhel9'], indirect=True)
     @pytest.mark.post_upgrade(depend_on=test_pre_subscription_scenario_auto_attach)
-    def test_post_subscription_scenario_auto_attach(self, request, target_sat, pre_upgrade_data):
+    def test_post_subscription_scenario_auto_attach(
+        self, request, target_sat, pre_upgrade_data, function_entitlement_manifest
+    ):
         """Run subscription auto-attach on pre-upgrade content host registered
         with Satellite.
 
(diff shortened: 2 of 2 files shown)


synkd (commented):
PRT is currently down, but running the repositories test with these changes locally fails with an `AttributeError`:
```
$ pytest tests/foreman/api/test_repositories.py::test_negative_upload_expired_manifest
===================================================================================================================================== test session starts =====================================================================================================================================

*** snip ***

collected 1 item                                                                                                                                              

tests/foreman/api/test_repositories.py F                                                                                                                                                                                                                                                [100%]

========================================================================================================================================== FAILURES ===========================================================================================================================================
____________________________________________________________________________________________________________________________ test_negative_upload_expired_manifest ____________________________________________________________________________________________________________________________

*** snip ***

>       manifest = function_entitlement_manifest.get_manifest()
E       AttributeError: 'Response' object has no attribute 'get_manifest'
```
Debugging the test, we can see that `function_entitlement_manifest` is of type `requests.models.Response`, not a Manifester object:
```
-> manifest = function_entitlement_manifest.get_manifest()
(Pdb) type(function_entitlement_manifest)
<class 'requests.models.Response'>
(Pdb) function_entitlement_manifest.get_manifest()
*** AttributeError: 'Response' object has no attribute 'get_manifest'
```
As noted above, there are some test scenarios that currently require importing and instantiating Manifester in the test module. The two tests modified by this PR are examples of this type of scenario and will fail with these changes. I would therefore suggest that this PR be closed.

dosas:
@synkd Thanks for testing. 

Now I get it, the problem is in the manifester context: https://github.com/SatelliteQE/manifester/blob/master/manifester/manifester.py#L421 it does not return an instance of manifester but the result of `get_manifest()` ...

In that case it should be closed. Would you be fine if I add a marker (`pytest.mark.manifester`) directly to those two tests?

synkd:
@dosas You're welcome, and yes, you have the issue exactly right.

I have no problem with adding the manifester marker directly to these two tests.

ColeHiggins2:
@dosas @synkd. @synkd is correct. Calling ```function_entitlement_manifest.get_manifest() will fail with ```AttributeError: 'Response' object has no attribute 'get_manifest'```. I chose to use Manifester to directly act upon the manifest object for this specific test scenario. We do not want to delete the subscription allocation for other tests that may use the function_entitlement_manifest fixture within the same scope

tests/foreman/api/test_repositories.py
```diff
@@ -199,10 +198,9 @@ def test_negative_upload_expired_manifest(module_org, target_sat):
 
     :expectedresults: Manifest refresh should fail and return error message
     """
-    manifester = Manifester(manifest_category=settings.manifest.entitlement)
-    manifest = manifester.get_manifest()
+    manifest = function_entitlement_manifest.get_manifest()
```
Gauravtalreja1, line 201: @dosas As you've mentioned `it does not return an instance of manifester but the result of get_manifest()`, so what if we can just use function_entitlement_manifest for manifest? 
CC @synkd @jyejare wdyt?
```suggestion
    target_sat.upload_manifest(module_org.id, function_entitlement_manifest.content)
    request.addfinalizer(function_entitlement_manifest.delete_subscription_allocation)
```

synkd: @Gauravtalreja1 This would be technically possible, but currently, it's a bit simpler to generate the manifest from a local Manifester instance. Manifester's `delete_subscription_allocation` method does allow passing in an arbitrary UUID argument for deletion [1], but the allocation UUID is not immediately accessible from the `function_entitlement_manifest` object. It's embedded in `function_entitlement_manifest.url` but would need to be extracted from a URL path like this one [2]. 

With that said, it should be a fairly simple change to add a line like the following to the end of Manifester's `trigger_manifest_export` method [3]:
```
manifest.uuid = self.allocation_uuid
``` 
The UUID would then be accessible as `function_entitlement_manifest.uuid`, which could be passed to `Manifester.delete_subscription_allocation()`. From the looks of it, this same change would also enable `tests/upgrades/test_subscription.py::test_post_subscription_auto_attach` to use a Manifester fixture.

If this is the direction that people want to go, then I'll also note that I would welcome a community contribution to Manifester making the necessary change.

[1] https://github.com/SatelliteQE/manifester/blob/8ff1e296d5dff89ccd9d76deffca8592807064ff/manifester/manifester.py#L184

[2] 'https://api.access.redhat.com/management/v1/allocations/0df90e19-9fda-49d8-8a3f-c7bc29ef1b23/export/2c94e7b58e14afed018e43cacf5b4db9'

[3] https://github.com/SatelliteQE/manifester/blob/8ff1e296d5dff89ccd9d76deffca8592807064ff/manifester/manifester.py#L400

dosas: @synkd I can continue here once a new release of manifester is out

synkd: @dosas I just cut a new Manifester release that includes this change. However, the build/publish to PyPi job failed, and I will be OOO the rest of this week and next week. We'll need to hold off on merging this change until the new Manifester release is published, which may need to wait until I return.

jyejare:
@dosas , Sorry to let you know that this has to be hold on until @synkd is back which will not happen this month.

tests/foreman/api/test_repositories.py
dosas: I doubt that the uuid change will enable the usage of the fixture since it still does not provide an instance of manifester

rplevka: @synkd hello, mind revisiting this? it's been sitting here for a while. Thanks!

dosas:
@synkd Are you okay with these changes?

dosas:
Not relevant anymore

PR #14424, opened 2024-03-15
Add a test case to read Capsules via non-admin user

vsedmik:
### Problem Statement
There was a bug where a non-admin user could not list particular Capsule but was able to list all Capsules, without any permissions. He should be able to list both when provided with `Reader` permissions and prevented to do so without them.

Covers [BZ#2096930](https://bugzilla.redhat.com/show_bug.cgi?id=2096930)

### Solution
Test both cases in one test case.

### PRT test Cases example
trigger: test-robottelo
pytest: tests/foreman/api/test_capsulecontent.py -k non_admin_user

PR #14434, opened 2024-03-18
[6.15.z] Add a test case to read Capsules via non-admin user

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/14424

### Problem Statement
There was a bug where a non-admin user could not list particular Capsule but was able to list all Capsules, without any permissions. He should be able to list both when provided with `Reader` permissions and prevented to do so without them.

Covers [BZ#2096930](https://bugzilla.redhat.com/show_bug.cgi?id=2096930)

### Solution
Test both cases in one test case.

### PRT test Cases example
trigger: test-robottelo
pytest: tests/foreman/api/test_capsulecontent.py -k non_admin_user

### DIFF
--- a/tests/foreman/api/test_capsulecontent.py
+++ b/tests/foreman/api/test_capsulecontent.py
@@ -17,8 +17,10 @@ import re
 from time import sleep
 
 from nailgun import client
+from nailgun.config import ServerConfig
 from nailgun.entity_mixins import call_entity_method_with_timeout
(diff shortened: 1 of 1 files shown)


Satellite-QE:
trigger: test-robottelo
pytest: tests/foreman/api/test_capsulecontent.py -k non_admin_user

PR #14449, opened 2024-03-19
Fix omitting credentials when CLI has already been instantiated

lhellebr:
### Problem Statement
Omitting credentials doesn't work when CLI has already been instantiated

### Solution
If it's been instantiated, set its omitting attribute to `True` for all the classes

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -1848,8 +1848,30 @@ class Satellite(Capsule, SatelliteMixins):
     @contextmanager
     def omit_credentials(self):
         self.omitting_credentials = True
+        # if CLI is already created
+        if self._cli._configured:
+            for file in Path('robottelo/cli/').iterdir():
+                if file.suffix == '.py' and not file.name.startswith('_'):
+                    cli_module = importlib.import_module(f'robottelo.cli.{file.stem}')
+                    for name, obj in cli_module.__dict__.items():
+                        try:
+                            if Base in obj.mro():
+                                getattr(self._cli, name).omitting_credentials = True
+                        except AttributeError:
+                            # not everything has an mro method, we don't care about them
+                            pass
         yield
         self.omitting_credentials = False
+        for file in Path('robottelo/cli/').iterdir():
+            if file.suffix == '.py' and not file.name.startswith('_'):
+                cli_module = importlib.import_module(f'robottelo.cli.{file.stem}')
+                for name, obj in cli_module.__dict__.items():
+                    try:
+                        if Base in obj.mro():
+                            getattr(self._cli, name).omitting_credentials = False
+                    except AttributeError:
+                        # not everything has an mro method, we don't care about them
+                        pass
 
     @contextmanager
     def ui_session(self, testname=None, user=None, password=None, url=None, login=True):
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -1867,7 +1867,7 @@ class Satellite(Capsule, SatelliteMixins):
                         AttributeError
                     ):  # not everything has an mro method, we don't care about them
                         if Base in obj.mro():
-                            getattr(self._cli, name).omitting_credentials = True
+                            getattr(self._cli, name).omitting_credentials = False
 
     @contextmanager
     def ui_session(self, testname=None, user=None, password=None, url=None, login=True):
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -1847,31 +1847,27 @@ class Satellite(Capsule, SatelliteMixins):
 
     @contextmanager
     def omit_credentials(self):
-        self.omitting_credentials = True
-        # if CLI is already created
-        if self._cli._configured:
-            for file in Path('robottelo/cli/').iterdir():
-                if file.suffix == '.py' and not file.name.startswith('_'):
-                    cli_module = importlib.import_module(f'robottelo.cli.{file.stem}')
-                    for name, obj in cli_module.__dict__.items():
-                        try:
-                            if Base in obj.mro():
-                                getattr(self._cli, name).omitting_credentials = True
-                        except AttributeError:
-                            # not everything has an mro method, we don't care about them
-                            pass
-        yield
-        self.omitting_credentials = False
-        for file in Path('robottelo/cli/').iterdir():
-            if file.suffix == '.py' and not file.name.startswith('_'):
-                cli_module = importlib.import_module(f'robottelo.cli.{file.stem}')
-                for name, obj in cli_module.__dict__.items():
-                    try:
+        change = not self.omitting_credentials  # if not already set to omit
+        if change:
+            self.omitting_credentials = True
+            # if CLI is already created
+            if self._cli._configured:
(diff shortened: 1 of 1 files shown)


lhellebr:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_ldap_authentication.py::test_positive_negotiate_manual_with_autonegotiation_disabled

pondrejk (approved):

robottelo/hosts.py
```diff
@@ -1848,8 +1848,30 @@ def cli(self):
...
+            for file in Path('robottelo/cli/').iterdir():
+                if file.suffix == '.py' and not file.name.startswith('_'):
+                    cli_module = importlib.import_module(f'robottelo.cli.{file.stem}')
+                    for name, obj in cli_module.__dict__.items():
+                        try:
+                            if Base in obj.mro():
+                                getattr(self._cli, name).omitting_credentials = True
+                        except AttributeError:
+                            # not everything has an mro method, we don't care about them
+                            pass
```
ogajduse, line 1862: Initially, @jameerpathan111 and I had some concerns about code repetition. But looking at it further, we concluded that it could be possibly remade into the following:
```suggestion
            for name, obj in self._cli.__dict__.items():
                with contextlib.suppress(AttributeError):  # not everything has an mro method, we don't care about them
                    if Base in obj.mro():
                        getattr(self._cli, name).omitting_credentials = True
```

With this change, it does not seem that much repetitive. :smiley:  We are not sure if it would work, though. But if the `omit_credentials` property assumes `self._cli._configured`, then we can rely on the classes being already populated there instead of going over `robottelo/cli/`.

robottelo/hosts.py
```diff
@@ -1848,8 +1848,30 @@ def cli(self):
...
         yield
         self.omitting_credentials = False
```
jameerpathan111, line 1864: @lhellebr should we check `self._cli._configured` is configured in the teardown phase too? what happens if we don't?

jameerpathan111 (commented):
Co-reviewed  with @ogajduse

ogajduse (commented):
co-reviewed with @jameerpathan111

robottelo/hosts.py
lhellebr: I think if someone does
```
with parametrized_enrolled_sat.omit_credentials():
  pass
```
Then CLI won't get instantiated at all (because `parametrized_enrolled_sat.cli`) hasn't been called. I haven't tested that though. In any case, I employed this just as a defensive programming technique to avoid error in cases we may not even think about.

robottelo/hosts.py
lhellebr: I think cycling through `robottelo/cli` ensures we won't run into some attribute of `cli` that is for some reason descendant of `Base` - but it's not an entity we should instantiate `cli` for. I don't know if any such exists and more so, I don't know if any will exist in the future. I understand your point about repetition. Please consider what I just wrote and if you don't change your opinion, I can test your version.

lhellebr: @ogajduse

robottelo/hosts.py
rplevka, line 1864: This is not necessarily `False`. One can initialize Satellite with `True`. In such scenario, this wouldn't reset the attribute to the correct original version.
Perhaps this contextmanager function could check if the `ommiting_credentials`  is set to `False` first, otherwise, it has no reason to run and could be skipped straight away.

rplevka (changes requested):

lhellebr:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_ldap_authentication.py::test_positive_negotiate_manual_with_autonegotiation_disabled tests/foreman/destructive/test_ldap_authentication.py::test_positive_autonegotiate

Satellite-QE:
**PRT Result**
```
Build Number: 6506
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/destructive/test_ldap_authentication.py::test_positive_negotiate_manual_with_autonegotiation_disabled tests/foreman/destructive/test_ldap_authentication.py::test_positive_autonegotiate --external-logging
Test Result : ================= 4 passed, 864 warnings in 9264.29s (2:34:24) =================
```

robottelo/hosts.py
```diff
@@ -1847,9 +1847,27 @@ def cli(self):
...
-        self.omitting_credentials = False
+        if change:
+            self.omitting_credentials = False
+            if self._cli._configured:
+                for name, obj in self._cli.__dict__.items():
+                    with contextlib.suppress(
+                        AttributeError
+                    ):  # not everything has an mro method, we don't care about them
+                        if Base in obj.mro():
+                            getattr(self._cli, name).omitting_credentials = True
```
JacobCallahan, line 1870: should this be True or False, if it's a reversion of the previous action?

JacobCallahan (changes requested):
You can likely simplify the reversion by adding all modified objects into a list called `modified`, then iterate over that list post-yield to update `omitting_credentials` again.

jameerpathan111 (approved):

Satellite-QE:
**PRT Result**
```
Build Number: 6647
Build Status: UNSTABLE
PRT Comment: pytest tests/foreman/destructive/test_ldap_authentication.py::test_positive_negotiate_manual_with_autonegotiation_disabled tests/foreman/destructive/test_ldap_authentication.py::test_positive_autonegotiate --external-logging
Test Result : =========== 4 passed, 1006 warnings, 1 error in 10758.13s (2:59:18) ============
```

lhellebr:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_ldap_authentication.py::test_positive_negotiate_manual_with_autonegotiation_disabled tests/foreman/destructive/test_ldap_authentication.py::test_positive_autonegotiate

Satellite-QE:
**PRT Result**
```
Build Number: 6658
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/destructive/test_ldap_authentication.py::test_positive_negotiate_manual_with_autonegotiation_disabled tests/foreman/destructive/test_ldap_authentication.py::test_positive_autonegotiate --external-logging
Test Result : ================ 4 passed, 986 warnings in 10078.11s (2:47:58) =================
```

JacobCallahan:
@lhellebr if you need help with my recommended implementation, let me know.

JacobCallahan:
@lhellebr I don't see any priority labels on this PR, so would still like you to consider my suggestion. Here it is in code form:
```python
    def omit_credentials(self):
        change = not self.omitting_credentials  # if not already set to omit
        changed = []
        if change:
            self.omitting_credentials = True
            # if CLI is already created
            if self._cli._configured:
                for name, obj in self._cli.__dict__.items():
                    with contextlib.suppress(
                        AttributeError
                    ):  # not everything has an mro method, we don't care about them
                        if Base in obj.mro():
                            attr = getattr(self._cli, name)
                            attr.omitting_credentials = True
                            changed.append(attr)
        yield
        if change:
            self.omitting_credentials = False
            for attr in changed:
                attr.omitting_credentials = False
```

PR #14474, opened 2024-03-21
Possibility to filter tests that create capsule and satellite hosts

dosas:

### DIFF
--- a/tests/foreman/conftest.py
+++ b/tests/foreman/conftest.py
@@ -32,6 +32,10 @@ def pytest_collection_modifyitems(session, items, config):
     for item in items:
         if any("manifest" in f for f in getattr(item, "fixturenames", ())):
             item.add_marker("manifester")
+        if any("satellite_host" in f for f in getattr(item, "fixturenames", ())):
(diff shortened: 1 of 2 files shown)


tests/foreman/conftest.py
```diff
@@ -32,6 +32,10 @@ def pytest_collection_modifyitems(session, items, config):
     for item in items:
         if any("manifest" in f for f in getattr(item, "fixturenames", ())):
             item.add_marker("manifester")
+        if any("satellite_host" in f for f in getattr(item, "fixturenames", ())):
```
jyejare, line 35: Can I ask the reason behind this PR implementation? Asking because there are many ways satellite/capsule is being created in tests.

This wont be able to collect all those tests, it would only collect that has `satellite_host`.

dosas: @jyejare Thanks! I think this is exactly what I need. I will check and close the MR.

lhellebr:
I suppose you are going to close this. If you still require review, let us know.

PR #14475, opened 2024-03-21
Fix upstream api permission tests

dosas:
### Problem Statement

API permission tests failing for upstream

### Solution

* add more checks for installed rpm packages
* speedup by querying rpm -qa ony once
* some tests were not updated with this commit https://github.com/SatelliteQE/robottelo/commit/6651dd63b4c308b3f7f3601f0e69d76b4c602b29#diff-9a55ae74469aa91ad5efb2b9068f73174467cd0c00d45abf7d35f9f2277a7632

Relevant tests:
```
tests/foreman/cli/test_role.py
tests/foreman/api/test_role.py
tests/foreman/api/test_permission.py

```

### DIFF
--- a/robottelo/constants/__init__.py
+++ b/robottelo/constants/__init__.py
@@ -970,13 +970,6 @@ PERMISSIONS = {
         'view_discovery_rules',
     ],
     'Domain': ['view_domains', 'create_domains', 'edit_domains', 'destroy_domains'],
-    #    'Environment': [
-    #        'view_environments',
-    #        'create_environments',
-    #        'edit_environments',
-    #        'destroy_environments',
-    #        'import_environments',
-    #    ],
     'ExternalUsergroup': [
         'view_external_usergroups',
         'create_external_usergroups',
@@ -1054,19 +1047,35 @@ PERMISSIONS = {
         'destroy_hostgroups',
         'play_roles_on_hostgroup',
     ],
-    #    'Puppetclass': [
-    #        'view_puppetclasses',
-    #        'create_puppetclasses',
-    #        'edit_puppetclasses',
-    #        'destroy_puppetclasses',
-    #        'import_puppetclasses',
-    #    ],
-    #    'PuppetclassLookupKey': [
-    #        'view_external_parameters',
-    #        'create_external_parameters',
-    #        'edit_external_parameters',
-    #        'destroy_external_parameters',
-    #    ],
+    'ForemanPuppet::ConfigGroup': [
+        'view_config_groups',
+        'create_config_groups',
+        'edit_config_groups',
+        'destroy_config_groups',
+    ],
+    'ForemanPuppet::Environment': [
+        'view_environments',
+        'create_environments',
+        'edit_environments',
+        'destroy_environments',
+        'import_environments',
+    ],
+    'ForemanPuppet::HostClass': [
+        'edit_classes',
+    ],
+    'ForemanPuppet::Puppetclass': [
+        'view_puppetclasses',
+        'create_puppetclasses',
+        'edit_puppetclasses',
+        'destroy_puppetclasses',
+        'import_puppetclasses',
+    ],
+    'ForemanPuppet::PuppetclassLookupKey': [
+        'view_external_parameters',
+        'create_external_parameters',
+        'edit_external_parameters',
+        'destroy_external_parameters',
+    ],
     'HttpProxy': [
         'view_http_proxies',
         'create_http_proxies',
--- a/tests/foreman/api/test_permission.py
+++ b/tests/foreman/api/test_permission.py
@@ -38,28 +38,35 @@ class TestPermission:
         # workaround for setting class variables
         cls = type(self)
         cls.permissions = PERMISSIONS.copy()
-        if class_target_sat.is_upstream:
-            cls.permissions[None].extend(cls.permissions.pop('DiscoveryRule'))
-            cls.permissions[None].remove('app_root')
-            cls.permissions[None].remove('attachments')
-            cls.permissions[None].remove('configuration')
-            cls.permissions[None].remove('logs')
-            cls.permissions[None].remove('view_cases')
-            cls.permissions[None].remove('view_log_viewer')
 
-        result = class_target_sat.execute('rpm -qa | grep rubygem-foreman_openscap')
-        if result.status != 0:
+        rpm_packages = class_target_sat.execute('rpm -qa').stdout
+        if 'rubygem-foreman_rh_cloud' not in rpm_packages:
+            cls.permissions.pop('InsightsHit')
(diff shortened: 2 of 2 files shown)


ogajduse:
@pondrejk @lhellebr @pnovotny Can you please take a look at @dosas's PR? It is touching a module that belongs to your team.

tests/foreman/api/test_permission.py
```diff
@@ -38,28 +38,41 @@ def create_permissions(self, class_target_sat):
...
-            cls.permissions[None].remove('attachments')
-            cls.permissions[None].remove('configuration')
-            cls.permissions[None].remove('logs')
-            cls.permissions[None].remove('view_cases')
-            cls.permissions[None].remove('view_log_viewer')
-
-        result = class_target_sat.execute('rpm -qa | grep rubygem-foreman_openscap')
-        if result.status != 0:
+        if not class_target_sat.is_upstream:
+            cls.permissions.pop('LookupValue')
```
lhellebr, line 42: This doesn't seem right. On my Satellite (stream and 6.15):
```
# hammer filter available-permissions | grep -i lookup
213 | view_lookup_values                              | LookupValue                    
214 | create_lookup_values                            | LookupValue                    
215 | edit_lookup_values                              | LookupValue                    
216 | destroy_lookup_values                           | LookupValue
```

robottelo/constants/__init__.py
```diff
@@ -1104,6 +1113,12 @@ class Colored(Box):
         'edit_operatingsystems',
         'destroy_operatingsystems',
     ],
+    'OrcharhinoCore': [
```
lhellebr, line 1116: I'm not sure about these 🐋 Orcharhino 🦏 parts. I'm not sure they should be present in this repo. I also don't think it's my decision. What do others think? @SatelliteQE/robottelo-tier-1-reviewers @SatelliteQE/robottelo-tier-2-reviewers @lvrtelov ?

tests/foreman/api/test_permission.py
```diff
@@ -38,28 +38,41 @@ def create_permissions(self, class_target_sat):
...
             cls.permissions[None].remove('view_arf_reports')
             cls.permissions[None].remove('create_arf_reports')
-        result = class_target_sat.execute('rpm -qa | grep rubygem-foreman_remote_execution')
-        if result.status != 0:
+        if 'rubygem-foreman_remote_execution' not in rpm_packages:
             cls.permissions.pop('JobInvocation')
             cls.permissions.pop('JobTemplate')
             cls.permissions.pop('RemoteExecutionFeature')
             cls.permissions.pop('TemplateInvocation')
+        if 'rubygem-foreman_puppet' not in rpm_packages:
```
lhellebr, line 66: There are more rpm conditions than previously. Have tests failed on these in upstream?

dosas: Since upstream is defined as 
```
# rpm -q satellite-capsule
package satellite-capsule is not installed
```
yes they failed.

tests/foreman/api/test_permission.py
dosas: Good catch. I have been looking through the diff but I don;t know why I thought this would not be present on satellite ...

tests/foreman/api/test_permission.py
ogajduse: https://github.com/SatelliteQE/robottelo/blob/21580fc40d64e0daaa31bf72d135bdd4d1fc9469/robottelo/hosts.py#L1540-L1547

What I got from your comment is that the current implementation of `is_upstream` does not suit you. Is that correct?

dosas: I do not really have an issue with it. It is just more a check for 'not downstream satellite'.
I am not really running the tests on upstream but a different downstream so it kind of helps me I guess.

robottelo/constants/__init__.py
lhellebr: I neither agree or disagree with you.
But where does it end? Will we have all potential Foreman forks' code here in this repo?

dosas: :D would that be a bad thing ;)

Can we maybe find a solution to (additionally) define it in the settings?

lhellebr: @jyejare If you are asking me whether Satellite supports Orcharhino... then no, no it doesn't

lhellebr: @dosas "We" can definitely find a solution but I'm afraid I'm not the person who should decide this. It's a management question, not a technical one.

dosas: sounds like it. In that case I will remove the orcharhino part and maintain it downstream.

JacobCallahan: @dosas @lhellebr while this repo did start out as more foreman-focused, it grew/matured almost entirely downstream/Satellite environment. Due to this, there is quite a bit of Satellite-specific language in the codebase. Is this something that could be changed, likely yes. Is this change something we're likely to prioritize in the near-term, probably not.
However, that doesn't mean that this repo's design can't be re-evaluated when changes like this might expose a design limitation.

tests/foreman/api/test_permission.py
JacobCallahan: @dosas is correct here that the missing presence of Satellite rpms isn't the most accurate test of whether we're in an upstream/downstream environment or not.

dosas: @lhellebr To answer your original questions: Yes there are more conditions because there are also more permissions that were added.

IMO the definition of `is_upstream` is out of scope of this PR.

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/api/test_permission.py -k test_positive_search

Satellite-QE:
**PRT Result**
```
Build Number: 6545
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/api/test_permission.py -k test_positive_search --external-logging
Test Result : ========== 3 passed, 13 deselected, 114 warnings in 812.58s (0:13:32) ==========
```

dosas:
@pondrejk Why did you remove the labels?

pondrejk:
hi @dosas, I checked out this branch locally and ran some tests from the affected files, all seems well, prt also passes so I'm ok with merging this, thanks for the contribution. I reckoned that downstream branches would not be of interest to you, so I changed the labels. If you need those cherry picks I can put them back

pondrejk (approved):

dosas:
> hi @dosas, I checked out this branch locally and ran some tests from the affected files, all seems well, prt also passes so I'm ok with merging this, thanks for the contribution. I reckoned that downstream branches would not be of interest to you, so I changed the labels. If you need those cherry picks I can put them back

I would require 6.15.z please

PR #14519, opened 2024-03-26
Add coverage for BZ1964539

rmynar:
Adding a test case to cover BZ1964539
Satellite and any of it's dependencies should not require package nmap-ncat.
The test is parametrized, so it can be easily extended to test different packages.


rmynar:
trigger: test-robottelo
pytest: tests/foreman/installer/test_installer.py -k 'test_weak_dependency or test_content_guarded_distributions_option or test_capsule_installation or test_foreman_rails_cache_store'

rmynar:
trigger: test-robottelo
pytest: tests/foreman/installer/test_installer.py -k test_weak_dependency

tests/foreman/installer/test_installer.py
```diff
@@ -1808,3 +1813,32 @@ def test_satellite_installation(installer_satellite):
     assert installer_satellite.execute('rpm -q foreman-redis').status == 0
     settings_file = installer_satellite.load_remote_yaml_file(FOREMAN_SETTINGS_YML)
     assert settings_file.rails_cache_store.type == 'redis'
+
+
+@pytest.mark.parametrize('package', ['nmap-ncat'])
+def test_weak_dependency(sat_non_default_install, package):
+    """Check if Satellite and its (sub)components do not require certain (potentianally insecure) packages
```
Griffin-Sullivan, line 1820:
```suggestion
    """Check if Satellite and its (sub)components do not require certain (potentially insecure) packages
```

Griffin-Sullivan (commented):
Seems like the test will cover what was in the BZ. My main concern is if this test is really necessary since we are only testing the one packages. I'd love to find a way to make this have a bigger impact with more than just one package. @lpramuk @jameerpathan111 @ehelms @evgeni @ekohl any thoughts around this?

tests/foreman/installer/test_installer.py
rmynar: Typo resolved. Thanks

ekohl:
I wonder what the best way of approaching this is. I'd imagine you can test this out without installing things which is a lot faster. How about this:

* Get a plain RHEL
* Assert `nmap` is not installed
* Set up repositories
* `assert 'nmap' not in run('dnf --setopt=install_weak_deps=False --assumeno install satellite')`
* `assert 'nmap' not in run('dnf --setopt=install_weak_deps=True --assumeno install satellite')`

You may need to tune the package list and you can probably parametrize it, or even skip the `False` case but it wouldn't save a lot of time in this case while it may make debugging easier.

Another approach would be to work with delivery to get this in a regular pipeline. They will investigate the failures of those and prevent it from even promoting in the first place.

lhellebr:
Why test using `dnf remove`?

Also, Ewould has a point, it may be better to check for package presence somewhere else. My suggestion would be logs, the same way installer log is checked. Additionally, at the time when the logs are checked (whatever it is), you could maybe directly check for rpm presence.

rmynar:
I would definitely like to make this bigger, that's why it is parametrized.

Current solution uses the fixture _sat_non_default_install_ which is common for more testcases in this module. So there's no significant additional time cost.

As said in the BZ, the unwanted dependency does not usually come from the main package itself (satellite.*rpm), but from some other subcomponent/feature which is not installed by default. This is the reason why I chose the opposite approach, i.e. enable various non-default subcomponents/features, attempt to remove unwanted package(s) and check if it does not remove any of our packages. As a benefit we also test if the non-default subcomponents/features can be installed during the fixture setup.

lpramuk:
Currently `nmap` is a weak dependency of `foreman-discovery-image`.
The test is going to ensure that it is going to stay that way:
customers are able to uninstall nmap if they care about nmap not present

lpramuk (approved):

tests/foreman/installer/test_installer.py
```diff
@@ -1808,3 +1813,32 @@ def test_satellite_installation(installer_satellite):
...
+    """Check if Satellite and its (sub)components do not require certain (potentially insecure) packages
+
+    :id: c7988920-2f8c-4646-bde9-8823a3ca96bb
+
+    :steps:
+        1. Use satellite with non-default setup (for 'nmap-ncat' enable foreman discovery plugin and install foreman-discovery-image)
+        2. Attempt to remove the package
+
+    :expectedresults:
+        1. The package can be either not installed or can be removed without removing any Satellite or Foreman packages
```
ogajduse, line 1829: Running the test for `nmap-ncat` package makes sense only if the package was installed.
What benefit do you see in running the test if  `nmap-ncat` is not installed on the system?

Note: If you decide to modify the `expectedresults` field, the last assertion also needs to be updated.

lpramuk: +1
You cannot test in a single test both that rpm is not present and that rpm gets removed.
This test is targeted to test that weak dependencies stay weak by removal transaction.

So please lets not mix things together, there should be assert only about removal so that we  will know in future FDI no longer requires nmap (won't get installed anymore) 

With broad `OR` assert we won't never get to know.

lpramuk: I've change my mind after discussion with @rmynar 
The test not only tests weak dependency but also no dependency - we do not care if it is `no` OR `weak` dependency as long as it is so we are good.

rmynar:
I slightly updated test docstring to make it more clear that current solution tries to be as close as possible to customer experience. As reported in the BZ, the customer had an existing instance of Satellite and he could not remove potentially harmful package because it would also remove some satellite or foreman package(s).
I added pit marker so we're able to catch dependency error earlier.

lpramuk:
> I wonder what the best way of approaching this is. I'd imagine you can test this out without installing things which is a lot faster. How about this:
> 
>     * Get a plain RHEL
> 
>     * Assert `nmap` is not installed
> 
>     * Set up repositories
> 
>     * `assert 'nmap' not in run('dnf --setopt=install_weak_deps=False --assumeno install satellite')`
> 
>     * `assert 'nmap' not in run('dnf --setopt=install_weak_deps=True --assumeno install satellite')`
> 
> 
> You may need to tune the package list and you can probably parametrize it, or even skip the `False` case but it wouldn't save a lot of time in this case while it may make debugging easier.
> 
> Another approach would be to work with delivery to get this in a regular pipeline. They will investigate the failures of those and prevent it from even promoting in the first place.

This PR covers the BZ
if you are interested in generic packaging testing then this is definitely out of scope of this PR

ogajduse (approved):
This test fully covers the related BZ. Radek chose a different approach from what @ekohl suggested, but it still covers the product bug. The test will pass also in a situation when this dependency is not present in the system or if it got renamed. That is a downside of the current form of the test. 
LGTM otherwise. I see no reason to block it at the moment.

PR #14538, opened 2024-03-27
Add case for special chars in HTTP proxy password

vsedmik:
### Problem Statement
We are missing coverage for a [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1844840) where the HTTP proxy is created with a user containing special characters in his password.

### Solution
Not only we need to create the proxy at the Satellite side, we also need to create the user at the proxy side to verify the content operations actually work. In the end I decided to create such users dynamically during the test run, so I needed to add a new `ProxyHost` class with a few functions to handle that. The class is instantiated in a session-scoped fixture and used in test and other fixture.

### PRT test Cases example
trigger: test-robottelo
pytest: tests/foreman/ui/test_http_proxy.py -k special

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -170,6 +170,10 @@ class IPAHostError(Exception):
     pass
 
 
+class ProxyHostError(Exception):
+    pass
+
+
 class ContentHost(Host, ContentHostMixins):
     run = Host.execute
     default_timeout = settings.server.ssh_client.command_timeout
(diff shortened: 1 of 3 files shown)


vsedmik:
trigger: test-robottelo
pytest: tests/foreman/ui/test_http_proxy.py -k special

pondrejk (approved):

robottelo/hosts.py
```diff
@@ -2545,3 +2549,42 @@ def remove_user_from_usergroup(self, member_username, member_group):
         )
         if result.status != 0:
             raise IPAHostError('Failed to remove the user from user group')
+
+
+class ProxyHost(Host):
+    """Class representing HTTP Proxy host"""
+
+    def __init__(self, url, **kwargs):
+        self._conf_dir = '/etc/squid/'
```
Griffin-Sullivan, line 2558:
```suggestion
        self._conf_dir = '/etc/squid'
```
Strings below would look better if you did this and then called `f'{self._conf_dir}/passwd'`

Griffin-Sullivan (approved):

PR #14573, opened 2024-03-29
extend the timeout for poor network

yanpliu:
### Problem Statement
when the network is poor, it will not enough time to get the correct message and mapping info. so extend the timeout


omkarkhatavkar (approved):

Gauravtalreja1 (approved):

jyejare:
@yanpliu Looking at the timeout you have set to `20` , your network doesnt look slow it must be very fast :)

In general we have bigger timeouts like 600 or more.

PR #14568, opened 2024-03-28
Use fixtures instead of finalizers

dosas:
### Problem Statement

The teardown order was not correct
and host was not deleted before hostgroup
https://github.com/pytest-dev/pytest/issues/10023#issuecomment-1148514756

### Solution

Use fixtures

PR #13620, opened 2024-01-05
Add a test for parameter precedence on the host

Gauravtalreja1:
### Problem Statement
Missing test to verify parameter precedence on the host

### Solution
Applying the component audit feedback to cover this scenario and adding test to verify parameter precedence on the host for Global/HostGroup/Host parameters

PR #14620, opened 2024-04-04
Ensure correct finalizer order

dosas:
### Problem Statement

the host was not deleted before the hostgroup

see: https://github.com/pytest-dev/pytest/issues/10023#issuecomment-1148514756

### Solution

Execute host finalizer after hostgroup finalizer

### Note

This should replace https://github.com/SatelliteQE/robottelo/pull/14568

### DIFF
--- a/tests/foreman/api/test_parameters.py
+++ b/tests/foreman/api/test_parameters.py
@@ -26,8 +26,8 @@ def test_positive_parameter_precedence_impact(
 
     :steps:
         1. Create Global Parameter
-        2. Create host and verify global parameter is assigned
-        3. Create Host Group with parameter
+        2. Create Host Group with parameter
+        3. Create host and verify global parameter is assigned
         4. Assign hostgroup to host created above and verify hostgroup parameter is assigned.
         5. Add parameter on the host directly, and verify that this should take precedence
             over Host group and Global Parameter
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan (approved):
Interesting how the finalizers work as a stack. My recommendation would be to flip the order of creation. Make the HG then the host and leave finalizers right after each create. Although this works too.

tests/foreman/api/test_parameters.py
```diff
@@ -51,6 +50,7 @@ def test_positive_parameter_precedence_impact(
         group_parameters_attributes=[{'name': param_name, 'value': param_value}],
     ).create()
     request.addfinalizer(hg.delete)
+    request.addfinalizer(host.delete)
```
Gauravtalreja1, line 53: I think here host should be deleted before hg, as host is linked to hg.
And, Here instead of 2-request.addfinalizer's, I'd suggest using a finalizer method like in the original PR for this test, which is used work correctly for teardown
https://github.com/SatelliteQE/robottelo/pull/13620

ogajduse: @Gauravtalreja1 
To your point about the order of deletion:
```
Finalizers are executed in a first-in-last-out order.
```
https://docs.pytest.org/en/latest/how-to/fixtures.html#note-on-finalizer-order

What is your motivation behind the suggestion to use a single finalizer method?

dosas:
> Interesting how the finalizers work as a stack. My recommendation would be to flip the order of creation. Make the HG then the host and leave finalizers right after each create. Although this works too.

> My recommendation would be to flip the order of creation.
Wouldn't that change the test logic? https://github.com/SatelliteQE/robottelo/blob/cb296b3f6498c13dc4497cca3d77b459ec7e91b9/tests/foreman/api/test_parameters.py#L28-L30

tests/foreman/api/test_parameters.py
Gauravtalreja1: Just to simplify this, Its better to write a single method with the execution order we provide, instead of pytest to decide this order for finalizers

dosas: What about the reasons for changing it to finalizers in https://github.com/SatelliteQE/robottelo/pull/14036 ?

>  so basically when some assertion defined in the test before the finalizer part fails, finalizer is not added and so not executed.

This is also why I first wanted to go for the fixture approach.

JacobCallahan (changes requested):
I overall like this approach better than the fixture-heavy one. However, I will block on changing creation order (and yes the test logic order) to create the hostgroup first, then add its finalizer before creating the host. This is the best balance between ensuring finalizer execution order and actual execution if something goes wrong between entity creation and finalizer addition.

Satellite-QE:
**PRT Result**
```
Build Number: 6453
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/api/test_parameters.py --external-logging
Test Result : ================== 1 passed, 9 warnings in 655.80s (0:10:55) ===================
```

JacobCallahan (approved):
ACK, thanks for taking this on!

PR #14707, opened 2024-04-10
Update PRT Label Remove

shweta83:
### Problem Statement
When a PRT runs on a PR, it adds a PRT-Passed or PRT-Failed label to it. When we push a new commit to the PR, the labels are removed and the PR looks all green even if the PRT failed earlier.

### Solution
Updated the workflow to only remove the PRT-label on new commit to the PR if the PRT was passed in previous run, and keep it as it is otherwise.

### DIFF
--- a/.github/workflows/prt_labels.yml
+++ b/.github/workflows/prt_labels.yml
@@ -8,7 +8,7 @@ jobs:
     prt_labels_remover:
       name: remove the PRT label when amendments or new commits added to PR
       runs-on: ubuntu-latest
-      if: "(contains(github.event.pull_request.labels.*.name, 'PRT-Passed') || contains(github.event.pull_request.labels.*.name, 'PRT-Failed'))"
+      if: "(contains(github.event.pull_request.labels.*.name, 'PRT-Passed'))"
       steps:
         - name: Avoid the race condition as PRT result will be cleaned
           run: |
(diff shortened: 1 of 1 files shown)


omkarkhatavkar (approved):
ACK

lpramuk (approved):

lpramuk:
@shweta83 
Thanks for this.
Consider adding labels for cherrypicking.

.github/workflows/prt_labels.yml
```diff
@@ -35,7 +35,7 @@ jobs:
                 repo: context.repo.repo,
                 issue_number: prNumber,
               });
-              const labelsToRemove = ['PRT-Failed', 'PRT-Passed'];
```
Gauravtalreja1, line 38: Could you elaborate on ` labels are removed and the PR looks all green even if the PRT failed earlier.`?
And, shouldn't we remove both of thses labels if new commit is added/amend'ed?

shweta83: If some check is failed on the PR, it shows as crossed on the PR but when all the PRT labels are removed then it shows the PR as green in the Pull Requests list. With this change, we are trying to keep the PR in failed state unless PRT is passed on it.

Gauravtalreja1: Yes, the change makes sense, and it shows PR as green with other checks but not PRT which isn't run yet, and we check PRT results only.
In earlier approach we kept no PRT results labels after new commit added, and now we keep PRT failed/passed until we run PRT after new commit is added (iiuc). Although I'd prefer the earlier approach, I am fine with either one, and as @omkarkhatavkar already ack'ed, so I wouldn't block this.

Gauravtalreja1 (approved):

PR #14772, opened 2024-04-15
Fix Provisioning Tests

shubhamsg199:
Fixing failing provisioning tests

### DIFF
--- a/tests/foreman/api/test_provisioning.py
+++ b/tests/foreman/api/test_provisioning.py
@@ -138,20 +138,32 @@ def test_rhel_pxe_provisioning(
     # Host is not blank anymore
     provisioning_host.blank = False
 
+    provisioning_host.key_filename = None
     # Wait for the host to be rebooted and SSH daemon to be started.
     provisioning_host.wait_for_connection()
+
+    # For RHEL9 since root login is disabled
+    disable_ssh_cmd = (
+        'echo -e "\nPermitRootLogin yes" >> /etc/ssh/sshd_config; systemctl restart sshd'
+    )
+    # Run a command on the host using REX to verify that Satellite's SSH key is present on the host
+    template_id = (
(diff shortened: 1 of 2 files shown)


tests/foreman/api/test_provisioning.py
```diff
@@ -534,20 +522,32 @@ def test_rhel_pxe_provisioning_fips_enabled(
     # Host is not blank anymore
     provisioning_host.blank = False
 
+    provisioning_host.key_filename = None
     # Wait for the host to be rebooted and SSH daemon to be started.
     provisioning_host.wait_for_connection()
 
+    # For RHEL9 since root login is disabled
```
jameerpathan111, line 529: I see a lot of code repetition, and we'll probably need to use it in other tests too, so it's probably better to move this to https://github.com/SatelliteQE/robottelo/blob/master/robottelo/hosts.py#L1756 and only for RHEL9 Satellite.

jameerpathan111 (changes requested):

tests/foreman/api/test_provisioning.py
```diff
@@ -138,20 +138,32 @@ def test_rhel_pxe_provisioning(
...
+    disable_ssh_cmd = (
```
shweta83, line 146: This will be true for all your provisioning tests for RHEL9 so can we keep this common for all the  tests?

Gauravtalreja1: @shubhamsg199 How about creating a new helper to run and verify this REX job to enable or permit root login in ProvisioningSetup host_helpers (run just for EL9) and run other REX task for RHEL7/8/9 to verify hostname as we do now?

shubhamsg199:
Closing this as not needed anymore. broker 0.5 has been released which fixes the isssue

PR #14803, opened 2024-04-17
Fix test_positive_clone_backup

jameerpathan111:
### Problem Statement
- `test_positive_clone_backup` is failing with `Satellite is already installed. Satellite-clone should be run on a clean RHEL machine.` error.

### Solution
- Only enable the Satellite module but do not install the package.

### Related Issues
- SAT-24576

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -1739,8 +1739,8 @@ class Capsule(ContentHost, CapsuleMixins):
         self._cli._configured = True
         return self._cli
 
-    def install_satellite_or_capsule_package(self):
-        """Install Satellite/Capsule package.
+    def enable_satellite_or_capsule_module_for_rhel8(self):
+        """Enable Satellite/Capsule module for RHEL8.
         Note: Make sure required repos are enabled before using this.
         """
         if self.os_version.major == 8:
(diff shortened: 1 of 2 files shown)


jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_clone.py -k "test_positive_clone_backup[include_pulp-offline]"

pondrejk (approved):

robottelo/hosts.py
```diff
@@ -1739,7 +1739,7 @@ def cli(self):
         self._cli._configured = True
         return self._cli
 
-    def install_satellite_or_capsule_package(self):
+    def install_satellite_or_capsule_package(self, install_package=True):
```
jyejare, line 1742: If installing the package is optional then we should modify the name of this function as its made/sounds to install the package only.

jyejare (commented):
ACK pending comment.

robottelo/hosts.py
jameerpathan111: I was thinking the same but I'm having a hard time coming up with something new.

jameerpathan111: got any suggestions?

jameerpathan111: how about `optional_satellite_or_capsule_package_install`

jameerpathan111: updated the pr. I split the helper into `install_satellite_or_capsule_package` and `enable_satellite_or_capsule_module`

jameerpathan111: PRT has passed last time so I'm not retriggering it again as the change was not destructive.

jyejare (approved):

PR #14837, opened 2024-04-19
Update Provisioning test workflow names

shubhamsg199:
### Problem Statement
The workflow names were recently updated and now they have `iac` prepended to the them,  the tests were failing because we had old workflow names in the test.

### Solution
Update them to the correct workflow names


shweta83 (approved):
Ack

ogajduse:
Can you consider making the WF name configurable through settings?

shubhamsg199:
> Can you consider making the WF name configurable through settings?

Updated

Gauravtalreja1 (approved):

PR #14842, opened 2024-04-19
update scope of fixtures from hosts-content longrun test_inc_updates

vijaysawant:
### Problem Statement
`longrun/test_inc_updates.py` has 2 tests, module scope fixture creates one organization and same organization has been used in two test cases, after uploading manifest second time test was failing due to below error message

**test name:** test_positive_incremental_update_time
**Error:**
`['Owner has already imported from another subscription management application. The following conflicts were found: [ DISTRIBUTOR_CONFLICT ]']`

### Solution
changed scope of fixture of one test case so it will not use previously created organization from same test session.

### Related Issues
N/A

### PRT test Cases example
trigger: test-robottelo
pytest: tests/foreman/longrun/test_inc_updates.py

### DIFF
--- a/tests/foreman/longrun/test_inc_updates.py
+++ b/tests/foreman/longrun/test_inc_updates.py
@@ -231,7 +231,7 @@ def test_positive_noapply_api(
 
 
 @pytest.mark.tier3
-def test_positive_incremental_update_time(module_target_sat, module_sca_manifest_org):
+def test_positive_incremental_update_time(target_sat, function_sca_manifest_org):
     """Incremental update should not take a long time.
 
     :id: a9cdcc58-2d10-42cf-8e24-f7bec3b79d6b
@@ -256,33 +256,31 @@ def test_positive_incremental_update_time(module_target_sat, module_sca_manifest
 
     """
     # create content view
-    cv = module_target_sat.cli_factory.make_content_view(
-        {'organization-id': module_sca_manifest_org.id}
-    )
+    cv = target_sat.cli_factory.make_content_view({'organization-id': function_sca_manifest_org.id})
     repo_sync_timestamp = (
         datetime.utcnow().replace(microsecond=0) - timedelta(seconds=1)
     ).strftime('%Y-%m-%d %H:%M')
     # setup rh repositories, add to cv, begin sync
     for _repo in ['rhel8_bos', 'rhst8', 'rhsclient8']:
-        rh_repo_id = module_target_sat.api_factory.enable_rhrepo_and_fetchid(
+        rh_repo_id = target_sat.api_factory.enable_rhrepo_and_fetchid(
             basearch=DEFAULT_ARCHITECTURE,
-            org_id=module_sca_manifest_org.id,
+            org_id=function_sca_manifest_org.id,
             product=PRDS['rhel8'],
             repo=REPOS[_repo]['name'],
             reposet=REPOSET[_repo],
             releasever=REPOS[_repo]['releasever'],
         )
-        module_target_sat.cli.ContentView.add_repository(
+        target_sat.cli.ContentView.add_repository(
             {
                 'id': cv['id'],
-                'organization-id': module_sca_manifest_org.id,
+                'organization-id': function_sca_manifest_org.id,
                 'repository-id': rh_repo_id,
             }
         )
-        module_target_sat.api.Repository(id=rh_repo_id).sync(synchronous=False)
+        target_sat.api.Repository(id=rh_repo_id).sync(synchronous=False)
 
     # wait for all repo sync tasks
-    sync_tasks = module_target_sat.wait_for_tasks(
+    sync_tasks = target_sat.wait_for_tasks(
         search_query=(
             'label = Actions::Katello::Repository::Sync'
             f' and started_at >= "{repo_sync_timestamp}"'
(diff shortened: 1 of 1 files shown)


vijaysawant:
trigger: test-robottelo
pytest: tests/foreman/longrun/test_inc_updates.py

Satellite-QE:
**PRT Result**
```
Build Number: 6623
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/longrun/test_inc_updates.py --external-logging
Test Result : ================= 2 passed, 112 warnings in 1126.00s (0:18:46) =================
```

damoore044 (approved):
nice fix! I could not reproduce either

tests/foreman/longrun/test_inc_updates.py
```diff
@@ -231,7 +231,7 @@ def test_positive_noapply_api(
 
 
 @pytest.mark.tier3
-def test_positive_incremental_update_time(module_target_sat, module_sca_manifest_org):
+def test_positive_incremental_update_time(target_sat, function_sca_manifest_org):
```
shweta83, line 234: Why are you using sca enabled manifest?

vijaysawant: - During 6.14.z TFA, I find this test was failing due to reimporting of manifest in same orgnaization (created using module scope fixture)
- I just want to keep `*_sca_manifest_org` as it is so just changed scope of fixture.

Gauravtalreja1: @vijaysawant I had the same question while reading this, but this issue should be mentioned in the PR description in problem/solution.
 And could you be little more descriptive in PR title and commit messge, which could help you in future to identity the cause of this failures?

vijaysawant: Sure @Gauravtalreja1 , this PR initially open for testing purpose and wanted to see behaviour of when test running in Jenkins (as same changes/ same tests in `6.14` & `6.15` behave differently if I talk about test results from latest build, we @ColeHiggins2 and @vsedmik monitor robottelo logs closely and took decision to test changes remotely)

jameerpathan111: @vijaysawant I agree with Shweta here, this can be easily fixed by using `module_entitlement_manifest_org`.

shweta83: Can you try `function_entitilement_manifest_org` here? I am sure that would also work.

jameerpathan111: why use `function_entitilement_manifest_org` when we can use existing org with manifest imported i.e `module_entitlement_manifest_org` here? That's the point of using module level fixtures, reuse the existing setup so we can reduce code duplication, test setup time and it's easier to maintain.

vijaysawant: - @jameerpathan111 if you follow thread I have mentioned "import new manifest in already created organization leads with an error" so I have decided to keep "*_sca_manifest_org" fixture, just by changing it's scope so it would create new (`function scope`) org
- Previous author of test used `module_sca_manifest_org` and it's reason is not mentioned why they use sca manifest org, so just to make test case run with same functionality I am using `function_sca_manifest_org`
Let me know still you want to make this change or not, it's working.
@shweta83 @jameerpathan111

jameerpathan111: it's fine by me

jameerpathan111 (approved):

tests/foreman/longrun/test_inc_updates.py
vsedmik: @jameerpathan111 @shweta83, from what I understood from the subscription team, we should move away from the entitlement manifests/orgs rather sooner than later. So, if there is no specific reason to use the entitlement manifest (e.g. testing subscription attach etc.), then we should rather go SCA.

shweta83: This change was going on 6.14.z branch so to keep the things consistent, I suggested the above change.

jameerpathan111: @vsedmik I understand but my main point was using a module-level fixture in both the tests, whether it's `module_sca_manifest_org` or `module_entitlement_manifest_org`. Anyway, it was more of a suggestion and I'm fine with the current approach too.

vijaysawant: @jameerpathan111 due to module scope fixture test setup fails- because it creates organization and reuse in another test, the second test case uploads newly created manifest and due to which it end up with en Error
`['Owner has already imported from another subscription management application. The following conflicts were found: [ DISTRIBUTOR_CONFLICT ]']`
May be I unable to explain it in correct way, @synkd @vsedmik @ColeHiggins2 were discussing this issue last week, so we took dicision to change scope of fixture (module to function) rather change functionality (sca to non-sca)

Satellite-QE:
**PRT Result**
```
Build Number: 6683
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/longrun/test_inc_updates.py --external-logging
Test Result : ================= 2 passed, 111 warnings in 1242.64s (0:20:42) =================
```

jameerpathan111 (approved):

vsedmik (approved):

PR #14856, opened 2024-04-22
Add coverage for clone from el8 to el9

lpramuk:
### Problem Statement
Stream/6.16 supports both el8 and el9 versions. We need to ensure that customers are able to upgrade underlying OS using `satellite-clone` tool

### Solution
If we are running for el8 then parametrize target `sat_ready_rhel`  to be both 8 and 9 extending matrix of testing datapoints. 

### Related Issues
SAT-24714

### DIFF
--- a/tests/foreman/destructive/test_clone.py
+++ b/tests/foreman/destructive/test_clone.py
@@ -15,13 +15,18 @@ import pytest
 
 from robottelo import constants
 from robottelo.config import settings
-from robottelo.hosts import Satellite
+from robottelo.hosts import Satellite, get_sat_rhel_version
 
 SSH_PASS = settings.server.ssh_password
 pytestmark = pytest.mark.destructive
 
 
 @pytest.mark.e2e
+@pytest.mark.parametrize(
+    "sat_ready_rhel",
+    [8, 9] if get_sat_rhel_version().major < 9 else [9],
+    indirect=True,
+)
 @pytest.mark.parametrize('backup_type', ['online', 'offline'])
 @pytest.mark.parametrize('skip_pulp', [False, True], ids=['include_pulp', 'skip_pulp'])
 def test_positive_clone_backup(


lpramuk:
trigger: test-robottelo

tests/foreman/destructive/test_clone.py
```diff
@@ -15,13 +15,18 @@
...
 SSH_PASS = settings.server.ssh_password
 pytestmark = pytest.mark.destructive
 
 
 @pytest.mark.e2e
+@pytest.mark.parametrize(
+    "sat_ready_rhel",
+    [8, 9] if get_sat_rhel_version().major < 9 else [9],
+    indirect=True,
+)
```
jameerpathan111, line 29: @lpramuk I'm not 100% sure but you'll probably  have to fix [sat_ready_rhel.download_repofile](https://github.com/SatelliteQE/robottelo/pull/14856/files#diff-0136ad9f3a21adf5716d5013efd837789362d2624fb5c83f55a226650e0c0470R91) part as well. I think it'll use rhel version of `target_satellite` or from server config and might get the wrong rhel version repo. To avoid this it's better to pass the required RHEL version explicitly.

jameerpathan111 (changes requested):

tests/foreman/destructive/test_clone.py
jameerpathan111: `download_repofile` doesn't support passing rhel version atm so you'll have to find another way or update it to support that.😅

lpramuk: This bubbles thru and gets to `def _os_release(self):` which reads `/etc/release` 
Safe I'd say :-)

jameerpathan111: ah cool

jameerpathan111 (approved):

Satellite-QE:
**PRT Result**
```
Build Number: 6640
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/destructive/test_clone.py::test_positive_clone_backup --external-logging
Test Result : ================ 8 passed, 1624 warnings in 25604.39s (7:06:44) ================
```

PR #14866, opened 2024-04-23
Copy config files for FAM to server

Griffin-Sullivan:
### Problem Statement
Some tests are failing in production because we are missing live config settings for them. We also created some tech debt on the last PR, because we were doing inline config replacement with `sed`.

### Solution
Host the config files ourselves and evaluate them with dynaconf before sending them to the server.

I won't be able to test this in PRT at the moment, but will open a jenkins PR for the right config settings. The current templates I copied into here are directly from https://github.com/theforeman/foreman-ansible-modules/tree/develop/tests/test_playbooks/vars

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -883,12 +883,17 @@ class ContentHost(Host, ContentHostMixins):
         """Get a remote file from the broker virtual machine."""
         self.session.sftp_read(source=remote_path, destination=local_path)
 
-    def put(self, local_path, remote_path=None):
+    def put(self, local_path, remote_path=None, temp_file=False):
         """Put a local file to the broker virtual machine.
         If local_path is a manifest object, write its contents to a temporary file
         then continue with the upload.
         """
-        if 'utils.manifest' in str(local_path):
+        if temp_file:
+            with NamedTemporaryFile(dir=robottelo_tmp_dir) as content_file:
+                content_file.write(str.encode(local_path))
+                content_file.flush()
+                self.session.sftp_write(source=content_file.name, destination=remote_path)
+        elif 'utils.manifest' in str(local_path):
             with NamedTemporaryFile(dir=robottelo_tmp_dir) as content_file:
                 content_file.write(local_path.content.read())
                 content_file.flush()
--- a/conf/fam.yaml.template
+++ b/conf/fam.yaml.template
@@ -1,0 +1,101 @@
+FAM:
+  SERVER:
+    # Parameter for all tests
+    foreman_username: admin
+    foreman_password: "changeme"
+    foreman_server_url: "https://foreman.example.com"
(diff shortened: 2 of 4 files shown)


rplevka:
I'd suggest to keep all config related to `FAM` under a single config file, probably `fam.yaml`.
then have separate keys to denote the individual "sub" sections of it.
```
fam.yaml
|_ server: {...}
|_ compute_resource {...}
```

rplevka (changes requested):
pending config consolidation under a single file

robottelo/hosts.py
```diff
@@ -883,12 +883,17 @@ def get(self, remote_path, local_path=None):
...
-    def put(self, local_path, remote_path=None):
+    def put(self, local_path, remote_path=None, temp_file=False):
         """Put a local file to the broker virtual machine.
         If local_path is a manifest object, write its contents to a temporary file
         then continue with the upload.
         """
-        if 'utils.manifest' in str(local_path):
+        if temp_file:
+            with NamedTemporaryFile(dir=robottelo_tmp_dir) as content_file:
+                content_file.write(str.encode(local_path))
```
rplevka, line 893: what is going on here? how is this different from the `utils.manifest` case?

Griffin-Sullivan: I honestly don't know how the path is being passed for the case where the local path contains `utils.manifest`. For my instance, I can at least say you aren't actually passing a path but rather a string. Whereas the manifest case does some sort of `read()` function. Maybe @JacobCallahan or @synkd could give us some insight here?

rplevka (approved):

Griffin-Sullivan:
Need to wait for https://github.com/theforeman/foreman-ansible-modules/pull/1732 to be merged now

robottelo/hosts.py
synkd: The path is being passed by the `upload_manifest` helper [1], specifically in cases where we're uploading a manifest via CLI, rather than imported directly via API. In those cases, we need to pass the path of the manifest file needs to be `put` onto the Satellite before being imported with hammer. The third bullet point in this comment [2] describes these scenarios.

I think that, in cases where Manifester times out and we fail over to cloned manifests, the contents of the cloned manifest object need to be written to a file before transferring that file to the Satellite, whereas with Manifester, the file would already be present on the local file system. 

I hope this provides some measure of clarity. It is admittedly rather confusing.

[1] https://github.com/SatelliteQE/robottelo/blob/bba4931f4aec31af9288f58e0bff3cbabc9cc6ca/robottelo/host_helpers/satellite_mixins.py#L137
[2] https://github.com/SatelliteQE/robottelo/pull/12605#issuecomment-1718143962

synkd (approved):

PR #14896, opened 2024-04-26
Lint: Black -> Ruff formatter

ogajduse:
### Problem Statement
#14781 attempts to upgrade some of our pre-commit hooks including black. The new version of [black formats](https://results.pre-commit.ci/run/github/4000279/1714059668.imtZ-VkMSyK3E5Wo_AicJw) 150+ files. Given the patch size, I propose switching to Ruff formatter as the patch size is comparable to the patch for the new version of black.

#### Reason to switch to Ruff
(copy&paste from one of the comments in this PR)
Astral's page sums it up pretty well. [astral.sh/blog/the-ruff-formatter](https://astral.sh/blog/the-ruff-formatter)

I would like us to switch to Ruff mainly because of its speed compared to black. It is incredibly fast and the compatibility with Black is 99%. Looking at the patch, it might not look like 99%, but taking into account how big our project is and what changes were made, it is not that bad.

### Solution
* Remove the black formatter and substitute it with Ruff formatter,
* bump the version of ruff pre-commit hook.

### Related Issues
Related PR: #14781

### DIFF
--- a/tests/foreman/api/test_errata.py
+++ b/tests/foreman/api/test_errata.py
@@ -123,7 +123,7 @@ def _validate_errata_counts(host, errata_type, expected_value, timeout=120):
         sleep(5)
     else:
         pytest.fail(
-            'Host {} contains {} {} errata, but expected to contain ' '{} of them'.format(
+            'Host {} contains {} {} errata, but expected to contain {} of them'.format(
                 host.name,
                 host.content_facet_attributes['errata_counts'][errata_type],
                 errata_type,
@@ -149,7 +149,7 @@ def _fetch_available_errata(host, expected_amount=None, timeout=120):
         errata = host.errata()
     else:
         pytest.fail(
-            'Host {} contains {} available errata, but expected to ' 'contain {} of them'.format(
+            'Host {} contains {} available errata, but expected to contain {} of them'.format(
                 host.name,
                 len(errata['results']),
                 expected_amount if not None else 'No expected_amount provided',
--- a/tests/foreman/api/test_errata.py
+++ b/tests/foreman/api/test_errata.py
@@ -11,6 +11,7 @@
 :CaseImportance: High
 
 """
+
 # For ease of use hc refers to host-collection throughout this document
 from time import sleep, time
 
@@ -122,8 +123,7 @@ def _validate_errata_counts(host, errata_type, expected_value, timeout=120):
         sleep(5)
     else:
         pytest.fail(
-            'Host {} contains {} {} errata, but expected to contain '
-            '{} of them'.format(
+            'Host {} contains {} {} errata, but expected to contain ' '{} of them'.format(
                 host.name,
                 host.content_facet_attributes['errata_counts'][errata_type],
                 errata_type,
@@ -149,8 +149,7 @@ def _fetch_available_errata(host, expected_amount=None, timeout=120):
         errata = host.errata()
     else:
         pytest.fail(
-            'Host {} contains {} available errata, but expected to '
-            'contain {} of them'.format(
+            'Host {} contains {} available errata, but expected to ' 'contain {} of them'.format(
                 host.name,
                 len(errata['results']),
                 expected_amount if not None else 'No expected_amount provided',
@@ -400,7 +399,7 @@ def package_applicability_changed_as_expected(
         output = host.execute(f'rpm -q {package_basename}').stdout
         current_package = output[:-1]
         assert package_basename in current_package
-        if current_package == package_filename:
+        if current_package == package_filename:  # noqa: SIM108
             # we have already checked if applicable package count changed,
             # in case the same version as prior was installed and present.
             prior_package = None  # package must not have been present before this modification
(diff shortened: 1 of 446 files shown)


synkd (commented):
A couple of small, non-blocking suggestions, but looks good to me overall.

tests/foreman/ui/test_ansible.py
```diff
@@ -182,7 +183,7 @@ def _finalize():
         result = rhel_contenthost.execute(command)
         assert result.status == 0, f'Failed to register host: {result.stderr}'
         target_host = rhel_contenthost.nailgun_host
-        default_value = '[\"test\"]'
+        default_value = '["test"]'
```
Gauravtalreja1, line 186: I'd ask @shweta83 to confirm if this could change test behavior; she originally wrote this test and tried this string.

tests/foreman/api/test_notifications.py
```diff
@@ -231,14 +232,14 @@ def long_running_task(target_sat):
...
-    sql_date_2_days_ago = "now() - INTERVAL \'2 days\'"
+    sql_date_2_days_ago = "now() - INTERVAL '2 days'"
     result = target_sat.execute(
         "su - postgres -c \"psql foreman postgres <<EOF\n"
         "UPDATE foreman_tasks_tasks "
         f"SET start_at = {sql_date_2_days_ago}, "
         f" started_at = {sql_date_2_days_ago}, "
         f" state_updated_at = {sql_date_2_days_ago} "
-        f"WHERE id=\'{job['task']['id']}\';\nEOF\n\" "
+        f"WHERE id='{job['task']['id']}';\nEOF\n\" "
```
Gauravtalreja1, line 242: is this expected in this DB query?

ogajduse: Not sure, but Python sees both as the same string:
```python
[ins] In [1]: job = {'task':{'id': 123}}

[ins] In [2]: f"WHERE id=\'{job['task']['id']}\';\nEOF\n\" " == f"WHERE id='{job['task']['id']}';\nEOF\n\" "
Out[2]: True

[ins] In [3]:
```

tests/foreman/ui/test_ansible.py
ogajduse: For Python interpreter, both are the same. Therefore this change should not have any negative effects. 

```python
[ins] In [1]: '[\"test\"]' == '["test"]'
Out[1]: True
```

JacobCallahan (approved):
ACK. Merge pending Danny's recommendations

tests/foreman/ui/test_ansible.py
shweta83: I tried with interpreter and it worked but somehow when it fails to read as a string when read from the page. So I added this.

ogajduse:
> I think this is OK but is there some specific reason for us to prefer Ruff?

Astral's page sums it up pretty well. https://astral.sh/blog/the-ruff-formatter

I would like us to switch to Ruff mainly because of its speed compared to black. It is incredibly fast and the compatibility with Black is 99%. Looking at the patch, it might not look like 99%, but taking into account how big our project is and what changes were made, it is not that bad.

tests/foreman/api/test_errata.py
```diff
@@ -149,8 +149,7 @@ def _fetch_available_errata(host, expected_amount=None, timeout=120):
         errata = host.errata()
     else:
         pytest.fail(
-            'Host {} contains {} available errata, but expected to '
-            'contain {} of them'.format(
+            'Host {} contains {} available errata, but expected to ' 'contain {} of them'.format(
```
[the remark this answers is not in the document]
ogajduse: Addressed in 9d648a751733570d4fe4b5dd3925cd240342caac

tests/foreman/api/test_errata.py
```diff
@@ -122,8 +123,7 @@ def _validate_errata_counts(host, errata_type, expected_value, timeout=120):
         sleep(5)
     else:
         pytest.fail(
-            'Host {} contains {} {} errata, but expected to contain '
-            '{} of them'.format(
+            'Host {} contains {} {} errata, but expected to contain ' '{} of them'.format(
```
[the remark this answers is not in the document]
ogajduse: Addressed in 9d648a751733570d4fe4b5dd3925cd240342caac. Thanks for catching it!

tests/foreman/ui/test_ansible.py
ogajduse: I have reverted the patch in 9d648a751733570d4fe4b5dd3925cd240342caac and I have added a comment that will prevent formatted from touching this chunk of code.

tests/foreman/api/test_notifications.py
ogajduse: Same as in the other comment.

I have reverted the patch in https://github.com/SatelliteQE/robottelo/commit/9d648a751733570d4fe4b5dd3925cd240342caac and I have added a comment that will prevent formatted from touching this chunk of code.

tests/foreman/ui/test_ansible.py
shweta83: Thank you @ogajduse

jyejare:
@ogajduse Could we run some tests before merge it ? Especially the code where reviewers commented/discussed concerns ?

ogajduse:
> @ogajduse Could we run some tests before merge it ? Especially the code where reviewers commented/discussed concerns ?

Sure we can! I do not really see a need to do so, but @Gauravtalreja1 and/or @shweta83, feel free to trigger PRT here.

jyejare:
> Sure we can! I do not really see a need to do so

AFAIR, In history, we have broken the sanity when we merged a large formatting PR, So I see the need here.

ogajduse:
trigger: test-robottelo
pytest: tests/foreman/api/test_notifications.py tests/foreman/ui/test_ansible.py::TestAnsibleCfgMgmt::test_positive_assign_ansible_role_variable_on_host -k "long_running_task or test_positive_assign_ansible_role_variable_on_host"

PR #15048, opened 2024-05-13
Add Phoenix team based codeowners

vsedmik:
### Problem Statement
Right now we have generic teams like `robottelo-tier-X-reviewers` who are tagged to provide review based on the CODEOWNERS file. It would be nice to get a review request based on the team owning the particular file where the changes are proposed to.

### Solution
This PR adds files owned by Phoenix team into CODEOWNERS, so @SatelliteQE/phoenix folks should get tagged automatically. We can add other teams too if this approach proves to be helpful in future.


sambible (approved):
Good idea. Will allow me to make a seperate folder for just our team's work - the normal github notification pipeline is truly a fire hose

JacobCallahan (approved):

jyejare:
Good to see this is being done even before we start discussing this in large audience. We can take it as a POC.

PR #15163, opened 2024-05-23
Modify test_positive_ansible_custom_role to cover non-admin user scenario

Gauravtalreja1:
### Problem Statement
Currently, test_positive_ansible_custom_role covers only admin scenario, and as part of component audit its suggest to modify this test to cover non-admin user scenario

### Solution
Parametrizing the existing test to cover both admin and non-admin user scenarios with few more changes

PR #15164, opened 2024-05-23
Remove nailgun entity_mixin import from discovery tests

Gauravtalreja1:
### Problem Statement
Remove nailgun entity_mixin import from discovery tests, and replace entity_mixins.DEFAULT_SERVER_CONFIG 

### Solution
Replacing entity_mixins.DEFAULT_SERVER_CONFIG with new default_nailgun_config method, which is same as existing user_nailgun_config method
And, we don't have any automated tests under this module, verifying default_nailgun_config() method in https://github.com/SatelliteQE/robottelo/pull/15163

### DIFF
--- a/robottelo/config/__init__.py
+++ b/robottelo/config/__init__.py
@@ -90,6 +90,15 @@ def get_url():
     return urlunsplit((scheme, hostname, '', '', ''))
 
 
+def admin_nailgun_config():
+    """Return a NailGun configuration file constructed from default admin user credentials.
+
+    :return: ``nailgun.config.ServerConfig`` object, populated from admin user credentials.
+
+    """
+    return ServerConfig(get_url(), get_credentials(), verify=settings.server.verify_ca)
+
+
 def user_nailgun_config(username=None, password=None):
     """Return a NailGun configuration file constructed from default values.
 
--- a/tests/foreman/destructive/test_discoveredhost.py
+++ b/tests/foreman/destructive/test_discoveredhost.py
@@ -13,10 +13,10 @@
 from copy import copy
 import re
 
-from nailgun import entity_mixins
 import pytest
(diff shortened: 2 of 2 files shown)


omkarkhatavkar (approved):

robottelo/config/__init__.py
```diff
@@ -90,6 +90,19 @@ def get_url():
     return urlunsplit((scheme, hostname, '', '', ''))
 
 
+def default_nailgun_config():
```
synkd, line 93: Can you please explain the rationale for adding a new function for this instead of passing the admin credentials to `user_nailgun_config`?

Gauravtalreja1: yes, its indeed possible, but I just wanted to keep it separate from existing user_nailgun_config, as similar to what we had in nailgun. entity_mixin and its using `get_credentials` func to get creds so we don't need to pass and import settings just for ssh username and passwd in the tests

synkd: I'm not totally convinced that adding a function with very similar functionality to an existing function is preferable to importing settings in a test module, especially since https://github.com/SatelliteQE/robottelo/pull/9936 suggests that `user_nailgun_config` can be used with both admin and non-admin users. I won't block the PR based on that, though.

However, I do think the name of the function and the docstring should make it clear why the new function exists. I would suggest changing the name of the function to `admin_server_config`, since it is designed specifically to return a `ServerConfig` object with the admin credentials. In the docstring, I would remove the reference to `:param user:` since the function has no parameters, and I would also make it clear in the docstring that the config file will be constructed from the admin credentials.

What do you think about these suggestions?

Gauravtalreja1: @synkd Good point, renamed the func and updated the docstrings

synkd (approved):

PR #15242, opened 2024-05-30
Bump el8 kickstart version to 8.10

Gauravtalreja1:
### Problem Statement
New el8.10 is GA'ed, and kickstart constants requires an update.

### Solution
we need to manually bump this kickstart constants for provisioning tests

PRT is expected to fail due to known broker issue and a known BZ, so manually tested the el8.10 provisioning.
```
➜  ~ ssh root@<host-ip> "cat /etc/redhat-release"
Red Hat Enterprise Linux release 8.10 (Ootpa)
```


Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/cli/test_leapp_client.py -k test_leapp_upgrade_rhel

tests/foreman/cli/test_leapp_client.py
```diff
@@ -24,7 +24,7 @@
 synced_repos = pytest.StashKey[dict]
 
 RHEL7_VER = '7.9'
-RHEL8_VER = '8.9'
```
shweta83, line 27: We can skip it for now. We need new leapp version for it.

Gauravtalreja1: Removed this, you can update it later when you've new leapp version.
Now, as its gone, we don't care about PRT results, we can merge this as is

omkarkhatavkar (approved):

PR #15298, opened 2024-06-04
updated server.hostname validator for resolve rhel 8.10 formatting issue

omkarkhatavkar:
### Problem Statement
Currently the settings.server.hostname validator updates the settings.server.version.rhel_version from string value of `8.10` to `8.1` 

### Solution
Removed the validator settigns.server.hostname

### DIFF
--- a/robottelo/config/validators.py
+++ b/robottelo/config/validators.py
@@ -10,7 +10,7 @@ VALIDATORS = dict(
         ),
     ],
     server=[
-        Validator('server.hostname', default=''),
+        Validator('server.hostname', must_exist=True),
         Validator('server.hostnames', must_exist=True, is_type_of=list),
         Validator('server.version.release', must_exist=True),
         Validator('server.version.source', must_exist=True),


robottelo/config/validators.py
```diff
@@ -10,7 +10,7 @@
         ),
     ],
     server=[
-        Validator('server.hostname', default=''),
+        Validator('server.hostname', must_exist=True),
```
Gauravtalreja1, line 13: we don't have any field for `settings.server.hostname` in server.yml, we set it when assigning a xdist worker from settings.server.hostnames, so we can remove `must_exist=True` from this validator, right?
https://github.com/SatelliteQE/robottelo/blob/master/pytest_fixtures/core/xdist.py#L45

JacobCallahan (changes requested):
This field shouldn't be set by a user, it is set by the xdist alignment fixture as @Gauravtalreja1 pointed out

omkarkhatavkar:
I dont think this is required as the solution was on the exporting env variables.

PR #15340, opened 2024-06-06
removed server.hostname validator for resolve rhel 8.10 formatting issue

omkarkhatavkar:


devendra104 (approved):

jyejare (approved):
To curb the need of sign-off of 6.15.1 GA , merging this on priority as this has been verified locally that the rhel version is being translated as expected. 

The FIPS pipeline has run with this change by exporting env var with very similar approach and it worked.

Considering all this in mind, merging the PR now.

PR #15359, opened 2024-06-06
Use jira issue to get bug status

jameerpathan111:
### Problem Statement
We have migrated to Jira and some of the bz have been closed without being fixed/resolved. It's causing is_open function to return false status and we're now running the assertion which fails.

### Solution
- Use Jira issues to get the bug status

### DIFF
--- a/tests/foreman/api/test_ansible.py
+++ b/tests/foreman/api/test_ansible.py
@@ -299,7 +299,7 @@ class TestAnsibleCfgMgmt:
 
         # gather ansible facts by running ansible roles on the host
         host.play_ansible_roles()
-        if is_open('BZ:2216471'):
+        if is_open('SAT-18656'):
             wait_for(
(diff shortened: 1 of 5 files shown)


JacobCallahan:
@jameerpathan111 why the choice not to cherry-pick this?

jameerpathan111:
> @jameerpathan111 why the choice not to cherry-pick this?

Marked for cherry-pick.

JacobCallahan (approved):

devendra104 (approved):

shweta83:
Should we need to update docstring too for Bugzilla to Jira?

devendra104:
> Should we need to update docstring too for Bugzilla to Jira?

I feel this should be updated but we can cover it separately @jameerpathan111 what do you say?

jameerpathan111:
> Should we need to update docstring too for Bugzilla to Jira?

I'd suggest doing that in a separate PR. For now, let's merge it as it is.

PR #15378, opened 2024-06-07
[Customer Case] Applied Errata report with invalid errata

ColeHiggins2:
Customer case automation for generating an applied errata report after applying an invalid errata. 

All we need to test here is to make sure the report is generated without failure. However, because we never actually apply any errata, the report is an empty list. We can just assume it generated if there are no failures, but if anyone has any ideas on a potential assert that can be made im open to suggestions.

### DIFF
--- a/tests/foreman/api/test_reporttemplates.py
+++ b/tests/foreman/api/test_reporttemplates.py
@@ -458,6 +458,82 @@ def test_positive_applied_errata(
 @pytest.mark.tier2
 @pytest.mark.rhel_ver_match('[^6]')
 @pytest.mark.no_containers
+def test_positive_applied_errata_report_with_invalid_errata(
+    function_org, function_location, function_lce, rhel_contenthost, target_sat
+):
+    """Generate an Applied Errata report after an invalid errata has been applied
+
+    :id: cf64f193-870d-4053-ae4b-28148424b2e2
+
+    :setup: A Host with some invalid applied errata.
+
+    :steps:
+
+        1. Apply invalid errata
+        2. Generate an Applied Errata report
+
+    :expectedresults: A report is generated without failures
+
+    :CaseImportance: High
+
+    :BZ: 2176368
+    """
+    activation_key = target_sat.api.ActivationKey(
+        environment=function_lce, organization=function_org
+    ).create()
+    cv = target_sat.api.ContentView(organization=function_org).create()
+    target_sat.cli_factory.setup_org_for_a_custom_repo(
+        {
+            'url': settings.repos.yum_6.url,
+            'organization-id': function_org.id,
+            'content-view-id': cv.id,
+            'lifecycle-environment-id': function_lce.id,
+            'activationkey-id': activation_key.id,
+        }
+    )
+    result = rhel_contenthost.register(
+        function_org, function_location, activation_key.name, target_sat
+    )
+    assert f'The registered system name is: {rhel_contenthost.hostname}' in result.stdout
+    assert rhel_contenthost.subscribed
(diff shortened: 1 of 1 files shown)


damoore044 (approved):
:shipit:

tests/foreman/api/test_reporttemplates.py
```diff
@@ -455,6 +455,82 @@ def test_positive_applied_errata(
...
+
+        1. Apply invalid errata
+        2. Generate an Applied Errata report
+
+    :expectedresults: A report is generated without failures
+
+    :CaseImportance: High
+
+    :BZ: 2176368
+    """
```
jameerpathan111, line 480:
```suggestion

    :customerscenario: true
    
    """
```

jameerpathan111 (changes requested):

tests/foreman/api/test_reporttemplates.py
```diff
@@ -455,6 +455,82 @@ def test_positive_applied_errata(
...
+            'organization_id': function_org.id,
+            'report_format': 'json',
+            'input_values': {
+                'Filter Errata Type': 'all',
+                'Include Last Reboot': 'no',
+                'Status': 'all',
+            },
+        }
+    )
+
```
vijaysawant, line 532: Do you think any other assertion can be implementated after generating errata report?

vijaysawant (commented):
May be we can assert these errors in output
```
undefined method `[]' for nil:NilClass"
undefined method `last' for nil:NilClass"
```

tests/foreman/api/test_reporttemplates.py
```diff
@@ -455,6 +455,82 @@ def test_positive_applied_errata(
...
+    :expectedresults: A report is generated without failures
+
+    :CaseImportance: High
+
+    :BZ: 2176368
+    """
+    activation_key = target_sat.api.ActivationKey(
+        environment=function_lce, organization=function_org
+    ).create()
+    cv = target_sat.api.ContentView(organization=function_org).create()
```
vsedmik, line 484:
```suggestion
```
We could probably use `function_ak_with_cv` fixture instead (and `function_promoted_cv` it uses inside).

vsedmik (changes requested):
Just two small proposals, otherwise looks good.

ColeHiggins2:
trigger: test-robottelo
pytest: tests/foreman/api/test_reporttemplates.py -k test_positive_applied_errata_report_with_invalid_errata

tests/foreman/api/test_reporttemplates.py
ColeHiggins2: Had some issues with lce using that fixture, Test was failing, Lets just keep it how it is

tests/foreman/api/test_reporttemplates.py
ColeHiggins2: We talked about this in chat, There are not any good ways to make an assertion here. All we need to test is that this report is generated. A failing tests is a good enough indicator here

tests/foreman/api/test_reporttemplates.py
damoore044: I have also found that fixture fails to setup as expected

Satellite-QE:
**PRT Result**
```
Build Number: 7460
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/api/test_reporttemplates.py -k test_positive_applied_errata_report_with_invalid_errata --external-logging
Test Result : ========= 3 passed, 38 deselected, 286 warnings in 2164.84s (0:36:04) ==========
```

vsedmik (approved):

PR #15289, opened 2024-06-04
Enforce the Dynaconf validaton

ogajduse:
### Problem Statement
Currently, the configuration is not consistently validated. This can lead to:
  * Uncaught errors in production: Invalid configurations can cause crashes, or unexpected behavior if not detected before deployment.
  * Missed configuration changes: If validators are not updated alongside code changes, invalid configurations might slip through unnoticed.

### Solution
Enforce the config validation.

Additionally, remove unused configuration files and their validators.

Benefits:
* prevent errors - catch configuration issues early to prevent crashes, and unexpected behavior,
* consistent configs - define valid formats and value ranges to reduce surprises.
* improve maintainability -  validators act as documentation for configuration settings, making it easier for new developers to understand the project.
* early detection -  identify configuration issues during development or deployment to save time and resources compared to fixing them in production.

### Notes
I am open to changing the setting name and placement, `robottelo.unit_test` might not be the best choice.

PR #15413, opened 2024-06-14
Fix validation: validate IPv6 proxy URL only when IS_IPV6 is set to True

ogajduse:
### Problem Statement
Using latest robottelo, including the IPv6-related changes, I am running into
```
dynaconf.validator.ValidationError: server.http_proxy_ipv6_url must is_type_of <class 'str'> but it is None in env main
```

This error is caused by a validator that enforces string type, but it is setting the value to None itself. :meow_shrugs:
```
Validator('server.http_proxy_ipv6_url', is_type_of=str, default=None),
```

### Solution
Utilize conditional validation - validate the IPv6 proxy URL setting only when IS_IPV6 is set to True
https://www.dynaconf.com/validation/#conditional-validation

### Related Issues
This should be fixed before #15289 goes in

### DIFF
--- a/robottelo/config/validators.py
+++ b/robottelo/config/validators.py
@@ -36,7 +36,12 @@ VALIDATORS = dict(
         Validator('server.ssh_password', default=None),
         Validator('server.verify_ca', default=False),
         Validator('server.is_ipv6', is_type_of=bool, default=False),
-        Validator('server.http_proxy_ipv6_url', is_type_of=str, default=None),
+        # validate http_proxy_ipv6_url only if is_ipv6 is True
+        Validator(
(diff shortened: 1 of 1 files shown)


jyejare (approved):

rmynar (approved):
Maybe `default=''` might be helpful, but not necessary. ACK

jyejare:
@ogajduse I just backported Ipv6 to 6.15 and it likely won't go beyond that. If needed we can do cherry-pick later.

lpramuk (approved):

lpramuk:
Cherrypicking this PR should follow cherrypicking of previous validators fixes - #15277 
Setting labels accordingly.

PR #15433, opened 2024-06-17
Add multi-cv enable helper and fixture

vsedmik:
### Problem Statement
There is a "Multi-CV" feature partially implemented into 6.16 as a tech preview and we need to deliver related coverage for the implemented bits. For this reason we need to enable the feature in a non-standard way until it's released.

### Solution
This PR just adds a helper to enable related setting (to be used with any Satellite) and a session-scoped fixture which would provide Satellite with that setting enabled and set.

### DIFF
--- a/pytest_fixtures/component/contentview.py
+++ b/pytest_fixtures/component/contentview.py
@@ -68,3 +68,11 @@ def module_cv_repo(module_org, module_repository, module_lce, module_target_sat)
     content_view = content_view.read()
     content_view.version[0].promote(data={'environment_ids': module_lce.id, 'force': False})
     return content_view
+
+
+@pytest.fixture(scope='session')
+def session_multicv_sat(session_satellite_host):
+    """Satellite with multi-CV enabled"""
+    session_satellite_host.enable_multicv_setting()
+    session_satellite_host.update_setting('allow_multiple_content_views', 'True')
+    return session_satellite_host
--- a/robottelo/host_helpers/satellite_mixins.py
+++ b/robottelo/host_helpers/satellite_mixins.py
@@ -39,6 +39,18 @@ class EnablePluginsSatellite:
         assert 'Success!' in result.stdout
         return self
 
+    def enable_multicv_setting(self):
+        cfg_file = 'upstream_only_settings.rb'
+        cfg_path = self.execute(f'find /usr/share/gems/gems/ -name {cfg_file}').stdout.strip()
+        assert cfg_file in cfg_path, 'Config file not found'
+        self.execute(
+            f'sed -i "s/allow_multiple_content_views/#allow_multiple_content_views/g" {cfg_path}'
+        )
+        self.cli.Service.restart()
+        assert len(
+            self.api.Setting().search(query={'search': f'name={"allow_multiple_content_views"}'})
+        ), 'Multi-CV enablement failed'
+
 
 class ContentInfo:
     """Miscellaneous content helper methods"""
--- a/robottelo/host_helpers/satellite_mixins.py
+++ b/robottelo/host_helpers/satellite_mixins.py
@@ -40,6 +40,7 @@ class EnablePluginsSatellite:
         return self
 
     def enable_multicv_setting(self):
+        """Makes multi-CV setting available in the downstream Satellite"""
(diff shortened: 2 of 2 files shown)


robottelo/host_helpers/satellite_mixins.py
```diff
@@ -39,6 +39,18 @@ def enable_puppet_satellite(self):
         assert 'Success!' in result.stdout
         return self
 
+    def enable_multicv_setting(self):
+        cfg_file = 'upstream_only_settings.rb'
+        cfg_path = self.execute(f'find /usr/share/gems/gems/ -name {cfg_file}').stdout.strip()
+        assert cfg_file in cfg_path, 'Config file not found'
+        self.execute(
+            f'sed -i "s/allow_multiple_content_views/#allow_multiple_content_views/g" {cfg_path}'
```
lpramuk, line 47: function is about enable but `sed` actually disables the setting ...

pytest_fixtures/component/contentview.py
```diff
@@ -68,3 +68,15 @@ def module_cv_repo(module_org, module_repository, module_lce, module_target_sat)
...
+
+@pytest.fixture(scope='session')
+def session_multicv_sat(session_satellite_host):
+    """Satellite with multi-CV enabled"""
+    session_satellite_host.enable_multicv_setting()
+    multicv_setting = session_satellite_host.api.Setting().search(
+        query={'search': f'name={"allow_multiple_content_views"}'}
+    )[0]
+    multicv_setting.value = 'True'
+    multicv_setting.update({'value'})
```
Gauravtalreja1, line 81:
```suggestion
    multicv_setting = session_satellite_host.update_setting('allow_multiple_content_views', 'True')
```

robottelo/host_helpers/satellite_mixins.py
```diff
@@ -39,6 +39,18 @@ def enable_puppet_satellite(self):
...
+        )
+        self.cli.Service.restart()
+        assert len(
+            self.api.Setting().search(query={'search': f'name={"allow_multiple_content_views"}'})
+        ), 'Multi-CV enablement failed'
+
```
Gauravtalreja1, line 53: what are we checking here exactly, could you explain?

pytest_fixtures/component/contentview.py
```diff
@@ -68,3 +68,15 @@ def module_cv_repo(module_org, module_repository, module_lce, module_target_sat)
     content_view = content_view.read()
     content_view.version[0].promote(data={'environment_ids': module_lce.id, 'force': False})
     return content_view
+
+
+@pytest.fixture(scope='session')
+def session_multicv_sat(session_satellite_host):
+    """Satellite with multi-CV enabled"""
```
Gauravtalreja1, line 75: Are you going to add tests related to this fixture later or missed pushing?

robottelo/host_helpers/satellite_mixins.py
lpramuk: ah! 
UpstreamOnlySettings - so you need to excluded that so it gets applied to Stream/Downstream

robottelo/host_helpers/satellite_mixins.py
vsedmik: We are checking that after rb file manipulation and service restart the setting became available.

pytest_fixtures/component/contentview.py
vsedmik: Yes, @jeremylenz is about to add a few based on this PR.

robottelo/host_helpers/satellite_mixins.py
vsedmik: Yeah it looks odd, but here we actually remove the setting from `UpstreamOnlySettings`, making it available downstream.

vsedmik:
Tested locally with stream snap, looks to work well

image

robottelo/host_helpers/satellite_mixins.py
Gauravtalreja1: I mean with this assertion, if setting exists or not iiuc? if yes, I'd suggest adding a comment which could help understand this fixture

Gauravtalreja1 (approved):

PR #15172, opened 2024-05-23
Remove sleeps and make cron expression more accurate

sambible:
### Problem Statement
The sleeps in these tests result in inconsistent behavior and the cron expression is malformed, leading to 2 seperate faliures. This PR will fix both.

### Solution
Removed the sleep by setting max_tries to 20 on the task poll, reformatted the cron expression to always accurately set the sync time to 5 minutes after the plan is created. 5 minutes is required here due to slow interactions with the UI sometimes resulting in the product not being added until after the sync time had passed.

trigger: test-robottelo
pytest: tests/foreman/ui/test_contenthost.py -k 'test_positive_synchronize_custom_product_custom_cron_real_time'

PR #15452, opened 2024-06-19
Increase search rate for syncplan polling

dosas:
### Problem Statement

The timediff was increased to 5 minutes.
On 'fast' UIs this can lead to poll timeouts.

### Solution

Increasing the search rate to 15 fixes this:
15 * 20 / 60 = 5

### Related Issues
https://github.com/SatelliteQE/robottelo/pull/15172

### Relevant tests
tests/foreman/ui/test_syncplan.py::test_positive_synchronize_custom_product_custom_cron_real_time
tests/foreman/ui/test_syncplan.py::test_positive_synchronize_custom_product_custom_cron_past_sync_date

### DIFF
--- a/tests/foreman/ui/test_syncplan.py
+++ b/tests/foreman/ui/test_syncplan.py
@@ -230,7 +230,7 @@ def test_positive_synchronize_custom_product_custom_cron_real_time(session, modu
             f' and organization_id = {module_org.id}'
             f' and resource_id = {repo.id}'
             ' and resource_type = Katello::Repository',
-            search_rate=10,
+            search_rate=15,
             max_tries=20,
         )
         validate_repo_content(repo, ['erratum', 'rpm', 'package_group'])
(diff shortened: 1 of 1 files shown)


omkarkhatavkar:

ogajduse:
Can you please describe the issue you are seeing using the current implementation? Please also expand on what you mean by fast UI. I would like to understand better where you are coming from.

dosas:
> Can you please describe the issue you are seeing using the current implementation? Please also expand on what you mean by fast UI. I would like to understand better where you are coming from.

Okay.

As discussed in the related PR https://github.com/SatelliteQE/robottelo/pull/15172 the sync interval was changed form 3 to 5 minutes. Allegedly because the navigation was (sometimes?) to slow and the task was executed before the polling (slow UI).

After I merged these changes in the downstream tests I had the issue that the polling went into a timeout because it stopped before the task was executed (fast UI). 

So the fix is to increase the search rate so that the five minute mark is always reached.

jameerpathan111 (approved):

ogajduse (approved):
@dosas Thank you for explaining it. All clear now. LGTM

PR #15479, opened 2024-06-21
Replace skip_if with BlockedBy for migrated BZ

vsedmik:
### Problem Statement
BZ was CLOSED MIGRATED (to Jira) without the actual fix. We need to keep skipping based on Jira status instead.

### Solution
This PR.

### DIFF
--- a/tests/foreman/cli/test_satellitesync.py
+++ b/tests/foreman/cli/test_satellitesync.py
@@ -1669,7 +1669,6 @@ class TestContentViewSync:
         assert len(importing_cvv) == 1
 
     @pytest.mark.tier3
-    @pytest.mark.skip_if_open("BZ:2262379")
     def test_postive_export_import_ansible_collection_repo(
         self,
         target_sat,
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_satellitesync.py
```diff
@@ -1669,7 +1669,6 @@ def test_positive_export_rerun_failed_import(
         assert len(importing_cvv) == 1
 
     @pytest.mark.tier3
-    @pytest.mark.skip_if_open("BZ:2262379")
     def test_postive_export_import_ansible_collection_repo(
```
damoore044, line 1672:
```suggestion
    @pytest.mark.skip_if_open("SAT-23051")
    def test_postive_export_import_ansible_collection_repo(
```

damoore044: Since this issue is open, should we still use the pytest marker with the SAT issue instead?
Should work the same as the BZ being passed.

damoore044 (changes requested):

tests/foreman/cli/test_satellitesync.py
damoore044: Seems they function the same, non-blocking

damoore044 (approved):

ColeHiggins2 (approved):

Gauravtalreja1 (approved):

PR #15607, opened 2024-07-09
removing attach and auto-attach related items from cli hosts

pondrejk:
### Problem Statement
these tests no longer apply to 6.16 environment

### DIFF
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -32,10 +32,8 @@ from robottelo.constants import (
     PRDS,
     REPOS,
     REPOSET,
-    SM_OVERALL_STATUS,
 )
 from robottelo.exceptions import CLIFactoryError, CLIReturnCodeError
-from robottelo.hosts import ContentHostError
 from robottelo.logging import logger
 from robottelo.utils.datafactory import (
     invalid_values_list,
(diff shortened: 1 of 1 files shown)


tests/foreman/cli/test_host.py
```diff
@@ -2216,7 +1912,6 @@ def test_positive_unregister_host_subscription(
     res = rhel_contenthost.register(module_org, None, ak_with_subscription.name, target_sat)
```
synkd, line 1912: Do we still need this test? If so, shouldn't the activation key be one without subscriptions?

tests/foreman/cli/test_host.py
```diff
@@ -1900,304 +1894,6 @@ def test_positive_register(
...
-    client_status = rhel_contenthost.subscription_manager_status()
-    assert SM_OVERALL_STATUS['current'] in client_status.stdout
-    repo_list = rhel_contenthost.subscription_manager_list_repos()
-    assert "This system has no repositories available through subscriptions." in repo_list.stdout
-
-
-@pytest.mark.rhel_ver_match('7')
-@pytest.mark.cli_host_subscription
-@pytest.mark.tier3
-def test_negative_without_attach_with_lce(
```
synkd, line 2043: This still seems like a valid test case to me, although I think it should be renamed to a positive test for 6.16+. Can you please explain your decision to remove it?

synkd (changes requested):
Looks good overall, but I have questions about two of the affected tests.

jyejare:
@pondrejk Do you really need Tier2 review here ?

lhellebr:
I'd guess no because this is a test change, not a framework change?

tests/foreman/cli/test_host.py
pondrejk: makes sense, I brought it back with updated name

pondrejk:
@synkd Hi, sorry for delay, mind revisiting?

synkd (approved):

PR #15608, opened 2024-07-09
adding :BlockedBy to org/loc tests

pondrejk:
### Problem Statement
bz closed as migrated, tests no longer skipped


Gauravtalreja1 (approved):
ACK for the changes, and @pondrejk  I think we need this cherrypick to all supported zstream branches as well, else this test won't be skipped, or is this intended to not cherrypick?

jameerpathan111 (approved):

lhellebr (approved):
I also think this needs to be cherrypicked.
Otherwise, ACK.

PR #15609, opened 2024-07-09
removed outdated webhook action from parametrization

pondrejk:
### Problem Statement
  "actions.remote_execution.run_host_job_ansible_run_insights_plan_succeeded" no longer around in stream 

### Solution
confirmed with dev that this is an intentional change, related to https://github.com/theforeman/foreman_ansible/pull/712


Gauravtalreja1:
@pondrejk Do we've any test which can be run in PRT to confirm this removal in stream? and requesting review from @jameerpathan111 since he reviewed upstream PR for removal

ogajduse (approved):

jameerpathan111 (approved):

pondrejk:
> @pondrejk Do we've any test which can be run in PRT to confirm this removal in stream? and requesting review from @jameerpathan111 since he reviewed upstream PR for removal

hi, what do you mean by confirming the removal? There is a test parametrized by WEBHOOK_EVENTS, I could run it with --collect-only, but I'd say its safe to assume the removed line won't be used...

PR #15647, opened 2024-07-12
FAM: don't set a password for auth_source_ldap tests

evgeni:
### Problem Statement

the password field is protected, so the API doesn't return it and thus we always detect a change, failing tests

### Solution

don't set the password field


Griffin-Sullivan (approved):

ogajduse (approved):
ACK
@Griffin-Sullivan Can you please add the desired labels and merge the PR?

PR #15666, opened 2024-07-15
[RHCloud]Adding PIT markers for interop

ColeHiggins2:
Adding PIT markers for RHCloud


tests/foreman/api/test_rhc.py
```diff
@@ -32,6 +32,8 @@ def fixture_enable_rhc_repos(target_sat):
 
 
 @pytest.mark.e2e
+@pytest.mark.pit_server
+@pytest.mark.pit_client
```
vijaysawant, line 36: As per my understanding,
In order to use `pit_server` & `pit_client` marker test should satiesfy some conditions
```
server tests will contains-  non destructive, not fips
client tests will contains- `--no-containers` marker, not fips
```
- If we look this test, it uses `destructive` marker, we can't use `pit_server` marker here.
- Talking about `pit_client` scenario, so test doesn't have anything related to rhel content host, good to remove client marker.

vijaysawant (changes requested):
PIT marker not require for rhc > api test

tests/foreman/api/test_rhc.py
ogajduse: Correct, I agree with Vijay.

ogajduse: One small correction, `--no-containers` is a parameter that controls whether content hosts will be containers or VMs. If the test is already marked with `@pytest.mark.no_containers`, that is fine. In the PIT pipeline, we force all tests to run on VM by using `--no-containers` as a pytest argument.

tests/foreman/ui/test_rhcloud_insights.py
```diff
@@ -300,6 +302,8 @@ def test_host_details_page(
 
 
 @pytest.mark.e2e
+@pytest.mark.pit_server
```
ogajduse, line 305: Marking this test with `pit_server` might be an issue since our internal infra does not have an `install-capsule-on-top-of-rhel-compose` workflow defined. We need to define that workflow first and also make sure that our PIT pipeline script changes the capsule config as desired. 

Without this workflow, the capsule would be installed on top of the GA'd RHEL version instead of the RHEL version that we are supposed to test on.

Hit me up on Slack, I can guide you.

ogajduse (changes requested):
One blocking comment left. Currently, we can not mark tests that utilize Capsule(s) with the `pit_server` marker before we define a respective workflow in our AAP.

ogajduse:
If you wish the newly marked tests to be picked up also in 6.13 and 6.14 PIT testing, please add the respective labels.

vijaysawant (approved):

sambible (approved):

PR #15673, opened 2024-07-16
Replace skip_if with BlockedBy for migrated httpboot BZ

Gauravtalreja1:
### Problem Statement
BZ was CLOSED MIGRATED (to Jira) without the actual fix. We need to keep skipping based on Jira status instead.

### Solution
Replace skip_if with BlockedBy for migrated BZ


ColeHiggins2 (approved):

synkd (approved):

amolpati30 (approved):

shubhamsg199 (approved):
Ack

PR #15686, opened 2024-07-17
Add jira comments based on Jira labels

jameerpathan111:
### Problem Statement
- Reduce the frequency of commenting test results on Jira. 

### Solution
Take action based on labels
- Initially set a Pass/Fail label based on the test result
- If the state changes add a comment
- If the state is already failing, and test is failing, still add a comment
- If the state is already passing, and the test passes, don’t add a comment

### Related Issues
- SAT-26505 and SAT-26504

### DIFF
--- a/pytest_plugins/jira_comments.py
+++ b/pytest_plugins/jira_comments.py
@@ -87,6 +87,7 @@ def pytest_sessionfinish(session, exitstatus):
         user = os.environ.get('USER')
         build_url = os.environ.get('BUILD_URL')
         for issue in session.config.issue_to_tests_map:
+            all_tests_passed = True
             comment_body = (
                 f'This is an automated comment from job/user: {build_url if build_url else user} for a Robottelo test run.\n'
                 f'Satellite/Capsule: {settings.server.version.release} Snap: {settings.server.version.snap} \n'
@@ -94,8 +95,15 @@ def pytest_sessionfinish(session, exitstatus):
             )
             for item in session.config.issue_to_tests_map[issue]:
                 comment_body += f'{item["nodeid"]} : {item["outcome"]} \n'
+                if item["outcome"] == 'failed':
+                    all_tests_passed = False
             try:
-                add_comment_on_jira(issue, comment_body)
+                labels = (
+                    [{'add': 'tests_passed'}, {'remove': 'tests_failed'}]
+                    if all_tests_passed
+                    else [{'add': 'tests_failed'}, {'remove': 'tests_passed'}]
+                )
+                add_comment_on_jira(issue, comment_body, labels=labels)
             except Exception as e:
                 # Handle any errors in adding comments to Jira
                 logger.warning(f'Failed to add comment to Jira issue {issue}: {e}')
--- a/pytest_plugins/jira_comments.py
+++ b/pytest_plugins/jira_comments.py
@@ -4,9 +4,10 @@ import os
 import pytest
 
 from robottelo.config import settings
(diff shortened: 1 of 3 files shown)


robottelo/constants/__init__.py
```diff
@@ -1765,6 +1765,8 @@
 CLOSED_STATUSES = ("ON_QA", "VERIFIED", "RELEASE_PENDING", "CLOSED")
 WONTFIX_RESOLUTIONS = ("WONTFIX", "CANTFIX", "DEFERRED")
 # Jira statuses used by Robottelo issue handler.
+JIRA_TESTS_PASSED_LABEL = "tests-passed"
+JIRA_TESTS_FAILED_LABEL = "tests-failed"
```
jameerpathan111, line 1769: @JacobCallahan @jyejare do you have any suggestions for the labels? For ex. `automation-passed`/`automation-failed` or `automation-tests-passed` or `automation-tests-failed`.

jyejare: I think as discussed in the past `automation` is a reserved keyword for ansible and in general `test` could have multiple meanings including unit tests, so maybe DEVs wants to push the results tomorrow for unit tests they will have to use a separate label.

Hence I would suggest `qe-passed` / `qe-failed` which covers al QE tests types and refers to QEs tests. Open for suggestions through.

jameerpathan111: @jyejare I can understand the point for `automation` but the Dev one I'd argue about. Currently, we're the ones who need it and Dev's don't. And I'd not discard it based on future needs of Dev, which we're not even sure of. And yes, `test` could mean unit test but I don't think we need to differentiate labels based on that, at least not yet. 
As for `qe-passed` / `qe-failed`, this seems like it could have multiple meanings. I'd recommend something which has `test` in the name. For example, `qe-tests-passed` / `qe-tests-failed` or `robottelo-tests-passed` / `robottelo-tests-failed`

jyejare: @jameerpathan111 Its good to think ahead in some cases like this that would ask you to rethink / rework. Though we don't have yet from DEV, keeping that as a possibility would be a good path for us, needs no/less rethinking.

I respect your opinion to have `test` and hence I can vote for `qe-tests-*`.

pytest_plugins/jira_comments.py
```diff
@@ -87,15 +88,37 @@ def pytest_sessionfinish(session, exitstatus):
...
+                # If the state is already passing, and the test passes, don’t add a comment
+                if (data["status"] in settings.jira.issue_status) and (
+                    not all_tests_passed or JIRA_TESTS_PASSED_LABEL not in data["labels"]
+                ):
+                    add_comment_on_jira(issue, comment_body, labels=labels)
+                else:
+                    logger.warning(
+                        f"Jira comments are currently disabled for {issue} issue. It could be because jira is in {data['status']} "
+                        f"state or that there are no failing tests. "
+                        f"Please update issue_status in jira.conf to override this behaviour."
```
JacobCallahan, line 120:
```suggestion
                        "state or that there are no failing tests. "
                        "Please update issue_status in jira.conf to override this behaviour."
```

JacobCallahan (changes requested):
Just a minor change needed

JacobCallahan (approved):

jyejare (approved):
ACKed the commints in another commit with some non-blocking questions. The commits would be rebased in that PR later.

PR #15687, opened 2024-07-17
Color code and sort Jira comment based on the test result.

jameerpathan111:
### Problem Statement
- Currently, the Jira test result comment is not escaped for special characters which makes some parts of the comment to be colored red. Also, the comment is not sorted which makes it less readable.

### Solution
- Make Jira comment escape special characters.
- Sort test results based on the outcome.
- Color code test result outcome.

### Related Issues
- SAT-26353 and SAT-26352
- See SAT-25230 and SAT-24796 for results/improvements.
- Depends on https://github.com/SatelliteQE/robottelo/pull/15686 (Do not merge before it)

### DIFF
--- a/pytest_plugins/jira_comments.py
+++ b/pytest_plugins/jira_comments.py
@@ -82,12 +82,21 @@ def pytest_runtest_makereport(item, call):
             update_issue_to_tests_map(item, blocked_by_issues, test_result)
 
 
+def escape_special_characters(text):
+    special_chars = '[]'
+    for char in special_chars:
+        text = text.replace(char, f'\\{char}')
+    return text
+
+
 def pytest_sessionfinish(session, exitstatus):
     """Add test result comment to related Jira issues."""
     if hasattr(session.config, 'issue_to_tests_map'):
         user = os.environ.get('USER')
         build_url = os.environ.get('BUILD_URL')
         for issue in session.config.issue_to_tests_map:
+            # Sort test result based on the outcome.
+            session.config.issue_to_tests_map[issue].sort(key=lambda x: x['outcome'])
             all_tests_passed = True
             comment_body = (
                 f'This is an automated comment from job/user: {build_url if build_url else user} for a Robottelo test run.\n'
@@ -95,9 +104,15 @@ def pytest_sessionfinish(session, exitstatus):
                 f'Result for tests linked with issue: {issue} \n'
             )
             for item in session.config.issue_to_tests_map[issue]:
-                comment_body += f'{item["nodeid"]} : {item["outcome"]} \n'
+                color_code = '{color:green}'
                 if item['outcome'] == 'failed':
                     all_tests_passed = False
+                    color_code = '{color:red}'
+                # Color code test outcome
+                color_coded_result = f'{color_code}{item["outcome"]}{{color}}'
+                # Escape special characters in the node_id.
+                escaped_node_id = escape_special_characters(item['nodeid'])
+                comment_body += f'{escaped_node_id} : {color_coded_result} \n'
             try:
                 labels = (
                     [{'add': JIRA_TESTS_PASSED_LABEL}, {'remove': JIRA_TESTS_FAILED_LABEL}]


pytest_plugins/jira_comments.py
```diff
@@ -81,21 +82,59 @@ def pytest_runtest_makereport(item, call):
             update_issue_to_tests_map(item, blocked_by_issues, test_result)
 
 
+# Function to escape special characters
+def escape_special_characters(text):
+    special_chars = ['[', ']']
```
JacobCallahan, line 87: Minor, but if these are always going to be individual characters, then you can just use a simple string to iterate over.
```suggestion
    special_chars = '[]'
```

pytest_plugins/jira_comments.py
```diff
@@ -81,21 +82,59 @@ def pytest_runtest_makereport(item, call):
...
                 f'Result for tests linked with issue: {issue} \n'
             )
             for item in session.config.issue_to_tests_map[issue]:
-                comment_body += f'{item["nodeid"]} : {item["outcome"]} \n'
+                color_code = '{color:green}'
+                if item["outcome"] == 'failed':
+                    all_tests_passed = False
+                    color_code = '{color:red}'
+                # Color code test outcome
+                color_coded_result = color_code + f'{item["outcome"]}' + '{color}'
```
JacobCallahan, line 113: Why not do this line in a single f-stting instead of combining an f-stting with multiple concats?

JacobCallahan (changes requested):
Like the idea and the results, just some minor comments

pytest_plugins/jira_comments.py
jameerpathan111: Updated the PR

JacobCallahan (approved):
ACK. Another great addition

pytest_plugins/jira_comments.py
```diff
@@ -81,21 +82,58 @@ def pytest_runtest_makereport(item, call):
...
+                if (data["status"] in settings.jira.issue_status) and (
+                    not all_tests_passed or JIRA_TESTS_PASSED_LABEL not in data["labels"]
+                ):
+                    add_comment_on_jira(issue, comment_body, labels=labels)
+                else:
+                    logger.warning(
+                        f"Jira comments are currently disabled for {issue} issue. It could be because jira is in {data['status']} "
+                        f"state or that there are no failing tests. "
+                        f"Please update issue_status in jira.conf to override this behaviour."
+                    )
```
jyejare, line 136: You could have fit all this logic in the `add_comment_on_jira` function itself.

Non-blocking improvement comment.

pytest_plugins/jira_comments.py
```diff
@@ -81,21 +82,58 @@ def pytest_runtest_makereport(item, call):
...
+                    color_code = '{color:red}'
+                # Color code test outcome
+                color_coded_result = f'{color_code}{item["outcome"]}{{color}}'
+                # Escape special characters in the node_id.
+                escaped_node_id = escape_special_characters(item["nodeid"])
+                comment_body += f'{escaped_node_id} : {color_coded_result} \n'
             try:
-                add_comment_on_jira(issue, comment_body)
+                labels = (
+                    [{'add': JIRA_TESTS_PASSED_LABEL}, {'remove': JIRA_TESTS_FAILED_LABEL}]
```
jyejare, line 118: Is it requested from JIRA APIs to feed in this format where its a list of dictionaries, could this accept in the following format?:

```suggestion
                    {'add': JIRA_TESTS_PASSED_LABEL, 'remove': JIRA_TESTS_FAILED_LABEL}
```
if yes, please do, else just resolve the comment without any action.

Non-blocking comment.

jyejare (approved):
ACK with some non-blocking suggestions.

Also if we merge this as is, we don't need to merge https://github.com/SatelliteQE/robottelo/pull/15686/ as you have pull both the commits in this PR.

jameerpathan111:
> Also if we merge this as is, we don't need to merge #15686 as you have pull both the commits in this PR.

We can but I'd like to keep them separate for future reference, if that's alright.

pytest_plugins/jira_comments.py
jameerpathan111: It made sense to put it here as I needed data from the issue about the label here. It also avoids adding a lot of restrictions on the usage of `add_comment_on_jira` function as well.

pytest_plugins/jira_comments.py
jameerpathan111: yes, it's based on JIRA APIs:/

pytest_plugins/jira_comments.py
jameerpathan111: I prefer existing approach, but I appreciate your suggestion, thanks:)

PR #15719, opened 2024-07-22
Update tests for change in network configuration

shweta83:
### Problem Statement
As part of fix in network configuration SAT-23034, ifcfg created during provisioning of host has been removed and keyfile created by anaconda will be the only path for configuring network.

### Solution
Updated tests of provisioning to remove ifcfg and update it with keyfile created by anaconda.


ColeHiggins2 (approved):

shubhamsg199 (approved):
Ack

PR #15731, opened 2024-07-22
[Comp Eval] Removing Duplicate Test From Upgrades

ColeHiggins2:
Removing a duplicate test as part of the upgrades comp eval

### DIFF
--- a/tests/foreman/api/test_rhc.py
+++ b/tests/foreman/api/test_rhc.py
@@ -32,6 +32,8 @@ def fixture_enable_rhc_repos(target_sat):
 
 
 @pytest.mark.e2e
+@pytest.mark.pit_server
+@pytest.mark.pit_client
 @pytest.mark.tier3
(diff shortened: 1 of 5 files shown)


ColeHiggins2:
Open to adding labels to cherrypick into previous versions

damoore044 (approved):

LadislavVasina1 (approved):

sambible (approved):
This test is duplicated in CLI and API testing, and that's not necessary for coverage, so it's being removed. LGTM

jameerpathan111:
> This test is duplicated in CLI and API testing, and that's not necessary for coverage, so it's being removed. LGTM

If that's the case, wouldn't keeping the CLI test and removing the API one be better?

jameerpathan111 (approved):

Gauravtalreja1 (approved):

PR #15756, opened 2024-07-24
pit marker for rex

pondrejk:
...in addition to existing markers in rex cli module


ogajduse (changes requested):
Please see https://github.com/SatelliteQE/robottelo/pull/15666#discussion_r1679606584.

For the same reason, applying `pit_server` marker is not apropriate.

vijaysawant (approved):
Ack.. need to remove `pit server` marker

ogajduse (approved):

PR #15812, opened 2024-07-31
[Task 1 - Customer Scenario Checks] Replacing Migrated BZs docstrings token to Jira based 

jyejare:
### Problem Statement 1
The customer scenario check is based on the Bugzilla today, but since bug management is moved to Jira and the bugs have been closed as `CLOSED MIGRATED`, the customer scenario check is broken.

Also, the new bugs are being created on the Jira, hence it's vital to update the test docstrings to have the details from Jira instead.

### Solution 1
- Migrate all the bugs closed as `CLOSED MIGRATED` to get their latest updates from Jira as a new destination.

Also,
The BZs which are not migrated are kept for reference 
Some of the tests have already covered the verified token and hence only the BZ reference is removed.

### Next
- Step 2 / Next PR would switch the logic of customer scenario checks from BZ to Jira.

### Notes

Why are we linking to the parent Jira to the automated test even though there are clones an subtasks?
- The today's focus is Just to remediate the broken  customer scenario check based on BZ
- The `fix_version` was missing on some of the clones at the time of writing this PR
- Parents has the customer cases attached, whereas some clones don’t hence finding the link to customer scenarios is hard

The issue that could happen to link the test to parent Jira ?
- The broken status-based test collection or any other logic

Should we be needing the revisit, if some logic/implementation needs stricter status of a jira ids linked to the tests?
- Yes, we may. But for now, fixing the customer scenario checks is more important.

### DIFF
--- a/tests/foreman/ui/test_organization.py
+++ b/tests/foreman/ui/test_organization.py
@@ -199,7 +199,7 @@ def test_positive_create_with_all_users(session, module_target_sat):
 
     :expectedresults: Organization and user entities assigned to each other
 
-    :BZ: 1321543
+    :verifies: SAT-25386
 
     :BlockedBy: SAT-25386
     """
--- a/tests/foreman/api/test_capsulecontent.py
+++ b/tests/foreman/api/test_capsulecontent.py
@@ -1709,7 +1709,7 @@ class TestCapsuleContentManagement:
 
         :BlockedBy: SAT-25503
 
-        :BZ: 2284027
+        :verifies: SAT-26453
 
         :customerscenario: true
         """
--- a/tests/foreman/api/test_computeresource_vmware.py
+++ b/tests/foreman/api/test_computeresource_vmware.py
@@ -61,6 +61,8 @@ def test_positive_provision_end_to_end(
     :customerscenario: true
 
     :BZ: 2186114
+
+    :verifies: SAT-18721
     """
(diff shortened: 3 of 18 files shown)


shweta83 (approved):
Ack

lhellebr (approved):

tests/foreman/ui/test_organization.py
```diff
@@ -199,7 +199,7 @@ def test_positive_create_with_all_users(session, module_target_sat):
 
     :expectedresults: Organization and user entities assigned to each other
 
-    :BZ: 1321543
+    :verifies: SAT-25386
```
jameerpathan111, line 202:
```suggestion
    :Verifies: SAT-25386
```
@jyejare can we keep the token consistent? let's use `Verifies`.

jameerpathan111 (approved):
@jyejare Ack, but let's be consistent and use uppercase `Verifies`.

shweta83:
> @jyejare Ack, but let's be consistent and use uppercase `Verifies`.

I am in favor to not stick with case, and we have added ignorecase for the same reason so we don't have to worry about it.

jyejare:
@jameerpathan111 @vsedmik @shweta83 Yes, Tokens are case-insensitive. I mean I would do it in this PR but cant promise this would be followed by everyone later. Since there is no impact of case here, let it be any case.

PR #15878, opened 2024-08-07
Jira replacing bz for customer scenarios check

jyejare:
### Problem Statement
- Broken Customer Scenarios Tagging as we have switched from Bugzilla to Jira
- The newly created and migrated BZs to Jira are not being tracked for customer scenarios
- Broken Customer Scenarios report in polarion

### Solution
- Switching the Customer Scenarios Tagging identification from Bugzilla to Jira
- New jira specific Customer Scenarios script functions are added 
- The existing Jira module is modified to work for all the possible scenarios  like jira test selection / commenting  and customer scenarios scripting
- The weekly check GHA now uses jira based integration for customer scenario checks

### Results
```
# ./scripts/customer_scenarios.py --jira
[####################################]  100%          
The following tests need customerscenario tags:
tests/foreman/cli/test_remoteexecution.py ['test_positive_timeout_to_kill', 'SAT-25243']
tests/foreman/cli/test_satellitesync.py ['test_positive_export_library_custom_repo', 'SAT-24884']
```

### DIFF
--- a/robottelo/utils/issue_handlers/jira.py
+++ b/robottelo/utils/issue_handlers/jira.py
@@ -20,6 +20,23 @@ from robottelo.logging import logger
 # The .version group being a `d.d` string that can be casted to Version()
 VERSION_RE = re.compile(r'(?:sat-)*?(?P<version>\d\.\d)\.\w*')
 
+mapped_response_fields = {
+    'key': "{obj_name}['key']",
+    'summary': "{obj_name}['fields']['summary']",
+    'status': "{obj_name}['fields']['status']['name']",
+    'labels': "{obj_name}['fields']['labels']",
+    'resolution': "{obj_name}['fields']['resolution']['name'] if {obj_name}['fields']['resolution'] else ''",
+    'fixVersions': "[ver['name'] for ver in {obj_name}['fields']['fixVersions']] if {obj_name}['fields']['fixVersions'] else []",
+    # Custom Field - SFDC Cases Counter
+    'customfield_12313440': "{obj_name}['fields']['customfield_12313440']",
+}
+
+
+def sanitized_issue_data(issue, out_fields):
+    return {
+        field: eval(mapped_response_fields[field].format(obj_name='issue')) for field in out_fields
+    }
+
 
 def is_open_jira(issue_id, data=None):
     """Check if specific Jira is open consulting a cached `data` dict or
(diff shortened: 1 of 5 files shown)


scripts/customer_scenarios.py
```diff
@@ -39,6 +40,29 @@ def get_bz_data(paths):
     return result
 
 
+def get_tests_path_without_customer_tag(paths):
+    """Returns the path and test name that does not have customerscenario token even though
+    it has verifies token when necessary
```
jameerpathan111, line 45:
```suggestion
    """Returns the path and test name that does not have customerscenario token even 
    though it has verifies token when necessary
```

scripts/customer_scenarios.py
```diff
@@ -39,6 +40,29 @@ def get_bz_data(paths):
...
+        paths {list} -- List of test modules paths
+    """
+    testcases = testimony.get_testcases(paths)
+    result = {}
+    for path, tests in testcases.items():
+        path_result = []
+        for test in tests:
+            test_dict = test.to_dict()
+            test_data = {**test_dict['tokens'], **test_dict['invalid-tokens']}
+            if 'verifies' in test_data and (
```
jameerpathan111, line 57: Is this case-sensitive?

robottelo/utils/issue_handlers/jira.py
```diff
@@ -169,12 +195,34 @@ def collect_dupes(jira, collected_data, cached_data=None):  # pragma: no cover
...
+    response = requests.get(
+        f"{settings.jira.url}/rest/api/latest/search/",
+        params=params,
+        headers={"Authorization": f"Bearer {settings.jira.api_key}"},
+    )
+    response.raise_for_status()
+    return response
+
+
+def get_data_jira(issue_ids, cached_data=None, jira_fields=None):  # pragma: no cover
```
jameerpathan111, line 219:
```suggestion
def get_data_jira(issue_ids, cached_data=None, jira_fields=common_jira_fields):  # pragma: no cover
```
Wouldn't it be better to use `common_jira_fields`? It'll avoid getting a ton of unnecessary data.

jameerpathan111 (changes requested):

robottelo/utils/issue_handlers/jira.py
jyejare: Handled within the code instead of arguments to param as its not a good practice.

jyejare:
@jameerpathan111 Please revisit, the comments has been resolved.

jameerpathan111 (approved):

JacobCallahan (approved):

PR #15931, opened 2024-08-14
Remove install_katello_ca for good

vsedmik:
### Problem Statement
Katello-ca is going to be removed from 6.16 and we should use GR everywhere. The last place where is has been used is `module_container_contenthost`.

### Solution
Update `module_container_contenthost` to use GR and remove the `install_katello_ca` for good.

### Related Issues
https://issues.redhat.com/browse/SAT-25401

### PRT test Cases example
trigger: test-robottelo
pytest: tests/foreman/cli/test_container_management.py -k pull_image

### DIFF
--- a/pytest_fixtures/core/contenthosts.py
+++ b/pytest_fixtures/core/contenthosts.py
@@ -198,7 +198,7 @@ def katello_host_tools_tracer_host(rex_contenthost, target_sat):
 
 
 @pytest.fixture(scope='module')
-def module_container_contenthost(request, module_target_sat):
+def module_container_contenthost(request, module_target_sat, default_org):
     """Fixture that installs docker on the content host"""
     request.param = {
         "rhel_version": "8",
@@ -207,8 +207,6 @@ def module_container_contenthost(request, module_target_sat):
     }
     with Broker(**host_conf(request), host_class=ContentHost) as host:
         host.register_to_cdn()
-        # needed for docker commands to accept Satellite's cert
-        host.install_katello_ca(module_target_sat)
         for client in constants.CONTAINER_CLIENTS:
             assert (
                 host.execute(f'yum -y install {client}').status == 0
@@ -216,6 +214,16 @@ def module_container_contenthost(request, module_target_sat):
         assert (
             host.execute('systemctl enable --now podman').status == 0
         ), 'Start of podman service failed'
+        host.unregister()
+        # register the host to Satellite to accept its cert
+        ak = module_target_sat.cli_factory.make_activation_key(
+            {
+                'lifecycle-environment': 'Library',
+                'content-view': 'Default Organization View',
+                'organization-id': default_org.id,
+            }
+        )
+        assert host.register(default_org, None, ak.name, module_target_sat).status == 0
         yield host
 
 
--- a/pytest_fixtures/core/contenthosts.py
+++ b/pytest_fixtures/core/contenthosts.py
@@ -198,7 +198,7 @@ def katello_host_tools_tracer_host(rex_contenthost, target_sat):
 
 
 @pytest.fixture(scope='module')
-def module_container_contenthost(request, module_target_sat, default_org):
+def module_container_contenthost(request, module_target_sat, module_org, module_activation_key):
     """Fixture that installs docker on the content host"""
     request.param = {
         "rhel_version": "8",
@@ -215,15 +215,10 @@ def module_container_contenthost(request, module_target_sat, default_org):
             host.execute('systemctl enable --now podman').status == 0
         ), 'Start of podman service failed'
         host.unregister()
-        # register the host to Satellite to accept its cert
-        ak = module_target_sat.cli_factory.make_activation_key(
-            {
-                'lifecycle-environment': 'Library',
-                'content-view': 'Default Organization View',
-                'organization-id': default_org.id,
-            }
+        assert (
+            host.register(module_org, None, module_activation_key.name, module_target_sat).status
+            == 0
         )
-        assert host.register(default_org, None, ak.name, module_target_sat).status == 0
         yield host
 
 
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -335,7 +335,7 @@ class ContentHost(Host, ContentHostMixins):
     def setup(self):
         logger.debug('START: setting up host %s', self)
         if not self.blank:
-            self.remove_katello_ca()
+            self.reset_rhsm()
 
         logger.debug('END: setting up host %s', self)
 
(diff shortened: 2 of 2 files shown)


vsedmik:
trigger: test-robottelo
pytest: tests/foreman/cli/test_container_management.py -k pull_image

Satellite-QE:
**PRT Result**
```
Build Number: 8107
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/cli/test_container_management.py -k pull_image --external-logging
Test Result : ========== 1 passed, 3 deselected, 16 warnings in 1459.93s (0:24:19) ===========
```

robottelo/hosts.py
```diff
@@ -569,39 +568,6 @@ def install_katello_host_tools(self):
...
-            self.execute('rpm -Uvh katello-ca-consumer-latest.noarch.rpm')
-        # Not checking the status here, as rpm could be installed before
-        # and installation may fail
-        result = self.execute(f'rpm -q katello-ca-consumer-{satellite.hostname}')
-        # Checking the status here to verify katello-ca rpm is actually
-        # present in the system
-        if satellite.hostname not in result.stdout:
-            raise ContentHostError('Failed to download and install the katello-ca rpm')
-
     def remove_katello_ca(self):
```
Gauravtalreja1, line 571: if we're not going to install katello_ca, so can we remove helper remove_katello_ca as well?

pytest_fixtures/core/contenthosts.py
```diff
@@ -198,7 +198,7 @@ def katello_host_tools_tracer_host(rex_contenthost, target_sat):
 
 
 @pytest.fixture(scope='module')
-def module_container_contenthost(request, module_target_sat):
+def module_container_contenthost(request, module_target_sat, default_org):
```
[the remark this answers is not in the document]
vsedmik: It is used regardless the Org (it does not matter which Org it is registered to), so I suppose we can go with any of them.

robottelo/hosts.py
vsedmik: Hmm, maybe we could. I just need to figure out if `subman unregister` is enough or we should do other cleanup (since the cert is used anyway under the hood by the GR)

pytest_fixtures/core/contenthosts.py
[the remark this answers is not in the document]
Gauravtalreja1: if host is registered in default_org then it won't be accessible in other orgs, which might fail a most of tests
I think we'll need to create a func scoped and module scoped fixtures for same purpose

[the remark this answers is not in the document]
vsedmik: No. The host does not need to be registered in any Org to search and pull images, it was not registered at all till now.
The only two things it needs are:
1. podman/docker client installed
2. accept Sat's cert to avoid need for `--tls-verify=false`

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/cli/test_container_management.py -k pull_image

Satellite-QE:
**PRT Result**
```
Build Number: 8121
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/cli/test_container_management.py -k pull_image --external-logging
Test Result : ========== 1 passed, 3 deselected, 16 warnings in 1544.21s (0:25:44) ===========
```

robottelo/hosts.py
vsedmik: I have replaced the `remove_katello_ca` by `reset_rhsm` since it is used at a couple of places.

robottelo/hosts.py
```diff
@@ -569,55 +568,11 @@ def install_katello_host_tools(self):
...
-        :return: None.
-        :raises robottelo.hosts.ContentHostError: If katello-ca wasn't
-            installed.
-        """
-        warnings.warn(
-            message='The install_katello_ca method is deprecated, use the register method instead.',
-            category=DeprecationWarning,
-            stacklevel=2,
-        )
-        self._satellite = satellite
```
ogajduse, line 586: Since you are removing the logic that ensures the connection between the `Satellite` host and the `ContentHost` object, I would like to see it being put in `register` and `api_register`

I am aware of the fact that ContentHost can be registered to Satellite **or** Capsule, but that is not something I would worry about too much at the moment.

vsedmik: Good catch, I'll take a look where to add it.

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/cli/test_container_management.py -k pull_image

Satellite-QE:
**PRT Result**
```
Build Number: 8127
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/cli/test_container_management.py -k pull_image --external-logging
Test Result : ========== 1 passed, 3 deselected, 16 warnings in 1052.07s (0:17:32) ===========
```

ogajduse (approved):
LGTM.
Waiting on @Gauravtalreja1 to rereview.

jameerpathan111 (approved):

synkd (approved):

robottelo/hosts.py
```diff
@@ -569,55 +568,11 @@ def install_katello_host_tools(self):
...
-        self.execute('subscription-manager unregister')
-        # Not checking the status here, as rpm can be not even installed
-        # and deleting may fail
-        self.execute('yum erase -y $(rpm -qa |grep katello-ca-consumer)')
-        # Checking the status here to verify katello-ca rpm is actually
-        # not present in the system
-        result = self.execute('rpm -qa |grep katello-ca-consumer')
-        if result.status == 0:
-            raise ContentHostError(f'katello-ca rpm(s) are still installed: {result.stdout}')
+        self.execute(r'\cp -f /etc/rhsm/rhsm.conf{.bak,}')
```
Gauravtalreja1, line 575: I think when a host is never registered to any system other than CDN, where `/etc/rhsm/rhsm.conf.bak` won't exist, so I think it could fail here while copying, so can we add this to try and run sub-man clean in finally to avoid the file not present error?

vsedmik: If the host was never registered to a SAT (and so the `rhsm.conf.bak` does not exist) the `cp` command would fail with status 1 and `self.execute` would just return the status and not throw any exception to be caught. Since we don't assert the status here (on purpose for this case) the code execution should not be interrupted.

```
In [5]: host.execute(r'\cp -f /etc/rhsm/rhsm.conf{.bak,}')
[DEBUG 240821 09:48:38] sat.redhat.com executing command: \cp -f /etc/rhsm/rhsm.conf{.bak,}
[DEBUG 240821 09:48:38] sat.redhat.com command result:
    stdout:
    
    stderr:
    cp: cannot stat '/etc/rhsm/rhsm.conf.bak': No such file or directory
    
    status: 1
Out[5]: 
stdout:

stderr:
cp: cannot stat '/etc/rhsm/rhsm.conf.bak': No such file or directory

status: 1
```

Gauravtalreja1: yes, it would just fail silently and it won't throw any exception, However, I feel it would be better to handle it in a try-finally block to avoid any errors

ogajduse: It is still unclear to me why you would like to handle it and what would be a part of the logical block that handles the non-zero return code. What is there to handle? I might be missing something.

vsedmik: I think it's not needed. There is actually nothing we should do in case there is no `rhsm.conf.bak` (blank host will just keep pointing to the CDN),  so it doesn't make sense to me ho handle that return code. And since the `self.execute` does not throw any exception in that case I can see no reason to put it in the `try` block. Both of it just seem redundant to me and thus undesirable.

Gauravtalreja1 (approved):

PR #15932, opened 2024-08-14
Satisfy validators, enable template config validation

ogajduse:
### Problem Statement
1. Dynaconf validation fails when the `.template` configuration files are used
2. `settings.local.yaml` is not loaded by Dynaconf due to its [settings loading policy](https://www.dynaconf.com/settings_files/?h=root_path#loading-setting-files)
3. If `2.` is fixed, config sections specified in `settings.local.yaml` override the preloaded settings (the YAML files in `conf/`)

### Solution
1. fix the `.template` configuration files to satisfy validation and stop ignoring validation errors in GH Actions
2. initialize the settings object with `root_path`
3. use [Dynaconf's merge functionality](https://www.dynaconf.com/merging/?h=mer#merging-existing-data-structures) to merge preloaded settings with `settings.local.yaml`

### Note
This change allows us to have all configuration files in the `conf/` directory symlinked from the configuration repository where we store the production configuration. After this PR is merged, users can utilize `settings.local.yaml` to override the config as they like without the need to touch settings in `conf/`.

### DIFF
--- a/conf/rh_cloud.yaml.template
+++ b/conf/rh_cloud.yaml.template
@@ -1,4 +1,5 @@
 RH_CLOUD:
+  TOKEN: this-is-the-token
   INSTALL_RHC: false
   ORGANIZATION: org_name
   ACTIVATION_KEY: ak_name
--- a/conf/rh_cloud.yaml.template
+++ b/conf/rh_cloud.yaml.template
@@ -1,5 +1,5 @@
 RH_CLOUD:
-  TOKEN: this-is-the-token
+  TOKEN: this-isnt-the-token
   INSTALL_RHC: false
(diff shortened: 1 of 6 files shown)


pondrejk (approved):

conf/rh_cloud.yaml.template
```diff
@@ -1,4 +1,5 @@
 RH_CLOUD:
+  TOKEN: this-is-the-token
```
JacobCallahan, line 2: non-blocking 😉 
```suggestion
  TOKEN: this-isnt-the-token
```

JacobCallahan (approved):

PR #15939, opened 2024-08-15
pit marker for hammer ping test

pondrejk:
### Problem Statement
just adding the pit marker for the hammer ping test

### DIFF
--- a/tests/foreman/cli/test_ping.py
+++ b/tests/foreman/cli/test_ping.py
@@ -17,6 +17,7 @@ import pytest
 pytestmark = [pytest.mark.tier1, pytest.mark.upgrade]
 
 
+@pytest.mark.pit_server
 @pytest.mark.parametrize('switch_user', [False, True], ids=['root', 'non-root'])
 def test_positive_ping(target_sat, switch_user):
     """hammer ping return code


tests/foreman/cli/test_ping.py
```diff
@@ -17,6 +17,7 @@
 pytestmark = [pytest.mark.tier1, pytest.mark.upgrade]
 
 
+@pytest.mark.pit_server
 @pytest.mark.parametrize('switch_user', [False, True], ids=['root', 'non-root'])
 def test_positive_ping(target_sat, switch_user):
```
Gauravtalreja1, line 22: We also have the API ping test, Both of these tests could be good candidates for pit testing, right?

jameerpathan111 (approved):

pnovotny (approved):

tests/foreman/cli/test_ping.py
Gauravtalreja1: @pondrejk ^

pondrejk: hi, well, I'd say one is enough for this scenario, hammer goes through the api anyway, so if there's any problem there, we should see that with this test too

PR #16115, opened 2024-08-28
Fix delete host test through new All host page

LadislavVasina1:
### Problem Statement
Fix the test that is deleting the host via the new All Host UI. It was failing due to a variance in the table header.

### Solution
This PR fixes that by setting the table header to just `Name` and then returning it back to the state at the beginning.


LadislavVasina1:
```
trigger: test-robottelopytest: tests/foreman/ui/test_host.py -k 'test_all_hosts_delete'
```

LadislavVasina1:
```
trigger: test-robottelo
pytest: tests/foreman/ui/test_host.py -k 'test_all_hosts_delete'
```

sambible (approved):

PR #16203, opened 2024-09-03
Start using uv

JacobCallahan:
UV[1] is turning out to be a great new replacement for pip and other tools in the packaging ecosystem.
This change starts with moving to uv for our GitHub actions, Dockerfile, and introduces UV's script dependency mangement.
This last part will allow people to simply `uv run scripts/some_script.py` without having to worry about non-robottelo dependencies.

[1] - https://docs.astral.sh/uv/

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -17,6 +17,9 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.10', '3.11', '3.12']
+    env:
+      UV_CACHE_DIR: /tmp/.uv-cache
+      UV_SYSTEM_PYTHON: 1
     steps:
       - name: Checkout Robottelo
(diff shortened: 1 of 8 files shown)


JacobCallahan:
Speed comparisons for the Install Dependencies step

Original:  2m 11s
New:       1m 8s
New w/cache: 32s (5s of which is uv)

ogajduse (approved):
TIL
* https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies
  * https://packaging.python.org/en/latest/specifications/inline-script-metadata/#inline-script-metadata
* https://docs.astral.sh/uv/guides/integration/docker/#installing-uv and
* https://docs.astral.sh/uv/guides/integration/github/

LGTM. Thank you, Jake!

lpramuk (approved):

PR #16214, opened 2024-09-04
Avoid calling decode on str

ekohl:
### Problem Statement

In tests we see this failure:

    AttributeError: 'str' object has no attribute 'decode'. Did you mean: 'encode'?

### Solution

This implies that the strerr value is a str, which you can't decode. I can't find the specific change, but this implies the result used to be a byte string and is now decoded elsewhere.

### DIFF
--- a/tests/foreman/destructive/test_rename.py
+++ b/tests/foreman/destructive/test_rename.py
@@ -193,8 +193,8 @@ def test_negative_rename_sat_wrong_passwd(module_target_sat):
     assert result.status == 1
     assert BAD_CREDS_MSG in result.stderr
     # assert no changes were made
-    result = module_target_sat.execute('hostname')
-    assert original_name == result.stdout.strip(), "Invalid hostame assigned"
+    hostname_result = module_target_sat.execute('hostname')
+    assert original_name == hostname_result.stdout.strip(), "Invalid hostame assigned"
 
 
 @pytest.mark.stubbed
--- a/tests/foreman/destructive/test_rename.py
+++ b/tests/foreman/destructive/test_rename.py
@@ -191,7 +191,10 @@ def test_negative_rename_sat_wrong_passwd(module_target_sat):
         f'satellite-change-hostname -y {new_hostname} -u {username} -p {password}'
     )
     assert result.status == 1
-    assert BAD_CREDS_MSG in result.stderr[1]
+    assert BAD_CREDS_MSG in result.stderr
+    # assert no changes were made
+    result = module_target_sat.execute('hostname')
+    assert original_name == result.stdout.strip(), "Invalid hostame assigned"
 
 
 @pytest.mark.stubbed
--- a/tests/foreman/destructive/test_rename.py
+++ b/tests/foreman/destructive/test_rename.py
@@ -191,7 +191,7 @@ def test_negative_rename_sat_wrong_passwd(module_target_sat):
         f'satellite-change-hostname -y {new_hostname} -u {username} -p {password}'
     )
     assert result.status == 1
-    assert BAD_CREDS_MSG in result.stderr[1].decode()
+    assert BAD_CREDS_MSG in result.stderr[1]
 
 
 @pytest.mark.stubbed


omkarkhatavkar:

pondrejk:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_rename.py::test_negative_rename_sat_wrong_passwd

jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_rename.py::test_negative_rename_sat_wrong_passwd

lpramuk (approved):

rmynar (approved):

pondrejk:
@ekohl prt run failed like this
```
15:23:21  >       assert BAD_CREDS_MSG in result.stderr[1]
15:23:21  E       AssertionError: assert 'Unable to authenticate user admin' in 'n'
```
I guess  result.stder is just string, not list. But trying that out locally would be the best way, reach out if you need help with setup or anything

pondrejk (changes requested):
see my comment about the prt result

ekohl:
> I guess result.stder is just string, not list. But trying that out locally would be the best way, reach out if you need help with setup or anything

From reading that message that would also be my guess.

Having said that, 67f2a6b67a7c1301c78d09b2281e7b4b6b2afb60 and 1a1fc4573db24414e758f72b41c312f017adbda8 were the last changes to this test. Has this test ever passed in the past 2 years?

Also looking at the test it says:
https://github.com/SatelliteQE/robottelo/blob/88e23d965782ba66204b2a9c77152ac175d83b14/tests/foreman/destructive/test_rename.py#L133-L134

I don't see any assertion that the hostname is unchanged. What value does this test even bring?

Reading the bugs linked:
* https://bugzilla.redhat.com/show_bug.cgi?id=1485884 generic "needs better error checking"
* https://bugzilla.redhat.com/show_bug.cgi?id=1897360 sometimes candlepin fails
* https://bugzilla.redhat.com/show_bug.cgi?id=1925616 SELinux denial during hostname checking

In other words, only the first BZ looks vaguely related.

Would this test be a good candidate to drop instead?

pondrejk:
@ekohl on a general note, much of these questions are for discussion within the team that maintains the component. Here as a reviewer I'm mostly concerned if your changes make it pass green. But anyway, my 2 cents:

> Having said that, https://github.com/SatelliteQE/robottelo/commit/67f2a6b67a7c1301c78d09b2281e7b4b6b2afb60 and https://github.com/SatelliteQE/robottelo/commit/1a1fc4573db24414e758f72b41c312f017adbda8 were the last changes to this test. Has this test ever passed in the past 2 years?

Possibly yes, resutl.stdout structure is not decided in this test so the last changes here are not indicative. Sometime in the past the structure changed and the assertion started to fail, the job of the test maintainer is to catch that in TFA and adjust.

> I don't see any assertion that the hostname is unchanged. What value does this test even bring?

As you have a PR opened, feel free to add any assertions you see fit :)

> In other words, only the first BZ looks vaguely related.

Could have been past blockers. But past BZs are not the main/only thing that determines the test value. Just ask yourself (or component maintainer) if you want to cover this scenario in the future and catch regressions there

ekohl:
@ehelms @evgeni any thoughts on just dropping the test?

pondrejk:
@ekohl to me it's an obvious keeper (you want to be sure that s-c-h accepts credentials), but even if I agreed with a drop I'd rather be sure that the scenario actually works in all versions and wasn't silently failing while this test was off. So I suggest patching this up, doing al the cherry-picks and then checking the results. I'd move the drop-related discussion to a better place, or maybe leave it up to component evaluation

ekohl:
That makes sense. I now included your suggestion. I don't think I can trigger robotello though.

jameerpathan111:
Retriggered PRT job, see robottelo-pr-testing#8536

tests/foreman/destructive/test_rename.py
```diff
@@ -191,7 +191,10 @@ def test_negative_rename_sat_wrong_passwd(module_target_sat):
         f'satellite-change-hostname -y {new_hostname} -u {username} -p {password}'
     )
     assert result.status == 1
-    assert BAD_CREDS_MSG in result.stderr[1].decode()
+    assert BAD_CREDS_MSG in result.stderr
+    # assert no changes were made
+    result = module_target_sat.execute('hostname')
+    assert original_name == result.stdout.strip(), "Invalid hostame assigned"
```
evgeni, line 197:
```suggestion
    hostname_result = module_target_sat.execute('hostname')
    assert original_name == hostname_result.stdout.strip(), "Invalid hostame assigned"
```

evgeni (approved):
ACK on the change itself.
ACK on keeping the test.
I'd just not reuse the `result` var, it's confusing when reading results (lol)

jameerpathan111:
PRT is passing

jameerpathan111 (approved):

PR #16218, opened 2024-09-05
Fix SATELLITE_VERSION and SAT_NON_GA_VERSIONS in constants

jameerpathan111:
### Problem Statement
`SATELLITE_VERSION` and `SAT_NON_GA_VERSIONS` constants are wrongly set as 6.16 is not GA yet. This is causing `test_positive_capsule_repositories_setup` and `test_positive_satellite_repositories_setup` to fail and probably some other tests too.
### Solution
Fix `SATELLITE_VERSION` and `SAT_NON_GA_VERSIONS` in constants

### Related Issues
- SAT-27753

### DIFF
--- a/robottelo/constants/__init__.py
+++ b/robottelo/constants/__init__.py
@@ -6,9 +6,9 @@ from box import Box
 from nailgun import entities
 
 # This should be updated after each version branch
-SATELLITE_VERSION = "6.17"
+SATELLITE_VERSION = "6.16"
 SATELLITE_OS_VERSION = "8"
-SAT_NON_GA_VERSIONS = ['6.17']
+SAT_NON_GA_VERSIONS = ['6.16', '6.17']
 
 # Default system ports
 HTTPS_PORT = '443'


jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/maintain/test_advanced.py -k repositories_setup

Satellite-QE:
**PRT Result**
```
Build Number: 8479
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/maintain/test_advanced.py -k repositories_setup --external-logging
Test Result : ========== 2 passed, 9 deselected, 65 warnings in 1558.32s (0:25:58) ===========
```

jyejare:
@jameerpathan111 Right, earlier the branching used to happen on GA but things have been changed this time ?

@omkarkhatavkar We need to update the GHA for this then.

robottelo/constants/__init__.py
```diff
@@ -6,9 +6,9 @@
 from nailgun import entities
 
 # This should be updated after each version branch
-SATELLITE_VERSION = "6.17"
+SATELLITE_VERSION = "6.16"
 SATELLITE_OS_VERSION = "8"
-SAT_NON_GA_VERSIONS = ['6.17']
+SAT_NON_GA_VERSIONS = ['6.16', '6.17']
```
jyejare, line 11: Need to asses the impact having two versions here as if they are being looped to operate on both versions.

jyejare (commented):
ACK pending question.

robottelo/constants/__init__.py
```diff
@@ -6,9 +6,9 @@
 from nailgun import entities
 
 # This should be updated after each version branch
-SATELLITE_VERSION = "6.17"
+SATELLITE_VERSION = "6.16"
```
lpramuk, line 9: This should stay `6.17` for `master` branch

robottelo/constants/__init__.py
lpramuk: @jameerpathan111 What about `['6.16']` to skip the need for an assessment  ?

Gauravtalreja1: @jyejare SAT_NON_GA_VERSIONS is used in only repositories_setup tests, which just validates the repositories usage for upgrades, and this tests are passing in PRT

Gauravtalreja1 (approved):

robottelo/constants/__init__.py
lpramuk: You can check `6.13.z`, `6.14.z`, `6.15.z`, `6.16.z` branches that we have everything unreleased as nobody goes back to N-2 branch to adjust this constant.
E.g you can see in `6.13.z` branch:
```
SAT_NON_GA_VERSIONS = ['6.14', '6.15']
```
But that is very much useless as automation in 6.12.z branch does care only about `6.12` in SAT_NON_GA_VERSIONS  - this would be better just boolean flag or yet better...


This shall be config value.

lpramuk: I've assessed it and the only operation is `not in SAT_NON_GA_VERSIONS` so it is going to be OK with multiple values.   (there is no operation like `SAT_NON_GA_VERSIONS[0]`)

lpramuk (approved):

PR #16290, opened 2024-09-09
Adjust rhel version under test

sambible:
Small adjustment to test on RHEL8 instead of RHEL9, like the rest of the testing suite.

### DIFF
--- a/tests/foreman/ui/test_host.py
+++ b/tests/foreman/ui/test_host.py
@@ -2563,7 +2563,7 @@ def test_positive_manage_packages(
 )
 @pytest.mark.no_containers
 @pytest.mark.tier2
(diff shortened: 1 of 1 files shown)


ColeHiggins2 (approved):
Looks good to me

ogajduse:
A small adjustment, but potentially a big change. Can you please provide a bit of background about the patch?
Why is this patch needed?

sambible:
There is currently an issue in CI with the RHEL 9 repo enablement - I'd like to try using RHEL8 like the other hosts tests utilizing this fixture and see if it fixes the issue. If there's any problems with RHEL9 they won't be in the actual feature this is testing, so I'm not worried about coverage for this feature being impacted.

PR #16340, opened 2024-09-12
Fix pxeless discovery test

shubhamsg199:
Problem Statement
pxeless discovery tests  are going to fail due to change in workflow var

Solution
This pr updates the var in the fixture

Related Issues


shubhamsg199:
Manual PRT passed on Build 8640

Gauravtalreja1 (approved):

PR #16347, opened 2024-09-13
Remove unused code related to is_open bz

jameerpathan111:
### Problem Statement
- We have unused code related to `is_open` and `skip_if_open` bugzilla. We have migrated from using Bugzilla to Jira we don't need it anymore. 

### Solution
- Remove unused code related to is_open bz

### Related Issues
- SAT-27966

### DIFF
--- a/tests/foreman/cli/test_host.py
+++ b/tests/foreman/cli/test_host.py
@@ -705,14 +705,14 @@ def test_positive_list_infrastructure_hosts(
     target_sat.cli.Host.update({'name': target_sat.hostname, 'new-organization-id': module_org.id})
     # list satellite hosts
     hosts = target_sat.cli.Host.list({'search': 'infrastructure_facet.foreman=true'})
-    assert len(hosts) == 1
+    assert len(hosts) == 2
     hostnames = [host['name'] for host in hosts]
     assert rhel7_contenthost.hostname not in hostnames
     assert target_sat.hostname in hostnames
     # list capsule hosts
     hosts = target_sat.cli.Host.list({'search': 'infrastructure_facet.smart_proxy_id=1'})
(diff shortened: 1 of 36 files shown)


tests/foreman/cli/test_host.py
```diff
@@ -706,14 +705,14 @@ def test_positive_list_infrastructure_hosts(
     target_sat.cli.Host.update({'name': target_sat.hostname, 'new-organization-id': module_org.id})
     # list satellite hosts
     hosts = target_sat.cli.Host.list({'search': 'infrastructure_facet.foreman=true'})
-    assert len(hosts) == 2 if is_open('BZ:1994685') else len(hosts) == 1
+    assert len(hosts) == 1
```
pondrejk, line 708: as 1994685 was closed wontfix, so maybe we should keep 2. On the other hand, the issue might not persist, so this warrants a local run

jameerpathan111: It doesn't matter as so far the code was checking the else part `len(hosts) == 1` as BZ: 1994685 has been closed(no matter the kind of resolution) for more than 5 months.

pondrejk: hm, it does matter if it passes or fails, digging through past runs 
image
so if you could flip it to 2 in both cases I'd be thankful

jameerpathan111: Updated the PR

vsedmik (approved):
Looks good to me (looked at the Phoenix files only).

pondrejk (approved):

synkd (approved):

Gauravtalreja1 (approved):

PR #16354, opened 2024-09-13
Remove sanity marker from GCE test until issue is fixed on IPv6 satellite

shweta83:
### Problem Statement
tests/foreman/api/test_computeresource_gce.py::test_positive_gce_host_provisioned test is failing on IPv6 Satellite with 500 error. This test is marked for sanity and thus causing sanity pipeline to fail. 

### Solution
Until we fix this issue on IPv6, we are  removing sanity marker from the test.

### DIFF
--- a/tests/foreman/api/test_computeresource_gce.py
+++ b/tests/foreman/api/test_computeresource_gce.py
@@ -22,6 +22,7 @@ import pytest
 
 from robottelo.config import settings
 from robottelo.constants import GCE_RHEL_CLOUD_PROJECTS, VALID_GCE_ZONES
+from robottelo.utils.issue_handlers import is_open
 
 
 @pytest.mark.skip_if_not_set('gce')
@@ -228,6 +229,10 @@ class TestGCEHostProvisioningTestCase:
     @pytest.mark.tier1
     @pytest.mark.pit_server
     @pytest.mark.build_sanity
+    @pytest.mark.skipif(
+        ((settings.server.is_ipv6) and is_open('SAT-27997')),
+        reason='Google CR APIs failing for IPv6',
+    )
     @pytest.mark.parametrize('sat_gce', ['sat', 'puppet_sat'], indirect=True)
     def test_positive_gce_host_provisioned(self, class_host, google_host):
         """Host can be provisioned on Google Cloud


devendra104 (approved):

tests/foreman/api/test_computeresource_gce.py
```diff
@@ -227,7 +227,7 @@ def google_host(self, googleclient):
     @pytest.mark.e2e
     @pytest.mark.tier1
     @pytest.mark.pit_server
-    @pytest.mark.build_sanity
+    # @pytest.mark.build_sanity
```
devendra104, line 230: @shweta83 Can't we skip only for the ipv6?

jyejare: @devendra104 @shweta83 There is the other way we can use is:
1. File a Jira for the issue 
2. Use the `is_open` function within test function to skip the test execution as below:
```
@pytest ......
def test_positive_gce_host_provisione(.......
"""Host ........
if is_open('SAT-<issue_id>') and settings.server.is_ipv6:
    pytest.skip('Skipping for Ipv6 due to SAT-<issue_id>')
```

jyejare (changes requested):
Update the code as requested.

tests/foreman/api/test_computeresource_gce.py
shweta83: Can't we use skipif marker?

shweta83:
trigger: test-robottelo
pytest: tests/foreman/api/test_computeresource_gce.py -k test_positive_gce_host_provisioned
network_type: ipv6

tests/foreman/api/test_computeresource_gce.py
jyejare: @shweta83 With only ipv6 setting, this would be skipped unless we explicitly remove it which we tend to forget in general.

Combine it with `Is_open` jira with `and` operator.

shweta83: @jyejare I intentionally skipped the above implementation as it would unnecessarily execute all the fixtures and setup just to skip the whole test. Also the test fails in setup where we are calling Google APIs so it would be better to skip this way.

Gauravtalreja1 (approved):

tests/foreman/api/test_computeresource_gce.py
jyejare: @shweta83 Nop, I am talking using `skipif` marker itself to check both the things.
```
@pytest.mark.skipif((settings.server.is_ipv6 and is_open('SAT-1234')), reason='Google CR APIs failing for IPv6')
```

shweta83: Updated.

shweta83:
trigger: test-robottelo
pytest: tests/foreman/api/test_computeresource_gce.py -k test_positive_gce_host_provisioned
network_type: ipv6

jyejare (approved):

PR #16410, opened 2024-09-18
Update download_repofile helper to create repo file with product name

jameerpathan111:
### Problem Statement
We are creating all product repofiles with `dogfood.repo` name and it's causing capsule n-1 to test to fail with `"Satellite_Engineering_Satellite_6_15_Composes_Satellite_Capsule_6_15_RHEL8" is already defined, cannot redefine (file: /etc/yum.repos.d/dogfood.repo)` as the workflow used in the pipeline create the same repofiles too, but with specific product name, for example `satellite.repo`. This causes installer to fail as repo is now defined more than once in the configuration.

### Solution
- Update `download_repofile` helper to download repo to repofile named after the product.

### Related Issues
- SAT-27944


ogajduse (approved):
LGTM, thanks!

PR #16423, opened 2024-09-19
Fetch pre_upgrade data based on node properties

vsedmik:
### Problem Statement
During TFA I've seen several `post_upgrade` failures for "`rhel_contenthost`-parametrized" upgrade scenario tests where the pre-data could not be fetched from the `scenario_entities` file:
```
failed on setup with "Failed: Invalid test parameter: rhel7. Test data not found."
```

The reason is that the pre-upgrade test names are now suffixed with `-ipv4` or `-ipv6` so the `scenario_entities` hold items like this:
```
{"tests/upgrades/test_repository.py::TestSimpleContentAccessOnly::test_pre_simple_content_access_only[rhel7-ipv4]": {...}}
```
while [this piece of code](https://github.com/SatelliteQE/robottelo/blob/master/tests/upgrades/conftest.py#L267) requires the exact param, which is just `rhel7`, `rhel8`, `rhel9`.

### Solution(?)
There are probably more than one solution and I'm not sure which one would be the best.

Initially I thought we could go regardless the `ipvX` suffix since I supposed the pipelines were separated (first commit).
That's probably not a good solution when we start decorating tests with `['ipv4', 'ipv6']` explicitly, so I took the network type from the settings.

Please let me know if there is a better place where to get that information from in scope of the `pre_upgrade_data` fixture.

### DIFF
--- a/tests/upgrades/conftest.py
+++ b/tests/upgrades/conftest.py
@@ -264,7 +264,9 @@ def pre_upgrade_data(request):
     if len(upgrade_data) == 1:
         param_value = next(iter(upgrade_data.values()))
     else:
-        param_value = upgrade_data.get(request.param)
+        param_value = upgrade_data.get(
+            next(key for key in upgrade_data if key.startswith(request.param))
+        )
         if param_value is None:
             pytest.fail(f"Invalid test parameter: {request.param}. Test data not found.")
     return Box(param_value)
--- a/tests/upgrades/conftest.py
+++ b/tests/upgrades/conftest.py
@@ -90,6 +90,7 @@ import os
 from box import Box
 import pytest
 
+from robottelo.config import settings
 from robottelo.logging import logger
 from robottelo.utils.decorators.func_locker import lock_function
 
@@ -265,7 +266,7 @@ def pre_upgrade_data(request):
     if len(upgrade_data) == 1:
         param_value = next(iter(upgrade_data.values()))
     else:
-        network_type = dict(request.keywords.node.user_properties).get('SatelliteNetworkType')
+        network_type = 'ipv6' if settings.server.is_ipv6 else 'ipv4'
         param_value = upgrade_data.get(f'{request.param}-{network_type}')
         if param_value is None:
             pytest.fail(f"Invalid test parameter: {request.param}. Test data not found.")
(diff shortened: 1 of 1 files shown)


tests/upgrades/conftest.py
```diff
@@ -260,11 +260,13 @@ def test_something_post_upgrade(pre_upgrade_data):
...
         end_index = test_node_id.find(']')
         extracted_value = test_node_id[start_index:end_index]
-        upgrade_data[extracted_value] = _read_test_data(test_node_id)
+        if request.param in extracted_value:
+            upgrade_data[extracted_value] = _read_test_data(test_node_id)
     if len(upgrade_data) == 1:
         param_value = next(iter(upgrade_data.values()))
     else:
-        param_value = upgrade_data.get(request.param)
+        network_type = dict(request.keywords.node.user_properties).get('SatelliteNetworkType')
```
jyejare, line 268: The network type is not always `SatelliteNetworkType`, it is first checked in tests param and if not then only in SatelliteNetworkType

jyejare (changes requested):

tests/upgrades/conftest.py
vsedmik: The logic is that `extracted_value` can be anything like `rhel7`, `rhel7-ipv4`, `rhel7-ipv6` and other rhel variants.
Also `request.param` can be anything like that - with or without ipvX. When it is decorated _with_ ipvX, only that single value is added to the `upgrade_data` list due to the `if` on line 263. On the other hand, when `request.param` is not decorated with ipvX (i.e. just plain `rhel7`) the `upgrade_data` gets filled with everything that contains `rhel7`, like `['rhel7-ipv4', 'rhel7-ipv6']`. Only in that case we read the `SatelliteNetworkType` to resolve which one should be picked from `upgrade_data`.

jyejare (approved):

jyejare:
@vsedmik We accept this as a half solution and we need to communicate to the team that, for now we are not supporting upgrade scenarios parametrization for network type.

pondrejk (approved):

ColeHiggins2 (approved):

PR #16497, opened 2024-09-25
[RHCloud] Update configure_rhai_client registration

ColeHiggins2:
Registering a host with insights client was failing. 

After further investigation, It appears that we were are not actually syncing rhel 6 - 9 repos but instead "mocking" them to be enabled...which are needed for insights setup.

This process must have started failing during our update to global registration across components?

My solution to this is to simply move the registration portion of ```confiure_rhai_client``` to be executed AFTER we setup the correct "mock" repositories needed and packages installed.

The other option in my mind was to manually register insights after we setup the repos using 
```insights-client --register```

But the first way allows us to register insights upon Satellite registration (which is the expected behavior)

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -1052,17 +1052,6 @@ class ContentHost(Host, ContentHostMixins):
         :param register: Whether to register client to insights
         :return: None
         """
-        if register:
-            if not activation_key:
-                activation_key = satellite.api.ActivationKey(
-                    content_view=org.default_content_view.id,
-                    environment=org.library.id,
-                    organization=org,
-                ).create()
-            self.register(
-                org, None, activation_key.name, satellite, setup_insights=register_insights
-            )
-
         # Red Hat Insights requires RHEL 6/7/8 repo and it is not
         # possible to sync the repo during the tests, Adding repo file.
         distro_repo_map = {
@@ -1084,6 +1073,17 @@ class ContentHost(Host, ContentHostMixins):
         # Ensure insights-client rpm is installed
         if self.execute('yum install -y insights-client').status != 0:
             raise ContentHostError('Unable to install insights-client rpm')
+        # attempt to register host
+        if register:
+            if not activation_key:
+                activation_key = satellite.api.ActivationKey(
+                    content_view=org.default_content_view.id,
+                    environment=org.library.id,
+                    organization=org,
+                ).create()
+            self.register(
+                org, None, activation_key.name, satellite, setup_insights=register_insights
+            )
 
     def unregister_insights(self):
         """Unregister insights client.
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -1038,7 +1038,7 @@ class ContentHost(Host, ContentHostMixins):
             host.host_parameters_attributes = host_parameters
             host.update(['host_parameters_attributes'])
 
-    def configure_rhai_client(
+    def configure_insights_client(
         self, satellite, activation_key, org, rhel_distro, register_insights=True, register=True
     ):
         """Configures a Red Hat Access Insights service on the system by
@@ -1052,7 +1052,7 @@ class ContentHost(Host, ContentHostMixins):
         :param register: Whether to register client to insights
         :return: None
         """
-        # Red Hat Insights requires RHEL 6/7/8 repo and it is not
+        # Red Hat Insights requires RHEL 6/7/8/9 repo and it is not
         # possible to sync the repo during the tests, Adding repo file.
         distro_repo_map = {
             'rhel6': settings.repos.rhel6_os,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by ColeHiggins2** [comment]
trigger: test-robottelo
pytest: tests/foreman/cli/test_rhcloud_inventory.py -k 'test_positive_sync_inventory_status'

**Comment by Satellite-QE** [comment]
**PRT Result**
```
Build Number: 8801
Build Status: UNSTABLE
PRT Comment: pytest tests/foreman/cli/test_rhcloud_inventory.py -k test_positive_sync_inventory_status --external-logging
Test Result : ===== 5 passed, 16 deselected, 618 warnings, 1 error in 2447.09s (0:40:47) =====
```

**Review by vsedmik — APPROVED** [review]
Do you know what is the error in registration specifically?
I just wonder if it was failed `insights-client` installation during GR, which would explain why it works now when GR is done _after_ client pre-installation.

Otherwise I agree to exercise the `insights-client --register` via GR rather than separately.

File: robottelo/hosts.py [hunk 1]
```diff
@@ -1084,6 +1073,17 @@ def configure_rhai_client(
         # Ensure insights-client rpm is installed
         if self.execute('yum install -y insights-client').status != 0:
```
Gauravtalreja1, line 1074: Can this be moved to below the registration to check if insights-client package is installed? 
and it can be installed with yum when `register=False`>

robottelo/hosts.py
```diff
@@ -1052,17 +1052,6 @@ def configure_rhai_client(
...
-                activation_key = satellite.api.ActivationKey(
-                    content_view=org.default_content_view.id,
-                    environment=org.library.id,
-                    organization=org,
-                ).create()
-            self.register(
-                org, None, activation_key.name, satellite, setup_insights=register_insights
-            )
-
         # Red Hat Insights requires RHEL 6/7/8 repo and it is not
```
Gauravtalreja1, line 1055:
```suggestion
        # Red Hat Insights requires RHEL 6/7/8/9 repo and it is not
```

robottelo/hosts.py
```diff
@@ -1052,17 +1052,6 @@ def configure_rhai_client(
         :param register: Whether to register client to insights
```
Gauravtalreja1, line 1052: Could you use the same PR to describe use of `register_insights` param here?

ColeHiggins2:
> Do you know what is the error in registration specifically? I just wonder if it was failed `insights-client` installation during GR, which would explain why it works now when GR is done _after_ client pre-installation.
> 
> Otherwise I agree to exercise the `insights-client --register` via GR rather than separately.

I honestly dont know what changed between global registration and the prior. I was hoping to hear back from @shweta83 or @tpapaioa on this.

tpapaioa (changes requested):
The fix looks good to me. My only request is to update the method's name and docstring to remove references to rhai / red hat access insights, which is the old package name. `configure_rhai_client` could instead be `configure_insights_client`.

robottelo/hosts.py
ColeHiggins2: In this case, I think logicially it should stay where it is... We need to make sure those packages are installed before we register

robottelo/hosts.py
ColeHiggins2: its a setting when registering a client to a sat. You can set up insights during this process. This is true by default

ColeHiggins2:
trigger: test-robottelo
pytest: tests/foreman/cli/test_rhcloud_inventory.py -k 'test_positive_sync_inventory_status'

Satellite-QE:
**PRT Result**
```
Build Number: 8860
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/cli/test_rhcloud_inventory.py -k test_positive_sync_inventory_status --external-logging
Test Result : ========= 6 passed, 16 deselected, 706 warnings in 2645.36s (0:44:05) ==========
```

sambible (approved):
Would be nice to understand what changed here, but I don't think it's necessary, as this patch is logically sound.

robottelo/hosts.py
Gauravtalreja1: @ColeHiggins2 yes, but I meant if it could be described in the function metadata, which it isn't

robottelo/hosts.py
Gauravtalreja1: @ColeHiggins2 yes, but this step would be covered with `setup_insights=register_insights` when `register=True`, where this isn't required, but its required to install insights-client manually, when `register=False`, right?

Gauravtalreja1:
@sambible There is no point merging the PRs if the threads are open and discussions are happening, especially if the code owners require Tier2 review. OR if you feel the feedback isn't necessary or non-blocking, at least you should comment on those threads.

PR #16522, opened 2024-09-27
Add test coverage for SAT-26014

Gauravtalreja1:
### Problem Statement
Missing test coverage for SAT-26014

### Solution
Adding a destructive test to cover SAT-26014

### DIFF
--- a/tests/foreman/destructive/test_installer.py
+++ b/tests/foreman/destructive/test_installer.py
@@ -225,3 +225,33 @@ def test_negative_handle_invalid_certificate(cert_setup_destructive_teardown):
 
     result = satellite.execute('hammer ping')
     assert result.status == 0, f'Hammer Ping failed:\n{result.stderr}'
+
+
+def test_negative_installer_invalid_dns_forwarder(target_sat):
+    """Satellite installer should not do any impact if any invalid values are set for DNS forwarder
+
+    :id: 97b72faf-4684-4d8c-ae0e-1ebd5085620d
+
+    :steps:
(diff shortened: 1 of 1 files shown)


Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/destructive/test_installer.py -k test_negative_installer_invalid_dns_forwarder

Satellite-QE:
**PRT Result**
```
Build Number: 8823
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/destructive/test_installer.py -k test_negative_installer_invalid_dns_forwarder --external-logging
Test Result : ========== 1 passed, 5 deselected, 18 warnings in 1252.58s (0:20:52) ===========
```

rmynar (approved):
Looks good to me.

Just non-blocking comments:
Would you consider adding a positive scenario? It may be easily added into non-default installation: https://github.com/SatelliteQE/robottelo/blob/8109a0f8f86fabebd989b6bc033a8f5a8742df1d/tests/foreman/installer/test_installer.py#L296

What about cherry picking into older versions? 6.15, maybe 6.14...?

Gauravtalreja1:
@rmynar Thanks for your comments. 
If you see the related issue, then its about a negative scenario only which breaks the answer file, and for positive scenario its already been tested in our misc-provisioning Jenkins jobs, so I don't think if we should cover in new test separately, and about cherrypicking it to 6.15 or 6.14, so its only fixed for stream and 6.16 for now.

shubhamsg199 (approved):
Ack, looks good.

PR #16409, opened 2024-09-18
[6.16.z] Use client-2 repo against Satellite 6.16+

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/16397

### Problem Statement
Satellite 6.16 includes **Puppet 8** so there is now new `Client-2` repo that provides Puppet 8 agent while `Client` repo continues to provide Puppet 7 agent.

### Solution
Use `Client-2` repo when running against Satellite 6.16 and newer

PR #16537, opened 2024-09-30
[6.16.z] Switch back to client-1 repo against Satellite 6.16

lpramuk:
Reverts SatelliteQE/robottelo#16409
(revert 6.16.z cherrypick only)

### Problem
There are not yet **client tests** identified therefore we cannot cover both client-1 and client-2 repos w/o duplicating whole test matrix

### Solution
Temporary solution is to use **client-1** repo against Satellite **6.16** and use **client-2** repo against Satellite **Stream** so that we don't duplicate whole matrix while we have coverage for both client repos.
 (We are missing testing specific client repo across multiple satellite versions anyway.)

### DIFF
--- a/conf/dynaconf_hooks.py
+++ b/conf/dynaconf_hooks.py
@@ -149,9 +149,9 @@ def get_dogfood_satclient_repos(settings):
     for ver in rhels:
         data[f'RHEL{ver}'] = get_ohsnap_repo_url(
             settings,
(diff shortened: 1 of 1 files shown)


jyejare:
@lpramuk Provide more details in the description or link related PRs.

Gauravtalreja1 (approved):

PR #16550, opened 2024-10-02
[6.16.z] Fetch pre_upgrade data based on node properties

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/16423

### Problem Statement
During TFA I've seen several `post_upgrade` failures for "`rhel_contenthost`-parametrized" upgrade scenario tests where the pre-data could not be fetched from the `scenario_entities` file:
```
failed on setup with "Failed: Invalid test parameter: rhel7. Test data not found."
```

The reason is that the pre-upgrade test names are now suffixed with `-ipv4` or `-ipv6` so the `scenario_entities` hold items like this:
```
{"tests/upgrades/test_repository.py::TestSimpleContentAccessOnly::test_pre_simple_content_access_only[rhel7-ipv4]": {...}}
```
while [this piece of code](https://github.com/SatelliteQE/robottelo/blob/master/tests/upgrades/conftest.py#L267) requires the exact param, which is just `rhel7`, `rhel8`, `rhel9`.

### Solution(?)
There are probably more than one solution and I'm not sure which one would be the best.

Initially I thought we could go regardless the `ipvX` suffix since I supposed the pipelines were separated (first commit).
That's probably not a good solution when we start decorating tests with `['ipv4', 'ipv6']` explicitly, so I took the network type from the settings.

Please let me know if there is a better place where to get that information from in scope of the `pre_upgrade_data` fixture.

### DIFF
--- a/tests/upgrades/conftest.py
+++ b/tests/upgrades/conftest.py
@@ -90,6 +90,7 @@ import os
 from box import Box
 import pytest
 
+from robottelo.config import settings
 from robottelo.logging import logger
 from robottelo.utils.decorators.func_locker import lock_function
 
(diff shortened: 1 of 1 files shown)


vsedmik (approved):
CP looks good

PR #16553, opened 2024-10-02
avoiding host details will be seen in the test result

omkarkhatavkar:
### Problem Statement
The current pytest parameterization exposes sensitive host details, such as proxy URLs, in the test results. This can lead to unintended exposure of confidential information in logs or reports, which is undesirable when sharing test results over PRT

### Solution
To address this, we will update the pytest parameterization to use a custom ID (un_auth_proxy) instead of displaying the actual proxy URL. By doing so, we ensure that the test results are cleaner and more secure, as sensitive host details are hidden.


ColeHiggins2 (approved):

ogajduse (approved):
Thanks for addressing this.

PR #16575, opened 2024-10-04
Add conditioning for pool attach

vsedmik:
### Problem Statement
The default RH-Sat-QE account will be flipped to the SCA mode soon (currently the SCA is disabled there). As a result pools manipulation like attach/auto-attach won't be applicable when registering hosts via that account.

### Solution
We should get rid of using pools. But before we do so and before we really flip for SCA, we could check the subscription status of the host being registered and attach pool only when it is still in the entitlement mode.

### Testing
I tested the dummie bellow locally and ensured the registration succeeds for both account modes, SCA and entitlement.
Also tried Manifester using the offline token from my SCA account and worked well.
```
def test_dummie(target_sat, rhel_contenthost, module_sca_manifest_org):
    rhel_contenthost.register_to_cdn()
    rhel_contenthost.unregister()
```
Result:
```
(venv) [vsedmik@fedora robottelo]$ pytest tests/foreman/cli/test_host.py -k dummie
=============================== test session starts ===============================
collected 102 items / 96 deselected / 6 selected                                                                                                                                                                                       

tests/foreman/cli/test_host.py ......                                       [100%]

=========== 6 passed, 96 deselected, 286 warnings in 1569.86s (0:26:09) ===========
```

### Other considerations
There are still several areas in robottelo where the subscriptions are attached like in the entitlement mode - typically Activation Keys and Hosts. Since this functionality is still supported by Satellite, I'm not going to remove that yet.

### DIFF
--- a/robottelo/hosts.py
+++ b/robottelo/hosts.py
@@ -46,6 +46,7 @@ from robottelo.constants import (
     RHSSO_RESET_PASSWORD,
     RHSSO_USER_UPDATE,
     SATELLITE_VERSION,
+    SM_OVERALL_STATUS,
 )
 from robottelo.exceptions import CLIFactoryError, DownloadFileError, HostPingFailed
 from robottelo.host_helpers import CapsuleMixins, ContentHostMixins, SatelliteMixins
@@ -1402,8 +1403,6 @@ class ContentHost(Host, ContentHostMixins):
 
     def register_to_cdn(self, pool_ids=None, enable_proxy=False):
         """Subscribe satellite to CDN"""
-        if pool_ids is None:
-            pool_ids = [settings.subscription.rhn_poolid]
         self.reset_rhsm()
         cmd_result = self.register_contenthost(
             org=None,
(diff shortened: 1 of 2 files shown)


robottelo/hosts.py
```diff
@@ -1416,11 +1415,16 @@ def register_to_cdn(self, pool_ids=None, enable_proxy=False):
...
             )
-        cmd_result = self.subscription_manager_attach_pool(pool_ids)[0]
-        if cmd_result.status != 0:
-            raise ContentHostError(
-                f'Error during pool attachment, command output: {cmd_result.stdout}'
-            )
+        # Attach a pool only if the Org isn't SCA yet
+        sub_status = self.subscription_manager_status().stdout
+        if SM_OVERALL_STATUS['disabled'] not in sub_status:
+            if pool_ids is None:
```
rplevka, line 1421:
```
if pool_ids in [None, []]:
```
Might save us from an eventual `IndexError` in case someone passes empty list.
Also, I realized, we don't verify the type of the passed value.
adding it to the beginning of the method could allow us to bail out fast.

rplevka (approved):

rplevka (approved):
Pretty straightforward change.
Left a minor comment regarding something not introduced by this MR, so I am leaving that optional and giving my ACK straight away

robottelo/hosts.py
```diff
@@ -1416,11 +1415,16 @@ def register_to_cdn(self, pool_ids=None, enable_proxy=False):
...
+                pool_ids = [settings.subscription.rhn_poolid]
+            cmd_result = self.subscription_manager_attach_pool(pool_ids)[0]
```
rplevka, line 1423: wait. only now i realized - we accept a list of pool ids but finally only use the first one? hasn't this been a bug in the method?

robottelo/hosts.py
vsedmik: I can add this check, it makes sense. This and maybe one to check the passed value is a HEX number?
I intended to make as few changes as possible to make this PR clear/straightforward and avoid curious questions. :)
Anyway, the pools are going away soon with SCA, so... we can add better validations but they will be in unreachable code section.

robottelo/hosts.py
vsedmik: Yeah, this is how it was [since beginning](https://github.com/SatelliteQE/robottelo/commit/f54efe9c62009473808b6eb1f473ccb9f176e466). I don't know the reason behind, we could ask @Gauravtalreja1 or we don't need to slam there `assert all` and that's it. But again, the pools are going away soon ... unreachable check.

robottelo/hosts.py
rplevka: You don't have to, that's why I gave you an ACK

vsedmik: Hah, too late! They are in now. :smiley:

pondrejk (approved):

synkd (approved):

PR #16594, opened 2024-10-07
Update is_open_jira issue handler functionality

jameerpathan111:
### Problem Statement
- `fix_version` doesn't have to be part of `is_open_jira` and is actually giving false positive result

### Solution
- Remove the condition.


pondrejk (approved):
I checked the patch on a test with closed jira in the :BlockedBy element, the test gets collected successfully

PR #16684, opened 2024-10-16
Actions comment pull request bump fix

omkarkhatavkar:
### Problem Statement
After upgrade to from version 2 to 3 for thollander/actions-comment-pull-request@v3 causing issues in PRT and other GHA's after looking at closely https://github.com/SatelliteQE/robottelo/actions/runs/11361084949/job/31600069885#step:2:2 I believe this mostly because of the changes in the parameters 

### Solution
updated the GHA to use the updated parameters

PR #16689, opened 2024-10-16
parameters names have changed after the actions-comment-pull-request …

omkarkhatavkar:
cherrypick of https://github.com/SatelliteQE/robottelo/pull/16684

### Problem Statement
After upgrade to from version 2 to 3 for thollander/actions-comment-pull-request@v3 causing issues in PRT and other GHA's after looking at closely https://github.com/SatelliteQE/robottelo/actions/runs/11361084949/job/31600069885#step:2:2 I believe this mostly because of the changes in the parameters 

### Solution
updated the GHA to use the updated parameters


omkarkhatavkar:
let's avoid this change right now as it is working in z streams

PR #16694, opened 2024-10-17
use packaged pytest, not some venv

evgeni:
### Problem Statement

The current FAM test setup relies on being able to reach the Internet, which is "bad design"™ but also actually broken in IPv6-only environments because our lab network has IPv6, but no route to the broader Internet.

### Solution

Instead of installing test dependencies via `pip` from the Internet, install them via `dnf` from the RH CDN and use them in the tests here.
There is an accompanying PR against the repo that carries the setup playbooks.

### DIFF
--- a/tests/foreman/sys/test_fam.py
+++ b/tests/foreman/sys/test_fam.py
@@ -74,6 +74,13 @@ def setup_fam(module_target_sat, module_sca_manifest, install_import_ansible_rol
     # Edit Makefile to not try to rebuild the collection when tests run
     module_target_sat.execute(f"sed -i '/^live/ s/$(MANIFEST)//' {FAM_ROOT_DIR}/Makefile")
 
+    # Edit Makefile to use passed-in pytest
+    # Can be removed once https://github.com/theforeman/foreman-ansible-modules/pull/1788
+    # is present in all relevant branches
+    module_target_sat.execute(
+        f"sed -i '/test_crud/ s/pytest/$(PYTEST_COMMAND)/' {FAM_ROOT_DIR}/Makefile"
+    )
+
     # Upload manifest to test playbooks directory
     module_target_sat.put(str(module_sca_manifest.path), str(module_sca_manifest.name))
     module_target_sat.execute(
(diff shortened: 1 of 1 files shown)


jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman/sys/test_fam.py
network_type: ipv6

evgeni:
@jameerpathan111 given the executed `fam-test-setup` playbook is the same for every Satellite branch, I think this needs cherry-picking to all supported branches.

Satellite-QE:
**PRT Result**
```
Build Number: 8994
Build Status: UNSTABLE
PRT Comment: pytest tests/foreman/sys/test_fam.py --external-logging
Test Result : =========== 108 failed, 2 passed, 472 warnings in 2113.86s (0:35:13) ===========
```

evgeni:
The PRT failure is expected, you need to run it against an AAP that has my deployment patches in.

jameerpathan111:
@evgeni Acked. Can you fix pre-commit failures?

evgeni:
Done!

tests/foreman/sys/test_fam.py
```diff
@@ -74,6 +74,13 @@ def setup_fam(module_target_sat, module_sca_manifest, install_import_ansible_rol
     # Edit Makefile to not try to rebuild the collection when tests run
     module_target_sat.execute(f"sed -i '/^live/ s/$(MANIFEST)//' {FAM_ROOT_DIR}/Makefile")
```
ekohl, line 75: Not a blocker, but could we introduce an environment variable for this as well so we can avoid patching code altogether?

evgeni: Maybe, probably. But not today ;)

evgeni:
Rebased and updated to match the last version of https://github.com/theforeman/foreman-ansible-modules/pull/1788 -- now only replacing `pytest` to `$(PYTEST_COMMAND)` and letting the rest of the `-vv` handling to other means.

PR #16728, opened 2024-10-22
prepare puppet classes needed for FAM tests in setup_fam

evgeni:
### Problem Statement

* FAM tests need Puppet modules for two things:
  * Testing of assignment of Puppet classes to hosts and hostgroups (`host` and `hostgroup` tests)
  * Testing of import of Puppet classes (`puppetclasses_import` test)
* For assignment, those Puppet classes need to be already imported into Foreman
* For importing "obviously" not
* Additionally, `puppet module install <module>` requires access to Puppet Forge, which might or might not be present during the `fam-test-setup` workflow

### Solution

* Instead of handling this in the `fam-test-setup` workflow, move the whole setup to `robottelo`
* Use `puppetlabs-ntp` from https://github.com/SatelliteQE/fedorapeople-repos/, as we use parameters from that module to test parameter assignment
* Fake `puppet-prometheus` as we don't use anything of the module, just the fact that the classes are present
* Fake another `fakemodule` that can be imported when the tests run.

### Example

```
trigger: test-robottelo
pytest: tests/foreman/sys/test_fam.py
```

### DIFF
--- a/tests/foreman/sys/test_fam.py
+++ b/tests/foreman/sys/test_fam.py
@@ -91,6 +91,38 @@ def setup_fam(module_target_sat, module_sca_manifest, install_import_ansible_rol
         f'''sed -i 's|subscription_manifest_path:.*|subscription_manifest_path: "data/{module_sca_manifest.name}"|g' {config_file}'''
     )
 
+    repo_path = '/fake_puppet1/system/releases/p/puppetlabs/'
+    module_tarball = 'puppetlabs-ntp-3.0.3.tar.gz'
+    local_path = '/tmp'
+    module_target_sat.execute(
+        f'curl --output {local_path}/{module_tarball} {settings.robottelo.repos_hosting_url}{repo_path}{module_tarball}',
+    )
+    module_target_sat.execute(
+        f'puppet module install --ignore-dependencies {local_path}/{module_tarball}'
+    )
+
+    def create_fake_module(module_target_sat, module_name, module_classes):
+        base_dir = '/etc/puppetlabs/code/environments/production/modules'
+        module_dir = f'{base_dir}/{module_name}'
+        manifest_dir = f'{module_dir}/manifests'
+        module_target_sat.execute(f'mkdir -p {manifest_dir}')
+        for module_class in module_classes:
+            full_class = module_name if module_class == 'init' else f'{module_name}::{module_class}'
+            module_target_sat.execute(
+                f'echo "class {full_class}(){{}}" > {manifest_dir}/{module_class}.pp'
+            )
+
+    create_fake_module(
+        module_target_sat,
+        'prometheus',
+        ['init', 'haproxy_exporter', 'redis_exporter', 'statsd_exporter'],
+    )
+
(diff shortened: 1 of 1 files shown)


tests/foreman/sys/test_fam.py
```diff
@@ -91,6 +91,42 @@ def setup_fam(module_target_sat, module_sca_manifest, install_import_ansible_rol
...
+        module_target_sat,
+        'prometheus',
+        ['init', 'haproxy_exporter', 'redis_exporter', 'statsd_exporter'],
+    )
+
+    smart_proxy = (
+        module_target_sat.api.SmartProxy()
+        .search(query={'search': f'name={module_target_sat.hostname}'})[0]
+        .read()
+    )
```
jameerpathan111, line 125:
```suggestion
    smart_proxy = module_target_sat.nailgun_smart_proxy.read()
```

jameerpathan111 (approved):
Looks alright to me, left small nitpick.

tests/foreman/sys/test_fam.py
evgeni: TIL, thanks!

tests/foreman/sys/test_fam.py
```diff
@@ -91,6 +91,38 @@ def setup_fam(module_target_sat, module_sca_manifest, install_import_ansible_rol
...
+    module_tarball = 'puppetlabs-ntp-3.0.3.tar.gz'
+    local_path = '/tmp'
+    module_target_sat.execute(
+        f'curl --output {local_path}/{module_tarball} {settings.robottelo.repos_hosting_url}{repo_path}{module_tarball}',
+    )
+    module_target_sat.execute(
+        f'puppet module install --ignore-dependencies {local_path}/{module_tarball}'
+    )
+
+    def create_fake_module(module_target_sat, module_name, module_classes):
```
evgeni, line 104: Jameer asked via Slack what's exactly happening here, and I think we can write it down here in the PR for anyone who will come back and read it later:

I need puppet modules for the tests. 1/ the prometheus module (with those exporter subclasses) and 2/ "some rando module, just something that can be imported"
in both cases, I don't actually care about them to have any working puppet code in them, as long as the names (of modules/classes) are as I want them to

a trivial Puppet class definition is:
```
class some_name(){
}
```
or also
```
class some_name(){}
```
if I want to ignore newlines/readability

so the code (`create_fake_module`) goes and given the name of a module (e.g. `prometheus`) and classes inside that module (e.g. `['init', 'haproxy_exporter', 'redis_exporter', 'statsd_exporter']`) it:
1. creates the folder where these files would live (`/etc/puppetlabs/code/environments/production/modules/prometheus/manifests`) (and it's parents, `mkdir -p` does that for us)
2. in this folder it creates "empty" class files, an `init.pp` with `class prometheus(){}`, a `haproxy_exporter.pp` with `class prometheus::haproxy_exporter(){}`, etc…

For Puppet/Foreman this looks like totally legit normal Puppet modules, they can be imported into Foreman and applied to Hosts/Hostsgroups, and that's all we care about

PR #16753, opened 2024-10-24
Remove reverse proxy assertion code from test_capsule_installation

jameerpathan111:
### Problem Statement
We have dropped support for port 8443 as a reverse proxy in Stream and it's causing test_capsule_installation to fail.

### Solution
- Remove code for reverse proxy


LadislavVasina1 (approved):
Seems good to me.

rmynar (approved):

PR #16810, opened 2024-10-29
[6.16.z] Add conditioning for pool attach

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/16575

### Problem Statement
The default RH-Sat-QE account will be flipped to the SCA mode soon (currently the SCA is disabled there). As a result pools manipulation like attach/auto-attach won't be applicable when registering hosts via that account.

### Solution
We should get rid of using pools. But before we do so and before we really flip for SCA, we could check the subscription status of the host being registered and attach pool only when it is still in the entitlement mode.

### Testing
I tested the dummie bellow locally and ensured the registration succeeds for both account modes, SCA and entitlement.
Also tried Manifester using the offline token from my SCA account and worked well.
```
def test_dummie(target_sat, rhel_contenthost, module_sca_manifest_org):
    rhel_contenthost.register_to_cdn()
    rhel_contenthost.unregister()
```
Result:
```
(venv) [vsedmik@fedora robottelo]$ pytest tests/foreman/cli/test_host.py -k dummie
=============================== test session starts ===============================
collected 102 items / 96 deselected / 6 selected                                                                                                                                                                                       

tests/foreman/cli/test_host.py ......                                       [100%]

=========== 6 passed, 96 deselected, 286 warnings in 1569.86s (0:26:09) ===========
```

### Other considerations
There are still several areas in robottelo where the subscriptions are attached like in the entitlement mode - typically Activation Keys and Hosts. Since this functionality is still supported by Satellite, I'm not going to remove that yet.

### DIFF
--- a/robottelo/constants/__init__.py
+++ b/robottelo/constants/__init__.py
@@ -317,6 +317,7 @@ SM_OVERALL_STATUS = {
     'current': 'Overall Status: Current',
     'invalid': 'Overall Status: Invalid',
     'insufficient': 'Overall Status: Insufficient',
+    'disabled': 'Overall Status: Disabled',
     'unknown': 'Overall Status: Unknown',
 }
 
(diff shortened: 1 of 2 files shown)


vsedmik (approved):
Looks CPed properly.

PR #16903, opened 2024-11-13
Post deployments for content hosts, starting with FIPS

jyejare:
### Problem Statement
- The existing FIPS WF is going to be deleted as RHEL10 demands the enablement of FIPS as a separate task
- Resulting the WF to be deleted from Lab and introduce new `enable-fips` WF for any RHEL
- Currently, robottelo does not support 2 steps installation of FIPS and hence that needs to handle in robottelo

### Solution
- Introducing the post deployments for content hosts which should run procedures on content host based on conditions
- Starting with FIPS, when the content host is entitled for FIPS the `enable_fips` WF should run on the given host hence supporting 2 steps FIPS Content host
- Also, the existing content host template is updated for FIPS to deploy_rhel instead of the previous (To be Removed) WF.

### DIFF
--- a/conf/content_host.yaml.template
+++ b/conf/content_host.yaml.template
@@ -2,7 +2,7 @@ content_host:
   default_rhel_version: 7
   rhel6:
     vm:
-      workflow: deploy-base-rhel
+      workflow: deploy-rhel
       deploy_rhel_version: '6'
       target_memory: 1GiB
       target_cores: 1
@@ -10,7 +10,7 @@ content_host:
       container_host: rhel6:latest
   rhel7:
     vm:
-      workflow: deploy-base-rhel
+      workflow: deploy-rhel
       deploy_rhel_version: '7'
       target_memory: 1GiB
       target_cores: 1
@@ -18,13 +18,13 @@ content_host:
       container_host: ubi7:latest
   rhel7_fips:
     vm:
-      workflow: deploy-base-rhel-fips
+      workflow: deploy-rhel
       deploy_rhel_version: '7'
       target_memory: 1GiB
       target_cores: 1
   rhel8:
     vm:
-      workflow: deploy-base-rhel
+      workflow: deploy-rhel
       deploy_rhel_version: '8'
       target_memory: 1536 MiB
       target_cores: 1
@@ -32,13 +32,13 @@ content_host:
       container_host: ubi8:latest
   rhel8_fips:
     vm:
-      workflow: deploy-base-rhel-fips
+      workflow: deploy-rhel
       deploy_rhel_version: '8'
       target_memory: 1536 MiB
       target_cores: 1
   rhel9:
     vm:
-      workflow: deploy-base-rhel
+      workflow: deploy-rhel
       deploy_rhel_version: '9'
       target_memory: 1536 MiB
       target_cores: 1
@@ -46,7 +46,7 @@ content_host:
       container_host: ubi9:latest
   rhel9_fips:
     vm:
-      workflow: deploy-base-rhel-fips
+      workflow: deploy-rhel
       deploy_rhel_version: '9'
       target_memory: 1536 MiB
       target_cores: 1
--- a/conf/content_host.yaml.template
+++ b/conf/content_host.yaml.template
@@ -50,6 +50,12 @@ content_host:
       deploy_rhel_version: '9'
       target_memory: 1536 MiB
       target_cores: 1
+  rhel10_fips:
+    vm:
+      workflow: deploy-template
+      deploy_rhel_version: '10'
+      target_memory: 1536 MiB
+      target_cores: 1
   centos7:
     vm:
       workflow: deploy-centos
--- a/pytest_fixtures/core/contenthosts.py
+++ b/pytest_fixtures/core/contenthosts.py
@@ -41,12 +41,22 @@ def host_conf(request):
     return conf
 
 
+def run_post_deployments(request, host):
+    """Execute post deploy"""
+    params = {}
+    if hasattr(request, 'param'):
+        params = request.param
+    if 'fips' in params.get('rhel_version'):
+        Broker().execute(workflow='enable-fips', target_host=host.name)
+
+
 @pytest.fixture
 def rhel_contenthost(request):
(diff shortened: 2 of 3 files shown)


pytest_fixtures/core/contenthosts.py
```diff
@@ -41,12 +41,22 @@ def host_conf(request):
     return conf
 
 
+def run_post_deployments(request, host):
+    """Execute post deploy"""
+    params = {}
+    if hasattr(request, 'param'):
+        params = request.param
+    if 'fips' in params.get('rhel_version'):
+        Broker().execute(workflow='enable-fips', target_host=host.name)
```
ogajduse, line 50: Did you consider putting this logic into `ContentHost.setup` method? If we would go that route, we would need to pass `params['rhel_version']` (as a metadata) to broker through the `host_conf` method, so we can read it later on in the `setup` method. This is something we already do with `blank=True`.

https://github.com/ogajduse/robottelo/blob/8fff2ede1a86937f0dd6f4275b3b603de39c8c3b/pytest_fixtures/component/provision_pxe.py#L231

conf/content_host.yaml.template
```diff
@@ -2,51 +2,51 @@ content_host:
...
+      workflow: deploy-rhel
       deploy_rhel_version: '9'
       target_memory: 1536 MiB
       target_cores: 1
     container:
       container_host: ubi9:latest
   rhel9_fips:
     vm:
-      workflow: deploy-base-rhel-fips
+      workflow: deploy-rhel
```
bherrin3, line 49: @jyejare , no matter what, don't you need to add an entry for RHEL 10 and the correct workflow call there?

Please note that if implemented this way, the value for `workflow` will need to be changed on GA and the new non-GA will have to be inserted afterwards.

```suggestion
      workflow: deploy-template
      deploy_rhel_version: '10'
```

conf/content_host.yaml.template
```diff
@@ -2,51 +2,51 @@ content_host:
   default_rhel_version: 7
   rhel6:
     vm:
-      workflow: deploy-base-rhel
+      workflow: deploy-rhel
       deploy_rhel_version: '6'
       target_memory: 1GiB
```
bherrin3, line 7: Please note, at this time (may be changing soon), the values for `target_memory` and `target_cores` do nothing

pytest_fixtures/core/contenthosts.py
jyejare: Did not consider that for sure, but It was a perfect suggestion.

I choose this route to keep the content host procedures native to content host fixtures in this module for easy understanding and debug.

But anyways now we had to close the PR :(

conf/content_host.yaml.template
jyejare: Yes it needs update for rhel10.  The main change is in CI as the actual conf resides there, this is just a template.

conf/content_host.yaml.template
jyejare: Yeah its a old config template and can be updated. BTW why are we going away from memory and cores conf? cause we Do have profiles to choose ?

bherrin3: You currently cannot select target_*. You choose a flavor only.

jyejare: BTW Not a blocker for this MR, so once `may be changing soon` is confirmed, we will change this in the template.

conf/content_host.yaml.template
jyejare: Addressed by adding a new config for RHEL10.

pytest_fixtures/core/contenthosts.py
jyejare: @ogajduse We are again allowed to go this path after some brainstorming in the Lab team. I have addressed your comment and its subject to verification post `enable-fips` WF merged in Tower.

conf/content_host.yaml.template
```diff
@@ -2,54 +2,60 @@ content_host:
...
     container:
       container_host: ubi9:latest
   rhel9_fips:
     vm:
-      workflow: deploy-base-rhel-fips
+      workflow: deploy-rhel
       deploy_rhel_version: '9'
       target_memory: 1536 MiB
       target_cores: 1
+  rhel10_fips:
```
Gauravtalreja1, line 53:
```suggestion
  rhel10:
```

conf/content_host.yaml.template
```diff
@@ -2,54 +2,60 @@ content_host:
...
       deploy_rhel_version: '9'
       target_memory: 1536 MiB
       target_cores: 1
     container:
       container_host: ubi9:latest
   rhel9_fips:
     vm:
-      workflow: deploy-base-rhel-fips
+      workflow: deploy-rhel
       deploy_rhel_version: '9'
```
Gauravtalreja1, line 50:
```suggestion
      deploy_rhel_version: '9'
      deploy_fips_enable: true
```

conf/content_host.yaml.template
```diff
@@ -2,54 +2,60 @@ content_host:
...
       deploy_rhel_version: '7'
       target_memory: 1GiB
       target_cores: 1
   rhel8:
     vm:
-      workflow: deploy-base-rhel
+      workflow: deploy-rhel
       deploy_rhel_version: '8'
       target_memory: 1536 MiB
       target_cores: 1
```
Gauravtalreja1, line 30: target_cores and target_memory options can be removed since it won't do anything for VM deployed in RHOS

pytest_fixtures/core/contenthosts.py
```diff
@@ -37,6 +37,9 @@ def host_conf(request):
         deploy_kwargs = settings.content_host.get(_rhelver).to_dict().get('vm', {})
         if network := params.get('network'):
             deploy_kwargs.update({'deploy_network_type': network})
+    # Post_deploy_workflow_runs to be run in ContentHost.setup method
+    if 'fips' in params.get('rhel_version'):
```
Gauravtalreja1, line 41:
```suggestion
    if 'rhel10_fips' in params.get('rhel_version'):
```

PR #16922, opened 2024-11-14
Stub out the ostree tests

vsedmik:
### Problem Statement
Ostree content type is not supported in downstream since 6.10. Ostree tests used to be been skipped based on the BZ status (CLOSED_WONTFIX) until #16347 was merged and now they are failing with validation error.

### Solution
Stub the tests until the ostree is supported again.


chris1984 (approved):
Looks fine to me, all 🍏  with PRT 🚢 🇮🇹

LadislavVasina1 (approved):

PR #16928, opened 2024-11-15
Refactor ipv6 http_proxy code

jameerpathan111:
### Problem Statement
- The current design of `enable_ipv6_http_proxy` and `disable_ipv6_http_proxy` is incorrect and confusing. 
- `Satellite` and `Capsule` classes of `robottelo/hosts.py` have `enable_ipv6_http_proxy`. The one present in Satellite is for setting HTTP proxy in Satellite settings(content and general HTTP proxy). And the one in Capsule is for enabling HTTP proxy for dnf and rhsm.
- This design doesn't allow Satellite to set dnf and rhsm proxies.

### Solution
- Remove`enable_ipv6_http_proxy` and `disable_ipv6_http_proxy` from `Satellite` and `Capsule` classes.
- Add `enable_satellite_http_proxy` and `disable_satellite_http_proxy` to enable/disable ipv4/ipv6 HTTP proxy.  
- Add `enable_satellite_ipv6_http_proxy` in `Satellite` class to set Ipv6 HTTP proxy in Satellite settings, rhsm, and dnf.
- Add `enable_ipv6_dnf_and_rhsm_proxy` to `ContenHost` class so all the three kind of hosts can use it (Satellite, Capsule, Client).
- This PR also removes `test_installer_modules_check` and covers it in `test_satellite_installation`.
- It also remove http_proxy test parameterization for `test_capsule_installation` as it'll be covered by default (Satellite will have HTTP proxy enabled)

### Related Issues
- SAT-29467
- Depends on satellite-jenkins#1556

### DIFF
--- a/conf/http_proxy.yaml.template
+++ b/conf/http_proxy.yaml.template
@@ -1,5 +1,6 @@
 HTTP_PROXY:
   UN_AUTH_PROXY_URL:  # http://proxy-01.example.com:3423
+  HTTP_PROXY_IPV6_URL:  # http://proxy-01.ipv6.example.com:3423
   AUTH_PROXY_URL:  # http://proxy-02.example.com:3423
   USERNAME: auth-proxy-user
   PASSWORD: auth-proxy-password
--- a/conf/migrations.py
+++ b/conf/migrations.py
@@ -35,3 +35,15 @@ def migration_231129_deploy_workflow(settings, data):
             logger.info(
                 f'Migrated {product_type}.DEPLOY_WORKFLOW to {product_type}.DEPLOY_WORKFLOWS'
             )
+
+
+def migration_241120_http_proxy_ipv6_url(settings, data):
+    """Migrates server.http_proxy_ipv6_url to http_proxy.http_proxy_ipv6_url"""
+    if (
+        settings.server.get('http_proxy_ipv6_url')
+        and isinstance(settings.server.http_proxy_ipv6_url, str)
+        and not settings.http_proxy.get('http_proxy_ipv6_url')
+    ):
+        data.http_proxy = {}
+        data.http_proxy.http_proxy_ipv6_url = settings.server.http_proxy_ipv6_url
+        logger.info('Migrated SERVER.HTTP_PROXY_IPv6_URL to HTTP_PROXY.HTTP_PROXY_IPV6_URL')
--- a/pytest_fixtures/component/maintain.py
+++ b/pytest_fixtures/component/maintain.py
@@ -25,7 +25,7 @@ def module_stash(request):
 def sat_maintain(request, module_target_sat, module_capsule_configured):
     if settings.remotedb.server:
         sat = Satellite(settings.remotedb.server)
-        sat.enable_ipv6_http_proxy()
+        sat.enable_satellite_ipv6_http_proxy()
         yield sat
     else:
         module_target_sat.register_to_cdn(pool_ids=settings.subscription.fm_rhn_poolid.split())
--- a/pytest_fixtures/core/sat_cap_factory.py
+++ b/pytest_fixtures/core/sat_cap_factory.py
@@ -31,7 +31,7 @@ def resolve_deploy_args(args_dict):
 def _target_satellite_host(request, satellite_factory):
     if 'sanity' not in request.config.option.markexpr:
         new_sat = satellite_factory()
-        new_sat.enable_ipv6_http_proxy()
+        new_sat.enable_satellite_ipv6_http_proxy()
         yield new_sat
         new_sat.teardown()
         Broker(hosts=[new_sat]).checkin()
@@ -49,7 +49,7 @@ def cached_capsule_cdn_register(hostname=None):
 def _target_capsule_host(request, capsule_factory):
     if 'sanity' not in request.config.option.markexpr and not request.config.option.n_minus:
         new_cap = capsule_factory()
-        new_cap.enable_ipv6_http_proxy()
+        new_cap.enable_ipv6_dnf_and_rhsm_proxy()
         yield new_cap
         new_cap.teardown()
         Broker(hosts=[new_cap]).checkin()
@@ -96,7 +96,7 @@ def satellite_factory():
 def large_capsule_host(capsule_factory):
     """A fixture that provides a Capsule based on config settings"""
     new_cap = capsule_factory(deploy_flavor=settings.flavors.custom_db)
-    new_cap.enable_ipv6_http_proxy()
+    new_cap.enable_ipv6_dnf_and_rhsm_proxy()
     yield new_cap
     new_cap.teardown()
     Broker(hosts=[new_cap]).checkin()
@@ -242,7 +242,7 @@ def module_lb_capsule(retry_limit=3, delay=300, **broker_args):
         )
         cap_hosts = wait_for(hosts.checkout, timeout=timeout, delay=delay)
 
-    [cap.enable_ipv6_http_proxy() for cap in cap_hosts.out]
+    [cap.enable_ipv6_dnf_and_rhsm_proxy() for cap in cap_hosts.out]
     yield cap_hosts.out
 
     [cap.teardown() for cap in cap_hosts.out]
@@ -277,7 +277,7 @@ def parametrized_enrolled_sat(
 ):
     """Yields a Satellite enrolled into [IDM, AD] as parameter."""
     new_sat = satellite_factory()
-    new_sat.enable_ipv6_http_proxy()
+    new_sat.enable_satellite_ipv6_http_proxy()
     ipa_host = IPAHost(new_sat)
     new_sat.register_to_cdn()
     if 'IDM' in request.param:
@@ -337,7 +337,7 @@ def cap_ready_rhel():
         'workflow': settings.capsule.deploy_workflows.os,
     }
     with Broker(**deploy_args, host_class=Capsule) as host:
-        host.enable_ipv6_http_proxy()
+        host.enable_ipv6_dnf_and_rhsm_proxy()
         yield host
 
 
@@ -395,7 +395,7 @@ def installer_satellite(request):
         ).get_command(),
         timeout='30m',
     )
-    sat.enable_ipv6_http_proxy()
+    sat.enable_satellite_ipv6_http_proxy()
     if 'sanity' in request.config.option.markexpr:
         configure_nailgun()
         configure_airgun()
--- a/robottelo/config/validators.py
+++ b/robottelo/config/validators.py
@@ -36,12 +36,6 @@ VALIDATORS = dict(
         Validator('server.ssh_password', default=None),
         Validator('server.verify_ca', default=False),
         Validator('server.is_ipv6', is_type_of=bool, default=False),
-        # validate http_proxy_ipv6_url only if is_ipv6 is True
-        Validator(
-            'server.http_proxy_ipv6_url',
-            is_type_of=str,
-            when=Validator('server.is_ipv6', eq=True),
-        ),
     ],
     content_host=[
         Validator('content_host.default_rhel_version', must_exist=True),
@@ -160,6 +154,12 @@ VALIDATORS = dict(
             'http_proxy.username',
             'http_proxy.password',
             must_exist=True,
+        ),
+        # validate http_proxy_ipv6_url only if server.is_ipv6 is True
+        Validator(
+            'http_proxy.http_proxy_ipv6_url',
+            is_type_of=str,
+            when=Validator('server.is_ipv6', eq=True),
         ),
     ],
     ipa=[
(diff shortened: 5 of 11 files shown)


conf/http_proxy.yaml.template
```diff
@@ -1,5 +1,6 @@
 HTTP_PROXY:
   UN_AUTH_PROXY_URL:  # http://proxy-01.example.com:3423
+  HTTP_PROXY_IPv6_URL:  # http://proxy-01.ipv6.example.com:3423
```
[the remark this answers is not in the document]
jameerpathan111: I think it makes sense to have all the HTTP proxies in one place and not all over the configs. Let's not think of http_proxy.yaml as just component-level config. The usage of proxies there is no different than IPv6 proxy.
And it's not specific to robottelo, it's being used in Satellite settings too, and would be needed in HTTP proxy component moving forward. I belive if someone is looking for a proxy then the first thought would be to check http_proxy.yaml

pytest_fixtures/component/maintain.py
```diff
@@ -25,7 +25,7 @@ def module_stash(request):
 def sat_maintain(request, module_target_sat, module_capsule_configured):
     if settings.remotedb.server:
         sat = Satellite(settings.remotedb.server)
-        sat.enable_ipv6_http_proxy()
+        sat.enable_satellite_ipv6_http_proxy()
```
jyejare, line 28:
```suggestion
        sat.enable_satellite_ipv6_http_proxy()
```
Arent we duplicating the `sat` here?

The object name `sat` itself prouds the reference onto which HTTP proxy is being enabled. Hence the explicit `satellite` keyword in function name should now be needed.

pytest_fixtures/core/sat_cap_factory.py
```diff
@@ -96,7 +96,7 @@ def factory(retry_limit=3, delay=300, workflow=None, **broker_args):
 def large_capsule_host(capsule_factory):
     """A fixture that provides a Capsule based on config settings"""
     new_cap = capsule_factory(deploy_flavor=settings.flavors.custom_db)
-    new_cap.enable_ipv6_http_proxy()
+    new_cap.enable_ipv6_dnf_and_rhsm_proxy()
```
jyejare, line 99: A change that makes sense.

robottelo/config/validators.py
```diff
@@ -161,6 +155,12 @@
             'http_proxy.password',
             must_exist=True,
         ),
+        # validate http_proxy_ipv6_url only if server.is_ipv6 is True
+        Validator(
+            'http_proxy.http_proxy_ipv6_url',
```
jyejare, line 160: I think keeping the global HTTP Proxy for all satellite applications, the server conf is a good place.

HTTP Proxy conf is made to configure within the satellite, whereas IPV6 HTTP proxy is inside and outside as well to have cross network communication.

jyejare (changes requested):

robottelo/config/validators.py
jameerpathan111: `HTTP Proxy conf is made to configure within the satellite` It's not, it can be used to configure on system level as well. It's not different from Ipv6 HTTP proxy. It can do things ipv6 can do. so I don't see a reason to differentiate. https://github.com/SatelliteQE/robottelo/pull/16928#discussion_r1844266242

jyejare: Another reason keeping it in `server.yaml` is that's the yaml sets global things for robottelo including the `is_ipv6` switch. Though its duplicating the key from http_proxy yaml, having it under `server.yaml` and near to `is_ipv6` setting increases the usability of robottelo.

pytest_fixtures/component/maintain.py
jameerpathan111: I don't understand what you mean; your code suggestion has no change.

jyejare:
```suggestion
        sat.enable_ipv6_http_proxy()
```

robottelo/config/validators.py
jameerpathan111: By that logic, we'll be making Robottelo unnecessarily bulky and a lot of other settings could fall into such reasoning. Not to mention that we'll be basically ignoring the very general rule of clustering similar settings in same yaml. But I'd leave the final decision to you/JPL.

pytest_fixtures/component/maintain.py
ogajduse: Isn't it better to be explicit in the name? This method does setup related to Satellite - it changes Satellite's settings through API. 
I, as a code reader, find it more straightforward to read `enable_satellite_ipv6_http_proxy` or `enable_sat_ipv6_http_proxy` as it tells me that it does something with Satellite itself.

The `enable_ipv6_http_proxy` method can be defined in the `ContentHost` class to do something completely different from modifying Satellite settings.

Host object methods count is constantly increasing, therefore I advocate for explicitness - for keeping `satellite` or `sat` in the method name.

jameerpathan111:
I ran ipv6 installer tests and all the tests passed. See satellite-jenkins#1556 for test result link

conf/http_proxy.yaml.template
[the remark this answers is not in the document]
jameerpathan111: Updated the PR

pondrejk (approved):

jameerpathan111:
Added logger.info code in the last push.

ogajduse:
@jameerpathan111 
Since you are moving the settings, please write a [configuration migration](https://github.com/SatelliteQE/robottelo/blob/master/conf/migrations.py). It will make the transition time smoother and less stressful and users will not be required to update their configs.

tests/foreman/installer/test_installer.py
```diff
@@ -593,25 +588,6 @@ def test_positive_check_installer_hammer_ping(target_sat):
             assert 'ok' in line
 
 
-@pytest.mark.upgrade
-@pytest.mark.tier1
-@pytest.mark.build_sanity
-def test_installer_modules_check(target_sat):
```
lpramuk, line 599: This change is completely unrelated to ipv6 and http proxy setups.
Would you please split it into separate PR to make life easier during cherrypicking.

This test lives only in master and 6.16.z so you would definitely hit conflicts in older branches. Thanks.

lpramuk (changes requested):
Split unrelated changes into separate PR.
Otherwise ipv6 and http proxy code **LGTM**.

tests/foreman/installer/test_installer.py
jameerpathan111: I was not planning to cherry-pick it to 6.16.z? Should we?

jameerpathan111:
> @jameerpathan111 Since you are moving the settings, please write a [configuration migration](https://github.com/SatelliteQE/robottelo/blob/master/conf/migrations.py). It will make the transition time smoother and less stressful and users will not be required to update their configs.

added migration

robottelo/hosts.py
```diff
@@ -854,22 +854,39 @@ def put_ssh_key(self, source_key_path, destination_key_name):
...
     def enable_dnf_proxy(self, hostname, scheme=None, port=None):
-        """Configures proxy for dnf"""
+        """Configures HTTP proxy for dnf"""
         if not scheme:
             scheme = 'http'
         cmd = f"echo -e 'proxy = {scheme}://{hostname}"
         if port:
             cmd += f':{port}'
         cmd += "' >> /etc/dnf/dnf.conf"
+        logger.info(f'Configuring {hostname} HTTP proxy for dnf.')
```
ogajduse, line 872:
```suggestion
        logger.info(f'Configured {hostname} HTTP proxy for dnf.')
```

robottelo/hosts.py
```diff
@@ -854,22 +854,39 @@ def put_ssh_key(self, source_key_path, destination_key_name):
             raise CLIFactoryError(f'Failed to chmod ssh key file:\n{result.stderr}')
 
     def enable_rhsm_proxy(self, hostname, port=None):
-        """Configures proxy for subscription manager"""
+        """Configures HTTP proxy for subscription manager"""
         cmd = f"subscription-manager config --server.proxy_hostname={hostname}"
         if port:
             cmd += f' --server.proxy_port={port}'
+        logger.info(f'Configuring {hostname} HTTP proxy for subscription manager.')
```
ogajduse, line 861: Please do either of:
1. Move the logger call before the actual action is performed or
2. Change the wording of the log message

```suggestion
        logger.info(f'Configured {hostname} HTTP proxy for subscription manager.')
```

conf/migrations.py
```diff
@@ -35,3 +35,15 @@ def migration_231129_deploy_workflow(settings, data):
...
+
+def migration_241120_http_proxy_ipv6_url(settings, data):
+    """Migrates server.http_proxy_ipv6_url to http_proxy.http_proxy_ipv6_url"""
+    if (
+        settings.server.get('http_proxy_ipv6_url')
+        and isinstance(settings.server.http_proxy_ipv6_url, str)
+        and not settings.http_proxy.get('http_proxy_ipv6_url')
+    ):
+        data['http_proxy'] = {}
+        data['http_proxy'].http_proxy_ipv6_url = settings.server.http_proxy_ipv6_url
```
ogajduse, line 48: Optional:

I know you copied that from the migration I wrote a couple of months back, but I see a small improvement in terms of readability. Note that `data` is a `Box` object already.

```suggestion
        data.http_proxy = {}
        data.http_proxy.http_proxy_ipv6_url = settings.server.http_proxy_ipv6_url
```

one might wish getting rid of assigning the empty dict as well, but that is needed:

```python
[ins] In [35]: data = Box()

[ins] In [36]: data.http_proxy.http_proxy_ipv6_url = "foo"
KeyError                                  Traceback (most recent call last)
File ~/.virtualenvs/robottelo/lib64/python3.12/site-packages/box/box.py:593, in box.box.Box.__getitem__()

KeyError: 'http_proxy'

The above exception was the direct cause of the following exception:

BoxKeyError                               Traceback (most recent call last)
File ~/.virtualenvs/robottelo/lib64/python3.12/site-packages/box/box.py:632, in box.box.Box.__getattr__()

File ~/.virtualenvs/robottelo/lib64/python3.12/site-packages/box/box.py:620, in box.box.Box.__getitem__()

BoxKeyError: "'http_proxy'"

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
File ~/.virtualenvs/robottelo/lib64/python3.12/site-packages/box/box.py:634, in box.box.Box.__getattr__()

AttributeError: 'Box' object has no attribute 'http_proxy'

The above exception was the direct cause of the following exception:

BoxKeyError                               Traceback (most recent call last)
Cell In[36], line 1
----> 1 data.http_proxy.http_proxy_ipv6_url = "foo"

File ~/.virtualenvs/robottelo/lib64/python3.12/site-packages/box/box.py:648, in box.box.Box.__getattr__()

BoxKeyError: "'Box' object has no attribute 'http_proxy'"

[ins] In [37]: data.http_proxy = {}

[ins] In [38]: data.http_proxy.http_proxy_ipv6_url = "foo"
```

ogajduse (commented):
Two minor comments left.

@jameerpathan111 Please choose one or more tests and run PRT to test your changes.

robottelo/hosts.py
jameerpathan111: Well, it is running before the `self.execute(cmd)` that's why the word `configuring` and not `configured`. The first option you're suggesting.

jameerpathan111:
trigger: test-robottelo
pytest: tests/foreman -k "test_capsule_installation or test_positive_on_demand_sync"
env:
ROBOTTELO_server__is_ipv6: true

conf/migrations.py
jameerpathan111: updated the MR

robottelo/hosts.py
ogajduse: My bad, apologies. Resolving

lpramuk: @ogajduse Do you plan to go fishing ? :fishing_pole_and_fish:

jameerpathan111:
> @jameerpathan111 Please choose one or more tests and run PRT to test your changes.

@ogajduse For some reason PRT didn't add comment for the test result. But the job ran and the 2 tests I ran passed. See build 9408

ogajduse:
I will merge this PR so it is ready before we trigger test automation for the next Sat Stream snap. If anyone wants to question the patch, please do so in a GH issue or initiate a discussion in any of our channels.

pytest_fixtures/component/maintain.py
JacobCallahan: i agree that `satellite` is redundant here, the vast majority of the functionality in this repo acts against a Satellite and this method is defined under a Satellite class. 

how about we really clarify what's going on with the name by changing the name to reflect that this is updating proxy settings to use an ipv6 proxy.

perhaps something like `ensure_ipv6_proxy_settings`

PR #16981, opened 2024-11-21
Add a contenthost_factory to perform post-deploy actions

JacobCallahan:
The intent of this change is to enable us to add post-deploy actions when requesting specific types of content hosts, like fips-enabled. Users can define a set of host_post_configs in conf/content_host.yaml file. They can then add the post config action under the host definition.

## Summary by Sourcery

Introduce a reusable content host factory that centralizes host checkout and optional post-deploy configuration, and update fixtures and upgrade tests to use it while improving network defaults and host cleanup.

New Features:
- Add a contenthost_factory context manager to create and optionally post-configure content hosts based on settings-defined post-config profiles.
- Add a host_post_config helper to run configurable post-deploy broker actions on one or more content hosts.
- Introduce a registered_hosts fixture that provisions and registers multiple content hosts to Satellite based on the current setup.

Enhancements:
- Default content host network configuration to settings.content_host.network_type when not explicitly provided in fixture parameters.
- Refactor existing content host fixtures to rely on the shared contenthost_factory for consistent provisioning behavior, including _count-based multi-host scenarios and Satellite-based custom hosts.
- Adjust CentOS and Oracle host fixtures to derive deploy_network_type from the server IPv4/IPv6 configuration.
- Update the sat_upgrade_chost fixture to use the common contenthost_factory and UBI8 container host configuration instead of directly invoking Broker.

Tests:
- Add explicit cleanup of the post-upgrade RHEL client in the repository upgrade test via a finalizer that checks in the host through Broker.

### DIFF
--- a/pytest_fixtures/core/contenthosts.py
+++ b/pytest_fixtures/core/contenthosts.py
@@ -5,6 +5,8 @@ The functions in this module are read in the pytest_plugins/fixture_markers.py m
 All functions in this module will be treated as fixtures that apply the contenthost mark
 """
 
+from contextlib import contextmanager
+
 from broker import Broker
 import pytest
 
@@ -20,7 +22,7 @@ def host_conf(request):
     if hasattr(request, 'param'):
         params = request.param
     distro = params.get('distro', 'rhel')
-    network = params.get('network')
+    network = params.get('network', settings.content_host.network_type)
     _rhelver = f"{distro}{params.get('rhel_version', settings.content_host.default_rhel_version)}"
 
     # check to see if no-containers is passed as an argument to pytest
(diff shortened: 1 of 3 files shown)


jyejare:
@JacobCallahan I think the approach shown in https://github.com/SatelliteQE/robottelo/pull/16903/files is much more straightforward and has fewer changes needed to make this change today and in future as well.

Though I agree that the post deploy configs are not visible within the module and happen behind the scenes but its not impossible to find those. Hense I sold on the idea of https://github.com/SatelliteQE/robottelo/pull/16903#discussion_r1839969056 , else I too had a similar structure defined before changing my mind.

CC: @ogajduse

pytest_fixtures/core/contenthosts.py
```diff
@@ -41,93 +43,115 @@ def host_conf(request):
...
 
 
+def host_post_config(hosts, config_name):
+    """A function that runs a specified post config on a list of content hosts."""
+    broker_args = settings.content_host.host_post_configs.get(config_name).to_dict()
+    for host in hosts:
+        for key, val in broker_args.items():
+            if "{" in val:
+                broker_args[key] = val.format(host=host)
+        Broker(**broker_args).execute()
```
pcreech, line 53: This is might be a random silly question, but it doesn't seem obvious on the surface to me upon an initial quick read.

My brain got caught up on the "why are we telling broker to take an action here, when it wasn't taking action before"

So in digging further, I'm inferring the following:

With the new approach, we now have actions we have to actually take time to execute.  In this case, either the 'enable-fips' workflow, or the 'enable-fapolicyd' workflow, as part of the 'factory<=>builder" relationship we're introducing.  When we encounter this, we then tell broker to take these specific actions before moving on to the next thing, which in this case would be yeilding the host back up to whatever called the contenthost_factory.

Does that seem generally correct?

JacobCallahan: that's exactly right! it's like making sure the manufacturer puts all the extras on your vehicle before it leaves the factory.

ogajduse (changes requested):
LGTM overall. Two changes need to be made before merging.

ogajduse (approved):

vsedmik (commented):
Looks good to me, just proposed two removals if they don't feel too unrelated to the PR subject.

pytest_fixtures/core/contenthosts.py
```diff
@@ -47,90 +49,136 @@ def host_conf(request):
...
+    with contenthost_factory(request=request) as host:
         yield host
 
 
 @pytest.fixture(scope="class", params=[{'rhel_version': '7'}])
 def rhel7_contenthost_class(request):
     """A fixture for use with unittest classes. Provides a rhel7 Content Host object"""
-    with Broker(**host_conf(request), host_class=ContentHost) as host:
+    with contenthost_factory(request=request) as host:
         yield host
```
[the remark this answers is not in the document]
vsedmik: Hmm, okay. We could omit the extraction from the `6.16.z` auto cherry-pick PR (I can do it if you want me to).
Although I don't expect we would be executing the suite against `6.16.z` anymore, I don't like breaking the `6.16.z` branch by that removal even more.

ogajduse:
@vsedmik, how are you feeling about merging this one?

vsedmik:
> @vsedmik, how are you feeling about merging this one?

Quite comfortable, @ogajduse

vsedmik (approved):

PR #16967, opened 2024-11-20
better logging of fam failures

evgeni:
### Problem Statement

The output of FAM tests is mangled due to how pytest tries to print the big `result` object.
There are no linebreaks and it's super hard to find the failure.

Additionally the double assert is pointless, as it asserts twice the same thing: the result is successful.

### Solution

Stop asserting `PASSED` in stdout (additionally saves the case where there is the string `PASSED` in the output for some reason, but the test actually failed).
Pass a custom error output to the `status == 0` assertion, including stdout, stderr and the status code.

PR #17001, opened 2024-11-26
[6.16.z] better logging of fam failures

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/robottelo/pull/16967

### Problem Statement

The output of FAM tests is mangled due to how pytest tries to print the big `result` object.
There are no linebreaks and it's super hard to find the failure.

Additionally the double assert is pointless, as it asserts twice the same thing: the result is successful.

### Solution

Stop asserting `PASSED` in stdout (additionally saves the case where there is the string `PASSED` in the output for some reason, but the test actually failed).
Pass a custom error output to the `status == 0` assertion, including stdout, stderr and the status code.


Satellite-QE:
**PRT Result**
```
Build Number: 9463
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/sys/test_fam.py::test_positive_import_run_roles --external-logging
Test Result : ================== 1 passed, 9 warnings in 560.04s (0:09:20) ===================
```

pondrejk (approved):

